PR #920, opened 2020-09-18
Temporarily remove fallback codeowners rule

eirikurn:
# Temporarily remove fallback codeowners rule

## What

Changed the core maintainer rule to only focus on root files and the .github directory.

Will re-enable the fallback rule when we've found a way to remove codegen schemas from git. 

## Why

This is to unblock the web team while they're working on the island.is launch.

Any change that touches the graphql schema causes changes in multiple generated schema files. This unneccessarily triggers code reviews for multiple codeowners, which slows down development.

## Checklist:


pshomov (approved):
LGTM

PR #927, opened 2020-09-19
chore: ignore generated files

jeremybarbet:
# Remove all generated files from the repository

- [x] fix CI broken
- [x] move open API generated files to the script as well
- [x] update documentation
- [x] fix ts-node

## What

- Removing all generated files (schema, types definitions, fragment types, open API, **except contentfulTypes.d.ts**) from git and generating them locally on post-install/after pulling changes.

- `contentfulTypes.d.ts` will still be part of the GitHub repository. This [PR](https://github.com/island-is/island.is/pull/1144) removed the need to have a local `CONTENTFUL_MANAGEMENT_ACCESS_TOKEN` to generate the contentful types definitions and is now done on a private build step. Because of the external dependency on contentful, we want to keep the file on Github to avoid breaking our whole build pipeline in case contentful is down for some reason.

I had to do some changes here and there to make all schemas/types automatically _generatable_

**All projects using `autoSchemaFile` on the GraphQLModule:**
- Exported all resolvers to be able to generate the schema file with the script

**All projects using open API:**
- Exported the `DocumentBuilder()` into a separate file to import it for the buildOpenApi script

**air-discount-scheme-backend/application-system-api**
- Added an `INIT_SCHEMA` to ignore Redis initialization when used. Otherwise, the process keeps ending and never stop when building the open API schema

**content-search**
- I moved the dto/enums folder form the sub-level folder to the root folder (libs/api/domains/content-search/src/enums -> libs/api/content-search/src/enums). It was creating circular references when running the build schema script

- You can use `EXCLUDES_SCHEMAS=air-discount-scheme-api,gjafakort-api,skilavottord-ws` if you are not interested in building schema/types for some specific projects
- I also removed the HelloWorld module

## Why

All theses generated files creates a lot of noise in PR reviews and add more code owners on each PR for no real reason

## Screenshots / Gifs

— 

## Checklist:


libs/infra-nest-server/src/lib/execShellCommand.ts
```diff
@@ -0,0 +1,15 @@
+import { logger } from '@island.is/logging'
+import { exec } from 'child_process'
+
+export const execShellCommand = (cmd: string): Promise<string> =>
+  new Promise((resolve, reject) => {
+    exec(cmd, (error, stdout, stderr) => {
+      if (error) {
+        console.log(error)
```
eirikurn, line 8: Should use logger

libs/logging/src/lib/logging.ts
```diff
@@ -18,5 +18,5 @@ if (process.env.NODE_ENV === 'production') {
 export const logger = createLogger({
   level: logLevel,
   format: logFormat,
-  transports: [new transports.Console(), new SentryTransport()],
+  // transports: [new transports.Console(), new SentryTransport()],
```
eirikurn, line 21: What is this for?

libs/infra-nest-server/src/lib/generateSchema.ts
```diff
@@ -0,0 +1,119 @@
...
+    })
+  }
+}
+
+const run = async () => {
+  logger.info('Running generate schema script...')
+  logger.info('Generating graphql schema files...')
+
+  const excludes = GENERATE_EXCLUDES_SCHEMA?.replace(/\s/g, '')?.split(',')
+  console.log('-excludes', excludes)
```
eirikurn, line 69: logger? Need to add eslint error for console.log.

scripts/ts-node.js
```diff
@@ -9,7 +9,13 @@
...
 //      yarn ts-node
 
+require('dotenv').config()
 require('ts-node').register({
-  compilerOptions: { module: 'commonjs', esModuleInterop: true },
+  compilerOptions: {
+    module: 'commonjs',
+    esModuleInterop: true,
+    noUnusedLocals: false,
+    strict: false,
```
eirikurn, line 18: Why are these required? I'd think it should pick up the correct tsconfig of the executing file.

workspace.json
```diff
@@ -345,6 +345,16 @@
             "buildTarget": "api:build"
           }
         },
+        "generate-schema": {
+          "builder": "@nrwl/workspace:run-commands",
+          "options": {
+            "commands": [
+              {
+                "command": "yarn ts-node apps/api/generateSchema.ts"
```
eirikurn, line 353: Does this file exist?

workspace.json
```diff
@@ -3142,6 +3152,34 @@
           }
         }
       }
+    },
+    "generate-schema": {
+      "root": "libs/generate-schema",
```
eirikurn, line 3157: Does this lib exist?

eirikurn (changes requested):
infra-nest-server should only be a consumable library for nest servers and not depend on all other libs (will cause nx circular dependency warnings).

I'd prefer each app/lib to define their own generate-schema and codegen scripts and use nx to run all of them `postinstall`.

Check out how `infra-nest-server/bootstrap` generates an openapi schema based on a CLI argument. We can do the same for graphql apps if they pass in an array of resolvers.

eirikurn:
(sorry I looked. Know this is wip, but wanted to sneak in some implementation suggestions. Reach out if you wanna chat)

jeremybarbet:
> infra-nest-server should only be a consumable library for nest servers and not depend on all other libs (will cause nx circular dependency warnings).
> 
> I'd prefer each app/lib to define their own generate-schema and codegen scripts and use nx to run all of them `postinstall`.
> 
> Check out how `infra-nest-server/bootstrap` generates an openapi schema based on a CLI argument. We can do the same for graphql apps if they pass in an array of resolvers.

Sounds very good, I wasn't sure where to include it yet, it was more to get something working first. Thanks for pointing out the direction. 

The rest is very much wip and messy so most of it will be reverted/deleted once everything is working properly, I will keep you updated

package.json
```diff
@@ -20,6 +20,7 @@
     "affected:lint": "nx affected:lint",
     "affected:dep-graph": "nx affected:dep-graph",
     "affected:codegen": "nx affected --target=codegen",
+    "affected:schemas": "nx affected --target=init-schema --parallel --maxParallel=5 --exclude=$(grep EXCLUDES_SCHEMAS .env | cut -d '=' -f2)",
```
jeremybarbet, line 23: Let's say I'm only working on api/web project, I can do `EXCLUDES_SCHEMAS=air-discount-scheme-api,gjafakort-api,skilavottord-ws`. It saves a few seconds

package.json
```diff
@@ -34,7 +35,9 @@
     "dep-graph": "nx dep-graph",
     "help": "nx help",
     "storybook:island-ui": "nx run island-ui-storybook:storybook",
-    "ts-node": "node -r ./scripts/ts-node"
+    "ts-node": "node -r ./scripts/ts-node",
+    "schemas": "test -f libs/api/schema/src/lib/schema.d.ts || echo 'export default () => {}' >> libs/api/schema/src/lib/schema.d.ts && yarn nx run api:contentful-types && yarn affected:schemas",
```
jeremybarbet, line 39: That's a bit ugly but the thing is `libs/api/schema/src/lib/schema.d.ts` is ignored, even though, without it, we can't build the API, because the file is required for some [import](https://github.com/island-is/island.is/blob/master/libs/api/schema/src/index.ts#L1). So if the file doesn't exist I created it with an empty function, and it gets overwritten once the codegen is done.

I will tackle that issue in another PR to remove this file and move all schema locally to each apps, so we won't depend on shared schemas for the apps. It already has been done for the [web project](https://github.com/island-is/island.is/blob/master/apps/web/codegen.yml)

.github/workflows/build.yml
```diff
@@ -152,6 +152,22 @@ jobs:
...
         run: echo "::set-env name=BRANCH::${GIT_BRANCH}"
 
+      - name: Cache for generated files
+        id: generated-files-cache
+        uses: actions/cache@v2
+        with:
+          path: generated_files.tar.gz
+          key: ${{ runner.os }}-${{ hashFiles('libs/api/domains/cms/src/lib/generated/contentfulTypes.d.ts', 'apps/**/models/**/*.tsx?', 'libs/**/models/**/*.tsx?', 'apps/**/dto/**/*.tsx?', 'libs/**/dto/**/*.tsx?', 'apps/**/enums/**/*.tsx?', 'libs/**/enums/**/*.tsx?', 'apps/**/queries/**/*.tsx?', 'libs/**/queries/**/*.tsx?', 'apps/air-discount-scheme/web/screens/**/*.tsx', 'apps/air-discount-scheme/web/i18n/withLocale.tsx', 'apps/air-discount-scheme/web/components/AppLayout/AppLayout.tsx', 'apps/air-discount-scheme/web/components/Header/Header.tsx', 'apps/gjafakort/api/src/**/typeDefs.ts') }}-generated-files
+
+      - name: Debug hashkey
```
sindrig, line 162: Remove debug

.github/workflows/build.yml
```diff
@@ -232,13 +248,28 @@ jobs:
...
           path: node_modules
           key: ${{ runner.os }}-${{ hashFiles('yarn.lock') }}-yarn
 
+      - name: Cache for generated files
+        uses: actions/cache@v2
+        with:
+          path: generated_files.tar.gz
+          key: ${{ runner.os }}-${{ hashFiles('libs/api/domains/cms/src/lib/generated/contentfulTypes.d.ts', 'apps/**/models/**/*.tsx?', 'libs/**/models/**/*.tsx?', 'apps/**/dto/**/*.tsx?', 'libs/**/dto/**/*.tsx?', 'apps/**/enums/**/*.tsx?', 'libs/**/enums/**/*.tsx?', 'apps/**/queries/**/*.tsx?', 'libs/**/queries/**/*.tsx?', 'apps/air-discount-scheme/web/screens/**/*.tsx', 'apps/air-discount-scheme/web/i18n/withLocale.tsx', 'apps/air-discount-scheme/web/components/AppLayout/AppLayout.tsx', 'apps/air-discount-scheme/web/components/Header/Header.tsx', 'apps/gjafakort/api/src/**/typeDefs.ts') }}-generated-files
+
+      - name: Debug hashkey
```
sindrig, line 265: ditto

.github/workflows/build.yml
```diff
@@ -292,6 +324,20 @@ jobs:
...
```
sindrig, line 333: ditto

.github/workflows/build.yml
```diff
@@ -311,13 +357,29 @@ jobs:
...
       - uses: actions/checkout@v2
         if: ${{ matrix.projects != 'dummy' }}
+
+      - name: Cache for generated files
+        uses: actions/cache@v2
+        with:
+          path: generated_files.tar.gz
+          key: ${{ runner.os }}-${{ hashFiles('libs/api/domains/cms/src/lib/generated/contentfulTypes.d.ts', 'apps/**/models/**/*.tsx?', 'libs/**/models/**/*.tsx?', 'apps/**/dto/**/*.tsx?', 'libs/**/dto/**/*.tsx?', 'apps/**/enums/**/*.tsx?', 'libs/**/enums/**/*.tsx?', 'apps/**/queries/**/*.tsx?', 'libs/**/queries/**/*.tsx?', 'apps/air-discount-scheme/web/screens/**/*.tsx', 'apps/air-discount-scheme/web/i18n/withLocale.tsx', 'apps/air-discount-scheme/web/components/AppLayout/AppLayout.tsx', 'apps/air-discount-scheme/web/components/Header/Header.tsx', 'apps/gjafakort/api/src/**/typeDefs.ts') }}-generated-files
+
+      - name: Debug hashkey
```
sindrig, line 367: ditto

.github/workflows/build.yml
```diff
@@ -358,13 +420,29 @@ jobs:
...
```
sindrig, line 430: ditto

.github/workflows/build.yml
```diff
@@ -405,13 +483,29 @@ jobs:
...
```
sindrig, line 493: ditto

.gitignore
```diff
@@ -50,3 +50,22 @@ helm/**/*.tgz
...
+apps/gjafakort/web/graphql/possibleTypes.json
+apps/air-discount-scheme/api.graphql
+apps/air-discount-scheme/web/graphql/schema.tsx
+apps/air-discount-scheme/web/graphql/possibleTypes.json
+apps/air-discount-scheme/backend/src/openapi.yaml
+apps/skilavottord/ws/src/app/api.graphql
+apps/judicial-system/api/src/openapi.yaml
+apps/application-system/api/src/openapi.yaml
+apps/reference-backend/src/openapi.yaml
+# libs/api/domains/cms/src/lib/generated/contentfulTypes.d.ts
```
sindrig, line 70: confusing to keep this here, remove?

package.json
```diff
@@ -20,6 +20,7 @@
     "affected:lint": "nx affected:lint",
     "affected:dep-graph": "nx affected:dep-graph",
     "affected:codegen": "nx affected --target=codegen",
+    "affected:schemas": "nx affected --target=init-schema --parallel --maxParallel=5 $(test -f .env && echo \"--exclude=$(grep EXCLUDES_SCHEMAS .env | cut -d '=' -f2)\")",
```
sindrig, line 23: When and how is the `EXCLUDE_SCHEMAS` used?

.github/workflows/build.yml
```diff
@@ -405,13 +483,29 @@ jobs:
     steps:
       - uses: actions/checkout@v2
         if: ${{ matrix.projects != 'dummy' }}
+
+      - name: Cache for generated files
+        uses: actions/cache@v2
+        with:
+          path: generated_files.tar.gz
+          key: ${{ runner.os }}-${{ hashFiles('libs/api/domains/cms/src/lib/generated/contentfulTypes.d.ts', 'apps/**/models/**/*.tsx?', 'libs/**/models/**/*.tsx?', 'apps/**/dto/**/*.tsx?', 'libs/**/dto/**/*.tsx?', 'apps/**/enums/**/*.tsx?', 'libs/**/enums/**/*.tsx?', 'apps/**/queries/**/*.tsx?', 'libs/**/queries/**/*.tsx?', 'apps/air-discount-scheme/web/screens/**/*.tsx', 'apps/air-discount-scheme/web/i18n/withLocale.tsx', 'apps/air-discount-scheme/web/components/AppLayout/AppLayout.tsx', 'apps/air-discount-scheme/web/components/Header/Header.tsx', 'apps/gjafakort/api/src/**/typeDefs.ts') }}-generated-files
```
darrikonn, line 491: Would like to see a variable for this, which can be reused for all the keys. Something like:
```yaml
cache_key: &cache_key "hashFiles('libs/api/domains/cms/src/lib/generated/contentfulTypes.d.ts', 'apps/**/models/**/*.tsx?', 'libs/**/models/**/*.tsx?', 'apps/**/dto/**/*.tsx?', 'libs/**/dto/**/*.tsx?', 'apps/**/enums/**/*.tsx?', 'libs/**/enums/**/*.tsx?', 'apps/**/queries/**/*.tsx?', 'libs/**/queries/**/*.tsx?', 'apps/air-discount-scheme/web/screens/**/*.tsx', 'apps/air-discount-scheme/web/i18n/withLocale.tsx', 'apps/air-discount-scheme/web/components/AppLayout/AppLayout.tsx', 'apps/air-discount-scheme/web/components/Header/Header.tsx', 'apps/gjafakort/api/src/**/typeDefs.ts')"
```

then it can be referenced with:
```yaml
key: ${{ runner.os }}-${{ *cache_key }}-generated-files
```
(something on these lines)

darrikonn: Will developers have to edit this variable when they create a new project?

.gitignore
```diff
@@ -50,3 +50,22 @@ helm/**/*.tgz
...
+libs/api/schema/src/lib/schema.d.ts
```
darrikonn, line 71: Can you change this to a wildcard.

package.json
```diff
@@ -34,7 +35,9 @@
     "dep-graph": "nx dep-graph",
     "help": "nx help",
     "storybook:island-ui": "nx run island-ui-storybook:storybook",
-    "ts-node": "node -r ./scripts/ts-node"
+    "ts-node": "ts-node -r tsconfig-paths/register -O '{\"module\":\"commonjs\",\"esModuleInterop\":true}'",
+    "schemas": "test -f libs/api/schema/src/lib/schema.d.ts || echo 'export default () => {}' >> libs/api/schema/src/lib/schema.d.ts && yarn nx run api:contentful-types && yarn affected:schemas",
+    "postinstall": "if [[ ! -z $DOCKER_BUILD ]]; then exit 0; else yarn schemas; fi"
```
sindrig, line 40: Minor: This will break if DOCKER_BUILD contains spaces or is an empty string. Consider `if [ ! -z ${DOCKER_BUILD+x} ]`

scripts/clean-generated-files.sh
```diff
@@ -0,0 +1,21 @@
...
+rm apps/gjafakort/web/graphql/possibleTypes.json
+rm apps/air-discount-scheme/api.graphql
+rm apps/air-discount-scheme/web/graphql/schema.tsx
+rm apps/air-discount-scheme/web/graphql/possibleTypes.json
+rm apps/air-discount-scheme/backend/src/openapi.yaml
+rm apps/skilavottord/ws/src/app/api.graphql
+rm apps/judicial-system/api/src/openapi.yaml
+rm apps/application-system/api/src/openapi.yaml
+rm apps/reference-backend/src/openapi.yaml
+# rm libs/api/domains/cms/src/lib/generated/contentfulTypes.d.ts
```
sindrig, line 20: ditto - remove commented line?

sindrig (changes requested):
Minor changes but a bit of cleanup required :clap:

scripts/clean-generated-files.sh
```diff
@@ -0,0 +1,21 @@
...
+rm libs/api/schema/src/lib/schema.d.ts
```
darrikonn, line 21: Same with wildcard.
Trying to avoid other developers having to edit this file.

darrikonn (changes requested):

package.json
jeremybarbet: It's only for local dev

Part of the update on the documentation I wrote:

"However, if you are only focusing on one project (e.g. the api) you can opt-out for others projects to generate their respective files with using an environnement variable into your .env file e.g. `EXCLUDES_SCHEMAS=air-discount-scheme-api,gjafakort-api,skilavottord-ws`"

.github/workflows/build.yml
jeremybarbet: That's a good question and this is the weakest part of the script.

I added the folders (the one that I saw) that trigger the scripts to generate new files (autoSchemeFile/openApi etc). Most of the projects have common folders where it happens e.g. `queries/*`, `models/*`, etc but it's not always the case.

If a new project is created and follows the most common way to create queries/models etc, it will be okay. However, if some files at some specific paths must trigger newly generated files and it's not part of the wildcard it will fail to generate the new schemas.

An alternative would be to never cache theses files on the build steps with the hash key and generate them all the time, but it will increase the build time of ~30sec * number of steps needed

cc/ @eirikurn @sindrig

apps/application-system/api/src/app/modules/application/application.module.ts
```diff
@@ -1,25 +1,34 @@
...
 import { Application } from './application.model'
 import { ApplicationController } from './application.controller'
 import { ApplicationService } from './application.service'
-import { FileStorageModule } from '@island.is/file-storage'
 import { UploadProcessor } from './upload.processor'
 
+let BullModule: DynamicModule
+
+if (process.env.INIT_SCHEMA === 'true') {
+  BullModule = NestBullModule.registerQueueAsync()
```
eirikurn, line 14: Probably needs to have the same name (upload). I think it's better to put this kind of stuff inside the async useFactory (possibly inline with the @Module declaration like it was):

`
useFactory: () => (inMemory ? {} : { redis: ... }),
`

apps/air-discount-scheme/backend/src/app/modules/cache/cache.module.ts
```diff
@@ -4,9 +4,9 @@ import { createNestJSCache } from '@island.is/cache'
...
 
 const { redis, production } = environment
-
 let CacheModule: DynamicModule
-if (process.env.NODE_ENV === 'test') {
+
+if (process.env.NODE_ENV === 'test' || process.env.INIT_SCHEMA === 'true') {
   CacheModule = NestCacheModule.register()
 } else {
   CacheModule = NestCacheModule.register({
```
eirikurn, line 12: I think `registerAsync` is intended to handle dynamic configuration. cc @darrikonn

apps/judicial-system/api/src/buildOpenApi.ts
```diff
@@ -0,0 +1,11 @@
+import { buildOpenApi } from '@island.is/infra-nest-server'
+
+import { AppModule } from './app/app.module'
+import { openApi } from './openApi'
+
+// TODO - redis
```
eirikurn, line 6: Still valid?

package.json
eirikurn: This assumes people are using dotenv, but it does not work for normal environment variables, nor for direnv (which some are pushing for).

I think we need to pick one or the other soon.

package.json
```diff
@@ -34,7 +35,9 @@
     "dep-graph": "nx dep-graph",
     "help": "nx help",
     "storybook:island-ui": "nx run island-ui-storybook:storybook",
-    "ts-node": "node -r ./scripts/ts-node"
+    "ts-node": "ts-node -r tsconfig-paths/register -O '{\"module\":\"commonjs\",\"esModuleInterop\":true}'",
```
eirikurn, line 38: I actually want the `scripts/ts-node` file so I can later require it from js files (like `next.config.js`) and then import shaerd typescript code.

Did you have issues with it?

.github/workflows/build.yml
sindrig: surprise surprise I keep forgetting about yaml variables. we should look into creating a parameter node in the build's root for this.

as for if developers will have to edit this variable. like @jeremybarbet says, if the most common way to create queries/models is used, there will be no problem. and if a developer uses some obscure way, the build will fail and he/she will be faced with the decision to either update this variable (and the .gitignore and removal script probably) or simply use the common way.

package.json
eirikurn: It's common to check for a CI variable. Does that make sense here?

scripts/get-files-touched-by.sh
```diff
@@ -0,0 +1,11 @@
...
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )/.." >/dev/null 2>&1 && pwd )"
+marker=$(mktemp)
+touch $marker -r ${BASH_SOURCE[0]}
+
+CMD=$1
+shift
+
+$CMD "$@" 1>&2
+
+find $DIR/apps $DIR/libs -type f -newer $marker
```
eirikurn, line 11: Magic! 😮 

Maybe a bash comment is appropriate :)

workspace.json
```diff
@@ -313,6 +313,16 @@
...
           }
         },
+        "init-schema": {
+          "builder": "@nrwl/workspace:run-commands",
+          "options": {
+            "commands": [
+              {
+                "command": "yarn nx run gjafakort-api:codegen"
+              }
+            ]
```
eirikurn, line 323: All of these `run-commands` can be simplified after you merge from master.

workspace.json
```diff
@@ -345,6 +355,23 @@
...
+            "commands": [
+              {
+                "command": "yarn ts-node -P apps/api/tsconfig.app.json apps/api/src/buildSchema.ts"
+              },
+              {
+                "command": "yarn nx run api:codegen"
+              },
+              {
+                "command": "yarn nx run web:codegen"
+              }
```
eirikurn, line 371: Can web have its own `init-schema`? Since we'll end up with a lot of clients to this api.

workspace.json
```diff
@@ -814,6 +810,18 @@
             "cwd": "apps/reference-backend"
           }
         },
+        "init-schema": {
+          "builder": "@nrwl/workspace:run-commands",
+          "options": {
+            "parallel": false,
```
eirikurn, line 816: Parallel not needed.

eirikurn (dismissed):
Awesome job. Minor comments

package.json
sindrig: The original `CI` variable isn't available inside the docker build. But I wonder if we should name it `CI` for consistency's sake?

package.json
sindrig: Agreed - can we skip the file check and simply use the environment variable `EXCLUDE_SCHEMAS`? Then the question on dotenv vs direnv can be postponed (or at least not tied to this usecase).

apps/application-system/api/src/app/modules/application/application.module.ts
jeremybarbet: I don't know why, but with this syntax, the process still keeps running and doesn't stop

package.json
jeremybarbet: Oh my bad, it wasn't used so I thought I could overwrite it. 

Yes I couldn't pass down the arguments for ts-node e.g. `ts-node -P tsconfig.json file.ts`, probably because it's using `node` instead of `ts-node`.

I can reintroduce another `ts-node-whatever` script specifically for the workspace usage

What do you think?

workspace.json
jeremybarbet: Yes I saw that, hopefully, won't get too many conflicts 😬

package.json
jeremybarbet: So basically doing:

```suggestion
    "affected:schemas": "nx affected --target=init-schema --parallel --maxParallel=5 --exclude=$EXCLUDES_SCHEMAS",
```

but it won't pick up the exclude from the .env file and/or direnv (I don't know how direnv works)?

package.json
jeremybarbet: I renamed it to CI

apps/air-discount-scheme/backend/src/app/modules/cache/cache.module.ts
darrikonn: Ahh ok 👍

package.json
sindrig: yes. direnv is basically a shellscript where you can export certain environment variables when you enter a directory (or do whatever you want, when you enter a directory)

package.json
eirikurn: Ah, I see. Makes sense.

Just undelete it for now. No need to create another script in package.json.

apps/application-system/api/src/app/modules/application/application.module.ts
eirikurn: Ah ... Maybe because of the name parameter, triggering the consumer to start waiting for upload jobs.

package.json
jeremybarbet: Looks nice. So it's probably better I also remove the section in the document I added about adding this variable inside the .env file until we decided a way to handle these env variables?

apps/air-discount-scheme/backend/src/app/modules/cache/cache.module.ts
jeremybarbet: I shouldn't change it on this PR right?

package.json
sindrig: Agreed

apps/application-system/api/src/app/modules/application/application.module.ts
jeremybarbet: Yes adding the name, trigger the upload job.

Should I keep it this way?

package.json
jeremybarbet: Not sure to understand your comment _Just undelete it for now_. Did you mean _delete_?

.github/workflows/build.yml
sindrig: Unfortunately, GitHub actions don't support anchors :scream_cat: 
https://github.com/island-is/island.is/issues/1117

.gitignore
jeremybarbet: I added the changes for both `gitignore` and `clean-generated-files`. However, if someone decides to rename a generated file into `custom-name.d.ts` it will need to be added to the list but it shouldn't be the case

scripts/get-files-touched-by.sh
```diff
@@ -0,0 +1,12 @@
...
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )/.." >/dev/null 2>&1 && pwd )"
+marker=$(mktemp)
+touch $marker -r ${BASH_SOURCE[0]}
+
+CMD=$1
+shift
+
+$CMD "$@" 1>&2
+
+# Find all the files paths that are generated out of the yarn schemas script
```
sindrig, line 11: This file doesn't really have anything to do with `yarn schemas`, although only used for it now. What about `# Find all files that were generated by running the command in argv`?

sindrig (approved):
LGTM

scripts/get-files-touched-by.sh
jeremybarbet: You are right I took a shortcut since the script is only used along with yarn schemas but it could be used for other scripts. I will change that, thanks

sindrig (approved):
:clap:

package.json
eirikurn: Yeah, I meant to restore the old scripts/ts-node.js file and leave it even though it's not used anymore in `yarn ts-node`.

apps/application-system/api/src/app/modules/application/application.module.ts
eirikurn: Sure, until we find a better way...

apps/air-discount-scheme/backend/src/app/modules/cache/cache.module.ts
eirikurn: No worries.

workspace.json
```diff
@@ -319,6 +331,18 @@
...
+          "options": {
+            "command": "yarn ts-node -P apps/api/tsconfig.app.json apps/api/src/buildSchema.ts"
+          }
+        },
+        "init-schema": {
+          "builder": "@nrwl/workspace:run-commands",
+          "options": {
+            "command": "yarn nx run api:codegen"
+          }
+        },
```
eirikurn, line 345: What's the difference between `build-schema`, `init-schema` and `codegen` (for developer workflow)?

eirikurn (approved):
Looks great. Just one clarification question.

workspace.json
jeremybarbet: - `build-schema` is used on the graphql module where autoSchemaFile is defined to generate the `api.graphql` file
- `codegen` is used in projects where a codegen.yml file exists
- `init-schema` is the common script used with the nx affected. For nestjs project it will run in general `build-schema`, for web projects it will run `codegen` scripts and for open api projects it will run `buildOpenApi`.

jeremybarbet: Actually, `build-schema` is only used once for the API because I run it before all the affected `init-schema`. 

Otherwise, `init-schema` run either or together `buildSchema.ts`/`codegen`/`buildOpenApi.ts` if the project is using Nest modules, web stack or open api.

eirikurn (approved):

sindrig (approved):

darrikonn (approved):

darrikonn (approved):

eirikurn (dismissed):

obmagnusson (approved):
LGTM

sindrig (dismissed):

sindrig (approved):

sindrig (approved):

eirikurn (dismissed):

haukurh (dismissed):
Aside from one minor comment and the fact that I miss the HelloWorld module already...
This looks good to me

eirikurn (approved):

haukurh (dismissed):

gudjong (dismissed):

eirikurn (dismissed):

haukurh (dismissed):

PR #948, opened 2020-09-21
Ads/Csv for admin

dabbeg:
# CSV for admin
## What

I am activating the csv button to trigger a CSV download with the data that
is currently being filtered.

### DIFF
--- a/apps/air-discount-scheme/web/screens/Admin/Admin.tsx
+++ b/apps/air-discount-scheme/web/screens/Admin/Admin.tsx
@@ -2,6 +2,7 @@ import React, { useState, useContext } from 'react'
 import { useQuery } from '@apollo/client'
 import gql from 'graphql-tag'
 import { SubmitHandler } from 'react-hook-form'
+import CSVStringify from 'csv-stringify'
 
 import { NotFound } from '@island.is/air-discount-scheme-web/screens'
 import { Airlines } from '@island.is/air-discount-scheme/consts'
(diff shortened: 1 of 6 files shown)


apps/air-discount-scheme/web/screens/Admin/Admin.tsx
```diff
@@ -92,6 +93,48 @@ const Admin: Screen = ({}) => {
     setFilters(data)
   }
 
+  const downloadCSV = () => {
```
darrikonn, line 96: Would like to see this function in e.g.:
```typescript
import { downloadCSV } from './utils'
```
Where utils is in:
`apps/air-discount-scheme/web/screens/Admin/utils.ts`

apps/air-discount-scheme/web/screens/Admin/Admin.tsx
```diff
@@ -92,6 +93,48 @@ const Admin: Screen = ({}) => {
...
+    Object.keys(Airlines)
+      .filter(
+        (airline) => !filters.airline || filters.airline?.value === airline,
```
darrikonn, line 99: This has to be changed due to the new cooperation field. See the `getFlightLegs` function in `./components/Summary/Summary.tsx` + the `airline` object.

They should be moved to `./utils.ts` as well and reused.

darrikonn (changes requested):
My main concern is the UX when the filters are dirty and have not yet been applied.
Might be ok, but we should probably either disable the button due to dirty state, or specify in a descriptive text that the user needs to apply the filter, since it affects the downloaded report.

darrikonn (dismissed):

darrikonn (approved):

PR #963, opened 2020-09-21
Always serve index.html as fallback content

### LINKED ISSUE #962: React static apps don't get to route url routes

pshomov:
# React router gets to route

Fixes #962 

## What

See issue

## Why

The React-based static apps don't work.

## Checklist:

### DIFF
--- a/scripts/dockerfile-config/static-nginx.conf
+++ b/scripts/dockerfile-config/static-nginx.conf
@@ -6,6 +6,7 @@ server {
     location / {
         root   /usr/share/nginx/html;
         index  index.html index.htm;
+        # The following is necessary for React router based apps
         try_files $uri $uri/ /index.html =404;
     }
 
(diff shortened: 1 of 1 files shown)


scripts/dockerfile-config/static-nginx.conf
```diff
@@ -6,6 +6,8 @@ server {
     location / {
         root   /usr/share/nginx/html;
         index  index.html index.htm;
+        # The following is necessary for React router based apps
+        try_files $uri $uri/ /index.html =404;
```
gudjong, line 10: So, is this only telling you which file to load? So, /blabla would still be /blabla, but just resolved by index.html?

sindrig (approved):

scripts/dockerfile-config/static-nginx.conf
pshomov: Correct

pshomov: But existing assets would still be resolved from the file system.

sindrig (approved):

PR #978, opened 2020-09-22
Some updates/fixes to the about page

maggisk:
# \<Description\>

## What

* Change the "scroll-spy" to an updated implementation
* Remove "PageHeader" from the slice union type

## Why

* Old scroll-spy was slow and buggy
* PageHeader being a slice type was causing issues without any real benefit

## Checklist:

### DIFF
--- a/apps/api/src/api.graphql
+++ b/apps/api/src/api.graphql
@@ -16,11 +16,6 @@ type Image {
   height: Int!
 }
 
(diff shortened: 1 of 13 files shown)


barabrian (approved):
lgtm, but I could not spot the difference why this scroll spy should be smoother?

PR #982, opened 2020-09-22
Extend Case

oddsson:
# Adds an field to Case, courtCaseNumber

## What

See description

## Why

To be able to save a case with a court case number

## Checklist:

### DIFF
--- a/apps/judicial-system/api/migrations/20200922125905-update-case.js
+++ b/apps/judicial-system/api/migrations/20200922125905-update-case.js
@@ -5,8 +5,7 @@ module.exports = {
     return Promise.all([
       queryInterface.addColumn('case', 'court_case_number', {
         type: Sequelize.STRING,
-        allowNull: false,
-        defaultValue: 'MISSING',
+        allowNull: true,
       }),
     ])
   },
--- a/apps/judicial-system/api/migrations/20200922125905-update-case.js
+++ b/apps/judicial-system/api/migrations/20200922125905-update-case.js
@@ -1,0 +1,19 @@
+'use strict'
+
+module.exports = {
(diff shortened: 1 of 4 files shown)


apps/judicial-system/api/src/app/modules/case/models/case.model.ts
```diff
@@ -170,6 +170,14 @@ export class Case extends Model<Case> {
   // Athugasemdir til dómara
   comments: string
 
+  @Column({
+    type: DataType.STRING,
+    allowNull: false,
```
gudjong, line 175: Allow null ætti að vera true

apps/judicial-system/api/src/app/modules/case/models/case.model.ts
```diff
@@ -170,6 +170,14 @@ export class Case extends Model<Case> {
...
+  })
+  @ApiProperty()
```
gudjong, line 177: Nota ApiPropertyOptional

apps/judicial-system/api/migrations/20200922125905-update-case.js
```diff
@@ -0,0 +1,19 @@
+'use strict'
+
+module.exports = {
+  up: (queryInterface, Sequelize) => {
+    return Promise.all([
+      queryInterface.addColumn('case', 'court_case_number', {
+        type: Sequelize.STRING,
+        allowNull: false,
+        defaultValue: 'MISSING',
```
gudjong, line 9: allowNull true og ekkert default value

gudjong (changes requested):

gudjong (approved):

PR #984, opened 2020-09-22
Web news list from elasticsearch

stefanhar:
# News list from elasticsearch

## What

- Made web get news list from elasticsearch
- Made searc content type resolver consistent with code in cms slice resolver

## Why

- To ensure content exists when clicked on web since single news is loaded from search engine
- Altered code to make resolver logic more consistent

## Screenshots / Gifs

Screenshot 2020-09-22 at 13 29 14

## Checklist:


libs/api/domains/cms/src/lib/models/news.model.ts
```diff
@@ -32,12 +35,13 @@ export class News {
 }
 
 export const mapNews = ({ fields, sys }: INews): News => ({
+  typename: 'News',
   id: sys.id,
   slug: fields?.slug ?? '',
   title: fields?.title ?? '',
   subtitle: fields?.subtitle ?? '',
   intro: fields?.intro ?? '',
   image: mapImage(fields.image),
```
jeremybarbet, line 44:
```suggestion
  image: fields.image?.fields?.file ? mapImage(fields.image) : null,
```

jeremybarbet (approved):

libs/api/domains/cms/src/lib/models/news.model.ts
haukurh: ```mapImage(fields.image ?? '')``` 
works fine as well

stefanhar: `mapImage` is quite bulletproof atm.
We can assume fields is allways set so, worst case is fields.image will pass undefined which is handled inside mapImage.
We are providing empty image object instead of null atm since null is probably not handled in frontend.
It makes sense to pass null once that has been handled, but we will tackle that as a separate issue

maggisk (approved):

PR #1000, opened 2020-09-22
Adds integration with document signing service from dokobit

gudjong:
# Wraps Dokobit's document signing API

## What

A service for signing PDF documents using Dokobit's API. The service can be extended later to include other types of documents and/or other capabilities supported by Dokobit's API.

## Why

This service will first be used to sign verdicts in detention request cases.

## Checklist:

### DIFF
--- a/libs/dokobit-signing/src/lib/signing.service.spec.ts
+++ b/libs/dokobit-signing/src/lib/signing.service.spec.ts
@@ -22,7 +22,7 @@ const testDocumentContent = 'Test Document Content'
 const testSignPath = `sign.json?access_token=${testOptions.accessToken}`
 const testSignResponse = {
   status: 'ok',
-  control_code: 'Test Control Code',
+  control_code: 'Test Control Code', // eslint-disable-line @typescript-eslint/camelcase
   token: 'Test Document Token',
 }
 const testStatusPath = `sign/status/${testSignResponse.token}.json?access_token=${testOptions.accessToken}`
@@ -39,8 +39,8 @@ const testStatusResponse = {
 }
 const postMock = jest.fn(function(
   path: string,
-  body?: Body,
-  init?: RequestInit,
+  body?: Body, // eslint-disable-line @typescript-eslint/no-unused-vars
+  init?: RequestInit, // eslint-disable-line @typescript-eslint/no-unused-vars
 ) {
   switch (path) {
     case testSignPath:
--- a/libs/dokobit-signing/src/lib/signing.service.spec.ts
+++ b/libs/dokobit-signing/src/lib/signing.service.spec.ts
@@ -1,0 +1,133 @@
+import { mock } from 'jest-mock-extended'
+import { Base64 } from 'js-base64'
+import { createHash } from 'crypto'
+
+import { Test, TestingModule } from '@nestjs/testing'
+
+import { Logger, LOGGER_PROVIDER } from '@island.is/logging'
+
+import { SigningService } from './signing.service'
+
+const testOptions = {
+  url: 'Test Url',
+  accessToken: 'Test Access Token',
+}
+const testMobileNumber = '1111111'
+const testMessage = 'Test Message'
+const testContact = 'Test Contact'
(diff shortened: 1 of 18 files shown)


libs/dokobit-signing/src/lib/signing.service.spec.ts
```diff
@@ -0,0 +1,133 @@
...
+    name: testDocumentName,
+    digest: createHash('sha256')
+      .update(testSignedDocumentContent)
+      .digest('hex'),
+    content: testSignedDocumentContent,
+  },
+}
+const postMock = jest.fn(function(
+  path: string,
+  body?: Body, // eslint-disable-line @typescript-eslint/no-unused-vars
```
oddsson, line 42: Why include these parameters if we don't use them?

libs/dokobit-signing/src/lib/signing.service.ts
```diff
@@ -0,0 +1,159 @@
...
+    )
+
+    const base64 = Base64.encode(documentContent)
+    const digest = createHash('sha256')
+      .update(base64)
+      .digest('hex')
+
+    const body = new FormData()
+    body.append('phone', mobileNumber)
+    body.append('message', `${message} `)
```
oddsson, line 83: Please consider replacing 

```
 body.append('message', `${message} `)
```

with
 
```
 body.append('message', message)
```

for code readability

libs/dokobit-signing/src/lib/signing.service.ts
```diff
@@ -0,0 +1,159 @@
...
+
+  async getDocument(
+    documentName: string,
+    documentToken: string,
+  ): Promise<string> {
+    let resStatus: DokobitStatusResponse
+
+    // Try to retrieve the signed document
+    // Need to try longer than the mobile signature timeout, but not too long
+    for (let i = 1; i < 120; i++) {
```
oddsson, line 116: This seems like a code smell. Why do we need this  if we are awaiting `this.post`

libs/nova-sms/src/lib/sms.service.spec.ts
```diff
@@ -12,8 +12,8 @@ const testSendSms = 'SendSms'
 const testCode = '0'
 const postMock = jest.fn(function(
   path: string,
-  body?: Body,
-  init?: RequestInit,
+  body?: Body, // eslint-disable-line @typescript-eslint/no-unused-vars
```
oddsson, line 15: Same question as above ☝️

oddsson (changes requested):

libs/dokobit-signing/src/lib/signing.service.spec.ts
gudjong: This is a mock of a function and the mock does not grab the call if I exclude the arguments. Later, I match calls to the mocked function.

libs/dokobit-signing/src/lib/signing.service.ts
gudjong: I need the extra space at the end. A four digit security code is added to the end of this message on the mobile phone. Without the space, the display will be kind of ugly.

libs/nova-sms/src/lib/sms.service.spec.ts
gudjong: Same answer.

oddsson: Same 👍

libs/dokobit-signing/src/lib/signing.service.ts
gudjong: I agree, but this is how the Dokobit API works and how the code examples from Dokobit do this.

The API returns pretty much immediately, whether the document has been signed or not. At the same time, the user gets a long time to respond to the signature request on the mobile phone.

The result is that we need to try repeatedly until we get an answer other than ´waiting´. I don't know of a good way to do this differently.

oddsson (approved):

Issue #253, opened 2020-07-02
Switch to stricter TypeScript config


eirikurn:
The default TypeScript config from NX does not have any strict flags set, eg it allows implicit any. This makes it easy to forget to add type definitions to expressions and declarations and end up with unsafe untyped code.

It&#39;s an easy change in tsconfig.json, however, we already have a lot of code this will break.

To start this migration we should turn off strict in the root tsconfig. For projects that have a significant number of errors with this flag set, we can unset it in the project-specific tsconfig. Later, someone (optimally from the project team) has time to fix these type issues and re-enable strict mode.

dabbeg:
So this is the current status of the strict setup even though me and @arnifreyrs took removed a lot of errors:
```
Errors Project
378 libs/api/domains/cms
378 apps/web
179 apps/judicial-system/api
174 apps/adgerdir
167 apps/air-discount-scheme/web
143 apps/gjafakort/web
111 apps/gjafakort/api
69  libs/island-ui/contentful
66  apps/air-discount-scheme/api
65  apps/judicial-system/web
54  libs/api/domains/content-search
41  apps/skilavottord/web
32  apps/application-system/api
29  libs/api/domains/application
28  apps/services/search-indexer
27  libs/api/domains/documents
18  libs/api/content-search
18  apps/gjafakort/application
11  libs/contentful-importer
10  apps/reference-next-app
10  apps/gjafakort/queue-listener
7   apps/skilavottord/ws
7   apps/reference-backend
6   libs/message-queue
5   libs/dokobit-signing
4   libs/infra-express-server
3   libs/island-ui/core
3   libs/elastic-indexing
3   apps/application-system/form
2   libs/service-portal/graphql
2   libs/service-portal/family
2   libs/api/domains/translations
1   libs/service-portal/documents
1   apps/api
```

Can we delegate fixing these type errors and removing `"strict": false` out of tsconfig.json to the codeowners? @eirikurn

Fixing the libs should be the highest priority because new projects with strict enabled cannot use a lib that has ts errors. Even though you explicitly disable strict in the lib.

dabbeg:
If we split this up by CODEOWNERS we have something like:

```
kosmos-kaos
174 apps/adgerdir

kosmos-kaos og aranja
378 libs/api/domains/cms
378 apps/web

aranja
29  libs/api/domains/application
32  apps/application-system/api
3   apps/application-system/form

vice-versa
66  apps/air-discount-scheme/api
[...]

PR #1008, opened 2020-09-23
Ads/Adding strict to air-discount-scheme-backend

dabbeg:
# Strict typescript for air-discount-scheme-backend
#253 

## What

Making air-discount-scheme follow the strict typescript protocol

## Why

A step in the direction of making all the monorepo strict

## Checklist:


apps/air-discount-scheme/backend/src/app/modules/common/test/unit/auth.guard.spec.ts
```diff
@@ -7,7 +7,7 @@ const authGuard = new AuthGuard()
 
 describe('AuthGuard', () => {
   it('Api key should work for Ernir', async () => {
-    const request = {
+    const request: any = {
```
dabbeg, line 10: I was unable to mock an express request and unable to cast this object to HttpRequest

apps/air-discount-scheme/backend/src/app/modules/flight/flight.model.ts
```diff
@@ -77,7 +77,7 @@ export class FlightLeg extends Model<FlightLeg> implements TFlightLeg {
 
   // eslint-disable-next-line
   @BelongsTo(() => Flight)
-  flight
+  flight: any
```
dabbeg, line 80: The app crashes it has the Flight type, there is a circular dependency going on here because Flight hasMany FlightLegs and FlightLeg has one Flight.
@darrikonn Do we need Flight here? Are you referencing Flight from a FlightLeg in the admin view?

apps/air-discount-scheme/backend/src/app/modules/flight/test/unit/flight.controller.spec.ts
```diff
@@ -114,6 +114,7 @@ describe('PublicFlightController', () => {
     }
 
     it('should create flight', async () => {
+      const request: any = { airline }
```
dabbeg, line 117: Same issue here with mocking express request

apps/air-discount-scheme/backend/src/app/modules/flight/flight.model.ts
darrikonn: Yup, need it both in `admin` and `min-rettindi`.

apps/air-discount-scheme/backend/src/app/modules/flight/test/integration/flight.service.spec.ts
```diff
@@ -25,23 +25,23 @@ describe('create', () => {
         originalPrice: 50000,
         discountPrice: 30000,
         date: new Date('2021-03-12T12:35:50.971Z'),
-        cooperation: null,
+        cooperation: '',
```
darrikonn, line 28: null is actually the expected behaviour. The column allows null with no default value.

apps/air-discount-scheme/backend/src/app/modules/user/user.controller.ts
```diff
@@ -65,11 +65,12 @@ export class PrivateUserController {
...
   ): Promise<User[]> {
     const relations = await this.userService.getRelations(params.nationalId)
-    return Promise.all([
+    const users = await Promise.all([
       this.userService.getUserInfoByNationalId(params.nationalId),
       ...relations.map((nationalId) =>
         this.userService.getUserInfoByNationalId(nationalId),
       ),
     ])
+    return users.filter((user) => user) as User[]
```
darrikonn, line 74: 🙈

darrikonn (commented):
Nice!

lots of `any` though :)

apps/air-discount-scheme/backend/src/app/modules/flight/test/integration/flight.service.spec.ts
dabbeg: Checkout the latest commit, I made the cooperation attribute optional (as it should have been) and I removed cooperation from the tests where they were null. That should be more accurate to when a request comes in and has no cooperation.

darrikonn (dismissed):

apps/air-discount-scheme/backend/tsconfig.json
```diff
@@ -2,7 +2,9 @@
   "extends": "../../../tsconfig.json",
   "compilerOptions": {
     "allowJs": true,
-    "types": ["node", "jest", "./types"]
+    "types": ["node", "jest", "./types"],
+    "strict": true,
+    "strictPropertyInitialization": false
```
arnifreyrs, line 7: I had some problems with this, but I managed to make it work. https://mariusschulz.com/blog/strict-property-initialization-in-typescript 

just food for thought

arnifreyrs (approved):
Nice!

apps/air-discount-scheme/backend/tsconfig.json
dabbeg: Nice, I enabled strictPropertyInitialization again and used ! to ignore certain properties that it should ignore.

darrikonn (approved):

PR #1013, opened 2020-09-23
Lazy load the form descriptor

arnifreyrs:
# Lazy load form descriptors in application system

## What

Lazy load form descriptors in application system instead of having them always part of each role in each state in each template. 

## Why

This is the first step in addressing bundle size optimisation for the application system. Form descriptors can be huge and we do not need to load all forms for a given application template. The system should not expose form descriptors for forms that are not visible to the user.

## Checklist:


apps/application-system/form/src/routes/Application.tsx
```diff
@@ -21,19 +29,47 @@ export const Application = () => {
     skip: !id,
   })
 
+  const application = data?.getApplication
+  const nationalRegistryId = userInfo?.profile?.natreg
+
+  useEffect(() => {
```
magnearun, line 35: I feel like this belongs in the application form lib, not in the route. Maybe we should move this logic and the application query down one level, so the Route only needs to add 
`
<ApplicationForm
applicationId={applicationId}
nationalRegistryId={nationalRegistryId} />
`

madebynoam (dismissed):
Agree with Magnea's comment, otherwise lgtm!

madebynoam (approved):
lgtm

PR #1028, opened 2020-09-23
Ads/removing zero and O from discount code because users tend to mix them up

dabbeg:
## What

Removing zero and O from discount code

## Why

Because users tend to mix them up

## Checklist:


darrikonn (approved):

darrikonn (commented):
Would like to see some tests though :)

dabbeg:
> Would like to see some tests though :)

Yes me too and I would like some help on how to test a random generator :)

PR #1032, opened 2020-09-23
Env file indexer fix

stefanhar:
# Env file indexer fix

## What

- Env file indexer fix

## Why

- Does not work as expected


siggakogk (approved):
🚀

PR #1036, opened 2020-09-23
Use scrollspy fix

stefanhar:
# Use scrollspy fix

## What

- useScrollSpy was causing errors on single articles


siggakogk (approved):
lets nlot throw errors!

stjanilofts (approved):

PR #1048, opened 2020-09-24
Add minimal config for Kodiak.

sindrig:
Prep work so we can continue with #938

## What

Prepare for kodiakhq.com merge bot

## Why

Because we want merge bots


pshomov (approved):
LGTM

PR #1052, opened 2020-09-24
{WIP} Add contentful page reference to link content type

stjanilofts:
# Add Contentful page reference to link content type

https://app.asana.com/0/1183498341031105/1195557775159433

## What

When getting menu entries with links there are only absolute paths which is fine for pages other than island.is that use the same menu links (such as ferdagjof/loftbru). This adds a reference to content type inside links.

## Why

For better performance on island.is we need a reference to the content type to prevent full page reloads and make the page prefetch. 

## Checklist:


oddsson:
I feel like I can't give a good review on this due to my inexperience with GraphQL and the lack of knowledge of the problem at hand so I'm removing myself as a reviewer 😬

apps/web/routes/index.ts
```diff
@@ -7,7 +8,9 @@ const routes = {
     page: 'stofnanir', // quick fix for launch
     Page: 'stofnanir',
     category: 'flokkur',
+    articleCategory: 'flokkur',
     ArticleCategory: 'flokkur',
+    contentCategory: 'flokkur',
     ContentCategory: 'flokkur',
```
barabrian, line 14: are these duplicates?

apps/web/routes/index.ts
```diff
@@ -21,7 +24,9 @@ const routes = {
     page: 'organizations',
     Page: 'organizations',
     category: 'category',
+    articleCategory: 'category',
     ArticleCategory: 'category',
+    contentCategory: 'category',
     ContentCategory: 'category',
```
barabrian, line 30: are these duplicates?

barabrian (commented):
I like the idea and this looks good to me. But there is a-lot going on in this PR and it would probably be best to get a second opinion maybe from @stefanhar

PR #1058, opened 2020-09-24
Lazy load application templates to minimize bundle size.

arnifreyrs:
## What
Lazy load big files that have no business being part of our bundle

## Why
We have to lazy load big files that have no business being part of our bundle. In the UI, there can only be one template with one form present at any given time, so this will not scale going forward.

## Checklist:


libs/application/form/src/lib/ApplicationForm.tsx
```diff
@@ -29,8 +29,13 @@ export const ApplicationForm: FC<{
   useEffect(() => {
     async function populateForm() {
       if (application !== undefined && form === undefined) {
-        const template = getApplicationTemplateByTypeId(application.typeId)
-        const stateInformation = getApplicationStateInformation(application)
+        // TODO wrap in try or?
```
madebynoam, line 32: Yeah, maybe for the next PR, catch in case it fails?...

madebynoam (approved):
Lgtm

PR #1083, opened 2020-09-28
Stop using types from api in island-ui/contentful

maggisk:
# \<Description\>

Using types from the graphql api made it a requirement for the client to fetch all fields from
the api for the types to match, which is causing several issues (e.g. it
breaks the client when fields are added to the Slice union type in the
graphql server)

## Checklist:


libs/island-ui/contentful/src/lib/richTextRendering.tsx
```diff
@@ -9,33 +9,79 @@ import {
...
-import { TeamList } from './TeamList/TeamList'
-import { ContactUs } from './ContactUs/ContactUs'
+import {
+  SectionWithImage,
+  SectionWithImageProps,
+} from './SectionWithImage/SectionWithImage'
+import { TeamList, TeamListProps } from './TeamList/TeamList'
+import { ContactUs, ContactUsProps } from './ContactUs/ContactUs'
+
+type HtmlSlice = { __typename: 'Html'; document: any }
```
jeremybarbet, line 37: Is document not of type `import { Document } from '@contentful/rich-text-types'`

jeremybarbet (dismissed):

libs/island-ui/contentful/src/lib/richTextRendering.tsx
maggisk: That is true, but the graphql server (or at least the type generated from the query) returns it as a JSON ({ [key: string]: any }) type, so if the use the Document type here, the types won't match

jeremybarbet: Make sense!

stjanilofts (approved):

PR #1086, opened 2020-09-28
Accordion hash urls

siggakogk:
# \<Accordion hash urls \>

## What

Add selected/expanded accordion group names to hash

## Why

Stop blocking users from keeping more than one group opened
Ability to route back to previous page

## Screenshots / Gifs

Screen Recording 2020-09-28 at 12 25 29

## Checklist:


apps/web/screens/Category/Category.tsx
```diff
@@ -167,6 +152,25 @@ const Category: Screen<CategoryProps> = ({
...
+      if (hashArray.indexOf(groupSlug) > -1) {
+        hashArray.splice(hashArray.indexOf(groupSlug), 1)
+        newHash = hashArray.join(',')
+      } else {
+        newHash = newHash.concat(`,${groupSlug}`)
+      }
+    }
+
+    setHash(newHash)
+    Router.push(makePath('ArticleCategory', category.slug + `#${newHash}`))
```
maggisk, line 171: Should we use Router.replace ?
If I understand correctly, this would add to history each time the user expands/collapses a group, and would require several back button clicks to get back to the page he came from?

siggakogk: Yep totally agree! Do not want to click "back" 10 times just because you expanded a lot of groups 😅

stjanilofts (approved):

PR #1077, opened 2020-09-25
Adds a large radio button

oddsson:
# Adds a prop to the Radio button component to make it big

## What

According to the design there should be variant sizes of radio buttons. This PR implements a large radio button

## Why

See ☝️

## Screenshots / Gifs

Screenshot 2020-09-25 at 15 14 49
Regular ☝️

Screenshot 2020-09-25 at 15 14 35
Large ☝️

## Checklist:

- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] Formatting passes locally with my changes
- [x] I have rebased against master before asking for a review

PR #1095, opened 2020-09-28
Large radio buttons now have a bigger clickable area

oddsson:
# Large radio buttons now have a bigger clickable area

## What

In #1077, I added a `large` prop on the radio button component. That component was made large by adding padding to the radio button container. That meant that since only the label is clickable, the clickable area was very small. I fixed this by setting the padding on the label instead, making the clickable area much larger.

## Why

For usability.

## Screenshots / Gifs

ezgif com-video-to-gif (2)

ezgif com-video-to-gif (3)

## Checklist:


barabrian (approved):
LGTM!

Issue #679, opened 2020-09-07
Add dedicated healthecheck paths to NextJS apps


pshomov:
They should be `/liveness` and `/readiness`

PR #1097, opened 2020-09-28
Feature/api healthchecks

stefanhar:
# API healthchecks

## What

- API is currently responding when it is up instead of when it is actually ready

## Why

- So the API is set to ready state when it is ready

## Checklist:


haukurh:
why are environment.prod.ts removed?

haukurh (approved):
I find it strange to remove environment.prod, if it's intended to implement/work on it in the future.
But otherwise LGTM

haukurh (approved):

pshomov:
@stefanhar Can you please add to the description of the PR that it fixes #679

stefanhar:
> @stefanhar Can you please add to the description of the PR that it fixes #679
This is not a fix for this issue, this exists to provide a base for correct readiness checks in the API, currently readiness and liveliness are the same endpoint.
We had some issues with the API declaring itself ready when it seems it wasn't during stress tests.
I inserted dns resolve checks as a starting point for cms readiness check, but It is likely not enough but we will go into further diagnostics on that issue once this is merged

stefanhar:
> why are environment.prod.ts removed?

It seems to me that the environment files are not swaped out during build (based on missing `fileReplacements` in libaries in workspace.json) hence `environment.prod.ts` is never swaped out and never used.
I removed these files to reduce the risk of errors until this is handled properly

arnifreyrs (approved):
LGTM but I am not knowledgable in this area, so some other can provide better review

PR #1101, opened 2020-09-28
Service portal/localization

atligudl:
# \<Description\>

## What

Bootstrap localization in service portal app and libs

## Why

To provide language options to the service portal

## Screenshots / Gifs

## Checklist:


libs/service-portal/core/src/lib/service-portal-core.ts
```diff
@@ -4,11 +4,19 @@ import { User } from 'oidc-client'
 import { ServicePortalPath } from './navigation/paths'
 import { ApolloClient, NormalizedCacheObject } from '@apollo/client'
 
+export interface ReactIntlMessage {
```
arnifreyrs, line 7: I am wondering if this should be part of the localization package instead?

arnifreyrs (approved):
LGTM, just one comment

libs/service-portal/core/src/lib/service-portal-core.ts
atligudl: Yeah, that would make sense.

@magnearun should I move it to the localization lib?

Thorrik58 (approved):
LGTM

libs/service-portal/core/src/lib/service-portal-core.ts
magnearun: I think we don't need this interface.

The formatMessage method looks like this formatMessage(descriptor: MessageDescriptor, values: <Record<K, V>>) MessageDescriptor is exported by 'react-intl' as 
MessageDescriptor {
    id?: string | number;
    description?: string | object;
    defaultMessage?: string | MessageFormatElement[];
}

name: ReactIntlMessage | string 

should rather be

name: MessageDescriptor | string

magnearun (changes requested):
LGTM, just one minor comment about the message types.

magnearun (approved):

obmagnusson (approved):
LGTM

workspace.json
```diff
@@ -1186,6 +1186,17 @@
             "passWithNoTests": true
           }
         },
+        "extract-strings": {
```
pshomov, line 1189: Do we need to call these targets during the build process?

atligudl: Not atm, this is done manually

@magnearun correct me if I'm wrong

arnifreyrs (approved):

workspace.json
magnearun: This is manual atm, not sure when in the process we want to do this. This extracts the strings from the code and uploads them into Contentful

PR #1103, opened 2020-09-28
Fix CI due to new build

sindrig:
New build for codeowners check was being considered a last good build
and we could not retag any images.

TODO: Fail build if we can't retag images!


.github/actions/main.ts
```diff
@@ -4,15 +4,20 @@ import { findLastGoodBuild, WorkflowQueries } from './detection'
 const octokit = new Octokit()
 // const octokit = new Octokit({ auth: process.env.GITHUB_TOKEN });
 const [owner, repo] = process.env.GITHUB_REPOSITORY.split('/')
+const workflow_file_name = 'build.yml'
```
eirikurn, line 7: Don’t think this passes eslint naming conventions (if they applied to this file) but meh.

eirikurn (approved):

PR #1104, opened 2020-09-29
Contentful sync: retrieve root entry from processEntry

haukurh:
# Contentful sync: retrieve root entry from processEntry

## What

Retrieve root entry based of processEntry id, during sync, to include in ElasticSearch import

## Why

If embedded processEntries are updated in Contentful, we don't get a signal which root element it belongs to.
We need to resolve which root entries said embedded entry belongs to, in order to index it to ElasticSearch.

## Checklist:


libs/api/domains/cms/src/lib/search/contentful.service.ts
```diff
@@ -183,14 +196,39 @@ export class ContentfulService {
       deletedEntries,
     } = await this.getSyncData(typeOfSync)
 
+    const processEntries = entries
+      .filter((entry) => entry.sys.contentType.sys.id === 'processEntry')
+      .map((entry) => entry.sys.id)
+
```
stefanhar, line 202: This needs to cast a wider net `entry.sys.contentType.sys.id === 'processEntry'` -> `nestedContentTypes.includes(entry.sys.contentType.sys.id)`
Where `nestedContentTypes` is a list of all content types nested in all page types

libs/api/domains/cms/src/lib/search/contentful.service.ts
```diff
@@ -183,14 +196,39 @@ export class ContentfulService {
...
     logger.info('Sync found entries', {
       entries: entries.length,
       deletedEntries: deletedEntries.length,
+      processEntries: processEntries.length,
```
stefanhar, line 206: Update `processEntries` to `nestedEntries` as per comment above

libs/api/domains/cms/src/lib/search/contentful.service.ts
```diff
@@ -170,6 +170,19 @@ export class ContentfulService {
     return _.flatten(chunkedChanges)
   }
 
+  private async linksToEntry(
```
stefanhar, line 173: Maybe calling this function `getParentEntries` might be more descriptive

stefanhar (changes requested):

libs/api/domains/cms/src/lib/search/contentful.service.ts
```diff
@@ -183,14 +196,39 @@ export class ContentfulService {
...
     const items = await this.getAllEntriesFromContentful(entries, locale)
 
+    // In case of delta updates, we need to resolve embedded entries to their root model
+    if (!fullSync && processEntries) {
+      logger.info('Finding root entries from processEntries')
+      const alreadyProcessedIds = items.map((entry) => entry.sys.id)
+      for (const entryId of processEntries) {
+        // Due to the limitation of Contentful Sync API, we need to query every entry one at a time
+        // with regular sync, triggered by a webhook, these calls 1 - 2 at most
+        const linkedEntries = await this.linksToEntry(entryId, locale)
```
stefanhar, line 219: This should use the limiter to reduce risk of throttling

stefanhar (changes requested):

libs/api/domains/cms/src/lib/search/contentful.service.ts
haukurh: I decided to let the name reflect the Contentful API.

According to [Contentful documentation](https://www.contentful.com/developers/docs/references/content-delivery-api/#/reference/search-parameters/links-to-entry), this functionality has more possible use cases than we're using it for.
> To search for entries which have a field linking to a specific entry

I do agree this name is not very descriptive, when in use, but I think it's also good that it reflects their API.
Some other possible approaches, I would like to take up for discussion:
- create `getParentEntries` as an alias function of `linksToEntry`, which would let code "document/describe" it self
- comment linksToEntry, with what it does
- comment when use, what you're trying to achieve
- just rename the function like you suggest, because it probably won't be used in any other way.

any thoughts?

libs/api/domains/cms/src/lib/environments/environment.ts
```diff
@@ -9,6 +9,7 @@ export default {
     'news',
     'page',
   ],
+  nestedContentTypes: ['processEntry'],
```
stefanhar, line 12: We should probably provide a complete list of types here to fully complete this issue

libs/api/domains/cms/src/lib/search/contentful.service.ts
stefanhar: Well our code does not reflect the Contentful API already e.g. `getContentfulData` is not called `getEntriesByIds`.
But I don't see this as a big issue, I'm ok with this either way

libs/api/domains/cms/src/lib/search/contentful.service.ts
```diff
@@ -170,6 +170,41 @@ export class ContentfulService {
...
+   * @param linkId
+   * @param locale
+   * @private
+   */
+  private async getParentEntries(
+    linkId: string,
+    locale: string,
+  ): Promise<Entry<any>[]> {
+    return this.linksToEntry(linkId, locale)
+  }
```
stefanhar, line 206: Maybe we should rather just use the original function?
To me this increases the complexity of this file with little benefits

stefanhar (changes requested):

stefanhar (approved):

PR #1113, opened 2020-09-29
Each application template is a library

arnifreyrs:
# Each application template is a library

## What

Build new libraries for all application templates, so each application template lib has all the specs for said application type.

## Why

This is to decrease bundle size, and also, when a developer is working on some **application A**, it should not need to trigger builds for other applications.

TODO - in a later PR I will upgrade all the README files for the application system.

## Checklist:

- [] I have added tests that prove my fix is effective or that my feature works


sindrig:
LGTM but I'm not confident enough to review this :grimacing:

madebynoam (approved):
Lgtm!

workspace.json
```diff
@@ -3098,6 +3098,122 @@
           }
         }
       }
+    },
```
magnearun, line 3101: We will need to move the extract strings script to each template lib

magnearun (changes requested):
LGTM! 
Just a small fix to get the extract-strings script to work

workspace.json
arnifreyrs: done

eirikurn (approved):

magnearun (approved):

PR #1120, opened 2020-09-29
Always use linefeed (lf) not OS specific line endings

eirikurn:
## What

Prettier 2.0 changes the default line ending from auto to always be "lf". For this to work well for Windows users, we need to configure git to checkout code with lf as well.

## Why

This is to avoid mixed end-line characters across the repo and inconsistent prettier behaviour across operating systems.

See here: https://prettier.io/docs/en/options.html#end-of-line


pshomov (approved):
👍🏻

sindrig (approved):
:+1: :end:

PR #1133, opened 2020-09-30
Fix bug in scrollspy + firefox

maggisk:
# \<Description\>

scrollspy hook is buggy in firefox sometimes, causing the navigation indication
bullet to move to an incorrect element

## What

Fix bug in scrollspy hook that only affects Firefox

## Why

FF returns getBoundingClientRect().top with floating point
precision but window.scrollY as integer, causing issues when
comparing scroll position to an element position

## Screenshots / Gifs

Bug behavior:
Screen Recording 2020-09-30 at 10 10 21 mov

## Checklist:


siggakogk (approved):
Nice! GJ 🤙

PR #1138, opened 2020-09-30
Language toggler that checks if content exists

stjanilofts:
## What

Show modal when switching to english indicating that the english content is not complete. It also checks if there is english content available and redirects to that instead of the frontpage.

## Why

"Better" experience for english users.

## Screenshots / Gifs

image

## Checklist:


apps/web/graphql/client.ts
```diff
@@ -43,6 +47,14 @@ function create(initialState?: any) {
     ssrMode: !isBrowser, // Disables forceFetch on the server (so queries are only run once)
     link: httpLink,
     cache: new InMemoryCache({
+      dataIdFromObject(obj: IdGetterObj) {
+        switch (obj.__typename) {
+          case 'ArticleCategory':
```
stefanhar, line 52: It should be safe to make this more generic as in always return the id with locale appended.
Maybe remove the switch and return defaultDataIfFromObject with locale appended to it?

stefanhar (changes requested):

apps/web/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,138 @@
...
+
+    useLayoutEffect(function() {
+      setMounted(true)
+    }, [])
+
+    return mounted && <div className={styles.backdrop} {...props} ref={ref} />
+  },
+)
+
+export const Modal: FC<ModalProps> = ({
```
arnifreyrs, line 47: I am wondering why we have no modal in the island-ui kit

apps/web/graphql/withApollo.tsx
```diff
@@ -1,12 +1,24 @@
...
 export const withApollo = (Component) => {
   const NewComponent = ({ apolloState, pageProps }) => {
+    const Router = useRouter()
+
+    const { asPath } = Router
+
+    let clientLocale = null
+
+    if (asPath) {
+      clientLocale = getLocaleFromPath(asPath)
```
arnifreyrs, line 17: so clientLocale could be null if `asPath` is falsy. Is that alright? should it be defaulted to Icelandic instead of being null?

arnifreyrs (approved):
LGTM, had only two questions

apps/web/components/Modal/Modal.tsx
stjanilofts: I will probably move this eventually to the island-ui kit but for now I wanted to test it out for this small use case where it shows an alert dialog when switching to english.

apps/web/graphql/withApollo.tsx
stjanilofts: Right, I have made some changes here (there was no need to check for asPath).

apps/web/graphql/withApollo.tsx
```diff
@@ -1,19 +1,24 @@
...
+import { getLocaleFromPath } from '@island.is/web/i18n/withLocale'
 import initApollo from './client'
 
 export const withApollo = (Component) => {
   const NewComponent = ({ apolloState, pageProps }) => {
+    const { asPath } = useRouter()
+    const clientLocale = getLocaleFromPath(asPath)
     return (
-      <ApolloProvider client={initApollo(apolloState)}>
+      <ApolloProvider client={initApollo({ ...apolloState, clientLocale })}>
```
maggisk, line 13: Maybe clientLocale should be a separate paramater to initApollo, rather than add data to apolloState unrelated to apolloState?

apps/web/screens/Article.tsx
```diff
@@ -340,11 +349,20 @@ export interface ArticleProps {
 }
 
 const ArticleScreen: Screen<ArticleProps> = ({ article, namespace }) => {
+  const { setContentfulId } = useContext(GlobalContext)
   const n = useNamespace(namespace)
   const { query } = useRouter()
   const { activeLocale } = useI18n()
   const { makePath } = routeNames(activeLocale)
 
+  useEffect(() => {
```
maggisk, line 358: Could we make a custom hook for this? Like useContentfulId rather than using useEffect everywhere and having to remember to cleanup

stjanilofts: True, I moved it into a hook.

maggisk (approved):

libs/island-ui/core/src/lib/Button/Button.tsx
```diff
@@ -64,6 +64,7 @@ export const Button = forwardRef<
       white,
       tabIndex,
       rounded = false,
+      ...rest
```
barabrian, line 67: we should not have the button component spread like this, we are rewriting the button component so this will change

barabrian (approved):

PR #1144, opened 2020-09-30
Use API to generate contentful types

sindrig:
This closes island-is/infrastructure#148

## What

Use central API to generate contentful types (and communicate with contentful)

## Why

We don't want developers to have production access keys to contentful.

## Checklist:


libs/api/domains/cms/scripts/generateContentfulTypes.ts
```diff
@@ -1,21 +1,26 @@
-import { codegen } from '@jeremybarbet/contentful-typescript-codegen'
+import 'isomorphic-fetch'
+import { writeFileSync } from 'fs'
+import { format, resolveConfig } from 'prettier'
 import { createClient } from 'contentful-management'
```
sindrig, line 4: @jeremybarbet without this import I get this error: ```sindri->~/dev/andes/island.is (master) (shared:internal-s) $ npx nx run api:contentful-types

> nx run api:contentful-types 
⨯ Unable to compile TypeScript:
libs/api/domains/cms/scripts/generateContentfulTypes.ts:8:16 - error TS2705: An async function or method in ES5/ES3 requires the 'Promise' constructor.  Make sure you have a declaration for the 'Promise' constructor or include 'ES2015' in your `--lib` option.

8 async function main() {
                 ~~~~
```

Any idea?

File: libs/api/domains/cms/src/lib/generated/contentfulTypes.d.ts [hunk 2]
```diff
@@ -18,6 +18,9 @@ export interface IAboutSubPageFields {
 
   /** Content */
   content: Document
+
```
darrikonn, line 21: Should this file be tracked?

cc. @jeremybarbet

arnifreyrs (approved):
LGTM but I am not knowledgable in this area, so some other can provide better review

libs/api/domains/cms/src/lib/generated/contentfulTypes.d.ts
jeremybarbet: We talked about it with Sindri last night, and it's safer to keep the file on the repository. Removing it would create a huge dependency on contentful and in case contentful is down for some reason it would break all builds on our pipeline

jeremybarbet (approved):
🙌🏻

darrikonn (approved):

dabbeg (approved):
:+1:

Issue #389, opened 2020-08-17
Document how to run the CI process locally


pshomov:
To help people troubleshoot CI issues on their dev machines

PR #1152, opened 2020-10-01
Refactoring CI pipeline in preparation for documenting it

pshomov:
Supports #389 

## What

Pull parts of the GHA in scripts that can be run locally

## Why

See https://github.com/island-is/handbook/blob/master/docs/adr/0002-continuous-integration.md#decision-drivers-policy

## Checklist:


eirikurn (approved):
Nice

.github/workflows/build.yml
```diff
@@ -111,7 +89,7 @@ jobs:
           else
             export PUBLISH=true
           fi
-          echo "::set-env name=PUBLISH::${PUBLISH}"
+          export PUBLISH=true
```
sindrig, line 92: i know this is WIP but still

.github/workflows/build.yml
```diff
@@ -134,10 +112,6 @@ jobs:
           path: node_modules
           key: ${{ runner.os }}-${{ hashFiles('yarn.lock') }}-yarn
 
-      - name: Building NodeJS dependencies - host OS
```
sindrig, line 137: probably ditto?

.github/workflows/build.yml
```diff
@@ -154,61 +128,45 @@ jobs:
         run: echo "::set-env name=BRANCH::${GIT_BRANCH}"
 
       - name: Security audit Node modules
-        run: yarn run security-audit
+        run: ./scripts/security-audit.sh
```
sindrig, line 131: why create scripts for one liners?

scripts/license-audit.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+set -euxo pipefail
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+
+source $DIR/_common.sh
+
+cd $PROJECT_ROOT
```
sindrig, line 8: suggest doing this in subshell

scripts/prepare_deps.sh
```diff
@@ -7,6 +7,10 @@ source $DIR/_common.sh
 
 # Build the node_modules as well as the base image for the final outputs and store it in the shared cache so it can be reused later
 
+cd $PROJECT_ROOT
```
sindrig, line 10: same, subshell

pshomov: 👍🏻

scripts/license-audit.sh
pshomov: 👍🏻

.github/workflows/build.yml
pshomov: Longer discussion, let's have it "live" ;)

scripts/license-audit.sh
pshomov: 👍🏻

.github/workflows/build.yml
pshomov: actually we put both dependence updates in the same script

.github/workflows/build.yml
```diff
@@ -148,67 +123,51 @@ jobs:
 
       - name: Building NodeJS dependencies
         if: steps.cache-deps.outputs.cache-hit != 'true'
```
sindrig, line 125: Only thing I can see is that we might want to check the cache-hit status of `steps.node-modules.outputs.cache-hit` as well - I know it depends on the same file but if one was invalidated somehow or something... /shrug

sindrig (approved):

PR #1160, opened 2020-10-01
ads/register payment

dabbeg:
- Updating kennitala package as it now includes ts types
- Ads/adding column to track the date when financial_state is updated
- Ads/Updating financial_state_updated when a flight is cancelled

# \<Description\>

## What

Add financial state updated column

## Why

To be able to know when the state changed for reporting reasons

## Checklist:


darrikonn (dismissed):

darrikonn (approved):

apps/air-discount-scheme/backend/migrations/20201001142253-fix-financial-state-default-value.js
```diff
@@ -0,0 +1,19 @@
+'use strict'
+
+module.exports = {
+  up: (queryInterface) => {
+    return queryInterface.sequelize.query(`
+      BEGIN;
+
+        ALTER TABLE flight_leg ALTER COLUMN financial_state SET DEFAULT 'AWAITING_DEBIT';
```
sindrig, line 8: do you also need to `update table flight_leg set financial_state='AWAITING_DEBIT' where financial_state='awaitingDebig';`?

darrikonn: Nope.. no records for it -> thankfully handled in the app layer.

However, it doesn't hurt to be extra careful and add it.

dabbeg: No we checked with the lambda thing, there were no records with `awaitingDebit`

sindrig: :hindu_temple: :church:

PR #1192, opened 2020-10-05
Better side effect handling

oddsson:
# Sanity checks when working with data from localStorage.

## What

When running the project without a user, it would crash. This would happen because sometimes cases in localstorage are not set and we need to make sure that that data is set if we want to work with it.

## Checklist:

### DIFF
--- a/apps/judicial-system/web/src/api/index.ts
+++ b/apps/judicial-system/web/src/api/index.ts
@@ -40,18 +40,25 @@ export const getCases: () => Promise<DetentionRequest[]> = async () => {
 export const getCaseById: (
   caseId: string,
 ) => Promise<GetCaseByIdResponse> = async (caseId: string) => {
-  const response = await fetch(`/api/case/${caseId}`)
+  try {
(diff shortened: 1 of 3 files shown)


apps/judicial-system/web/src/api/index.ts
```diff
@@ -40,18 +40,25 @@ export const getCases: () => Promise<DetentionRequest[]> = async () => {
 export const getCaseById: (
   caseId: string,
 ) => Promise<GetCaseByIdResponse> = async (caseId: string) => {
-  const response = await fetch(`/api/case/${caseId}`)
+  try {
+    const response = await fetch(`/api/case/${caseId}`, {
+      method: 'get',
+      headers: { Authorization: `Bearer ${csrfToken}` },
```
gudjong, line 46: Very important - how did we miss this?

gudjong (approved):

apps/judicial-system/web/src/api/index.ts
oddsson: No idea 🙃  How was this working before this?

PR #1195, opened 2020-10-05
chore: missing auto-generated folder

jeremybarbet:
# chore: missing auto-generated folder

## What

I removed `gen/fetch` directories that are automatically generated.

## Why

- Try to remove as much as we can of the generated files from the repository.
- This script needs to be run first outside of the `affected` loop otherwise it fails to build other code.

## Screenshots / Gifs

—

## Checklist:


package.json
```diff
@@ -36,7 +36,7 @@
     "help": "nx help",
     "storybook:island-ui": "nx run island-ui-storybook:storybook",
     "ts-node": "ts-node -r tsconfig-paths/register -O '{\"module\":\"commonjs\",\"esModuleInterop\":true}'",
-    "schemas": "test -f libs/api/schema/src/lib/schema.d.ts || echo 'export default () => {}' >> libs/api/schema/src/lib/schema.d.ts && yarn nx run api:contentful-types && yarn nx run api:build-schema && yarn affected:schemas",
+    "schemas": "test -f libs/api/schema/src/lib/schema.d.ts || echo 'export default () => {}' >> libs/api/schema/src/lib/schema.d.ts && yarn nx run api:contentful-types && yarn nx run application-system-api:build-schema && yarn nx run api-domains-application:codegen && yarn nx run api:build-schema && yarn affected:schemas",
```
sindrig, line 39: This is quite long now - does it make sense to extract stuff like this from package.json?

jeremybarbet: Yes, it's getting quite long. Since these scripts need to be run first, it needs to be out of the nx/workspace configuration.

I  could do something like this inside the workspace defining a new target to group all scripts calls:

```
"postinstall": {
  "root": "",
  "sourceRoot": "",
  "projectType": "",
  "schematics": {},
  "architect": {
    "init": {
      "options": {
        "parallel": false,
        "commands": [
          "yarn nx run api:contentful-types",
          "yarn nx run application-system-api:build-schema",
          "yarn nx run api-domains-application:codegen",
          "yarn nx run api:build-schema",
          "yarn affected:schemas"
        ]
      }
    }
  }
},
```

and then use it as
```bash
"schemas": "test -f libs/api/schema/src/lib/schema.d.ts || echo 'export default () => {}' >> libs/api/schema/src/lib/schema.d.ts && yarn nx run postinstall:init",
```

@eirikurn What do you think?

jeremybarbet: Hey, we talked about it with @eirikurn and I'm cleaning up this logic on another PR with a better solution that handles implicit dependencies between projects Cf: https://github.com/island-is/island.is/pull/1225

In the meantime, this PR can be merged 👍

sindrig (approved):

DavidBachmann (approved):

PR #1206, opened 2020-10-05
Add Datefield to application system

arnifreyrs:
## What

Create new library: `shared-form-fields` for components used in a "react hook form"-form. Moved all the controllers from island-ui into that new lib. Added `DatePickerController`

## Why

We need a datepicker in the application system.

## Checklist:


libs/shared/form-fields/src/lib/DatePickerController/DatePickerController.spec.tsx
```diff
@@ -0,0 +1,72 @@
...
+          id="id"
+          label="date"
+          placeholder="great placeholder"
+        />
+      </Wrapper>,
+    )
+    expect(getByText('great placeholder')).toBeInTheDocument()
+  })
+
+  it('should render no placeholder if the daatepicker has a selected value', () => {
```
madebynoam, line 48: minor, typo `daatepicker ` :)

apps/application-system/form/src/styles.scss
```diff
@@ -0,0 +1,45 @@
+@font-face {
```
madebynoam, line 1: Should these be part of core theme?

madebynoam (approved):
Lgtm!

apps/application-system/form/src/styles.scss
arnifreyrs: probably, I just don't know how to do it in another way :/ just followed Kolibri's lead on this.

madebynoam (approved):

pshomov (approved):
LGTM

apps/application-system/form/src/styles.scss
jeremybarbet: Agree, it should probably be moved to the island-ui. One scss file that could be imported along with the fonts folder

workspace.json
```diff
@@ -1866,7 +1866,7 @@
               "apps/application-system/form/src/favicon.ico",
               "apps/application-system/form/src/assets"
             ],
-            "styles": [],
+            "styles": ["apps/application-system/form/src/styles.scss"],
```
jeremybarbet, line 1869: Could the logic be done here: `libs/island-ui/core/src/global/fonts.treat.ts`

and above
```
"assets" : [
  ...,
  libs/island-ui/core/src/assets/fonts
]
```

this is the theory, not sure if it works

jeremybarbet (approved):
LGTM, apart from the font question

workspace.json
oddsson: I believe there is an open issue on the Treat repo that addresses this and according to them, setting global font-face in Treat does not work as expected. I'll post this on Slack to see if anyone has the answer to this.

https://github.com/seek-oss/treat/issues/93

oddsson (commented):
This looks good to me, however [this](https://github.com/island-is/island.is/pull/1206/files#r500360521) issue stands out. I don't know if we want to try to fix it here or later.

workspace.json
arnifreyrs: Then I am thinking this should be resolved in a later PR :)

oddsson (approved):

PR #1213, opened 2020-10-06
Skilavottord/Fix language toggler

dd-sdang:
# Skilavottord - Fix language toggler

## What

Fix language toggler so that it switches between is/en routes and uses correct translation files

## Checklist:


tommigitgaur (approved):
Fine

PR #1226, opened 2020-10-06
Adding support for running local proxy of ES service in Dev

### LINKED ISSUE #1215: Allow developers to access Dev env service proxies from local machines
This is coming from multiple people in the org. The goal is two fold:
 * to remove the need for local development replicas of services we use in AWS - ElasticSearch, Redis, etc.
 * to allow access to services that are only allowed to run in AWS - X-Road

pshomov:
Fixes #1215 

## What

Running a local proxy for ES. Can be easily refactored for proxying more services

## Why

See #1215 

## Checklist:


scripts/run-es-proxy.sh
```diff
@@ -0,0 +1,12 @@
...
+
+# This scripts sets up a local proxy for ElastSearch running in our Dev env
+#
+# You need to login to AWS portal and get some env variables as in step 1 here - https://github.com/island-is/handbook/blob/master/dockerizing.md#troubleshooting
+# Then simply run the script
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+
+docker build -f $DIR/Dockerfile.proxy -t es-proxy . 
+docker run --rm -e AWS_ACCESS_KEY_ID -e AWS_SECRET_ACCESS_KEY -e AWS_SESSION_TOKEN -p 9200:9200 es-proxy
```
sindrig, line 12: Since we're not referencing the `AWS_*` variables here, this will fail quite late. Maybe check for existence and fail if not set?

sindrig (approved):

scripts/run-es-proxy.sh
pshomov: good idea, will address in the follow up PR which is coming right up

PR #1229, opened 2020-10-06
Implements pdf generation for prosecutor requests

gudjong:
# PDF for prosecutor requests

## What

Implements PDF generation for prosecutor requests

## Why

Prosecutors want to file their requests on disk and pdf is a nice format for that.

## Checklist:


.github/workflows/build.yml
```diff
@@ -185,7 +185,7 @@ jobs:
   tests:
     needs: prepare
     runs-on: ubuntu-20.04
-    timeout-minutes: 25
+    timeout-minutes: 45
```
sindrig, line 188: this seems excessive. if there are tests that take >25 minutes to run we should rewrite them rather than this.

i'm working on improvements for the prep/test times. as this doesn't seem to help with the overall issue (tests never finish due to some unknown reason) I say remove this for now.

sindrig (changes requested):

.github/workflows/build.yml
gudjong: I was seeing random failures of various tasks. In two cases, tests were timing out, not the same tests each time. In one case formatting was timing out. And then there was some 40 min timeout, I could explain.

So, I don't think the tests as such were to blame.

But I agree 45 min is a bit excessive, but I was just tired of rerunning the build without success.

gudjong: And I really need to get this build through the pipeline.

sindrig: I get the frustration. But it looks to me to have been extremely random - at least none of the test steps in your latest run exceeded 16 minutes.

The good news is that looking at this reinforced me in my belief that we need to re-think how we're doing integration tests, and that's next on my to-do list. The bad news is it might take some days.

Bottom line - do what you need to do to get this through for now, and the future will be brighter... :sun_behind_large_cloud:

sindrig (approved):

apps/judicial-system/api/src/app/modules/case/case.pdf.ts
```diff
@@ -73,19 +90,18 @@ export async function generateRulingPdf(existingCase: Case): Promise<string> {
   const html = compiledFunction({
     courtCaseNumber: existingCase.courtCaseNumber,
     policeCaseNumber: existingCase.policeCaseNumber,
-    courtStartTime: formatTime(existingCase.courtStartTime),
-    courtEndTime: formatTime(existingCase.courtEndTime),
-    courtDate: formatDate(existingCase.courtStartTime),
+    courtStartTime: formatDate(existingCase.courtStartTime, 'HH:mm'),
```
oddsson, line 93: We have a constant for time format in Constants 👉  https://github.com/island-is/island.is/blob/master/apps/judicial-system/web/src/utils/constants.ts#L13

Consider using that instead

apps/judicial-system/api/templates/ruling.pug
```diff
@@ -37,9 +37,9 @@ html
 
     h3 Úrskurðarorð
     if isolation
-      p Kærði, <strong>#{accusedName} kt. #{nationalId}</strong>, skal sæta gæsluvarðhaldi, þó ekki lengur en <strong>til #{custodyEndDay} #{custodyEndDate} kl. #{custodyEndTime}.</strong> Kærði skal <strong>sæta einangrun</strong> á meðan á gæsluvarðhaldinu stendur.
+      p Kærði, <strong>#{accusedName} kt. #{accusedNationalId}</strong>, skal sæta gæsluvarðhaldi, þó ekki lengur en <strong>til #{custodyEndDate} kl. #{custodyEndTime}.</strong> Kærði skal <strong>sæta einangrun</strong> á meðan á gæsluvarðhaldinu stendur.
```
oddsson, line 40: Custody restrictions are missing here. Example string is 

Kærði, Gunnar Apoi kt.1112902539 skal sæta gæsluvarðhaldi, þó ekki lengur en til 22. október 2020 kl. 12:31. Kærði skal sæta fjölmiðlabanni, heimsóknarbanni og einangrun á meðan á gæsluvarðhaldinu stendur.

gudjong: Ég skal laga það en væri til í að fá þessa PR í gegn fyrst. Þetta er work in progress. Ef ég breyti einhverju og pusha fer af stað endalaus keðja að buildum og mergjum sem tekur allan daginn:(

oddsson (dismissed):

oddsson:
Approved

oddsson (dismissed):

sindrig (approved):

oddsson (approved):

sindrig (approved):

PR #1233, opened 2020-10-06
Website / Modifications to DrawerMenu

stjanilofts:
## Why

The drawer menu was showing too much empty space when it had only a few links. Now it only shows as much space as is required. Also, you can now scroll the drawer menu if it is taller than the viewport.

## Screenshots / Gifs

When the content height is lower than the viewport height you can scroll the document body while it is open and it opens only as much as is needed:
small

When the content is taller than the viewport height you can scroll the drawer menu meanwhile the document body stays in place:
full

## Checklist:


DavidBachmann (dismissed):
Cool. Now that `styles.root` is no longer the root you might want to rename it :)

siggakogk (approved):
nice!

PR #1242, opened 2020-10-07
Split up build yml

sindrig:
# Split build.yaml up in 2

Closes #1242 

## What

Faster builds, especially in feature branches

## Why

Because we're losing patience


sindrig:
image
these are two full builds - still a lot of speed to be gained in the on pull request.

.github/workflows/pullrequest.yml
```diff
@@ -0,0 +1,275 @@
...
+    if: always()
+    needs:
+      - prepare
+      - tests
+      - linting
+      - formatting
+      - build
+    steps:
+      - name: Check tests success
+        run: '[[ ${{ needs.tests.result }} != "failure" ]] || exit 1'
```
pshomov, line 247: nice, was wondering how to get that.

scripts/_chunk.js
```diff
@@ -16,6 +16,4 @@ let chunks = chunk(
   .map((ch) => ch.join(','))
   .filter((job) => job.length > 0)
 
-if (chunks.length == 0) chunks = ['dummy']
```
pshomov, line 19: 👍🏻

pshomov (approved):

.github/workflows/pullrequest.yml
pshomov: It is right there in the docs, not that I know about it ;)

PR #1246, opened 2020-10-07
Search indexer resync in migration

stefanhar:
# \<Description\>

https://app.asana.com/0/1183498341031105/1194377777305561

## What

- Changes in mapping may change document source hence elastic reindex is unreliable

## Why

- Ensure data in new indexes are up to date

## Checklist:


libs/api/domains/cms/src/lib/search/contentful.service.ts
```diff
@@ -109,25 +108,26 @@ export class ContentfulService {
...
     // write sync token to elastic here as it's own type
     logger.info('Writing next sync token to elasticsearch index')
-    return this.elasticService.index(
-      SearchIndexes[locale],
-      nextSyncTokenDocument,
-    )
+    return this.elasticService.index(elasticIndex, nextSyncTokenDocument)
   }
 
+  interface
```
jeremybarbet, line 114: interface alone?

apps/services/search-indexer/src/migrate/index.ts
```diff
@@ -9,6 +9,8 @@ interface PromiseStatus {
   error?: Error
 }
 
+export type locales = 'is' | 'en'
```
jeremybarbet, line 12: Is this type not available from somewhere else already defined? and it would be better to upperfirst it

libs/api/domains/cms/src/lib/search/contentful.service.ts
stefanhar: Hmm intresting that this ran without troubles, removed

jeremybarbet (approved):

PR #1251, opened 2020-10-08
Application system - optional field description

madebynoam:
## What

Allow developers of an application to include a description in any of the standard fields.

## Why

To fit the latest application system designs

## Screenshots / Gifs

image

## Notes
In a future PR, we will further investigate integrating Contentul Rich Text into these locale "strings".

## Checklist:


libs/application/ui-fields/src/lib/CheckboxFormField.tsx
```diff
@@ -17,14 +18,17 @@ const CheckboxFormField: FC<Props> = ({
...
+  const { id, name, description, options, disabled } = field
   const { formatMessage } = useLocale()
 
   return (
     <div>
       {showFieldName && (
         <Typography variant="p">{formatMessage(name)}</Typography>
       )}
+
+      {description && <Description description={formatMessage(description)} />}
```
arnifreyrs, line 30: I am curious how a field with description in a multifield will look

arnifreyrs (approved):
LGTM

libs/application/ui-fields/src/lib/CheckboxFormField.tsx
madebynoam: Good point, I assume it would be designed for only one description, so they could leave the second field description blank. Will keep it in mind if we want to enhance.

arnifreyrs (approved):

PR #1252, opened 2020-10-08
island-ui / Rename button imports

barabrian:
# Deprecating old button component

## What

We want to rename the old Button to ButtonDeprecated and add deprecated warnings to it and export new button as Button from island-ui

## Why

Old button was drifting apart from the updated UI Library and was getting harder to maintain, the new Button RC will replace it.

## Checklist:

### DIFF
--- a/libs/island-ui/core/src/index.ts
+++ b/libs/island-ui/core/src/index.ts
@@ -1,7 +1,7 @@
 // Components
 export * from './lib/Pagination/Pagination'
 export * from './lib/AsyncSearch/AsyncSearch'
-export * from './lib/Button/Button'
+export { Button as ButtonDeprecated } from './lib/Button/Button'
 export * from './lib/Logo/Logo'
 export * from './lib/Typography/Typography'
 export * from './lib/Text/Text'
--- a/libs/island-ui/core/src/index.ts
+++ b/libs/island-ui/core/src/index.ts
@@ -2,6 +2,7 @@
 export * from './lib/Pagination/Pagination'
 export * from './lib/AsyncSearch/AsyncSearch'
 export { Button as ButtonDeprecated } from './lib/Button/Button'
+export * from './lib/ButtonRC/Button'
 export * from './lib/Logo/Logo'
 export * from './lib/Typography/Typography'
 export * from './lib/Text/Text'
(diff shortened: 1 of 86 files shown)


libs/island-ui/core/src/index.ts
```diff
@@ -1,7 +1,7 @@
 // Components
 export * from './lib/Pagination/Pagination'
 export * from './lib/AsyncSearch/AsyncSearch'
-export * from './lib/Button/Button'
+export { Button as ButtonDeprecated } from './lib/Button/Button'
```
darrikonn, line 4: Would like to see the `useDeprecatedComponent` hook inside `Button`.

Since this is a global rename, the name of `ButtonDeprecated` (which is very descriptive) might be missed by developers.

darrikonn (approved):
One comment, otherwise approved!

libs/island-ui/core/src/index.ts
barabrian: good point! this is the first step in switching to the new button, there is still some internal wiring needed in island-ui, we will surely add this in the next reiterations

DavidBachmann (dismissed):
Good phase 1

libs/island-ui/core/src/lib/Button/Button.tsx
```diff
@@ -68,6 +69,9 @@ export const Button = forwardRef<
     },
     ref,
   ) => {
+    if (process.env.NODE_ENV === 'development') {
```
darrikonn, line 72: We already have an existing hook for this.

stjanilofts (dismissed):

libs/island-ui/core/src/lib/private/useDeprecatedComponent.ts
```diff
@@ -1,9 +1,12 @@
...
+const useDeprecatedComponent = (
+  componentName: string,
+  customMsg = `${componentName} is deprecated`,
+) =>
   useEffect(() => {
     if (process.env.NODE_ENV === 'development') {
-      console.warn(`${componentName} is deprecated`)
+      console.warn(customMsg)
     }
   }, [componentName])
```
oddsson, line 11: Can we remove `componentName` from the dependency array, since we are no longer using it in the `useEffect`?

oddsson (changes requested):
See comment ☝️

libs/island-ui/core/src/lib/private/useDeprecatedComponent.ts
barabrian: good point, this should probably be customMsg, I'll cahnge it

oddsson (dismissed):

atligudl (approved):

DavidBachmann (approved):

darrikonn (approved):

siggakogk (approved):
🚢

PR #1261, opened 2020-10-08
Make `yarn postinstall` cross-env

sindrig:
It should work on windows without the linux subsystem

PR #1262, opened 2020-10-08
fix: remove api:contentful-types from postinstall

jeremybarbet:
# remove api:contentful-types from postinstall

— 

## What

- Remove api:contentful-types from postinstall
- Add documentation about it in the cms domain readme

## Why

- So people don't get unwanted changes to the `contentfulTypes.d.ts` when they are running postinstall

## Screenshots / Gifs

— 

## Checklist:


sindrig:
NB already fixed in https://github.com/island-is/island.is/pull/1261

PR #1265, opened 2020-10-08
Accessibility improvements

ingolfurs:
# Accessibility improvements
Fixing accessibility issues reported by Blindrafelagid, Icelandic Association of the Visually Impaired (BIAVI)

## What
- Screen readers will read the language toggle button with more descriptive text than "EN" or "IS", using aria-label with the value "Breyta tungumáli í ensku"
- Side menu made accessible by adding appropriate aria attributes, making the toggle button refer to the menu element it controls. Left and right keyboard arrows are used to navigate between side menu tabs. Had to remove keyboard navigation for frontpage tabs because it conflicted with other keyboard events, and was listening to left/right keyboard click regardless of focus location.
- Frontpage tabs prev / next buttons have appropriate role attributes, required for screen readers
- Tabbing focus order on frontpage tabs will make more sense to keyboard users, by relocating prev/next buttons below carousel content, using absolute style to get them back to it's position
- Frontpage tabs images will not be read by screen readers 
- Category view sidebar and content columns swapped in html but styled so sidebar appears in left side. This way skip to nav link moves focus first to the content of the page and then to the sidebar. Otherwise, keyboard uses first have to tab through all the sidebar links before reaching the actual content
- Category view sub headings should use h3 instead of p element

## Why
- Because accessibility matters and we want to make island.is accessible for everyone

## Screenshots / Gifs
Category headings:
headings

## Checklist:


DavidBachmann (approved):
Fantastic 👏

stjanilofts (commented):
Nice

PR #1268, opened 2020-10-08
Website / Allow articles to be shown in other categories

stjanilofts:
## Description

This allows for articles to be visible in categories/groups/subgroups other than their default ones. Also adds importance field to groups for sorting.

## Why

It is a feature request from the content administrators.

## Checklist:


DavidBachmann (dismissed):
LGTM

DavidBachmann (approved):

stefanhar:
Why are we creating otherCategory and otherGroup, could we maybe change category and group to multiple reference and handle this as arrays?

stjanilofts:
> Why are we creating otherCategory and otherGroup, could we maybe change category and group to multiple reference and handle this as arrays?

stefanhar:
> We could but then we have no default category/group/subgroup to show in the breadcrumbs of the article.

Any reason why we cant just show the first?

stjanilofts:
> > We could but then we have no default category/group/subgroup to show in the breadcrumbs of the article.
> 
> Any reason why we cant just show the first?

For example if the article is in 2 or more categories but also in 2 or more groups, then both of those groups will show up in both of those categories. The way it works now is, it will only show up in those "other" categories/groups/subgroups if they already are being displayed by other articles that have those groups as their main/default groups.

haukurh (approved):
LGTM

PR #1272, opened 2020-10-08
Application system - emphasize radio field

madebynoam:
## What

Add option in the Application System to make radio buttons placed behind a blue background ("emphasize") and "large" in size.

## Why

To match the latest designs

## Screenshots / Gifs

image

## Checklist:


libs/application/core/src/lib/fieldBuilders.ts
```diff
@@ -119,6 +119,8 @@ export function buildRadioField(data: {
   description?: MessageDescriptor | string
   options: Option[]
   required?: boolean
+  emphasize?: boolean
+  largeButtons?: boolean
```
arnifreyrs, line 123: I am just wondering, would you ever want to emphasize without having large buttons? 

what I am trying to get at, is maybe this should just be one attribute? 

I do not have a strong preference, it is fine as is

arnifreyrs (approved):
great stuff, just had one question

libs/application/core/src/lib/fieldBuilders.ts
madebynoam: I thought about this as well :) There will be a case where you want just large buttons without emphasis:
image

But we could simplify, that if you have `emphasize` you won't have to set 'largeButtons'...

arnifreyrs: nono, that is too much. this alone is enough of a case for having both props :)

DavidBachmann (approved):

libs/application/core/src/lib/fieldBuilders.ts
madebynoam: Updated it now.

arnifreyrs (approved):

libs/application/core/src/lib/fieldBuilders.ts
madebynoam: hehe, oops checked it in, it's a one liner in RadioController `<RadioButton large={largeButtons || emphasize}`, so will leave for now :)

DavidBachmann (approved):

PR #1295, opened 2020-10-09
Website/aria labels

ingolfurs:
# \<Description\>

## What

Add missing aria labels to buttons, fix aria-labelledby references
If search form has no label, only placeholder attribute on the search input, we need to add label element that has "srOnly" styles, 
otherwise we get errors because input element and parent div are referring to label through aria-labelledby attribute. 
This way, screen readers will be able to read the label, even though it is not visible on screens.

## Why
We want the site to be accessible in screen readers.

## Checklist:


stjanilofts (dismissed):

jeremybarbet (dismissed):
LGTM

stjanilofts (approved):

PR #1297, opened 2020-10-09
Run tests in minimal image instead of the monster

### LINKED ISSUE #1231: Stop running tests in docker
Prerequisite for #1230. Find a way to a) build simple docker images that can run tests mounted from `.`, b) "hook" into docker-compose network during testing or c) expose different ports for projects (maybe random...?) and feed that to the testrunner.

sindrig:
This closes #1231

# Avoid docker cache in pull request build

## What

Tests that need external dependencies such as db run in docker. We want to avoid using our production image to run the tests.

## Why

Because we like speed

## Checklist:


darrikonn (approved):
🎉

pshomov (approved):
Excellent work!

gudjong (approved):
nice

scripts/Dockerfile-test
```diff
@@ -0,0 +1,12 @@
+FROM node:12.19.0-alpine3.9
+
+WORKDIR "/code"
+VOLUME ["/code"]
```
eirikurn, line 4: This assumes the host has a compatible architecture (and node_modules) with node:alpine. I guess that's mostly true with Github Actions agents?

Don't know if it's an issue, but it will probably cause issues when run by developers on a mac host. One of the reasons for using docker extensively in the CI was ease of reproduction for developers (@pshomov).

Guess we just need to document how devs can reproduce different CI issues.

sindrig: The other route we could go would be to dynamically expose random db port for these tests. I felt like this was easier, but if we run into problems with it I'm open to taking another stab at this.

eirikurn (approved):

PR #1298, opened 2020-10-09
Prettier plugin to sort nx project lists

eirikurn:
## What

Make sure nx project lists are always sorted (in workspace.json, nx.json and tsconfig.base.json).

This is implemented as a prettier plugin, so it is applied every time we run `yarn format`. `yarn generate` also runs prettier on all added/updated files. "Just Works ™️ "

## Why

To reduce the chance of conflicts when teams are generating nx apps and libs.

## Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,6 @@
     "generate": "nx generate",
     "env-secrets": "./scripts/secrets.sh",
     "format": "nx format:write",
-    "format:write": "nx format:write",
     "format:check": "nx format:check",
     "update": "nx migrate latest",
     "workspace-schematic": "nx workspace-schematic",
--- a/scripts/prettier-plugins/sort-projects.js
+++ b/scripts/prettier-plugins/sort-projects.js
@@ -1,0 +1,65 @@
+const { parsers } = require('prettier/parser-babel')
+const parser = parsers['json']
+const _ = require('lodash')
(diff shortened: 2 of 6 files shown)


darrikonn (approved):
Nice

scripts/prettier-plugins/sort-projects.js
```diff
@@ -0,0 +1,65 @@
+const { parsers } = require('prettier/parser-babel')
+const parser = parsers['json']
+const _ = require('lodash')
```
sindrig, line 3: long time no see i love lodash @arnifreyrs

sindrig (approved):
LGTM :tada:

package.json
```diff
@@ -28,7 +28,6 @@
     "generate": "nx generate",
     "env-secrets": "./scripts/secrets.sh",
     "format": "nx format:write",
-    "format:write": "nx format:write",
```
jeremybarbet, line 31: Does the sort script also format at the same time?

eirikurn: This line is just removing a redundant script which I don't see used anywhere in the repo.

But yeah, the sort script runs as part of prettier, and calls the standard prettier json formatter.

arnifreyrs (dismissed):
awesome

PR #1302, opened 2020-10-09
Judicial system/fixes

gudjong:
# Fixes

https://app.asana.com/0/1189271618154957/1197323947467631

## What

Fixes a null pointer in capitalize and uses boolean in api for rejecting field

## Why

These were bugs

## Checklist:


libs/judicial-system/formatters/src/lib/formatters.spec.ts
```diff
@@ -33,31 +33,42 @@ describe('formatDate', () => {
...
+describe('capitalize', () => {
+  test('should return empty string if text is empty', () => {
+    // Arrange
+    let text = (undefined as unknown) as string
+
+    // Act
+    const r = capitalize(text)
+
+    // Assert
+    expect(r).toBeUndefined()
```
oddsson, line 72: Shouldn't this be `expect(r).toEqual("")`?

libs/judicial-system/formatters/src/lib/formatters.ts
```diff
@@ -37,6 +37,10 @@ export function formatDate(
 
 // Credit: https://dzone.com/articles/capitalize-first-letter-string-javascript
 export const capitalize = (text: string): string => {
+  if (!text) {
+    return text
```
oddsson, line 41: I'd think this should be `return ""`

gudjong: ok, makes sense

oddsson (approved):

PR #1225, opened 2020-10-06
chore: change the way to call the postinstall scripts

jeremybarbet:
## What

- Refactor and make a better documentation on how the auto-schemas files work
- Check [this readme](https://github.com/island-is/island.is/blob/c8cc5b7a7c5ad1ea9accdc542be12c5ebd9676c9/SCHEMAS.md) for more info

## Checklist:

- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] Formatting passes locally with my changes
- [x] I have rebased against master before asking for a review

PR #1309, opened 2020-10-10
Generate all schemas

eirikurn:
## What

Make sure we always have all of the schemas. 

## Why

"Affected" logic does not make sense for code generation unless we have all of the schemas in the beginning and always run `generate` with the correct HEAD/BASE as we pull more code. This is tricky for CI. Even more so for developers.

For new developers, the `schemas` command is failing, since there is no HEAD/BASE variables, and there's no affected projects, so it doesn't generate any schemas.

This is a quick fix until most of this logic is replaced in #1225.

## Checklist:

### DIFF
--- a/scripts/generate-schemas.js
+++ b/scripts/generate-schemas.js
@@ -2,15 +2,12 @@ const fs = require('fs')
 const { exec } = require('child_process')
 const { promisify } = require('util')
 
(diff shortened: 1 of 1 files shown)


sindrig (approved):
This probably makes more sense for most cases... 🙏

PR #1311, opened 2020-10-11
Judicial System: Rewrites pdf generation

gudjong:
# PDF generation without headless browser

## What

Rewrites PDF generation using pdfkit.

## Why

Using pug and puppeteer caused problems with Docker deployments.

## Checklist:


apps/judicial-system/api/src/app/modules/case/case.pdf.ts
```diff
@@ -57,114 +62,315 @@ export const formatConclusion = (existingCase: Case) => {
...
-    caseFacts: existingCase.caseFacts,
-    witnessAccounts: existingCase.witnessAccounts,
-    investigationProgress: existingCase.investigationProgress,
-    legalArguments: existingCase.legalArguments,
-    prosecutorName: existingCase.prosecutor.name,
-    prosecutorTitle: existingCase.prosecutor.title,
+  const doc = new PDFDocument({
+    size: 'A4',
   })
+  const stream = doc.pipe(new streamBuffers.WritableStreamBuffer())
```
oddsson, line 85: This is an interesting way to create text :D

oddsson (commented):
Looks awesome! I see that you are not deleting the PUG files you created earlier. Do we want to keep those files?

gudjong:
> Looks awesome! I see that you are not deleting the PUG files you created earlier. Do we want to keep those files?

No, we want to delete them, but not yet. The build is copying the folder that contains them and would fail if we delete it too early.

pshomov (approved):

pshomov:
👍🏻

oddsson (approved):

PR #1312, opened 2020-10-12
island-ui / Figma Storybook

DavidBachmann:
# Figma links in Storybook

## What

Add 'design' addon to Storybook to show live Figma design inside stories

## Why

For easier access

## Screenshots / Gifs

image

## Checklist:


libs/island-ui/core/src/lib/Tabs/Tabs.stories.tsx
```diff
@@ -1,19 +1,27 @@
 import React from 'react'
+import { withDesign } from 'storybook-addon-designs'
 import { Tabs } from './Tabs'
 import { Stack } from '../Stack/Stack'
 import { Typography } from '../Typography/Typography'
 import { ContentBlock } from '../ContentBlock'
 
+const figmaLink =
+  'https://www.figma.com/file/pDczqgdlWxgn3YugWZfe1v/UI-Library-%E2%80%93-%F0%9F%96%A5%EF%B8%8F-Desktop?node-id=208%3A697'
```
pshomov, line 9: Shouldn't this be a config rather?

libs/island-ui/core/src/lib/Select/Select.stories.tsx
```diff
@@ -1,18 +1,26 @@
 import React from 'react'
+import { withDesign } from 'storybook-addon-designs'
 import { Select } from './Select'
 import { Box } from '../Box/Box'
 import { ContentBlock } from '../ContentBlock'
 
+const figmaLink =
+  'https://www.figma.com/file/pDczqgdlWxgn3YugWZfe1v/UI-Library-%E2%80%93-%F0%9F%96%A5%EF%B8%8F-Desktop?node-id=50%3A101'
```
pshomov, line 8: Same here - config?

jeremybarbet (approved):

libs/island-ui/core/src/lib/Tabs/Tabs.stories.tsx
DavidBachmann: I started with a config like so:

```
import { figmaLinks } from '../../../storybook'
const figmaLink = figmaLinks['FormStepper']
```
but changed my mind 🤷 not sure where to put the config.

jeremybarbet: You could create an `environment.ts` file where you define the Figma links array

DavidBachmann: I don't think this warrants an environment file

arnifreyrs: I personally do not think this is a problem unless the designers will change the URLs rapidly. If this turns ugly at some point, we should consider moving these links to a config where it could be simpler to update the links without making a PR going through the CD pipeline.

arnifreyrs (approved):
see my comment. I think this is fine as is

PR #1317, opened 2020-10-12
Skilavottord/add auth

dd-sdang:
# Skilavottord - added auth to routes

## What

- Wrapped protected routes with authentication.
- Includes fix for i18n routing

## Why

- Routes should not be public and needed to be wrapped with authentication (inskranning)
- There was a bug in the language toggler when swicthing between routes

## Checklist:


tommigitgaur (approved):
Fine

PR #1327, opened 2020-10-12
Fully disabling `console` usage for now at least

pshomov:
## What

Disable logging using `console` object

## Why

We want to get everyone to use our standard logging library

## Checklist:


tslint.json
```diff
@@ -27,7 +27,7 @@
     ],
     "no-arg": true,
     "no-bitwise": true,
-    "no-console": [true, "debug", "info", "time", "timeEnd", "trace"],
+    "no-console": true,
```
eirikurn, line 30: We're not using tslint. I think this file can be removed. Configure .eslintrc instead.

eirikurn (changes requested):

tslint.json
pshomov: 👍🏻

arnifreyrs (dismissed):

eirikurn:
@pshomov I see that .eslintrc is not an implicit dependency for all projects, so the CI did not lint all the projects. Did you lint the whole repo and are there no new errors? (Is that what almost destroyed your computer?)

I'd rather not break master with this.

pshomov:
> @pshomov I see that .eslintrc is not an implicit dependency for all projects, so the CI did not lint all the projects. Did you lint the whole repo and are there no new errors? (Is that what almost destroyed your computer?)
> 
> I'd rather not break master with this.

I mean if it did not break the CI then it will not break master if that's the breakage you are referring to.

pshomov:
> @pshomov I see that .eslintrc is not an implicit dependency for all projects, so the CI did not lint all the projects. Did you lint the whole repo and are there no new errors? (Is that what almost destroyed your computer?)
> 
> I'd rather not break master with this.

@eirikurn Just noticed `.eslint.react.json` which seems to be used by front-end apps. Do we want the `no-console` rule there as well? Isn't Sentry the front-end logging platform?

eirikurn:
@pshomov no worries.  .eslint.react.json extends .eslintrc.

pshomov:
> @pshomov no worries. .eslint.react.json extends .eslintrc.

I am aware, my question was if we would like to allow some console usage in the frontend.

eirikurn:
Ah. Then no. We don't want any console.log in production by default.

eirikurn (approved):
Approved pending full-repo lint.

oddsson:
Can we put a automerge label on this so it get's merged?

Issue #1315, opened 2020-10-12
Remove sass support


darrikonn:
There are still some Sass files in our codebase, as well as `node-sass` dependency.

These sass files are located in:
```
apps/judicial-system/web/src/styles.scss
apps/application-system/form/src/styles.scss
```
^ might be autogenerated by `nx`. Not sure if this is used?

darrikonn:
@arnifreyrs assigning you since the `application-system` file is from you.

darrikonn:
@gudjong ^ same

gudjong:
@oddsson ^ can you look at this?

oddsson:
The reason we're using scss in the judicial-system project is for @font-face declarations. It's currently not possible to do this with Treat yet (source: https://github.com/seek-oss/treat/issues/93). NextJS apps have a way of doing this but that does not apply to plain React apps. @jeremybarbet suggested we'd have a root level .css file that projects import for this instead, that just hasn't been implemented. I'll put it on my backlog and fix it this week.

PR #1330, opened 2020-10-12
Remove sass support

oddsson:
# Remove sass support

https://github.com/island-is/island.is/issues/1315

## What

Importing font-face declarations is currently not supported in Treat. NextJS projects have their way of doing this but there has not been a good way to do this until now. In this PR, we are creating a root level css file with font-face declarations that individual apps can import, rather than creating separate declaration for each project. Since we don't use sass in any app, we also remove the support for it, to clean up dependencies.

## Why

Consistency between projects and one less dependency to worry about. 

## Checklist:


gudjong (approved):

sindrig (dismissed):
LGTM

apps/shared/fonts.css
```diff
@@ -0,0 +1,45 @@
+@font-face {
+  font-family: 'IBM Plex Sans';
+  font-style: normal;
+  font-weight: 300;
+  font-display: swap;
+  src: local('IBM Plex Sans Light'), local('IBMPlexSans-Light'),
+    url('./fonts/ibm-plex-sans-v7-latin-300.woff2') format('woff2'),
+    url('./fonts/ibm-plex-sans-v7-latin-300.woff') format('woff');
```
eirikurn, line 8: Are these references processed by webpack? I.e. does it bundle the fonts on build?

apps/application-system/form/src/styles.css
```diff
@@ -1,3 +1,5 @@
+@import '../../../shared/fonts.css';
+
 @font-face {
   font-family: 'IBM Plex Sans';
   font-style: normal;
```
eirikurn, line 5: Remove these?

dd-sdang (approved):
Looks good

apps/shared/fonts.css
```diff
@@ -0,0 +1,45 @@
+@font-face {
```
eirikurn, line 1: I'd like this to be in a lib, like `libs/island-ui/fonts`.

Can that be referenced in a styles array in workspaces.json, or from css like this:

```
@import '@island.is/island-ui/fonts';
```

I guess not, since these path aliases are defined in tsconfig rather than webpack, but worth trying.

arnifreyrs (approved):
Great!

apps/shared/fonts.css
oddsson: To be honest, I don't really know for sure. First guess is _no_ because neither the Webpack config in the Judicial system project f.x. nor the nrwlConfig webpack config handles font files. However, the Judicial system builds without a problem and generates font files in `/dist/apps/judicial-system` and `nx build` is running webpack under the hood. I'm guessing there is some magic happening here and that they actually are, even though I can't find where it's being done 🤷‍♂️

apps/shared/fonts.css
oddsson: I've moved this to libs but was unable to import like you mention 😕

apps/judicial-system/web/src/styles.css
```diff
@@ -0,0 +1,31 @@
+@import '../../../../libs/island-ui/fonts/fonts.css';
```
arnifreyrs, line 1: doesnt this have to be `@import '@island.is/fonts/fonts.css'`

oddsson: I believe it is, everything is building without an error, the fonts are being downloaded and everything seems to work.

jeremybarbet (approved):
Very nice!

atligudl (approved):

gudjong (approved):

dd-sdang (approved):

sindrig (approved):

jeremybarbet (approved):

PR #1333, opened 2020-10-12
island-ui / Icon RC

barabrian:
# Icon RC

## What

Convert ionicons designer pack to react components.
Use one single component to lazyload desired icon

## Why

Decision has been made to use Ionicons, current icon component is limited, hard to maintain and not using the same assets that are used by designers (mostly viewbox related)

## Screenshots / Gifs

icons
icon-list

## Checklist:


libs/island-ui/core/src/lib/IconRC/icons/Accessibility.tsx
```diff
@@ -0,0 +1,28 @@
+import * as React from 'react'
+interface SVGRProps {
```
jeremybarbet, line 2: To make it easier to maintain if we need to change it, I would probably move this interface and the type out of each icon so you can have:

```
import { IconType } from '../Icon'

function SvgAccessibility({ title, titled, ...props }): IconType
```

The replace all feature will be very useful in this case 😅

oddsson (dismissed):
Awesome 😛

libs/island-ui/core/src/lib/IconRC/Icon.tsx
```diff
@@ -0,0 +1,66 @@
...
+  color = 'currentColor',
+  size = 'medium',
+  className,
+  title,
+  titleId,
+}: IconProps & IconPropsType) => {
+  const path = iconMap[type][icon]
+  const IconSvg = React.lazy(() => import('./icons/' + path))
+  const optionalProps: OptionalProps = {}
+  if (className) optionalProps.className = className
```
jeremybarbet, line 54: For better readability, I would follow the "curly brace" convention here

DavidBachmann:
+46,032 −3 😂

libs/island-ui/core/src/lib/IconRC/icons/Accessibility.tsx
barabrian: for consistency we could make a [custom template](https://react-svgr.com/docs/custom-templates/) for the generator and maybe keep the file in the Icon folder if the icon pack changes

jeremybarbet: That would make sense! It would overwrite all the react components with the svg you are providing as entry files?

barabrian: I removed all the svg files, wasn't sure they belonged in the UI library. I could include them in the Icon folder for now and moving forward we can relocate them if we want.

jeremybarbet: I think it's fine to not have them in the repo. I would just add the documentation on how to generate the react components and the template for svgr

jeremybarbet (approved):

PR #1346, opened 2020-10-13
Website / Taktikal fill and sign modals on process entry

stjanilofts:
# Description

These modals were used on the old website but are now only external links. This will make the modals work again.

https://app.taktikal.is/fill-and-sign-documents/ffd3425725b4
https://app.taktikal.is/docs/fill-and-sign-integration

## What

Adds Taktikal Drop & Sign npm package and implements its openDocument method to open process entry links that have a dropSignKeyFile. If the value exists it will open the drop and sign modal, otherwise it will open the default process entry link.

## Why

It is a better user experience. Check out article /dreifing-osku to test it out.

## Checklist:


pshomov (dismissed):
LGTM

package.json
```diff
@@ -73,6 +73,7 @@
     "@sentry/react": "^5.22.3",
     "@sentry/tracing": "^5.22.3",
     "@sindresorhus/slugify": "1.0.0",
+    "@taktikal/fillandsign": "^3.1.0",
```
pshomov, line 76: Actually I think there is a preference to pin down the versions of the packages as the one above.

eirikurn (commented):
This logic (and css) should be lazyloaded on demand, if possible.

package.json
eirikurn: We should add a yarn config file that enforces this as a default.

stjanilofts:
> This logic (and css) should be lazyloaded on demand, if possible.

I made some changes. I think it should be better now. However, the initial animation when the modal opens does not work at the moment. Using animations in globalStyle seems to be bugged: https://github.com/seek-oss/treat/issues/137

darrikonn (approved):

libs/island-ui/contentful/src/lib/ProcessEntry/ProcessEntry.tsx
```diff
@@ -21,18 +26,61 @@ export const Titles: { [k: string]: string } = {
...
+    ...(!dropSignFileKey && {
+      href: processLink,
+      icon: 'external' as IconProps['type'],
+    }),
+    ...(dropSignFileKey && {
+      onClick: () => openDocument(dropSignFileKey),
+    }),
+  }
+}
+
```
stefanhar, line 58: This logic should probably be in it's own fill and sign component.
My concern is this might diverge from process entry both in design and functionality
e.g we might want to display an icon for fill and sign.

We might also use other similar services in future adding every service into this component might get complex.

Maybe the process entry component should take in props so we can wrap it with a fill and sign component?

stefanhar (changes requested):

PR #1349, opened 2020-10-13
Judicial System: Displays a special message if user cancels electronic signature

gudjong:
# Electronic Signature Cancel Message

## What

Displays a special cancel message if the user cancels an electronic signature.

## Why

This is part of the requirements.

image

## Checklist:


apps/judicial-system/api/src/app/modules/case/case.controller.ts
```diff
@@ -255,7 +273,22 @@ export class CaseController {
       )
     }
 
-    return this.caseService.confirrmSignature(existingCase, documentToken)
+    try {
+      const response = await this.caseService.confirrmSignature(
```
oddsson, line 277: Typo... confirrmSignature 🖌 confirmSignature

apps/judicial-system/web/src/routes/Judge/Confirmation/Confirmation.tsx
```diff
@@ -406,15 +406,17 @@ export const Confirmation: React.FC = () => {
       {modalVisible && (
         <Modal
           title={
-            confirmSignatureResponse?.httpStatusCode > 400
-              ? 'Auðkenning tókst ekki'
+            confirmSignatureResponse?.httpStatusCode >= 400
+              ? confirmSignatureResponse?.code === 7023
```
oddsson, line 410: Consider creating an enum for dokobit error codes

oddsson (changes requested):

apps/judicial-system/api/src/app/modules/case/case.controller.ts
gudjong: Fixed

apps/judicial-system/web/src/routes/Judge/Confirmation/Confirmation.tsx
gudjong: Just commented for now.

oddsson (approved):

PR #1350, opened 2020-10-13
Fix generate schemas

sindrig:
## What

Fix generate schemas. Change how we detect the root folder.

## Why

It stopped working when we moved the script that detects changed files down one level.


jeremybarbet (approved):
🎉

eirikurn (approved):

PR #1359, opened 2020-10-14
Add pre-check for checking if we should be running push

sindrig:
See https://github.com/island-is/island.is/runs/1254033281?check_suite_focus=true for a branch that should not run push
See https://github.com/island-is/island.is/runs/1254041876?check_suite_focus=true for a branch taht should run push

Bad thing is that this will create a "red" build. Don't know if/how we can do anything about it, or if we should simply skip this and only use workflow trigger


pshomov (approved):
🙏🏻

PR #1365, opened 2020-10-15
Allow contentful documents as messages and display with <Richtext> co…

magnearun:
…mponent

TODO find a better way to mark richtext messages in defineMessage, atm it will asume it is richtext if defaultMessage is left out.

# \<Description\>

## What

Localized richtext from contentful using translations lib

## Why

We want some messages to be richtext

## Screenshots / Gifs

## Checklist:

### DIFF
--- a/apps/application-system/form/src/routes/Application.tsx
+++ b/apps/application-system/form/src/routes/Application.tsx
@@ -7,7 +7,12 @@ import useAuth from '../hooks/useAuth'
 export const Application = () => {
   const { id } = useParams()
   const { userInfo } = useAuth()
-  useNamespaces(['dl.application', 'pl.application', 'application.system'])
+  useNamespaces([
+    'dl.application',
+    'pl.application',
+    'application.system',
+    'example.application',
+  ])
 
   const nationalRegistryId = userInfo?.profile?.natreg
 
--- a/apps/application-system/form/tsconfig.json
+++ b/apps/application-system/form/tsconfig.json
@@ -5,7 +5,8 @@
     "allowJs": true,
     "esModuleInterop": true,
     "allowSyntheticDefaultImports": true,
-    "types": ["node", "jest"]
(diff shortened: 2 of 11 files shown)


eirikurn (changes requested):
I don't think we want to depend on Contentful's rich text system for system translations in all our projects? 

I think we should keep richtext system translations fairly limited, to not load up a bunch of unused code on every screen. Just bold, italic, links and maybe lists?

Contentful richtext ends up as bloated propriatery json. I also assume `@island.is/island-ui/contentful#richtext` will become quite big to support a ton of stuff for island.is content. (though we could maybe split it up into SimpleRichText vs ExtendedRichText)

Maybe it should just be markdown. Then it's always a string, but we'll need to implement (and load) a secondary markdown richtext renderer.  But the rendering side could control if/how it wraps translations in a markdown renderer (some screens/projects may not need it).

I would also prefer to keep the defaultValue somehow. It's useful when we run against mock api.

magnearun:
@eirikurn I agree that we might not want to depend on the Contentful richtext system. We could use richtext formatting like here https://formatjs.io/docs/react-intl/components/#rich-text-formatting , we could extend this to use Markdown as well.
Skipping the defaultValue is a ugly hack to tell the extract script that it should mark this message as richtext to know in the translation extension that it should render the Contentful richtext editor, this would not be a problem if we skip using the contentful richtext system.

eirikurn:
Oh nice, FormattedMessage looks really good. I like especially that each context can control which tags are allowed and how they are rendered. And not needing an extra markdown parser/renderer.

I think we can expect more technical knowledge from system translation editors, to manually write markdown or HTML tags, and understanding the syntax for variable tokens, plurality, etc.

magnearun:
If the tags b, i, li are used in the string and one of them is not provided in the values it will not format the message. You can play around with it in the link above, it has a live editor

PR #1368, opened 2020-10-15
Dev path for graphql in service portal

Thorrik58:
We dont want to use the normal basepath while developing, this fixes that problem by checking location.

## What

Fixing graphql fetch on localhost

## Why

Nothing is being fetched locally

## Screenshots / Gifs

## Checklist:

### DIFF
--- a/libs/service-portal/graphql/src/lib/client.ts
+++ b/libs/service-portal/graphql/src/lib/client.ts
@@ -9,8 +9,13 @@ import { onError } from '@apollo/client/link/error'
 import { RetryLink } from '@apollo/client/link/retry'
 import { setContext } from '@apollo/client/link/context'
 
+const apiUri =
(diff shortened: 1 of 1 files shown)


libs/service-portal/graphql/src/lib/client.ts
```diff
@@ -9,8 +9,13 @@ import { onError } from '@apollo/client/link/error'
 import { RetryLink } from '@apollo/client/link/retry'
 import { setContext } from '@apollo/client/link/context'
 
+const apiUri =
```
atligudl, line 12: I think this would be described better as a runtime config instead of doing a location check

Thorrik58: We could do that

obmagnusson (approved):
LGTM

atligudl (approved):

PR #1372, opened 2020-10-15
Fix master build

sindrig:
Seems to build for now: 

https://github.com/island-is/island.is/actions/runs/309564817

and

https://github.com/island-is/island.is/actions/runs/309656157


pshomov (approved):
LGTM

pshomov (approved):

PR #1394, opened 2020-10-16
Only be allowed to write answers and external data that are permitted…

arnifreyrs:
… in current application state

Remove external id from application
Remove owner id attribute from Form
Form ID can be a string
Form Fields do no longer have the required attribute

## Why

A user with a certain role in some state might not be allowed to overwrite answers and external data that some other user has already written to the application.

## Checklist:


magnearun (approved):
LGTM

atligudl (approved):

PR #1418, opened 2020-10-19
Use anchor instead of nextjs link to make use of the redirect on the index page

dabbeg:
## What

Using anchors instead of Link to make the browser refresh

## Why

Because that will trigger the redirect on the ingress to redirect all pages at loftbru.island.is -> island.is/loftbru
You can check the redirect out on https://loftbru.dev01.devland.is

## Checklist:


apps/air-discount-scheme/web/components/Header/Header.tsx
```diff
@@ -39,16 +36,13 @@ function Header({ routeKey, localeKey }: PropTypes) {
   const { toRoute, activeLocale, switchLanguage } = useI18n()
 
   const nextLanguage = activeLocale === 'is' ? 'en' : 'is'
+  const nextPath = activeLocale === 'is' ? '/' : '/en'
```
darrikonn, line 39: This should actually be `'is' ? 'https://island.is/loftbru' : 'https://island.is/en/lower-airfares-for-residents-in-rural-areas'`

darrikonn (changes requested):

barabrian (approved):
Not sure what redirect is on the index page but this looks fine to me

apps/air-discount-scheme/web/components/Header/Header.tsx
dabbeg: The redirect will handle that for us

darrikonn: aaaahhh sooooo smaaaaart

darrikonn (approved):

apps/air-discount-scheme/web/components/Header/Header.tsx
darrikonn: What about local development?

PR #1425, opened 2020-10-19
GraphQL API mocks and mocking library

eirikurn:
## What

Create API mocks for the GraphQL API so that frontend projects can be started without any API dependencies.

Uses [Mock Service Worker](https://mswjs.io/) behind the scene with a custom GraphQL handler and mocking helpers.

This is split into two libs:

`libs/api/mocks` has mock resolvers and data for the main GraphQL API, which can be integrated into any web frontend which uses the API. Currently, it only includes mocks for GraphQL domains and resolvers used by the `web` project.

`libs/shared/mocking` has mocking utilities which can be used to mock different APIs.

Mocks are removed from production bundles and only apply with an environment variable: `API_MOCKS=true`

## Why

This is critical for running browser-based E2E tests in CI (without setting up databases and APIs) but also for open source contributors who do not have access to our private APIs. It's also very nice when working on frontend tasks to not need to have any backend solutions and databases running locally.

## Why MSW, not MirageJS?

Mock APIs should share the same schema as the real API, and we should use TypeScript to enforce this, i.e. mock data should match schema object types, and resolvers should match schema query types. MirageJS's "simple" database and factory system made this pretty much impossible.

This makes DX much better (type checking and auto-complete in object types, inputs and payloads), and the CI will catch many cases where an API change breaks the mocks.

MSW also has a few large benefits over MirageJS:

* It works in Node.JS, so we can use it to mock API calls in the backend.
* That means it also works in NextJS server rendering, sharing the same mocks on server and client. (MirageJS only supports client-side API requests in NextJS).
* It provides a much better developer experience since it uses real requests in the browser, as these show up in the network tab.

I still miss MirageJS's database, managing mock data separately from API handlers to support mutating data, smarter handlers and different testing environments. So I implemented a similar system with `createFactory` and `createStore`, which integrate better with TypeScript and GraphQL schemas.

For more information, check out the README files.

## Future work

- Set up more Cypress tests.
- Configure CI to run E2E tests.
- Support piping mocked requests through Cypress so that tests can override data and resolvers. ([like Mirage does](https://miragejs.com/quickstarts/cypress/))
- Support some kind of "MOCK_ENV" variable in .env and browser query, to enable different mock data scenarios.
- Integrate Service Portal.
- Integrate Application System forms.

## Screenshots / Gifs

https://share.getcloudapp.com/geuqp8bQ

Screenshot 2020-10-19 at 22 15 41
Screenshot 2020-10-19 at 22 17 35
Screenshot 2020-10-19 at 22 20 20
Screenshot 2020-10-19 at 22 16 34
Screenshot 2020-10-19 at 22 20 52
Screenshot 2020-10-19 at 22 21 35
Screenshot 2020-10-19 at 22 22 19

## Checklist:

- [x] I have commented on my code, particularly in hard-to-understand areas

## Why are you working? Didn't you just have a baby?

Yeah. I was actually wrapping up this work just hours before the water broke. I really hope we can get this merged. 😊


arnifreyrs (approved):
Amazing stuff!

PR #1429, opened 2020-10-20
island-ui/ CategoryCard

DavidBachmann:
# Create CategoryCard

## Screenshots / Gifs

image

## Checklist:


libs/island-ui/core/src/lib/CategoryCard/CategoryCard.tsx
```diff
@@ -0,0 +1,74 @@
...
+  href?: string
+  onClick?: () => void
+}
+
+type CategoryCardProps = {
+  heading: string
+  text: string
+  tags?: Tag[]
+  href?: string
+  colorScheme?: 'blue' | 'red'
```
barabrian, line 19: should this be more descriptive towards the main usage, like 'primary', 'secondary' or anything that would make more sense regarding how this should be used?

libs/island-ui/core/src/lib/CategoryCard/CategoryCard.tsx
```diff
@@ -0,0 +1,74 @@
...
+    >
+      <Text variant="h3" color={textColor}>
+        {heading}
+      </Text>
+      <Text paddingTop={1}>{text}</Text>
+      {hasTags && (
+        <Box paddingTop={3}>
+          <Inline space={['smallGutter', 'smallGutter', 'gutter']}>
+            {tags.map((tag) => (
+              <Tag variant={tagVariant} href={tag.href} onClick={tag.onClick}>
```
barabrian, line 65: should the main card be clickable or is it mainly a frame for the tags?

barabrian (commented):
some minor comments

libs/island-ui/core/src/lib/CategoryCard/CategoryCard.tsx
DavidBachmann: It should be clickable with clickable tags inside

libs/island-ui/core/src/lib/CategoryCard/CategoryCard.tsx
DavidBachmann: When we know exactly how this will be used we should do that. Right now we only know of 2 projects that will use this component and each of these projects wants a different color, hence the two available ones.

DavidBachmann: image

image

barabrian (approved):
LGTM

PR #1446, opened 2020-10-20
Animated loading icon

stjanilofts:
## What

An animated loading icon.

## Why

Requested by designers.

## Checklist:


darrikonn (dismissed):
Snake remake ✅

darrikonn (dismissed):

package.json
```diff
@@ -84,6 +84,7 @@
     "@types/styled-components": "^5.1.4",
     "@xstate/react": "0.8.1",
     "amazon-s3-uri": "^0.0.3",
+    "animejs": "^3.2.1",
```
barabrian, line 87: we should probably lock the package version
```suggestion
    "animejs": "3.2.1",
```

barabrian (commented):
I would like to see a discussion regarding animation library and which one to use rather than including multiple, but for now I think this is fine as long as we change it if we make a decision on which library to use

barabrian (approved):

darrikonn (approved):

PR #1405, opened 2020-10-19
Feature/api catalog

saevarma:
# Viskuausan 

## What
First merge for Viskuausan into the master branch. 
Getting Viskuausan into the devland environment.

## Why

To prepare for the release of Viskuausan

## Checklist:

- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] Formatting passes locally with my changes
- [X] I have rebased against master before asking for a review

PR #1460, opened 2020-10-21
island-ui/ Responsify Column component 

DavidBachmann:
# Responsify Column component by making it handle array of widths

## Why

To be able to set widths like `['1/3', '1/5', '3/12', 'content']`, see https://github.com/island-is/island.is/pull/1405#discussion_r509122773

## Checklist:


DavidBachmann:
Won't merge:

After discussion we have found that people are MOSTLY using `Columns` where they should be using `Grid`. We will look to deprecate this component instead of adding more features.

PR #1474, opened 2020-10-21
island-ui/ ModalBase + DialogPrompt

DavidBachmann:
# Modal solutions

## What

Basic Modal foundation and DialogPrompt, a dialog that presents the user with a yes/no choice, that builds on BaseModal.

## Screenshots / Gifs

image

## Checklist:


DavidBachmann:
Discussion to be had before merging:

**Should we also provide a more basic modal component and build DialogPrompt on that?**
My opinion: Reakit's Dialog is a great building block for projects that need more advanced modals. island-ui shouldn't try to abstract that any more than Reakit is already doing.

libs/island-ui/core/src/lib/DialogPrompt/DialogPrompt.treat.ts
```diff
@@ -0,0 +1,69 @@
+import { style } from 'treat'
+import { theme } from '@island.is/island-ui/theme'
+
+const hexToRgb = (hex: string) => {
```
jeremybarbet, line 4: I've seen this in multiples places, I would move it to utils and export it from islandui

jeremybarbet (dismissed):

libs/island-ui/core/src/lib/DialogPrompt/DialogPrompt.treat.ts
DavidBachmann: You're right. A util like
```
const hexToRgba = (hex: string, alpha: number) => {
  // Requires 6 number hexadecimal
  const rgb = /^#?([a-f\d]{2})([a-f\d]{2})([a-f\d]{2})$/i.exec(hex)

  if (!rgb) {
    return null
  }

  const r = parseInt(rgb[1], 16)
  const g = parseInt(rgb[2], 16)
  const b = parseInt(rgb[3], 16)

  return `rgba(${r}, ${g}, ${b}, ${alpha})`
}
```

barabrian:
> Discussion to be had before merging:
> 
> **Should we also provide a more basic modal component and build DialogPrompt on that?**
> My opinion: Reakit's Dialog is a great building block for projects that need more advanced modals. island-ui shouldn't try to abstract that any more than Reakit is already doing.

I agree Reakit is doing a-lot for us, but I think if there are more components going to use this pattern I think abstracting the core boilerplate and styles for a modal component could be useful, but that totally depends on what the designers have in mind 😄

libs/island-ui/core/src/lib/DialogPrompt/DialogPrompt.treat.ts
jeremybarbet: I would probably re-use where it was defined locally in each projects at the same time

DavidBachmann: I don't want to touch other projects in this particular PR. Will make another for that.

DavidBachmann:
> I agree Reakit is doing a-lot for us, but I think if there are more components going to use this pattern I think abstracting the core boilerplate and styles for a modal component could be useful, but that totally depends on what the designers have in mind 😄

Sold. I'll experiment with that.

libs/island-ui/core/src/lib/DialogPrompt/DialogPrompt.tsx
```diff
@@ -0,0 +1,118 @@
...
+export const DialogPrompt = ({
+  title,
+  description,
+  ariaLabel,
+  baseId,
+  disclosureElement,
+  onConfirm,
+  onCancel,
+  buttonTextCancel = 'Cancel',
+  buttonTextConfirm = 'Confirm',
```
barabrian, line 44: not sure about using defaults here, maybe we could use this to conditionally show the buttons, maybe some cases only have one button or no buttons

barabrian (commented):
LGTM! I only have one comment we could discuss

libs/island-ui/core/src/lib/DialogPrompt/DialogPrompt.tsx
```diff
@@ -0,0 +1,118 @@
...
+}: DialogPromptProps) => (
+  <ModalBase
+    disclosure={disclosureElement}
+    baseId={baseId}
+    aria-label={ariaLabel}
+    className={styles.dialog}
+  >
+    {({ closeModal }) => {
+      const handleClose = () => {
+        onCancel && onCancel()
```
jeremybarbet, line 54: You can do 
```suggestion
        onCancel?.()
```

libs/island-ui/core/src/lib/DialogPrompt/DialogPrompt.tsx
jeremybarbet: I agree, we should be able to hide the buttons if needed. But I like having default values 😅

barabrian: if we use defaults, wouldn't we need to make it translate-able?

jeremybarbet: Could point, and we don't want to handle translation inside the island-ui, probably good to stick to what you said above then 👍

libs/island-ui/core/src/lib/DialogPrompt/DialogPrompt.tsx
DavidBachmann: Sorcery

libs/island-ui/core/src/lib/DialogPrompt/DialogPrompt.tsx
DavidBachmann: Addressed in https://github.com/island-is/island.is/pull/1474/commits/97880b29409df6fc36c69d4fdd0e8725a71c4a39

barabrian (approved):
Nice! LGTM

barabrian (approved):

PR #1476, opened 2020-10-21
Adding common libs needed for api-catalogue and xroad-collector

saevarma:
# Common libs for api-catalogue and xroad-collector

## What

To break down PR #1405  we are creating smallar PR with isolated units.
This contains the libs part that are common for api-catalogue and xroad-collector

## Why

To get api-catalogue and xroad-collector to master but in smaller PRs

## Checklist:


libs/api-catalogue/elastic/src/lib/types/types.model.ts
```diff
@@ -0,0 +1,41 @@
+interface ShardsResponse {
```
jeremybarbet, line 1: Is this file automatically generated?

Addips: No as of now the definitions are not available from elasticsearch, but this based on TypeScript support description here: https://www.elastic.co/guide/en/elasticsearch/client/javascript-api/current/typescript.html

jeremybarbet: I see, then it should be similar to this one used on the api/elastic-search https://github.com/island-is/island.is/blob/master/libs/api/content-search/src/types/index.ts I wonder if we could share some part of this file somewhere instead of typing it locally on each project

Addips: Yes, I agree, would be nice to have a lib or something of that sorts for the shared types :)

saevarma: So should we maybe create a island-ui/elastic library that other libraries and projects can reuse these basics types?

jeremybarbet: Since it's only types, it might be confusing to create a whole lib for it. I wonder if it could be instead inside `libs/island-shared/types/elastic.d.ts` for example

saevarma: Sounds good, I will move it there and update the types used in api/content-search as well

arnifreyrs: I like that idea @jeremybarbet

saevarma: there is a  `libs/shared` with the two libs `form-fields` and `mocking`. Any preference on adding the `types` library there or in a new `libs/island-shared/` ?

arnifreyrs: let's use `libs/shared` for this

apps/service-portal/src/components/GlobalModules/GlobalModules.tsx
```diff
@@ -0,0 +1,80 @@
+import {
```
arnifreyrs, line 1: is this file intended to be part of this PR? how does service-portal and api-catalogue connect?

arnifreyrs (commented):
There are various changes to apps that I am unsure if should be part of this PR

apps/service-portal/src/components/GlobalModules/GlobalModules.tsx
saevarma: I did a `yarn format` to format my changes, admit I reviewed the staged file in a hurry. There is no connection and should no be a part of this PR. I will revert and fix this. Well spotted!

PR #1484, opened 2020-10-22
Be able to submit writable answers when submitting the application.

arnifreyrs:
# Submit application fixes and improvements

## What

Rework the submit field so it can take over the submit button in the footer if desired.

Move footer logic into a separate component.

Hide go back button when on the first screen in the flow.

## Why

Necessary improvements and fixes

## Checklist:


libs/application/templates/reference-template/src/forms/ExampleForm.ts
```diff
@@ -114,11 +117,24 @@ export const ExampleForm: Form = buildForm({
...
+                { event: 'SUBMIT', name: 'Senda inn umsókn', type: 'primary' },
+              ],
+            }),
+            buildIntroductionField({
+              id: 'overview',
+              name: 'Takk fyrir að sækja um',
+              introduction:
+                'Með því að smella á "Senda" hér að neðan, þá sendist umsóknin inn til úrvinnslu. Við látum þig vita þegar hún er samþykkt eða henni er hafnað.',
+            }),
+          ],
```
arnifreyrs, line 137: @fjandakornid great spot in the other PR, I fixed it here.

Now you need to include the submit field somewhere, otherwise the application will never really be submitted...We can discuss the pros and cons here.

libs/application/ui-shell/src/components/ScreenFooter.tsx
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  function getSubmitButtonText(): StaticText {
+    if (submitField?.placement === 'footer') {
+      const { actions } = submitField
+      if (actions.length === 1) {
+        return actions[0].name
+      }
+    }
+    return {
```
magnearun, line 48: I think you might need to define this message with
` const submitText = defineMessage({
      id: 'application.system:button.submit',
      defaultMessage: 'Submit',
      description: 'Submit button text',
})
`
and then return that for the extract script to work, not sure though, but you could try and see if this message appears in the messages.json in the localization lib.

magnearun (dismissed):

libs/application/ui-shell/src/components/ScreenFooter.tsx
arnifreyrs: yeah, I moved this just from the render function really. I think we need to take a look at these strings in a future PR when we will apply stricter rules on translations before going live :)

libs/application/templates/driving-lessons/src/forms/DrivingLessonsApplication.ts
```diff
@@ -217,10 +220,28 @@ export const DrivingLessonsApplication: Form = buildForm({
...
+              name: 'sick',
+              actions: [
+                {
+                  event: 'SUBMIT',
+                  name: 'Smelltu hér til að senda inn umsókn',
+                  type: 'primary',
+                },
+              ],
+            }),
+            buildIntroductionField({
```
madebynoam, line 239: Maybe we'll want to re-name `IntroductionField` in the future since it's not just used as an introduction. Something like `buildExplanationField`?

madebynoam (dismissed):

libs/application/templates/driving-lessons/src/forms/DrivingLessonsApplication.ts
arnifreyrs: yeah, or StaticContentField or something. It is definitely not `IntroductionField` any more :)

madebynoam (dismissed):

libs/application/templates/reference-template/src/forms/ExampleForm.ts
fjandakornid: Looks good to me :+1: Will do.

magnearun (approved):

PR #1497, opened 2020-10-22
Feature/catalogue xroad collector common libs

saevarma:
# Common libs for api-catalogue and xroad-collector

## What

To break down PR #1405  we are creating smallar PR with isolated units.
This contains the libs part that are common for api-catalogue and xroad-collector

## Why

To get api-catalogue and xroad-collector to master but in smaller PRs

## Checklist:


arnifreyrs (dismissed):
LGTM

libs/api-catalogue/services/src/lib/provider.service.ts
```diff
@@ -0,0 +1,68 @@
...
+                xroadInstance: item.id?.xroadInstance ?? '',
+                memberClass: item.id?.memberClass ?? '',
+                memberCode: item.id?.memberCode ?? '',
+                subsystemCode: item.id?.subsystemCode ?? '',
+              }
+            },
+          )
+      }
+    } catch (err) {
+      exceptionHandler(err)
```
pshomov, line 61: I am concerned with this pattern of swallowing exceptions and returning successfully. Aren't these critical exceptions? Shouldn't they be propagated?

libs/api-catalogue/elastic/src/lib/queries/search.model.ts
```diff
@@ -0,0 +1,101 @@
...
+              query: query + '*',
+              fields: ['name', 'owner', 'description'],
+              // eslint-disable-next-line @typescript-eslint/camelcase
+              analyze_wildcard: true,
+            },
+          },
+          {
+            bool: {
+              should,
+              // eslint-disable-next-line @typescript-eslint/camelcase
```
darrikonn, line 82: Would like to see this globally ignored for this library (in the .eslintrc config), since snake_case is a requirement for the typescript version of elasticsearch

libs/api-catalogue/services/src/lib/restmetadata.service.ts
```diff
@@ -0,0 +1,173 @@
...
+   * @param xroadIdentifier Object identifying the service in X-Road
+   * @returns OpenAPI object if spec is found, otherwise null
+   */
+  async getOpenApi(xroadIdentifier: XroadIdentifier): Promise<OpenApi | null> {
+    try {
+      return YamlParser.safeLoad(
+        await this.getOpenApiString(xroadIdentifier),
+      ) as OpenApi
+    } catch (err) {
+      exceptionHandler(err)
```
pshomov, line 96: Ditto

libs/api-catalogue/services/src/lib/restmetadata.service.ts
```diff
@@ -0,0 +1,173 @@
...
+
+          if (map) {
+            map.push(mappedItem)
+          } else {
+            serviceMap.set(serviceCode, [mappedItem])
+          }
+        }
+      })
+    } catch (err) {
+      exceptionHandler(err)
```
pshomov, line 168: Ditto

package.json
```diff
@@ -155,6 +155,7 @@
     "next-treat": "1.3.0",
     "nodemailer": "^6.4.13",
     "oidc-client": "^1.10.1",
+    "openapi3-ts": "^2.0.0",
```
darrikonn, line 158: pin the version.

libs/api-catalogue/services/src/lib/api-catalogue-services.ts
```diff
@@ -0,0 +1,3 @@
+export function apiCatalogueServices(): string {
```
darrikonn, line 1: Looks like this function isn't used, but was gonna comment on the name.
`getApiCatalogueServices` > `apiCatalogueServices`
(or change this to a constant)

darrikonn (changes requested):

libs/api-catalogue/services/src/lib/provider.service.ts
saevarma: This is a valid point. But the XRoad api will throw exceptions if there is some connection issues with other Xroad clients or if rest service are registered manually instead of using openapi. We can solve these issues if we add usage of the new Admin API from XRoad and that is on the roadmap for these services. But until we can do that the idea was to catch these errors, log them but return empty answers, so the whole process won't crash because of one client or one service being misconfigured. 

But I can change the exceptionHandler function to throw the error if it is not XRoad communication error, so other unexpected errors won't be swallowed?

libs/api-catalogue/services/src/lib/api-catalogue-services.ts
saevarma: Removed, was just autogenerated with new lib and forgot to remove

package.json
saevarma: Done

libs/api-catalogue/elastic/src/lib/queries/search.model.ts
saevarma: Done

package.json
```diff
@@ -155,7 +155,7 @@
     "next-treat": "1.3.0",
     "nodemailer": "^6.4.13",
     "oidc-client": "^1.10.1",
-    "openapi3-ts": "^2.0.0",
+    "openapi3-ts": "2.0.0",
```
darrikonn, line 158: Need to yarn install afterwards to update the lock file.

saevarma: Yup saw the failed check, was commiting updated yarn.lock

darrikonn (approved):

arnifreyrs (approved):
lgtm

PR #1498, opened 2020-10-22
chore: add issue repro

jeremybarbet:
## What

See: https://github.com/island-is/island.is/pull/1498/files#diff-593e15c1908ada6a0bafe0a2e7862af1a05c624b40137646f7bb6c3db1899066R2-R6

There is an issue going on between treat and the island-ui/core but can't find what's wrong. It results in:

Screenshot 2020-10-22 at 14 22 48

However, if I have the same function imported from island-ui/theme it just works.


jeremybarbet:
Screenshot 2020-10-22 at 14 22 48

PR #1499, opened 2020-10-22
Judicial-system/Save case fix

oddsson:
# Save case with all fields on the first screen for prosecutors.

https://app.asana.com/0/1197628843274291/1197773320394651

## What

See ☝️

## Why

If the name of the accused is entered first, then police case number and national id, the name would not be saved with the case.

## Checklist:


gudjong (approved):
Nice

PR #1506, opened 2020-10-22
Feature/es rank eval

haukurh:
# Rank evaluation on ElasticSearch

ElasticSearch rank evaluation as requested

## What

Implement rank evaluation on ElasticSearch

## Why

To be able to evaluate changes on ElasticSearch queries, in order see if improve or reduce the search results quality

## Screenshots / Gifs

## Checklist:


libs/elastic-indexing/src/lib/indexing.service.ts
```diff
@@ -43,4 +46,48 @@ export class IndexingService {
...
+  }
+
+  protected async getRankQueries(locale: string) {
+    if (this.rankQueries) {
+      return this.rankQueries
+    }
+
+    const index = await this.getIndex(locale)
+    this.rankQueries = rankQueries.map((entry) => {
+      const query = this.elasticService.getSearchQuery({
```
stefanhar, line 78: You don't have to export `getSearchQuery` if this query generation is moved to the content-search lib.
Also getting search query might be handled in the rankEvaluation query generation by importing the searchQuery there

libs/elastic-indexing/src/lib/indexing.service.ts
```diff
@@ -43,4 +46,48 @@ export class IndexingService {
     logger.info('Done with sync')
     return true
   }
+
+  async rankEvaluation(locale: string) {
+    const requests = await this.getRankQueries(locale)
+    const index = await this.getIndex(locale)
```
stefanhar, line 52: You should be able to get the ES indexes from `SearchIndexes[locale]`

stefanhar (changes requested):

libs/elastic-indexing/src/lib/indexing.service.ts
```diff
@@ -43,4 +46,48 @@ export class IndexingService {
...
+
+    return this.elasticService.precisionEvaluation(index, requests, 10)
```
stefanhar, line 54: We might want to format the results to a more readable format for each query.
We should probably also document this endpoint

haukurh: I do agree to document this endpoint, that's an oversight on my part. I thought of this endpoint as an endpoint where developers can run the rank evaluation when they choose (during changes to the search engine) and also this could be hook some automation if needed/wanted.

But how would you like to see this formatted ?

I personally find json easy to decipher and with some basic browser extension, it's very easy to read
image

libs/api/content-search/src/services/elastic.service.ts
```diff
@@ -148,17 +153,49 @@ export class ElasticService {
...
       size,
       page,
       types,
       tags,
       countTag,
     })
+  }
+
+  async search(index: SearchIndexes, query: SearcherInput) {
+    const requestBody = this.getSearchQuery(query)
```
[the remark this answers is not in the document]
haukurh: huhh, can you explain what you mean? :) 

but searchQuery requires SearchInput as an input, which SearcherInput is not fully compatible with, resulting in a warning, unless you extract the attributes you need, like in the new getSearchQuery function of this class.

therefor I made getSearchQuery accept both, which results in fewer lines

libs/api/content-search/src/services/elastic.service.ts
```diff
@@ -148,17 +153,49 @@ export class ElasticService {
...
+      body: query,
+    })
+  }
+
+  async precisionEvaluation(
+    index: string | string[],
+    requests: Request[],
+    size = 10,
+  ) {
+    const requestBody = precisionEvaluation(requests, size)
```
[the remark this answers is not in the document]
haukurh: Will fix

libs/elastic-indexing/src/lib/indexing.service.ts
stefanhar: The alias points to the current index, is there a need to rank other index versions?

libs/elastic-indexing/src/lib/indexing.service.ts
stefanhar: JSON is fine, I'm suggesting we format the response.
We have some information in this response that are not relevant e.g. meta.connection.resurrectTimeout
We might create a rank type e.g.
```ts
interface SearchRanking {
  score: number
  relevantDocsRetrived: number
  unratedDocsRetrived: number
  description: string <- what this ranking is testing
}
```
And return it in a ranking response e.g.
```ts
interface RankingResponse {
  [requestedString: string]: SearchRanking
}
```

libs/api/content-search/src/services/elastic.service.ts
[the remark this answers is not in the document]
stefanhar: Search should just accept search input, I believe using dto here is not optimal.
This is being rectified in #1411

stefanhar:
This has been merged into #2109

PR #1509, opened 2020-10-22
Website / Remove sticky if sticky content is taller than the window height

stjanilofts:
## What

A wrapper div that is a sticky will now automatically be a sticky if the content height is less than the window height. Can be forced to be always on with "constantSticky" prop.

## Why

Users cannot see the bottom half of the sticky sidebar until they have reached the end of the page. This will at least make the sidebar more usable on long pages with tall sidebars.

## Checklist:


DavidBachmann (approved):
LGTM

PR #1511, opened 2020-10-23
Website/siteimprove

ingolfurs:
# Accessibility fixes based on Siteimprove errors

## What
- Only add img role and aria-label to BackgroundImage element if alt text exists
- Fix _Link text used for multiple different destinations_ issue in FrontpageTabs
- Hide illustration from screen readers in frontpage IntroductionSection
- Fix Element ID is not unique in LanguageToggler, because it is included twice, in Header for desktop and for SideMenu for mobile.
- Support lang attribute on modal element
- Surround english in LanguageToggler text with lang attribute, so screen readers will notice this text is in english
- Add lang attribute to language toggle button, it now reads english text on icelandic page and vice versa
- Fix _WAI-ARIA alert is present when page loads_ issue in Modal, switch to english warning should just be rendered on display
- Fix headings order in news list
- Fix _Missing required WAI-ARIA states or properties_ issue by add missing aria-controls to combobox
- Fix _Redundant WAI-ARIA attribute_ issue by remove aria-label from SideMenu content tabpanels
- Fix _Image with no alt attribute_ issue by add an empty alt attribute to AboutPage image
- Hide mobile tab select element on About page "Hvað breytist?" section, as an way to avoid error that dummyInput element creates because it is empty label: Form control label is missing text. Could not find a way to make <Select element not create this dummy input element that is not used in this case.

## Why
island.is must validate to accessibility standards, at least Level AA. There are too many Level A issues that must be fixed.
I recommend [Siteimprove accessibility checker chrome extension](https://support.siteimprove.com/hc/en-gb/articles/115002413812-Siteimprove-Accessibility-Checker-Chrome-Extension) to check for accessibility issues.

## Checklist:


DavidBachmann (approved):
👍

PR #1530, opened 2020-10-26
Translation fallbacks

magnearun:
Handle translation fallbacks in translation service

## What

Handle translation fallbacks in translation service, before this change the fallback support was handled in the Contentful extension. Since fallbacks are not set for locales in Contentful for various reasons (it was causing bugs on the website), I declared the used locales and fallbacks in code in the translation service, but using the same format as Contentful so this can be replaced if we set the fallbacks in contentful in the future.

## Why

If requested locale is English but the is not translation found, the icelandic string will be returned for that key.

## Checklist:


libs/api/domains/translations/src/lib/translations.service.ts
```diff
@@ -23,19 +30,25 @@ export class TranslationsService {
     namespaces?: string[],
     lang?: string,
   ): Promise<TranslationsDict | null> => {
+    const locale = locales.find((l) => l.code === localeMap[lang])
+
     const result = await this.contentfulRepository
-      .getLocalizedEntries<any>(lang, {
+      .getLocalizedEntries<any>('*', {
```
arnifreyrs, line 36: so this fetches everything?

libs/api/domains/translations/src/lib/translations.service.ts
```diff
@@ -23,19 +30,25 @@ export class TranslationsService {
...
       })
       .catch(errorHandler('getNamespace'))
 
-    const withFallbacks = result?.items?.map(({ fields }) =>
-      mergeWith({}, fields.fallback, fields.strings, (o, s) =>
-        isEmpty(s) ? o : s,
-      ),
-    )
+    const withFallbacks = result?.items?.map(({ fields }) => {
+      console.log('fields', fields)
```
arnifreyrs, line 44: away with you

libs/application/templates/reference-template/src/forms/messages.ts
```diff
@@ -13,7 +13,8 @@ export const m = defineMessages({
   },
   introIntroduction: {
     id: 'example.application:intro.introduction',
-    defaultMessage: 'Halló {name}',
+    defaultMessage:
```
arnifreyrs, line 16: this should not be part of the PR

libs/application/ui-fields/src/lib/IntroductionFormField.tsx
```diff
@@ -21,7 +22,9 @@ const IntroductionFormField: FC<{
           {formatText(field.name, application, formatMessage)}
         </Text>
       )}
-      <Text>{formatText(field.introduction, application, formatMessage)}</Text>
+      <Markdown>
```
arnifreyrs, line 25: nor this

arnifreyrs (changes requested):

libs/api/domains/translations/src/lib/translations.service.ts
magnearun: yes this fetches the field in all locales

libs/application/templates/reference-template/src/forms/messages.ts
magnearun: No, I forgot I branched out of another branch, not master. That should be merged, will rebase

arnifreyrs (approved):

atligudl (approved):
Good upgrade!
I wonder if there's anything we can do about text flickering from the default message to the locale message on the initial fetch?

PR #1538, opened 2020-10-26
Feature/xroad collector

saevarma:
# XRoad-Collector scheduler service

## What

Adding xroad-collector scheduler service

## Why

To test the xroad-collector in dev.

## Checklist:


apps/services/xroad-collector/src/app/restservicecollector.service.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  private async indexProviders(providers: Array<Provider>): Promise<void> {
+    // Remove the index so we can recreate it
+    // with the latest state in X-Road
+    await this.elasticService.deleteIndex()
+
+    for (const provider of providers) {
+      // For each provider get list af all REST services
+      // currently supporting those who were registered using OpenAPI
+      const services = await this.restMetadataService.getServices(provider)
```
pshomov, line 36: Was this the service call that swallows exceptions? If 'yes' - I would think that here would be a better place to handle that.

apps/services/xroad-collector/src/app/restservicecollector.service.ts
```diff
@@ -0,0 +1,42 @@
...
+
+      // Insert into Elastic
+      await this.elasticService.bulk(services)
```
pshomov, line 39: Maybe handle exceptions here in case ES cannot insert some services metadata. One failure should not stop the other services from getting in ES IMO.

apps/services/xroad-collector/src/app/restservicecollector.service.ts
saevarma: Agreed. Will change.

apps/services/xroad-collector/src/app/restservicecollector.service.ts
saevarma: The bulk method was using try/catch and swallowing. I will change it to not do that, and we do it here as in the above comment :)

apps/services/xroad-collector/src/app/restservicecollector.service.ts
saevarma: Pushed new commit with changed exception handling. All swallowing removed, except in RestMetadataService.getOpenApiString. Added comment in the code why and when it should be changed.

apps/services/xroad-collector/src/app/restservicecollector.service.ts
saevarma: Changed exception handling and now it is done here.

Addips (approved):
LGTM

pshomov (approved):
LGTM

PR #1545, opened 2020-10-27
GraphQL API for Judicial System

gudjong:
# GraphQL API for Judicial System

https://app.asana.com/0/1197628843274291/1192548337265729

## What

☝️ 

## Why

Digital Iceland policy

## Checklist:


darrikonn:
Don't have access to the Asana link in the description.

Can you either invite darri@viceversa.is or explain why this api isn't using the shared one?

apps/judicial-system/api/src/app/modules/case/models/pendingSignature.model.ts
```diff
@@ -0,0 +1,12 @@
+import { Field, ObjectType } from '@nestjs/graphql'
+
+import { PendingSignature as TPendingSignature } from '@island.is/judicial-system/types'
+
+@ObjectType()
+export class PendingSignature implements TPendingSignature {
+  @Field()
+  controlCode: string
```
darrikonn, line 8: I'm guessing this is the ID for the type. Then this needs to be marked as:
```typescript
import { Field, ID } from '@nestjs/graphql'
...
@Field((_1) => ID)
```
for better caching in the front-end.

darrikonn: ^ More of an awareness for all your GraphQL types

apps/judicial-system/api/src/app/modules/user/user.model.ts
```diff
@@ -1,79 +1,24 @@
...
-} from 'sequelize-typescript'
+import { Field, ObjectType } from '@nestjs/graphql'
 
-import { ApiProperty } from '@nestjs/swagger'
+import { User as TUser, UserRole } from '@island.is/judicial-system/types'
 
-import { UserRole } from './user.types'
+@ObjectType()
+export class User implements TUser {
+  @Field()
```
darrikonn, line 7: ditto (won't comment more on this, you can check the rest of your types)

apps/judicial-system/backend/src/app/modules/case/models/case.model.ts
```diff
@@ -0,0 +1,309 @@
...
+  })
+  @ApiProperty()
+  requestedCustodyEndDate: Date
+
+  @Column({
+    type: DataType.STRING,
+    allowNull: true,
+  })
+  @ApiProperty()
+  // Lagaákvæði sem brot varða við
```
darrikonn, line 117: Icelandic 😢  😄

apps/judicial-system/web/src/graphql/gql.ts
```diff
@@ -0,0 +1,353 @@
...
+      }
+      # notifications
+    }
+  }
+`
+
+export const UpdateCaseMutation = gql`
+  mutation UpdateCaseMutation($id: ID!, $updateCase: UpdateCaseDto!) {
+    updateCase(id: $id, updateCase: $updateCase) {
+      id
```
darrikonn, line 202: Why are the queries/mutations split up from the components?

This most often leads to objects being requested that aren't needed in the components (looks like these mutations are requesting all the objects).

Each component should only request what they need.

apps/judicial-system/web/src/graphql/gql.ts
```diff
@@ -0,0 +1,353 @@
...
+      controlCode
+      documentToken
+    }
+  }
+`
+
+export const ConfirmSignatureQuery = gql`
+  query ConfirmSignatureQuery($id: ID!, $documentToken: String!) {
+    confirmSignature(id: $id, documentToken: $documentToken) {
+      documentSigned
```
darrikonn, line 348: If an ID is requested as well, then Apollo will know how to update its store.

apps/judicial-system/web/src/routes/Judge/Confirmation/Confirmation.tsx
```diff
@@ -18,29 +13,116 @@ import { TIME_FORMAT, formatDate } from '@island.is/judicial-system/formatters'
...
+          : signatureResponse.documentSigned
+          ? 'Tilkynning hefur verið send á ákæranda og dómara sem kvað upp úrskurð.'
+          : 'Vinsamlegast reynið aftur svo hægt sé að senda úrskurðinn með undirritun.'
+      }
+      secondaryButtonText={
+        !signatureResponse
+          ? null
+          : signatureResponse.documentSigned
+          ? 'Loka glugga'
+          : 'Loka og reyna aftur'
```
darrikonn, line 103: Will this project have to support English as well?

darrikonn: And what about content management?
Wondering if it's only supposed to support Icelandic, should the "translations" be stored in a json file so someone from content management can more easily update the texts.

apps/judicial-system/web/src/routes/Judge/Overview/Overview.tsx
```diff
@@ -34,24 +40,22 @@ export const JudgeOverview: React.FC = () => {
     setCourtCaseNumberErrorMessage,
   ] = useState('')
   const [workingCase, setWorkingCase] = useWorkingCase()
-  const uContext = useContext(userContext)
 
   useEffect(() => {
     document.title = 'Yfirlit kröfu - Réttarvörslugátt'
   }, [])
 
+  const { data } = useQuery(CaseQuery, { variables: { id: id } })
```
darrikonn, line 48: Queries and mutations should follow the `Input objects naming conventions`: https://github.com/island-is/handbook/blob/master/docs/api-design-guide/graphql-naming-conventions.md#input-objects-naming-conventions

darrikonn (changes requested):

apps/judicial-system/api/src/app/modules/case/models/pendingSignature.model.ts
gudjong: Í þessu tilviki er ekki um raunverulegan object að ræða og ekkert id. Þetta er svar sem við fáum frá undirskriftarþjónustu sem notar rafræn skilríki á sími. Þetta svæði (controlCode) er fjögurra tölustafa kóðinn sem birtist á símanum þínum þegar þú undirritar.

apps/judicial-system/api/src/app/modules/user/user.model.ts
gudjong: Við erum ekki að skila user id úr grunninum heldur bara kennitölunni. Finnst þér ástæða til að merkja kennitöluna með ID?

apps/judicial-system/backend/src/app/modules/case/models/case.model.ts
gudjong: Það er ástæða fyrir þessu íslenska commenti. Þýðingar eru ekki alltaf augljósar og til að tengja við texta sem birtast í UI þá settum við comment á íslensku.

apps/judicial-system/api/src/app/modules/user/user.model.ts
darrikonn: Yup, the Apollo store needs to be able to match future objects to the objects in its cache, using the GraphQL ID provided.

apps/judicial-system/web/src/graphql/gql.ts
gudjong: Not sure I understand this comment. We are only returning the updated object.

apps/judicial-system/web/src/graphql/gql.ts
gudjong: I do not understand this comment. Here we are querying a document signature service, asking for the result of a previously requested document signature. The signature is related to a court case that is stored in a database and has an ID, but the the court case itself is not returned from this query. The returned object is transient, not stored and only relevant while showing the user whether the document signing succeeded or not.

apps/judicial-system/web/src/routes/Judge/Confirmation/Confirmation.tsx
gudjong: This project will not be in other languages. It is used by judges and prosecutors in the Icelandic judicial system.

apps/judicial-system/web/src/routes/Judge/Overview/Overview.tsx
gudjong: I was not aware of that. Will modify the code accordingly.

apps/judicial-system/api/src/app/modules/user/user.model.ts
gudjong: Will modify then, probably return the surrogate key instead.

apps/judicial-system/web/src/graphql/gql.ts
darrikonn: e.g. in `apps/judicial-system/web/src/routes/DetentionRequests/DetentionRequests.tsx` you're only working with 
```
      id
      policeCaseNumber
      accusedName
      accusedNationalId
      state
      custodyEndDate
```
.. but that query is requesting:
```typescript
      id
      created
      modified
      state
      policeCaseNumber
      accusedNationalId
      accusedName
      accusedAddress
      court
      arrestDate
      requestedCourtDate
      requestedCustodyEndDate
      lawsBroken
      custodyProvisions
      requestedCustodyRestrictions
      caseFacts
      witnessAccounts
      investigationProgress
      legalArguments
      comments
      # prosecutorId
      prosecutor {
        nationalId
        name
        title
        mobileNumber
        email
        role
      }
      courtCaseNumber
      courtStartTime
      courtEndTime
      courtAttendees
      policeDemands
      accusedPlea
      litigationPresentations
      ruling
      rejecting
      custodyEndDate
      custodyRestrictions
      accusedAppealDecision
      accusedAppealAnnouncement
      prosecutorAppealDecision
      prosecutorAppealAnnouncement
      # judgeId
      judge {
        nationalId
        name
        title
        mobileNumber
        email
        role
      }
      # notifications
    }
```
which is a bloated payload and not what the component needs.
Here, the `GraphQL` is being used like `REST`, removing the power of GraphQL.

Same for the `mutations`. What is it that the individual component is about to change using the mutation? The component should then only request the fields that are about to be changed.

If you specify the `GraphQL ID` (like mentioned in previous comments) then the Apollo store will respectfully update the corresponding objects that are later requested with additional data.

apps/judicial-system/web/src/graphql/gql.ts
darrikonn: My confusion was due to the name of the parameter `id`.
Is this `id` the id of the confirm-signature which is being queried, or the id of another object, e.g. document-signature, which would lead to a more descriptive query:
```graphql
confirmSignature(documentSignatureId: $id, ...) {
  ...
}
```

gudjong: Got you, I will clarify the naming. I should not use ID here I guess as I do not want Apollo to update anything.

apps/judicial-system/web/src/graphql/gql.ts
gudjong: Your comment on the query used in DetentionRequests makes sense. I will limit the fields returned to what is actually used.

Regarding the mutations and GraphQL automatically (respectfully :) ) updating its store, this makes sense as long as we have a single user working on the same court case. That is the situation we expect for now, but later, we will likely have multiple people working on the same court case and need to guard against that. In such cases, we will need to be able to ask Apollo to invalidate its cache or not use the cache at all. Do you have any thoughts/pointers regarding this?

darrikonn: Yeah, best case scenario would be to setup GraphQL Subscription instead of Queries -> over websocket

apps/judicial-system/api/src/app/modules/user/user.model.ts
gudjong: Modified and pushed.

apps/judicial-system/web/src/routes/Judge/Overview/Overview.tsx
gudjong: I think I have addressed all naming issues.

apps/judicial-system/web/src/graphql/gql.ts
gudjong: Right, we fight that battle in due time.

gudjong: I have now addressed most of the issues you mentioned here.

One thing I was struggling with and did not address was updates to court cases. In most cases the client updates a single field at a time or when the corresponding input field looses focus. Ideally, I would wan the mutation to query for the exact same fields that are being updated, plus some select other fields, such as a modified timestamp.

This would something like this:

mutation UpdateCaseMutation($input: UpdateCaseInput!) {
    updateCase(input: $input) {
      id
      modified
      ...$input
    }
}

However, I don't know how to create such a query, except perhaps by using dynamic queries generated at run time.

So, I will submit this PR for another review. I would appreciate any suggestions as to how I can handle these update mutations. Alternatively, it would be great if you would approve the PR and we would improve this later.

darrikonn: I believe the main issue here is that `updateCase` is too broad. It doesn't specify "what" it's updating, going against the GraphQL Mutation principal: `Make mutations as specific as possible.`.
There you end up with the same problem as before, querying all.
Moving the AST to the component would also document quite well what the component is doing, e.g.
```graphql
query JudicialSystemUserQuery {
  judicialSystemUser {
    id
    name
    address
  }
}

mutation JudicialSystemChangeAddressMutation($input: JudicialSystemChangeAddressInput) {
  judicialSystemChangeAddress(input: $input) {
    id
    address
  }
}
```

You can even batch up the mutations, to save round-trips (mutations run in series):
```graphql
mutation JudicialSystemUpdateUserMutation($input: JudicialSystemUpdateUserInput) {
  judicialSystemChangeAddress(..) {
    id
    address
  }
  judicialSystemChangeName(..) {
    id
    name
  }
}
```

gudjong: I understand the issue, but with this approach I will end up with a lot of mutations. One for each single field update and one for any combination of multi field updates possible.

The general rule in the UI is to save each field separately, but there are a few cases where more than one field are saved at the same time.

Later on, we will also need to deal with the case where some of the mutations will fail, because of network issues and such. Then, we will try to save all edited, but not yet saved fields. Creating a hard coded mutation for each case is not feasible. 

So my question is rather: How do people handle such cases in general? Do they generate dynamic mutations at run time?

gudjong: I'll do some experiments, but what is your opinion on dynamically creating mutations at run-time?

darrikonn: Yeah I see :/.. I've had a similar problem which I didn't look any deeper into.
The dynamic field fetching sounds like a good solution!
Having a top-level function that returns GraphQL AST :+1:

gudjong: 👍

gudjong: Turns out we don't want Apollo to cache the queries. This is because more than one user can be modifying/creating court cases at the same time and we want the latest data/updates to be loaded when queries are executed. This means that we really don't need the update mutation to return any data, so I modified the mutation to only return a "modified" timestamp.

gudjong: Would be great to get another review.

darrikonn (approved):

oddsson (approved):

PR #1547, opened 2020-10-27
Website/ Performance sprint

DavidBachmann:
# Performance sprint

## Main issues:
[x] Hero slider is way too heavy
[x] Hero causes page jumping
[ ] Initial JS payload is over 600kb

### Hero animations are being temporarily removed to establish a baseline performance and to improve the homepage ASAP. Animations will be put back in using a different strategy soon.

Animation strategy going forward:
- Keep animation files local and pick them by ID from Contentful if they have to be picked.
- Serve a basic SVG version and enhance it with a dynamically loaded lottie-web canvas implementation.


apps/web/components/FrontpageTabs/illustrations/ManOnBenchIllustration.tsx
```diff
@@ -0,0 +1,73 @@
+import React from 'react'
+
+function ManOnBenchIllustration() {
```
arnifreyrs, line 3: other projects need these illustrations. I suggest that these illustrations are moved to a shared library.

stefanhar:
Did we view alternative methods that allow us to keep the animations e.g. loading them lazily or load them as assets instead of string?

apps/web/components/FrontpageTabs/illustrations/ManOnBenchIllustration.tsx
DavidBachmann: Good call

DavidBachmann:
> Did we view alternative methods that allow us to keep the animations e.g. loading them lazily or load them as assets instead of string?

This Slider change will increase performance ASAP because it's unacceptable right now — it will **also** give us a baseline to try to incorporate the animations performantly while taking the time we need to do so.

stefanhar (approved):

darrikonn (dismissed):

DavidBachmann:
Closed in favor of #1576

PR #1551, opened 2020-10-27
Always run retag unaffected, even though there are no builds

sindrig:


pshomov (approved):
LGTM

sindrig:
example run: https://github.com/island-is/island.is/runs/1313980884?check_suite_focus=true

PR #1559, opened 2020-10-27
Avoid red builds on branch creation

sindrig:
Avoid red builds on branch creation
    
This started happening after #1359 was merged


pshomov (approved):
LGTM

PR #1573, opened 2020-10-27
Judicial system/Stop using localstorage to store form data.

oddsson:
# Rethink how we use form data

Fixes: https://app.asana.com/0/1197628843274291/1198491524080605

## What

This PR does the following

- Removes all references to local storage.
- Add's a loading indicator while fetching a case from API.
- Add's error handling when a case is not found. 
- Add's a `reactLink` to AlertBanner component because that component assumes that the consumer is using NextJS.

## Why

Keeping sensitive data in local or session storage is generally a bad idea. Instead of managing the data via browser storage, we're now always reading the data from API.

The other things are for UX.

### Loading indicator

ezgif com-gif-maker (1)

### Case not found 

Screenshot 2020-10-27 at 20 32 24

## Checklist:


oddsson:
Closed in favour of https://github.com/island-is/island.is/pull/1643

PR #1592, opened 2020-10-28
updating details page to use query to prevent contacting wrong endpoint

Addips:
## What

Updating service detail to fetch using useQuery instead of using serversideprops

## Why

To prevent the graphql call to use the wrong endpoint during build and shorting possible page build time

## Checklist:


saevarma (approved):
LGTM

PR #1604, opened 2020-10-29
Remove all semver prefixes from package.json

arnifreyrs:
# No semver prefixes in package.json nor when new packages are added

Make sure all newly added packages have no prefix.

## Why

This is to make sure that all builds of a certain branch will behave the same, and not be sensitive to accidental buggy version releases of our various dependencies. 

This poses a question how we should handle updating the dependencies of this repo. One way could be to do that in refactor-weeks between sprints, or have an automatic dependency update PRs generated by a bot that could be run every week. This is something that needs to be discussed.

## Checklist:


dabbeg (commented):
I totally agree with this but I wonder if pinning the devDependencies is too far.
But that ties to another problem, that we are installing all dependencies in our docker containers (including devDependencies).

pshomov (approved):
LGTM

dabbeg:
> I totally agree with this but I wonder if pinning the devDependencies is too far.
> But that ties to another problem, that we are installing all dependencies in our docker containers (including devDependencies).

But yeah that's another problem for another time, this is the right move forward.

I am wondering though instead of just removing the hat if we should also make the version match the one that is used in the yarn.lock file so we are not changing the version?

arnifreyrs:
> > I totally agree with this but I wonder if pinning the devDependencies is too far.
> > But that ties to another problem, that we are installing all dependencies in our docker containers (including devDependencies).
> 
> But yeah that's another problem for another time, this is the right move forward.
> 
> I am wondering though instead of just removing the hat if we should also make the version match the one that is used in the yarn.lock file so we are not changing the version?

That is a great idea, I'll try to make that work somehow, could be loads of manual labour :D but I like it

arnifreyrs:
This is continued in  https://github.com/island-is/island.is/pull/1609

PR #1617, opened 2020-10-29
Identity server url fetched from env variable

Thorrik58:
# Use env for identity server url

## What

## Why

## Screenshots / Gifs

## Checklist:


apps/service-portal/src/environments/environment.prod.ts
```diff
@@ -1,3 +1,6 @@
-export const environment = {
+export default {
   production: true,
+  identityServer: {
+    IDENTITY_SERVER_BASE_URL:: process.env.IDENTITY_SERVER_BASE_URL,
```
obmagnusson, line 4: double ::

obmagnusson (changes requested):

apps/service-portal/src/utils/userManager.ts
```diff
@@ -4,9 +4,10 @@ import {
   InMemoryWebStorage,
   UserManagerSettings,
 } from 'oidc-client'
+import { environment } from '../environments'
 
 const settings: UserManagerSettings = {
-  authority: 'https://siidentityserverweb20200805020732.azurewebsites.net/',
+  authority: environment.identityServer.IDENTITY_SERVER_BASE_URL,
```
pshomov, line 10: I think this should be just fine, do you have the environment config replacement thing in workspace.json?
Also not sure if it matters but you had originally a trailing forward slash and now it is missing.

pshomov (commented):
LGTM

apps/service-portal/src/utils/userManager.ts
Thorrik58: @pshomov 
Yes we have the file replacement line in our workspace.json.

Ill double check the trailing slash. 

Trying to load this env variable locally from the environment file by pointing towards process.env.IDENTITY_SERVER_BASE_URL in environments.ts doesnt work. But maybe its just not working locally for once :P

pshomov: if you want to try it locally then you should rename your .prod file to be your default one and then try it. Alternatively you can `yarn build` your project with `--prod` flag to make it build a prod distro and run it directly.

obmagnusson (approved):
LGTM give it a try

PR #1687, opened 2020-11-03
Skilavottord - Minor fix in method that filters car

dd-sdang:
# Skilavotttord - Minor fix in method that filters car

## What

Added a missing return in the switch/case for case 'deregistered' cars

## Why

Because it didnt return any value otherwise and fell through to the default case.

## Checklist:


tommigitgaur (approved):
LFTM

PR #1705, opened 2020-11-04
Service portal :  Rewrite document client

obmagnusson:
# \<Description\>
Rewrite document REST client to user HttpModule instead of openapi generated code from an .yml file.
Rehydration of token implemented.

## What

REST client credentials communication with Island.is Postholf.

## Why

Previous setup did not allow for rehydration of tokens. And this should be a much more efficient.

## Screenshots / Gifs

## Checklist:


libs/api/domains/documents/src/lib/client/document.connection.ts
```diff
@@ -0,0 +1,50 @@
...
+      logger.info(
+        'No credentials provided to fetchToken for DocumentOauthConnection',
+      )
+      return {
+        token: '',
+        expiresIn: 0,
+      }
+    }
+
+    logger.debug('Fetching token for Document Service')
```
atligudl, line 26: Remove debug logs?

libs/api/domains/documents/src/lib/client/documentClient.ts
```diff
@@ -0,0 +1,80 @@
...
+  private async rehydrateToken() {
+    if (
+      !this.accessTokenExpiryDate ||
+      this.accessTokenExpiryDate < new Date()
+    ) {
+      const { token, expiresIn } = await DocumentOauthConnection.fetchToken(
+        this.clientConfig,
+      )
+      this.accessToken = token
+      let expiryTime = new Date()
```
atligudl, line 40: const?

libs/api/domains/documents/src/lib/client/documentClient.ts
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  private async getRequest<T>(requestRoute: string): Promise<T> {
+    await this.rehydrateToken()
+    const config: AxiosRequestConfig = {
+      headers: {
+        Authorization: `Bearer ${this.accessToken}`,
+      },
+    }
+    console.log(`Get : ${this.clientConfig.basePath}${requestRoute}`)
```
atligudl, line 53: console log

atligudl (commented):
Few minor comments

libs/api/domains/documents/src/lib/client/document.connection.ts
obmagnusson: no as i understand these only fire on debugging.

eirikurn:
Fyi, looks to me like we can implement token rehydration in a client middleware (as part of the codegen api configuration).

The "pre" hook receives the "url" and "init" fetch arguments, and can change do async logic to change any part of the request. Your call.

obmagnusson (commented):
I tried that at one point. Ran into some problems at with the client not properly picking up the middleware. Guessing the yml structure was improperly set up. Maybe something we will revisit when the Postholf service gets reworked and integrated into X-Road.

andriandresson (approved):
LGTM 🎉

atligudl (approved):

arnifreyrs (approved):
From my codeowner standpoint this looks good.

PR #1708, opened 2020-11-04
Parental leave review screen

madebynoam:
* First pass at the Review screen for the Parental Leave application
* Added a `type="button"` attribute to core `AccordionItem` so that it doesn't submit forms. (@jeremybarbet if you could confirm for Island UI)

## Screenshots / Gifs

image

## Todo:
* Change the submit button to green (@arnifreyrs do you know how we can do that?)...
* Implement the timeline chart buttons into the flow (where "delete period" and "add another period" takes you back into the flow).

## Checklist:


libs/application/templates/parental-leave/src/fields/Review/index.tsx
```diff
@@ -0,0 +1,576 @@
...
+                      {
+                        label: '6 personal months',
+                        bulletStyle: 'blue',
+                      },
+                      {
+                        label: '1 granted month from other parent',
+                        bulletStyle: 'greenWithLines',
+                      },
+                    ]}
+                  /> */}
```
arnifreyrs, line 488: remove?

libs/application/templates/parental-leave/src/fields/Review/index.tsx
```diff
@@ -0,0 +1,576 @@
...
+  Accordion,
+  AccordionItem,
+  Box,
+  GridColumn,
+  GridRow,
+  Input,
+  Text,
+} from '@island.is/island-ui/core'
+import {
+  FieldDescription,
```
arnifreyrs, line 18: I feel like this file is too big. Maybe we can reuse better some of the components? or split this into more isolated components, one for each accordion item. This is a tad too much

arnifreyrs:
Regarding the change of colour for the submit button, I will do that in another PR. Also the repeater actions

libs/application/templates/parental-leave/src/fields/Review/index.tsx
madebynoam: oops, should be gone now.

libs/application/templates/parental-leave/src/fields/parentalLeaveUtils.ts
```diff
@@ -22,3 +26,37 @@ export function getEstimatedMonthlyPay(application: Application): number {
...
+  })
+  return timelinePeriods
+}
+
+/*
+ *  Takes in a number (ex: 119000) and
+ *  returns a formated ISK value "119.000 kr."
+ */
+export const formatIsk = (value: number): string =>
+  value.toString().replace(/\B(?=(\d{3})+(?!\d))/g, '.') + ' kr.'
```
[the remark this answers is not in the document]
madebynoam: Yeah this one came in from the pitch project we did, I'll work on it (+ tests) in the next PR :)

libs/application/templates/parental-leave/src/fields/Review/index.tsx
madebynoam: Just pushed two refactors:
`PaymentsTable` and `YourRightsBoxChart` which cut out a bunch of repeated code from the Review screen. 
I think we can probably also wrap the periods component...
What do you think?

arnifreyrs: I like this! I think we can do that in a future PR, this is great for now

libs/application/templates/parental-leave/src/fields/Review/index.tsx
```diff
@@ -0,0 +1,403 @@
...
+              </GridRow>
+            </Box>
+          </AccordionItem>
+        </Accordion>
+      </Box>
+    </div>
+  )
+}
+
+export default Review
```
arnifreyrs, line 403: I like the changes here, the only thing I would add later is error handling. I think we need that at some point. 

More importantly, I think the application system should know how to dynamically render a summary screen so this repetitive coding is unnecessary, but that can wait until later :)

arnifreyrs (dismissed):

libs/application/templates/parental-leave/src/fields/Review/index.tsx
madebynoam: Agree, both error handling and validation. 
And yeah I think it would be nice to make it part of the general Application System.

jeremybarbet (approved):
LGTM

arnifreyrs (approved):

jeremybarbet (approved):

PR #1728, opened 2020-11-05
Setting maxWorkers for builds to lower memory consumption

dabbeg:
## What

Lowering memory consumption for `nx serve`

## Why

Projects that use either `node:build` or `web:build` were taking around 5GB per project.
When running multiple projects this overloads most machines (even though they have 32GB ram).
By default the number of workers equal the number of CPU cores, which is 12 for many new machines.

Fun side affect: the startup time is also reduced :+1: 

## Checklist:


darrikonn (approved):
Noice debugging leading to this :+1:

pshomov (approved):
Awesome!

arnifreyrs (approved):
Wow, amazing!

PR #1731, opened 2020-11-05
Island-ui / Datepicker: input and design changes

barabrian:
# Datepicker: input and design changes

#### Most noticeable changes
* Change datepicker to use island-ui Input instead of custom button
* Change month header to select instead of text
* Update design to match Figma

## Why

Fulfilling design requirements 

## Side-effects

* Text component
  * Warp with forward ref
* Input component
  * add icon support - icons are blue by default and red with error state
  * fix mobile padding right
  * add support for fixed focus state
  * added click and keydown events

## Screenshots / Gifs

image

---

is
datepicker-is

---

en
datepicker-en

---

pl
datepicker-pl

---

## Checklist:


libs/island-ui/core/src/lib/DatePicker/DatePicker.tsx
```diff
@@ -140,50 +124,118 @@ export const DatePicker: React.FC<DatePickerProps> = ({
...
+    {...props}
+    icon="calendar"
+    iconType="outline"
+    fixedFocusState={fixedFocusState}
+    placeholder={placeholderText}
+  />
+)
+
+const getLocale = (locale: Locale) => {
+  return locale === 'is' ? is : locale === 'pl' ? pl : en
```
siggakogk, line 173: maybe do something where it would be easier to add a new language? b/c you never know!

libs/island-ui/core/src/lib/DatePicker/DatePicker.tsx
```diff
@@ -140,50 +124,118 @@ export const DatePicker: React.FC<DatePickerProps> = ({
             setDatePickerState('closed')
             handleCloseCalendar && handleCloseCalendar(startDate)
           }}
-          onChange={(date: Date) => {
+          onChange={(date: any) => {
```
siggakogk: 👾  why?

libs/island-ui/core/src/lib/DatePicker/DatePicker.tsx
```diff
@@ -17,32 +15,62 @@ import { Text } from '../Text/Text'
...
   locale?: Locale
-  value?: ReactDatePickerProps['value']
   minDate?: ReactDatePickerProps['minDate']
   selected?: ReactDatePickerProps['selected']
   disabled?: boolean
   hasError?: boolean
   errorMessage?: string
   id?: string
-  handleChange?: (date: Date) => void
+  handleChange?: (startDate: Date) => void
```
siggakogk, line 33: nice!

libs/island-ui/core/src/lib/DatePicker/DatePicker.tsx
barabrian: lets just say this is still a draft 😄  good spot by the way!

libs/island-ui/core/src/lib/DatePicker/DatePicker.tsx
barabrian: I'll make this more reuseable

libs/island-ui/core/src/lib/DatePicker/DatePicker.stories.tsx
```diff
@@ -9,17 +9,20 @@ import { DatePicker } from './DatePicker'
 export default {
   title: 'Form/DatePicker',
   component: DatePicker,
-  decorators: [withDesign],
-  parameters: withFigma({
-    desktop:
-      'https://www.figma.com/file/pDczqgdlWxgn3YugWZfe1v/UI-Library-%E2%80%93-%F0%9F%96%A5%EF%B8%8F-Desktop?node-id=50%3A155',
```
jeremybarbet, line 15: Any reason the figma link has been removed?

barabrian: No real reason, just forgot to put it back in. There is something weird going on with the generated porps table, it won't list all the props from the interface only the currently used ones. I was thinking we need to do some debugging around this, at least it doesn't look like figma is causing the problem

jeremybarbet: You need to use global import for React `import * as React` [here](https://github.com/island-is/island.is/pull/1731/files#diff-82d164b904bafdf368831dfb85e330a8cfae0df04ea86866271f1021d94081fbL1).

It's a very sad [issue](https://github.com/island-is/island.is/tree/master/libs/island-ui/storybook#props-issue) regarding the docgen plugin that pickup the props

barabrian: You're right! did not know that, thanks! 😄

libs/island-ui/core/src/lib/DatePicker/DatePicker.tsx
```diff
@@ -17,32 +16,64 @@ import { Text } from '../Text/Text'
 
 import * as styles from './DatePicker.treat'
 import * as coreStyles from './react-datepicker.treat'
+import { Input, InputProps } from '../Input/Input'
+import { VisuallyHidden } from 'reakit'
+
+const languageConfig = {
```
jeremybarbet, line 22: Only found one occurence in the project [here](https://github.com/island-is/island.is/blob/ee476d0da2000cb4f854b577a8ca1fea4c6c553c/libs/service-portal/core/src/components/ActionCard/ActionCard.tsx#L45) but I'm wondering if this config should be exported and be reused when possible. It could be moved into the island-ui/utils. What do you think?

libs/island-ui/core/src/lib/Text/Text.tsx
```diff
@@ -45,58 +45,67 @@ export interface TextProps {
...
-        (child: any) => {
-          if (child?.props?.href && child?.props?.as) {
-            // Checking to see if the child  using "href" and "as" props, which indicates it
-            // is (most likely) a next.js link since the linkRenderer breaks this functionality.
-            // TODO: Make linkRenderer handle next.js links.
-            return child
-          } else if (child?.props?.href && typeof linkRenderer === 'function') {
-            return linkRenderer(child.props.href, child.props.children)
-          }
+export const Text = forwardRef<any, TextProps>(
```
jeremybarbet, line 48: Doesn't work if you replace any with `typeof Box`?

barabrian: `typeof Box` is not working, I probably should be using `HTMLElement` I'll change it

jeremybarbet (approved):
LGTM

atligudl (approved):
LGTM

PR #1733, opened 2020-11-05
Fix duration slider for parental leave application.

arnifreyrs:
Add Employer Approval screen.

## Checklist:


libs/application/templates/parental-leave/src/fields/FirstPeriodStart/index.tsx
```diff
@@ -0,0 +1,87 @@
...
+            {
+              label: 'I will start from the date of birth',
+              value: 'dateOfBirth',
+              tooltip:
+                'If the child is born on another date than the expected date of birth, the parental leave and its duration will adjust to the real date of birth',
+            },
+            {
+              label: 'I will start on a specific date',
+              tooltip:
+                'If the child is born on another date than the expected date of birth, the parental leave and its duration will !!!!NOT!!!! adjust to the real date of birth',
```
madebynoam, line 57: Minor can we add `<em>` instead of `!!!!NOT!!!!` :)

madebynoam (approved):
Nice, lgtm!

libs/application/templates/parental-leave/src/fields/FirstPeriodStart/index.tsx
arnifreyrs: haha hadnt noticed it. I think we need to just start using contentful more instead of these oh so many hardcoded strings :/

PR #1742, opened 2020-11-05
Service Portal Remove Mirage Server

atligudl:
# \<Service Portal Remove Mirage Server\>

## What

Remove Mirage Server from project

**Passenger**
Add label flag to all non interactable tags in service portal

## Why

Mirage is not being used anymore

**Passenger**
Tags have a hover effect on them by default which caused confusion with users when they were not interactable

## Checklist:


obmagnusson (approved):
LGTM

PR #1762, opened 2020-11-06
Skilavottord/company list

stjanilofts:
## What

Show custom "connected" components as slices for articles.

## Why

Skilavottord has a component that shows a list of addresses/phone numbers they want/need to show on the "Recycle your car" article on island.is. These components can use optional JSON data from contentful.

## Checklist:


libs/api/domains/cms/src/lib/models/connectedComponent.model.ts
```diff
@@ -0,0 +1,69 @@
...
+  id: string
+
+  @Field()
+  title: string
+
+  @Field({ nullable: true })
+  type?: string
+
+  @Field(() => graphqlTypeJson, { nullable: true })
+  json?: Record<string, any>
```
stefanhar, line 23: I believe we intended this json to be temporary while we test out the display of the component, maybe we should indicate that with a comment?

stefanhar (changes requested):

arnifreyrs:
Why is this new library added to island-ui? Does it belong there? Should it perhaps rather live under `libs/shared/connected-components/` or something?

libs/api/domains/cms/src/lib/models/connectedComponent.model.ts
```diff
@@ -0,0 +1,69 @@
...
+const parseJson = (fields: ISliceConnectedComponentFields) => {
+  const json = fields?.json ?? null
+
+  if (!json) return null
+
+  switch (fields.type) {
+    case 'Skilavottord/CompanyListConnected':
+      if (typeof json === 'object' && Object.keys(json).length) {
+        return {
+          graphqlLink: json.graphqlLink,
```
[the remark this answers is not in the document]
stjanilofts: This is optional to quickly overwrite the api endpoint in the code and does not need to be provided through contentful.

libs/api/domains/cms/src/lib/models/connectedComponent.model.ts
stjanilofts: Ok, I have added a comment mentioning that this is a temporary field for trying out the first component.

libs/shared/connected/src/lib/skilavottord/CompanyList/CompanyListConnected.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+  const { data, error, loading } = useQuery(RECYCLING_PARTNERS, {
+    client: new ApolloClient({
+      name: 'shared-connected-skilavottord-client',
+      version: '0.1',
+      link: new HttpLink({
+        uri: graphqlLink,
+      }),
+      cache: new InMemoryCache(),
+    }),
```
arnifreyrs, line 39: why do you need to introduce a new client to this lib? Apps that use this `skilavottorð` lib will then most likely have multiple clients which can cause all sorts of problems

stjanilofts: This is kind of an experiment with this component. The idea was that it could be pulled into any project (for example apps/web) and work by itself with or without any apollo provider. Ideally the useQuery would not have its own apollo client and use the apollo provider client in the project that imports the component but for now, skilavottord has its own web service.

jeremybarbet: What about creating a new client [here](https://github.com/island-is/island.is/blob/master/apps/skilavottord/web/graphql/client.ts) that will pull the data from the graphql endpoint you want, and then pass the client as a props to the component, instead of passing an [url](https://github.com/island-is/island.is/pull/1762/files#diff-d007f0a626b12a2300321a3f331ead1d47407e90955022960b0b2c1cd3458a8eR29).

Having the client on the component itself will by-pass the SSR and do the query on the client all the time, which could be a pitfall if we use more and more this kind of component in the projects.

stefanhar: This CMP is an attempt at creating a reusable standalone component that fetches it's own data.
One complexity here is that skilavottorð is not in the central API yet, this forces us to create a client to route the useQuery request to the correct API.

Optimally (and hopefully in future) the data for this CMP would be in the central API that way this CMP would use the client form the embedding app (through ApolloProvider) and optimistically use the useQuery hook.

We see SSR as an issue better tackled at a later time since web (the embedding app) talks to a different API.
We think SSR might be solved by having the CMP export a query variable that the embedding app can use to fetch data from the API and the pass the data into the CMP in getInitialProps

stjanilofts: I think it might be overly complicated (or maybe not) to make the article (on island.is) check if this component is being used as one of the content slices and then make the second api call. First await for the article content, then check the content slices for this component, then do the fetch (to the skilavottord web service). SSR for the data that this type of component was not really a priority (in my opinion).

Maybe we should just scrap the client idea and just go with the json from contentful for now (until skilavottord-ws gets moved to the same central api that island.is uses, which I assume will happen someday?)

dd-sdang: How should we proceed? Should you allow this to go in so we could test it or should we remove the apollo client completely and use json data from Contentful until the endpoint is in the central api, as Kristján suggested?

PR #1796, opened 2020-11-09
Service portal/documents found text

andriandresson:
# \<Updated Documents found text\>

https://app.asana.com/0/1199140985009301/1199129883562050
## What

Added singular and plural text.

### Singular: "1 Skjal fannst"
image

### Plural: "X Skjöl fundust"
image

### Edge Case
Documents found is greater than 11, ends with "1" and lang is "IS"
image

## Checklist:


libs/service-portal/documents/src/screens/Overview/Overview.tsx
```diff
@@ -155,6 +155,33 @@ export const ServicePortalDocuments: ServicePortalModuleComponent = ({
 
   const hasActiveFilters = () => !isEqual(filterValue, defaultFilterValues)
 
+  const documentsFoundText = () => {
+    // default text format, singular & plural
+    let foundText =
+      filteredDocuments.length === 1
+        ? formatMessage({
```
atligudl, line 162: formatMessage should only be used when rendering, you can use defineMessage when declaring variables and then format it in the markup

libs/service-portal/documents/src/screens/Overview/Overview.tsx
```diff
@@ -155,6 +155,33 @@ export const ServicePortalDocuments: ServicePortalModuleComponent = ({
...
+            id: 'sp.documents:found.singular',
```
atligudl, line 163: a dot in a variable name will break when extracting messages

libs/service-portal/documents/src/screens/Overview/Overview.tsx
```diff
@@ -155,6 +155,33 @@ export const ServicePortalDocuments: ServicePortalModuleComponent = ({
...
+            defaultMessage: 'Skjal fannst',
```
atligudl, line 164: skjal should be lowercase right?

"21 skjal fannst"

atligudl (changes requested):

atligudl (approved):

PR #1799, opened 2020-11-09
Judicial system/Strict TS setting pt.1

oddsson:
# Start using strict TS.

https://app.asana.com/0/1197628843274291/1199143603683037

## What

See ☝️

## Why

It's a requirement from island-is. Also it's generally good practice.

## Checklist:


gudjong (approved):

apps/judicial-system/web/src/routes/DetentionRequests/DetentionRequests.tsx
```diff
@@ -42,7 +42,7 @@ export const CasesQuery = gql`
 `
 
 export const DetentionRequests: React.FC = () => {
-  const [cases, setCases] = useState<Case[]>(null)
+  const [cases, setCases] = useState<Case[] | null>(null)
```
addi, line 45: Should we not use
`useState<Case[]?>(undefined)`?

libs/judicial-system/formatters/src/lib/formatters.ts
```diff
@@ -12,24 +12,33 @@ const getAsDate = (date: Date | string | undefined | null): Date => {
...
+export function formatDate(
+  date: null,
+  formatPattern: string,
+): string | undefined
 
 export function formatDate(
   date: Date | string | undefined | null,
   formatPattern: string,
-): string | null {
+): string | undefined {
```
addi, line 35: string?

addi (changes requested):

apps/judicial-system/web/src/routes/DetentionRequests/DetentionRequests.tsx
oddsson: That's not a valid syntax 👇

Screenshot 2020-11-10 at 13 37 22

libs/judicial-system/formatters/src/lib/formatters.ts
oddsson: All dates are treated as strings because that's how they are sent to the server.

addi (approved):

PR #1800, opened 2020-11-09
Application system should use new identity-server

dabbeg:
## What

Add identity-server to env and switch to new stable identity-server

## Why

To get the application system working on dev

## Checklist:


darrikonn (approved):

arnifreyrs (approved):
@magnearun had already started work on this in a bigger PR, adding the new authguards to the backend, but this is still a welcome addition

apps/application-system/form/src/environments/environment.prod.ts
```diff
@@ -1,4 +1,7 @@
 export default {
   production: true,
   baseApiUrl: process.env.API_URL,
+  identityServer: {
```
darrikonn, line 4: Should the front-end handle the redirection?

dabbeg: Hmm I am not sure, I don't see any reason why it should not do so.
If you can think of a reason we should take this up in the `audkenning_og_heimild` slack channel as there are more projects configured this way.

dabbeg:
> @magnearun had already started work on this in a bigger PR, adding the new authguards to the backend, but this is still a welcome addition

Awesome! Good to know, I only wanted to make this work on dev. I will not make any more changes to this auth stuff :slightly_smiling_face:

PR #1830, opened 2020-11-11
As/meta application fixes

magnearun:
Fixes to the Meta application

- Fetch organizations and ministries from Contentful and display in a Select field.
- Display repeater data in table

This also moves the Table component that was used in the ParentalLeave application into the shared lib, this component might belong in the ui library but will be here until that is decided. 


libs/application/templates/meta-application/src/fields/DataTable/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+  const emptyMessage = formatText(
+    isDraft ? m.dataEmptyDraft : m.dataEmpty,
+    application,
+    formatMessage,
+  )
+
+  const data = useMemo(
+    () => (application as { answers?: { data?: Data[] } }).answers?.data ?? [],
+    [application.answers?.data],
```
arnifreyrs, line 41: answers can never by undefined

libs/application/templates/meta-application/src/fields/DataTable/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+  const isDraft = application.state === 'draft'
+
+  const emptyMessage = formatText(
+    isDraft ? m.dataEmptyDraft : m.dataEmpty,
+    application,
+    formatMessage,
+  )
+
+  const data = useMemo(
+    () => (application as { answers?: { data?: Data[] } }).answers?.data ?? [],
```
arnifreyrs, line 40: I do not understand this useMemo, what are you trying to achieve?

could you do: `const data = application.answers.data ?? []` instead ?

arnifreyrs (approved):
lgtm!

libs/application/templates/meta-application/src/fields/DataTable/index.tsx
magnearun: React-table requires the data to be memoized https://react-table.tanstack.com/docs/api/useTable

arnifreyrs: ok makes sense

PR #1879, opened 2020-11-13
Add printHidden to Box

stjanilofts:
# What/why

`<Hidden>` is a nice tool when you easily and visibly (in code) want to hide things from print such as in this case we want to not show the sidebar. But `Sticky` does not work correctly when wrapped with `<Hidden>`. It needs the parent element to be full height and when wrapped with Hidden it no longer works. I added boxProps to Hidden so that it can be given `height: full`.

## Checklist:


barabrian (commented):
I don't see the value in having a Hidden component if it's just a disguised box component, I would rather want to see Box component implement something line `printHidden`

stjanilofts:
> I don't see the value in having a `Hidden` component if it's just a disguised `Box` component, I would rather want to see `Box` component implement something like `printHidden`

Yeah I was thinking something something similar. This is more of a quick fix solution to fix the sticky and also keep the easy to spot Hidden wrapper. I'll look into adding `printHidden` on `Box` instead.

barabrian (approved):

PR #1896, opened 2020-11-13
Support/api catalogue strict ts

Addips:
## What

Updating libs for api-catalogue to set strict as true.
The api-catalogue app will be set to strict once techincal decision for 

## Why

To help ensure typesafety

## Checklist:


saevarma (changes requested):
Move default search limit to elastic library.
Is it possible to use specific response type in elastic.service.ts instead of any ?

Addips:
Closing due git history issue

PR #1904, opened 2020-11-16
Fix release branch good build detection

sindrig:
This closes island-is/infrastructure#287
Also package lock detection script and run push on release branch
creation like it was supposed to all the time.

### DIFF
--- a/.github/actions/main.ts
+++ b/.github/actions/main.ts
@@ -38,7 +38,7 @@ rl.on('line', function (line) {
 rl.on('close', async function () {
   const result = await findLastGoodBuild(
     shas,
-    process.env.BRANCH,
+    process.env.BRANCH || process.env.GIT_BRANCH,
     process.env.BASE_BRANCH || 'main',
     new GitHubWorkflowQueries(),
   )
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -45,13 +45,14 @@ jobs:
                 echo "Checking branch against pattern '$pattern'"
                 if [[ "$GIT_BRANCH" =~ $pattern ]]; then
                   echo "'$GIT_BRANCH' matches '$pattern', continuing with build"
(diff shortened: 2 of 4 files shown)


.github/actions/main.ts
```diff
@@ -38,7 +38,7 @@ rl.on('line', function (line) {
 rl.on('close', async function () {
   const result = await findLastGoodBuild(
     shas,
-    process.env.BRANCH,
+    process.env.BRANCH || process.env.GIT_BRANCH,
```
sindrig, line 41: This is basically a case of "Computer, do what I want you to do, not what I tell you to do."

.github/workflows/push.yml
```diff
@@ -45,13 +45,14 @@ jobs:
...
                   echo "'$GIT_BRANCH' matches '$pattern', continuing with build"
+                  echo "::set-output name=PRE_CHECK_OK::$GITHUB_EVENT_NAME"
                   exit 0
                 fi
             done
             echo "No pattern matches '$GIT_BRANCH', exiting with exit code 1"
             exit 1
           fi
-          echo "::set-output name=PRE_CHECK_OK::true"
+          echo "::set-output name=PRE_CHECK_OK::$GITHUB_EVENT_NAME"
```
pshomov, line 55: Not quite sure what's the improvement here? Thought that "true" was clearer if anything. Could not find clear rules about typecasting from string => bool but I guess JS rules apply.

sindrig: Not really an improvement, just some leftover from when I intended to use exit code for flow control.

I felt it was better to have some distinction between output values depending on if it was set by a matching branch or not-create. 

I'm pretty sure we're dealing with js truthy rules, so empty string/null vs any string contents.

pshomov: 👍🏻

pshomov (approved):
LGTM

PR #1907, opened 2020-11-16
Judicial System / Added the formal request to overview screens.

addi:
Added the formal request to overview screens.

https://app.asana.com/0/1199153462262248/1198962953338955

## What

Adding more formality to the process.

## Why

## Screenshots / Gifs

Screenshot 2020-11-16 at 13 37 50

## Checklist:


gudjong:
Tvær athugasemdir varðandi orðalag í dómkröfunni:

- Nafn kærða ætti að vera í þágufalli í þessari setningu. Þetta er líka vandamál í PDF-inu og spurning hvort við getum ekki fengið þessari setningu breytt þanning að nefnifall passi.

- ".. á meðan gæsluvarðhaldinu stendur." hljómar ekki vel. Í PDF-inu notaði ég "... meðan á gæsluvarðhaldinu stendur." sem er réttara.

apps/judicial-system/web/src/routes/Judge/Overview/Overview.tsx
```diff
@@ -49,30 +50,41 @@ export const JudgeOverview: React.FC = () => {
...
 
   useEffect(() => {
     document.title = 'Yfirlit kröfu - Réttarvörslugátt'
   }, [])
 
   useEffect(() => {
-    if (!workingCase && resCase) {
-      setWorkingCase(resCase)
+    if (!workingCase && data) {
+      setWorkingCase(data.case)
```
gudjong, line 76: Probably makes no difference, but data.case may be undefined even if data is not. Thus, we may be setting the working case to undefined.

apps/judicial-system/web/src/routes/Judge/Overview/Overview.tsx
```diff
@@ -49,30 +50,41 @@ export const JudgeOverview: React.FC = () => {
...
     }
-  }, [workingCase, setWorkingCase, resCase])
+  }, [workingCase, setWorkingCase, data])
+
+  const hasIsolation = (rass: Case): boolean => {
+    const restrictions =
+      rass?.custodyRestrictions !== undefined
+        ? rass.custodyRestrictions
```
gudjong, line 83: I don't think we should use custodyRestrictions here. This should always be based on the request that came from the prosecutor regardless of what the judge decided to do later. This should be exactly the same as in the prosecutor overview.

gudjong (changes requested):

apps/judicial-system/web/src/routes/Judge/Overview/Overview.tsx
addi: I see. I was thinking about the case when the judge had ruled and signed, and would only see the overview. But thats another story i guess.

addi: Lol ups the variable name is rass.

gudjong (approved):

PR #1908, opened 2020-11-16
New layout component

siggakogk:
# Navigation Sidebar Layout

Created a simple sidebar layout component that will be used on all pages that use sidebar
The sidebar is not a part of the grid so it will need to handle it's own width depending on viewport
Sidebar will not resize with content, so it's less busy for the eye

Note: on the gif below the sidebar has yet to be styled
Screen Recording 2020-11-16 at 13 55 33

## Checklist:


apps/web/screens/Layouts/SidebarLayout.tsx
```diff
@@ -0,0 +1,62 @@
...
+}) => (
+  <GridContainer position="none">
+    <Box
+      display="flex"
+      flexDirection="row"
+      height="full"
+      paddingBottom={6}
+      style={{
+        position: absolute ? undefined : 'relative',
+      }}
```
barabrian, line 35: Is it possible to use `Box` props here instead of style?
```suggestion
      position={absolute ? undefined : 'relative'}
```

apps/web/screens/Layouts/SidebarLayout.treat.ts
```diff
@@ -0,0 +1,24 @@
...
+})
+
+export const sticky = style({
+  position: 'sticky',
+  alignSelf: 'flex-start',
+})
+
+export const absolute = style({
+  position: undefined,
+})
```
barabrian, line 24: isn't this kind of misleading?

apps/web/screens/Layouts/SidebarLayout.tsx
```diff
@@ -0,0 +1,62 @@
...
+    >
+      <Hidden
+        print
+        position={absolute ? 'none' : 'relative'}
+        below={hiddenBelow}
+      >
```
barabrian, line 41: we might be able to solve this using `Box`, then we won't need to modify the `Hidden` component
```suggestion
      <Box
        printHidden
        position={absolute ? 'none' : 'relative'}
        display={['none', 'none', 'none', 'block']}
      >
```
this is just an example, might have to tweak it 😄

apps/web/screens/Layouts/SidebarLayout.tsx
```diff
@@ -0,0 +1,62 @@
...
+        <Box
+          className={cn(
+            styles.sidebarWrapper,
+            absolute ? styles.absolute : styles.sticky,
```
barabrian, line 45: it might be time to add `position: sticky` to the `position` prop on the `Box` component that would probably simplify some of the code

barabrian (changes requested):

apps/web/screens/Layouts/SidebarLayout.treat.ts
siggakogk: no 🙄 yes lol!

apps/web/screens/Layouts/SidebarLayout.tsx
siggakogk: mmmmm interesting

barabrian (dismissed):

apps/web/screens/Layouts/SidebarLayout.tsx
```diff
@@ -0,0 +1,57 @@
...
+      paddingBottom={6}
+      position={isSticky ? 'relative' : undefined}
+    >
+      <Box
+        printHidden
+        className={cn(styles.sidebarWrapper, isSticky ? styles.sticky : null)}
+        display={
+          hiddenOnTablet
+            ? ['none', 'none', 'none', 'block', 'block']
+            : ['none', 'none', 'block', 'block', 'block']
```
barabrian, line 41: No need to repeat the last breakpoints if its the same value
```suggestion
            ? ['none', 'none', 'none', 'block']
            : ['none', 'none', 'block']
```

apps/web/screens/Layouts/SidebarLayout.tsx
```diff
@@ -0,0 +1,57 @@
...
+        }
+      >
+        {sidebarContent}
+      </Box>
+      <GridContainer>
+        <GridRow>
+          <GridColumn span={'12/12'}>
+            <Box paddingLeft={[0, 2, 4, 6, 6]}>{children}</Box>
```
barabrian, line 49: No need to repeat the last breakpoints if its the same value
```suggestion
            <Box paddingLeft={[0, 2, 4, 6]}>{children}</Box>
```

apps/web/screens/Layouts/SidebarLayout.tsx
```diff
@@ -0,0 +1,57 @@
...
+    <Box
+      display="flex"
+      flexDirection="row"
+      height="full"
+      paddingBottom={6}
+      position={isSticky ? 'relative' : undefined}
+    >
+      <Box
+        printHidden
+        className={cn(styles.sidebarWrapper, isSticky ? styles.sticky : null)}
```
barabrian, line 37: classnames has a built in solution for conditional styles using object
```suggestion
        className={cn(styles.sidebarWrapper, {[styles.sticky]: isSticky})}
```

barabrian (approved):

barabrian (approved):
Muy bonito!

PR #1915, opened 2020-11-17
Feature/api cata contentful

saevar43:
Connect API catalogue to island.is contentful space.

## What

Fetch static content from island.is contentful space via elastic/graphQL.

## Why

We want our static content to be stored in island.is contentful space and fetched in the same way as elsewhere on the the island.is web.

## Screenshots / Gifs

N/A

## Checklist:


saevarma (dismissed):
Looks good to me.
One minor comment `apps/api-catalogue/types.tsx` doesn't need to be a tsx file, could be .ts or maybe .d.ts?

Another comment for our selves:
`@island.is/web` dependencies would probably need to update as we make this as a `sub-app` under the `web` project. If common graphql schemas like for contentful will be moved into a library for all sub-apps under `web` to use.

PR #1951, opened 2020-11-19
Judicial system/Lock fields pt2

oddsson:
# Lock fields by using Loadash' `isEmpty` function instead of comparing with undefined

## What

Lock fields by using Loadash' `isEmpty` function instead of comparing with undefined. Also, we are now validating the defender email on the hearing arrangement screen.

## Why

Previously we were locking fields if they were undefined. In reality, the fields could be null so that was a bug.

## Checklist:


gudjong (approved):
excellente

PR #1969, opened 2020-11-19
Add table of contents component to island ui

siggakogk:
# Table of contents

Storybook screenshot:
image

Figma screenshot:
image
[Figma link](https://www.figma.com/file/cl7qnkUWOFwgMQKt2FsiZn/H%C3%B6nnun?node-id=4701%3A0)

## Checklist:


libs/island-ui/core/src/lib/TableOfContents/TableOfContents.tsx
```diff
@@ -0,0 +1,36 @@
...
+      <Text variant="h5" as="h5">
+        {title}
+      </Text>
+      {headings.map((text, id) => (
+        <Box
+          key={id}
+          component="button"
+          type="button"
+          textAlign="left"
+          onClick={() => onClick(text)}
```
barabrian, line 25: will it be enough to only take in the text? won't we need some kind of anchor id or something?

barabrian (changes requested):

libs/island-ui/core/src/lib/TableOfContents/TableOfContents.tsx
```diff
@@ -0,0 +1,44 @@
...
+        {tableOfContentsTitle}
+      </Text>
+      {headings.map(({ headingTitle, headingId }, id) => (
+        <Box
+          key={id}
+          component="button"
+          type="button"
+          textAlign="left"
+          onClick={() => onClick(headingId)}
+        >
```
barabrian, line 34: maybe use reakit button instead of Box, I'm not totally sure if we need to adjust the styling but it will provide us with some nice accessibility features

barabrian (approved):
lgtm

PR #1978, opened 2020-11-20
"Feature flag" application templates so they do not appear on production

arnifreyrs:
## What

 Because we have no feature flagging mechanisms in place, I added a boolean attribute to application templates in order to ban POST/PUT these kinds of applications, as well as invalidating the routes for those application types.

## Why

Users of island.is should not be able to create applications like "ExampleForm" and in progress work like ParentalLeave and more.

## Checklist:


libs/application/ui-shell/src/lib/ApplicationForm.tsx
```diff
@@ -35,7 +40,10 @@ const ShellWrapper: FC<{
         const template = await getApplicationTemplateByTypeId(
           application.typeId,
         )
-        if (template !== null) {
+        if (
```
magnearun, line 43: Will this just show a blank page ? We could also redirect to a 404 page ?

arnifreyrs: yes this will show a blank page and this is how it is currently. We have an asana task for rendering better error states and loading states and what not

magnearun (approved):

PR #1979, opened 2020-11-20
Service portal/documents modal

andriandresson:
# \<Document exception modal\>

https://app.asana.com/0/1199218948638353/1199195379824200

## What

- Displaying exception modal to notify the user that external documents are not supported.
- Updated Modal- & ModalBase components.

## Why
- External documents have broken URL's.
- In order to have correct state of the modal visibility, I had to add optional cb func that is fired when the modal visibility is hidden, either by an CTA button or when the user clicks on the DialogBackdrop.

#### Desktop: 
image

#### Mobile: 
image

## Checklist:


Thorrik58 (approved):

libs/island-ui/core/src/lib/ModalBase/ModalBase.tsx
```diff
@@ -71,6 +76,10 @@ export const ModalBase: FC<ModalBaseProps> = ({
     if (toggleClose) closeModal()
   }, [toggleClose])
 
+  useEffect(() => {
+    if (!modal.visible && onClose) onClose()
```
barabrian, line 80: Would it make sense to have this something like?
```suggestion
    onChange(modal.visible)
```
or maybe `onVisibilityChange`, just so we won't need to extend this in the future and change the API, or will that not work for your use case?

barabrian (commented):
LGTM, minor comments

libs/service-portal/documents/src/components/DocumentCard/DocumentCard.treat.ts
```diff
@@ -0,0 +1,13 @@
+import { style } from 'treat'
+import { themeUtils } from '@island.is/island-ui/theme'
+
+export const modalButtonWrapper = style({
+  ...themeUtils.responsiveStyle({
+    sm: {
+      width: '100%',
+    },
+    lg: {
+      width: 'fit-content',
```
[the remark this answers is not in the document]
andriandresson: According to [our handbook](https://github.com/island-is/island.is/blob/main/handbook/technical-overview/technical-direction.md) I believe we do not have to support IE  😅

barabrian (dismissed):
LGTM!

libs/island-ui/core/src/lib/ModalBase/ModalBase.tsx
andriandresson: Good point 👍

Thorrik58 (approved):

atligudl (approved):

PR #1982, opened 2020-11-20
fix: hashfiles

jeremybarbet:
## What

- `apps/**.ts` doesn't match anything https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#patterns-to-match-file-paths
- `apps/**/*.ts` should match root and sub files


sindrig (approved):
Holy cow this list is getting long

PR #1987, opened 2020-11-20
Feature/ids db changes

unnarb:
#Table and model changes to Client

## What

Adding new fields to Client (National Id and Client Type)

## Why

## Screenshots / Gifs

## Checklist:


libs/auth-api-lib/src/lib/entities/models/client.model.ts
```diff
@@ -37,6 +37,24 @@ export class Client extends Model<Client> {
...
   clientId: string
 
+  @Column({
+    type: DataType.STRING,
+    allowNull: true,
+  })
+  @ApiProperty({
+    example: '1234567890',
+  })
+  nationalId: string
```
valurefugl, line 47: Suggestion: maybe we can restrict null here, even though the db allows nulls for now?

valurefugl (dismissed):

libs/auth-api-lib/src/lib/entities/dto/base/client-base-dto.ts
```diff
@@ -8,6 +8,20 @@ import {
 import { ApiProperty } from '@nestjs/swagger'
 
 export abstract class ClientBaseDTO {
+  @IsString()
```
jeremybarbet, line 11: Should this file be named `client-base.dto.ts` to follow the same conventions as others filenames? (e.g. `client.model.ts`)

unnarb: Will fix in another commit

valurefugl (approved):

PR #1995, opened 2020-11-23
As/async select

magnearun:
Adds a Async select component to the application form components

## Screenshots / Gifs

## Checklist:


libs/application/templates/parental-leave/src/forms/ParentalLeaveForm.ts
```diff
@@ -24,6 +25,38 @@ import {
   getEstimatedMonthlyPay,
   getNameAndIdOfSpouse,
 } from '../fields/parentalLeaveUtils'
+import gql from 'graphql-tag'
+
+const GetUnions = gql`
```
magnearun, line 30: @arnifreyrs where do we want to keep these?

libs/application/ui-fields/src/lib/AsyncSelectFormField.tsx
```diff
@@ -0,0 +1,67 @@
...
+  const [options, setOptions] = useState<Option[]>([])
+
+  useEffect(() => {
+    async function load() {
+      const loaded = await loadOptions({ application, apolloClient })
+      setOptions(loaded)
+    }
+
+    load()
+  }, [])
```
arnifreyrs, line 32: you probably have to add the dependencies into the array or?

libs/application/ui-fields/src/lib/AsyncSelectFormField.tsx
```diff
@@ -0,0 +1,67 @@
...
+  FieldDescription,
+} from '@island.is/shared/form-fields'
+import { useLocale } from '@island.is/localization'
+import { useApolloClient } from '@apollo/client/react'
+import { Option } from '@island.is/application/core'
+
+interface Props extends FieldBaseProps {
+  field: AsyncSelectField
+}
+const AsyncSelectFormField: FC<Props> = ({ application, error, field }) => {
```
arnifreyrs, line 19: don't we need to know the loading state or the error from the graphql call?

libs/application/core/src/types/Fields.ts
```diff
@@ -91,6 +99,13 @@ export interface SelectField extends BaseField {
   placeholder?: FormText
 }
 
+export interface AsyncSelectField extends BaseField {
+  readonly type: FieldTypes.ASYNC_SELECT
+  component: FieldComponents.ASYNC_SELECT
+  placeholder?: FormText
+  loadOptions: (c: Context) => Promise<any>
```
arnifreyrs, line 106: `Promise<{label: string, value: unknown}[]>`

arnifreyrs (commented):
I just have some comments. I really like how we can use the client to load the options

libs/application/core/src/lib/fieldBuilders.ts
```diff
@@ -182,6 +185,41 @@ export function buildSelectField(data: {
   }
 }
 
+export function buildAsyncSelectField(data: {
+  condition?: Condition
+  id: string
+  name: FormText
+  description?: FormText
+  placeholder?: string
+  loadOptions: (c: Context) => Promise<any>
```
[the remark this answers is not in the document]
magnearun: yes, was going to get back to that and forgot, Promise<Option[]> right ?

libs/application/ui-fields/src/lib/AsyncSelectFormField.tsx
magnearun: Well...I just want it to run in the first render, but I guess the loadOptions func should be a dep ?

libs/application/ui-fields/src/lib/AsyncSelectFormField.tsx
magnearun: I don't think we need the loading state, options will just be a empty array until it is filled, but I'm adding the error state

libs/application/templates/parental-leave/src/forms/ParentalLeaveForm.ts
magnearun: yes, I'll add that

libs/application/ui-fields/src/lib/AsyncSelectFormField.tsx
arnifreyrs: I just thought the linter would complain

arnifreyrs (approved):
awesome

PR #2001, opened 2020-11-24
chore: definition files

jeremybarbet:
## Why

- Following up on the issue we had on the CI build yesterday

## What

- Move local npm typing to shared folder
- Includes types into tsconfig.base paths to be picked up by others apps/libs
- Re-enable strict mode on ts-node commands

## Checklist:


apps/judicial-system/api/tsconfig.json
```diff
@@ -6,7 +6,6 @@
     "target": "es2015",
     "strict": true
   },
-  "allowJs": true,
```
jeremybarbet, line 9: I deleted it because `allowJs` is part of the `compilerOptions` object and since there is no javascript file in the folder, it's most likely unnecessary

gudjong: 👍  Should be ok.

gudjong (approved):
Nice

pshomov (approved):
LGTM

tsconfig.base.json
```diff
@@ -18,6 +18,7 @@
     "baseUrl": ".",
     "strict": true,
     "paths": {
+      "*": ["libs/shared/types/src/lib/*", "*"],
```
darrikonn, line 21: Why do you need this?
And shouldn't this be in the typeroots section?

jeremybarbet: We spent some time with @arnifreyrs trying to debug that, moving this path as a `typeRoots` never worked. It was giving the following issue:

```text
Could not find a declaration file for module 'entropy-string'. 'node_modules/entropy-string/entropy-string.js' implicitly has an 'any' type.
  Try `npm install @types/entropy-string` if it exists or add a new declaration (.d.ts) file containing `declare module 'entropy-string';`ts(7016)
```

According to this [comment](https://stackoverflow.com/a/40326719), "`typeRoots` are NOT general-purpose ways to load arbitrary declaration (*.d.ts) files."

And it results using this path `*` that matches a library with a definition file if it exists in the directory `libs/shared/types/src/lib/*`.

darrikonn (approved):

sindrig (approved):

dd-sdang (approved):

PR #2003, opened 2020-11-24
Feature/api catalogue webpages

saevar43:
Add first API Catalogue pages to the Web app

## What

Add basic webpages to the web app in the correct places.

## Why

The decision was made to move the API Catalogue into the web app and this is the first step in doing so.

## Screenshots / Gifs

N/A

## Checklist:


saevarma (dismissed):
LGTM

stefanhar:
I assume this is the starting point for the api catalog integration into the web.
Code looks good to me, but I have a few questions about how this should appear in production since we should always try to keep the main branch of web production ready.
Are we ok with this becoming accessible as incomplete pages on the web?
Are `/throun/vefthjonustur/vorulisti/` and `/throun/vefthjonustur/vorulisti/[slug]` the final paths or is the plan to have the api catalog on its own path like `/covid-adgerdir`?

saevarma:
> I assume this is the starting point for the api catalog integration into the web.
> Code looks good to me, but I have a few questions about how this should appear in production since we should always try to keep the main branch of web production ready.
> Are we ok with this becoming accessible as incomplete pages on the web?
> Are `/throun/vefthjonustur/vorulisti/` and `/throun/vefthjonustur/vorulisti/[slug]` the final paths or is the plan to have the api catalog on its own path like `/covid-adgerdir`?

These paths are final but the question about production ready is a good question.
I guess we don't want these pages in production until they are final, but long living feature branches are discouraged as the repo is moving fast and we have also bad experience with being to long in our own branch.
There was some talk about feature flags some while ago, I don't know the status on that but we could look into that.

stefanhar:
A very simple approach to making this not accessible on prod would be something like this in the top of screen components:
``` ts
import getConfig from 'next/config'

const { disableApiCatalog } = getConfig()
if(disableApiCatalog): {
  throw new CustomNextError(404, 'Not found')
}
```

Then pass an environment variable through `next.config.js`

saevar43:
I've added configuration that disables the Api Catalogue pages in the web.

apps/web/next.config.js
```diff
@@ -77,6 +77,7 @@ module.exports = withTreat(
         graphqlUrl: '',
         graphqlEndpoint: graphqlPath,
         SENTRY_DSN,
+        disableApiCatalog: true,
```
stefanhar, line 80: Should this maybe be an environment variable so we can have these pages accessible on dev and staging for testing?

saevar43: Ah yeah, that's a good point. I'll fix that.

saevar43:
I've made disableApiCatalog an env variable now.

stefanhar (approved):
LGTM

PR #2010, opened 2020-11-24
Judicial System: Audit Trail

gudjong:
# Audit Trail Library

https://app.asana.com/0/1199153462262248/1189677453581106

## What

Creates a library for creating an audit trail in AWS CloudWatch and uses it to audit which users view which cases.

## Why

We need to track which cases judicial system users view.

## Checklist:


pshomov (commented):
For this to start working we need to grant your service access to CloudWatch. What is the name of the log group you are using and is it maybe configurable?

gudjong:
> For this to start working we need to grant your service access to CloudWatch. What is the name of the log group you are using and is it maybe configurable?

I configured helm so that the normal logger is used for now.

I expect the audit trail to get three env variables from helm:
- log group name
- log stream name
- aws region

So, the log group name is configurable and it would be great if the log stream name could be unique to each pod. If I need to do that in code, then I will make some changes.

These are the options that the audit trail lib accepts:
export interface AuditTrailOptions {
  useGenericLogger: boolean // should be false in production environments
  groupName?: string
  streamName?: string
  region?: string
}

I will read these options from the environment in the projects that use the trail.

oddsson (approved):

addi (approved):

pshomov (approved):
LGTM

PR #2023, opened 2020-11-25
updating the method used to create ids for services

Addips:
Updating the way we create IDs for services from xroad, so the URLs stay the same after night schedule run

## What

Create a single reusable id for each service

## Why

So that URLs can be reused

## Checklist:


saevarma (approved):
Looks good

PR #2024, opened 2020-11-25
Updated route path resolver

siggakogk:
# new path resolver for next links 

## Checklist:


apps/web/i18n/routes.tsx
```diff
@@ -0,0 +1,109 @@
...
+  | 'page'
+  | 'category'
+  | 'contentcategory'
+  | 'articlecategory'
+  | 'news'
+  | 'search'
+  | 'lifeevent'
+  | 'lifeeventpage'
+  | 'adgerdir'
+  | 'vidspyrna-frontpage'
```
stefanhar, line 20: Looks like some types are missing e.g. vidspyrnapage.
Is the plan to add missing types when we integrate this into the rest of the web?

apps/web/i18n/routes.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+export const routes: Record<ContentType, Record<Locale, AnchorAttributes>> = {
+  article: {
+    is: { as: '/[slug]', href: '/[slug]' },
+    en: { as: '/en/[slug]', href: '/en/[slug]' },
+  },
+  page: {
+    is: { as: '/stofnanir/[slug]', href: '/stofnanir/stafraent-island' }, // TODO FOLDER STRUCTURE NEEDS FIXING
```
stefanhar, line 28: This looks like the only instance where `as !== href`, maybe we can simplify the `routes` mapping object by changing `{as: string, href: string}`to string e.g. `{is: '/[slug]', en: '/en/[slug]'}`.
Then we can handle the temporary edgecase of `/stofnanir/stafraent-island` inside the resolver function

apps/web/i18n/routes.tsx
```diff
@@ -0,0 +1,109 @@
...
+  slugs?: Array<string>,
+): AnchorAttributes => {
+  const typePath = routes[type.toLowerCase()][locale]
+  const path: AnchorAttributes = typePath
+
+  if (type === 'vidspyrna-frontpage') {
+    path.as = removeSlugFromString(path.as)
+    path.href = removeSlugFromString(path.href)
+    return path
+  }
```
stefanhar, line 94: Is this needed? We might be able to do this in the routes mapping object:
```ts
'vidspyrna-frontpage': {
  is: '/covid-adgerdir',
  en: '/en/covid-operations'
}
```

apps/web/i18n/routes.tsx
siggakogk: either way works fine to me :) i just moved over the types that were already defined in the project, if you can point out some overview of all types i can add the missing ones in, or add them in one by one

apps/web/i18n/routes.tsx
```diff
@@ -0,0 +1,101 @@
...
+  let path: AnchorAttributes = { as: '/', href: '/' }
+  const type = String(contentType).toLowerCase()
+
+  if (routes[type]) {
+    const typePath: string = routes[type][locale]
+    path = { as: typePath, href: typePath }
+
+    if (slugs && slugs.length > 0) {
+      for (let i = 0; i < slugs.length; i++) {
+        path.as = typePath.replace('[slug]', slugs[i])
```
stefanhar, line 87: The url template string might include path params that are not called slug.
Maybe we could use regex to replace instead of slug?
```suggestion
        path.as = typePath.replace(/\[\w+\]/, slugs[i])
```

stefanhar (approved):

PR #2026, opened 2020-11-25
[WIP] As/email

magnearun:
Let application templates send emails on state changes

## Screenshots / Gifs

## Checklist:


libs/application/core/src/lib/ApplicationTemplateHelper.ts
```diff
@@ -32,11 +41,14 @@ export class ApplicationTemplateHelper<
...
-  private initializeStateMachine(stateMachineContext?: TContext) {
+  private initializeStateMachine(
+    stateMachineContext?: TContext,
+    stateMachineOptions?: Partial<MachineOptions<TContext, TEvents>>,
+  ) {
     this.stateMachine = createApplicationMachine(
       this.application,
       this.template.stateMachineConfig,
-      this.template.stateMachineOptions,
+      merge({}, stateMachineOptions, this.template.stateMachineOptions),
```
arnifreyrs: not sure you need merge, you could do:

{...stateMachineOptions, ...this.template.stateMachineOptions}

libs/application/core/src/lib/ApplicationTemplateHelper.ts
```diff
@@ -32,11 +41,14 @@ export class ApplicationTemplateHelper<
...
-  private initializeStateMachine(stateMachineContext?: TContext) {
+  private initializeStateMachine(
+    stateMachineContext?: TContext,
+    stateMachineOptions?: Partial<MachineOptions<TContext, TEvents>>,
+  ) {
     this.stateMachine = createApplicationMachine(
       this.application,
       this.template.stateMachineConfig,
-      this.template.stateMachineOptions,
+      merge(stateMachineOptions, this.template.stateMachineOptions),
```
arnifreyrs, line 51: you can also just do {...stateMachineOptions, ...this.template.stateMachineOptions} right?

libs/application/core/src/lib/ApplicationTemplateHelper.ts
```diff
@@ -58,8 +70,22 @@ export class ApplicationTemplateHelper<
...
+    emailService: any,
+  ): [boolean, string, Application] {
+    this.initializeStateMachine(undefined, {
+      services: {
+        emailService: async (
+          context: TContext,
+          event: TEvents,
+          { src }: InvokeMeta,
+        ) => {
+          const template = src.template(context)
```
arnifreyrs, line 84: maybe use emailTemplate instead as template is used for the application template, just a thought

libs/application/core/src/lib/ApplicationTemplateHelper.ts
```diff
@@ -58,8 +70,22 @@ export class ApplicationTemplateHelper<
...
+          return emailService.sendEmail(template)
```
arnifreyrs, line 85: maybe we have to NOT do this if there is no emailTemplate?

arnifreyrs (commented):
Awesome, I feel like this is a simple approach and shows us how we can do this in a straight forward manner, maybe the only change needed is to not send an email if there is no emailtemplate for this state transition

libs/application/core/src/lib/ApplicationTemplateHelper.ts
magnearun: I want to merge the { services: { ... } } part, and not overide other services that have been declared in the state machine from the template

libs/application/core/src/lib/ApplicationTemplateHelper.ts
magnearun: yes, good point

libs/application/core/src/lib/ApplicationTemplateHelper.ts
magnearun: This function is not called unless the
`invoke: {
          src: {
            type: 'emailService',
            emailTemplate: assignParentTemplate,
          },
        },
`
part is added to the state, meaning it will invoke the emailService when transitioning to this state, but adding a check for the emailTemplate is a good thing, I'm also going to fix the types.

libs/application/core/src/lib/ApplicationTemplateHelper.ts
arnifreyrs: of course!

libs/application/core/src/lib/ApplicationTemplateHelper.ts
```diff
@@ -58,8 +77,34 @@ export class ApplicationTemplateHelper<
    * @param event A state machine event
    * returns [hasChanged, newState, newApplication] where newApplication has the updated state value
    */
-  changeState(event: Event<TEvents>): [boolean, string, Application] {
-    this.initializeStateMachine()
+  changeState(
+    event: Event<TEvents>,
+    sendEmail: (emailTemplate: SendMailOptions) => Promise<string>,
```
arnifreyrs, line 82: could this be undefined and we just provide default behavior (do nothing?) when this is undefined?

arnifreyrs (approved):
looks good

libs/application/core/src/lib/ApplicationTemplateHelper.ts
magnearun: yes, I'll fix that

PR #2034, opened 2020-11-26
island-ui/Update the Tag component

oddsson:
# Update the Tag component

## What

Remove prop `label` and `bordered` and add a prop `outlined` to the Tag component.

## Why

According to the design, the tag component should be able to be filled or outlined. The filled version is clickable while the outlined is not. Also, according to the desing, we should not offer filled and bordered tags.

## Screenshots / Gifs

Screen Shot 2020-11-26 at 11 38 36

## Checklist:


addi (dismissed):

atligudl (approved):
Service Portal looks good

atligudl (approved):

barabrian (approved):
LGTM

jeremybarbet (approved):
LGTM

dd-sdang (approved):

libs/island-ui/core/src/lib/Tag/Tag.tsx
```diff
@@ -70,7 +67,7 @@ export const Tag = forwardRef<HTMLButtonElement & HTMLAnchorElement, TagProps>(
       </Text>
     )
 
-    if (label) {
+    if (outlined) {
```
barabrian, line 70: Just a though for another PR.. I'm not sure we want to define clickable or not by the outlined value. Would it make sense to use a button if the `onClick` function is present? something like what's happening with the `href` attribute? so if there is no `href` and no `onClick` then we would return this span?

addi (approved):

PR #2036, opened 2020-11-26
Update providers to include the org name from xroad

Addips:
## What

Updating the service registration to use the org name available from x-road

## Why

So that the org name is tied to x-road and prevent users from setting multiple org names by accident

## Checklist:

### DIFF
--- a/libs/api-catalogue/services/src/lib/provider.service.ts
+++ b/libs/api-catalogue/services/src/lib/provider.service.ts
@@ -46,7 +46,7 @@ export class ProviderService {
           (item: XroadIdentifier): Provider => {
             return {
               type: providerType,
-              name: item.name ?? '',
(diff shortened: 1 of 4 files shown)


libs/api-catalogue/services/src/lib/provider.service.ts
```diff
@@ -46,6 +46,7 @@ export class ProviderService {
           (item: XroadIdentifier): Provider => {
             return {
               type: providerType,
+              name: item.name ?? '',
```
saevarma, line 49: Maybe we should have the default name behavior happening here instead of in L64 in `restmetadata.service.ts`
As here we are responsible to set the provider, and each provider is classified as the subsystem in question.

saevarma (changes requested):
See my comment below.

saevarma (approved):
LGTM

PR #2046, opened 2020-11-26
Application system - mobile refinements 1

madebynoam:
Various mobile refinements and fixes for the application system:

1. Fix cases where a horizontal scroll bar appears
2. Update Continue and Back buttons to match latest designs
3. Fix logo not appearing
4. Turn off auto focus fields
5. Scroll to top on question transition
6. Make the stepper sticky on mobile size
7. Make the stepper move to the active section on mobiles size
8. Wrap the the table with `overflowX: scroll` so that it doesn't create horizontal scroll bars for the main shell.
9. Start setting field types to more semantic attributes (for the relevant mobile keyboards).

Remaining (for next PR):
1. Make chapter images full width on mobile size
2. Add a sub-section progress bar, since the stepper only shows main sections on mobile size.
3. Continue investigating how to do `escapedGrid` blue containers.

## Screenshots / Gifs

application-shell-mobile

## Checklist:

### DIFF
--- a/libs/application/ui-shell/src/components/FormMultiField.tsx
+++ b/libs/application/ui-shell/src/components/FormMultiField.tsx
@@ -40,7 +40,6 @@ const FormMultiField: FC<{
                 showFieldName
                 field={field as FieldDef}
                 key={field.id}
-                autoFocus={index === 0}
                 errors={errors}
               />
             </Box>
--- a/libs/island-ui/core/src/lib/FormStepper/FormStepperSection.tsx
+++ b/libs/island-ui/core/src/lib/FormStepper/FormStepperSection.tsx
@@ -41,22 +41,29 @@ export const FormStepperSection: FC<{
   const subSections = getSubSectionsInSection(section, subSection)
   const hasSubSections = subSections.length > 0
   const containerRef = useRef<HTMLDivElement>(null)
-  const { height: activeHeight, width: containerWidth } = useComponentSize(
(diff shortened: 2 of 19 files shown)


jeremybarbet (approved):
LGTM

libs/island-ui/core/src/lib/FormStepper/FormStepperSection.tsx
```diff
@@ -40,57 +41,92 @@ export const FormStepperSection: FC<{
...
+    }
+
+    handleResize()
+
+    window.addEventListener('resize', handleResize)
+
+    return () => {
+      window.removeEventListener('resize', handleResize)
+    }
+  }, [, isClient])
```
arnifreyrs, line 86: do you need to have empty first element in the dependency array?

libs/application/ui-shell/src/components/FormMultiField.tsx
```diff
@@ -40,7 +40,6 @@ const FormMultiField: FC<{
                 showFieldName
                 field={field as FieldDef}
                 key={field.id}
-                autoFocus={index === 0}
```
arnifreyrs, line 43: should we remove `autoFocus` from props in `FormField` as well?

arnifreyrs (commented):
Awesome stuff! My comments are just minor and something you could consider fixing in the next PR

PR #2066, opened 2020-11-28
Update auth configuration for service portal

valurefugl:
# Suggested changes to auth configuration

## What

- Enable silent renew
- Store user in session storage
- Only re-authenticate if no user in session storage
- offline_access scope not requested (refresh tokens not needed)
- Add post_logout_redirect_uri

## Why

Streamline login and logout flows. Harden security.

## Screenshots / Gifs

## Checklist:

- [ x] I have performed a self-review of my own code
- [ x] My changes generate no new warnings


apps/service-portal/src/utils/userManager.ts
```diff
@@ -11,17 +11,18 @@ const settings: UserManagerSettings = {
...
+  // eslint-disable-next-line @typescript-eslint/camelcase
   response_type: 'code',
   revokeAccessTokenOnSignout: true,
   loadUserInfo: true,
-  automaticSilentRenew: false,
-  scope: 'openid profile offline_access api_resource.scope',
-  userStore: new WebStorageStateStore({ store: new InMemoryWebStorage() }),
+  automaticSilentRenew: true,
+  scope: 'openid profile api_resource.scope',
+  userStore: new WebStorageStateStore({ store: window.sessionStorage }),
```
obmagnusson, line 25: Why the change to session storage? Wouldnt this be vulnarable to XSS?

valurefugl: Potentially yes, and we should mitigate that risk. The upside is that we don't need to contact auth server on refresh.
Note that the in-memory storage also makes the tokens accessible in the browser.

apps/service-portal/src/hooks/useAuth/useAuth.ts
```diff
@@ -1,6 +1,7 @@
 import { useStore } from '../../store/stateProvider'
 import { ActionType } from '../../store/actions'
 import { userManager } from '../../utils/userManager'
+import { User } from '@sentry/react'
```
obmagnusson, line 4: This is not used.

apps/service-portal/src/utils/userManager.ts
obmagnusson: Yes but the risk I would guess is less when dealing with In-memory storage.  But maybe the upside outweighs the potentential risk. Am assuming this will or has been cleared with Syndis.

obmagnusson (changes requested):
Miniscule changes

apps/service-portal/src/utils/userManager.ts
valurefugl: Haven't cleared with Syndis but we should. As I see it the tokens are either stored in the browser or not, in-memory storage might be a bit more obscure but the tokens are still there.

apps/service-portal/src/hooks/useAuth/useAuth.ts
valurefugl: Fixed.

obmagnusson (approved):
LGTM

PR #2074, opened 2020-11-30
Allow splitting radio buttons in half

darrikonn:
# Allow splitting radio buttons in half in applications

## What

Allow splitting radio buttons in half

## Why

**width="half"** will only split the component width in half, but radio button renders multiple
children for a single component. A different action is thus needed.

### halve=true commit e0cc09ed91b5611f413246e2085795997b1d6d73 ... width="half" commit 28e6d00a3b311c3aad769d67ca0d6a35099d463a
image

### halve=false commit e0cc09ed91b5611f413246e2085795997b1d6d73 ... width="full" commit 28e6d00a3b311c3aad769d67ca0d6a35099d463a
image

### width="half" only related to commit e0cc09ed91b5611f413246e2085795997b1d6d73 and previous behaviour
image

## Checklist:


darrikonn:
A better API would be to use the `width="half"` prop, ignore it in `FormMultiField`, and pass it down to the field.

But then we would have to take component specific action in `FormMultiField`:
```typescript
const IGNORED_HALF_TYPES: FieldTypes[] = ["RADIO"]
...
const isHalfColumn = !IGNORED_HALF_TYPES.includes(field.type) && field?.width === 'half'
const span = isHalfColumn ? '1/2' : '1/1'
```

But this might lead to "hard to debug" problem for other developers.
Your thoughts!?

libs/application/templates/driving-license/src/forms/application.ts
```diff
@@ -29,12 +29,15 @@ export const application: Form = buildForm({
...
+              width: 'half',
               options: [
-                { value: 'car', label: 'Bifreiða- eða bifhjólaréttindi' },
+                { value: 'general', label: 'Almenn ökuréttindi' },
                 { value: 'truck', label: 'Vöru- eða hópbifreiðaréttindi' },
-                { value: 'tractor', label: 'Dráttarvélaréttindi' },
+                { value: 'bike', label: 'Bifhjólaréttindi' },
                 { value: 'taxi', label: 'Leigubílaréttindi' },
+                { value: 'tractor', label: 'Dráttarvélaréttindi' },
+                { value: 'trailer', label: 'Kerrur og eftirvagnar' },
```
dabbeg, line 40: Are you not using the messages.ts for translatable names?

dabbeg (dismissed):
SLGMTM

libs/application/templates/driving-license/src/forms/application.ts
darrikonn: Nope, not yet. Started like this to get familiar with the application system.

arnifreyrs:
The only thing I think about in relation to this, is maybe `width=half | full` is too restrictive in general. What if we want to fit three radio answers in one line? or more? But we can solve it later.

I would love feedback from @madebynoam on this :)

arnifreyrs (approved):
Like it! just one thought about the `width` attribute in general

darrikonn:
> The only thing I think about in relation to this, is maybe `width=half | full` is too restrictive in general. What if we want to fit three radio answers in one line? or more? But we can solve it later.
> 
> I would love feedback from @madebynoam on this :)

Yeah, it's just in line with current application UI setup, and all these changes are only related to the application system.
I agree that the current width might be too restrictive, but that task is broader than only these Radio Button changes and should be implemented in another PR.

darrikonn:
> The only thing I think about in relation to this, is maybe `width=half | full` is too restrictive in general. What if we want to fit three radio answers in one line? or more? But we can solve it later.
> 
> I would love feedback from @madebynoam on this :)

Yeah, the "width" prop is related to the application system, and is in line with current application UI setup.
What needs to be taken further is the new `halve` boolean prop. That should probably be a `splitwise` prop that takes in positive integers. We could modify `flex-basis` based on the split.

```typescript.jsx
<RadioButton splitwise={2} />
```

```typescript
export const root = style({
  flexGrow: 1,
  flexShrink: 1,
})

export const split2 = style({
  flexBasis: '50%',
})

export const split3 = style({
  flexBasis: '33.3%',
})

...
```

libs/shared/form-fields/src/lib/RadioController/RadioController.tsx
```diff
@@ -58,10 +78,32 @@ export const RadioController: FC<Props> = ({
...
+              {renderRadioButton((child) => (
+                <Box
+                  paddingLeft={leftSpace}
+                  paddingTop={topSpace}
+                  className={styles.coverHalf}
+                >
+                  {child}
+                </Box>
+              ))}
+            </Box>
```
barabrian, line 102: is there any reason why we are using custom layout here rather than using `GridRow` and `GridColumn`?

darrikonn: Ahh nice, that would also solve allowing further splitting :+1:

libs/shared/form-fields/src/lib/RadioController/RadioController.tsx
```diff
@@ -1,7 +1,18 @@
 import React, { FC } from 'react'
-import { Box, RadioButton, Stack, Tooltip } from '@island.is/island-ui/core'
+import cs from 'classnames'
```
barabrian: I've seen classnames use the acronym `cn` in the codebase, maybe we should decide what acronym to go with moving forward to keep consistency

darrikonn:
Updated. Now supports multi-splitting the radio buttons

barabrian (approved):
LGTM

libs/shared/form-fields/src/lib/RadioController/RadioController.tsx
darrikonn: I already removed the import and force-pushed :).

But yeah, this was supposed to be cn. Both `classnames` and `cn` are being used -> the reason for why default exports are not goooood :D

darrikonn: default exports are good and are super descriptive if they are reexported as named exports.

dabbeg (approved):

arnifreyrs (approved):

PR #2077, opened 2020-11-30
Feature/document provider api

fjandakornid:
# DocumentProviderAPI

## What

API that serves both the onboarding application for document-providers and the document-provider admin portal (not yet implemented). Only mocked data at the moment.

## Why

Same as the above

## Checklist:


pshomov (dismissed):
LGTM

libs/api/domains/document-provider/README.md
```diff
@@ -0,0 +1,7 @@
+# Document-Provider-API
+
+This library was generated with [Nx](https://nx.dev).
```
arnifreyrs, line 3: Consider improving the README

libs/api/domains/document-provider/src/lib/document-provider.resolver.ts
```diff
@@ -0,0 +1,37 @@
...
+
+@Resolver()
+export class DocumentProviderResolver {
+  constructor(private documentProviderService: DocumentProviderService) {}
+
+  @Mutation(() => ClientCredentials)
+  async registerProvider(
+    @Args('input') input: RegisterProviderInput,
+  ): Promise<ClientCredentials> {
+    return this.documentProviderService.registerProvider(input.nationalId)
```
arnifreyrs, line 18: you could read the national id from the auth token that associates the request

arnifreyrs (dismissed):
Looks good to me, had two comments that can be improved in a future pr

libs/api/domains/document-provider/README.md
fjandakornid: Added minor documentation... will add more when there's more meat on the bones

libs/api/domains/document-provider/src/lib/document-provider.resolver.ts
fjandakornid: that's true and we will.. just added this for now since there will be some parameters but we don't know which ones atm..

arnifreyrs (approved):
lgtm

PR #2078, opened 2020-11-30
dl/populate countries

darrikonn:
# Populate select box with countries

## Checklist:

### DIFF
--- a/libs/application/core/src/types/Fields.ts
+++ b/libs/application/core/src/types/Fields.ts
@@ -89,7 +89,6 @@ export interface RadioField extends BaseField {
   component: FieldComponents.RADIO
   options: MaybeWithApplication<Option[]>
   emphasize?: boolean
-  halve?: boolean
   largeButtons?: boolean
 }
 
--- a/libs/application/core/src/types/Fields.ts
+++ b/libs/application/core/src/types/Fields.ts
@@ -89,6 +89,7 @@ export interface RadioField extends BaseField {
   component: FieldComponents.RADIO
   options: MaybeWithApplication<Option[]>
   emphasize?: boolean
+  halve?: boolean
   largeButtons?: boolean
 }
 
--- a/libs/application/core/src/types/Fields.ts
+++ b/libs/application/core/src/types/Fields.ts
@@ -96,6 +96,7 @@ export interface SelectField extends BaseField {
   readonly type: FieldTypes.SELECT
   component: FieldComponents.SELECT
   options: MaybeWithApplication<Option[]>
+  defaultValue?: Option
   placeholder?: FormText
 }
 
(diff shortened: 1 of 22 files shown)


libs/shared/data/src/lib/countries.ts
```diff
@@ -0,0 +1,998 @@
...
+  },
+  {
+    value: 'KE',
+    label: 'Kenya',
+  },
+  {
+    value: 'KI',
+    label: 'Kiribati',
+  },
+  {
```
arnifreyrs, line 470: where is Kosovo? :/

libs/application/core/src/types/Fields.ts
```diff
@@ -96,6 +96,7 @@ export interface SelectField extends BaseField {
   readonly type: FieldTypes.SELECT
   component: FieldComponents.SELECT
   options: MaybeWithApplication<Option[]>
+  defaultValue?: Option
```
arnifreyrs, line 99: nice, I like this. We will definitely need to add `defaultValue` to many other fields later on, maybe even `BaseField`

arnifreyrs (dismissed):
great stuff

libs/shared/data/src/lib/countries.ts
darrikonn: Hahah, you just went through the whole country list :D ... good spot!

darrikonn: Soooo.. which native part should I pick:
```
Republika e Kosovës  (Albanian)
Република Косово  (Serbian)
```

arnifreyrs: Woah, tricky. Albanian is the most spoken language, so let's go for that https://en.wikipedia.org/wiki/Demographics_of_Kosovo#Languages

sindrig (commented):
Shared data module sounds kinda scary TBH. Right now and for countries it's probably okay (not everyone relies on it and it needs some major event to change). But the name sounds ambiguous enough that someone might add all the shops in Kringlan at one point, and when one of the shops goes bankrupt we need to rebuild everything, even the modules that only rely on the country list and not the shops.

Just my 2c, maybe you want to think about renaming the module to something more concrete like `countries-list-data`.

darrikonn:
> Shared data module sounds kinda scary TBH. Right now and for countries it's probably okay (not everyone relies on it and it needs some major event to change). But the name sounds ambiguous enough that someone might add all the shops in Kringlan at one point, and when one of the shops goes bankrupt we need to rebuild everything, even the modules that only rely on the country list and not the shops.
> 
> Just my 2c, maybe you want to think about renaming the module to something more concrete like `countries-list-data`.

Yeah, wasn't quite pleased with the name. But then again, I wasn't pleased with the name `shared` either since it's super broad. What is shared?

I agree, renaming the module is better!

darrikonn:
Change in design, won't need it.

Agree with you @sindrig, the name was baaaad and too broad

PR #2089, opened 2020-12-01
Add personal allowance sub section to the flow.

arnifreyrs:
Remove blue boxed radio buttons in most cases (requested by designers).

Change wording related to transferring rights.

# Screenshots

New personal allowance screen:
Screenshot 2020-12-01 at 15 41 50

Blue radio buttons before:
Screenshot 2020-12-01 at 15 41 10

And after:
Screenshot 2020-12-01 at 15 39 23

## Checklist:

### DIFF
--- a/libs/application/templates/parental-leave/src/forms/ParentalLeaveForm.ts
+++ b/libs/application/templates/parental-leave/src/forms/ParentalLeaveForm.ts
@@ -25,6 +25,7 @@ import {
   getNameAndIdOfSpouse,
 } from '../fields/parentalLeaveUtils'
 import { GetPensionFunds, GetUnions } from '../graphql/queries'
+import { NO, YES } from '../constants'
 
 interface SelectItem {
   id: string
@@ -115,7 +116,7 @@ export const ParentalLeaveForm: Form = buildForm({
                     )
                     const options: Option[] = [
                       {
-                        value: 'no',
+                        value: NO,
                         label: m.noOtherParent,
                       },
                       { value: 'manual', label: m.otherParentOption },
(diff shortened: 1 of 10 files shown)


madebynoam (approved):
lgtm!

libs/application/templates/parental-leave/src/forms/ParentalLeaveForm.ts
```diff
@@ -245,6 +232,106 @@ export const ParentalLeaveForm: Form = buildForm({
...
+                buildSelectField({
+                  // This should probably be a text input with a format, and type number
+                  name: m.paymentInformationPersonalAllowance,
+                  id: 'personalAllowance.usage',
+                  width: 'half',
+                  options: [
+                    { label: '100%', value: '100' },
+                    { label: '75%', value: '75' },
+                    { label: '50%', value: '50' },
+                    { label: '25%', value: '25' },
```
baering, line 262: Are values always strings?
Does the schema handle converting this to a number?

Same below

libs/application/templates/parental-leave/src/forms/ParentalLeaveForm.ts
```diff
@@ -245,6 +232,106 @@ export const ParentalLeaveForm: Form = buildForm({
...
         }),
+        buildSubSection({
+          id: 'personalAllowanceSubSection',
+          name: m.personalAllowanceName,
+          children: [
+            buildRadioField({
+              id: 'usePersonalAllowance',
+              name: m.usePersonalAllowance,
+              options: [
+                { label: m.yesOptionLabel, value: 'yes' },
```
baering, line 243: From experience I would recommend extracting string constants like these (`yes`/`no`), since they have to also be referenced from the data schema, avoiding duplication and magic values.

baering (dismissed):
LGTM 👍

libs/application/templates/parental-leave/src/forms/ParentalLeaveForm.ts
arnifreyrs: So this is one of the things that definitely needs improvement in the application system. React hook form gets all the values from the inputs that are registered on each screen, and the values are almost always of type string. This can be manipulated by the use of `setValue` and casting to the desired type, but that is problematic as well.

Zod, the dataschema library, does not yet support transformation properly, but version 2 which is in beta does. Then we can use the dataschema lib to transform the strings into desired formats when needed and store the answers like that properly. 

I am all ears if you have any ideas how we can improve this, as I agree that this is not good enough

madebynoam (dismissed):

magnearun (approved):

magnearun (dismissed):

madebynoam (approved):

PR #2094, opened 2020-12-02
Judicial system: PDF Preview

gudjong:
# PDF Preview

https://app.asana.com/0/1199153462262248/1199153462262162

## What

Lets the judge review the ruling PDF before signing it.

## Why

Being able to preview documents before signing them is preferred.

## Screenshots / Gifs

image

## Checklist:


apps/judicial-system/backend/src/app/modules/case/e2e/case.spec.ts
```diff
@@ -332,6 +341,7 @@ describe('Case', () => {
       await Case.create(getCaseData()).then(async () => {
```
oddsson, line 341: Just want to point out that you are mixing `async` and `await` with `then`. I'm assuming that all instances of `then` could be `await`. Not asking for this to change just pointing it out in case you were unaware :)

oddsson (approved):

sindrig (approved):
LGTM

PR #2096, opened 2020-12-02
Website/landmarks

ingolfurs:
# Use ARIA landmarks

## What
Use HTML5 semantic tags or ARIA landmarks to identify page regions.

## Why
Using aria landmarks to define page regions allows screen reader user the ability to easily jump from one section to another. 
According to W3 WCAG ([Using ARIA landmarks to identify regions of a page](https://www.w3.org/TR/WCAG20-TECHS/ARIA11.html)):
> It is a best practice to include ALL content on the page in landmarks, so that screen reader users who rely on them to navigate from section to section do not lose track of content.

## Screenshots / Gifs
Frontpage landmarks:
Before:
one-landmark

After:
landmarks

## Checklist:


libs/island-ui/core/src/lib/Main/Main.tsx
```diff
@@ -0,0 +1,20 @@
+import React, { useRef } from 'react'
```
jeremybarbet, line 1: Does this main component has to be part of the island-ui? maybe it could be local to the web project?

libs/island-ui/core/src/lib/Page/Page.tsx
```diff
@@ -1,7 +1,14 @@
 import React from 'react'
 
 import * as styles from './Page.treat'
+import { Box, BoxProps } from '../..'
```
jeremybarbet, line 4:
```suggestion
import { Box, BoxProps } from '../../Box/Box'
```

libs/island-ui/core/src/lib/Main/Main.tsx
```diff
@@ -0,0 +1,20 @@
...
+import { Box, BoxProps } from '../..'
```
jeremybarbet, line 2:
```suggestion
import { Box, BoxProps } from '../../Box/Box'
```

libs/island-ui/core/src/lib/Main/Main.tsx
ingolfurs: I've moved it to the web project

jeremybarbet (approved):
LGTM on the island-ui part 👍

PR #2103, opened 2020-12-02
Island-ui/Add the ability to have icons come before text in buttons

oddsson:
# Add the ability to have icons come before text in buttons

## What

Add prop to the button component to make icons appear before the button text. 

## Why

According to the desing, this should be available. 

## Screenshots / Gifs

Screen Shot 2020-12-02 at 15 00 14
Screen Shot 2020-12-02 at 15 00 04

## Checklist:


libs/island-ui/core/src/lib/Button/Button.tsx
```diff
@@ -50,6 +50,7 @@ export interface ButtonProps {
   fluid?: boolean
   icon?: IconType
   iconType?: Type
+  iconPosition?: 'start' | 'end'
```
barabrian, line 53: Maybe we should try implementing 'start' as a separate icon prop, that would make it a bit more flexible just incase we need a Button with icons on both sides like a user dropdown e.g. `[ 👤  User options V ]` (hope you get the idea 😄 ). Not sure what we should call it, maybe something with "pre" attached to it. And we could let the right side icon be default `icon` as it will probably be used more frequently

oddsson: I think that variant of the button wouldn't have icons on both sides, it would have a `[image text icon]`, not `[icon text icon]` see 👇

Screen Shot 2020-12-03 at 08 34 58

There is no mention of a `[icon text icon]` button in the design and I think we should implement that when/if it's ever added to the design.

barabrian: Maybe that was a bad example 😄  but basically what I'm thinking is when we were implementing the button there was a rule that icons should only be on the right, but now that has changed. So I asked the designers if there is a possibility in the future that they would use icons on both sides and they said that could happen. So it would probably be better to keep the option open on our side form the start

barabrian (commented):
https://github.com/island-is/island.is/pull/2103#discussion_r535054202

libs/island-ui/core/src/lib/Button/Button.tsx
oddsson: Right, I'll get right on that :)

barabrian (approved):
nice! LGTM

addi (approved):

PR #2106, opened 2020-12-02
Sjukratryggingar/initial setup of health insurance application

dd-sdang:
# Sjukratryggingar - Initial setup of health insurance application

## What

Added new library for health insurance application in Application System with initial structure and dummy template.

## Why

New application in the Application System for Health Insurance

## Checklist:


libs/application/templates/health-insurance/README.md
```diff
@@ -0,0 +1,9 @@
+<!-- gitbook-ignore -->
+
+# Application for health insurance
+
+Application for health insurance when moving back to Iceland after living abroad.
```
arnifreyrs, line 5: i am wondering, does a general application for health insurance exist somewhere for Sjúkratryggingar?

My concern is that maybe this application you are working on is specific to health insurance for people moving back to iceland, so the HEALTH_INSURANCE enum value is "taken".

libs/application/templates/health-insurance/src/lib/HealthInsuranceTemplate.ts
```diff
@@ -0,0 +1,68 @@
...
+  infoInput: z.string(),
+  summaryInput: z.string(),
+})
+
+const HealthInsuranceTemplate: ApplicationTemplate<
+  ApplicationContext,
+  ApplicationStateSchema<Events>,
+  Events
+> = {
+  type: ApplicationTypes.EXAMPLE,
```
arnifreyrs, line 28: here you have to use the `ApplicationTypes.HEALTH_INSURANCE`

dd-sdang: Great you spotted it, had missed this. Updated it 👍

darrikonn (commented):
This is missing the babel config, most likely not set up as a react lib.

libs/application/templates/health-insurance/README.md
dd-sdang: I have investigated it and it seems that this is the general application. If an Icelandic citizen never moves abroad they will get health insurance automatically and doesn't need to apply for it. But I could always make this enum specific anyway if we want to be safe. What do you think?

For more background, there will be applications in the future for insurance claim from the Health insurance for example sick pay and work accidents. But that might not be conflicting with this enum as they are more specific.

darrikonn: Good comment ^

dd-sdang:
> This is missing the babel config, most likely not set up as a react lib.

@darrikonn Oh no, i must have taken the wrong command and used the node library generation instead. How do I update this in the best way? Should I regenerate it?

darrikonn:
> > This is missing the babel config, most likely not set up as a react lib.
> 
> @darrikonn Oh no, i must have taken the wrong command and used the node library generation instead. How do I update this in the best way? Should I regenerate it?

You can just copy-paste the `tsconfig.*` + `(.?)babel*` + `.eslintrc` files from the reference template to your template.

libs/application/templates/health-insurance/README.md
arnifreyrs: great answer, I think that this enum value (HEALTH_INSURANCE) can safely be reserved for this application.

arnifreyrs (approved):

darrikonn (approved):

PR #2107, opened 2020-12-02
User profile data provider WIP

arnifreyrs:
# \<Description\>

We need the application system to have data providers which can communicate with other services in our system. One of them, is the User Profile data provider, which fetches the email and mobile number from the user profile service.

## What and why

In order to achieve this we have to use the graphql api gateway for service to service communication. Doing so requires us to propagate the access token from the graphql layer to the application system, where the dataproviders are initialized and they in turn call the graphql api again.

In this PR, there are also some minor improvements to the dev environment so it is easier to run the user profile service in parallel with other backend services. 

## Checklist:


libs/api/domains/application/src/lib/application.service.ts
```diff
@@ -103,12 +103,16 @@ export class ApplicationService {
...
+  async updateExternalData(
+    input: UpdateApplicationExternalDataInput,
+    accessToken?: string,
+  ) {
     const { id, ...populateExternalDataDto } = input
 
     return this.applicationApi.applicationControllerUpdateExternalData({
       id,
       populateExternalDataDto,
+      xAccessToken: accessToken,
```
eirikurn, line 115: Ideally, we can configure applicationApi with a preMiddleware, that can access `CurrentUser`, and add its access token to the request headers.

libs/application/core/src/types/BasicDataProvider.ts
```diff
@@ -13,13 +14,32 @@ export interface DataProvider {
...
   /**
    * Use this method to fetch data from external APIs
    * @param application: current application object which may or may not possess answers, and more information that
    * could be beneficial in the function body
    */
-  abstract provide(application: Application): Promise<unknown>
+  abstract async provide(application: Application): Promise<unknown>
+
+  protected async useGraphqlGateway(query: string): Promise<Response> {
+    return fetch(`http://localhost:4444/api/graphql`, {
```
eirikurn, line 31: Hard coded url.

libs/application/data-providers/src/providers/UserProfileProvider.ts
```diff
@@ -0,0 +1,59 @@
...
+    const query = `query GetUserProfile {
+      getUserProfile {
+        email
+        emailVerified
+        mobilePhoneNumber
+        mobilePhoneNumberVerified
+      }
+    }`
+
+    return this.useGraphqlGateway(query)
```
eirikurn, line 23: I'm not sure if I love a circular reference between the graphql API and services. For a start, it makes it very tricky to strongly type the requests since the API schemas depend on each other.

Is there an issue with configuring/allowing the application service to depend on other services directly @pshomov?

apps/application-system/api/src/app/modules/application/application.controller.ts
```diff
@@ -63,8 +64,13 @@ import { ApplicationSerializer } from './tools/application.serializer'
 import { UpdateApplicationStateDto } from './dto/updateApplicationState.dto'
 import { ApplicationResponseDto } from './dto/application.response.dto'
 import { EmailService } from '@island.is/email-service'
+
 // @UseGuards(IdsAuthGuard, ScopesGuard) TODO uncomment when IdsAuthGuard is fixes, always returns Unauthorized atm
 @ApiTags('applications')
+@ApiHeader({
+  name: 'X-AccessToken',
+  description: 'Access token',
+})
```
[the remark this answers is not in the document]
arnifreyrs: of course, I did not realize that I could "overwrite" that standard header functionality

apps/application-system/api/src/app/modules/application/application.controller.ts
```diff
@@ -229,18 +235,23 @@ export class ApplicationController {
...
   ): Promise<ApplicationResponseDto> {
     await validateIncomingExternalDataProviders(
       existingApplication as BaseApplication,
       externalDataDto,
     )
-
     const templateDataProviders = await getApplicationDataProviders(
       (existingApplication as BaseApplication).typeId,
     )
+    const headers = (req.headers as unknown) as { 'x-accesstoken'?: string }
```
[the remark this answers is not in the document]
arnifreyrs: The issue with that is pretty huge, and that is why I had to do all of these shenanigans. The issue is that the auth-api-lib is far from typescript strict, so importing it causes loads of problems. I could do it and make the application-system-api also typescript strict: false, until Fuglar have finished the typescript compliance. What do you guys think?

libs/api/domains/application/src/lib/application.service.ts
arnifreyrs: see answer above

libs/application/core/src/types/BasicDataProvider.ts
arnifreyrs: Is it alright to read environment variables inside a lib?

apps/application-system/api/src/app/modules/application/utils/externalDataUtils.ts
```diff
@@ -15,13 +15,17 @@ class NotImplemented extends BasicDataProvider {
...
+    new (accessToken?: string) => BasicDataProvider
+  >,
+  accessToken?: string,
 ): BasicDataProvider[] {
   const providers: BasicDataProvider[] = []
   externalDataDTO.dataProviders.forEach(({ type }) => {
     const Provider = templateDataProviders[type]
     if (Provider) {
-      providers.push(new Provider())
+      providers.push(new Provider(accessToken))
```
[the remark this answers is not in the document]
arnifreyrs: I think this is a very valid future improvement (implement this like NestjS providers). I will take a look at the rest

libs/application/data-providers/src/providers/UserProfileProvider.ts
pshomov: We wanted to try to avoid direct dependencies between services and rather go through the official entry point. Would be much simpler for everyone to reason about dependencies if we know the only way to access a service is through the graphql interface.

libs/application/core/src/types/BasicDataProvider.ts
pshomov: In a similar situation asked a developer to pass base url in as a param and then the app using the lib is using "Environment" resolvers for local and prod env.

apps/application-system/api/src/app/modules/application/application.controller.ts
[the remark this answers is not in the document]
arnifreyrs: I am having trouble with using `CurrentUser` inside our REST layer, some problem with Scopes and ScopesGuard. We will have to add it as future improvement

PR #2111, opened 2020-12-03
Updated version

stefanhar:
# \<Description\>

Update content search elasticsearch index template

## What

Bumping versions by one

## Why

To have the migrations update the index structures

## Checklist:


dabbeg (approved):

barabrian (approved):
lgtm

barabrian (approved):

apps/services/search-indexer/config/template-en.json
```diff
@@ -1,6 +1,6 @@
 {
   "order": 0,
-  "version": 4,
+  "version": 11,
```
haukurh, line 3: why not up to 5?
this might cause issues for the rollback, I'm guessing

stefanhar: We need to display a kibana dashboard that needs a index pattern to match a single version of indexes, this is the easiest way to achieve that.
A rollback would always need to be manual when it comes to indexes today. Errors in migrations look to be implemented so that they revert all changes on error. Meaning current index should not be affected on migration errors

haukurh: I see, then in that case 👍

PR #2113, opened 2020-12-03
Disable sentry during local dev

arnifreyrs:
# \<Description\>

## What

## Why

## Screenshots / Gifs

## Checklist:


libs/application/ui-shell/src/components/ErrorBoundary.tsx
```diff
@@ -26,12 +26,14 @@ class ErrorBoundary extends PureComponent<Props, StateTypes> {
 
   componentDidCatch(error: Error) {
     const { application, currentScreen } = this.props
-    Sentry.withScope((scope) => {
-      scope.setExtra('applicationType', application.typeId)
-      scope.setExtra('applicationState', application.state)
-      scope.setExtra('currentScreen', currentScreen.id)
-      Sentry.captureException(error)
-    })
+    if (!window.location.origin.includes('http://localhost')) {
```
jeremybarbet, line 29: Why not using NODE_ENV variable?

arnifreyrs: because this is a react web app, not server side rendered

madebynoam: Could we do it like this?:
https://create-react-app.dev/docs/adding-custom-environment-variables/

madebynoam: > There is also a built-in environment variable called NODE_ENV. You can read it from process.env.NODE_ENV. When you run npm start, it is always equal to 'development', when you run npm test it is always equal to 'test', and when you run npm run build to make a production bundle, it is always equal to 'production'. You cannot override NODE_ENV manually. This prevents developers from accidentally deploying a slow development build to production.

arnifreyrs: I see the error in my ways, will fix, thx @jeremybarbet and @madebynoam

madebynoam (approved):
lgtm!

PR #2114, opened 2020-12-03
Dpo refactoring

ssh16:
# More refactoring of the application

## Screenshots / Gifs

## Checklist:


libs/application/templates/document-provider-onboarding/src/forms/DocumentProviderApplication.ts
```diff
@@ -65,6 +68,8 @@ export const DocumentProviderOnboarding: Form = buildForm({
               id: 'applicant.phoneNumber',
               name: m.applicantPhoneNumber,
               variant: 'tel',
+              format: '###-####',
+              placeholder: '000-0000',
```
hinriksteinar, line 72: This makes it so we can only use icelandic phonenumbers. Might be some case where there is a phonenumber from different country ? Open for disscussion

libs/application/templates/document-provider-onboarding/src/forms/DocumentProviderApplication.ts
```diff
@@ -156,11 +167,101 @@ export const DocumentProviderOnboarding: Form = buildForm({
...
+              id: 'applicant.address',
+              name: m.applicantAddress,
+              width: 'half',
+            }),
+            buildTextField({
+              id: 'applicant.zipCode',
+              name: m.applicantZipCode,
+              width: 'half',
+            }),
+            buildTextField({
```
hinriksteinar, line 196: Was there a decision regarding which inputs should be half width ? I could imagine the email input can be to long for half width.

hinriksteinar (commented):
Minor comments, but looks good !

libs/application/templates/document-provider-onboarding/src/forms/DocumentProviderApplication.ts
ssh16: We decided to make all the inputs full width :)

libs/application/templates/document-provider-onboarding/src/forms/DocumentProviderApplication.ts
ssh16: True, good point. We will have to think about that later on and maybe change it

hinriksteinar (approved):
LGTM

PR #2120, opened 2020-12-04
Judicial system/update detention request table

oddsson:
# Update styles for detention request table

https://app.asana.com/0/1199153462262248/1199387306535837

## What

Update styles for detention request table and add the ability to sort table by accused name and created date. 

### NOTE. not according to design

According to the design, you should be able to sort the table by `state` and `custody end date` as well. I didn't implement that because it's hard.

### Why it's hard to sort by state

We are sorting by the values that come from the server. The states from the server are `ACCEPTED`, `REJECTED`, etc. so the order would not be correct since we display the states as `Gæsluvarðhald virkt`, `Gæsluvarðhaldi hafnað´ etc. We could sort by what's displayed on the screen when we sort by state but there is another complication. Accepted cases can show `Gæsluvarðhald virkt` and `Gæsluvarðhaldi lokið` and it's hard to determine exactly that happens.

### Why it's hard to sort by custody end date

We are sorting by the values that come from the server rather than what we show on the screen. Cases might have a custody end date but we are not showing it (we only show the custody end date on accepted case) making the order confusing. 

## Why

UX 

## Screenshots / Gifs

ezgif com-video-to-gif-3

## Checklist:


addi (dismissed):

barabrian (approved):
lgtm

gudjong (approved):

PR #2126, opened 2020-12-04
Feature/api cms rework union resolution

stefanhar:
# \<Description\>

https://app.asana.com/0/1199123945323117/1199523795314495
Union type resolution is polluting our generated types with a typename field.

## What

- Removed typename from model types
- Added a system metadata type
- Fixed minor bugs in importer
- Moved all union types to a seperate folder
- Appended union to union type names to make it more clear when using unions
- Removed newlines from imports in models

## Why

- Typename field is similar to __typename field and might cause confusion
- Typename field is only used for type resolution of cached data and should only be used internally in API
- We had no oversight over union types so moved to seperate folder to prevent duplicates

## Checklist:


libs/api/domains/cms/src/lib/search/importers/article.service.ts
```diff
@@ -31,16 +31,20 @@ export class ArticleSyncService implements CmsSyncProvider<IArticle> {
...
-            fields: { relatedArticles, ...prunedRelatedArticlesFields },
-          }) => ({
-            sys,
-            fields: prunedRelatedArticlesFields,
-          }),
-        )
+        // remove nested related articles from related articles
+        const relatedArticles = (entry.fields.relatedArticles || [])
+          .map(({ sys, fields }) => {
+            // handle if some one deletes an article without removing reference case, this will be fixed more permanently at a later time with nested resolvers
```
jeremybarbet, line 37:
```suggestion
            // handle if someone deletes an article without removing reference case, this will be fixed more permanently at a later time with nested resolvers
```

jeremybarbet (changes requested):
Much needed refactoring, looking really good! 👍 

Could you add the `*.union.ts` to the list here https://github.com/island-is/island.is/blob/main/.github/workflows/push.yml#L131 https://github.com/island-is/island.is/blob/main/.github/workflows/pullrequest.yml#L46 so it triggers schemas when this type of files are changed

darrikonn:
> Typename field is only used for type resolution of cached data and should only be used internally in API

^ are you talking about the GraphQL's `__typename` or the model's `typename` field?

pshomov (approved):
LGTM

stefanhar:
> > Typename field is only used for type resolution of cached data and should only be used internally in API
> 
> ^ are you talking about the GraphQL's `__typename` or the model's `typename` field?

The `typename` field of the models.
My concern is that if we keep the `typename` field in the models it will be generated into the types and might be used to identify types over the `__typename` field

darrikonn:
> > > Typename field is only used for type resolution of cached data and should only be used internally in API
> > 
> > 
> > ^ are you talking about the GraphQL's `__typename` or the model's `typename` field?
> 
> The `typename` field of the models.
> My concern is that if we keep the `typename` field in the models it will be generated into the types and might be used in code rather than the `__typename` field

Awesome, agree!

jeremybarbet (approved):

PR #2130, opened 2020-12-07
Filter in island-ui library

saevarma:
# [WIP]Filter in island-ui library

*This is currently contains only the desktop version, hence labeled WIP*
It would be great to get comments on how to separate the mobile version,
as the mobile version uses full screen overlay which can be found [here](https://www.figma.com/file/IQ006wy2vR1jIWJ6Uqwis5/Viskuausan?node-id=381%3A14)

## What

Filter component to be added to the island-ui library.
The design can currently be found in the [UI Library Sandbox](https://www.figma.com/file/Rb6Ga0RzPJluRZJf4i7eEv/Sandbox?node-id=1771%3A12876) and in [Viskuausan Re-design](https://www.figma.com/file/IQ006wy2vR1jIWJ6Uqwis5/Viskuausan?node-id=319%3A483)

## Why

To use the `Filter` in API Catalogue sites but also make the `Filter` part of the island-ui library for others to use.

## Checklist:


barabrian (commented):
nice! looks good so far

libs/island-ui/core/src/lib/Filter/FilterInput/FilterInput.tsx
```diff
@@ -0,0 +1,42 @@
...
+  onChange,
+}) => {
+  const { width } = useWindowSize()
+  const [isMobile, setIsMobile] = useState(false)
+  useEffect(() => {
+    if (width < theme.breakpoints.md) {
+      return setIsMobile(true)
+    }
+    setIsMobile(false)
+  }, [width])
```
barabrian, line 28: Maybe we should implement this the same way we do with FilterMultiChoice using responsive displays. Or we could possibly implement responsive background for the input component

barabrian (commented):
LGTM! only some minor comments

libs/island-ui/core/src/lib/Filter/FilterInput/FilterInput.tsx
saevarma: Yes I was actually about doing the background responsive on the input as it would save the additional responsive display markup.

barabrian (approved):
LGTM

PR #2140, opened 2020-12-08
Judicial system/user institution

gudjong:
# User Institution

https://app.asana.com/0/1199153462262248/1199240898801817

## What

☝️ 

## Why

This information is needed in PDF documents and on screen.

## Screenshots / Gifs

## Checklist:


libs/judicial-system/types/src/lib/types.ts
```diff
@@ -14,6 +14,7 @@ export interface User {
   mobileNumber: string
   email: string
   role: UserRole
+  institution?: string
```
oddsson, line 17: Would it be a good idea to use an enum for this?

gudjong: I would like to handle this and the court in a similar way. Currently, courts are stored in cases as strings, so I would like this to be a string as well. We may consider changing this later, especially if used in drop downs and such in the client.

oddsson (approved):

PR #2146, opened 2020-12-08
nr/speed up resolvers by fetching on demand

darrikonn:
# Speed up GraphQL querying for national registry

## What

Resolve individual fields on demand

## Why

Using a single resolver that makes 5 api requests synchronously is super slow.
Rather, use the power of GraphQL to make the requests on demand, i.e.
when and if the user requests it.

## Checklist:


obmagnusson (approved):
Fantastic. Good work.

PR #2149, opened 2020-12-08
Kibana migrator in search content service

dabbeg:
## What

Saving kibana objects to json files and creating migrate and sync logic.
Migrate updates the kibana objects according to the json in git repo, this was crucial so that the kibana objects use the newest version of the index available.
Sync updates the json files according to the kibana objects you have on your local running kibana instance.

## Checklist:


apps/services/search-indexer/src/migrate/index.ts
```diff
@@ -192,11 +214,29 @@ class App {
     logger.info('Elasticsearch migration completed')
     return processedMigrations
   }
+
+  private async migrateKibana() {
+    logger.info('Starting kibana migration')
+    const version = await elastic.getCurrentVersionFromConfig(locales[0])
+    await kibana.importObjects(version)
+    logger.info('Done')
+  }
```
stefanhar, line 223: I didn't notice any check to prevent this from running multiple times.
This migration script runs in init-container for `search-indexer`.
It's probably not optimal to run this on every init-container, we should look into making this run only when needed

stefanhar (changes requested):

workspace.json
```diff
@@ -4087,8 +4087,8 @@
         "migrate": {
           "builder": "@nrwl/workspace:run-commands",
           "options": {
-            "command": "node migrate.js",
-            "cwd": "dist/apps/services/search-indexer"
+            "command": "yarn ts-node -P apps/services/search-indexer/tsconfig.app.json apps/services/search-indexer/src/migrate/index.ts",
```
darrikonn, line 4090: Can skip yarn

apps/services/search-indexer/src/environments/environment.prod.ts
```diff
@@ -8,4 +8,5 @@ export const environment: Environment = {
   dictRepo: 'https://api.github.com/repos/island-is/elasticsearch-dictionaries',
   s3Folder: '',
   locales: ['is', 'en'],
+  configPath: './config',
```
darrikonn, line 11: Is there a different config file for prod?

(Since local dev references the grandparent directory)

dabbeg: Yes when developing it uses the default project structure but when the project has been built then it looks like this
```
config/
migrate.js
main.js
```

workspace.json
dabbeg: I need yarn, were you thinking of `package.json`?

apps/services/search-indexer/src/migrate/index.ts
dabbeg: I am not very concerned with running this multiple times as all of the json files are combined to a single ndjson file which is imported only once per migration. So this is one API call per pod. The import takes only a second or two.

workspace.json
darrikonn: No, thought you wouldn't need it since you run nx through yarn.. but it was nit-picky, can ignore.

apps/services/search-indexer/src/migrate/README.md
```diff
@@ -14,15 +14,34 @@ The migration file has several tasks:
...
+You can run a sync with
+
+```bash
+yarn nx run services-search-indexer:migrate --sync
+```
+
+A sync fetches data from running services and updates your local files. This
+finds the ids of the objects inside `./config/kibana` and searches for them on
+your local running Kibana instance and updates them accordingly. Remember to
+run `migrate` before running `sync`.
```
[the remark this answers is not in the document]
dabbeg: services-search-indexer runs `migrate` for multiple services, my thinking was creating `sync` that fetches data from remote services and updates local files.
But no problem, I will change the command and revise the README to only mention a sync to kibana

apps/services/search-indexer/src/migrate/index.ts
```diff
@@ -38,6 +55,12 @@ class App {
     logger.info('Done!')
   }
 
+  async sync() {
+    logger.info('Running local sync')
+
+    this.syncKibana()
+  }
```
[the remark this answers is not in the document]
dabbeg: I assume that you are talking about the sync function being to generic.
I removed that and modified the sync command to refer only to kibana.

stefanhar (approved):
LGTM

darrikonn (approved):

PR #2164, opened 2020-12-09
Be able to use external data in form conditions

arnifreyrs:
# \<Description\>

When building forms in the application system, we have to be able to add conditions to fields. Before, dynamic conditions were only passed the answers as a parameter. Now we have the external data too, and that enables us to add more functionality than before.

## Checklist:


madebynoam (approved):
Nice and good idea, lgtm!

PR #2167, opened 2020-12-09
Service Portal/Add plausible

Thorrik58:
# Add plausible to service portal

## What

Adding plausible for analytics

## Why

We need to have some idea of how many people are visiting the site and how they are using the page. Plausible helps with that with some added benefits. All events will be defined as custom goals which isnt great but we will need to look into other solutions later on.

## Screenshots / Gifs

## Checklist:


apps/service-portal/src/index.html
```diff
@@ -37,6 +37,12 @@
 
     <meta name="viewport" content="width=device-width, initial-scale=1" />
     <link rel="icon" type="image/x-icon" href="favicon.ico" />
+    <script
+      async
+      defer
+      data-domain="beta.minarsidur.island.is"
```
atligudl, line 43: Should we not be using the new domain here?

Thorrik58: Yep

libs/plausible/src/lib/BaseEvent.d.ts
```diff
@@ -0,0 +1,17 @@
+/*  This is the type declaration for all events sent to plausible.
```
atligudl, line 1: You can use typescript doc comments here.

```
/**
 * Comment
 */
```

libs/service-portal/core/src/components/UserInfoLine/UserInfoLine.tsx
```diff
@@ -9,7 +9,9 @@ import {
 } from '@island.is/island-ui/core'
 import { useLocale } from '@island.is/localization'
 import { MessageDescriptor } from 'react-intl'
-import { Link } from 'react-router-dom'
+import { Link, useLocation } from 'react-router-dom'
+import { ServicePortalOutboundLink } from '@island.is/plausible'
```
atligudl, line 13: Shouldn't the function name be lowercase?

Thorrik58: You are correct

libs/service-portal/documents/src/screens/Overview/Overview.tsx
```diff
@@ -57,55 +59,64 @@ type FilterValues = {
...
+    //   return fuse.search(searchQuery).map((elem) => {
+    //     // const fuseItem = (elem as unknown) as FuseItem
+    //     // return fuseItem.item
+    //     return elem.item
+    //   })
+    // }
+    if (searchQuery) {
+      if (!searchInteractionEventSent) {
+        if (pathname) {
+          documentsSearchDocumentsInitialized(pathname)
```
atligudl, line 108: Wouldn't this make more sense within the search event listener? `handleSearchChange`

You can also chain booleans instead of nesting if statements:
`if(searchQuery && !searchInteractionEventSent && pathname)`

libs/service-portal/settings/src/components/UserOnboardingModal/UserOnboardingModal.tsx
```diff
@@ -52,6 +58,9 @@ const UserOnboardingModal: ServicePortalModuleComponent = ({ userInfo }) => {
 
   const closeModal = () => {
     setToggleCloseModal(true)
+    if (pathname) {
```
atligudl, line 61: This will only catch events where the modal is closed on the final step. Might  belong better in `dropOnboardingSideEffects`
See implementation of `<Modal />` and `<BaseModal />`

atligudl (changes requested):

libs/service-portal/documents/src/screens/Overview/Overview.tsx
```diff
@@ -155,6 +149,10 @@ export const ServicePortalDocuments: ServicePortalModuleComponent = ({
   const handleSearchChange = useCallback((value: string) => {
     setPage(1)
     setFilterValue({ ...defaultFilterValues, searchQuery: value })
+    if (!searchInteractionEventSent && pathname) {
+      documentsSearchDocumentsInitialized(pathname)
+      searchInteractionEventSent = true
```
atligudl, line 154: Use a state variable, this will probably run on every keydown because the variable will reinitialize to false on every render
You can also move useFilteredDocuments back out of the component

darrikonn:
Was there an ADR written about the choice of Plausible?

Thorrik58:
@darrikonn nope not that I know of. I got told that we would be using plausible just like ytri síður as requested by Andri Kristinsson and Linda Lyngmo.

atligudl (dismissed):

libs/service-portal/documents/src/screens/Overview/Overview.tsx
```diff
@@ -27,6 +27,8 @@ import isEqual from 'lodash/isEqual'
 import { ValueType } from 'react-select'
 import DocumentCard from '../../components/DocumentCard/DocumentCard'
 import { defineMessage } from 'react-intl'
+import { documentsSearchDocumentsInitialized } from '@island.is/plausible'
+import { useLocation } from 'react-use'
```
darrikonn, line 31: Is there a reason for using `react-use` instead of `react-router(-dom?)`?
Since the location history is already tracked with `react-router` in the service-portal.

darrikonn (dismissed):

libs/service-portal/documents/src/screens/Overview/Overview.tsx
Thorrik58: not really, just used what was already there. Ill use the react-router instead.

Thorrik58: Fixed also removes the need for a check of pathname, good point @darrikonn

darrikonn (approved):

atligudl (approved):

sindrig (approved):
LGTM

PR #2171, opened 2020-12-10
Flip it to the left side

siggakogk:
# Flip it to the left side

Flipping navigation to the left on about page and about sub pages

 ## What

using new sidebar layout component instead of custom ones
using new navigation sidebar

## Screenshots / Gifs

image

image

Screen Recording 2020-12-11 at 09 35 19

image

## Checklist:


libs/island-ui/core/src/lib/Navigation/Navigation.treat.ts
```diff
@@ -6,7 +6,9 @@ export const divider = style({
   height: 1,
 })
 
-export const root = style({})
+export const root = style({
+  transition: 'all 150ms',
```
barabrian, line 10: We should probably be more specific here regarding targeted transitions, `all` could have unwanted side effects and some performance impacts

barabrian (commented):
LGTM, one minor comment

barabrian (approved):
LGTM

PR #2174, opened 2020-12-10
Improve routing in general for the application system. 

arnifreyrs:
Use `umsoknir` as the primary path instead of the English alternative. Try to fix the weird freeze error by changing how login is handled and protected routes rendered.

## Screenshots / Gifs
Screenshot 2020-12-10 at 11 11 47

## Checklist:


apps/application-system/form/src/components/Authenticator.tsx
```diff
@@ -0,0 +1,29 @@
...
+        .then(() => {
+          setPending(false)
+        })
+        .catch(() => {
+          setPending(false)
+        })
+    }
+  }, [pendingAuthentication, userInfo, userInfoState, signInUser])
+
+  return <>{userInfo ? children : <AuthenticationLoading />}</>
```
madebynoam, line 26: For future PR: Would be nice if we had an error screen for the catch, otherwise they would just think it's loading and stuck.

apps/application-system/form/src/routes/Application.tsx
```diff
@@ -17,10 +17,15 @@ export const Application = () => {
...
 
   if (!id) {
-    return <p>Error there is no id</p>
+    return <NotFound />
   }
   if (!nationalRegistryId) {
-    return null
+    return (
+      <NotFound
+        title="Þú þarft að vera skrá þig inn."
```
madebynoam, line 25: Next PR: Localise strings? :)

apps/application-system/form/src/routes/Applications.tsx
```diff
@@ -64,7 +65,20 @@ export const Applications: FC = () => {
     }
   }, [data])
 
-  if (createError) return <p>Error! {createError.message}</p>
+  if (applicationsError)
+    return (
+      <NotFound
+        title="Þessi gerð umsókna er ekki til"
```
madebynoam, line 71: Same here: locale strings

libs/application/ui-shell/src/lib/ApplicationForm.tsx
```diff
@@ -71,11 +72,8 @@ const ShellWrapper: FC<{
     populateForm()
   }, [fieldsDispatch, application, form, nationalRegistryId])
 
-  if (!applicationId) {
-    return <p>Error there is no id</p>
-  }
-  if (error) {
-    return <p>{error}</p>
+  if (!applicationId || error) {
+    return <NotFound />
```
madebynoam, line 76: Future PR: Would be nice to have more details for Error situations, so the user could tell someone what  error they are getting. Will need design for that type of screen.

madebynoam (approved):
Just noted refinements that we can tackle in a next task/PR. 
Otherwise lgtm!

apps/application-system/form/src/routes/Application.tsx
arnifreyrs: definitely

PR #2182, opened 2020-12-10
gql/genarate schema from code first approach

darrikonn:
# Use NestJS module to create schema from resolvers

## What

Use NestJS module to create schema from resolvers and remove buildSchema script

## Why

Should be a single source of truth, and NestJS automatically picks up the resolvers
to create its schema using code-first approach, so we can just use that to
create the schema. This enables us to use the single-source of truth for the resolvers
and removes the need of having to add every single resolver to the buildSchema function
in addition to importing the module.

## Checklist:


apps/api/src/app/app.module.ts
```diff
@@ -20,19 +20,14 @@ import { DocumentProviderModule } from '@island.is/api/domains/document-provider
...
- * So the automatically generated schemas won't be failing when running.
- */
 @Module({
   controllers: [HealthController],
   imports: [
     GraphQLModule.forRoot({
       debug,
       playground,
-      autoSchemaFile,
+      autoSchemaFile: 'apps/api/src/api.graphql',
```
pshomov, line 30: Not sure this will work in the Docker container. The account running the process does not have permissions to write to the file system

handbook/repository/schemas.md
```diff
@@ -140,35 +140,6 @@ Add the following script to the `workspace.json`'s project.
...
-  path: 'PATH/api.graphql',
-  resolvers: [...], // Define all the resolvers used by the api
-})
-```
-
-> IMPORTANT! When adding new resolvers to your modules don't forget to update this array as well.
-
-Then you can add it to the `workspace.json`
-
-```json
```
jeremybarbet, line 160: This part is still needed in the workspace

scripts/build-graphql-schema.ts
```diff
@@ -0,0 +1,17 @@
+import yargs from 'yargs'
```
jeremybarbet, line 1: I would probably keep the same name between the script and the command name inside the workspace

apps/api/src/app/app.module.ts
```diff
@@ -20,19 +20,14 @@ import { DocumentProviderModule } from '@island.is/api/domains/document-provider
 
 const debug = process.env.NODE_ENV === 'development'
 const playground = debug || process.env.GQL_PLAYGROUND_ENABLED === 'true'
-const autoSchemaFile = debug ? 'apps/api/src/api.graphql' : true
```
jeremybarbet, line 23: I believe that was there for some Docker related purpose

jeremybarbet (commented):
That's definitely a very nice improvement 👍

scripts/build-graphql-schema.ts
darrikonn: agree!

apps/api/src/app/app.module.ts
darrikonn: Yeah fixed it. Now in-memory for production builds.

darrikonn: Yuup, fixed it. Uses in-memory for production.

apps/air-discount-scheme/api/src/app/app.module.ts
```diff
@@ -10,15 +10,14 @@ import {
   FlightLegModule,
 } from './modules'
 import { BackendAPI } from '../services'
+import { environment } from '../environments'
 
 const debug = process.env.NODE_ENV === 'development'
 const playground = debug || process.env.GQL_PLAYGROUND_ENABLED === 'true'
-const autoSchemaFile = debug ? 'apps/air-discount-scheme/api.graphql' : true
+const autoSchemaFile = environment.production
```
dabbeg, line 17: Weird to have a variable that either has a boolean or a string

dabbeg (approved):
Good work!!

apps/air-discount-scheme/api/src/app/app.module.ts
darrikonn: Well, that's a decision made by the creators of `@nestjs/graphql`. I agree, it's super confusing. `true` means in-memory btw.

dabbeg: Then they are weird

pshomov (approved):
I am a bit lost in all the schema generation thing but let's see what happens ;)

jeremybarbet (approved):

essgje (approved):
lftm

scripts/build-graphql-schema.ts
```diff
@@ -0,0 +1,17 @@
...
+import { GraphQLDefinitionsFactory } from '@nestjs/graphql'
```
jeremybarbet, line 2: This import is not used

gudjong (approved):
LGTM

PR #2194, opened 2020-12-11
Fix freeze bug

arnifreyrs:
Fix application system freeze due to lazy loading of forms and templates

This happens when lazy loading forms and other files that are stored in memory after first load takes too long. 

## Checklist:


libs/application/ui-shell/src/lib/ApplicationForm.tsx
```diff
@@ -35,48 +35,70 @@ const ShellWrapper: FC<{
...
           template !== null &&
           !(isOnProduction() && !template.readyForProduction)
         ) {
-          const stateInformation = await getApplicationStateInformation(
-            application,
-          )
-          const applicationFields = await getApplicationUIFields(
-            application.typeId,
-          )
-          fieldsDispatch(applicationFields)
```
arnifreyrs, line 54: I think this is the bug actually. This changes the surrounding context, triggering a re-render of the whole tree. Then, this effect is run again because `form` is still undefined because the `formLoader` is being awaited still. This might be setting of a infinity loop.

I still might be wrong, and it is just a bug in how we lazy load templates and forms.

libs/application/ui-shell/src/lib/ApplicationForm.tsx
```diff
@@ -35,48 +35,70 @@ const ShellWrapper: FC<{
...
+  const [dataSchema, setDataSchema] = useState<Schema>()
+  const [form, setForm] = useState<Form>()
+  const [, fieldsDispatch] = useFields()
+
   useEffect(() => {
     async function populateForm() {
-      if (application !== undefined && form === undefined) {
+      console.log('popluate')
+      if (dataSchema === undefined && form === undefined) {
+        console.log('load templates')
```
madebynoam, line 65: Do you want to keep the log in?

madebynoam (approved):
lgtm.

PR #2209, opened 2020-12-14
Feature/advania ark client

stullifugl:
# \<Description\>

Adding an ark client as asked for from Ægir
Also updating a redirect-url for HMS client since I received wrong redirect url the first time

## What

New client


libs/auth-api-lib/seeders/20201210073716-island-is-add-payment-ark-client.js
```diff
@@ -0,0 +1,163 @@
...
+
+module.exports = {
+  up: (queryInterface, Sequelize) => {
+    const clients = [
+      {
+        client_id: 'island-ark-1',
+        require_client_secret: false,
+        enable_local_login: true,
+        require_pkce: true,
+        allow_offline_access: true,
```
valurefugl, line 11: do they need refresh tokens?

libs/auth-api-lib/seeders/20201210073716-island-is-add-payment-ark-client.js
```diff
@@ -0,0 +1,163 @@
...
+        client_id: 'island-ark-1',
+        grant_type: 'authorization_code',
+      },
+    ]
+
+    const resourceClaims = [
+      {
+        api_resource_name: 'api_resource',
+        claim_name: 'name',
+      },
```
valurefugl, line 81: is this the only claim they need in the payload?

libs/auth-api-lib/seeders/20201210073716-island-is-add-payment-ark-client.js
stullifugl: Refresh token implementation removed

libs/auth-api-lib/seeders/20201210073716-island-is-add-payment-ark-client.js
stullifugl: payload put in id_token, not access_token

valurefugl (approved):

PR #2218, opened 2020-12-14
island-ui/ feat: support truncating CategoryCard heading

DavidBachmann:
## Support truncating CategoryCard heading
When documenting APIs the heading text is often unbroken which breaks the layout.

## Screenshots / Gifs

image

instead of 

image

## Checklist:


barabrian (commented):
LGTM, I question if we want the text to truncate instead of breaking into multiple lines. And maybe we should add `title` attribute to the text if we truncate it?

barabrian (approved):
LGTM

PR #2228, opened 2020-12-15
Finally use map user to role properly across the application system. 

arnifreyrs:
This method is critical in making application answers and external data read/write-able across the system. I also changed the second parameter for the function so the whole application is accessible.

## Checklist:

### DIFF
--- a/apps/application-system/api/src/app/modules/application/e2e/application.spec.ts
+++ b/apps/application-system/api/src/app/modules/application/e2e/application.spec.ts
@@ -1,22 +1,36 @@
 import { setup } from '../../../../../test/setup'
 import request from 'supertest'
 import { INestApplication } from '@nestjs/common'
-
+import * as tokenUtils from '../utils/tokenUtils'
 import { environment } from '../../../../environments'
 
 let app: INestApplication
 
+const nationalId = '123456-4321'
 beforeAll(async () => {
   app = await setup()
 })
 
 describe('Application system API', () => {
+  let spy: jest.SpyInstance<
+    string | undefined,
+    [import('@nestjs/common').ExecutionContext]
+  >
+  beforeEach(() => {
+    spy = jest.spyOn(tokenUtils, 'getNationalIdFromToken')
+    spy.mockImplementation(() => {
+      return nationalId
+    })
+  })
+  afterAll(() => {
+    spy.mockRestore()
+  })
   it(`POST /application should register application`, async () => {
     // Act
     const response = await request(app.getHttpServer())
       .post('/applications')
       .send({
-        applicant: '123456-4321',
+        applicant: nationalId,
         state: 'draft',
         attachments: {},
         typeId: 'ParentalLeave',
@@ -38,7 +52,7 @@ describe('Application system API', () => {
     const failedResponse = await request(app.getHttpServer())
       .post('/applications')
       .send({
-        applicant: '123456-4321',
+        applicant: nationalId,
         state: 'draft',
         attachments: {},
         typeId: 'ExampleForm',
@@ -62,7 +76,7 @@ describe('Application system API', () => {
     const response = await server
       .post('/applications')
       .send({
-        applicant: '123456-4321',
+        applicant: nationalId,
         state: 'draft',
         attachments: {},
         typeId: 'ExampleForm',
@@ -92,7 +106,7 @@ describe('Application system API', () => {
     const response = await server
       .post('/applications')
       .send({
-        applicant: '123456-4321',
+        applicant: nationalId,
         state: 'draft',
         attachments: {},
         typeId: 'ExampleForm',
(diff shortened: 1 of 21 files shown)


madebynoam (dismissed):
The state machine changes look good to me! I like the changes to `mapUserToRole`. The other parts of the PR are out of my expertise, so best that @baering takes a look at those.

hinriksteinar (dismissed):
Looks good to me ( the parts concerning DocumentProviders, other parts not in my expertise ) :+1:

apps/application-system/api/src/app/modules/application/utils/tokenUtils.ts
```diff
@@ -0,0 +1,16 @@
...
+  ctx: ExecutionContext,
+): string | undefined {
+  const request = ctx.switchToHttp().getRequest()
+  try {
+    const decodedToken = decode(
+      request.headers.authorization?.replace('Bearer ', ''),
+    ) as { nationalId?: string }
+    return decodedToken?.nationalId
+  } catch (e) {
+    return undefined
```
baering, line 14: `@NationalId() nationalId: string | undefined`

If decodedToken did not contain a nationalId the user is not authorized. Should that not throw an error?

And should the decorator not always be written like this:
`@NationalId() nationalId: string`

So the developer is certain that if the handler in the controller runs, then the user has a valid token.

apps/application-system/api/src/app/modules/application/utils/validationUtils.ts
```diff
@@ -46,15 +46,15 @@ export async function validateIncomingAnswers(
   application: Application,
   newAnswers: FormValue | undefined,
   isStrict = true,
+  nationalId = '',
```
baering, line 49: Same arguments here, will this then not become a required parameter that is not initialised?

apps/application-system/api/src/app/modules/application/utils/tokenUtils.ts
arnifreyrs: true!

libs/api/domains/application/src/lib/application.service.ts
```diff
@@ -24,81 +24,93 @@ const handleError = (error: any) => {
...
 
-  async deleteAttachment(input: DeleteAttachmentInput) {
+  async deleteAttachment(input: DeleteAttachmentInput, authorization?: string) {
     const { id, ...deleteAttachmentDto } = input
 
     return await this.applicationApi
       .applicationControllerDeleteAttachment({
         id,
         deleteAttachmentDto,
+        authorization,
```
baering, line 113: `authorization` will become required here

libs/application/core/src/types/ApplicationTemplate.ts
```diff
@@ -25,5 +26,8 @@ export interface ApplicationTemplate<
     states: StatesConfig<TContext, TStateSchema, TEvents> // TODO Extend StatesConfig to completely enforce meta being required attribute
   }
   readonly stateMachineOptions?: Partial<MachineOptions<TContext, TEvents>>
-  mapUserToRole(id: string, state: string): ApplicationRole
+  mapUserToRole(
+    id: string,
+    application: Application,
+  ): ApplicationRole | undefined
```
baering, line 32: Wondering about the return type here. If a user has no role, shouldn't we throw an error?

libs/application/core/src/lib/ApplicationTemplateHelper.ts
```diff
@@ -166,8 +166,11 @@ export class ApplicationTemplateHelper<
     return returnValue
   }
   getWritableAnswersAndExternalData(
-    role: ApplicationRole,
+    role?: ApplicationRole,
   ): ReadWriteValues | undefined {
+    if (!role) {
```
baering, line 171: Same here with role, like with authentication - should this not just be assumed to be valid and caught elsewhere if not?

baering (approved):
Nice addition, added some ideas for minor improvements. LGTM 👍

libs/application/core/src/types/ApplicationTemplate.ts
arnifreyrs: i was wondering the same. Not sure what should happen. If there is no role, there will be no form loaded, nor any answers readable or writable any ways. I am a bit afraid of throwing an error, but maybe we can do that later ?

apps/application-system/api/src/app/modules/application/e2e/application.spec.ts
```diff
@@ -71,6 +71,26 @@ describe('Application system API', () => {
     environment.environment = envBefore
   })
 
+  it('should fail when trying to POST when not logged in', async () => {
```
baering, line 74: 💯

baering (approved):

hinriksteinar (approved):

dabbeg (approved):

PR #2254, opened 2020-12-16
Fix flaky tests

sindrig:
Using date constructs using current timestamp in mocks can lead to
flaky tests. In this case, the mock dates are evaluated when the mocks
are imported, while the dates that are comparing to them are evaluated
when the tests run. This leads to failed runs when the mocks are
imported, the clock reaches the next full minute and then the test
runs.

The solution is to either hardcode the dates (like I've done) or use
some test tool to freeze time (but that's harder here since it would
mean a change in architecture).


sindrig:
this is still flaky. will need something better. hang on.

gudjong (approved):
Sometimes, depending on the current date is necessary, but in these cases we don't need to do that.

sindrig:
> Sometimes, depending on the current date is necessary, but in these cases we don't need to do that.

Totally agree! But like I said, in these cases if you're evaluating the current date in different places, you need to have a way to freeze the time between evaluations :)

PR #2255, opened 2020-12-16
Judicial system/delete case

oddsson:
# Delete case

https://app.asana.com/0/322488613152836/1199337767669527

## What

Add the ability for prosecutors to delete drafts.

## Why

Reduces clutter on the detention request screen.

## Screenshots / Gifs

ezgif com-video-to-gif-4

## Checklist:


apps/judicial-system/web/src/routes/Judge/Confirmation/Confirmation.tsx
```diff
@@ -28,29 +28,14 @@ import {
...
-      documentSigned
-      code
-      message
-    }
-  }
-`
+import {
+  RequestSignatureMutation,
+  SignatureConfirmationQuery,
+} from '../../../utils/mutations'
```
gudjong, line 38: Not sure it is good to have all GraphQL querys/mutations is one place. They used to be all in sharedGql.ts under the graphql folder and it got a bit crowded. Best practice says the query belongs close to its used, so perhaps in a separate file along side the tex file?

But I'll let it slide for now as it is not good to have it inside the tsx file either.

apps/judicial-system/web/src/routes/Shared/DetentionRequests/DetentionRequests.tsx
```diff
@@ -157,6 +160,41 @@ export const DetentionRequests: React.FC = () => {
...
+
+        setRequestToRemoveIndex(undefined)
+
+        setTimeout(() => {
+          setCases(
+            cases?.filter((c: Case) => {
+              return c !== caseToDelete
+            }),
+          )
+        }, 800)
```
gudjong, line 189: Perhaps we should add timeouts to all queries and mutations.
But isn't a timeout of less than a second for a server request that updates the database a bit aggressive?

gudjong (commented):
A few comments

gudjong (approved):

PR #2258, opened 2020-12-16
Correcting kibana url for environments

dabbeg:
## What

Use the ELASTIC_NODE environment variable because it has the same path as the kibana api


stefanhar (approved):
LGTM

PR #2261, opened 2020-12-16
Feature/api catalogue content

saevar43:
# API Catalogue Content

## What

Changes made to how subpage header content is stored and displayed.

## Why

We want to be able to display subpage header content with or without images so we needed to create a new content type and hook it up to the web.

## Screenshots / Gifs

image

## Checklist:


libs/api/domains/cms/src/lib/models/subpageHeader.model.ts
```diff
@@ -0,0 +1,32 @@
...
+  @Field()
+  subpageId: string
+
+  @Field()
+  title: string
+
+  @Field()
+  summary: string
+
+  @Field({ nullable: true })
```
stefanhar, line 17: I'm not sure nest graphQL can resolve the type here.
You might need to add a type function to resolve the type:
```suggestion
  @Field(() => Image, { nullable: true })
```

libs/api/domains/cms/src/lib/models/subpageHeader.model.ts
```diff
@@ -0,0 +1,32 @@
...
+}
+
+export const mapSubpageHeader = ({
+  fields,
+}: ISubpageHeader): SubpageHeader => ({
+  subpageId: fields.subpageId ?? '',
+  title: fields.title ?? '',
+  summary: fields.summary ?? '',
+  featuredImage: mapImage(fields.featuredImage),
+  content: fields.content ? JSON.stringify(fields.content) : '',
```
stefanhar, line 31: Should this maybe be a slice type?
This should be a slice type if this a richText field in Contentful.
If that is the case this should be resolved like body in the article model.
Using JSON negates the type, we should avoid using JSON where able.

stefanhar (changes requested):
Will ask a frontend developer to review:
apps/web/screens/ApiCatalogue/ApiCatalogue.tsx

libs/api/domains/cms/src/lib/models/subpageHeader.model.ts
saevar43: Content is JSON in contentful , and I noticed that it was dealt with like this in model for GenericPage (misc field). 

The reason for having this as JSON was to be able to specify and reference content for multiple components, e.g. we need to have a button on this page. Maybe another page would need some other components and feed content into those which felt easier if the content was specified in JSON format. 

Could something similar be done with a slice union?

stefanhar: We have taken the approach of having components represented by content types in Contentful.
This allows us to have embeddable entries that reflect our components and maintain type safety.
This allows for great flexibility for editors as they can intertwine components and formatted text.
e.g. we have a content-type called "Process entry" that can be embedded into richText fields of articles.

This is then resolved into a slices union in API that the frontend can request. Then the RichText cmp is used to display the correct cmp with correct data.
This process can be seen by viewing Article.

This is suitable for generic pages that have dynamic content with components.

If you need just the html elements of richText I believe the Html type might be the way to go as seen implemented here:
https://github.com/island-is/island.is/pull/2236/files

stefanhar (approved):

saevarma (approved):
LGTM

PR #2272, opened 2020-12-16
island-ui / Hyphen

barabrian:
# Hyphen

## What

Add soft hyphens to icelandic text

## Screenshots / Gifs

long-text-hypen

## Checklist:


libs/island-ui/core/src/lib/Hyphen/Hyphen.tsx
```diff
@@ -0,0 +1,46 @@
...
+  minRight?: number
+  minLeft?: number
+}
+
+export const Hyphen = ({
+  children,
+  minRight = 4,
+  minLeft = 4,
+}: HyphenProps) => {
+  const text = useMemo(() => hypenateText(children, { minRight, minLeft }), [
```
DavidBachmann, line 38: 🥇

libs/island-ui/core/src/lib/Hyphen/Hyphen.tsx
```diff
@@ -0,0 +1,46 @@
+import React, { useMemo } from 'react'
+import ReactHtmlParser from 'react-html-parser'
+import Hypher from 'hypher'
+import { is } from './patterns'
```
DavidBachmann, line 4: I wonder if we can/should dynamically import this

DavidBachmann (dismissed):
I like it

libs/island-ui/core/src/lib/Hyphen/Hyphen.tsx
barabrian: That might be a good idea, I think this is sufficient while we only have one pattern but we should look into having this dynamically imported when/if we add locale support

DavidBachmann (approved):

sindrig (approved):
LGTM

PR #2275, opened 2020-12-17
Show chat button on mobile

siggakogk:
# \<Description\>

## What

## Why

## Screenshots / Gifs

## Checklist:

### DIFF
--- a/apps/web/components/ChatPanel/ChatPanel.tsx
+++ b/apps/web/components/ChatPanel/ChatPanel.tsx
@@ -15,7 +15,7 @@ declare global {
 let boost = null
 
 export const ChatPanel = () => {
-  const [visible, setVisible] = useState(false)
+  const [visible, setVisible] = useState(true)
 
   useEffect(() => {
     const conversationId =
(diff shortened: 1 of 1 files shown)


apps/web/components/ChatPanel/ChatPanel.tsx
```diff
@@ -15,7 +15,7 @@ declare global {
 let boost = null
 
 export const ChatPanel = () => {
-  const [visible, setVisible] = useState(false)
+  const [visible, setVisible] = useState(true)
```
jeremybarbet, line 18: I would probably use `react-use` and [`useWindowSize`](https://github.com/streamich/react-use/blob/master/docs/useWindowSize.md) or [`useMedia`](https://github.com/streamich/react-use/blob/master/docs/useMedia.md), so you listen to changes when the window get resized.

You would be able to do then something like:
```tsx
const isBigWindow = useMedia('min-width: 768px')
const [visible, setVisible] = useState(!isBigWindow)
```

PR #2281, opened 2020-12-17
Add email and phone number questions back.

arnifreyrs:
Add default value support to the form builder.

UserProfileProvider.ts never fails, just silently succeeds with empty object.

## Checklist:


madebynoam (approved):
Lgtm. My only note is if we want to add the information back to the final Review screen.

arnifreyrs:
> Lgtm. My only note is if we want to add the information back to the final Review screen.

Yeah definitely. The thing is we need to take a final overhaul of the review screen, I will do that later today or tomorrow

PR #2292, opened 2020-12-17
Adding labelColor to control the color for the AccordionItem

saevarma:
# Adding `labelColor` prop on `AccordionItem`

## What

Adding `labelColor` prop so the color of the label on `AccordionItem` can be controlled.

## Why

The `Filter` component is using this to change the color of the label to blue when there is 
a selected value for a specific group.

## Checklist:


DavidBachmann (approved):
👍

PR #2302, opened 2020-12-18
Catch expired sessions from useAuth hook

baering:
# \<Description\>

There was a regression where useAuth hook would think a session is valid when it has already expired

## Checklist:


arnifreyrs (approved):
nice! 

Some future work idea is to extract this into a shared lib so my pages and the application system use the same front end auth handling

arnifreyrs (approved):

PR #2313, opened 2020-12-18
nr/rewrite soap with new service

darrikonn:
# Complete rewrite of the National Registry Module

## What

A complete rewrite of the national registry module

## Why

- The previous module was developed against a wrong Soap client :(
so a full rewrite was needed.
- Additionally split up graphql into its own folder since it was confusing
working with graphql objects vs. types mapped in the services.
- Had to take out address and maritalStatus from the familyMember since the new
client doesn't return that. Needs a further refactor in the next slice if we want it.

## Checklist:


libs/service-portal/family/src/screens/FamilyMember/FamilyMember.tsx
```diff
@@ -28,9 +28,7 @@ const NationalRegistryFamilyQuery = gql`
     nationalRegistryFamily {
       fullName
       nationalId
```
obmagnusson, line 30: Is this information not available in the new Soap service?

darrikonn: Nope:
```
Had to take out address and maritalStatus from the familyMember since the new
client doesn't return that. Needs a further refactor in the next slice if we want it.
```

397b88d
"This needs to be split up into another component where a nationalRegistryUser is queried for the matching familyMember"

obmagnusson (approved):
Great work!

.envrc
```diff
@@ -2,7 +2,7 @@ source_env .env.secret
 
 # global
 
-export NODE_OPTIONS="--max-old-space-size=4096"
+export NODE_OPTIONS="--max-old-space-size=8192"
```
arnifreyrs, line 5: i like it

arnifreyrs (approved):

libs/api/domains/national-registry/src/lib/nationalRegistry.service.ts
```diff
@@ -1,32 +1,143 @@
...
+    if (this.isChild(person)) {
+      return FamilyRelation.CHILD
+    }
+    return FamilyRelation.SPOUSE
   }
 
-  getLegalResidence(houseCode: User['houseCode']): Promise<string> {
-    return this.nationalRegistryApi.getLegalResidence(houseCode)
+  private isParent(person: ISLFjolskyldan): boolean {
+    return ['1', '2'].includes(person.Kyn)
```
dabbeg, line 132: Transgender (7) should probably be in here also

.envrc
darrikonn: couldn't even lint!

dabbeg (approved):

atligudl (approved):

PR #2332, opened 2020-12-21
Skilavottord/Handle error on cancel recycling

dd-sdang:
# Skilavottord - handle error on cancel recycling

## What

Catch error if something goes wrong when user requests to cancel to recycle car.

## Why

Because errors on requesting cancellation is returned as if the mutation request passes, this fix will be able to catch if any error occurs and display it to the user.

## Checklist:


essgje (approved):
lgtm

PR #2363, opened 2020-12-23
Add custom answer validators to the application system

arnifreyrs:
Custom answer validators added to the application system to handle custom and contextual data validation.

## Why

ZOD dataschema validation is not enough.

Screenshot 2021-01-15 at 12 43 49
Screenshot 2021-01-15 at 12 43 00
Screenshot 2021-01-15 at 12 42 45
Screenshot 2021-01-15 at 12 42 36
Screenshot 2021-01-15 at 12 42 23


libs/application/templates/parental-leave/src/lib/answerValidators.ts
```diff
@@ -18,55 +22,78 @@ function buildValidationError(
 
 const FIRST_PERIOD_START = 'periods[0].startDate'
 const FIRST_PERIOD_END = 'periods[0].endDate'
+const FIRST_PERIOD_RATIO = 'periods[0].ratio'
 
+/**
+ * TODO
+ * Maybe it is enough to have a single 'periods' answer validator to handle all answers for periods?
+ * loop through all periods?
+ */
```
arnifreyrs, line 31: this is definitely something I wondered about

libs/application/templates/parental-leave/src/lib/answerValidators.ts
```diff
@@ -18,55 +22,78 @@ function buildValidationError(
...
     ) {
       return buildError('You cannot apply for a period shorter than 14 days')
     }
+
+    return undefined
+  },
+  [FIRST_PERIOD_RATIO]: (newAnswer: unknown, application: Application) => {
     return undefined
   },
-  // Maybe it is enough to have a single 'periods' answer validator to handle all answers for periods?
```
arnifreyrs: ah I did wonder about this :D

libs/application/templates/parental-leave/src/lib/answerValidators.ts
jeremybarbet: I haven't figured it all yet, I'm still going through everything to understand how everything is working together

libs/application/core/src/lib/formUtils.spec.ts
```diff
@@ -184,3 +186,126 @@ describe('extractRepeaterIndexFromField', () => {
...
+  it('should return a merge object', () => {
+    expect(
+      mergeAnswers(
+        { id: 'periodStart', value: '2020-20-20' },
+        { id: 'periodStart', value: '2021-01-01' },
+      ),
+    ).toStrictEqual({ id: 'periodStart', value: '2021-01-01' })
+  })
+
+  it('should return a merge array of object with new answers to overwrite the current answers', () => {
```
madebynoam, line 249: Just curious about one more merge case that might come up from file uploads, or from deleting a period (we had issues with it originally), perhaps we could add a test for it?:

```
 expect(
      mergeAnswers(
        [
          {
            periods: [{ startDate: '2023-12-12', ratio: 100 }],
          },
        ],
        [
          {
            periods: [], // They deleted the period
          },
        ],
      ),
    ).toStrictEqual([
      {
        periods: [],
      },
    ])
```

madebynoam (dismissed):
Great work, just one minor comment on adding a test for merge answers and lgtm!

libs/application/core/src/lib/formUtils.spec.ts
jeremybarbet: That's a good point, I will add it! thanks

libs/application/core/src/lib/ApplicationTemplateHelper.ts
```diff
@@ -187,4 +189,41 @@ export class ApplicationTemplateHelper<
...
+    if (!validators) {
+      return Promise.resolve(undefined)
+    }
+
+    let hasError = false
+    const errorMap: Record<string, string> = {}
+    const validatorPaths = Object.keys(validators)
+
+    for (let i = 0; i < validatorPaths.length; i++) {
+      const validatorPath = validatorPaths[i]
```
baering, line 207: Minor: since you're not using the index for anything else you could also use `for (const validatorPath of validatorPaths)`

libs/application/core/src/lib/formUtils.ts
```diff
@@ -17,11 +15,29 @@ import {
   SubSection,
 } from '../types/Form'
 
+const containsArray = (obj: Record<string, any>) => {
```
baering, line 18: I see a few occurrences of `Record<string, any>` in this file. Maybe extract it to a type and have it be `Record<string, unknown>`

jeremybarbet: I had to change a couple of places. The tests I did locally passed I will see if it passed with the CI now

libs/application/templates/parental-leave/src/lib/answerValidators.ts
```diff
@@ -0,0 +1,155 @@
...
+        )
+      }
+
+      // We check if the period already exists in the others periods
+      if (
+        newPeriods
+          .filter((_, index) => index !== newPeriodIndex)
+          .some(
+            (period) =>
+              period.startDate === startDate && period.endDate === endDate,
```
baering, line 128: I'm wondering if it should rather check for both `startDate` and `endDate` if either of them are within the period since I don't think you can have two periods happening simultaneously:

```javascript
const isWithinPeriod = (period: string, date: string) => {
  const periodStartEpoch = parseISO(period.startDate).getTime()
  const periodEndEpoch = parseISO(period.endDate).getTime()

  const dateEpoch = parseISO(date).getTime()

  return dateEpoch >= periodStartEpoch && dateEpoch <= periodEndDate
}

// ...
.some(period => {
  const startWithinExistingPeriod = isWithinPeriod(period, startDate)
  const endWithinExistingPeriod = isWithinPeriod(period, endDate)

  if (startWithinExistingPeriod && endWithinExistingPeriod) {
    return buildError('cannot have period within another period')
  } else if (startWithinExistingPeriod && !endWithinExistingPeriod) {
    return buildError('cannot start period within another period')
  } else if (!startWithinExistingPeriod && endWithinExistingPeriod) {
    return buildError('cannot end period within another period')
  }
)
```

libs/application/templates/parental-leave/src/parentalLeaveUtils.ts
```diff
@@ -79,7 +86,44 @@ export function formatPeriods(
...
+export const getAvailableRights = (application: Application) => {
+  const requestRights = getValueViaPath(
+    application.answers,
+    'requestRights',
+  ) as any
+  const giveRights = getValueViaPath(application.answers, 'giveRights') as any
+
+  let requestedDays = 0
+  let givenDays = 0
+  let months = defaultMonths
```
baering, line 112: Just as a reminder: once we connect to paternal leave we'll also need to check existing payment plans to see if applicant has reduced base available rights.

libs/application/ui-shell/src/utils.ts
```diff
@@ -69,3 +69,23 @@ export function extractAnswersToSubmitFromScreen(
...
+  return Boolean(message.match(/^\{.*\}/g))
+}
+
+export function parseMessage(message?: string) {
+  if (!message) {
+    return undefined
+  }
+
+  if (isJSONObject(message)) {
+    return JSON.parse(message)
```
baering, line 87: Wondering about this, from your test case above:

```javascript
it('return true if the message is an object', () => {
  expect(isJSONObject('{field:true}')).toBeTruthy()
})
```

If you were to run `JSON.parse('{field:true}')` you would get an error since it is not valid JSON.

baering (commented):
Great work Jeremy, I've only left some minor comments.

libs/application/templates/parental-leave/src/parentalLeaveUtils.ts
jeremybarbet: Yes, totally. I guess it will be part of the `externalData` so it should be easy to do. I will add a ticket for it

libs/application/ui-shell/src/utils.ts
jeremybarbet: Good catch! will update that

jeremybarbet: I don't know why I was doing this by myself when I can just use a try/catch around JSON.parse 🤦

libs/application/templates/parental-leave/src/lib/answerValidators.ts
jeremybarbet: Ahhh I didn't know that, indeed that would make sense, it's very unlikely you can create another period inside the same period. Unless we do the subtract on our side and merge both, but it sounds a bit too complicated approach and the easiest is probably check if it's within as you said. I will update this, thanks 👍

madebynoam (approved):
Nice! Lgtm.

PR #2368, opened 2020-12-28
Allow non-feature prefix feature deployment

sindrig:


sindrig:
the bad check with retag should be fixed via https://github.com/island-is/island.is/pull/2370

pshomov (approved):
LGTM

PR #2370, opened 2020-12-28
Skip failure notification if we're skipping workflow

sindrig:
Make sure we fail the workflow if we don't want to run it so the build doesn't count as a good build


pshomov (approved):
LGTM

PR #2376, opened 2020-12-29
Feature/add archived functionality

stullifugl:
## What

Adding archived functionality to client and resource models in auth-admin-api


apps/services/auth-admin-api/src/app/modules/resources/resources.controller.ts
```diff
@@ -77,7 +77,7 @@ export class ResourcesController {
   @Get('api-resources')
   @ApiQuery({ name: 'page', required: true })
   @ApiQuery({ name: 'count', required: true })
-  // TODO: Figure this out: @ApiOkResponse({  type: { rows: ApiResource[]; count: number }, isArray: true })
+  @ApiOkResponse({ type: ApiResource, isArray: true })
   async findAndCountAllApiResources(
```
unnarb, line 81: FindAndCountAll always returns { rows: array of type and count: number }
In this case: type: { rows: ApiResource[]; count: number }.
You have changed this like it returns ApiResource array. Was that the goal?

apps/services/auth-admin-api/src/app/modules/clients/clients.controller.ts
```diff
@@ -82,14 +82,14 @@ export class ClientsController {
     return await this.clientsService.update(client, id)
   }
 
-  /** Deletes a client by Id */
-  @Delete(':id')
+  /** Archive a client by Id */
+  @Put('archive/:id')
   @ApiCreatedResponse()
```
unnarb, line 87: Why not use @Delete ? If you don't want to use Delete since we aren't really deleting. Then I would use  @Patch instead of @Put.. since we aren't really doing a full update.

apps/services/auth-admin-api/src/app/modules/resources/resources.controller.ts
```diff
@@ -61,7 +61,7 @@ export class ResourcesController {
   @Get('api-scopes')
   @ApiQuery({ name: 'page', required: true })
   @ApiQuery({ name: 'count', required: true })
-  // TODO: Figure this out: @ApiOkResponse({  type: { rows: ApiScope[]; count: number }, isArray: true })
+  @ApiOkResponse({ type: ApiScope, isArray: true })
```
unnarb, line 64: Same here as below is this indented?

apps/services/auth-admin-api/src/app/modules/resources/resources.controller.ts
```diff
@@ -45,7 +45,7 @@ export class ResourcesController {
   @Get('identity-resources')
   @ApiQuery({ name: 'page', required: true })
   @ApiQuery({ name: 'count', required: true })
-  // TODO: Figure this out: @ApiOkResponse({  type: { rows: IdentityResource[]; count: number }, isArray: true })
+  @ApiOkResponse({ type: IdentityResource, isArray: true })
   async findAndCountAllIdentityResources(
     @Query('page') page: number,
```
unnarb, line 50: Same here as below is this indented?

libs/auth-api-lib/src/lib/services/clients.service.ts
```diff
@@ -81,6 +86,9 @@ export class ClientsService {
       limit: count,
       offset: offset,
       distinct: true,
+      where: {
+        archived: null,
+      },
     })
   }
```
unnarb, line 94: Should there be a function to view archived Clients or Api Resources?

libs/auth-api-lib/src/lib/services/resources.service.ts
```diff
@@ -322,15 +325,20 @@ export class ResourcesService {
...
-    return await this.apiResourceModel.destroy({ where: { name: name } })
+    const result = await this.apiResourceModel.update(
+      { archived: new Date() },
+      { where: { name: name } },
+    )
+
+    return result[0]
   }
 
   async getResourceUserClaims(name: string): Promise<any> {
```
unnarb, line 344: What about the Models and DTO's? Probably not needed in DTO's in this implementation, but isn't this field suppose to at least be in the models?

unnarb (changes requested):

libs/auth-api-lib/src/lib/services/clients.service.ts
stullifugl: Removed as requested

apps/services/auth-admin-api/src/app/modules/resources/resources.controller.ts
stullifugl: Figured out

apps/services/auth-admin-api/src/app/modules/resources/resources.controller.ts
stullifugl: Figured out

apps/services/auth-admin-api/src/app/modules/clients/clients.controller.ts
stullifugl: Changed to @Delete

libs/auth-api-lib/src/lib/services/resources.service.ts
stullifugl: Can´t figure out how to let sequelize.query return a model. Default it´s unknown. 
Set it as todo for now

libs/auth-api-lib/src/lib/services/clients.service.ts
```diff
@@ -81,6 +81,9 @@ export class ClientsService {
       limit: count,
       offset: offset,
       distinct: true,
+      where: {
+        archived: null,
+      },
     })
```
unnarb, line 87: This should be removed right?

unnarb (changes requested):

libs/auth-api-lib/src/lib/services/clients.service.ts
stullifugl: Yes, fixed

unnarb (approved):

PR #2383, opened 2021-01-04
Feature deployment harden and cleanup

sindrig:
- Deploy feature on all code changes in a pull request
- Deploy feature when label is set
- Destroy namespace when pull request is closed (pipeline remains to be implemented but should be trivial)


pshomov (dismissed):
LGTM

pshomov:
I think renaming the `push.yaml` flow will reset the build counter. Avoid that?

sindrig:
> I think renaming the `push.yaml` flow will reset the build counter. Avoid that?

Hadn't thought of that. Sure. Reverting name change.

Re-naming the workflow itself is okay though, e.g. the counter isn't reset, so I'm sticking with that as it's much more descriptive.

PR #2384, opened 2021-01-04
Judicial System: Moved court end time to the confirmation screen (#2307)

gudjong:
Co-authored-by: kodiakhq[bot] <49736102+kodiakhq[bot]@users.noreply.github.com>

# Move Court End Time

https://app.asana.com/0/1199153462262248/1199353523039760

## What

☝️ 

## Why

Minor release

## Checklist:

### DIFF
--- a/apps/judicial-system/web/src/routes/Judge/Confirmation/Confirmation.spec.tsx
+++ b/apps/judicial-system/web/src/routes/Judge/Confirmation/Confirmation.spec.tsx
@@ -1,17 +1,73 @@
 import React from 'react'
 import { render, waitFor, screen } from '@testing-library/react'
 import { Confirmation } from './Confirmation'
-import { CaseAppealDecision } from '@island.is/judicial-system/types'
+import {
+  CaseAppealDecision,
+  UpdateCase,
+} from '@island.is/judicial-system/types'
 import { MemoryRouter, Route } from 'react-router-dom'
 import * as Constants from '../../../utils/constants'
 import {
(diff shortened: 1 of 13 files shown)


apps/judicial-system/web/src/routes/Judge/Confirmation/Confirmation.spec.tsx
```diff
@@ -1,17 +1,73 @@
...
+
+    // Act
+    render(
+      <MockedProvider
+        mocks={[
+          ...mockCaseQueries,
+          ...mockJudgeQuery,
+          ...mockUpdateCaseMutation([
+            {
+              courtStartTime: '2020-09-16T15:55:000Z',
```
oddsson, line 31: Hmm, this is not correct 🤔

This should be `courtEndTime`. Could you please see if the test runs like that?

oddsson (changes requested):
One thing regarding a test.. I'll look into it

apps/judicial-system/web/src/routes/Judge/Confirmation/Confirmation.spec.tsx
gudjong: The test also succeeds with courtEndTime set and I have pushed that change. Note however that this change will not be merged to the main branch, so if it is also wrong there, then it needs to be fixed there as well.

pshomov (approved):
LGTM

oddsson (approved):

PR #2390, opened 2021-01-04
Enabling truncate flag on CategoryCard use in ServiceList

saevarma:
# Enabling truncate flag on CategoryCard 

## What

Enabling truncate flag on CategoryCard used in the ServiceList.

## Why

For some screen widths the service title can be come to long for the card so we need to truncate it.

## Screenshots / Gifs

image

## Checklist:


DavidBachmann (approved):
```
truncateHeading={true}
```
is the same as
```
truncateHeading
```

Doesn't matter to me though. Approved.

PR #2403, opened 2021-01-05
feat: add start and end date to slider

jeremybarbet:
## What

- Add start and end range to the slider dates selection

Screenshot 2021-01-05 at 13 07 36


madebynoam (approved):
Lgtm!

PR #2441, opened 2021-01-08
Only 16 and above can apply for driving licenses

darrikonn:
# Only 16 and above can apply for driving licenses

I'll move this to the backend when we implement the business logic, where we'll query if the user can apply for these licenses.


libs/service-portal/driving-license/src/screens/DrivingLicense/DrivingLicense.tsx
```diff
@@ -8,6 +8,7 @@ import { Box, GridColumn, GridRow, Stack } from '@island.is/island-ui/core'
 import { Application, Eligibility } from './components'
 
 const ELIGABLE_FOR_RENEWAL_AGE = 68
+const ELIGABLE_FOR_DRIVING_LICENSE_AGE = 16
```
sindrig, line 11: image

sindrig (dismissed):
LGTM

dabbeg (approved):

atligudl (approved):

PR #2442, opened 2021-01-08
Avoid bad check on all pull requests

sindrig:


sindrig:
Not good enough.

pshomov (approved):
LGTM

Issue #2360, opened 2020-12-23
E2E Test : 1) "before each" hook for "should display front page slider" - fails sometimes


guttih:
When pushing new code to a pull-request in the repository, the E2E test sometimes fails.

When this happens, I need to push some comment changes in order to trigger this test again which usually changes the test result to success.

The error displayed is:
    1) "before each" hook for "should display front page slider"

  0 passing (2m)
  1 failing

  1) web "before each" hook for "should display front page slider":
     CypressError: cy.visit() failed trying to load:

sindrig:
@guttih a link to a failed run would be great

jeremybarbet:
Looks like web-e2e is running against our contentful data, but the cypress test is using the mock for the test.

```
cy.get('h1').contains(cms.store.frontPageSliders[0].title)
```

I've been able to run it locally successfully with 

```
API_MOCKS=true yarn e2e web-e2e --skip-cache
```

we might want to add it to the `40_e2e.sh` script or pass the API_MOCKS=true through the workspace with an environment

guttih:
Link to a failing test: https://github.com/island-is/island.is/pull/2393/checks?check_run_id=1644920724

saevarma:
Fixed in #2446

PR #2446, opened 2021-01-08
Fix Cypress timeout

saevarma:
#Fix Cypress timeouts

Should fix Issue #2360 
## What

The solution is to run Cypress tests on production built apps.
Then Cypress is not waiting while NextJS is compiling and bundling.

## Why

Compilation of NextJS app is slow so Cypress test often fail with a timeout.

## Checklist:


scripts/ci/40_e2e.sh
```diff
@@ -1,4 +1,6 @@
 #!/bin/bash
 set -euxo pipefail
 
-yarn run e2e $1 --headless --production --record --group=$1
+HOST="http://localhost:4200"
+
+ yarn start-server-and-test "yarn start $(echo ${1%-*})" $HOST "yarn run e2e $1 --headless --production --record --group=$1 --base-url $HOST --devServerTarget '' "
```
sindrig, line 6: Pull the `echo ${1%-*}` to a variable and explain why you can't simply use `$1`.

sindrig: Maybe also pull the `yarn start ...` and `yarn run ....` strings to variables as well for readability.

sindrig: And add a newline at the end of the file :)

sindrig (changes requested):

scripts/ci/40_e2e.sh
```diff
@@ -1,4 +1,6 @@
 #!/bin/bash
 set -euxo pipefail
 
-yarn run e2e $1 --headless --production --record --group=$1
+HOST="http://localhost:4200"
```
eirikurn, line 4: Where/how can we tweak this based on the app if the ports differ?

scripts/ci/40_e2e.sh
darrikonn: Would be nice to see an example of how you run this script.

Another thing that I'm wondering...
is the api mocked?
if not, you might have to run multiple instances before running the tests. If `start-server-and-test` doesn't support that, then we already have a script that waits for ports to become active: `scripts/_wait-for.sh`

sindrig: The apiis mocked in all cypress tests as the environment variables `API_MOCKS` and `NX_API_MOCKS` are set in the github workflow. Wondering if they should be set in this file as well...? Probably?

scripts/ci/_wait-for.sh
```diff
@@ -21,21 +21,33 @@ USAGE
...
+  nc -z "$HOST" "$PORT" > /dev/null 2>&1
+
+  result=$?
+  if [ $result -eq 0 ] ; then
+    if [ $# -gt 0 ] ; then
+      exec "$@"
+    fi
+    exit 0
+  fi
+  sleep 1
```
sindrig, line 34: strange place for this sleep tbh. maybe pull it up in `wait_for` where the flow logic is?

scripts/ci/_wait-for.sh
```diff
@@ -21,21 +21,33 @@ USAGE
...
-  for i in `seq $TIMEOUT` ; do
-    nc -z "$HOST" "$PORT" > /dev/null 2>&1
+  
+  if [ $TIMEOUT -gt 0 ] ; then
+    for i in `seq $TIMEOUT` ; do
+      check_availability "$@"
+    done
+    echo "Operation timed out" >&2
+  else
+    while true ; do
```
sindrig, line 46: I don't particularly like the idea of possibly waiting for something forever in case the server doesn't come up. Can we come up with a reasonable default `TIMEOUT` instead? An hour?

    TIMEOUT=${TIMEOUT:-3600}

workspace.json
```diff
@@ -4895,6 +4973,18 @@
             }
           }
         },
+        "e2e-ci": {
+          "builder": "@nrwl/workspace:run-commands",
+          "options": {
+            "args": "--targetName=web --targetHost=localhost:4200",
+            "commands": [
+              "yarn start {args.targetName} &",
```
sindrig, line 4981: Since these 3 commands are duplicated all over this file, can we extract them to a file? e.g. `scripts/ci/e2e-ci.sh`?

sindrig: Also, if you do that, can you add a trap to kill the forked yarn start? something like

    trap "pkill -P $$" EXIT

sindrig (changes requested):

scripts/ci/_wait-for.sh
sindrig: If you do that you can get rid of the if/else and pull the function back into the business logic, so the exit logic is more readable.

handbook/repository/workspace-settings.md
```diff
@@ -0,0 +1,180 @@
...
+"static-serve": {
+  "builder": "@nrwl/workspace:run-commands",
+  "options": {
+    "commands": [
+      {
+        "command": "yarn serve -l 4200 -s dist/apps/service-portal",
+        "forwardAllArgs": false
+      }
+    ],
+    "readyWhen": "Accepting connections"
```
eirikurn, line 51: Just curious: Why is `forwardAllArgs` and `readyWhen` needed?

handbook/repository/workspace-settings.md
```diff
@@ -0,0 +1,180 @@
...
+
+  // Here you can add app specific config
+
+  return {
+    ...config,
+    node: {
+      process: true,
+      global: true,
+    },
+  }
```
eirikurn, line 104: I would just return the config, the node stuff may confuse the reader if it's not explained.

handbook/repository/workspace-settings.md
```diff
@@ -0,0 +1,180 @@
...
+    "parallel": false
+  }
+},
+```
+
+#### Configure
+
+> :information*source: `pkill -f 'serve -l'` is only needed for \_React*
+> apps as they use the `serve` npm package to serve from the production
+> build, and on failure it leaves the process dangling.
```
eirikurn, line 149: Strange that it's only left dangling on failure. 😕

scripts/ci/40_e2e.sh
```diff
@@ -1,4 +1,4 @@
 #!/bin/bash
 set -euxo pipefail
 
-yarn run e2e $1 --headless --production --record --group=$1
+yarn e2e-ci $1
```
eirikurn, line 4: Would it work to pass and forward some of these arguments from this script, so users can more easily run `e2e-ci` on their machines?

eirikurn (approved):
Awesome work. Minor comments. Let me know when you've fixed the lint and addressed Sindri's comments so I can re-review.

scripts/ci/40_e2e.sh
saevarma: No longer applicable changes.

scripts/ci/40_e2e.sh
saevarma: Moved individual project's config to workspace.json.

scripts/ci/_wait-for.sh
saevarma: Undid all changes to `_wait-for.sh`

scripts/ci/40_e2e.sh
saevarma: Yes, that would be ideal. `forwardAllArgs` was not working as I was hoping. But I will look closer at this, as I would prefer the user to be able to `yarn e2e-ci web-e2e` locally without needing to modify `workspace.json`

handbook/repository/workspace-settings.md
saevarma: When not using `forwardAllArgs` the execution hits this error:
```
ERROR: Unknown or unexpected option: --watch
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
Warning: @nrwl/run-commands command "yarn serve -l 4200 -s dist/apps/service-portal --watch=false" exited with non-zero status code
 ```
Nx seems to be adding `--watch=false` to all commands by default, which is failing the `serve` package.

`readyWhen` is needed to continue the process, otherwise `serve` blocks the process.

**Comment by saevarma** [comment]
This PR actually contains the changes from #2719. So they can safelly be merged as part of this PR and we can close #2719.

File: scripts/e2e-ci.js [hunk 10]
```diff
@@ -0,0 +1,114 @@
...
+  ],
+  TEST: `yarn nx run ${argv.name}:e2e:production --headless --production --base-url http://localhost:${argv.port}${
+    argv.ci ? 
+      ` --record --group=${argv.name}` 
+    : ''}`,
+}
+
+const main = async () => {
+  console.log(`Building ${target}...`)
+  await pexec(CMD.BUILD)
```
sindrig, line 80: await with then and catch? why?

Also probably means that an error in build won't stop the run but continue to serve/test

scripts/e2e-ci.js
```diff
@@ -0,0 +1,114 @@
...
+
+  await pexec(CMD.TEST)
+    .then((res) => console.log(res.stdout))
+    .catch((err) => {
+      console.log(err.stdout)
+      console.log(err.stderr)
+
+      if (serveChild) {
+        console.log('Clean up serve process...')
+        serveChild.kill()
```
sindrig, line 106: do you only want to kill the serve process if test failed?

sindrig (dismissed):
Minor comments but overall very good. I'm fine with moving ahead with this as is and addressing comments later :clap:

stullifugl (dismissed):

scripts/e2e-ci.js
saevarma: Yes agreed this is stupid. I'm changing this.

scripts/e2e-ci.js
saevarma: This is a left over from my previous expirements with the `serve` package, where the process was only left dangling on failed runs. But as I'm using custom static serve... I better always check if to kill this. Fixing this now.

Thorrik58 (dismissed):

apps/judicial-system/web/webpack.config.js
```diff
@@ -1,12 +1,9 @@
 // eslint-disable-next-line @typescript-eslint/no-var-requires
-const TreatPlugin = require('treat/webpack-plugin')
-// eslint-disable-next-line @typescript-eslint/no-var-requires
-const nrwlConfig = require('@nrwl/react/plugins/webpack.js')
+const nrwlConfig = require('./../../../libs/shared/webpack/nrwl-config')
```
gudjong, line 2: Shouldn't we use @island.is/... when importing?

gudjong (changes requested):
Only review the code in the judicial system.

apps/judicial-system/web/webpack.config.js
saevarma: Jubb definetely, VS Code auto import has been doing this trick on me, I will update this.

saevarma: Yes usually, but this is inside of `webpack.config.js` so it won't translate the `tsconfig.base.json` `paths` except if we would add usage of for example `tsconfig-paths-webpack-plugin`. But I will leave that for a separate discussion.

dd-sdang (approved):

oddsson (approved):

Addips (approved):
LGTM

darrikonn (approved):

Thorrik58 (approved):
LGTM

eirikurn (approved):
Love it!

dd-sdang (approved):

valurefugl (approved):

gudjong (approved):

Ingimarsson (approved):

stullifugl (approved):

barabrian (approved):
lgtm

PR #2470, opened 2021-01-11
Feature/content search remove index aliases

stefanhar:
# \<Description\>

https://app.asana.com/0/1199123945323117/1199701993113446
https://app.asana.com/0/1199123945323117/1199701993113450
https://app.asana.com/0/1199123945323117/1199701993113470

## What

- Moved index template config to a central index management library
- Removed cleanup code from migrate
- Removed rollback code from migrate
- Swapped migrate from is version higher check to does index exists check
- Made index use hash of index template files for version
- Removed dependency on index template files for version
- Removed dependency on elasticsearch alias

## Why

- Moved index templates to index management library to get hash values
- removed cleanup code to simplify migration this logic will be re-implemented on dev-ops side if required
- Removed rollback code since it should no longer be required, if index is created there is no way for it to have the wrong index template
- Checking if index exists is sufficient
- Using hash removes the human element ensuring index is always updated
- Removing aliases removes unnecessary complexity in alias management, alias also proved to be hard to keep in sync

## Checklist:


pshomov (approved):
Beautiful 👍🏻

pshomov:
Not sure what is behind those Asana links but for additional context this PR addresses the first of the two action points from this [incident report](https://www.notion.so/2020-12-16-1efd0511657041c59b78e3454ea9f894).

PR #2477, opened 2021-01-12
Adding driving license types to api domain

dabbeg:
## What

Adding driving license types to api domain

## Checklist:


libs/api/domains/driving-license/src/lib/graphql/drivingLicense.resolver.ts
```diff
@@ -27,4 +28,19 @@ export class DrivingLicenseResolver {
   drivingLicense(@CurrentUser() user: User) {
     return this.drivingLicenseService.getDrivingLicense(user.nationalId)
   }
+
+  @Query(() => [DrivingLicenseType])
```
darrikonn, line 32: Should write a separate resolver for `DrivingLicenseType`. You can see above in the decorator:
```typescript
@Resolver(() => DrivingLicense)
```
This is a resolver for `DrivingLicense`.

Thus you can not resolve individual fields inside the `DrivingLicenseType` (unless you separate it)

darrikonn (changes requested):
One comment regarding a separate GraphQL entity

libs/api/domains/driving-license/src/lib/graphql/drivingLicense.resolver.ts
dabbeg: As I understand it after reading the documentation the `@Resolver` decorator identifies this class as a resolver.

If the type is defined in the decorator like `@Resolver(DrivingLicense)` then this resolver should only resolve the DrivingLicense type as you note here above and that is why the type is required if `@ResolveField` is used .

But one could also just do `@Resolver()` without a type and then this resolver will resolve multiple types.

I want to remove the DrivingLicense type from the `@Resolver` decorator and keep this in the same resolver. If we are going to add `@ResolveField` to this resolver we should refactor and create multiple resolvers.

darrikonn (approved):

PR #2502, opened 2021-01-13
Unblock main

darrikonn:
# Fix main

## What

Fix main

## Why

I broke it (shame)

### DIFF
--- a/libs/application/core/src/types/Form.ts
+++ b/libs/application/core/src/types/Form.ts
@@ -2,6 +2,7 @@ import { ZodObject } from 'zod'
 import { Condition } from './Condition'
 import { Field } from './Fields'
 import { MessageDescriptor } from 'react-intl'
+import { BoxProps } from '@island.is/island-ui/core'
 import { Application } from './Application'
 
 export type StaticText = (MessageDescriptor & { values?: object }) | string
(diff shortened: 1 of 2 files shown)


barabrian (approved):
LGTM!

jeremybarbet:
Got also this when running the `application-system-api`

```bash
ERROR in /Users/jeremy/www/island.is/libs/application/templates/driving-license/src/forms/application.ts(144,17):
TS2532: Object is possibly 'undefined'.
ERROR in /Users/jeremy/www/island.is/libs/application/templates/driving-license/src/forms/application.ts(144,39):
TS2339: Property 'fullName' does not exist on type 'string | number | boolean | object'.
  Property 'fullName' does not exist on type 'string'.
ERROR in /Users/jeremy/www/island.is/libs/shared/form-fields/src/lib/CheckboxController/CheckboxController.tsx(76,19):
TS2322: Type 'ReactNode' is not assignable to type 'string | undefined'.
  Type 'null' is not assignable to type 'string | undefined'.
ERROR in /Users/jeremy/www/island.is/libs/shared/form-fields/src/lib/CheckboxController/CheckboxController.tsx(79,19):
TS2322: Type 'ReactNode' is not assignable to type 'string | undefined'.
  Type 'null' is not assignable to type 'string | undefined'.
Version: typescript 3.9.7
Time: 36230ms
```

barabrian (approved):

jeremybarbet:
On another note, I also believe you need to define at least an `id` for a new field in the AS. Otherwise, you won't be able to define a `dataSchema` for this specific field

darrikonn:
reverted the changes on main and will continue in another PR

PR #2508, opened 2021-01-13
Adding penalty points status to dl api domain

dabbeg:
## What

Adding penalty points to dl api domain

## Why

To be able to validate if a user is eligible to get a license

## Checklist:


libs/api/domains/driving-license/src/lib/graphql/main.resolver.ts
```diff
@@ -43,4 +44,9 @@ export class MainResolver {
   drivingLicenseRemarkTypes() {
     return this.drivingLicenseService.getRemarkTypes()
   }
+
+  @Query(() => PenaltyPointStatus)
```
darrikonn, line 48: Would like to see this as a field on driving license. Nice to make a single query:
```graphql
query {
  drivingLicense {
    nationalId
    isPenaltyPointsOk
  }
}
```

darrikonn (changes requested):

libs/api/domains/driving-license/src/lib/graphql/main.resolver.ts
darrikonn: On further notice, you don't have to have a driving license to have some points in your name.

darrikonn (approved):

PR #2509, opened 2021-01-13
Migrate admin auth web monorepo

sindrig:
# What?

Migrating https://github.com/island-is/auth-admin-web to the monorepo


darrikonn (approved):
core files look good

PR #2530, opened 2021-01-15
Web Performance improvements

eirikurn:
## What

Reduce bundle sizes in production by removing large dependencies (eg crypto-js), make large RichText sections dynamic, and properly bundle utils/components for each page with `babel-plugin-transform-imports`.

This PR gets rid of 1,2 MB from the main shared JS bundle, lowering the gzipped size of page bundles from 700KB to 300-350KB (depending on the page).

## Why

In production, most of our pages are downloading 700KB of gzipped javascript, and parsing and running multiple megabytes of JS. This hurts performance, especially on mobile phones. Consequently, it hurts our PageSpeed, which can negatively affect rankings in search engines.

## Future work:

* To avoid regression, we need to set up performance monitoring on our production website, with clear budgets and notifications when they're crossed.
* Currently, there's third-party "chatAgent" JS on all our pages which is 200KB gzipped, which we need to get rid of.
* There is definitely more stuff in our bundles worthy of follow-up discussion. 

## Screenshots / Gifs

Before:
image

After:
image

## Checklist:


libs/shared/babel/src/web.ts
```diff
@@ -0,0 +1,32 @@
...
+          '@island.is/web/components',
+          'apps/web/components/index.ts',
+        ),
+        lodash: {
+          transform: 'lodash/${member}',
+          preventFullImport: true,
+        },
+        'date-fns': {
+          transform: 'date-fns/${member}',
+          preventFullImport: true,
```
darrikonn, line 25: `lodash` and `date-fns` are already being handled by `eslint` via the `no-restricted-imports` rule

... but doesn't hurt :)

barabrian (approved):
Impressive, Well done! 🙌🏻

libs/shared/babel/src/exportFinder/exportFinder.ts
```diff
@@ -0,0 +1,228 @@
+import { parse } from '@babel/parser'
```
darrikonn, line 1: Can you give a little context on why we need these custom handlers?

darrikonn: Mainly wondering since the typescript aliases need to somehow be interpreted for jest, babel + webpack .. so nx must have been doing some magic before, right?

oddsson (approved):

jeremybarbet (approved):

DavidBachmann:
🎉

libs/shared/babel/src/web.ts
eirikurn: Ah, cool!

libs/shared/babel/src/exportFinder/exportFinder.ts
eirikurn: This is a library that parses an ES6 module and recursively finds where all its exports are actually defined.

Let's say we have a web-app which code-splits per page (like NextJS), and PageA uses a SectionA component and PageB uses a SectionB component, which are defined in our UI library. This creates the following chunks:

[](https://mermaid-js.github.io/mermaid-live-editor/#/edit/eyJjb2RlIjoiZ3JhcGggVEJcbnN1YmdyYXBoIGNodW5rIGFcblBhZ2VBXG5lbmRcbnN1YmdyYXBoIGNodW5rIGJcblBhZ2VCXG5lbmRcblBhZ2VBIC0tPiBVSVxuUGFnZUIgLS0-IFVJXG5zdWJncmFwaCBjaHVuayBzaGFyZWRcblVJW2lzbGFuZC11aS9pbmRleC50c11cblVJIC0tPiBTZWN0aW9uQVxuVUkgLS0-IFNlY3Rpb25CXG5lbmRcbiIsIm1lcm1haWQiOnsidGhlbWUiOiJkZWZhdWx0In0sInVwZGF0ZUVkaXRvciI6ZmFsc2V9)

Note that PageA gets code that is only used on PageB and vice-versa (😉), because they are imported from the index.ts file. Tree shaking only applies to code that is not used by the whole app (unfortunately).

For DX, and isolation, it's ideal to import everything from index.ts, but for optimal code splitting, it's ideal to import stuff from where it is actually defined:

[](https://mermaid-js.github.io/mermaid-live-editor/#/edit/eyJjb2RlIjoiZ3JhcGggVEJcbnN1YmdyYXBoIGNodW5rIGFcblBhZ2VBXG5QYWdlQSAtLT4gU2VjdGlvbkFcbmVuZFxuc3ViZ3JhcGggY2h1bmsgYlxuUGFnZUJcblBhZ2VCIC0tPiBTZWN0aW9uQlxuZW5kXG4iLCJtZXJtYWlkIjp7InRoZW1lIjoiZGVmYXVsdCJ9LCJ1cGRhdGVFZGl0b3IiOmZhbHNlfQ)

So we use babel-plugin-transform-imports to change the import statements, which uses this library (`exportFinder`) to recursively find the actual module that defines each export.

I'll add this explanation to the README in a separate PR (don't want to invalidate the code reviews).

darrikonn (approved):

PR #2541, opened 2021-01-15
dl/map data schema

darrikonn:
# Driving License

## What
* Fix half Grid for multi fields
* allow empty data providers (the reason why I didn't create a new field was that this is rendered as a child of dataProviders, so it still makes sense rendering this inside buildDataProviderItem, until a further refactor is needed)
* hook up data schema for driving license application

### Empty data provider (Buseta)
image

### Controlling space for multi fields
image

### Fix half grid
image

## Checklist:

### DIFF
--- a/libs/application/templates/driving-license/src/forms/application.tsx
+++ b/libs/application/templates/driving-license/src/forms/application.tsx
@@ -1,3 +1,5 @@
+import React from 'react'
+
 import {
   buildForm,
   buildDescriptionField,
@@ -16,7 +18,7 @@ import {
   Form,
   FormModes,
 } from '@island.is/application/core'
-import { NationalRegistryUser } from '@island.is/api/schema'
+import { NationalRegistryUser, UserProfile } from '@island.is/api/schema'
 import { m } from '../lib/messages'
 
 export const application: Form = buildForm({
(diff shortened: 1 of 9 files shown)


DavidBachmann (approved):

DavidBachmann (approved):

libs/application/templates/driving-license/src/forms/application.tsx
```diff
@@ -53,14 +55,15 @@ export const application: Form = buildForm({
               subTitle:
                 'Til þess að tryggja að notandi hafi heimild til þess að sækja um ökuskírteini út frá punktastöðu',
             }),
+            buildDataProviderItem({
+              id: 'residence',
+              type: undefined,
```
dabbeg, line 60: I don't fully understand why we need a untyped data provider for this. Can you elaborate?

darrikonn: We want a singular place where the user accepts some terms. These terms don't always fetch data but we still want the user to accept them, in this case búseta.

darrikonn: This needs a further refactor where we rename this field in another slice

dabbeg (approved):

PR #2556, opened 2021-01-18
ADR for feature flag service

sindrig:
# Feature flag service ADR

## What

Select a tool to use for feature flags

## Why

We want to be able to use feature flags


darrikonn (approved):
Nice

dabbeg (approved):
I like it, launch darkly is awesome!

sindrig:
FYI: We're going to do a POC on configcat to see if it would be enough for our use case as it's much cheaper

darrikonn (approved):
aiit... looking forward to see how it comes out

PR #2561, opened 2021-01-18
Set address fields nullable in national registry

dd-sdang:
# Set address fields nullable in national registry

## What

Address in NationalRegistryUser should be nullable.
PostalCode and StreetAddress in Address should also be nullable.

## Why

In the Health Insurance application we need to check if user has an Icelandic address, because Icelanders that move abroad are registered without an address and postalcode. These users should not be able to continue the health insurance application process.

## Checklist:


obmagnusson (approved):
LGTM

PR #2566, opened 2021-01-18
Domicile transfer/custom fields

VilliGunn:
# \<Description\>

https://app.asana.com/0/1199733748699731/1199584850352418

## What

- Added components, mostly with static data until we have data coming from the API
- Spacing is a mess in some places because of the application system API, needs to be discussed with the team.
- There is no 'generic' custom component, yet. If we find a use case for that we can add it.

## Why

To get better control of the layout

## Checklist:


libs/application/templates/children-domicile-transfer/src/fields/ChangeInformation/index.tsx
```diff
@@ -0,0 +1,56 @@
+import React from 'react'
+import { FieldBaseProps, getValueViaPath } from '@island.is/application/core'
+import { Box, Text } from '@island.is/island-ui/core'
+import { extractParentFromApplication } from '../../lib/utils'
+import { CheckboxController } from '@island.is/shared/form-fields'
+
+const ChangeInformation = ({ field, application, error }: FieldBaseProps) => {
+  const { id, disabled } = field
+  const parent = extractParentFromApplication(application)
+  const parentAddress = `${parent?.address}, ${parent?.postalCode} ${parent?.city}`
```
Rikkiabb, line 10: I was wondering with this line do we want to extract the formatting of the string into a helper function? 

I am guessing we are always going to show the address in this format so it could be nice to have it in one place if we decide to change the format.

libs/application/templates/children-domicile-transfer/src/forms/ChildrenDomicileTransferForm.ts
```diff
@@ -10,63 +10,12 @@ import {
...
-  return {
-    selectedChildren: application.answers.selectChild as string[],
-    selectedDuration: application.answers.selectDuration as string,
-    durationDate: application.answers.durationDate as string,
-  }
-}
+import {
+  extractParentFromApplication,
+  extractChildrenFromApplication,
+} from '../lib/utils'
```
Rikkiabb, line 19: Nice 👏

Rikkiabb (dismissed):
Looks good 🤩  Added one nitty question 😊

libs/application/templates/children-domicile-transfer/src/fields/ChangeInformation/index.tsx
VilliGunn: Done 💯

Rikkiabb (approved):
💯

PR #2567, opened 2021-01-19
Judicial system/Extend case

oddsson:
# Adds the ability to extend a case.

https://app.asana.com/0/1199153462262248/1199174045356543

## What

See ☝️

## Why

It's a requirement

## Checklist:


libs/judicial-system/formatters/src/lib/formatters.spec.ts
```diff
@@ -38,6 +38,17 @@ describe('formatDate', () => {
     expect(time).toEqual('09:36')
     expect(time2).toEqual('23:36')
   })
+
+  test('should shorten the day name if shortenDayName is set to true', () => {
```
oddsson, line 42: This test doesn't make sense. Fixing...

gudjong (approved):

PR #2570, opened 2021-01-19
feat: static AS/ui-core/ui-shell messages translated to Contentful + add markdown support + various fixes

jeremybarbet:
## What

- Move static copy from application-core and application-ui-shell to contentful
- Import @magnearun's translation extension into monorepo
- Did some refactoring to only show one field for each message inside contentful. Because there were 3 objects to handles the strings it was creating 3 times the fields for each message
- Added markdown support for translations messages

Textfield block:
Screenshot 2021-02-10 at 11 08 58

Deprecated message Textfield block:
Screenshot 2021-02-10 at 11 09 24

Markdown block:
Screenshot 2021-02-10 at 11 09 08


libs/contentful-translation-extension/README.md
```diff
@@ -0,0 +1,43 @@
...
+{% hint style="warning" %}
+To be able to run the `start` and `deploy` commands you will need to pass an `extensionId`, a `managementToken` and the `path` to the extension's folder.
+{% endhint %}
+
+### Start
+
+Will start the local development server. You are then able to go to Contentful to do your changes.
+
+```bash
+yarn cte start <extensionId> <managementToken> <path>
```
eirikurn, line 22: How about defining this as an NX workspace, with serve/build/deploy `run-commands` targets? Those targets can set the `extensionId` and `path` values, and pull the `managementToken` from the environment (documented in this readme). It should be fine to use NX, even though this is not using all of their fancy plugins logic.

I'm trying to avoid cluttering the package.json scripts when possible. Maybe we don't need `scripts/contentful-translation-extension.sh` either.

libs/contentful-translation-extension/README.md
```diff
@@ -0,0 +1,43 @@
...
+```
+
+{% hint style="warning" %}
+Be careful, once you are running this command with your extensionId, it will replaces the production's extension and only the local extension will be accessible. In short, others users trying to display the extension in their contentful dashboard will get an error message.
```
eirikurn, line 26: Good hint! :)

libs/api/domains/translations/src/lib/translations.service.ts
```diff
@@ -35,21 +41,24 @@ export class TranslationsService {
...
-    const withFallbacks = result?.items?.map(({ fields }) =>
-      mergeWith(
-        {},
-        locale.fallbackCode ? fields.strings[locale.fallbackCode] : {},
-        fields.strings[locale.code],
-        (o, s) => (isEmpty(s) ? o : s),
-      ),
+    return (
+      result?.items?.map(({ fields }) =>
+        fields.strings['is-IS'].reduce(
```
eirikurn, line 52: Guess you don't need all locales anymore? (`*` above)

libs/localization/scripts/extract.spec.ts
```diff
@@ -0,0 +1,101 @@
...
+    const local = convertToLocalArray(localValues)
+    const distant = convertToContentfulArray(contentfulValues as any)
+
+    expect(mergeArray(local, distant)).toStrictEqual([
+      {
+        id: 'application.system:no.value.in.contentful',
+        defaultMessage: 'Takk',
+        'is-IS': 'Takk',
+        en: 'Thanks',
+      },
```
eirikurn, line 40: We should store `description` in a special field and show it in the Contentful UI, since it's very useful to provide context to translators. 

Also, not sure if we should default English to the description, as it might discourage developers from writing helpful descriptions.

libs/localization/scripts/extract.ts
```diff
@@ -41,105 +40,91 @@ spawn.sync('npx', [
...
+  (namespace?.fields?.strings?.['is-IS'] ?? []).map((item: DictArray) => ({
+    id: item.id,
+    defaultMessage: item.defaultMessage,
+    'is-IS': item['is-IS'],
+    en: item.en,
+  }))
+
+export const updateNamespace = (namespace: Entry, messages: MessageDict) => {
+  const fromLocal = convertToLocalArray(messages)
+  const fromContentful = convertToContentfulArray(namespace)
```
eirikurn, line 102: Regarding the function names, maybe Objective C style `translationsFromLocal` and `translationsFromContentful` is clearer?

eirikurn (commented):
Really good to get the extension into the monorepo. Great job!

libs/api/domains/translations/src/lib/translations.service.ts
```diff
@@ -35,21 +41,24 @@ export class TranslationsService {
...
+          (acc: Record<string, string>, cur: DictArray) => ({
+            ...acc,
+            [cur.id]: isEmpty(cur[locale.code])
+              ? cur?.[locale.fallbackCode]
+              : cur?.[locale.code],
+          }),
+          {},
+        ),
+      )?.[0] ?? null
```
eirikurn, line 61: Is this only returning the first namespace requested?

libs/api/domains/translations/src/lib/translations.service.ts
```diff
@@ -35,21 +41,24 @@ export class TranslationsService {
...
-        {},
-        locale.fallbackCode ? fields.strings[locale.fallbackCode] : {},
-        fields.strings[locale.code],
-        (o, s) => (isEmpty(s) ? o : s),
-      ),
+    return (
+      result?.items?.map(({ fields }) =>
+        fields.strings['is-IS'].reduce(
+          (acc: Record<string, string>, cur: DictArray) => ({
+            ...acc,
```
eirikurn, line 54: I would suggest replacing the reduce with a simpler loop adding stuff to an object.

IMHO `reduce` is unnecessarily complicated and doing `...acc` is not well optimized in JS; for (larger) namespaces of N translations, it's creating N objects, with 1-N property assignments. Plus, this code runs a lot.

eirikurn (commented):
Some more comments.

libs/contentful-translation-extension/README.md
jeremybarbet: Great idea, I will do that!

At first, I started creating a custom schematic and definitely questioned myself if that was worth it, for now, that's why I took this shortcut instead and I didn't think about this nice alternative

libs/localization/scripts/extract.spec.ts
jeremybarbet: Yes, I've been noticing and doing it on my own as well that the `description` is mostly used to define the English translation to the Icelandic string in the `defaultMessage`.

It's just a very convenient/workaround way to define both strings at the same place at once, especially for non-native-Icelandic people to get quickly an interface they can "google translate" and switch to English for development until some proper translations are added through contentful. That's the reason I was using the `description` to populate the English field inside contentful. However, I agree it's not the real purpose of the `description` field.

I'm totally fine to stop doing that and going inside Contentful to define the English strings, but will it be the case of all developers and how can we enforce this so it's not used the wrong way? 

For the parental leave, for example, 90% of the strings were already translated inside the messages.ts, using the `description` field, but because we are not doing using it to populate the `English` fields inside contentful, all the English translation were missing once switching to the English language. I pretty much had to copy/paste everything between messages.ts and contentful to populate the English translation.

libs/api/domains/translations/src/lib/translations.service.ts
jeremybarbet: I realized that a template can use more than one namespace, I need to change that yep 👍

libs/localization/scripts/extract.ts
```diff
@@ -41,105 +40,91 @@ spawn.sync('npx', [
...
-    namespace.fields.defaults['is-IS'],
-    messages,
-  )
 
-  const newStrings = Object.keys(messages).reduce(
-    (arr, cur) => ({ ...arr, [cur]: '' }),
-    {},
-  )
+export const mergeArray = (local: DictArray[], distant: DictArray[]) =>
+  local.map((localObj) => {
```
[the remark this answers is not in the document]
jeremybarbet: I didn't add the remove functionality, but we can tackle that in another PR
 
Screenshot 2021-01-26 at 13 11 56

libs/application/ui-shell/src/hooks/useDocumentTitle.ts
```diff
@@ -0,0 +1,19 @@
+import { useEffect, useRef } from 'react'
```
jeremybarbet, line 1: I guess I could have use resolution on the package json to fix the react version. It was failing tests for I believe a mismatch of react versions

apps/judicial-system/web/src/routes/Judge/Confirmation/Confirmation.spec.tsx
```diff
@@ -60,7 +60,7 @@ describe('Confirmation route', () => {
         name: /Staðfesta og hefja undirritun/i,
       }),
     ).not.toBeDisabled()
-  })
+  }, 10000)
```
oddsson, line 63: We've set a global timeout for tests at 60 seconds, so this is not needed
```suggestion
  })
```

libs/application/core/src/lib/messages.ts
```diff
@@ -0,0 +1,127 @@
...
+  },
+  globalErrorTitle: {
+    id: 'application.system:boundary.error.title',
+    defaultMessage: 'Úps! Eitthvað fór úrskeiðis',
+    description: 'Oops! Something went wrong',
+  },
+  globalErrorMessage: {
+    id: 'application.system:boundary.error.message',
+    defaultMessage:
+      'Fyrirgefðu! eitthvað fór rosalega úrskeiðis og við erum að skoða það',
```
oddsson, line 43:
```suggestion
      'Fyrirgefðu! Eitthvað fór rosalega úrskeiðis og við erum að skoða það',
```

libs/application/core/src/lib/messages.ts
```diff
@@ -0,0 +1,127 @@
...
+    description:
+      'Sorry! something went terribly wrong and we are looking into it',
```
oddsson, line 45:
```suggestion
      'Sorry! Something went terribly wrong and we are looking into it',
```

libs/application/core/src/lib/messages.ts
```diff
@@ -0,0 +1,127 @@
...
+  thanksDescription: {
+    id: 'application.system:thanks.description',
+    defaultMessage:
+      'Úrvinnslu þinni er lokið. Umsókn er komin áfram í ferlinu.',
+    description:
+      'Your application is complete. The application has progressed in the process.',
+  },
+  notLoggedIn: {
+    id: 'application.system:not.logged.id',
+    defaultMessage: 'Þú þarft að vera skrá þig inn.',
```
oddsson, line 118: The word skrá is not correct here. It should be either skráður (male) or skráð (female/other).

libs/application/templates/parental-leave/src/forms/EmployerApproval.ts
```diff
@@ -7,23 +7,25 @@ import {
   buildSubmitField,
   Form,
   FormModes,
```
oddsson, line 9: This seems fishy 🐟

libs/application/templates/parental-leave/src/forms/InReview.ts
```diff
@@ -5,12 +5,13 @@ import {
   Form,
   FormModes,
 } from '@island.is/application/core'
+
 import Logo from '../assets/Logo'
-import { mm } from '../lib/messages'
+import { inReviewForm as m, mm } from '../lib/messages'
```
oddsson, line 10: This seems fishy 🐟

libs/application/templates/parental-leave/src/forms/OtherParentApproval.ts
```diff
@@ -6,33 +6,34 @@ import {
   buildSubmitField,
   Form,
   FormModes,
+  m as mm,
```
oddsson, line 9: This seems fishy 🐟

libs/application/templates/parental-leave/src/lib/messages.ts
```diff
@@ -437,6 +448,78 @@ export const m = defineMessages({
   },
 })
 
+export const employerForm = defineMessages({
+  formTitle: {
+    id: 'pl.application:form.title',
+    defaultMessage: 'Samþykki vinnuveitanda fyrir umsókn um foreldraorlof',
```
oddsson, line 454: I don't know if foreldraorlof is the correct word here, It's more common to use fæðingarorlof

libs/application/templates/parental-leave/src/lib/messages.ts
```diff
@@ -437,6 +448,78 @@ export const m = defineMessages({
...
+  },
+  reviewSection: {
+    id: 'pl.application:review.section',
+    defaultMessage: 'Samþykki vinnuveitanda',
+    description: 'Employer approval',
+  },
+  reviewMultiTitle: {
+    id: 'pl.application:review.multi.title',
+    defaultMessage:
+      'Starfsmaður þinn hefur sótt um foreldraorlof. Samþykkir þú valin tímabil hans?',
```
oddsson, line 465: Same as above with foreldraorlof. Also, `Samþykkir þú valin tímabil hans?` should be `Samþykkir þú valið tímabil hans?` or better yet, `Samþykkir þú tímabilið sem hann valdi?`

libs/application/templates/parental-leave/src/lib/messages.ts
```diff
@@ -437,6 +448,78 @@ export const m = defineMessages({
...
+    description:
+      'Your employee has applied for parental leave. Do you approve of his/her selected periods?',
```
oddsson, line 467: You are using his/her in english, I think you should do the same in Icelandic.

oddsson (changes requested):
I feel the copy needs some love in this PR 😉 

Would love to take another look when these issues have been resolved.

libs/application/core/src/lib/messages.ts
jeremybarbet: Not gonna lie all the translations are done by Google translate for the ones I found in English only. Would need someone better than me to review that

libs/application/templates/parental-leave/src/forms/InReview.ts
jeremybarbet: Yeah totally. The usage of the `import { m } from '.....'` is already used in every app so I didn't want to change the behavior in one app that would make the usage different.

A good alternative, in my opinion, would be to use the object directly and instead of using:

```typescript
formatMessage({
  id: 'm.my.key',
})
```

it would be:

```typescript
formatMessage({
  id: 'otherParentApprovalFormMessages.my.key',
})
```

more verbose, but clearer on what it is.

libs/application/core/src/lib/messages.ts
jeremybarbet: I'm gonna figure it out to proof read the copy :)

stefanhar (dismissed):
LGTM

oddsson (approved):

oddsson:
I'm going to be completely honest, this PR is going _way_ over my head. The changes to the judicial-system look good to me, but I didn't really understand the rest 🙈

apps/judicial-system/web/src/utils/utils.spec.tsx
```diff
@@ -378,17 +382,21 @@ describe('Step helper', () => {
       // Assert
       expect(
         await screen.findByText((_, node) => {
-          // Credit: https://www.polvara.me/posts/five-things-you-didnt-know-about-testing-library/
+          // Credit: https://www.polvara.me/posts/five-things-you-didnt-know-about-testing-library/)
```
oddsson, line 385:
```suggestion
          // Credit: https://www.polvara.me/posts/five-things-you-didnt-know-about-testing-library/
```

libs/contentful-translation-extension/src/components/Button.tsx
```diff
@@ -0,0 +1,55 @@
...
+  { id: 'format_list_bulleted', src: require('../assets/list_bulleted.png') },
+]
+
+export const Button: FC<ButtonProps> = ({ active, icon, ...props }) => {
+  const getIcon = ICONS.find((item) => item.id === icon)?.src ?? ICONS[0].src
+
+  return (
+    <button
+      {...props}
+      style={{
```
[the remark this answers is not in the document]
jeremybarbet: Yes, the extension is using `create-contentful-extension` which is using behind the scene Parcel to manage bundles. I didn't find any easy way to setup Treat using Parcel, instead of webpack with our default setup with NX, so I didn't bother trying to figure it out more and just write plain CSS objects.

This extension is out of our common dev pipeline and is not using any of NX's builders and only rely on contenful's bundle packages

libs/contentful-translation-extension/src/plugins/withShortcuts.ts
```diff
@@ -0,0 +1,112 @@
...
+/**
+ * This plugin allow markdown shortcuts to format the content
+ */
+export const withShortcuts = (editor: ReactEditor) => {
+  const { deleteBackward, insertText } = editor
+
+  editor.insertText = (text) => {
+    const { selection } = editor
+
+    if (text === ' ' && selection && Range.isCollapsed(selection)) {
```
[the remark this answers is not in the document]
jeremybarbet: It's a plugin from [Slate](https://github.com/ianstormtaylor/slate/blob/master/site/examples/markdown-shortcuts.tsx) to allow markdown shortcuts

DavidBachmann (dismissed):
LGTM

eirikurn (approved):

oddsson (approved):

PR #2572, opened 2021-01-19
Parental leave -  Review screen action buttons

madebynoam:
1. Connects the InReview action buttons - allowing the user to view or edit the application they submitted.
2. Update the style of the conclusion screen to latest design.

## Screenshots / Gifs
image

image

image

image

## Known issues (for next PRs)
1. Looks like the strings on Contentful are incorrect, we should re-test when Jeremy completes investigating translations.
2. There is still placeholder data like `const otherParentPeriods` which will eventually come from the data-provider/api
3. The Icelandic translations that I've added are need to be reviewed (came from Google Translate).

## Checklist:


libs/application/templates/parental-leave/src/fields/Review/ReadOnlyReview.tsx
```diff
@@ -0,0 +1,337 @@
...
+                  <Text>
+                    {
+                      getValueViaPath(
+                        application.answers,
+                        'usePrivatePensionFund',
+                      ) as string[]
+                    }
+                  </Text>
+                </GridColumn>
+              </GridRow>
```
barabrian, line 199: Isn't grid row and column unnecessary here? Or are we looking for consistency in the markup? Maybe I'm overlooking something :)
```suggestion
              <Box marginTop={3}>
                  <Box marginTop={1} marginBottom={2}>
                    <Text variant="h5">
                      {formatMessage(m.privatePensionFundName)}
                    </Text>
                  </Box>
                  <Text>
                    {
                      getValueViaPath(
                        application.answers,
                        'usePrivatePensionFund',
                      ) as string[]
                    }
                  </Text>
              </Box>
```

libs/application/templates/parental-leave/src/fields/Review/ReadOnlyReview.tsx
```diff
@@ -0,0 +1,337 @@
...
+                    {
+                      getValueViaPath(
+                        application.answers,
+                        'employer.email',
+                      ) as string[]
+                    }
+                  </Text>
+                </GridColumn>
+              </GridRow>
+            </Box>
```
barabrian, line 248: Same as the last comment regarding unnecessary markup, this applies to all instances where we have a grid row with a single column with a span of 1 
```suggestion
            <Box paddingY={4}>
                  <Text>
                    {
                      getValueViaPath(
                        application.answers,
                        'employer.email',
                      ) as string[]
                    }
                  </Text>
            </Box>
```

libs/application/templates/parental-leave/src/fields/Review/ReadOnlyReview.tsx
```diff
@@ -0,0 +1,337 @@
...
+            startExpanded={allItemsExpanded}
+          >
+            <Box paddingY={4}>
+              <GridRow>
+                <GridColumn span="12/12">
+                  <Box marginTop={1} marginBottom={2}>
+                    <Text variant="h5">
+                      {formatMessage(mm.shareInformation.title)}
+                    </Text>
+                  </Box>
```
barabrian, line 318: If you're looking for less markup then you could use the margin props on the `Text` component
```suggestion
                    <Text variant="h5" marginTop={1} marginBottom={2}>
                      {formatMessage(mm.shareInformation.title)}
                    </Text>
```

barabrian (commented):
Looks good! just some minor comments regarding unnecessary markup

libs/application/templates/parental-leave/src/fields/InReviewSteps/index.tsx
```diff
@@ -33,63 +56,124 @@ const statesMap: statesMap = {
...
+                      id: application.id,
+                      event: 'EDIT',
+                      answers: application.answers,
+                    },
+                  },
+                })
+
+                if (res?.data) {
+                  // Takes them back to the editable Review screen
+                  window.location.reload()
```
jeremybarbet, line 134: I wonder if somehow, this could be using the `moveToScreen` util we have inside the ui-shell?

libs/application/templates/parental-leave/src/fields/Review/ReadOnlyReview.tsx
```diff
@@ -0,0 +1,337 @@
+import React, { FC, useState } from 'react'
```
jeremybarbet, line 1: I wonder if we could share some structure between the Review and ReadOnlyReview. If we change the order of one field inside one of the two screens, we would need to update the other, and we might forget we would lead to missing fields. 

The common ground I see between the two, is the getValueViaPath where the id of each field is defined. We could probably create an object where we list the object/keys we need and then we render easy in review or read only mode. Another solution, would be to create a common component for both use cath and use the state machine to determine if we are in review or edit mode. Might be the easier. It would end up something like:

```
{edit ? (
<Input
  id="otherParentName"
  name="otherParentName"
  label={formatMessage(m.otherParentName)}
  ref={register}
/>
) : (
<Text>
  {
    getValueViaPath(
      application.answers,
      'otherParentName',
    ) as string[]
  }
</Text>
)}
```

libs/application/templates/parental-leave/src/lib/ParentalLeaveTemplate.ts
```diff
@@ -217,16 +229,15 @@ const ParentalLeaveTemplate: ApplicationTemplate<
                 import('../forms/InReview').then((val) =>
                   Promise.resolve(val.InReview),
                 ),
-              read: {
-                answers: ['spread', 'periods'],
-                externalData: ['pregnancyStatus', 'parentalLeaves'],
-              },
+              read: 'all',
+              write: 'all',
```
jeremybarbet, line 233: I'm still unfamiliar with the state machine, but does it mean that the employer can write on all his own answers or write on the applicant answers?

libs/application/templates/parental-leave/src/fields/InReviewSteps/index.tsx
```diff
@@ -33,63 +56,124 @@ const statesMap: statesMap = {
...
-      </Box>
+      {(screenState === 'steps' && (
+        <Box marginTop={7} marginBottom={8}>
+          <ReviewSection
+            application={application}
+            index={1}
+            state={statesMap['otherParent'][application.state]}
+            title={formatMessage(mm.reviewScreen.otherParentTitle)}
+            description={formatMessage(mm.reviewScreen.otherParentDesc)}
+          />
```
baering, line 156: Just as a reminder, at some point this section will need to be conditionally displayed. Or maybe I'm not familiar enough with the business logic, does the other parent always need to approve?

libs/application/templates/parental-leave/src/lib/ParentalLeaveTemplate.ts
baering: In this case it would give the applicant (or whatever role defined in that scope) the rights to read and write all answers in the application.

libs/application/templates/parental-leave/src/forms/ParentalLeaveForm.ts
```diff
@@ -731,10 +730,10 @@ export const ParentalLeaveForm: Form = buildForm({
             }),
           ],
         }),
-        buildDescriptionField({
+        buildCustomField({
           id: 'thankYou',
           title: mm.finalScreen.title,
-          description: mm.finalScreen.description,
+          component: 'Conclusion',
```
baering, line 736: It would be nice to refactor out "magic" constants, custom components could be referenced through an enum defined in constants.ts

```typescript
// constants.ts
enum CustomComponents {
  Conclusion = 'Conclusion',
}

// ParentalLeaveForm.ts
buildCustomField({
  // ...
  component: CustomComponents.Conclusion
})
```

libs/application/templates/parental-leave/src/fields/InReviewSteps/index.tsx
madebynoam: Good point, curious how we could make it work:
When we submit the update to make the state machine go back to draft it will load a different form: From the `InReview` form to `ParentalLeaveForm`, so I'm not sure if we have access to those screens during this event.

Will confirm with Baering about the state machine change.

libs/application/templates/parental-leave/src/fields/Review/ReadOnlyReview.tsx
madebynoam: Thanks for these tips! Yeah was being consistent but I see all the unnecessary mark-up now. Should be cleaned up in the latest.

libs/application/templates/parental-leave/src/fields/Review/ReadOnlyReview.tsx
madebynoam: Agree, should be cleaned up in the latest as well.

libs/application/templates/parental-leave/src/fields/Review/ReadOnlyReview.tsx
madebynoam: Same here, now refactored.

libs/application/templates/parental-leave/src/fields/Review/ReadOnlyReview.tsx
madebynoam: Great point, I was wondering how to solve it.
The latest commit has it merged into one `Review` file with the `editable` prop deciding which part to show.

libs/application/templates/parental-leave/src/fields/InReviewSteps/index.tsx
madebynoam: I ended up going with a `refetch` approach, which should cause the re-render and avoid full page reload. Good idea to avoid full page reload.

libs/application/templates/parental-leave/src/fields/InReviewSteps/index.tsx
madebynoam: Good catch, should be conditional now.

libs/application/templates/parental-leave/src/lib/ParentalLeaveTemplate.ts
madebynoam: It should work as the employer only being able to read a certain part, and submit or reject. 
On line 216 we set the permission for the `Roles.ASSIGNEE` which would be the employer: 
`read: { answers: ['periods'], externalData: ['pregnancyStatus'] },` 

Then for `Roles.APPLICANT` we allow them to write/submit (so that they can set it back to 'draft').

@baering is that correct?

libs/application/templates/parental-leave/src/forms/ParentalLeaveForm.ts
madebynoam: Agree on this, I'll tackle in separate PR.

libs/application/templates/parental-leave/src/lib/messages.ts
```diff
@@ -920,17 +959,29 @@ export const mm: MessageDir = {
...
+    step3: {
+      id: 'pl.application:finalscreen.step3',
       defaultMessage:
-        'Hitt foreldrið þarf að samþykkja umsókn þína ef þú óskaðir eftir yfirfærslu á hluta réttinda þess til þín, eða nýtingu á persónuafslætti þess. Síðan þarf vinnuveitandinn þinn að samþykkja tilhögun þína til fæðingarorlofs. Að lokum fer umsóknin á borð Fæðingarorlofssjóðs þar sem lokaúrvinnsla hennar fer fram.',
+        'Að lokum fer umsóknin á borð Fæðingarorlofssjóðs þar sem lokaúrvinnsla hennar fer fram.',
       description:
-        'The other parent will need to approve your request to use their shared month (if you did so). Then, ' +
-        'your employer will approve your parental leave dates.' +
-        'And finally Vinnumálastofnun will review your application.',
+        'Finally, the application goes to the Maternity / Paternity Leave Fund, where its final processing takes place.',
```
jeremybarbet, line 984:
```suggestion
        'Finally, the application goes to the Parental Leave Fund, where its final processing takes place.',
```

libs/application/templates/parental-leave/src/fields/Review/index.tsx
```diff
@@ -325,24 +421,25 @@ const Review: FC<FieldBaseProps> = ({
...
+                periods={formatPeriods(
+                  application.answers.periods as Period[],
+                  otherParentPeriods,
+                )}
+              />
+              {editable && (
+                <Box paddingTop={3}>
+                  <Button
+                    size="small"
+                    onClick={() => goToScreen && goToScreen('periods')}
```
jeremybarbet, line 437: You can use TS optional chaining for function as well
```suggestion
                    onClick={() => goToScreen?.('periods')}
```

libs/application/templates/parental-leave/src/fields/InReviewSteps/index.tsx
```diff
@@ -32,64 +61,140 @@ const statesMap: statesMap = {
...
+                      id: application.id,
+                      event: 'EDIT',
+                      answers: application.answers,
+                    },
+                  },
+                })
+
+                if (res?.data) {
+                  // Takes them back to the editable Review screen
+                  refetch && refetch()
```
jeremybarbet, line 166: Ditto

libs/application/ui-shell/src/lib/ApplicationForm.tsx
```diff
@@ -24,12 +26,17 @@ const ApplicationLoader: FC<{
...
     variables: {
       input: {
         id: applicationId,
       },
     },
+    // Setting this so that refetch causes a re-render
+    // https://github.com/apollographql/react-apollo/issues/321#issuecomment-599087392
+    // We want to refetch after setting the application back to 'draft', so that
+    // it loads the correct form for the 'draft' state.
+    notifyOnNetworkStatusChange: true,
```
jeremybarbet, line 39: When using `notifyOnNetworkStatusChange`, you will also get a change to the loading boolean going from false, to true, to false again. How it's done at the moment, we gonna render a blank page until the loading pass to false again. Maybe we could use this [loading state](https://ui.devland.is/?path=/docs/navigation-loadingicon--template) instead.

libs/application/ui-shell/src/lib/ApplicationForm.tsx
```diff
@@ -42,11 +49,18 @@ const ApplicationLoader: FC<{
...
-      application={application}
-      nationalRegistryId={nationalRegistryId}
-    />
+    <RefetchProvider
+      value={() => {
+        refetch()
+      }}
+    >
+      <ShellWrapper
+        application={application}
```
jeremybarbet, line 60: Since we have access to the application object, I'm wondering if a state field in the object would make more sense. Then we can listen if the state is in draft/review and refetch accordingly when needed.

Using a provider in this case looks a bit off in my opinion, but I might be wrong

libs/application/templates/parental-leave/src/fields/Review/index.tsx
madebynoam: Oh nice! Will update

libs/application/ui-shell/src/lib/ApplicationForm.tsx
madebynoam: Good catch, updated it to use the loading icon

jeremybarbet (approved):

libs/application/ui-shell/src/lib/ApplicationForm.tsx
madebynoam: Will leave this as is for now, until we can figure out a better approach. 
Related: https://github.com/apollographql/apollo-client/issues/3633
We run into the same (non ui-update issue) when trying to use `refetchQueries`

PR #2582, opened 2021-01-20
dl/split up health declaration to use field id

darrikonn:
# Driving License application

## What

Split up health declaration to use field id

## Why

The application system relies heavily on the field id for its form, thus you cannot create
a custom component with multiple form fields, since it needs to be mapped to a single field.id

## Checklist:


libs/application/templates/driving-license/src/lib/template.ts
```diff
@@ -19,30 +19,19 @@ const dataSchema = z.object({
...
-    usesContactGlasses: z.boolean(),
-    hasEpilepsy: z.boolean(),
-    hasHeartDisease: z.boolean(),
-    hasMentalIllness: z.boolean(),
-    usesMedicalDrugs: z.boolean(),
-    isAlcoholic: z.boolean(),
-    hasDiabetes: z.boolean(),
-    isDisabled: z.boolean(),
-    hasOtherDiseases: z.boolean(),
+    usesContactGlasses: z.enum(['yes', 'no']),
```
dabbeg, line 23: Would be nice to create a consts file for this

dabbeg (approved):

libs/application/templates/driving-license/src/lib/template.ts
darrikonn: agree ... the application needs a cleanup PR (this includes the translation mappings)

libs/application/core/src/lib/fieldBuilders.ts
```diff
@@ -391,7 +391,7 @@ export function buildDividerField(data: {
 
 export function buildKeyValueField(data: {
   label: React.ReactNode
-  value: MaybeWithApplication<React.ReactNode>
+  value: MaybeWithApplication<React.ReactNode | React.ReactNode[]>
```
eirikurn, line 394: Why is this ReactNode, if it is passed to formatText which expects FormText and returns strings?

libs/application/ui-fields/src/lib/KeyValueFormField.tsx
```diff
@@ -13,13 +13,16 @@ const KeyValueFormField: FC<{
...
     <Box>
       <Text variant="h4">{field.label}</Text>
-      <Text>
-        {formatText(field.value as string, application, formatMessage)}
-      </Text>
+      {Array.isArray(values) ? (
+        values.map((value) => <Text>{value}</Text>)
+      ) : (
+        <Text>{values}</Text>
+      )}
```
eirikurn, line 25: Still getting back into this project.

Is KeyValueFormField a kind of read-only field? Is it (visually?) better to wrap each value in a Text than use newlines and some handy css?

eirikurn (commented):
Quick Qs

libs/application/core/src/lib/fieldBuilders.ts
darrikonn: Yup, this should be a string. Good spot!

libs/application/ui-fields/src/lib/KeyValueFormField.tsx
darrikonn: Yeah, KeyValueFormField is a read-only field.
I originally went with the newline, but raised concerns since these values can come from CMS; and therefore changed it to rendering multiple `<Text>`

eirikurn (approved):
Still unsure about the use case and API design of having KeyValueField accept an array of value. But we can revisit that later. Happy to get this merged in.

PR #2590, opened 2021-01-20
Fix web regressions

eirikurn:
## What

This PR does 3 things:

* Fixes global styles in production build.
* Fixes an import that broke recent chunk optimizations.
* Prevent bad relative imports in the future by making `apps/web/components/index.ts` throw an error in dev.

## Why

Importing stuff from large index.ts modules that re-export other stuff is great for dev experience and isolation (to hide internals), but really bad for optimal webpack chunking. So we create aliases for these modules (like `@island.is/web/components`), which in production builds are transformed into deep imports. 

But if someone accidentally imports the index.ts file using `../components` syntax, it breaks the optimization. This PR lets the user know that they shouldn't do this.

## Checklist:


oddsson (approved):

apps/web/components/index.ts
```diff
@@ -1,50 +1,9 @@
...
-export * from './InstitutionPanel/InstitutionPanel'
+// Required ES6 syntax for TypeScript isolatedModules.
+export * from './real'
+
+// This module should never be imported for performance reasons. Instead
+// import components from `@island.is/web/components`. Read
+// `libs/shared/babel/README.md` for more details.
+throw new Error(
+  'Do not import web components from components/index.ts. Use `@island.is/web/components` instead.',
+)
```
darrikonn, line 9: Did you only add the `reel` file to be able to throw the error on relative imports?

eirikurn: Yeah. I couldn't find another way (read: an eslint rule) to block relative imports of this module. Though it might be easy to write from scratch.

darrikonn (approved):

apps/web/components/index.ts
darrikonn: Yeah, an eslint rule would be optimal :) ... I can see someone importing straight from real 🙈

eirikurn: Yeah. I made sure that index.ts exports all the same stuff.

That makes most IDE's suggest/autocomplete either "@island.is/web/components" (good!) or "../components" (bad but caught immediately and hopefully fixed correctly). Third IDE suggestion is "../components/real" >_<

Thought about naming the file DO_NOT_IMPORT.ts 😝

Oh well, might be fun refactor week task to write a custom eslint rule.

darrikonn: haha `DO_NOT_IMPORT.ts` is actually pretty good!

DavidBachmann (approved):

apps/web/components/index.ts
jeremybarbet: Doesn't this work if you use [`no-restricted-imports`](https://github.com/island-is/island.is/blob/main/.eslintrc#L46)  and pass the relative path to `../components/*`?

PR #2614, opened 2021-01-21
Use the dynamic sticky for left sidebar

stjanilofts:
# Description

Reusing the Sticky component for article and category sidebars. Also fixed where it did not update correctly on resize.

## What

Detects if the screen is not tall enough for the whole sidebar to be visible and if so it will remove position = sticky.

## Why

On smaller screens (or lower resolutions) the bottom half of the sidebar (if it was a tall one) was not visible until you scrolled all the way down.

## Checklist:


siggakogk (approved):
LGTM

PR #2618, opened 2021-01-22
Custom nextjs server

eirikurn:
## What

Create `infra-next-server` lib and integrate as custom server in all next.js app.

This is the first stab at this problem. It only works in production and it's a bit boilerplate'y. But it uses only official nx builders (more future-proof?).

It includes most of the same stuff as `infra-nest-server`, but it additionally monkey patches console.* to use winston, since NextJS internals and most React code uses console directly.

Let me know what you think.

## Why

To get better logs, tracing and basic nodejs metrics.

## Checklist:


libs/infra-metrics/src/lib/startMetricServer.ts
```diff
@@ -1,9 +1,11 @@
 import express from 'express'
-import { register } from 'prom-client'
+import { collectDefaultMetrics, register } from "prom-client";
```
darrikonn, line 2: needs formatting ;)

darrikonn (commented):
Like it!

However, my main concern is that this only works on production (according to the PR's description). Why is that?

Looking at the `serve` target, it has the `build` target as its `buildTarget`; which in turn runs the `build-server` target.
Not sure how `nx` interprets the `:production` suffix .. will only the `:production` commands be run with the `production` flag set, or does it only affect the file replacements and other configurations?

eirikurn:
> However, my main concern is that this only works on production (according to the PR's description). Why is that?
> Looking at the `serve` target, it has the `build` target as its `buildTarget`; which in turn runs the `build-server` target.

Yeah, that's not ideal. The buildTarget is a mistake, it "needs" to point to a `@nrwl/next:build` target to pick up some configurations. Thing is the "next:server" target only supports a simple js custom server.

However, I've recently gotten some tips from @sebastienne on how we can support the custom server locally as well. I'll update this PR soon.

scripts/ci/Dockerfile
```diff
@@ -61,7 +61,7 @@ COPY --from=builder /build/${APP_HOME}/next-modules/ /webapp/next-modules/
 COPY --from=builder /build/${APP_HOME}/public /webapp/public
 COPY --from=builder /build/${APP_DIST_HOME}/.next /webapp/.next
 
-ENTRYPOINT [ "npx", "next", "start", "-p", "4200" ]
+ENTRYPOINT [ "node", "main.js" ]
```
sindrig, line 64: it would feel better to me if we would keep being explicit with the port in the Dockerfile

eirikurn: IMO it's more appropriate to configure the port with environment variables. All of our server projects (Next, Nest and Express) support this using the `PORT` variable. At least they should be consistent.

We could add an `ENV PORT=4200` line to this container?

eirikurn:
fyi @darrikonn I've updated the PR with a custom server that works both locally and in docker.

scripts/ci/Dockerfile
sindrig: Ah I didn't know that it could be configured via environment. Yes, please add it as a default env to the docker image 👍

valurefugl (approved):

darrikonn (dismissed):
Nice, good work!

darrikonn (dismissed):

darrikonn (approved):

oddsson (approved):
Changes to island-ui things look good to me, the other changes look alien to me so I can't effectively review them 👽

PR #2621, opened 2021-01-22
Input fields disabled background color

tthornvik:
# Disabled input fields no longer colored blue 

## What

Making it so that the default input component no longer turns blue when disabled.
You now have to set the background color manually when you want a blue background.

## Why

In the design system the disabled input fields background color is dependent on the background color of the page. So they are blue when disabled on a page with a blue background and vice versa. See attached images

## Screenshots / Gifs

Screenshot 2021-01-22 at 13 53 11
Screenshot 2021-01-22 at 13 52 19

## Checklist:


DavidBachmann (commented):
Looks good. Just two points:
1. `mixins.containerDisabled` and `mixins.inputDisabled` should be removed from the stylesheet if they aren't being used elsewhere.
2. If you think this introduces a visual regression to other projects using this component (now incorrectly) we need to broadcast that to the other teams.

DavidBachmann (approved):
LGTM

PR #2650, opened 2021-01-25
fix: add try/catch arround spawn

jeremybarbet:
## What

- Try/catch to get error messages


jeremybarbet:
I'm closing this, I'm adding to https://github.com/island-is/island.is/pull/2570 at the same time

PR #2663, opened 2021-01-26
infrastructure/secret-service - bash wrapper

busla:
# AWS cli wrapper script for creating secrets 

Part of infrastructure repository

## What
A wrapper script around `aws-cli` to create secrets in the parameters store.

## Why

- Kubernetes consumes secrets
- Secrets should be kept in a secure location / aws parameters store (ssm)
- Secrets should be created in a uniform fashion
- Secrets creation should be audited
- Developers should have a simple and uniform way to create secrets
## Checklist:


pshomov (dismissed):
LGTM

busla:
Not sure if we can close issues in a different project, but lets give it a try:

closes https://github.com/island-is/infrastructure/issues/441

handbook/technical-overview/devops/dockerizing.md
```diff
@@ -20,6 +20,20 @@ That's it.
...
+
+You can run the `create-secret` script and you will be prompted for input.
+
+```bash
+./scripts/create-secret
+```
+
+You will be prompted for a _secret name_ that will be added to the `/k8s/` secrets namespace and a _secret value_. The length of the _secret name_ should be from 6-16 characters long. Only alphanumeric characters and dash (-) is allowed.
+
+Make sure your credentials are set in your shell before you create a secret.
```
pshomov, line 35: Make sure your AWS credentials are set up in your shell before you create a secret.

handbook/technical-overview/devops/dockerizing.md
```diff
@@ -20,6 +20,20 @@ That's it.
...
+
+Yes you can!
+
+You can run the `create-secret` script and you will be prompted for input.
+
+```bash
+./scripts/create-secret
+```
+
+You will be prompted for a _secret name_ that will be added to the `/k8s/` secrets namespace and a _secret value_. The length of the _secret name_ should be from 6-16 characters long. Only alphanumeric characters and dash (-) is allowed.
```
pshomov, line 33: One can create secrets like `/k8s/part1/part2/partN`, right?

pshomov: I think 16 is way too short for a secret name, maybe put no limit on it ;)

busla: good catch!

No they can't, I will need to fix the regex to allow `/` and maybe I should increase the max allowed chars

busla:
Since we do not control the length of the secret-value it might be risky to set a minimum length, but I think we should set a bare minimum.

busla:
I need to add one more validation to the secret-value, I'll let you know when you can review.

busla:
phew.. done.

pshomov (dismissed):
LGTM

pshomov (dismissed):
LGTM

darrikonn (commented):
Nice work!

Can you also hook it up in package.json?

handbook/technical-overview/devops/dockerizing.md
busla: secret names ending with `/` are now also rejected.

busla:
> Nice work!
> 
> Can you also hook it up in package.json?

will do, thanks for pointing that out!

package.json
```diff
@@ -36,7 +36,8 @@
     "openapi-generator": "openapi-generator generate -g typescript-fetch --additional-properties=typescriptThreePlus=true",
     "schemas": "node ./scripts/schemas.js",
     "postinstall": "node ./scripts/postinstall.js",
-    "gitbook": "node ./scripts/gitbook.js"
+    "gitbook": "node ./scripts/gitbook.js",
+    "secret-service": "./scripts/create-secret.sh"
```
darrikonn, line 40: We might want to rename these script tags, since we also have
`"env-secrets": "./scripts/secrets.sh",`.

That one fetches development secrets from the param store and loads them to env.

so `yarn (fetch/get/load)-secrets` and `yarn create-secret` is probably more descriptive.

busla: makes sense, fixing.

handbook/technical-overview/devops/dockerizing.md
busla: missed the second comment, fixing now.

darrikonn (approved):
LGTM

PR #2669, opened 2021-01-26
Major NextJs link cleanup

siggakogk:
# Remove `as`from next links

As of Next.js 9.5.3 the `as` prop is optional.
Cleaning up and simplifying next links 

> `as` - Optional decorator for the path that will be shown in the browser URL bar. Before Next.js 9.5.3 this was used for dynamic routes, check our previous docs to see how it worked

## Checklist:


apps/web/screens/ApiCatalogue/ApiCatalogue.tsx
```diff
@@ -244,7 +244,7 @@ const ApiCatalogue: Screen<ApiCatalogueProps> = ({
               items={navigationItems}
               title={nn('linkThrounText')}
               titleLink={{
-                href: linkResolver('developerspage').as,
+                href: linkResolver('developerspage').href,
               }}
```
stefanhar, line 248: The wrapping object is technically not needed here, we might also write this like: ```titleLink={ linkResolver('developerspage')}```

apps/web/components/AboutLatestNews/AboutLatestNews.tsx
```diff
@@ -123,7 +120,7 @@ const BigNewsItem = ({
         {news.title}
       </Text>
       <Text variant="intro">{news.intro}</Text>
-      <Link as={as} href={href}>
+      <Link href={href} passHref>
```
stefanhar, line 123: We should probably apply `passHref` in island-UI to all links

apps/web/components/AboutLatestNews/AboutLatestNews.tsx
```diff
@@ -61,7 +61,7 @@ export const AboutLatestNews: FC<LatestNewsProps> = ({
         {first && (
           <BigNewsItem
             news={first}
-            {...linkResolver('news', [first.slug])}
+            href={linkResolver('news', [first.slug]).href}
```
stefanhar, line 64: We might also write all instances where the prop is `href` like this
```suggestion
            {...linkResolver('news', [first.slug])}
```

stefanhar (changes requested):

apps/web/screens/ApiCatalogue/ApiCatalogue.tsx
siggakogk: this object is unrelated to next.js so i don't want to be spreading the linkresolver object here, just cherrypicking the href string

stjanilofts (approved):

stefanhar (approved):
Awesome

PR #2676, opened 2021-01-26
Feature flag draft

sindrig:
# Feature flag draft

## What

A wrapping library for configcat (configcat.com)

## Why

We want to try configcat


libs/feature-flags/src/lib/types.ts
```diff
@@ -0,0 +1,16 @@
+export interface User {
+  uuid: string
```
eirikurn, line 2: Why uuid? Not sure if we have user uuids.

libs/feature-flags/src/lib/configcat.ts
```diff
@@ -0,0 +1,28 @@
...
+
+  constructor(config: FeatureFlagClientProps) {
+    const resolvedSdkKey = config.sdkKey ?? process.env.CONFIGCAT_SDK_KEY
+    if (!resolvedSdkKey) {
+      throw new Error(
+        'Trying to initialize configcat client without CONFIGCAT_SDK_KEY environment variable',
+      )
+    }
+    if (typeof window === 'undefined') {
+      this.configcat = require('configcat-node').createClient(resolvedSdkKey)
```
eirikurn, line 15: I noticed [an option called dataGovernance](https://configcat.com/docs/sdk-reference/js/#createclientwithautopollsdkkey-options), is this something we should set?

eirikurn (approved):
Nice. Would be awesome to create a NestJS module and example for this. Though I'm not sure if it's ideal to have Nest stuff in libs that browsers bundle. Maybe another lib.

libs/feature-flags/src/lib/types.ts
sindrig: I was under the impression we had some kind of GUID for every logged-in user?

But this could just a well be session id for unauthorized users. The name here is merely to emphasize that this should be a unique id.

libs/feature-flags/src/lib/configcat.ts
sindrig: Unless we'll be storing PII in configcat (e.g. a list of SSNs, full names, e-mails) I don't think we need it. Might be up for debate though.

PR #2678, opened 2021-01-26
Api Catalogue change Service model

saevarma:
# Api Catalogue change Service model

## What

Refactor Service model to have environment owning versions
as it fits better for the UI.

## Why

To provide more appropriate data structure for the UI.

## Checklist:


libs/api-catalogue/services/src/lib/restmetadata.service.ts
```diff
@@ -50,8 +53,12 @@ export class RestMetadataService {
         data: [],
         access: [AccessCategory.XROAD],
         type: [TypeCategory.REST],
-        environments: [Environment.DEV],
-        versions: [],
+        environments: [
+          {
+            environment: Environment.DEVELOPMENT, // TODO: Needs to be environment aware
```
stefanhar, line 58: I guess this will become environment aware at a later time?

libs/api-catalogue/services/src/lib/restmetadata.service.ts
```diff
@@ -71,26 +78,23 @@ export class RestMetadataService {
...
+            description: spec.info.description ?? '',
+            type: TypeCategory.REST,
+            data: spec.info['x-category'] ?? [],
+            pricing: spec.info['x-pricing'] ?? [],
+            links: {
+              responsibleParty: spec.info['x-links']?.responsibleParty ?? '',
+              bugReport: spec.info['x-links']?.bugReport ?? '',
+              documentation: spec.info['x-links']?.documentation ?? '',
+              featureRequest: spec.info['x-links']?.featureRequest ?? '',
+            },
```
stefanhar, line 96: I guess this will become environment aware at a later time?

libs/api-catalogue/services/src/lib/restmetadata.service.ts
saevarma: Yes, Guðjón is working on that feature in the XRoad Collector.
So I was just making necessary changes so it wouldn't break.

libs/api-catalogue/services/src/lib/restmetadata.service.ts
saevarma: Yes, Guðjón is working on that feature in the XRoad Collector.
So I was just making necessary changes so it wouldn't break.

stefanhar (approved):

guttih (approved):

PR #2694, opened 2021-01-27
Add Sýslumenn page

Ingimarsson:
# Add Sýslumenn page

## What

This is an implementation of the first organization page `/stofnanir/syslumenn` and organization subpages.

**Note:** Work in progress, includes broken links, only for demoing on beta.

## Why

This is a part of migrating https://syslumenn.is to https://island.is

## Screenshots / Gifs

image

## Checklist:


stefanhar (changes requested):
I think we need a review on this from @siggakogk or @barabrian

apps/web/components/Organization/OfficeCard/OfficeCard.tsx
```diff
@@ -0,0 +1,41 @@
...
+        boxShadow="subtle"
+        border="standard"
+        borderColor="blue200"
+        marginBottom={[1, 1, 1]}
+        marginTop={[1, 1, 1]}
+        paddingX={[3, 3, 3]}
+        paddingY={[3, 3, 3]}
+      >
+        <h1 className={styles.title}>{name}</h1>
+        <p style={{ fontWeight: 300, marginTop: 14 }}>{address}</p>
```
barabrian, line 33: We should try to avoid inline styles and use treat for styling if necessary. It looks like this could be solved using the default variant from `Text`. 
```suggestion
        <Text marginTop={1}>{address}</Text>
```
Regarding the spacing we should try to limit spacings to the defined spacing variables, currently they are mostly a multiple of 8, but I can see you are using the paragraph spacings defined in figma, so maybe that's something we should add to the defined list of spacings.

barabrian: I've added paragraph spacing to the theme so when https://github.com/island-is/island.is/pull/2705 will be merged this could be
```suggestion
        <Text marginTop='p3'>{address}</Text>
```

apps/web/components/Organization/OfficeCard/OfficeCard.tsx
```diff
@@ -0,0 +1,41 @@
...
+        <div style={{ paddingTop: 18, fontWeight: 'bold' }}>Opnunartími</div>
```
barabrian, line 34: Same regarding `Text` component
```suggestion
        <Text as='div' fontWeight='semiBold' paddingTop='p5'>Opnunartími</Text>
```
We should avoid using font weights that are not defined in our theme. in this case using "bold" which is equivalent to using 700 would result in the browser faux bolding the text because we do not have the defined font loaded for that weight

apps/web/components/Organization/OfficeCard/OfficeCard.tsx
```diff
@@ -0,0 +1,41 @@
...
+        {openingHours.split('\n').map((line) => (
+          <p style={{ fontWeight: 300, marginTop: 14 }}>{line}</p>
```
barabrian, line 36: Same regarding `Text` component, this should apply for all typography use cases so I won't comment on further instances

apps/web/components/Organization/Slice/Districts/DistrictsSlice.treat.ts
```diff
@@ -0,0 +1,35 @@
...
+    },
+  }),
+})
+
+export const districtsList = style({
+  columnGap: 30,
+  ...themeUtils.responsiveStyle({
+    xs: {
+      columnCount: 1,
+      marginTop: 40,
```
barabrian, line 21: could we use theme.spacing here for consistency?
```suggestion
      marginTop: theme.spacing[5],
```
same regarding the following margins, and try to follow the multiple of 8 like ex 75 could be rounded to 72 and use theme.spacing[9]

apps/web/components/Organization/Slice/Districts/DistrictsSlice.tsx
```diff
@@ -0,0 +1,58 @@
...
+  slice: Districts
+  organization?: Organization
+}
+
+export const DistrictsSlice: React.FC<SliceProps> = ({
+  organization,
+  slice,
+}) => {
+  return (
+    <>
```
barabrian, line 23: unnecessary fragment?

apps/web/components/Organization/Slice/Districts/DistrictsSlice.tsx
```diff
@@ -0,0 +1,58 @@
...
+            <GridRow>
+              <GridColumn span={['12/12', '12/12', '7/12']}>
+                <ul className={styles.districtsList}>
+                  {slice.links.map((link) => (
+                    <li className={styles.districtsListItem}>
+                      <Link href={link.url}>
+                        <a>
+                          <Button variant="text">{link.text}</Button>
+                        </a>
+                      </Link>
```
barabrian, line 44: Just wondering, do we need the anchor here?
```suggestion
                      <Link href={link.url}>
                        <Button variant="text">{link.text}</Button>
                      </Link>
```

barabrian (changes requested):
I would like to see the current comments resolved before finishing my review

apps/web/components/Organization/Slice/Districts/DistrictsSlice.tsx
Ingimarsson: We need the anchor so that the browser recognizes them as links (the `Link` component automatically passes the href to it's children). Although this is probably not the best method. 

There seem to be quite a few non-anchor links on island.is, shouldn't that be fixed? For example they can't be right clicked and opened in a new tab.

siggakogk: `Link` renders an anchor for you, whether it's wrapped in Next link or not so this anchor is unnecessary, it means that it will be rendered like so:
```
<a>
 <a>
  <Button></Button>
 </a>
</a>
```

barabrian (commented):
There are some recurring comments that apply to more cases, I would like to see the following comments resolved before continuing the review :)

libs/api/domains/cms/src/lib/models/districts.model.ts
```diff
@@ -0,0 +1,36 @@
...
+export const mapDistricts = ({
+  sys,
+  fields,
+}: IDistricts): SystemMetadata<Districts> => ({
+  typename: 'Districts',
+  id: sys.id,
+  title: fields.title,
+  description: fields.description,
+  image: mapImage(fields.image),
+  links: fields.links.map(mapLink),
```
stefanhar, line 35: We should provide defaults for all non optional fields to support preview since it can have empty fields

libs/api/domains/cms/src/lib/models/featuredArticles.model.ts
```diff
@@ -0,0 +1,57 @@
...
+export const mapFeaturedArticles = ({
+  sys,
+  fields,
+}: IFeaturedArticles): SystemMetadata<FeaturedArticles> => ({
+  typename: 'FeaturedArticles',
+  id: sys.id,
+  title: fields.title,
+  image: mapImage(fields.image),
+  articles: fields.articles.map(mapFeaturedArticle),
+})
```
stefanhar, line 57: It might be confusing to have two models in the same file

libs/api/domains/cms/src/lib/models/featuredArticles.model.ts
```diff
@@ -0,0 +1,57 @@
...
+
+export const mapFeaturedArticle = ({
+  sys,
+  fields,
+}: IArticle): FeaturedArticle => ({
+  id: sys.id,
+  title: fields.title ?? '',
+  slug: fields.slug ?? '',
+  isApplication: fields.processEntry ? true : false,
+})
```
stefanhar, line 31: This model might not be needed since we can get this data via the `Article` model.
Using the `Article` model might also be preferable to ensure you have access to all article fields in articles under `FeaturedArticles`

libs/api/domains/cms/src/lib/models/featuredArticles.model.ts
```diff
@@ -0,0 +1,57 @@
...
+
+export const mapFeaturedArticles = ({
+  sys,
+  fields,
+}: IFeaturedArticles): SystemMetadata<FeaturedArticles> => ({
+  typename: 'FeaturedArticles',
+  id: sys.id,
+  title: fields.title,
+  image: mapImage(fields.image),
+  articles: fields.articles.map(mapFeaturedArticle),
```
stefanhar, line 56: We should check if  `fields.articles` is empty and provide a default value else `mapFeaturedArticle` might throw an error

libs/api/domains/cms/src/lib/models/featuredArticles.model.ts
```diff
@@ -0,0 +1,57 @@
...
+  articles: Array<FeaturedArticle>
+}
+
+export const mapFeaturedArticles = ({
+  sys,
+  fields,
+}: IFeaturedArticles): SystemMetadata<FeaturedArticles> => ({
+  typename: 'FeaturedArticles',
+  id: sys.id,
+  title: fields.title,
```
stefanhar, line 54: We should provide default values for non optional fields to support preview

libs/api/domains/cms/src/lib/models/footerItem.model.ts
```diff
@@ -0,0 +1,22 @@
...
+
+  @Field(() => Link, { nullable: true })
+  link: Link
+
+  @Field({ nullable: true })
+  content?: string
+}
+
+export const mapFooterItem = ({ fields }: IFooterItem): FooterItem => ({
+  title: fields.title,
```
stefanhar, line 19: We should provide default values for non optional fields to support preview

libs/api/domains/cms/src/lib/models/linkGroup.model.ts
```diff
@@ -0,0 +1,23 @@
...
+  @Field(() => Link)
+  primaryLink: Link
+
+  @Field(() => [Link])
+  childrenLinks?: Array<Link>
+}
+
+export const mapLinkGroup = ({ fields }: ILinkGroup): LinkGroup => ({
+  name: fields.name ?? '',
+  primaryLink: mapLink(fields.primaryLink),
```
stefanhar, line 21: `mapLink` will throw an error on undefined so we should check if `fields.primaryLink` exists before mapping

libs/api/domains/cms/src/lib/models/districts.model.ts
stefanhar, line 35: We have a way to resolve internal and external links with references using a model called `ReferenceLink`. Our goal is to consolidate all link logic into this model in future to allow for dynamic link resolution for all pages types.
If you think this fits your case you can change link to a reference field an pass it's value into `mapReferenceLink` you can use the menu link for reference the model is called `MenuLink` and the content type exists [here](https://app.contentful.com/spaces/8k0h54kbe6bj/content_types/menuLink/fields) in Contentful.
This might apply to all links in the following models

libs/api/domains/cms/src/lib/models/offices.model.ts
```diff
@@ -0,0 +1,31 @@
...
+  offices?: Array<OrganizationOffice>
+}
+
+export const mapOffices = ({
+  sys,
+  fields,
+}: IOffices): SystemMetadata<Offices> => ({
+  typename: 'Offices',
+  id: sys.id,
+  title: fields.title,
```
stefanhar, line 29: We should provide default values for non optional fields to support preview

libs/api/domains/cms/src/lib/models/organization.model.ts
```diff
@@ -12,32 +12,31 @@ export class Organization {
...
+  link?: string
 }
 
 export const mapOrganization = ({
   fields,
   sys,
 }: IOrganization): Organization => ({
   id: sys.id,
-  title: fields.title ?? '',
+  title: fields.title,
```
stefanhar, line 36: We should provide default values for non optional fields to support preview

libs/api/domains/cms/src/lib/models/organizationNews.model.ts
```diff
@@ -0,0 +1,46 @@
...
+  fields,
+}: IOrganizationNews): OrganizationNews => ({
+  id: sys.id,
+  organization: mapOrganization(fields.organization),
+  title: fields.title,
+  date: fields.date ?? '',
+  introduction: fields.introduction,
+  featuredImage: mapImage(fields.featuredImage),
+  content: (fields.content && JSON.stringify(fields.content)) ?? null,
+  slug: fields.slug,
```
stefanhar, line 45: We should provide default values for non optional fields to support preview

libs/api/domains/cms/src/lib/models/organizationNews.model.ts
```diff
@@ -0,0 +1,46 @@
...
+})
```
stefanhar, line 46: We should maybe try and extend the default news model to support organizations. It would allow us to consolidate the news logic and maybe reuse some components/functionality

libs/api/domains/cms/src/lib/models/organizationPage.model.ts
```diff
@@ -0,0 +1,53 @@
...
+}
+
+export const mapOrganizationPage = ({
+  sys,
+  fields,
+}: IOrganizationPage): OrganizationPage => ({
+  id: sys.id,
+  title: fields.title,
+  slug: fields.slug,
+  description: fields.description,
```
stefanhar, line 47: We should provide default values for non optional fields to support preview

libs/api/domains/cms/src/lib/models/organizationSubpage.model.ts
```diff
@@ -0,0 +1,56 @@
...
+  organizationPage: OrganizationPage
+
+  @Field({ nullable: true })
+  featuredImage?: Image
+}
+
+export const mapOrganizationSubpage = ({
+  fields,
+}: IOrganizationSubpage): OrganizationSubpage => ({
+  title: fields.title,
```
stefanhar, line 46: We should provide default values for non optional fields to support preview

libs/api/domains/cms/src/lib/models/organizationSubpage.model.ts
```diff
@@ -0,0 +1,56 @@
...
+  slug: fields.slug,
+  description: fields.description ?? '',
+  links: (fields.links ?? []).map(mapLink),
+  sidebarCards: (fields.sidebarCards ?? []).map(mapStaffCard),
+  slices: (fields.slices ?? []).map(safelyMapSliceUnion),
+  menuItem: fields.menuItem?.fields,
+  parentSubpage: fields.parentSubpage?.fields.slug,
+  organizationPage: mapOrganizationPage(fields.organizationPage),
```
stefanhar, line 54: We should skip map here if `fields.organizationPage` is empty

libs/api/domains/cms/src/lib/models/organizationNews.model.ts
```diff
@@ -0,0 +1,46 @@
...
+  sys,
+  fields,
+}: IOrganizationNews): OrganizationNews => ({
+  id: sys.id,
+  organization: mapOrganization(fields.organization),
+  title: fields.title,
+  date: fields.date ?? '',
+  introduction: fields.introduction,
+  featuredImage: mapImage(fields.featuredImage),
+  content: (fields.content && JSON.stringify(fields.content)) ?? null,
```
stefanhar, line 44: We should try and map all objects to types, using JSON removes type safety between api and web for this field.
This might be already solved by using a richtext field and `SliceUnion`

stefanhar (changes requested):

libs/api/domains/cms/src/lib/models/featuredArticles.model.ts
Ingimarsson: That's what we tried first but it won't work as the depth limit in the Contentful client is set to 4. I'm not sure if it's a good idea to increase it.

apps/web/components/Organization/Wrapper/SidebarWrapper.tsx
```diff
@@ -0,0 +1,68 @@
...
+  fullWidthContent = false,
+  children,
+}) => {
+  const isMobile = useWindowSize().width < theme.breakpoints.md
+
+  return (
+    <GridContainer position="none">
+      <Box
+        display="flex"
+        flexDirection={isMobile ? 'column' : 'row'}
```
siggakogk, line 31:
```suggestion
        flexDirection={['column', 'column', 'row']}
```

libs/api/domains/cms/src/lib/models/featuredArticles.model.ts
stefanhar: I see, you can bypass this issue by using nested resolvers.
An example of this can be found in the `relatedArticles` field of the `Article` model, with the resolver being in `cms.resolver.ts`.

apps/web/screens/Syslumenn/Home.treat.ts
```diff
@@ -0,0 +1,18 @@
+import { style } from 'treat'
+import { themeUtils } from '@island.is/island-ui/theme'
+
+export const intro = style({
```
siggakogk, line 4: redundant, should use intro variant from the Text component

apps/web/screens/Syslumenn/Home.tsx
```diff
@@ -0,0 +1,164 @@
...
+      navigationData={{
+        title: n('navigationTitle', 'Efnisyfirlit'),
+        items: navList,
+        titleLink: {
+          href: linkResolver('organizationpage', [organizationPage.slug]).as,
+          active: false,
+        },
+      }}
+      mainContent={
+        <Box className={styles.intro}>{organizationPage.description}</Box>
```
siggakogk, line 76:
```suggestion
        <Text variant={intro}>{organizationPage.description}</Text>
```

apps/web/screens/Syslumenn/SubPage.treat.ts
```diff
@@ -0,0 +1,49 @@
+import { globalStyle, style } from 'treat'
+import { themeUtils } from '@island.is/island-ui/theme'
+
+export const intro = style({
```
siggakogk, line 4: you should use the `Text` component in relevant variants

apps/web/screens/Syslumenn/SubPage.tsx
```diff
@@ -0,0 +1,217 @@
...
+            {subpage.links.length > 0 && (
+              <GridColumn
+                span={['12/12', '12/12', '4/12']}
+                offset={[null, null, '1/12']}
+              >
+                <ul>
+                  {subpage.links.map((link) => (
+                    <li>
+                      <Link href={link.url}>
+                        <a className={styles.link} href={link.url}>
```
siggakogk, line 132: the `Link` component will render an anchor for you, so no need for the duplication here, you should use the `Text` component for styling the link

siggakogk (changes requested):
needs a bit more cleaning up

libs/api/domains/cms/src/lib/models/organizationNews.model.ts
Ingimarsson: Sounds good. We could add an `organization: Organization` field to the news model and make a small change to the web app so that only news entries with no organization are shown on the frontpage.

We also got a suggestion that it could be useful if important organization news could be shown on the frontpage, so maybe we should add something like an `important: boolean` field as well :)

stefanhar: We already have tags on `news`under the `genericTags` field.
You can already query news by tags.
If we create an Organization tag we should not need to create any additional fields or code to fetch the news in each Organization page.
Additionally we might create an important tag but we might need to solve the leaky nested queries problem before using that. I believe it would be worth the work but definitely something we should only cross when we get to it

stefanhar: I should also mention we are using the tags on the news overview atm so you can look at the logic there to see how to query by tag

apps/web/components/Organization/Slice/Districts/DistrictsSlice.tsx
ingolfurs: The problem was that we were importing `Link` from `next/link`, that did not render anchors.

apps/web/components/Organization/Wrapper/SidebarWrapper.tsx
ingolfurs: @siggakogk Of course... thanks for the suggestion

libs/api/domains/cms/src/lib/models/organizationNews.model.ts
Ingimarsson: Thanks, we will look into using the article tags for this. We are going to remove the news from this pull request and add it in a separate PR, so I'm marking this as resolved here.

libs/api/domains/cms/src/lib/models/featuredArticles.model.ts
Ingimarsson: Turns out that using `Article` works fine after all. I have simplified the model by using the `Article` model instead.

apps/web/components/Organization/Slice/FeaturedArticles/FeaturedArticlesSlice.tsx
```diff
@@ -0,0 +1,101 @@
...
+            </GridColumn>
+          </GridRow>
+          <Box
+            display="flex"
+            justifyContent="flexEnd"
+            paddingTop={4}
+            paddingBottom={1}
+          >
+            <Text variant="h5" as="p">
+              <Link href="#">
```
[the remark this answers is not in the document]
ingolfurs: The screen for Sýslumenn all services in not ready, so I removed it for now and will add it back when that is ready

apps/web/components/Organization/Wrapper/SidebarWrapper.tsx
```diff
@@ -0,0 +1,64 @@
...
+        display="flex"
+        flexDirection={['column', 'column', 'row']}
+        height="full"
+        paddingBottom={6}
+      >
+        <Box
+          printHidden
+          className={styles.sidebarWrapper}
+          display={
+            hideSidebarInMobile ? ['none', 'none', 'block'] : ['block', 'block']
```
barabrian, line 35:
```suggestion
            hideSidebarInMobile ? ['none', 'none', 'block'] : 'block'
```

apps/web/components/Organization/Wrapper/SidebarWrapper.tsx
```diff
@@ -0,0 +1,64 @@
...
+          }
+        >
+          {sidebarContent}
+        </Box>
+        <GridContainer>
+          <GridRow>
+            <GridColumn
+              offset={['0', '0', '0', '0', '1/9']}
```
barabrian, line 43: does this offset not rely on `fullWidthContent` like span?

apps/web/screens/Syslumenn/SubPage.treat.ts
```diff
@@ -0,0 +1,18 @@
...
+})
+
+export const smallDescription = style({})
+
+globalStyle(`${smallDescription} p`, {
+  fontSize: 18,
+  fontWeight: 300,
+  lineHeight: '34px',
+  marginBottom: 32,
+})
```
barabrian, line 18: are these styles being used anywhere?

barabrian (commented):
few comments :)

barabrian (approved):

stefanhar (approved):

PR #2696, opened 2021-01-27
Sjukratryggingar/old pending applications

dd-sdang:
# Sjukratryggingar - get old pending application from 

## What

- Remove national id from query
- Added query to get old pending applications in sjukra
- Show modal if pending applications exist

## Why

Because in the transition period of using this application and the old, there might be users that have active applications with the old. We want to inform those users with this feature.

## Screenshots / Gifs
Screenshot 2021-01-27 at 18 23 53

## Checklist:


tthornvik (approved):
Looks good to merge

PR #2702, opened 2021-01-28
Error page dash fix

stefanhar:
# \<Description\>

https://app.asana.com/0/1199123945323117/1199693578639224

## What

- Made route template regex match dash
- Added additional test for this case
- Stopped using type resolver on error page

## Why

- Dash in url causes 500 error on error page
- Make the type resolver more robust

## Checklist:


apps/web/hooks/useLinkResolver/useLinkResolver.spec.tsx
```diff
@@ -163,14 +163,28 @@ describe('Type resolver', () => {
...
   })
+
+  it('Should resolve paths with dashes', () => {
+    const types = typeResolver('/andes-foo/andes-bar')
+    expect(types).toEqual({
+      type: 'subarticle',
+      locale: 'is',
+      slug: ['andes-foo', 'andes-bar'],
+    })
+  })
```
siggakogk, line 187: nice!

siggakogk (approved):
LGTM

AndriMar:
If I understand this correctly this is only fixing the `-` character and not all other special and icelandic characters `.` `~` `æ`
I think this fix needs to be more generic if I understand this correctly.

eirikurn:
These were separate issues. The other one is an internal nextjs bug which is tricky to work around. I think it's been fixed in the latest nextjs and we'll upgrade during next refactor week.

AndriMar:
> These were separate issues. The other one is an internal nextjs bug which is tricky to work around. I think it's been fixed in the latest nextjs and we'll upgrade during next refactor week.

Ok let's see how this will behave after the nextjs update.

PR #2703, opened 2021-01-28
Add organization parameter to getArticles query

Ingimarsson:
# Add organization parameter to getArticles query

## What

This PR adds an organization tag to articles in the Elasticsearch index and changes the `getArticles` query so that we can fetch all articles from a specified organization.

This also makes the `category` parameter of the query optional.

## Why

This is needed for organization pages, where we want to show a list of services provided by a given organization, for example "Þjónustur Sýslumanna"

## Checklist:


libs/api/domains/cms/src/lib/search/importers/article.service.ts
```diff
@@ -104,6 +104,11 @@ export class ArticleSyncService implements CmsSyncProvider<IArticle> {
                 value: x.title,
                 type: 'category',
               })),
+              ...(entry.fields?.organization ?? []).map((x) => ({
+                key: x.fields?.slug,
```
stefanhar, line 108: Key is not optional, we should provide a default.
```suggestion
                key: x.fields?.slug ?? '',
```

libs/api/domains/cms/src/lib/cms.elasticsearch.service.ts
```diff
@@ -41,13 +41,21 @@ export class CmsElasticsearchService {
...
+      size: input.size,
+    }
+
+    if (input.category) {
+      query.tags.push({ type: 'category', key: input.category })
+    }
+
+    if (input.organization) {
+      query.tags.push({ type: 'organization', key: input.organization })
     }
```
stefanhar, line 59: This might not work as expected.
I believe the query underlying `getDocumentsByMetaData` might not scope tags as expected.
Meaning a query that requests a document by `category` and `organization` tag will return documents having either a matching category tag or matching organization tag.
While I think the expected behavior is returning documents that have both tags.

If you require this kind of scoping maybe we should look into changing the  underlying query to allow for this?

Ingimarsson: It seems to work as expected. I did some playing around and the following query returns 6 articles
```
{
  getArticles(input:{lang: "is", organization: "syslumenn", category: "akstur-og-bifreidar"}) {
    title
  }
}
```

But if I exclude the category parameter I get all 54 articles belonging to Sýslumenn.

stefanhar: Nice, never been so happy to be wrong.
I'll approve

stefanhar (approved):

PR #2712, opened 2021-01-28
Fix slug strings for table of context heading ids

stjanilofts:
## What

Recursively find the innermost text within the render node to create a slug from.

## Why

Sometimes the text was deeper in the render node (like h2 -> strong -> em) so the id was just "object-object". That caused the table of contents links to break for these types of headings. The easy fix would be to disable strong/em and such but we still want them to be available for other text inside articles so that is not an option.

## Checklist:


siggakogk (approved):
LGTM

PR #2718, opened 2021-01-28
Residence transfer/pdf creator

Rikkiabb:
# \<Description\>

Endpoint for creating pdf for children legal residence transfer. The process is the following:
1. Create pdf from application.
2. Upload pdf to s3 bucket.
3. Create presigned url for bucket.
4. Return presigned url to be used in client.

The endpoint takes in an application and a type which refers to a pdf type. Current types for pdf's is only one: children legal residence transfer, but when more pdf's are needed then we can add to this type and create a template for it and that's it. 

The endpoint uses pdfkit. I searched slack and saw talk about IText Pdf for pdf creation (https://islandis.slack.com/archives/C019VN1KG04/p1599483604046900?thread_ts=1599468993.039200&cid=C019VN1KG04) but didn't find any usage for it in the monorepo but found that pdfkit was used so went with that for now. This decision can and in my opinion should be revisited on a larger scale than this PR.

## Checklist:


apps/application-system/api/src/app/modules/application/utils/pdf.ts
```diff
@@ -0,0 +1,145 @@
...
+
```
Rikkiabb, line 29: I hate this data uri string, but it what we are working with in this pdf file builder. Wondering if it would be nicer to move to another place to hide the uglyness 🙈  Or if I should keep it as close to the code as possible since it is not being reused.

apps/application-system/api/src/app/modules/application/file.service.ts
```diff
@@ -0,0 +1,106 @@
...
+    const parentA: ParentResidenceChange = {
+      id: answers.ssn as string,
+      name: answers.name as string,
+      ssn: answers.ssn as string,
+      phoneNumber: answers.phoneNumber as string,
+      email: answers.email as string,
+      address: answers.address as string,
+      postalCode: answers.postalCode as string,
+      city: answers.city as string,
+    }
```
Rikkiabb, line 56: I wish there was a better way to get the types from the application, any suggestions/ideas/thoughts?

apps/application-system/api/src/app/modules/application/utils/pdf.ts
VilliGunn: Well, this is not ideal but I don't have a preference if it should be here or in an external file 😅

apps/application-system/api/src/app/modules/application/file.service.ts
```diff
@@ -0,0 +1,106 @@
...
+    const parentBNationalRegistry = externalData.parentNationalRegistry as FormValue
+    const childrenAppliedFor = (answers.selectChild as unknown) as Array<
+      PersonResidenceChange
+    >
+    const parentB = (parentBNationalRegistry.data as unknown) as ParentResidenceChange
+
+    parentB.email = answers.parentBEmail as string
+    parentB.phoneNumber = answers.parentBPhoneNumber as string
+
+    // TODO: Revisit once connection with national registry is up and we have actual schema for this data.
```
VilliGunn, line 46: This can be updated now, we have the schema here:
```
{ NationalRegistryUser } from '@island.is/api/schema'
```

VilliGunn (commented):
Just this one minor comment about the schema. The pdf generation will probably change slightly when we have real data and when we add translations 💯

Rikkiabb:
I am wondering if I should put the new files all in it's own folder and then add kolibri as codeowner to that folder?

apps/application-system/api/src/app/modules/application/files/file.service.ts
```diff
@@ -0,0 +1,108 @@
...
+    const pdfBuffer = await generateResidenceChangePdf(
+      childrenAppliedFor,
+      parentA,
+      parentB,
+      expiry,
+    )
+
+    const id = uuid()
+    const fileName = `${parentA.ssn}/${id}.pdf`
+    const bucket = `${process.env.NODE_ENV}-legal-residence-change`
```
sindrig, line 71: Can we pull this in as an environment variable/secret instead of hardcoding it? We'd be dead in the water if `production-legal-residence-change` has been taken by someone else :)

apps/application-system/api/src/app/modules/application/files/utils/pdf.ts
```diff
@@ -0,0 +1,153 @@
+import PDFDocument from 'pdfkit'
+import streamBuffers from 'stream-buffers'
+import {
+  ParentResidenceChange,
+  PersonResidenceChange,
+} from '@island.is/application/templates/children-residence-change'
+import { Constants } from './constants'
+import { DistrictCommissionerLogo } from './districtCommissionerLogo'
+
+export async function generateResidenceChangePdf(
```
sindrig, line 10: tests for any of this new stuff?

sindrig (changes requested):

apps/application-system/api/src/app/modules/application/files/file.service.ts
Rikkiabb: Makes sense. 😅  

First time adding a secret for this project. Could you tell me about that process? 

I get that they are stored in AWS Parameter store and need to be labeled with `dev` and can of course follow examples in the codebase but think it would be good to hear about it instead of guessing 🙏

sindrig: There's really nothing secret about bucket names, so no need for parameter store IMO. You can just add 3 different environment values to the `env` section in helm: https://github.com/island-is/helm/blob/main/src/services/application-system/application-system-api.ts#L11. Something like

```
FILE_SERVICE_PRESIGN_BUCKET: {
    dev: 'island-is-development-fs-presign-bucket',
    staging: 'island-is-staging-fs-presign-bucket',
    prod: 'island-is-production-fs-presign-bucket',
}
```

Then you'll get the environment variable `node.env.FILE_SERVICE_PRESIGN_BUCKET` in you pod in k8s.

Then ping one of the devops folks and we'll create the buckets for you :smiley:. We'll want some characteristics for the bucket, such as if we can delete files after some time, if we need versioned objects etc.

apps/application-system/api/src/environments/environment.ts
```diff
@@ -25,4 +25,5 @@ export default {
   auth: {
     jwtSecret: 'supersecret',
   },
+  fsS3Bucket: process.env.FILE_SERVICE_PRESIGN_BUCKET
```
Rikkiabb, line 28: Do you hard code the value for development or pick it up from the process @sindrig?

apps/application-system/api/src/app/modules/application/files/file.service.ts
```diff
@@ -0,0 +1,109 @@
...
+      .upload(uploadParams)
+      .promise()
+      .catch(() => {
+        return null
+      })
+
+    const presignedUrlParams = {
+      Bucket: bucket,
+      Key: fileName,
+      Expires: this.one_minute * 120, // TODO: Select length for presigned url's in island.is
```
Rikkiabb, line 100: Do you have any thoughts about the expiry @sindrig? This number is totally random atm.

madebynoam (dismissed):
Overall this looks good to me! @baering will be able to add more details as needed.

apps/application-system/api/src/app/modules/application/file.service.ts
madebynoam: Agree on this, we want it on the client as well... @baering potential future item?

Rikkiabb: Totally agree 🙏 
Would love to see when we start a new template for the application system that we start with defining a type for that application and then we could use that for type safety both in frontend and backend. 😍

PR #2731, opened 2021-02-01
Residence change/formexternaldataprovider label

VilliGunn:
# \<Description\>

https://app.asana.com/0/1199733748699731/1199674225481888

We need to be able to control/translate the checkbox label and sub title in the `FormExternalDataProvider`

## What

- Added `subTitle` and `checkboxLabel` props
- Fallback to the Icelandic values so other teams can start using the props when it suits them

## Why

- To control the text

## Screenshots / Gifs

Screenshot 2021-02-01 at 10 02 51

## Checklist:


libs/application/ui-shell/src/components/FormExternalDataProvider.tsx
```diff
@@ -125,7 +124,7 @@ const FormExternalDataProvider: FC<{
           <Icon icon="download" size="medium" color="blue400" type="outline" />
         </Box>
         <Text variant="h4">
-          Eftirfarandi gögn verða sótt rafrænt með þínu samþykki
+          {subTitle || 'Eftirfarandi gögn verða sótt rafrænt með þínu samþykki'}
```
Rikkiabb, line 127: Wondering if we want to have the hardcoded fallback here? And rather just do:
`subtitle && ....`
My worry is when we change the form and this doesn't update that it could be confusing. Just a thought 🤔

libs/application/core/src/types/Form.ts
```diff
@@ -89,6 +89,8 @@ export interface ExternalDataProvider extends FormItem {
   readonly children: undefined
   isPartOfRepeater?: boolean
   dataProviders: DataProviderItem[]
+  checkboxLabel?: MessageDescriptor | string
+  subTitle?: MessageDescriptor | string
```
Rikkiabb, line 93: Should this type be like this in the future or just a `MessageDescriptor`  or just a `string` ?

Rikkiabb (commented):
Two questions from me 😊

libs/application/core/src/types/Form.ts
VilliGunn: This is how things are currently typed in the application system. What I would like to see is just `string` and if there needs to be some processing of the value it should be done before passing it down as a prop.

libs/application/ui-shell/src/components/FormExternalDataProvider.tsx
VilliGunn: This is done for 'backwards compatibility' if I change this as you suggested I would need to change all the projects that currently use the `FormExternalDataProvider`

Rikkiabb (approved):
LGTM

DavidBachmann (approved):

PR #2732, opened 2021-02-01
Application System: FileUploadField updates

madebynoam:
Updates and fixes to the FileUploadField

1. Fix 400 errors from AWS
2. Catch non 200 responses
3. Move the core of the field to shared, so that it can be used in a review screen
4. Add work in progress readme.

## TODO (next PR)
1. Add functionality to block the Continue button when uploading (currently the user can make it so that the answer is not saved if they click Continue before the upload is done).
2. Localise error messages (when Jeremy introduces the application system messages file in his PR).
3. See remaining items in readme for future updates

## Checklist:


libs/application/ui-fields/README.md
```diff
@@ -9,3 +9,112 @@ Each Field has to have its own unique `type` which needs to be part of the `Fiel
...
+
+## FileUploadFormField
+
+The file upload form field provides a general way to upload files for your application.
+
+### To use the FileUploadFormField in an application
+
+1. Add it to the _schema_ using the following structure:
+
+```
```
jeremybarbet, line 23: You can add `typescript` to get nice color syntax for the following code snippets

```suggestion
```typescript
```

libs/shared/form-fields/src/lib/FileUploadController/index.tsx
```diff
@@ -0,0 +1,251 @@
...
+import { useLocale } from '@island.is/localization'
+
+type UploadFileAnswer = {
+  name: string
+  key?: string
+  url?: string
+}
+
+// Transform an uploaded file to an form answer.
+const transformToAnswer = (f: UploadFile): UploadFileAnswer => {
```
jeremybarbet, line 28: You could destructurate to make it easier for the return

```suggestion
const transformToAnswer = (({ name, key, url }): UploadFile): UploadFileAnswer => {
  return { name, key, url }
```

libs/shared/form-fields/src/lib/FileUploadController/index.tsx
```diff
@@ -0,0 +1,251 @@
...
+            key: fields.key,
+            url: `${response.url}/${fields.key}`,
+          },
+        },
+      })
+
+      // Done!
+      return Promise.resolve({ url: response.url, key: fields.key })
+    } catch (e) {
+      error = e
```
jeremybarbet, line 143: Wouldn't it be better to have a useState for that with the default value to be the error props and then we can use the setter to pass the new e object as an error?

libs/shared/form-fields/src/lib/FileUploadController/index.tsx
```diff
@@ -0,0 +1,251 @@
...
+    // with a spinner.
+    dispatch({
+      type: ActionTypes.ADD,
+      payload: {
+        newFiles: newUploadFiles,
+      },
+    })
+
+    // Upload each file.
+    newUploadFiles.forEach(async (f: UploadFile) => {
```
jeremybarbet, line 176: Using async/await and then/catch together is a bit weird. I would probably do

```typescript
Promise.all(newUploadFiles.map(async (f: UploadFile) => {
  try {
    const res = await uploadFileFlow(f);
    
    dispatch({
      type: ActionTypes.UPDATE,
      payload: {
        file: f,
        status: 'done',
        percent: 100,
        key: res.key,
        url: res.url,
      },
    })
  } catch (e) {
    ...
  }
}))
```

libs/application/ui-fields/README.md
madebynoam: Oh nice thanks! Updated now.

libs/shared/form-fields/src/lib/FileUploadController/index.tsx
madebynoam: Nice, agree, switched now.

libs/shared/form-fields/src/lib/FileUploadController/index.tsx
madebynoam: Good point thanks, updated now.

libs/shared/form-fields/src/lib/FileUploadController/index.tsx
madebynoam: Makes sense thanks, I've refactored it to use the `uploadError` state consistently - with the `error` prop as the default.

madebynoam:
Thanks for the feedback, should be ready for re-review.

jeremybarbet (approved):

PR #2733, opened 2021-02-01
Do not post success if retag failed

sindrig:
we don't want something like this to be green: https://github.com/island-is/island.is/runs/1806040890?check_suite_focus=true


pshomov (approved):
LGTM

PR #2752, opened 2021-02-02
Judicial-system/Fix null if no defender is set

oddsson:
# Add defender to pdfs and prevent showing "defender: null" if no defender is set in the email that goes out to the prisons.

https://app.asana.com/0/1199153462262248/1199584850352431

## What

See ☝️

## Why

The prison email thing is a bug and adding the defender to the pdf is for better ux.

## Screenshots / Gifs

Screen Shot 2021-02-02 at 13 07 37

Screen Shot 2021-02-02 at 13 10 31

## Checklist:


apps/judicial-system/backend/src/app/formatters/pdf.ts
```diff
@@ -199,6 +207,7 @@ export async function generateRulingPdf(
   existingCase: Case,
   user: User,
 ): Promise<string> {
+  console.log('DEFNAME', existingCase.defenderName === '')
```
gudjong, line 210: Remove console log

apps/judicial-system/backend/src/app/formatters/pdf.ts
```diff
@@ -242,6 +251,13 @@ export async function generateRulingPdf(
       lineGap: 6,
       paragraphGap: 0,
     })
+    .text(
+      `Verjandi sakbornings: ${
+        isFalsy(existingCase.defenderName)
+          ? 'Hefur ekki verið skráður.'
+          : existingCase.defenderName
+      }`,
+    )
```
gudjong, line 260: Why add the defender here? The defender is listed in the courtAttendees just above.

gudjong (changes requested):
Some comments

apps/judicial-system/backend/src/app/formatters/pdf.ts
oddsson: Not if the defender is not set. I've changed this to say nothing if the defender is set and appears in the courtAttendees and "Hefur ekki verið skráður" if not. Does that make sense to you?

gudjong: Not really. You don't know if courtAttendees includes a defender because it is s free text field. I think the ticket in Asana only specified that the prosecutor request should include the defender. The ruling is a different matter. Ask Anna or Baldur perhaps?

oddsson: Ah! I removed this code. Let's have tensions come up in review. Sounds good?

gudjong (approved):
Very nice

PR #2754, opened 2021-02-02
Feature flag documentation

sindrig:
This closes island-is/infrastructure#430


scripts/secrets.sh
```diff
@@ -20,6 +20,8 @@ function main {
     echo "export $secret" >> $env_secret_file
   done
 
+  echo "export CONFIGCAT_SDK_KEY='$(aws ssm --region eu-west-1 get-parameter --name=/k8s/configcat/sdk-key --profile=island-dev --with-decryption --query=Parameter.Value --output=text)'" >> $env_secret_file
```
darrikonn, line 23: This should not be needed. We can fetch this variable with:
`./secrets.sh configcat`

Would be nicer to rename the parameter in the parameter store as: `CONFIGCAT_SDK_KEY`

handbook/technical-overview/feature-flags.md
```diff
@@ -0,0 +1,18 @@
+# Feature Flags
+
+## Developer usage
+
+If you want to introduce something behind a feature flag you can follow these steps:
+
+1. Ask someone from DevOps for invite to ConfigCat.
+2. Once you're in (https://app.configcat.com/) you can add your feature flag. The initial values should always be "On" in Dev and (probably always to start with) "Off" in Production and Staging.
+3. Make sure that the CONFIGCAT_SDK_KEY environment variable is in `.env.secrets` in the root of the repository. You can fetch it using the `./scripts/secrets.sh` script.
```
darrikonn, line 9: Then update this to use the `configcat` argument.

handbook/technical-overview/feature-flags.md
```diff
@@ -0,0 +1,18 @@
...
+4. Start using your flag by using the package `@island.is/feature-flags`.
+
+## Naming convention
+
+When creating your feature flag, please try to have both the "Key for humans" and "key for programs" descriptive. If the feature flag will be used by a single service, make sure that the human key contains the service name.
```
darrikonn, line 14: Would be nice to show an example.

darrikonn (changes requested):
Nice

scripts/secrets.sh
darrikonn: Just needs the `dev` label (I added it on dev, but as I mentioned above, it needs a rename)

sindrig: That would override the secrets for the service you're developing. This environment variable is set for all services, always, so I felt it made more sense to include it for all services. That way you can keep doing `./scripts/secrets.sh gjafakort` (see https://github.com/island-is/helm/pull/275)

handbook/technical-overview/feature-flags.md
sindrig: I'll add an example

scripts/secrets.sh
darrikonn: No it doesn't. It only overrides if you pass the --rest flag with. 

So this would just safely append it to the list

sindrig: crawlintohole

sindrig: notthedroids

darrikonn (approved):
🎉

PR #2756, opened 2021-02-02
Standardized the parameter format in the contentful service

kksteini:
# \<Description\>

Some of the functions in the contentful service declared a parameter object whereas other did not.

## What

I standardized the contentful service to parameter object declaration instead of constructing it in the function call itself

## Why

consistency


stefanhar (approved):
LGTM

PR #2765, opened 2021-02-03
hotfix/secret-service-length-bug

busla:
Hotfixed `create-secret` script

- maxlength now 128
- improved error handling


sindrig (approved):
you've got a problem. you decide to use regex. now you've got two problems :)

busla:
> you've got a problem. you decide to use regex. now you've got two problems :)

According to @pshomov you two have a grand and gloriuos `cli` plan. Until that is resolved I'm stuck with `bash`.

PR #2767, opened 2021-02-03
Domicile transfer/pdf creator

Rikkiabb:
# \<Description\>
Endpoint for creating pdf for children legal residence transfer. The process is the following:

Create pdf from application.
Upload pdf to s3 bucket.
Create presigned url for bucket.
Return presigned url to be used in client.
The endpoint takes in an application and a type which refers to a pdf type. Current types for pdf's is only one: children legal residence transfer, but when more pdf's are needed then we can add to this type and create a template for it and that's it.

The endpoint uses pdfkit. I searched slack and saw talk about IText Pdf for pdf creation (https://islandis.slack.com/archives/C019VN1KG04/p1599483604046900?thread_ts=1599468993.039200&cid=C019VN1KG04) but didn't find any usage for it in the monorepo but found that pdfkit was used so went with that for now. This decision can and in my opinion should be revisited on a larger scale than this PR.

## Checklist:


darrikonn (approved):
core/devops files look good

apps/application-system/api/src/environments/environment.ts
```diff
@@ -25,4 +25,5 @@ export default {
   auth: {
     jwtSecret: 'supersecret',
   },
+  fsS3Bucket: process.env.FILE_SERVICE_PRESIGN_BUCKET,
```
Rikkiabb, line 28: Is this always read from the proccess environment or is sometimes hardcoded for local development?

apps/application-system/api/src/app/modules/application/files/file.service.ts
```diff
@@ -0,0 +1,126 @@
...
+      .upload(uploadParams)
+      .promise()
+      .catch(() => {
+        return null
+      })
+
+    const presignedUrlParams = {
+      Bucket: bucket,
+      Key: fileName,
+      Expires: this.one_minute * 120, // TODO: Select length for presigned url's in island.is
```
Rikkiabb, line 117: Does island.is have any guidelines on how long presigned urls should be active? Currently this is just a random number....

apps/application-system/api/src/app/modules/application/files/file.service.ts
```diff
@@ -0,0 +1,126 @@
...
+    const parentA: ParentResidenceChange = {
+      id: nationalRegistryData.nationalId,
+      name: nationalRegistryData.fullName,
+      ssn: nationalRegistryData.nationalId,
+      phoneNumber: answers.phoneNumber as string,
+      email: answers.email as string,
+      address: nationalRegistryData.address?.streetAddress as string,
+      postalCode: nationalRegistryData.address?.postalCode as string,
+      city: nationalRegistryData.address?.city as string,
+    }
```
Rikkiabb, line 67: I added tests for this. But is there a way to get type safety for this? Or do we always have to this thingamajig and trust the type will be cast to the correct type?

libs/application/api-template-utils/src/lib/types.ts
```diff
@@ -19,3 +19,7 @@ export type SendEmail = { type: 'sendEmail'; template: SendMailOptions }
 export type ApplicationAPITemplateAction =
   | SendEmail
   | AssignApplicationThroughEmail
+
+export enum PDF_TYPES {
```
jeremybarbet, line 23: A bit weird to have this type located here, while the api-template-utils doesn't do anything regarding PDF. I would define it locally in your app instead and if it's need to be shared between libraries and app then better to define inside shared/tyoes

VilliGunn (approved):
No further comments from me since the first review. Nice changes 💯

libs/application/api-template-utils/src/lib/types.ts
Rikkiabb: Totally agree, will move 👍

jeremybarbet (approved):
LGTM 👍

darrikonn (approved):
devops/core files look good

atligudl (approved):

pshomov (approved):
LGTM

VilliGunn (approved):

Thorrik58 (approved):

PR #2771, opened 2021-02-03
Feature/checkbox radio sublabel

VilliGunn:
# \<Description\>

https://app.asana.com/0/1199733748699731/1199584850352412

Introducing a `subLabel` that can be displayed below the `label` text in Radio buttons and Checkboxes when a `large` button is being used.

Only used in the Children Residence Change application for now.

This is according to designs shown below.

## What

Having the option of adding a sub label to checkboxes and radio buttons.

## Why

To have more detailed information more visible to users.

## Screenshots / Gifs

Radio Buttons
Screenshot 2021-02-04 at 14 35 08

Checkboxes
Screenshot 2021-02-04 at 14 35 24

## Checklist:


libs/application/ui-fields/src/lib/CheckboxFormField.tsx
```diff
@@ -58,9 +58,12 @@ const CheckboxFormField: FC<Props> = ({
             useDefaultValue(field, application)
           }
           error={error}
-          options={finalOptions.map(({ label, tooltip, ...o }) => ({
+          options={finalOptions.map(({ label, subLabel, tooltip, ...o }) => ({
             ...o,
             label: HtmlParser(formatText(label, application, formatMessage)),
+            ...(subLabel && {
+              subLabel: formatText(subLabel, application, formatMessage),
+            }),
```
VilliGunn, line 66: This is how things are done in multiple places in the application system. Strings are formatted somewhere on their way down the component tree.
I want to suggest that we change this behaviour and do all formatting needed at the highest level and pass down a string. Not in some cases a string and in other cases a MessageDescriptor. That would also make our types more easier to follow. I feel like we have too many places where types are too broad.

libs/island-ui/core/src/lib/Checkbox/Checkbox.tsx
```diff
@@ -79,7 +82,12 @@ export const Checkbox = ({
             color={checked ? 'white' : 'transparent'}
           />
         </div>
-        {label}
+        <span className={styles.labelText}>
```
oddsson, line 85: Consider using the Text component here instead of a span. That way, any changes made to the Typography in the future are reflected here as well. You can use `as="span"` to render Text as a span.

libs/island-ui/core/src/lib/RadioButton/RadioButton.tsx
```diff
@@ -74,7 +77,12 @@ export const RadioButton = ({
         >
           <div className={styles.checkMark} />
         </div>
-        {label}
+        <span className={styles.labelText}>
```
oddsson, line 80: Same as ☝️

libs/shared/form-fields/src/lib/CheckboxController/CheckboxController.spec.tsx
```diff
@@ -47,13 +47,35 @@ describe('CheckboxController', () => {
...
     )
     expect(getByText('nice tooltip')).toBeInTheDocument()
   })
+
+  it('should render a sublabel below an option if wished and when large set to true', () => {
+    const options = [
+      { label: 'some checkbox', value: 'off', subLabel: 'nice sublabel' },
+      { label: 'another checkbox', value: 'on' },
+    ]
+    const { getByText } = render(
```
oddsson, line 61: Please use screen.getByText instead of this. (source https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-using-screen)

libs/shared/form-fields/src/lib/RadioController/RadioController.spec.tsx
```diff
@@ -43,9 +43,35 @@ describe('RadioController', () => {
...
     )
     expect(getByText('nice tooltip')).toBeInTheDocument()
   })
+
+  it('should render a sublabel below an option if wished and largeButtons is set to true', () => {
+    const options = [
+      { label: 'some checkbox', value: 'off', subLabel: 'nice sublabel' },
+      { label: 'another checkbox', value: 'on' },
+    ]
+    const { getByText } = render(
```
oddsson, line 57: Same as ☝️

oddsson (changes requested):
Looks good! I had a few comments to address.

libs/island-ui/core/src/lib/Checkbox/Checkbox.tsx
```diff
@@ -16,12 +16,15 @@ export interface CheckboxProps {
   hasError?: boolean
   errorMessage?: string
   value?: string
-  large?: boolean
   filled?: boolean
+  large?: boolean
+  /** subLabel can only be used if the 'large' prop set to true */
+  subLabel?: string
```
jeremybarbet, line 22: You might want [to type guard](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgGIgApQPYAcDOABQN4BQyyAbnADYCuEA-AFzL5hSgDmA3OcjThQuTVgCNs2GhDghSAX1KlQkWIhQBlOmIAycMRBpY8RCAA9IIACb40mHAWRkKg4RHGTpsvhXza9BjQsbBzcfIqkYACeuCjoACoxKAC8dsaOAD7IABTo6UQAZMhauvqG+QCUSjB0IAhgwNggyAAWcPglAYbZ2GIAVqx5DrZZnWVGwxWsvX3IwLZjgfkk-FAQYHRQzT39yO3F-uOVAHR+pYHIAITJqbVWEDCgEFYKSghN7MgwzanZuMODewmZCjQ5LSbIZIAPhWFGAMBybQ6YO6-xMFQqsIoyDRBGO1HoEH4FFx+GOrhExJxw1OKJo-AiJJpBIY-FJ5KElKZJlp50MyAA9ALkAARbAQfDgZDmYBgAA0yAMCDgdHwKGwCLALRQ0ViyC4dCEVmQCG1CAA1q9qiBss4qLQGKwAOT4bAAW3WLW4ToUVVI31t-Ap7mQHAYct91sDLk5IfgNDVEd8dOdgkCToj8j9AbtZy6NFT4wzvqAA) this props so you have better check over the usage of the props

VilliGunn: Yes definitely! 💯

libs/island-ui/core/src/lib/Checkbox/Checkbox.treat.ts
```diff
@@ -34,6 +34,15 @@ export const label = style({
   fontSize: theme.typography.baseFontSize,
   fontWeight: theme.typography.light,
 })
+export const labelText = style({
+  display: 'flex',
+  flexDirection: 'column',
+})
+export const subLabel = style({
+  marginTop: '2px',
+  fontSize: '14px',
```
madebynoam, line 43: As mentioned below, we can replace the `fontSize` with `<Text variant="small">`

libs/island-ui/core/src/lib/Checkbox/Checkbox.tsx
VilliGunn: Yes, I will use the `<Text>` component for the `subLabel` but I will keep this span as is since it's only used as a container for the `label` and `subLabel` and no text styles are applied.

libs/island-ui/core/src/lib/Checkbox/Checkbox.tsx
VilliGunn: We are going to revisit this at a later date.

jeremybarbet (approved):
LGTM

libs/island-ui/core/src/lib/Checkbox/Checkbox.tsx
Rikkiabb: This is more of a gerneral question and not todo in this PR. 

But when do we decide to break components into smaller components and this for example would be: `CheckBox` and `LargeCheckbox` ? 

I am mainly wondering because of this logic and the styles below are dependent on the large variable currently.

Rikkiabb (approved):
:shipit:

libs/island-ui/core/src/lib/Checkbox/Checkbox.tsx
VilliGunn: I think we are at that point now where we should reconcider how both the `<Checkbox>` and the `<RadioButton>` components are set up.

Rikkiabb: I agree, but out of scope for this PR. Was mainly wondering what the process for it was. Could we pick it up if we feel there is a need for it or is it a decision of the codeowner or something else?

oddsson: It's totally up to you! If you feel a change is needed, feel free to implement it 🙏

oddsson (approved):

PR #2777, opened 2021-02-04
restrict document-provider module to list of nationalIDs

fjandakornid:
# Restrict document-provider module in service-portal to list of national IDs

## What

Restrict access to the document-provider module in service-portal to list of national IDs

## Why

Only few people should have access to the document-provider module. Since we don't have user claims in the frontend, this has to be fixed temporarily by getting list of nationalIDs from environment variable and check if current user is in that list. 
This fix only removes the link from the navigation menu, not the access to the route (if known by the user). That check will be implemented in the module's pages.

## Checklist:


apps/service-portal/src/hooks/useNavigation/useNavigation.ts
```diff
@@ -22,6 +23,12 @@ const filterNavigationTree = (
           route.path.includes(item.path)),
     ) !== undefined || item.systemRoute === true
 
+  // Only display the document-provider module for those who have access
+  if (item.path === '/skjalaveita') {
+    if (env.documentProviderAdmins?.indexOf(userInfo.profile.nationalId) === -1)
```
Thorrik58, line 28: Could also have used includes which I would prefer since its more readable and we dont really care about the index for this case..
```suggestion
    if (!env.documentProviderAdmins?.includes(userInfo.profile.nationalId))
```

fjandakornid: Good point... will fix

Thorrik58 (approved):
LGTM! 🎉

fjandakornid:
Merge is blocked by https://github.com/island-is/helm/pull/279

PR #2784, opened 2021-02-04
Judicial System / Select judge

addi:
# Select judge

https://app.asana.com/0/1199153462262248/1199174045356607

Screenshot 2021-02-04 at 15 32 19

## Checklist:


apps/judicial-system/backend/src/app/modules/case/case.controller.ts
```diff
@@ -197,22 +197,7 @@ export class CaseController {
...
-      user.role === UserRole.JUDGE
-    ) {
-      update = {
-        state,
-        judgeId: user.id,
-      } as UpdateCaseDto
-    } else {
-      update = { state } as UpdateCaseDto
-    }
+    let update = { state } as UpdateCaseDto
```
gudjong, line 200:
```suggestion
    const update = { state } as UpdateCaseDto
```

apps/judicial-system/backend/src/app/modules/case/case.controller.ts
```diff
@@ -305,15 +286,12 @@ export class CaseController {
   ) {
     const existingCase = await this.findCaseById(id)
 
-    if (user.role !== UserRole.JUDGE) {
-      throw new ForbiddenException('A ruling must be signed by a judge')
+    if (user.role !== UserRole.JUDGE && user.id !== existingCase.judge.id) {
```
gudjong, line 289:
```suggestion
    if (user.role !== UserRole.JUDGE && user.id !== existingCase.judgeId) {
```

The database stores the judge id in the column judge_id which is available in the case model as the field judgeId. The judge fieled is filled in automatically by querying the database for a user by judgeId. It is safer to compare to judgeId, because it could be null and if so judge will also be null and we will get a null pointer reference exception.

apps/judicial-system/web/src/routes/Judge/CourtRecord/CourtRecord.tsx
```diff
@@ -89,6 +89,10 @@ export const CourtRecord: React.FC = () => {
     const defaultCourtAttendees = (wc: Case): string => {
       let attendees = ''
 
+      if (wc.judge) {
+        attendees += `${wc.judge?.name} ${wc.judge?.title}\n`
```
gudjong, line 93:
```suggestion
        attendees += `${wc.judge.name} ${wc.judge.title}\n`
```

Question marks not necessary after null check above. Actually, the same applies to the prosecutor below.

gudjong (changes requested):

gudjong (approved):

apps/judicial-system/web/src/routes/Judge/Confirmation/Confirmation.tsx
```diff
@@ -612,10 +612,16 @@ export const Confirmation: React.FC = () => {
           </Box>
           <FormFooter
             nextUrl={Constants.REQUEST_LIST_ROUTE}
-            nextButtonText="Staðfesta og hefja undirritun"
+            nextButtonText={'Staðfesta og hefja undirritun'}
```
oddsson, line 615:
```suggestion
            nextButtonText="Staðfesta og hefja undirritun"
```

apps/judicial-system/web/src/routes/Judge/HearingArrangements/HearingArrangements.tsx
```diff
@@ -105,6 +125,16 @@ export const HearingArrangements: React.FC = () => {
     return data?.sendNotification?.notificationSent
   }
 
+  const judges = (userData?.users || [])
+    .filter((user: User) => user.role === UserRole.JUDGE)
+    .map((prosecutor: User) => {
```
oddsson, line 130: Shouldn't this be `judge: User ...`?

apps/judicial-system/web/src/routes/Judge/HearingArrangements/HearingArrangements.tsx
```diff
@@ -105,6 +125,16 @@ export const HearingArrangements: React.FC = () => {
...
+      return { label: prosecutor.name, value: prosecutor.id }
```
oddsson, line 131: Same ☝️

apps/judicial-system/web/src/routes/Judge/HearingArrangements/HearingArrangements.tsx
```diff
@@ -147,17 +177,29 @@ export const HearingArrangements: React.FC = () => {
     ]
 
     if (workingCase) {
-      setIsStepIllegal(isNextDisabled(requiredFields))
+      setIsStepIllegal(
+        isNextDisabled(requiredFields) || workingCase.judge === null,
```
oddsson, line 181: Would it be possible to write `!workingCase.judge` instead?

apps/judicial-system/web/src/routes/Judge/Ruling/StepTwo/RulingStepTwo.tsx
```diff
@@ -114,7 +114,7 @@ export const RulingStepTwo: React.FC = () => {
                   Úrskurðarorð
                 </Text>
               </Box>
-              <Box marginBottom={3}>getConclusion(workingCase)</Box>
+              <Box marginBottom={3}>{getConclusion(workingCase)}</Box>
```
oddsson, line 117: 🙈

apps/judicial-system/web/src/shared-components/FormFooter/FormFooter.tsx
```diff
@@ -43,6 +44,27 @@ const FormFooter: React.FC<Props> = (props: Props) => {
           {props.nextButtonText ?? 'Halda áfram'}
         </Button>
       )}
+      {props.infoBoxText && (
+        <Box
```
oddsson, line 48: Island-ui has a component for this, right? Why not use that?

oddsson (changes requested):
Just a few things to look into 👀

apps/judicial-system/web/src/shared-components/FormFooter/FormFooter.tsx
addi: It was only with bold text, looked weird. So i just copied what i needed. It's just a small component.

oddsson (approved):

oddsson (approved):

PR #2797, opened 2021-02-05
file size limit added

tthornvik:
# Health insurance upload file size limit added

## What

Added a limit to how big the added attachment can be, 10MB

## Checklist:


dd-sdang (approved):
Looks good! We need to remember to set this when we change the component in the Review section as well. When we do that change we could also make this value into a constant so that if it changes later, it will change everywhere :)

PR #2801, opened 2021-02-08
Children residence change/effect of change

Rikkiabb:
# \<Description\>

## What
- The partialSchemaValidation function in `libs/application/core/src/validation/validators.ts` now validates top level items as well as validating nested items if a parent item is invalid. Without this functionality it was impossible to validate partial schema for an array except for checking if the array was empty.  📦 
- Refactored the partialSchemaValidation function, to hopefully make it clearer 🙏 
- Updated tests to correspond to this change. ✅ 
- Added full template validation for children residence change. 👶 
- Noticed that `libs/application/ui-shell/src/components/FormExternalDataProvider.tsx` was not displaying error on invalid input, added that. 🔴 

-> UPDATED:
- Made validation for all templates match the new implementation of the validator. The issues with templates were:
1. Id was not in form which made the validation fail. This was fixed with changing to correct name in the schema or removing the id from the schema when it was no place for it in the form. [Example](https://github.com/island-is/island.is/pull/2801/commits/b8ba31d77b89994f97ba9ec4437256ddbeca9c86#diff-e78fb337f0812af425e6bb1e31aff1acd80f94fa486cede1cb707f9ac4ea18cfR33)
2. Items were required in the schema validation but were added later in the user flow which made the validation fail. [Example](https://github.com/island-is/island.is/pull/2801/files#diff-b3d5c8bab4ac60976b573e9111f135dbb1a7e88861ebb21da2bce6ac58f4b354R30) Here I changed it to optional, but the flow should probably be that this is not a single object since it is not on the same screen and then it can be made mandatory.
3. Items were mutated later but used as a single object. [Example](https://github.com/island-is/island.is/pull/2801/commits/b8ba31d77b89994f97ba9ec4437256ddbeca9c86#diff-5d6ea1b484040d01f8521dde9956f09485acc5bab2737e499730f24cf856442fR38)

## Why
- Without this functionality it was impossible to validate partial schema for an array except for checking if the array was empty. 
- Now the partial validation is more inline with how zod validates schemas.
- Now we only iterate through nested items if a top level item is invalid. This means that validation should be faster and more efficient.

## Checklist:


libs/application/core/src/validation/validators.ts
```diff
@@ -45,49 +44,37 @@ function partialSchemaValidation(
...
       : originalSchema
 
-    const answerIsArray = Array.isArray(answer)
-    const answerIsEmptyArray = Array.isArray(answer) && answer.length === 0
-    const answerIsObject = typeof answer === 'object' && !answerIsArray
-    const answerIsPrimitive = !answerIsArray && !answerIsObject
+    try {
+      trimmedSchema.parse({ [key]: answer })
+    } catch (e) {
+      error = populateError(error, e, newPath)
```
dabbeg, line 50: I would think this would show an error message in the form when we have an array or object, is that not the case?

Rikkiabb: It depends if the array or object is valid. 
Zod is able to check if an object/array/primitive is valid. 
So if we have an valid array/object then this will try to parse it and not throw an error and continue with life. 
But if the array/object is invalid then the function would iterate through nested values and see what items are invalid, then the error response would be something like this:
```
invalidArray: invalid input
invalidArray[${index}]: invalid input
invalidArray[${index}].invalidVarible: invalid input
```

So you get top down what items are invalid in the schema which means we can choice on which level we handle it in the template.
I tried to reflect the in the tests, but would be happy to add more tests if you have suggestions for any or add an comment if you think that would help 😊 
This function is not straight forward, so everything like this should help future devs with debugging/iterating on it.

dabbeg: Awesome, looks good!

dabbeg (dismissed):
Nice!

libs/application/templates/children-residence-change/src/lib/ChildrenResidenceChangeTemplate.ts
```diff
@@ -19,14 +19,30 @@ type Events =
...
+  }),
+  selectChild: z
+    .array(z.string())
+    .min(1, { message: 'Velja þarf að lágmarki eitt barn' }),
+  email: z
+    .string()
+    .email('Netfang þarf að vera löglegt, t.d. netfang@netfang.is'),
+  phoneNumber: z
+    .string()
+    .length(7, { message: 'Símanúmer þarf að vera 7 tölustafir' }),
```
VilliGunn, line 35: We should probably still be using `min(7)` here

VilliGunn (approved):
Just one minor comment from me 💯

VilliGunn (dismissed):

libs/application/ui-shell/src/components/FormExternalDataProvider.tsx
```diff
@@ -162,13 +170,15 @@ const FormExternalDataProvider: FC<{
                     activateBeforeSubmitCallback(isChecked)
                   }}
                   checked={value}
+                  hasError={error !== undefined}
```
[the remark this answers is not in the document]
Rikkiabb: Makes sense, I will move it :)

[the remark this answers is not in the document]
Rikkiabb: Actually tested it and the rendering is way off if I put it in the checkbox. 
Don't want to change the checkbox because it makes sense the way it is currently if there is no `Box` around it.

VilliGunn (approved):
Looks good from the Children Residence Change application. I've not tested all applications touched.

dabbeg (approved):

jeremybarbet (approved):
👍

dabbeg (approved):

VilliGunn (approved):

VilliGunn (approved):

PR #2805, opened 2021-02-08
using file upload controller

tthornvik:
# Using file upload controller - health insurance application

## What

Changed to using file upload controller instead of inputFileUpload component on status and children review page

## Checklist:

### DIFF
--- a/libs/application/templates/health-insurance/src/fields/Review/StatusAndChildren.tsx
+++ b/libs/application/templates/health-insurance/src/fields/Review/StatusAndChildren.tsx
@@ -1,15 +1,10 @@
 import React, { FC, useState } from 'react'
 import { formatText, getValueViaPath } from '@island.is/application/core'
-import {
-  Box,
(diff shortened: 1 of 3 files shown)


libs/application/templates/health-insurance/src/fields/Review/StatusAndChildren.tsx
```diff
@@ -113,9 +104,10 @@ const StatusAndChildren: FC<ReviewFieldProps> = ({
                   formatMessage,
                 )}
               />
-              <InputFileUpload
+              <FileUploadController
+                application={application}
                 id="confirmationOfStudies"
-                disabled={!isEditable}
+                maxSize={10000000}
```
dd-sdang, line 110: As we use this value in so many places, we should create a constant for it so that any changes updates everywehere it is used

dd-sdang (approved):
Looks great! :)

PR #2820, opened 2021-02-09
Judicial system: User Control

gudjong:
# User Control

https://app.asana.com/0/1199153462262248/1199582364552602

## What

☝️ 

## Why

Users are currently managed through database seeds, which is not ideal.

## Checklist:


apps/judicial-system/api/src/app/modules/user/dto/createUser.input.ts
```diff
@@ -0,0 +1,40 @@
...
+  @Allow()
+  @Field()
+  readonly mobileNumber!: string
+
+  @Allow()
+  @Field()
+  readonly email!: string
+
+  @Allow()
+  @Field(() => String)
```
oddsson, line 30: Is this correct? Why is this needed here but not in the other fields?

oddsson (approved):

apps/judicial-system/api/src/app/modules/user/dto/createUser.input.ts
gudjong: The reason is the enum type UserRole. The GraphQl schema generator cannot resolve it without help.

PR #2834, opened 2021-02-09
Fixing prefilling of data for address from nationalRegistry

tthornvik:
# Fixing prefilling of address fields

## What

Changing from `.postalCode` to `.address.postalCode` etc. to prefill with the actual data.

## Checklist:

### DIFF
--- a/libs/application/templates/health-insurance/src/forms/HealthInsuranceForm.ts
+++ b/libs/application/templates/health-insurance/src/forms/HealthInsuranceForm.ts
@@ -21,6 +21,7 @@ import {
 import { m } from './messages'
 import { YES, NO, FILE_SIZE_LIMIT } from '../constants'
 import { StatusTypes } from '../types'
+import { Address } from 'libs/api/schema/src/lib/schema'
 import Logo from '../assets/Logo'
 import {
   isEUCountry,
@@ -137,7 +138,9 @@ export const HealthInsuranceForm: Form = buildForm({
               width: 'half',
               disabled: true,
               defaultValue: (application: Application) =>
-                (application.externalData.nationalRegistry?.data as {
+                ((application.externalData.nationalRegistry?.data as {
+                  address?: Address
+                }).address as {
                   streetAddress?: string
                 })?.streetAddress,
             }),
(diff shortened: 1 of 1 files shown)


libs/application/templates/health-insurance/src/forms/HealthInsuranceForm.ts
```diff
@@ -21,6 +21,7 @@ import {
 import { m } from './messages'
 import { YES, NO, FILE_SIZE_LIMIT } from '../constants'
 import { StatusTypes } from '../types'
+import { Address } from 'libs/api/schema/src/lib/schema'
```
dd-sdang, line 24: Thinks this import needs to be something like `@island.is/api/schema`, otherwise linting will fail

libs/application/templates/health-insurance/src/forms/HealthInsuranceForm.ts
```diff
@@ -147,17 +150,21 @@ export const HealthInsuranceForm: Form = buildForm({
               width: 'half',
               disabled: true,
               defaultValue: (application: Application) =>
-                (application.externalData.nationalRegistry?.data as {
+                ((application.externalData.nationalRegistry?.data as {
+                  address?: Address
+                }).address as {
                   postalCode?: string
-                })?.postalCode,
+                })?.postalCode || '',
```
dd-sdang, line 157: Should we have this: `|| ''` or was it accidentally comitted?

libs/application/templates/health-insurance/src/forms/HealthInsuranceForm.ts
```diff
@@ -147,17 +150,21 @@ export const HealthInsuranceForm: Form = buildForm({
...
               title: m.city,
               width: 'half',
               disabled: true,
               defaultValue: (application: Application) =>
-                (application.externalData.nationalRegistry?.data as {
+                ((application.externalData.nationalRegistry?.data as {
+                  address?: Address
+                }).address as {
                   city?: string
                 })?.city,
```
dd-sdang, line 169: I don't know if you we need to define the type for city if address already has a type that contains city. Same with postalCode and streetAddress. Will typescript "complain" anyway?
What I mean is, does it work if we do like this:
`(application.externalData.nationalRegistry?.data as {
    address?: Address
}).address?.city`

tthornvik: Missed that, will fix 👍

libs/application/templates/health-insurance/src/forms/HealthInsuranceForm.ts
tthornvik: If we don't have this we get an error `Cannot convert undefined or null to object` so this is for us when we are testing so that we can move forward in the form. `|| ''` is only after postal code since it is the only one we don't get with our test user

dd-sdang (approved):

PR #2836, opened 2021-02-10
Create eslint rule for ssn

sindrig:
Start with warning for a small grace period.
This closes island-is/infrastructure#488


pshomov (approved):
LGTM

eirikurn (commented):
Sweet!

I wonder how we should handle test kennitalas, of which our repo is full of. Either:

* a list of allowed test kennitalas.
* a bunch of eslint-disable-next-line.
* override eslintrc rules for test files/folders.

sindrig:
@eirikurn out of the 3 cases you mention I'd only want to introduce case 1. But I'd say even better to just don't allow any valid kennitalas!

If we really need valid kennitalas, kennitala.js has a function to create valid, random kennitalas, that we should create on-demand when running the tests.

PR #2843, opened 2021-02-10
Judicial System: Authentication Error Handling

gudjong:
# Authentication Error Handling

https://app.asana.com/0/1199153462262248/1199180537866767

## What

☝️ 

We distinguish between three authentication errors:

- `/?villa=innskraning-ogild` returned by the server when the authentication fails
- `/?villa=innskraning-ekki-notandi` returned by the server when the authentication succeeds but the person logging in is not a user
- `/?villa=innskraning-utrunnin` set by the client when the GraphQL api responds with UNAUTHENTICATED. This typically happens when the authentication expires. However, it will also happen if an unauthenticated user tries to navigate directly to a valid client url. It should be safe to ignore the latter case.

## Why

To better inform the user of what is happening.

## Screenshots / Gifs

## Checklist:


gudjong (commented):
This is a very nice improvement.

oddsson (approved):

PR #2849, opened 2021-02-10
Judicial System: Access Control for Role Registrar

gudjong:
# Access Control for Role Registrar

https://app.asana.com/0/1199153462262248/1199175858230021

## What

☝️ 

The registrar role has a subset of the judge role privileges. The registrar cannot:
- Sign rulings
- Accept / reject cases (happens automatically immediately after signing)
- Send ruling notifications (happens automatically immediately after signing)

## Why

## Checklist:


apps/judicial-system/backend/src/app/modules/case/case.controller.ts
```diff
@@ -108,6 +108,37 @@ const judgeUpdateRule = {
   ],
 } as RolesRule
 
+// Allows judges to update a specific set of fields
```
oddsson, line 111:
```suggestion
// Allows registrars to update a specific set of fields
```

apps/judicial-system/backend/src/app/modules/case/case.controller.ts
```diff
@@ -132,6 +163,14 @@ const judgeTransitionRule = {
   ],
 } as RolesRule
 
+// Allows registrats to receive cases
```
oddsson, line 166:
```suggestion
// Allows registrars to receive cases
```

oddsson (changes requested):
Couple of typos

oddsson (approved):

PR #2868, opened 2021-02-11
Children residence change: pdf download

VilliGunn:
# \<Description\>

https://app.asana.com/0/1199733748699731/1199701171784808

## What

- Added a button to download a pdf version of the contract
- The `pdfUrl` will be fetched when the overview screen is loaded
- Button will be disabled and in a loading state when the url is being fetched
- Button will be disabled if there is an error fetching the url
- There is currently no other error handling, we will handle that in another task

## Why

- So users can see what they are about to sign

## Screenshots / Gifs

Screenshot 2021-02-12 at 09 50 36

## Checklist:


libs/application/templates/children-residence-change/src/fields/Overview/index.tsx
```diff
@@ -19,6 +21,28 @@ const Overview = ({ application }: FieldBaseProps) => {
   const children = extractChildrenFromApplication(application)
   const answers = extractAnswersFromApplication(application)
   const { formatMessage } = useIntl()
+
+  const [createPdfPresignedUrl, { loading: loadingUrl, data }] = useMutation(
+    CREATE_PDF_PRESIGNED_URL,
+    {
+      onError: (e) => console.log('error', e),
```
Rikkiabb, line 28: Do we want to add a TODO comment here or do you feel that a ticket on our board is enough?

libs/application/templates/children-residence-change/src/fields/Overview/index.tsx
```diff
@@ -19,6 +21,28 @@ const Overview = ({ application }: FieldBaseProps) => {
...
+    },
+  )
+
+  useEffect(() => {
+    createPdfPresignedUrl({
+      variables: {
+        input: {
+          id: application.id,
+          type: 'ChildrenResidenceChange',
```
Rikkiabb, line 37: Should we use `libs/application/core/src/types/PdfTypes.ts` and use the enum itself?

libs/application/templates/children-residence-change/src/fields/Overview/index.tsx
```diff
@@ -19,6 +21,28 @@ const Overview = ({ application }: FieldBaseProps) => {
...
+        },
+      },
+    })
+  }, [application.id, createPdfPresignedUrl])
+
+  const pdfUrl =
+    data?.createPdfPresignedUrl?.attachments?.ChildrenResidenceChange
```
Rikkiabb, line 44: Should we rename the data variable to something more explicit?

Rikkiabb (commented):
Nice 👏  Minor things from me

libs/application/templates/children-residence-change/src/fields/Overview/index.tsx
VilliGunn: I've added a ticket to our board. This will be a part of a bigger error handling ticket.

libs/application/templates/children-residence-change/src/fields/Overview/index.tsx
```diff
@@ -1,8 +1,10 @@
-import React from 'react'
+import React, { useEffect } from 'react'
 import { useIntl } from 'react-intl'
+import { useMutation } from '@apollo/client'
 import { FieldBaseProps } from '@island.is/application/core'
-import { Box, Text, AlertMessage } from '@island.is/island-ui/core'
-import { DescriptionText } from '../components'
+import { Box, Text, AlertMessage, Button } from '@island.is/island-ui/core'
+import { CREATE_PDF_PRESIGNED_URL } from '@island.is/application/graphql'
+import { PdfTypes } from '@island.is/application/core'
```
jeremybarbet, line 7: You already have the imported library 2 lines above

libs/application/templates/children-residence-change/src/fields/Overview/index.tsx
```diff
@@ -19,6 +22,30 @@ const Overview = ({ application }: FieldBaseProps) => {
   const children = extractChildrenFromApplication(application)
   const answers = extractAnswersFromApplication(application)
   const { formatMessage } = useIntl()
+
+  const [
+    createPdfPresignedUrl,
+    { loading: loadingUrl, data: response },
+  ] = useMutation(CREATE_PDF_PRESIGNED_URL, {
+    onError: (e) => console.log('error', e),
```
jeremybarbet, line 30: you should use our logging library instead of console.log

jeremybarbet (changes requested):

libs/application/templates/children-residence-change/src/fields/Overview/index.tsx
VilliGunn: 🤦  Shouldn't the linter throw an error when this happens? 🤷

jeremybarbet (approved):

libs/application/templates/children-residence-change/src/fields/Overview/index.tsx
jeremybarbet: I don't think so, because it's not wrong as is. We could probably enforce it with some eslint rules to avoid import duplication declarations

Rikkiabb (approved):
💯

PR #2883, opened 2021-02-12
Judicial System: Only return new cases to prosecutors

gudjong:
# Only return new cases to prosecutors

https://app.asana.com/0/1199153462262248/1193875241514678

## What

☝️ 

## Why

Added security

## Checklist:


apps/judicial-system/backend/src/app/modules/case/interceptors/case.filter.ts
```diff
@@ -0,0 +1,8 @@
+import { CaseState, UserRole } from '@island.is/judicial-system/types'
+
+export function isStateVisibleToRole(
+  state: CaseState,
+  role: UserRole,
+): unknown {
```
addi, line 6: Afhverju skilar þetta unknown?

gudjong: That is unknown :)  I have no idea how this got there.

addi (approved):

PR #2912, opened 2021-02-16
feat: define translations namespace with applications types

jeremybarbet:
## What

- Pick up translation namespace for the required application type
- Won't load multiple namespaces not required for the AS


libs/application/core/src/types/ApplicationTypes.ts
```diff
@@ -9,3 +9,15 @@ export enum ApplicationTypes {
   HEALTH_INSURANCE = 'HealthInsurance',
   CHILDREN_RESIDENCE_CHANGE = 'ChildrenResidenceChange',
 }
+
+export const ApplicationTranslations = {
+  ExampleForm: 'example.application',
+  Passport: null,
+  DrivingLessons: 'dl.application',
+  DrivingLicense: 'dl.application',
```
baering, line 17: I assume this being the same is intentional?

libs/localization/src/lib/useNamespaces.tsx
```diff
@@ -11,8 +11,8 @@ export function useNamespaces(namespaces?: string | string[]) {
   } = useContext(LocaleContext)
 
   useEffect(() => {
-    if (!isEmpty(namespaces)) {
-      loadMessages(namespaces as string | string[])
+    if (namespaces && !isEmpty(namespaces)) {
+      loadMessages(namespaces)
```
baering, line 15: Shouldn't we filter out `null` here rather than changing the signature of `loadMessages` to expect `null[]`?

baering (approved):
Some minor comments. LGTM 👍

libs/application/core/src/types/ApplicationTypes.ts
jeremybarbet: Yes, they are both using the same namespace

libs/localization/src/lib/useNamespaces.tsx
jeremybarbet: Tes that's true, it would be more logic, let me fix that, thanks :+1:

baering (approved):

PR #2915, opened 2021-02-16
Skilavottord - Update base path in frontend and api

dd-sdang:
# Skilavottord - update base path

## What

Updated tha basepath to be `/app/skilvottord` instead of only `/skilavottord` due to route conflict with island.is

## Checklist:


tommigitgaur (approved):
LFTM

PR #2929, opened 2021-02-17
Feature/save organisation on accept

fjandakornid:
# Save organisation on accept

## What

Added makeGraphqlQuery to the sharedTemplateApiService and saved the document-provider application to db on accept.

## Why

We had to make graphql query when accepting the application. Added makeGraphqlQuery to the sharedTemplateApiService.

## Checklist:


libs/application/template-api-modules/src/lib/modules/templates/document-provider-onboarding/document-provider-onboarding.service.ts
```diff
@@ -1,4 +1,7 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
baering, line 1: I would like to avoid the use of `any` where we can.

Until we take the next step and refactor the state machine out of templates and into the template-api-modules we can do something like:

```typescript
interface Contact {
  name?: string
  email: string
  phoneNumber: string
}

// ...
const applicant = get(application.answers, 'applicant') as string
const adminContact = get(
  application.answers,
  'administrativeContact',
) as Contact
const techContact = get(application.answers, 'technicalContact') as Contact
const helpdesk = get(application.answers, 'helpDesk') as Contact
```

libs/application/template-api-modules/src/lib/modules/shared/shared.service.ts
```diff
@@ -72,4 +72,21 @@ export class SharedTemplateApiService {
...
+
+    return fetch(`${baseApiUrl}/api/graphql`, {
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/json',
+        Accept: 'application/json',
+        authorization,
+      },
+      body: JSON.stringify({ query }),
+    })
```
baering, line 90: Nice, thanks for adding this 👍

baering (changes requested):
Minor comment on the use of `any`, other than that LGTM! 💯

libs/application/template-api-modules/src/lib/modules/templates/document-provider-onboarding/document-provider-onboarding.service.ts
fjandakornid: had to cast to `unknown` first

baering (approved):

PR #2930, opened 2021-02-17
Base application system template for party letter application

stefanhar:
# \<Description\>

A base template for developers to work agains

## What

- A raw incomplete template

## Why

-  As a starting point for development

## Screenshots / Gifs

Screenshot 2021-02-17 at 15 21 53

## Checklist:


siggakogk (dismissed):

sindrig (dismissed):

libs/application/templates/party-letter/src/fields/index.ts
```diff
@@ -0,0 +1 @@
+export {}
```
jeremybarbet, line 1: Is that working?

libs/application/templates/party-letter/src/forms/LetterApplicationForm.ts
```diff
@@ -0,0 +1,77 @@
...
+  buildRadioField,
+  buildSection,
+  buildSubmitField,
+  buildTextField,
+  Form,
+  FormModes,
+} from '@island.is/application/core'
+import { m } from './messages'
+
+const partyLetters = ['A', 'B', 'C', 'X', 'H', 'I', 'O', 'P', 'Q', 'T', 'R']
```
jeremybarbet, line 14: Do these letters mean something in regard to the law or something? if so, I would probably move that into an enum, e.g.

```typescript
enum PartyLetters {
  PartyOfAmazing = 'A',
  PartyOfBelieving = 'B'
}
```

and so on, but not sure if that's how it works. In any case, it could be a simpler enum, I guess these letters will be important allong the application flow/process

libs/application/templates/party-letter/src/forms/messages.ts
```diff
@@ -0,0 +1,24 @@
+import { defineMessages } from 'react-intl'
+
+export const m = defineMessages({
+  companySection: {
+    id: 'pl.application:company.section',
```
jeremybarbet, line 5: oooopss, ParentalLeave is already using `pl.application`. I belive if you run `extract-strings`, you will deprecate all the parentalLeave strings inside the existing contentful entry.

What about `ple.application` or `pal.application`, just something different 😅

libs/application/templates/party-letter/src/lib/PartyLetterApplicationTemplate.ts
```diff
@@ -0,0 +1,129 @@
...
+  ApplicationTypes,
+  ApplicationContext,
+  ApplicationRole,
+  ApplicationStateSchema,
+  Application,
+  DefaultEvents,
+} from '@island.is/application/core'
+import * as z from 'zod'
+
+const nationalIdRegex = /([0-9]){6}-?([0-9]){4}/ // TODO: make this use kennitala lib
```
jeremybarbet, line 12: You might want to do it already, since we already have it in the project for example [here](https://github.com/island-is/island.is/blob/main/libs/application/templates/parental-leave/src/lib/dataSchema.ts#L63)

workspace.json
```diff
@@ -1710,6 +1710,28 @@
...
+            "linter": "eslint",
+            "tsConfig": [
+              "libs/application/templates/party-letter/tsconfig.lib.json",
+              "libs/application/templates/party-letter/tsconfig.spec.json"
+            ],
+            "exclude": [
+              "**/node_modules/**",
+              "!libs/application/templates/party-letter/**/*"
+            ]
+          }
```
jeremybarbet, line 1731: Missing the `extract-string` [script](https://docs.devland.is/handbook/libs/localization#message-extraction) to handle your strings into contentful.

jeremybarbet (changes requested):

libs/application/templates/party-letter/src/forms/messages.ts
stefanhar: ple is fine, missed that

libs/application/templates/party-letter/src/lib/PartyLetterApplicationTemplate.ts
stefanhar: True, fixed

workspace.json
stefanhar: Fixed

libs/application/templates/party-letter/src/fields/index.ts
stefanhar: Yes wanted to keep the structure since we expect to use custom fields later on.
We are returning an empty module to prevent errors

libs/application/templates/party-letter/src/forms/LetterApplicationForm.ts
stefanhar: These are letters free to be assigned to a political party for voting.
This array is just a placeholder, we'll get this data from an external service later on

workspace.json
```diff
@@ -1710,6 +1710,34 @@
...
+        },
+        "extract-strings": {
+          "builder": "@nrwl/workspace:run-commands",
+          "options": {
+            "command": "yarn ts-node -P libs/localization/tsconfig.lib.json libs/localization/scripts/extract 'libs/application/core/src/**/*.{js,ts,tsx}'"
```
jeremybarbet, line 1736: Wrong path, should point to your partyLetter AS directory or else to messages.ts if you will only use one file for messages

stefanhar: Fixed

jeremybarbet (approved):
LGTM

sindrig (approved):

PR #2932, opened 2021-02-17
Feature/add citizenship countrycode

dd-sdang:
# Added country code to citizenship field

## What

- Updated the national registry api to return country code as well in the field for citizenship.
- Updated graphql queries that has the citizenship field
- Added lastUpdated selection to graphql query in application system
- Added logging in application system when querying national registry (I use console.log right now as the logging library doesnt work in frontend at the moment)

## Why

The country code is required in the application for health insurance

## Checklist:


libs/api/domains/national-registry/src/lib/graphql/user.resolver.ts
```diff
@@ -28,7 +28,7 @@ export class UserResolver {
     return this.nationalRegistryService.getUser(user.nationalId)
   }
 
-  @ResolveField('citizenship', () => String)
+  @ResolveField('citizenship', () => Citizenship)
   resolveCitizenship(@Parent() user: User): string {
     return user.citizenship.name
```
darrikonn, line 33: Need to change the resolver as well. Shouldn't just return the name

darrikonn (changes requested):
Otherwise looks good!

libs/api/domains/national-registry/src/lib/graphql/user.resolver.ts
dd-sdang: Got it! Updated :)

darrikonn (approved):
Nice

obmagnusson (approved):
LGTM

baering (approved):
LGTM 👍

PR #2946, opened 2021-02-18
Start using configcat feature flags instead of env in service portal

dabbeg:
## What

Integrate Configcat into service portal instead of using environment variables

## Why

Because Configcat is our newly decided feature flagging solution

## Checklist:


apps/service-portal/src/environments/environment.prod.ts
```diff
@@ -13,17 +13,5 @@ export default {
...
-    finance: true,
-    family: true,
-    health: false,
-    education: false,
-    delegation: false,
-    assets: false,
-    drivingLicense: false,
-    documentProvider: false,
-  },
+  configcatSdkKey: 'YcfYCOwBTUeI04mWOWpPdA/KgCHhUk0_k2BdiKMaNh3qA',
```
darrikonn, line 16: Change to an environment variable.

apps/service-portal/src/environments/environment.ts
```diff
@@ -11,17 +11,5 @@ export default {
...
-    finance: true,
-    family: true,
-    health: false,
-    education: true,
-    delegation: false,
-    assets: false,
-    drivingLicense: false,
-    documentProvider: true,
-  },
+  configcatSdkKey: 'YcfYCOwBTUeI04mWOWpPdA/KgCHhUk0_k2BdiKMaNh3qA',
```
darrikonn, line 14: ditto

apps/service-portal/src/hooks/useRoutes/useRoutes.ts
```diff
@@ -3,12 +3,14 @@ import { useEffect } from 'react'
 import { ServicePortalModule } from '@island.is/service-portal/core'
 import { Action, ActionType } from '../../store/actions'
 import { useModuleProps } from '../useModuleProps/useModuleProps'
+import useModules from '../useModules/useModules'
```
darrikonn, line 6: Can you change this to a named export, so the name won't differ between files.

darrikonn (changes requested):

apps/service-portal/src/environments/environment.prod.ts
dabbeg: Yeah my thinking was to have it here in clear text because this is not a secret but there might be a benefit having this in an environment variable so we don't need to repeat it through the projects.
I will change it to env variable.

apps/service-portal/src/hooks/useRoutes/useRoutes.ts
dabbeg: I was following the same naming pattern for most of the other hooks. Is this some standard that we use?

apps/service-portal/src/environments/environment.prod.ts
dabbeg: Ah no the reason is I can't get this out of an environment variable as `process.env.X` is evaluated in the browser.
So I at least don't know a good way to solve that easily.

apps/service-portal/src/environments/environment.ts
dabbeg: ditto

apps/service-portal/src/environments/environment.prod.ts
Thorrik58: This hasnt been solved yet as far as I know.

apps/service-portal/src/hooks/useRoutes/useRoutes.ts
darrikonn: But the hook above is a named export. This mainly comes from https://basarat.gitbook.io/typescript/main-1/defaultisbad and is the reason why we almost always reexport our default imports in an index file.
```typescript
// index.ts
export { default as useModules } from './useModules'
```

```typescript
// apps/service-portal/src/hooks/useRoutes/useRoutes.ts
import { useModules } from '../useModules'
```

Thorrik58 (approved):
LGTM, have all member of the SP been invited to the dashboard or is there a member limit?

apps/service-portal/src/environments/environment.prod.ts
darrikonn: Needs to be set at build time via webpack.

dabbeg: Yes because configcat sdk key does not change between environments we can build it into the bundle with something like webpack DefinePlugin.

I am not sure it is a good idea though, because we are adding the CI environment as a dependency to build the project with correct variables. I feel weird about it.

@sindrig thoughts?

dabbeg:
> LGTM, have all member of the SP been invited to the dashboard or is there a member limit?

When you say dashboard are you talking about the configcat dashboard? If so then no they have not been invited, there are some decisions around how the process will be for developers to enable/disable features in the Dashboard. @sindrig can go into more detail.

apps/service-portal/src/environments/environment.prod.ts
sindrig: The SDK indeed changes between environments, we've got different keys for dev, staging and prod! It's available to all pods via the `CONFIGCAT_SDK_KEY` environment variable.

dabbeg: :facepalm: ofcourse there are different keys between environments.

Okay then the solution is easy but ugly.

apps/service-portal/src/hooks/useRoutes/useRoutes.ts
dabbeg: Yeah they were mixed this was the only hook with named export and all the others had default export so I opted to be consistent.

But is this is a standard I will change this to named export. Someone else needs to refactor the other hooks and put in the interface

apps/service-portal/src/environments/environment.prod.ts
```diff
@@ -13,17 +15,10 @@ export default {
...
-    finance: true,
-    family: true,
-    health: false,
-    education: false,
-    delegation: false,
-    assets: false,
-    drivingLicense: false,
-    documentProvider: false,
-  },
+  featureFlagSdkKey:
```
sindrig, line 18: Erm... really? can't we somehow get this from the environment? You're at least missing `www.island.is`... but I don't want to maintain a list of possible ingress hosts in code...

libs/feature-flags/src/lib/configcat.ts
```diff
@@ -14,17 +14,19 @@ export class Client implements FeatureFlagClient {
...
-        resolvedSdkKey,
-        ccConfig,
-      )
-    } else {
-      this.configcat = require('configcat-js').createClient(
-        resolvedSdkKey,
-        ccConfig,
-      )
-    }
+    // TODO find a way to conditionally require this without
```
sindrig, line 17: Hmm I'm pretty sure I tried this for both webpack and server side, this was what @eirikurn came up for me. IMO we should figure this out now.

libs/feature-flags/src/lib/feature-flags.ts
```diff
@@ -1,6 +1,12 @@
 import { Client } from './configcat'
 import { FeatureFlagClient, FeatureFlagClientProps } from './types'
 
+export const sdkKeys = {
```
sindrig, line 4: I don't want these in code. They're provided in the environment :grimacing:

sindrig (changes requested):

apps/service-portal/src/environments/environment.prod.ts
dabbeg: I don't like this either... I am up for any ideas on how we could do this.
The service portal is a react project which uses the docker-static configuration.
I am not aware of any way to get values from environment when working with docker-static.

Is there something that I am missing?

sindrig: Some kind of env.json that we can generate on container startup via envsubst? And then fetch it on application startup? Either way it's not acceptable to *not* have these capabilities.

libs/feature-flags/src/lib/configcat.ts
dabbeg: The errors that come up are:
```
WARNING in /home/dabbeg/repos/island.is/node_modules/got/source/request-as-event-emitter.js 72:18-25
Critical dependency: require function is used in a way in which dependencies cannot be statically extracted

WARNING in /home/dabbeg/repos/island.is/node_modules/keyv/src/index.js 18:14-40
Critical dependency: the request of a dependency is an expression

ERROR in /home/dabbeg/repos/island.is/node_modules/got/source/utils/get-body-size.js
Module not found: Error: Can't resolve 'fs' in '/home/dabbeg/repos/island.is/node_modules/got/source/utils'

ERROR in /home/dabbeg/repos/island.is/node_modules/got/source/utils/timed-out.js
Module not found: Error: Can't resolve 'net' in '/home/dabbeg/repos/island.is/node_modules/got/source/utils'

ERROR in /home/dabbeg/repos/island.is/node_modules/tunnel/lib/tunnel.js
Module not found: Error: Can't resolve 'net' in '/home/dabbeg/repos/island.is/node_modules/tunnel/lib'

ERROR in /home/dabbeg/repos/island.is/node_modules/defer-to-connect/dist/index.js
Module not found: Error: Can't resolve 'tls' in '/home/dabbeg/repos/island.is/node_modules/defer-to-connect/dist'

ERROR in /home/dabbeg/repos/island.is/node_modules/tunnel/lib/tunnel.js
Module not found: Error: Can't resolve 'tls' in '/home/dabbeg/repos/island.is/node_modules/tunnel/lib'
ℹ ｢wdm｣: Failed to compile.
```

I was able to supress the errors by adding fs, net, tls to the webpack node config but that was not a very reliable long term solution. Maybe @eirikurn knows how to fix this?

libs/feature-flags/src/lib/feature-flags.ts
dabbeg: :grimacing: 
same problem as in the first thread, lets have this discussion there.

apps/service-portal/src/environments/environment.prod.ts
dabbeg: I thought this was on your backlog, but I didn't find an issue for this.
I created #2972

dabbeg:
I am closing the PR due to #2972. I will reopen when this has been resolved

PR #2947, opened 2021-02-18
fix: throw error when script fails

jeremybarbet:
## What

- Should break the CI when failing during NX schemas generations
- `process.exit(err.code)` err.code was undefined, I believe that's why it wasn't failing


jeremybarbet:
Closing in favor of https://github.com/island-is/island.is/pull/3059

PR #2967, opened 2021-02-19
Prevent infinite loop in rich text renderer

Ingimarsson:
# Prevent infinite loop in rich text renderer

## What

Some articles currently return error 500 because of an infinite loop (for example https://island.is/umgengni).

This is because `null` has type `object` so we need to add a check for that.

## Checklist:


barabrian (dismissed):
LGTM

barabrian (approved):

PR #2968, opened 2021-02-19
Children residence change/file signature

Rikkiabb:
# \<Description\>
Adds two new endpoints to support electronic signature for the application system. 
1. Is requesting a file signature from dokubit which returns a controlCode to display to the user and a documentToken. 
2. Uses the documentToken to see if user has authenticated and if so uploads the signed document to s3.

TODO: Get environment variables from Dokobit and add it to env.

## Why
Needed for children residence change application.

## Checklist:


apps/application-system/api/src/app/modules/application/files/file.service.spec.ts
```diff
@@ -1,114 +1,191 @@
...
+  }
+  const applicantsPhoneNumber = '111-2222'
+  const parentBPhoneNumber = '222-1111'
+  const applicantName = 'Test name'
+  const applicantSsn = '0113215029'
+  const applicationId = '1111-2222-3333-4444'
+
+  const createApplication = (
+    answers: object = {
+      email: 'email@email.is ',
```
VilliGunn, line 26:
```suggestion
      email: 'email@email.is',
```

apps/application-system/api/src/app/modules/application/files/file.service.spec.ts
```diff
@@ -1,114 +1,191 @@
...
+import { NotFoundException } from '@nestjs/common'
 
 describe('FileService', () => {
   let service: FileService
+  let signingService: SigningService
+  const signingServiceRequestSignatureResponse = {
+    controlCode: 'code',
+    documentToken: 'token',
+  }
+  const applicantsPhoneNumber = '111-2222'
```
VilliGunn, line 18: I am wondering if we want to have all the test data in the same place. I get that the data that is used in more than place has been extracted to a constant but in my opinion it would be super nice to have all the data in one place and constructed with an object:

```
const parentA: {
    phoneNumber: '111-2222',
    name: 'Test name',
}
const parentB: {
    phoneNumber: '222-1111',
    name: 'Parent B name',
}
```
etc.

apps/application-system/api/src/app/modules/application/files/utils/aws.ts
```diff
@@ -0,0 +1,55 @@
+import * as AWS from 'aws-sdk'
+import { environment } from '../../../../../environments'
```
VilliGunn, line 2: This feels super weird to see in a repo full of aliases. Should we create an alias if there isn't one for this?

apps/application-system/api/src/app/modules/application/files/utils/aws.ts
```diff
@@ -0,0 +1,55 @@
...
+
+const s3 = new AWS.S3()
+const oneMinute = 60
+const bucket = environment.fsS3Bucket ?? 'development-legal-residence-change'
```
VilliGunn, line 6: Do we want to hardcode the default here? Don't we want this to fail on prod for example if the env variable is missing?

apps/application-system/api/src/app/modules/application/files/utils/pdf.ts
```diff
@@ -4,7 +4,7 @@ import {
   ParentResidenceChange,
   PersonResidenceChange,
 } from '@island.is/application/templates/children-residence-change'
-import { Constants } from './constants'
+import { PdfConstants as Constants } from './constants'
```
VilliGunn, line 7: Why not just use `PdfConstants` instead of renaming to `Constants` ?

VilliGunn (commented):
Just some minor comments from me 💯

apps/application-system/api/src/app/modules/application/files/file.service.spec.ts
Rikkiabb: 🙈

apps/application-system/api/src/app/modules/application/files/utils/aws.ts
Rikkiabb: Totally agree, this is done this way in 35 places in the repo, so I am guessing there is some reason for why there is not an alias used for the environment. Feel that this is out of scope for this PR since this looks like the pattern that is being created for services in BE.

Rikkiabb: Just noticed that this is also fixed in this PR: https://github.com/island-is/island.is/pull/2973/files/9241da83202f9ed51dce3cf43ac0cbaf89e309e2..37882b8f601e6332ae7caf3837df9922186a591e

apps/application-system/api/src/app/modules/application/files/file.service.ts
```diff
@@ -40,97 +41,95 @@ export class FileService {
...
     }
 
-    const expiry = answers.expiry as string
-
-    return { parentA, parentB, childrenAppliedFor, expiry }
+    await this.signingService
+      .getSignedDocument(DokobitFileName[type], documentToken)
+      .then((file) => {
+        const s3FileName = `${BucketTypePrefix[type]}/${application.id}.pdf`
+        uploadFile(Buffer.from(file, 'binary'), bucket, s3FileName)
```
eirikurn, line 83: We should use Buffers instead of binary strings whenever possible.

apps/application-system/api/src/app/modules/application/files/file.service.ts
```diff
@@ -40,97 +41,95 @@ export class FileService {
...
 
-    await this.s3
-      .upload(uploadParams)
-      .promise()
-      .catch(() => {
-        return null
-      })
+    const s3FileName = `${BucketTypePrefix[type]}/${applicationId}.pdf`
+    const s3File = await getFile(bucket, s3FileName)
+    const fileContent = s3File.Body?.toString('binary')
```
eirikurn, line 120: We should use buffers instead of binary strings whenever possible.

apps/application-system/api/src/app/modules/application/application.controller.ts
```diff
@@ -572,4 +574,63 @@ export class ApplicationController {
...
+
+    const { updatedApplication } = await this.applicationService.update(
+      application.id,
+      {
+        attachments: {
+          ...application.attachments,
+          fileSignature: {
+            controlCode: controlCode,
+            documentToken: documentToken,
+          },
```
eirikurn, line 608: For now, I would rather you store this under `application.externalData.fileSignature`.

apps/application-system/api/src/app/modules/application/files/utils/aws.ts
```diff
@@ -0,0 +1,58 @@
...
+    ContentType: 'application/pdf',
+    Body: content,
+  }
+
+  await s3
+    .upload(uploadParams)
+    .promise()
+    .catch((err) => {
+      return null
+    })
```
eirikurn, line 39: Why swallow errors?

apps/application-system/api/src/app/modules/application/files/utils/aws.ts
```diff
@@ -0,0 +1,58 @@
...
+    Bucket: bucket,
+    Key: fileName,
+    Expires: oneMinute * 120, // TODO: Select length for presigned url's in island.is
+  }
+
+  return await new Promise((resolve, reject) => {
+    s3.getSignedUrl('getObject', presignedUrlParams, (err, url) => {
+      err ? reject(err) : resolve(url)
+    })
+  })
```
eirikurn, line 57: Why not getSignedUrlPromise?

apps/application-system/api/src/app/modules/application/files/utils/aws.ts
```diff
@@ -0,0 +1,58 @@
...
+): Promise<AWS.S3.GetObjectOutput> {
+  const downloadParams = {
+    Bucket: bucket,
+    Key: fileName,
+  }
+  return await new Promise((resolve, reject) => {
+    s3.getObject(downloadParams, (err, res) => {
+      err ? reject(err) : resolve(res)
+    })
+  })
```
eirikurn, line 17: Why not getObject(downloadParams).promise()?

apps/application-system/api/src/app/modules/application/files/file.service.spec.ts
```diff
@@ -1,122 +1,215 @@
...
+          },
+        },
+        SigningService,
         { provide: APPLICATION_CONFIG, useValue: config },
       ],
     }).compile()
 
+    jest
+      .spyOn(aws, 'getFile')
+      .mockImplementation(() => Promise.resolve({ Body: 'body' }))
```
eirikurn, line 101: Huh. Does this work?

apps/application-system/api/src/app/modules/application/files/utils/aws.ts
```diff
@@ -0,0 +1,58 @@
+import * as AWS from 'aws-sdk'
+
+const s3 = new AWS.S3()
```
eirikurn, line 3: Should this be a Nest module/service? To make it easier/cleaner to inject and stub.

libs/api/domains/application/src/lib/application.resolver.ts
```diff
@@ -142,4 +144,23 @@ export class ApplicationResolver {
...
+    )
+  }
+
+  @Mutation(() => Application, { nullable: true })
+  uploadSignedFile(
+    @Args('input') input: UploadSignedFileInput,
+    @CurrentUser() user: User,
+  ): Promise<Application> {
+    return this.applicationService.uploadSignedFile(input, user.authorization)
+  }
```
eirikurn, line 165: Where will this mutation be triggered? Shouldn't the user _just_ start the signing process, and when that's done, we somehow upload the file automatically?

apps/application-system/api/src/app/modules/application/files/utils/pdf.ts
```diff
@@ -14,12 +14,12 @@ export async function generateResidenceChangePdf(
...
-      top: Constants.VERTICAL_MARGIN,
-      bottom: Constants.VERTICAL_MARGIN,
-      left: Constants.HORIZONTAL_MARGIN,
-      right: Constants.HORIZONTAL_MARGIN,
+      top: PdfConstants.VERTICAL_MARGIN,
+      bottom: PdfConstants.VERTICAL_MARGIN,
+      left: PdfConstants.HORIZONTAL_MARGIN,
+      right: PdfConstants.HORIZONTAL_MARGIN,
     },
   })
```
eirikurn, line 24: I dream about having a PDF generation service, which supports multiple types of PDFs, accepts data through a POST body and renders the PDF either with react-pdf or plain react website that gets saved to PDF with the puppeteer.

Maybe one day :)

eirikurn (commented):
Generally, this could have more guards, since the order of operations is fairly brittle, also make sure it's only used in specific types of applications and signed by the right users.

Going forward, I'd really like to refactor this to an external data + template module system. We'll discuss that later.

apps/application-system/api/src/app/modules/application/application.module.ts
```diff
@@ -50,8 +51,13 @@ if (process.env.INIT_SCHEMA === 'true') {
...
   providers: [
     ApplicationService,
-    UploadProcessor,
     FileService,
+    UploadProcessor,
+    {
+      provide: SIGNING_OPTIONS,
+      useValue: environment.signingOptions,
+    },
+    SigningService,
```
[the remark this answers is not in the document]
gudjong: This one is on me and not really in the scope of this PR. I was not aware of the `register` approach when implementing the library. It is on my todo list to change this.

apps/application-system/api/src/app/modules/application/files/file.service.ts
Rikkiabb: Yes agree. 
When implementing this I tried that approach and when you rightfully poked at this then I tried at again but no luck 😭 
The problem I think is the way dokobit works. If I send anything else then a binary string to them then I get an error from their side or and empty document. I talked to @gudjong  about this since he implemented the signing service and he had the same experience. If you have any suggestions to try out them I am all ears. 🙏

libs/api/domains/application/src/lib/application.resolver.ts
Rikkiabb: Yes this flow is weird. So currently it is like this:
1. User requests for file signature. We send back a control code (pin) which is displayed to the user and a document token. 
2. When the user receives these items we send another request which calls the signing service with that document token to check if the signature was successful. In the meantime the user approves/rejects electronic signature on their phone.
It would be ideal that this was one call from the client but since we have to return the control code and then know if the file upload was successful in the client then I implemented this as two calls. This is the same way as it is implemented in the judicial system. 
We could go for server sent event here. But not sure how that works with a graphql layer or if that is a overkill. Any suggestions/thoughts?

apps/application-system/api/src/app/modules/application/files/file.service.spec.ts
Rikkiabb: Yup do you want to go for another approach?

apps/application-system/api/src/app/modules/application/files/utils/aws.ts
Rikkiabb: Fixed ✅  but want to pick your 🧠. Do you have any getlines that you follow when an utility is a service vs a module vs utility functions in the monorepo?

Rikkiabb:
> Generally, this could have more guards, since the order of operations is fairly brittle, also make sure it's only used in specific types of applications and signed by the right users.


All for adding guards here. To make sure we are on the same page. You would like to have three guards.
1. Order of operations
2. Only for specific type of applications
3. Signed by the correct users

Or are you thinking a single guard for all of this? And is there something I am missing? 😊

apps/application-system/api/src/app/modules/application/files/file.service.ts
eirikurn: I'm just talking about it on our end. 

The dokubit nest service is doing `Base64.btoa(string)` before sending the document to their API. So instead of us converting buffers to binary strings to base64, we can just pass buffers around and then do `buffer.toString('base64')` which should at least save one copy of the file in memory, and the base64 encoding probably runs in native code.

apps/application-system/api/src/app/modules/application/files/file.service.spec.ts
eirikurn: Just surprised at this Jest black-magic, since I thought this module had a separate "aws" object from what is imported elsewhere.

If it works, then its fine, otherwise I would have suggested overriding it in a Nest TestingModule.

apps/application-system/api/src/app/modules/application/files/utils/aws.ts
eirikurn: My thinking is if it makes sense to dependency inject it, have it configured externally, or if I want a "simple" way to override/stub it in tests, then we should inject it using NestJS.

Usually applies to stuff that has external dependencies or side effects that we want to configure or stub in tests.

While pure utility functions don't necessarily need to be part of the NestJS DI container.

libs/api/domains/application/src/lib/application.resolver.ts
eirikurn: Ok, thanks for the explanation. I think we can revisit this later.

eirikurn:
No, that's great. Your call if it makes sense to create Nest guards for this or just add `if` statements in the right places.

apps/application-system/api/src/app/modules/application/files/file.service.ts
Rikkiabb: Yes totally agree on paper. Tried this out at first and then did it again now to make sure I wasn't doing something silly. This is the response I get from Dokobit when I try that: 
`Uploaded document is not compliant with PDF standard` 🤷

Rikkiabb:
I have added guards for order of operations and application types. 
I added if statements because I don't have access to the application as a whole in the guards unless I would look it up in the db using the id and I don't want to do a double db look up for each api call on these endpoints. 
I had some trouble figuring out how I would be able to check who are the right users to sign. I currently have who owns the application and then parent B will be an assignee, but I would have to validate those ssn's against something to check if they have access to sign the files, but not sure what this something should be 🤷  Am I way off here or were you thinking something else @eirikurn?

apps/application-system/api/src/app/modules/application/files/file.service.ts
eirikurn: Whaaaaaat??? 🤯

apps/application-system/api/src/app/modules/application/application.controller.ts
```diff
@@ -558,6 +560,8 @@ export class ApplicationController {
   ): Promise<ApplicationResponseDto> {
     const { type } = input
 
+    this.fileService.validateApplicationType(application.typeId)
```
eirikurn, line 563: Why not put the guards inside the corresponding functions? 😶😊

eirikurn (approved):
Sweet! LGTM. Just minor comments that you may address in later PRs if you want.

apps/application-system/api/src/app/modules/application/files/file.service.ts
```diff
@@ -40,97 +46,141 @@ export class FileService {
...
-    fileName: string,
-  ): Promise<string> {
-    const uploadParams = {
-      Bucket: bucket,
-      Key: fileName,
-      ContentEncoding: 'base64',
-      ContentDisposition: 'inline',
-      ContentType: 'application/pdf',
-      Body: buffer,
+  validateApplicationType(applicationType: string) {
```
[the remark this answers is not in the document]
Rikkiabb: This is a very valid point, my feeling is iterate when we have the second pdf contract. Since then we have a more mature vision of this.
My current view of this and I hoped that this reflected that each applicaiton can only create a single pdf and then the pdf type enum should only have values that are reflected in the application type enum and you need to add to the pdf enum to be able to have a pdf from that applicaiton. 
I feel that it will be very valuable to get your input once the next pdf is created and if we will create that then I will make sure you are requested for that review.
Resolving this for now and we will iterate once we get there :)

apps/application-system/api/src/app/modules/application/application.controller.ts
Rikkiabb: I actually started that way and then I thought about who should have this responsibility and decided on this should be on a higher level. Up for discussion imo 😄

VilliGunn (approved):
Nothing further from me after this great thorough review from @eirikurn 💯

PR #2971, opened 2021-02-22
fix: update immer for security report

jeremybarbet:
## What

- `yarn run security-audit` repors high vulnerabilities regarding immer.
- We have warnings for two packages, storybook and slate/* both have open issues https://github.com/storybookjs/storybook/issues/13961 and https://github.com/ianstormtaylor/slate/pull/4050, it should be a matter of time a patch version is published, using resolutions shouldn't impact the usage of these two packages

### DIFF
--- a/package.json
+++ b/package.json
@@ -356,11 +356,12 @@
   },
   "resolutions": {
     "@testing-library/dom": "7.29.1",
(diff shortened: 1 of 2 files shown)


sindrig (approved):
LGTM

PR #2975, opened 2021-02-22
chore(application-system): refactor and improve parental leave submit

baering:
## What

- Add and test util functions for edge cases when transforming application to parental leave to send to VMST
- Include a new data provider for family members
- Fix an issue in formatText

## Why

To improve how application is transformed into parental leave object that is sent to VMST

### New data provider for family members
image

image

## Checklist:


libs/application/template-api-modules/tsconfig.spec.json
```diff
@@ -1,6 +1,7 @@
 {
   "extends": "./tsconfig.json",
   "compilerOptions": {
+    "esModuleInterop": true,
```
baering, line 4: To be able to `import get from 'lodash/get'` from unit test

libs/application/core/src/types/Form.ts
```diff
@@ -5,7 +5,8 @@ import { MessageDescriptor } from 'react-intl'
 import { BoxProps } from '@island.is/island-ui/core'
 import { Application } from './Application'
 
-export type StaticText = (MessageDescriptor & { values?: object }) | string
+export type StaticTextObject = MessageDescriptor & { values?: object }
```
jeremybarbet, line 8: Would be nicer to use `Record<string, unknown>` or probably somewhere a `RecordObject` exists somewhere iirc

libs/application/data-providers/src/providers/FamilyInformationProvider.ts
```diff
@@ -0,0 +1,53 @@
...
+
+        return Promise.resolve(response.data.nationalRegistryFamily)
+      })
+      .catch(() => {
+        return this.handleError()
+      })
+  }
+
+  handleError() {
+    return Promise.resolve([])
```
jeremybarbet, line 38: No errors handlings?

libs/application/template-api-modules/src/lib/modules/templates/parental-leave/parental-leave.utils.ts
```diff
@@ -77,6 +75,88 @@ const getEmployer = (
...
+    if (!spouse) {
+      throw new Error(
+        'transformApplicationToParentalLeaveDTO: Cannot find spouse. No family member with this relation.',
+      )
+    }
+
+    return spouse.nationalId
+  }
+
+  return otherParentId
```
jeremybarbet, line 113: Is it fine to return an empty string if nothing can be found?

libs/application/templates/parental-leave/src/lib/messages.ts
```diff
@@ -340,6 +340,16 @@ export const parentalLeaveFormMessages: MessageDir = {
       description:
         'In order to apply for this application we need your email and phone number',
     },
+    familyInformationTitle: {
+      id: 'pl.application:familyinformation.title',
+      defaultMessage: 'Upplýsingar um fjölskyldumeðlimi',
+      description: 'Your family information',
```
jeremybarbet, line 346: Just for a reminder, the description is not used to pass the English translation, but more for the translations people to know what the message it is referring to. The description won't be used to populate the Contentful English field but just used to be displayed as the "description" section

jeremybarbet (commented):
A few minor comments, overall really nice cleanup and PR 👍

libs/application/template-api-modules/src/lib/modules/templates/parental-leave/parental-leave.utils.ts
baering: Yes - I think so.

The otherParentId is required in the API payload but optional in the application since a primary parent is not required to enter the national registry id of the other parent.

libs/application/core/src/types/Form.ts
baering: Good catch. I gave it a go and it resulted in some weird ts errors in other places that I'll tackle some other time.

libs/application/data-providers/src/providers/FamilyInformationProvider.ts
baering: The `onProvideError` will get called with

```
{
  date: new Date(),
  reason: result,
  status: 'failure',
  data: result,
}
```

Where `result` is whatever handleErrors resolves to.

But I'll include logging when this happens.

madebynoam (approved):
Lgtm!

obmagnusson (approved):
LGTM

PR #2982, opened 2021-02-22
Feature/simplified frontpage

kksteini:
# What and why?

Created a new content type within contentful that merges together elements unique to the frontpage. This makes managing the content easier, keeps it all in one place and lessens the amount of API calls made to the contentful servers.

# How?

Introduced new queries, models and dto's to the API that deal with a new content type called "Frontpage" in contentful.


apps/web/screens/queries/Frontpage.ts
```diff
@@ -0,0 +1,52 @@
+import gql from 'graphql-tag'
+
+export const GET_FRONTPAGE_QUERY = gql`
+  query GetFrontpage($input: GetFrontpageInput!) {
+    getFrontpage(input: $input) {
```
stefanhar, line 5: Maybe we should add `id` field here to better support cache:
https://www.apollographql.com/docs/react/caching/cache-configuration/#default-identifier-generation

libs/api/domains/cms/src/lib/models/frontpage.model.ts
```diff
@@ -0,0 +1,32 @@
...
+
+  @Field(() => Namespace, { nullable: true })
+  namespace?: Namespace
+
+  @Field(() => [LifeEventPage])
+  lifeEvents?: Array<LifeEventPage>
+}
+
+export const mapFrontpage = ({ fields }: IFrontpage): Frontpage => ({
+  title: fields.title ?? '',
```
stefanhar, line 27: This should include id for cache

libs/api/domains/cms/src/lib/cms.contentful.service.ts
```diff
@@ -568,6 +569,20 @@ export class CmsContentfulService {
...
+      ['content_type']: 'frontpage',
+      include: 10,
+      order: '-sys.createdAt',
+    }
+
+    const result = await this.contentfulRepository
+      .getLocalizedEntries<types.IFrontpageFields>(lang, params)
+      .catch(errorHandler('getFrontpage'))
+
+    return result.items.map(mapFrontpage)[0]
```
stefanhar, line 583: Maybe we should add a page identifier field like we do on the "Generic Overview Page" that is more robust

stefanhar (changes requested):

apps/web/screens/Home.tsx
```diff
@@ -44,28 +35,19 @@ import { FRONTPAGE_NEWS_TAG_ID } from '@island.is/web/constants'
...
   const { activeLocale, t } = useI18n()
   const { globalNamespace } = useContext(GlobalContext)
   const n = useNamespace(namespace)
   const gn = useNamespace(globalNamespace)
   const { linkResolver } = useLinkResolver()
 
-  if (!lifeEvents || !lifeEvents.length) {
+  if (!page.lifeEvents || !page.lifeEvents.length) {
     return null
   }
```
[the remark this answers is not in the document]
kksteini: Agree

kksteini:
New query sanity check: 
image

libs/api/domains/cms/src/lib/cms.contentful.service.ts
kksteini: May I ask why?
What does that add?

stefanhar (approved):
LGTM

eirikurn (approved):
Only reviewed mocks. We should configure the right codeowner for each mock domain.

PR #2984, opened 2021-02-22
New validation of status and former country

tthornvik:
# New validation of status and former country

## What

Added validator for status on status and children page
Added validators on former country page:
- registration
- country
- confirmation of residency 
- personal Id
- entitlement
- entitlement reason

## Checklist:


libs/application/templates/health-insurance/src/lib/answerValidators.ts
```diff
@@ -0,0 +1,132 @@
...
+    if (
+      formerInsurance.registration !== YES &&
+      formerInsurance.registration !== NO
+    ) {
+      const field = `${FORMER_INSURANCE}.registration`
+      const buildError = buildValidationError(field)
+      return buildError('You must select one of the above', field)
+    }
+
+    // Check that country is a string and not empty
```
dd-sdang, line 74: Hm realised that we aren't really checking if it is a string, only if it is empty. So we should update this comment

dd-sdang (approved):
Looks good, just a few comments that we should update in next branch

PR #2992, opened 2021-02-22
Fix cleanup of feature deployments

sindrig:
Feature deployments strip underscores, and limit the length of the branch


busla (approved):
LGTM although not familiar with the codebase. Looks like a cleanup of a hardcoded feature branch?

PR #2993, opened 2021-02-22
Crc/custom component multi field

VilliGunn:
# \<Description\>

https://app.asana.com/0/1199733748699731/1199914212397854

- Added an option to have multiple named fields in one custom component instead of array of inputs.
- All form errors are now accessible to the component rather than just one error per comonent.

## What

- Added `childInputIds` to the `buildCustomField` function
- In the `extractAnswersToSubmitFromScreen` we now check if the `screen.type` is custom and if the `childInputIds` is available. If so then we use the array instead of the `screen.id`

## Why

To have better control of the answers in custom components with multiple field inputs.

## Screenshots / Gifs

Screenshot 2021-02-23 at 14 00 25

## Checklist:


libs/application/templates/children-residence-change/src/fields/ContactInfo/index.tsx
```diff
@@ -30,22 +35,24 @@ const ContactInfo = ({ error, application, field }: FieldBaseProps) => {
       </Box>
       <Box marginTop={5}>
         <Controller
-          name="email"
-          defaultValue={getValue('contactInfo')?.[0]}
+          name={emailId}
+          defaultValue={getValue(phoneNumberId)}
```
Rikkiabb, line 39: Should this be `emailId` ?

libs/application/templates/children-residence-change/src/fields/ContactInfo/index.tsx
```diff
@@ -30,22 +35,24 @@ const ContactInfo = ({ error, application, field }: FieldBaseProps) => {
...
+                ref={register}
+                id={emailId}
+                name={emailId}
                 backgroundColor="blue"
-                type="text"
+                type="email"
                 label={formatMessage(otherParent.inputs.emailLabel)}
                 value={value}
-                hasError={!!error}
+                hasError={!!emailError}
```
Rikkiabb, line 50: Do we have to use not not here? Wondering if we know the type `undefined | string` and can then check if it is undefined or do we ever get an empty string?

libs/application/templates/children-residence-change/src/fields/ContactInfo/index.tsx
```diff
@@ -6,15 +6,20 @@ import { Controller, useFormContext } from 'react-hook-form'
 import { otherParent } from '../../lib/messages'
 import { extractParentFromApplication } from '../../lib/utils'
 
-const ContactInfo = ({ error, application, field }: FieldBaseProps) => {
-  const { id } = field
+const emailId = 'email'
+const phoneNumberId = 'phoneNumber'
+
+export const contactInfoIds = [emailId, phoneNumberId]
```
Rikkiabb, line 12: Like the approach that the component "owns" the id's

Rikkiabb (commented):
Small questions from me :) Nice stuff 👏

libs/application/templates/children-residence-change/src/fields/ContactInfo/index.tsx
VilliGunn: I though about it but decided to go with the same approach as all the wrapper controller comopnents - same name for the controller and the input fields:
https://github.com/island-is/island.is/blob/main/libs/shared/form-fields/src/lib/InputController/InputController.tsx#L143

libs/application/templates/children-residence-change/src/fields/ContactInfo/index.tsx
VilliGunn: No we shouldn't get an empty string here. Always just undefined or some error value

Rikkiabb (approved):
Nice stuff 👏

jeremybarbet (approved):
LGTM

PR #2977, opened 2021-02-22
Crc/localization contentful sync

VilliGunn:
# \<Description\>

- Added the `extract-strings` to the Children residence change application workspace
- Added a new translation string namespace
- Fixed duplicate string id

## Why

We need this script to be able to sync translation string with Contentful

## Checklist:

- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] Formatting passes locally with my changes
- [x] I have rebased against main before asking for a review

PR #2998, opened 2021-02-23
fix: contentful markdown fixes

jeremybarbet:
## What

- Line breaks were trim off by formatjs extract script, we need to escape them in order to send them over to Contentful to handle the markdown format


jeremybarbet:
Cherry-picked in here directly https://github.com/island-is/island.is/pull/2977

Issue #2972, opened 2021-02-22
Implement a way to use environment variables with docker-static deployments


dabbeg:
Pure react projects that use docker-static have no way of getting variables from the environment.

Examples of these projects are service-portal and application-system.

PR #2999, opened 2021-02-23
Extracting environment and putting it into index.html for docker-static

dabbeg:
# \<Description\>

Resolves: #2972

Checkout index.html at:
* [https://dynamic-environment-for-docker-static-beta.dev01.devland.is/minarsidur/](https://dynamic-environment-for-docker-static-beta.dev01.devland.is/minarsidur/)
* [https://dynamic-environment-for-docker-static-umsoknir.dev01.devland.is/](https://dynamic-environment-for-docker-static-umsoknir.dev01.devland.is/)

I mainly evaluated two implementations on how to do this:
1. Extract environment in docker entrypoint and insert into index.html.
2. Extract environment in docker entrypoint and serve through nginx as json data.

I chose nr 1 because then we don't need to make the environment values async which would change our current workflow regarding environments and we would also need to wait for the network request to be able to render the website.

This only works when we are running the docker containers, they are not for local development (at least not yet). So we can't use environment variables locally.

## What

A way to get environment variables from our environment to docker-static deployments

## Why

It is a more stable way to decide environment values than checking the current domain name

## Checklist:


apps/service-portal/src/environments/environment.prod.ts
```diff
@@ -1,18 +1,22 @@
...
-        ? 'https://innskra.island.is'
-        : 'https://identity-server.dev01.devland.is',
+    IDENTITY_SERVER_ISSUER_URL: window.ENV.IDENTITY_SERVER_ISSUER_URL,
   },
   sentry: {
     dsn:
       'https://3c45a55273774b91a897b85e0a1243d1@o406638.ingest.sentry.io/5501494',
   },
-  documentProviderAdmins: process.env.NX_DOCUMENT_PROVIDER_ADMINS || '',
+  documentProviderAdmins: window.ENV.NX_DOCUMENT_PROVIDER_ADMINS || '',
```
dabbeg, line 19: This is a secret in helm, is it a secret? if so it should not be in the service-portal as it will be bundled to the client.

apps/service-portal/src/hooks/useNavigation/useNavigation.ts
```diff
@@ -23,12 +23,6 @@ const filterNavigationTree = (
           route.path.includes(item.path)),
     ) !== undefined || item.systemRoute === true
 
-  // Only display the document-provider module for those who have access
-  if (item.path === '/skjalaveitur') {
```
dabbeg, line 27: This should be visible locally, if we really want to only show specific users we can do that in configcat later.

scripts/ci/Dockerfile
```diff
@@ -79,6 +79,7 @@ ENV GIT_SHA=${GIT_SHA}
 ENV APP=${APP}
 ENV BASEPATH=/
 
-RUN mkdir -p /etc/nginx/templates
-ADD scripts/ci/dockerfile-config/nginx/* /etc/nginx/templates
+RUN mkdir -p /etc/nginx/templates && apk add bash
```
sindrig, line 82: just `RUN apk add bash` in a separate layer. don't see what it has to do with the creation of this directory :)

scripts/ci/dockerfile-assets/bash/extract-environment.sh
```diff
@@ -0,0 +1,58 @@
...
+
+work_dir="/usr/share/nginx/html"
+file="$work_dir/index.html"
+
+script_start="<script id=\"environment\">"
+script_end="<\/script>"
+placeholder="$script_start.*$script_end"
+
+function extract_environment() {
+  environment_initial_value="{ "
```
sindrig, line 13: seems like a job for jq :grimacing:

eirikurn (dismissed):
I like having this in the html file.

Two requests:

* The paranoid side of me would like to prefix or whitelist the extracted env keys. Rather than expose anything referenced in the code. Feel free to argue and ignore :)
* This will cause issues for CSP. Can you add this to an application/json script tag, and create a getEnv('ENV_NAME') utility? (See NEXT_DATA on island.is)

scripts/ci/dockerfile-assets/bash/extract-environment.sh
```diff
@@ -0,0 +1,59 @@
...
+
+set -e
+
+work_dir="/usr/share/nginx/html"
+file="$work_dir/index.html"
+
+placeholder="<!-- environment placeholder -->"
+
+function extract_environment() {
+  env_prefix="SI_PUBLIC_"
```
darrikonn, line 15: Should the prefix maybe come from an environment variable in order to sync with the ts files?

dabbeg:
> I like having this in the html file.
> 
> Two requests:
> 
>     * The paranoid side of me would like to prefix or whitelist the extracted env keys. Rather than expose anything referenced in the code. Feel free to argue and ignore :)

I like that! I thought initially that it would be enough to trust the code but this is safer. I added a prefix `SI_PUBLIC_`.
The bash script will only pick up references with that prefix and the getStaticEnv function will throw an error to notify developers that forget to put the prefix.

>     * This will cause issues for CSP. Can you add this to an application/json script tag, and create a getEnv('ENV_NAME') utility? (See NEXT_DATA on island.is)

Added the application/json tag and created a library with the getStaticEnv function that wraps window.ENV with tests.

libs/utils/environment/src/lib/static-environment.ts
```diff
@@ -0,0 +1,14 @@
...
+  interface Window {
+    ENV?: any
+  }
+}
+
+export const ENV_PREFIX = 'SI_PUBLIC_'
+
+export const getStaticEnv = (environmentVariableName: string) => {
+  if (!environmentVariableName.startsWith(ENV_PREFIX)) {
+    throw new Error(`Variable must be prefixed with ${ENV_PREFIX}`)
```
darrikonn: Nice.

But I believe this also needs a documentation, most likely from the root README where the `<!-- environment placeholder -->` is also specified, since it won't be auto generated by nx.

scripts/ci/dockerfile-assets/bash/extract-environment.sh
dabbeg: I think that would be more confusing as it would need to be defined for every environment in helm and we want this to be the same in all environments.
And also we want the value of this to be visible for developers.

libs/utils/environment/src/lib/static-environment.ts
dabbeg: Good point, I added instructions on how to use this in the root README

Thorrik58 (approved):
Awesome!

pshomov (approved):
LGTM

scripts/ci/dockerfile-assets/bash/extract-environment.sh
```diff
@@ -0,0 +1,59 @@
...
+  environment="{}"
+  env_names=$(grep -ohr "$env_prefix\w*" "$work_dir/" | sort | uniq)
```
eirikurn, line 17: Bash question. If grep finds nothing when I run this command on my machine, it exits with an error status code. Will that crash the whole script? Maybe that's fine?

eirikurn (approved):
Love it!

scripts/ci/dockerfile-assets/bash/extract-environment.sh
dabbeg: Because we have `set -e` any command that exists with an error should crash the script.

The command should not exit with an error though, even if it does not find any references.
I have tried this on my linux and also in the Docker container that runs this script so we should be fine. Not sure if this is a mac thing.

PR #3007, opened 2021-02-23
Sjukratryggingar/Handle fetch failures

dd-sdang:
# Sjukratryggingar/Handle fetch failures

## What

Resolved providers with empty objects if fail.

## Why

If we can not determine if user
- has an active draft
- is already insured
- has pending application
(because of failed providers) this will let the user continue to create and submit a new application anyway.

## Checklist:


tthornvik (approved):
Looks good

PR #3018, opened 2021-02-24
Fix: Make address update date nullable

dd-sdang:
# National registry fix nullalbe address date

## What

Make last update date nullable.

## Why

Data from national registry api returns null even if users have correctly registered addresses. It is possibly that the data in the dev env is not complete. But this fix will prevent queries to fail.

## Checklist:


darrikonn (approved):

obmagnusson (approved):
LGTM

PR #3019, opened 2021-02-24
Crc/submit for parent a

Rikkiabb:
# \<Description\>

Changes application state on submit

## Checklist:

### DIFF
--- a/libs/application/templates/children-residence-change/src/fields/Overview/index.tsx
+++ b/libs/application/templates/children-residence-change/src/fields/Overview/index.tsx
@@ -47,7 +47,7 @@ const Overview = ({ application }: FieldBaseProps) => {
 
   return (
     <>
-      <Box marginTop={3}>
+      <Box marginTop={0}>
         <AlertMessage
           type="info"
           title={formatMessage(m.contract.alert.title)}
--- a/libs/application/templates/children-residence-change/src/forms/ChildrenResidenceChangeForm.ts
+++ b/libs/application/templates/children-residence-change/src/forms/ChildrenResidenceChangeForm.ts
@@ -9,6 +9,7 @@ import {
   buildSubSection,
   buildMultiField,
   buildSubmitField,
+  DefaultEvents,
 } from '@island.is/application/core'
 import Logo from '../../assets/Logo'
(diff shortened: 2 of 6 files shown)


libs/application/templates/children-residence-change/src/fields/Overview/index.tsx
```diff
@@ -47,7 +47,7 @@ const Overview = ({ application }: FieldBaseProps) => {
 
   return (
     <>
-      <Box marginTop={3}>
+      <Box marginTop={0}>
```
Rikkiabb, line 50: I changed this because we now have a `buildMultiField` from the application system 🤦  and that has a mandatory description 🤦  which I set to an empty string but it of course takes up vertical space. 🤦 
So even though marginTop here is set to zero it is still a little bit off. Thinking of having this as an TODO to look at later, any thoughts?

libs/application/templates/children-residence-change/src/forms/ChildrenResidenceChangeForm.ts
```diff
@@ -131,10 +134,27 @@ export const ChildrenResidenceChangeForm: Form = buildForm({
...
-          component: 'Overview',
+          children: [
+            buildCustomField({
+              id: 'residenceChangeReview',
+              title: m.contract.general.pageTitle,
+              component: 'Overview',
+            }),
+            buildSubmitField({
+              id: 'assign',
+              title: '',
```
Rikkiabb, line 148: ☹️

libs/application/templates/children-residence-change/src/fields/Overview/index.tsx
VilliGunn: The only thing I can think of right now is not using this `buildSubmitField` and create our own. But I guess this will do for now..

Rikkiabb: Yes agree, but it is big refactor since we need to replace the footer as well then and have our own custom custom component. But I think it is definitely a todo and the sooner the better 🙏 
IMO then this approach is a patch and then we will do it the right way when we have the time.

VilliGunn (approved):

PR #3037, opened 2021-02-25
feature(children residence change): pdf signature

VilliGunn:
# Enables digital signature for generated pdfs

- Parent A will be able to sign and submit the application

https://app.asana.com/0/1199733748699731/1199906048712046

## What

- Added the `setBeforeSubmitCallback` function to `FormField` and `FormMultiField`.
  - To be able to run a custom 'click handler' before form is submitted
- Added mutations for file signature and upload.
- Added a signature flow in the frontend for Parent A.
  - Opens a modal with a `controlCode` that should match the code on the users phone.
  - If everything works the pdf will be uploaded with the signature to s3.
  - If there is an error or if the user cancells the request there will be a default error message shown.
- Added props to the `BaseModal` component for usage without focusable element inside the modal.

### Note
- Only works locally with the `DOKOBIT_ACCESS_TOKEN` env variable set
  - Available in AWS parameter store
- Hardcoded values are needed on localhost for name and phonenumber
  - https://github.com/island-is/island.is/blob/main/apps/application-system/api/src/app/modules/application/files/file.service.ts#L109
  - https://github.com/island-is/island.is/blob/main/apps/application-system/api/src/app/modules/application/files/file.service.ts#L110
- Errors will be handled better once the backend supports that

## Why

To be able to sign and upload signed PDFs from the application flow.

## Screenshots / Gifs

Screenshot 2021-03-01 at 14 52 33
Screenshot 2021-03-01 at 14 52 39
Screenshot 2021-03-01 at 14 52 45

## Checklist:


libs/application/graphql/src/lib/mutations/uploadSignedFile.ts
```diff
@@ -0,0 +1,9 @@
+import { gql } from '@apollo/client'
+
+export const UPLOAD_SIGNED_FILE = gql`
+  mutation UploadSignedFile($input: UploadSignedFileInput!) {
+    uploadSignedFile(input: $input) {
+      externalData
```
Rikkiabb, line 6: Wondering about this flow. 
Maybe we shouldn't request anything since we are not adding anything to the application just waiting and seeing if the request fails or succeeds. 
What do you think?

libs/application/templates/children-residence-change/src/fields/Overview/index.tsx
```diff
@@ -13,47 +17,122 @@ import {
...
+  })
+  const [uploadSignedFile] = useMutation(UPLOAD_SIGNED_FILE, {
+    onError: () =>
+      dispatchFileSignature({
+        type: FileSignatureActionTypes.ERROR,
+        status: FileSignatureStatus.UPLOAD_ERROR,
+        error: '500',
+      }),
+    onCompleted: () =>
+      dispatchFileSignature({ type: FileSignatureActionTypes.SUCCESS }),
```
Rikkiabb, line 66: Can we add new line between each mutation for readability?

libs/application/templates/children-residence-change/src/fields/Overview/index.tsx
```diff
@@ -13,47 +17,122 @@ import {
...
+            id: application.id,
+            type: pdfType,
+          },
+        },
+      })
+      if (requestResponse?.data) {
+        const documentToken =
+          requestResponse.data.requestFileSignature?.externalData?.fileSignature
+            ?.data?.documentToken
+        const signedFileReponse = await uploadSignedFile({
```
Rikkiabb, line 100: typo: *signedFileResponse

libs/application/templates/parental-leave/src/forms/ParentalLeaveForm.ts
```diff
@@ -221,14 +221,12 @@ export const ParentalLeaveForm: Form = buildForm({
...
-                      PensionFundsQuery
-                    >({
-                      query: GetPensionFunds,
+                    const { data } = await apolloClient.query<UnionQuery>({
+                      query: GetUnions,
                     })
 
                     return (
-                      data?.getPensionFunds.map(({ id, name }) => ({
+                      data?.getUnions.map(({ id, name }) => ({
```
Rikkiabb, line 229: Should these changes be here?

libs/application/templates/children-residence-change/src/fields/Overview/SignatureModal/index.tsx
```diff
@@ -0,0 +1,107 @@
...
+      className={style.modal}
+      modalLabel={formatMessage(signatureModal.general.title)}
+      isVisible={modalOpen}
+      onVisibilityChange={(visibility: boolean) => {
+        if (!visibility) {
+          onClose()
+        }
+      }}
+      // When there is an error it should not be possible to close the modal
+      hideOnEsc={hasError}
```
Rikkiabb, line 45: How is the user flow when this happens?
Does the user see a retry button and is unable to close the modal?
Or what is the situation? 😄

libs/application/templates/children-residence-change/src/fields/Overview/index.tsx
VilliGunn: Nice catch 🧤

libs/application/templates/children-residence-change/src/fields/Overview/SignatureModal/index.tsx
VilliGunn: The user should not be able to close the modal with the `esc` key when the the modal is in a loading state or displaying the `controlCode` the modal should only close automatically on `success` or when there is an `error` and then the user will be able to press the `Close` button or the `esc` key to close.

libs/application/templates/parental-leave/src/forms/ParentalLeaveForm.ts
VilliGunn: Nope 😅

libs/application/graphql/src/lib/mutations/uploadSignedFile.ts
VilliGunn: I'm not sure how to go about doing this, I would like to investigate at a later date 👍

libs/application/templates/children-residence-change/src/fields/Overview/SignatureModal/index.tsx
Rikkiabb: Isn't the comment on line 44 then wrong? 😊

libs/application/templates/children-residence-change/src/fields/Overview/index.tsx
```diff
@@ -13,47 +17,124 @@ import {
...
+      dispatchFileSignature({ type: FileSignatureActionTypes.REQUEST })
+      const requestResponse = await requestFileSignature({
+        variables: {
+          input: {
+            id: application.id,
+            type: pdfType,
+          },
+        },
+      })
+      if (requestResponse?.data) {
```
Rikkiabb, line 98: Should we  use `.then()` here?

libs/application/templates/children-residence-change/src/lib/messages/signatureModal.ts
```diff
@@ -0,0 +1,45 @@
+import { defineMessages } from 'react-intl'
+
+// Signature
+export const signatureModal = {
+  general: defineMessages({
+    title: {
+      id: 'crc.application:section.contract.signatureModal.title',
```
Rikkiabb, line 7: Wondering about the id's in this file. Normally we have been following the rule `crc.application:section.{constName}.{messagesDefiniationName}.{variableName}` I am wondering in this scope if it makes sense to have just: `crc.application:signatureModal.title`?

I know that this is used in the contract section but just thinking if it makes sense to have as a part of the key here, since maybe we would like to use this modal elsewhere?

libs/application/templates/children-residence-change/src/fields/Overview/SignatureModal/index.tsx
```diff
@@ -0,0 +1,107 @@
...
+        boxShadow="large"
+        borderRadius="standard"
+        background="white"
+        paddingX={[3, 3, 5, 12]}
+        paddingY={[3, 3, 4, 10]}
+      >
+        <Text variant="h1" marginBottom={2}>
+          {formatMessage(signatureModal.general.title)}
+        </Text>
+        {(() => {
```
jeremybarbet, line 60: I would personally move this into a sub-component to make it clearer and easier to read, instead of having an anonymous function to handle the displaying logic

VilliGunn: True, was going to do later but decided to do some of the work now :)

jeremybarbet (approved):
LGTM

libs/application/templates/children-residence-change/src/fields/Overview/index.tsx
```diff
@@ -56,21 +75,88 @@ const Overview = ({ application }: FieldBaseProps) => {
...
+          variables: {
+            input: {
+              id: application.id,
+              documentToken: documentToken,
+              type: pdfType,
+            },
+          },
+        })
+          .then(() => {
+            return true
```
Rikkiabb, line 122: We could return response.data?.uploadSignedFile?.documentSigned here, if we want too :) 
Think it makes the code more explicit of what we want from the endpoint.

Rikkiabb (approved):
One small nit comment from me, your choice if you want to change it or ignore it 😊  Awesome work here 👏

libs/application/templates/children-residence-change/src/fields/Overview/index.tsx
VilliGunn: I agree and we should change this in another PR 😃

PR #3046, opened 2021-02-26
Adds flow for parent B

Rikkiabb:
# \<Description\>

There are two todo's as a follow up of this PR:
1. Add an endpoint for parent B which creates presigned url for signed document.
2. UI wise the flow is the same as for parent A this means in the overview screen we display a section 'Tengiliðaupplýsingar hins foreldris'. We are talking about changing this flow for parent A so this won't be displayed in their flow either but that decision has not been finalized. When we take that decision we will either remove this section from both flows or hide for parent b. For now I am just letting it be.

## Checklist:


libs/application/templates/children-residence-change/src/forms/ParentBForm.ts
```diff
@@ -15,12 +15,35 @@ export const ParentBForm: Form = buildForm({
...
-      title: m.section.backgroundInformation,
+      id: 'approveTermsParentB',
+      title: m.section.effect,
       children: [
-        buildTextField({
-          id: 'placeholderId',
-          title: 'placeholder',
+        buildCustomField({
+          id: 'approveTermsParentB',
+          title: 'Hvaða áhrif hefur breytingin?',
```
VilliGunn, line 23: There is a translation string for this but it seems to be missing here and in the parentA form as well for some reason.
`m.terms.general.pageTitle`

VilliGunn (commented):
Just this one minor comment 👍

libs/application/templates/children-residence-change/src/forms/ParentBForm.ts
Rikkiabb: Nice catch, this a copy/paste error 🙈 Fixed now :)

VilliGunn (approved):

PR #3052, opened 2021-02-26
Feature roletoenv

essgje:
# \<Description\>

## What

## Why

## Screenshots / Gifs

## Checklist:


tommigitgaur (dismissed):
LFTM

tommigitgaur (dismissed):
LFTM

tommigitgaur (approved):
LFTM

apps/skilavottord/ws/src/environments/environment.ts
```diff
@@ -24,5 +24,8 @@ export default {
     restUsername: process.env.FJARSYSLA_REST_USER,
     restPassword: process.env.FJARSYSLA_REST_PASS,
   },
+  skilavottord: {
+    userList: process.env.USER,
+  },
```
eirikurn, line 29: This is breaking schemas.ts on main. Please pick a better env name, like "SKILAVOTTORD_USER_LIST". On many machines, the USER env is the name of the OS user, in my case "eirikur", which crashes when passed to JSON.parse().

You should also:

* Add this to the environment.prod.ts.
* Add a default value so we can generate schemas without crashing the process, eg `process.env.SKILAVOTTORD_USER_LIST ?? "[]"`.
* Document the format of this environment variable in README.md to help other devs get started with this project.

tommigitgaur: We will fix that

tommigitgaur: I have fixed this and changes are in main now.n

PR #3059, opened 2021-02-27
Improve schema error handling

eirikurn:
## What

Make sure schemas.js and postinstall.js fail with non-zero error code if there is an error.

Also switch to spawn to stream stdout/stderr from sub commands.

## Why

Schema generation would silently fail after #3052 when running `yarn install`, which is confusing to developers who have out-of-date schemas even after running schema generation script.

## Checklist:

- [] I have added tests that prove my fix is effective or that my feature works


jeremybarbet (approved):
Nice 👍

PR #3063, opened 2021-03-01
Adds flow for parent b to view pdf contract

Rikkiabb:
# \<Description\>

Adds flow for parent b to be able to view signed contract from parent a. 

## Checklist:


libs/application/templates/children-residence-change/src/lib/ChildrenResidenceChangeTemplate.ts
```diff
@@ -13,7 +13,7 @@ import { dataSchema } from './dataSchema'
 
 type Events = { type: DefaultEvents.ASSIGN } | { type: DefaultEvents.SUBMIT }
 
-enum States {
+export enum States {
```
Rikkiabb, line 16: Wondering if we wan't to have a type file for the types we use in the template or if we wan't to export them directly like this when we need them.

libs/application/templates/children-residence-change/src/fields/Overview/index.tsx
```diff
@@ -24,24 +28,36 @@ const Overview = ({ application }: FieldBaseProps) => {
...
+      ? createPdfPresignedUrl(input)
+      : getPresignedUrl(input)
+  }, [application.id, createPdfPresignedUrl, getPresignedUrl])
 
   const pdfUrl =
-    response?.createPdfPresignedUrl?.attachments?.[
+    createResponse?.createPdfPresignedUrl?.attachments?.[
+      PdfTypes.CHILDREN_RESIDENCE_CHANGE
+    ] ||
+    getResponse?.getPresignedUrl?.attachments?.[
```
Rikkiabb, line 60: Not loving this, but discussed with Villi and we will do a refactor once frontend for signature is done since it is touching on similar things.

libs/application/templates/children-residence-change/src/fields/Overview/index.tsx
```diff
@@ -24,24 +28,36 @@ const Overview = ({ application }: FieldBaseProps) => {
...
   })
 
+  const [
+    getPresignedUrl,
+    { data: getResponse, loading: getLoadingUrl },
+  ] = useLazyQuery(GET_PRESIGNED_URL)
+
   useEffect(() => {
-    createPdfPresignedUrl({
+    let input = {
```
VilliGunn, line 42: `input` is not re-assigned so we should use `const` here

libs/application/templates/children-residence-change/src/lib/ChildrenResidenceChangeTemplate.ts
VilliGunn: I would say it just depends..One thing regarding this though is that I would probably change this enum to `ApplicationStates` and then we can get rid of `{ States as ApplicationStates }` where the type is being imported.

apps/application-system/api/src/app/modules/application/application.controller.ts
```diff
@@ -651,4 +652,38 @@ export class ApplicationController {
...
+
+    const { updatedApplication } = await this.applicationService.update(
+      application.id,
+      {
+        attachments: {
+          ...application.attachments,
+          [type]: url,
+        },
+      },
+    )
```
VilliGunn, line 685: Nitpick: Since we are destructuring the `input` we could also do that with the `application`

That being said I actually think that in this case the most readable code would be skipping the desctructuring of `input` so when reading the code it is clear where each prop is coming from.

VilliGunn (commented):
Nice one, just some minor comments 🌮

libs/application/templates/children-residence-change/src/fields/Overview/index.tsx
Rikkiabb: Nice catch, I thought the linter would catch this, is that incorrect?

apps/application-system/api/src/app/modules/application/application.controller.ts
Rikkiabb: The way this controller is architectured is that it takes an application id uses an interceptor to look that id up in the db and returns that application to the function. 
So you are talking about doing:
`const { id, typeId } = application`?
Agree with the second point since in the scope of this small function we do actually have two types, pdf type and application type so it is nice to have clarity there, will fix 😄

libs/application/templates/children-residence-change/src/fields/Overview/index.tsx
Rikkiabb: Nevermind, it did catch just didn't notice 🤦

apps/application-system/api/src/app/modules/application/application.controller.ts
```diff
@@ -651,4 +652,36 @@ export class ApplicationController {
 
     return application
   }
+
+  @Put('application/:id/presignedUrl')
```
jeremybarbet, line 656: Looks like we have some inconsistency over the names, we either use `applications/:id` or `application/:id`. We will need to fix that in another PR

apps/application-system/api/src/app/modules/application/application.controller.ts
```diff
@@ -651,4 +652,36 @@ export class ApplicationController {
...
+  @ApiParam({
+    name: 'id',
+    type: String,
+    required: true,
+    description: 'The id of the application to update the state for.',
```
jeremybarbet, line 661: The description is not very related to the function it's doing. I agree a lot of other endpoints are also not well documented

libs/api/domains/application/src/lib/application.resolver.ts
```diff
@@ -163,4 +164,12 @@ export class ApplicationResolver {
   ): Promise<Application> {
     return this.applicationService.uploadSignedFile(input, user.authorization)
   }
+
+  @Query(() => Application, { nullable: true })
+  getPresignedUrl(
+    @Args('input') input: GetPresignedUrlInput,
+    @CurrentUser() user: User,
+  ): Promise<Application> {
```
jeremybarbet, line 172: Is this returning the `Application` object? or a presigned URL?

apps/application-system/api/src/app/modules/application/application.controller.ts
Rikkiabb: Nice catch, we can fix this in this PR if you want. Did not notice that you had always `applications/:id`. Why is it plural? To me it makes sense that you are working with one application per that id. 
Anyways don't mind fixing this in this PR if you want 😊

apps/application-system/api/src/app/modules/application/application.controller.ts
Rikkiabb: Correct! :D Will fix :)

apps/application-system/api/src/app/modules/application/application.controller.ts
jeremybarbet: If you are up for it, go for it 👍  Yeah I also prefer application/:id for a single application endpoint

libs/api/domains/application/src/lib/application.resolver.ts
Rikkiabb: It returns an update application which contains the presigned url as an attachment. 
The reason why I went for that way was that I noticed that every endpoint in the AS returns an application. So I thought that was the architecture you were going for. Is that wrong? Should I return just the `Promise<string>`?

apps/application-system/api/src/app/modules/application/application.controller.ts
Rikkiabb: I am going to update the description as well on other endpoints I have created, hope you don't mind adding it to this PR 😊

apps/application-system/api/src/app/modules/application/application.controller.ts
Rikkiabb: One question about this. How will this work in the buildpipe and on local machines? 
Everybody will have to clean their cache for the autogenareted files for this to update since we are just changing the name of the endpoint, right?

libs/api/domains/application/src/lib/application.resolver.ts
jeremybarbet: I'm not sure about it, but it might be an architectural decision indeed. In my opinion, it looks a bit weird to have the wording of "getUrl" and you get an application in return where you have to look for the field you need but that's just my 2ct

apps/application-system/api/src/app/modules/application/application.controller.ts
jeremybarbet: no problem 👍

apps/application-system/api/src/app/modules/application/application.controller.ts
jeremybarbet: The CI is not using any cache, so it will regenerate all files. Locally, it's common that if you get a type issue to re-run `yarn schemas --skip-cache`

Rikkiabb: Fixed

libs/api/domains/application/src/lib/application.resolver.ts
Rikkiabb: Agree and fixed, I also fixed it for other endpoints so the PR is a little bit bigger 😅

apps/application-system/api/src/app/modules/application/e2e/application.spec.ts
```diff
@@ -440,17 +440,15 @@ describe('Application system API', () => {
...
 
-    const newState = await server
+    const res = await server
       .put(`/application/${postResponse.body.id}/createPdf`)
       .send({
         type: type,
       })
       .expect(200)
-
+    console.log(res)
```
VilliGunn, line 449: 👀

VilliGunn (commented):
Minor comments from me, we need to make sure to update #3037 after this is merged or the other way around.

apps/application-system/api/src/app/modules/application/application.controller.ts
```diff
@@ -649,6 +626,28 @@ export class ApplicationController {
...
+    allowEmptyValue: false,
+  })
+  @ApiOkResponse({ type: PresignedUrlResponseDto })
+  async getPresignedUrl(
+    @Param('id', new ParseUUIDPipe(), ApplicationByIdPipe)
+    application: Application,
+    @Param('pdfType') type: PdfTypes,
+  ): Promise<PresignedUrlResponseDto> {
+    this.fileService.validateApplicationType(application.typeId)
+    let url = this.fileService.getPresignedUrl(application.id, type)
```
VilliGunn, line 649: Same here with never reassigning `url`

jeremybarbet (dismissed):
LGTM

apps/application-system/api/src/app/modules/application/application.controller.ts
```diff
@@ -548,50 +553,39 @@ export class ApplicationController {
...
-    const { updatedApplication } = await this.applicationService.update(
-      application.id,
-      {
-        attachments: {
-          ...application.attachments,
-          [type]: url,
-        },
-      },
-    )
+    let url = await this.fileService.createPdf(application, type)
```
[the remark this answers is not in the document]
Rikkiabb: 🤦

apps/application-system/api/src/app/modules/application/e2e/application.spec.ts
Rikkiabb: 🤦 🤦

apps/application-system/api/src/app/modules/application/application.controller.ts
Rikkiabb: 🤦 🤦 🤦

VilliGunn (approved):
💯 🌮

jeremybarbet (approved):

PR #3070, opened 2021-03-01
Application System/Added background color prop on async select field

dd-sdang:
# Application System/Added background color prop on async select field

## What

Added the prop `backgroundColor` to the AsyncSelect field. 

## Why

According to the design system, enabled field should be blue and disabled should be white. Adding the prop allows us to decide the color.

## Checklist:


tthornvik (approved):
Looks good :)

DavidBachmann (approved):

PR #3080, opened 2021-03-02
Fix builds: free disk space in prepare step

sindrig:
This should be temporary - currently we're filling >10GBs in every build


pshomov (approved):
LGTM

PR #3086, opened 2021-03-02
Feature/search popularity score

Ingimarsson:
# Search popularity score

## What

This PR adds a `popularityScore` field to the ES indices. When articles or news are fetched the score is updated according to [this](https://stackoverflow.com/questions/11128086/simple-popularity-algorithm) formula. The `a` constant can be adjusted according to how many page hits should significantly change the ordering.

The `popularityScore` field persists when we run `re-sync` and `migrate` instead of getting zeroed each time

## Why

This allows us to order articles and search results by popularity.

## Screenshots / Gifs

image

## Checklist:


eirikurn (commented):
Note that this won't include all hits, since the pages are cached on the edge. Probably fine to consider this "approximate" popularity, since the updates aren't atomic. 

I assume it's "ok" load-wise to constantly update documents in elasticsearch?

I guess a better implementation would build this on top of some analytics/metrics service which we can hit for every request, and periodically sync to our index. I know there was interest in presenting this kind of data, gov.uk style. Do you have any update on that? Can we pull this kind of stuff from plausible apis?

Ingimarsson:
@eirikurn my first thought was to periodically fetch the data from Plausible and write the hit count to Elastic Search, with a cron job on midnight or something like that. But the feedback I got was that we might not want this to depend on an external service.

Plausible does have an API to fetch this data, but it's currently in beta and not documented.

I'm not sure how resource intensive the ES updates are, I might do some stress tests just in case. Since we have 5 sec micro caching on the GraphQL API I don't think there will be that many updates per second.

libs/api/domains/cms/src/lib/search/cmsSync.service.ts
```diff
@@ -187,4 +196,72 @@ export class CmsSyncService implements ContentSearchImporter<PostSyncOptions> {
...
+
+    const results = indices.body.match(
+      new RegExp(`island-${locale}-[a-z0-9]+`, 'gi'),
+    )
+
+    if (results.length < 1) {
+      return null
+    }
+    return results[0]
+  }
```
stefanhar, line 266: I'm not sure if this is the right approach.
Popularity score migration relates to index migration/creation not to index maintenance.
The cms sync is run periodically forcing us to have "when to run" logic and to dynamically find the target/source indexes.
We might get a cleaner approach by moving the migration logic into `island.is/apps/services/search-indexer/src/migrate/migrateElastic.ts` and not toughing the `popularityScore` inside imports.

stefanhar (changes requested):

libs/api/domains/cms/src/lib/search/cmsSync.service.ts
Ingimarsson: The popularity score migration is also done when doing `GET /sync` on the indexing service. Every time we want to update a document we need to do this score migration. It will read the old score of the documents from the index and put it in the documents in `importableData` before calling `client.bulk()` to write to ES. Otherwise the counter would be set to zero on every update, as the `index` function can not do partial updates to documents.

stefanhar: Yes missed that, we could look into alternative methods of importing data into the index like update/upsert allowing us to keep existing popularity fields on documents.
If we make sure, all fields are always defined (should be this way already) we should be able to safely update existing documents even if they are setting a subset of fields to empty

libs/api/domains/cms/src/lib/cms.elasticsearch.service.ts
```diff
@@ -46,10 +46,18 @@ export class CmsElasticsearchService {
...
       size: input.size,
     }
 
+    switch (input.sort) {
+      case 'popular':
+        query.sort = { popularityScore: 'desc' as sortDirection }
+        break
+      default:
+        query.sort = { 'title.sort': 'asc' as sortDirection }
+    }
```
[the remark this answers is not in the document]
Ingimarsson: Good idea.

libs/api/domains/cms/src/lib/dto/getArticles.input.ts
```diff
@@ -24,4 +24,9 @@ export class GetArticlesInput {
   @Max(1000)
   @IsOptional()
   size?: number = 100
+
+  @Field({ nullable: true })
+  @IsString()
+  @IsOptional()
+  sort?: 'title' | 'popular' = 'title'
```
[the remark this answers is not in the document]
Ingimarsson: I added the `@IsEnum()` so that GraphQL can return `sort must be a valid enum value` when an invalid value is used.

I tried creating an `enum sortDirection { title, popular }` but I'm not sure how to use it since `input.sort` is a string but `sortDirection.popular` is an integer.

libs/content-search-toolkit/src/services/elastic.service.ts
```diff
@@ -195,6 +195,39 @@ export class ElasticService {
...
+    const client = await this.getClient()
+    return client.update({
+      index,
+      id,
+      body: {
+        doc: {
+          popularityScore,
+        },
+      },
+    })
```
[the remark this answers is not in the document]
Ingimarsson: Changed this to a painless script and logging if an error is catched.

libs/api/domains/cms/src/lib/dto/getArticles.input.ts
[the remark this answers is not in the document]
stefanhar: Something like this might do it.
``` ts
import { registerEnumType } from '@nestjs/graphql'

export enum SortDirection {
  DESC = 'desc',
  ASC = 'asc',
}

registerEnumType(ContentLanguage, { name: 'SortDirection' })
```

Then use like this:
``` ts
@Field(() => SortDirection, { nullable: true })
```

libs/api/domains/cms/src/lib/search/cmsSync.service.ts
Ingimarsson: It would be cool to use something like upsert to persist the popularity score.

But from my understanding, the sync service does not know whether a document is being initialized or updated? (does the Contentful API provide this info?) We would need to pass that to the model mappers so that they know which fields to omit?

Maybe this is good enough for a first version? :)

libs/api/domains/cms/src/lib/dto/getArticles.input.ts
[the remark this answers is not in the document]
stefanhar: Something like this might do it.
``` ts
import { registerEnumType } from '@nestjs/graphql'

export enum SortDirection {
  DESC = 'desc',
  ASC = 'asc',
}

registerEnumType(SortDirection, { name: 'SortDirection' })
```

Then use like this:
``` ts
@Field(() => SortDirection, { nullable: true })
```

libs/api/domains/cms/src/lib/search/cmsSync.service.ts
Ingimarsson: I'm also wondering how permanent this feature would be? If we are thinking about presenting traffic data like [gov.uk](https://www.gov.uk/performance/site-activity) does, as @eirikurn mentioned, then we could get the hit count from there.

stefanhar: We need the popularityScore inside the document to be able to query by it so I see this as a permanent feature.
The challenges looks similar to me, I see the main difference in how the popularity score is updated in the document.

I believe adding `"doc_as_upsert":true` to the update query makes is behave as create or update (I haven't tested this).
If we can prevent the **sync destroys popularity score**&trade; issue we can handle this as a migration to be run only when new indexes are created

Ingimarsson: This sounds good. I have moved this logic to the migration scripts, among other things (like using painless scripts and upserts instead of insert)

apps/services/search-indexer/src/migrate/lib/elastic.ts
```diff
@@ -165,3 +169,92 @@ export const removeIndexIfExists = async (indexName: string) => {
...
+      requests.push({
+        update: { _index: newIndex, _id: obj._id },
+      })
+      requests.push({
+        doc: { popularityScore: oldScores[obj._id] },
+        // eslint-disable-next-line @typescript-eslint/camelcase
+        doc_as_upsert: true,
+      })
+    }
+  })
```
stefanhar, line 222: We might not need this since the `id` is known we can send the update command optimistically with the old `popularityScore` field, bulk allows for partial success. Update would fail if the document no longer exists.
We would have to guard against false success though e.g. on network errors.

Also we probably don't want `doc_as_upsert: true` this might cause documents to be created with only the `popularityScore` field set, e.g. if a document is deleted while this is running.

apps/services/search-indexer/src/migrate/lib/elastic.ts
```diff
@@ -165,3 +169,92 @@ export const removeIndexIfExists = async (indexName: string) => {
...
+    s: 'creation.date:desc',
+  })
+
+  const results = indices.body.match(
+    new RegExp(`island-${locale}-[a-z0-9]+`, 'gi'),
+  )
+
+  if (results.length < 2) {
+    return null
+  }
```
stefanhar, line 258: Maybe add comment here to make clear this is so we don't return the current index

stefanhar (changes requested):

libs/api/domains/cms/src/lib/cms.elasticsearch.service.ts
```diff
@@ -46,10 +47,14 @@ export class CmsElasticsearchService {
     const query = {
       types: ['webArticle'],
       tags: [],
-      sort: { 'title.sort': 'asc' as sortDirection },
+      sort: { 'title.sort': 'asc' as sortDirection } as sortableFields,
       size: input.size,
     }
 
+    if (input.sort === 'popular') {
+      query.sort = { popularityScore: 'desc' as sortDirection, ...query.sort }
```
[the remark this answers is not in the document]
Ingimarsson: This syntax works as well, and is used in all our other queries so I decided to stick with it :)

libs/content-search-toolkit/src/services/elastic.service.ts
```diff
@@ -195,6 +199,46 @@ export class ElasticService {
...
+              ctx._source.popularityScore = 0
+            }
+            `,
+            params: {
+              a: environment.popularityFactor,
+              t: Number(new Date()) / 1000,
+            },
+          },
+        },
+      })
```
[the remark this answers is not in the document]
Ingimarsson: I set retry_on_conflict to 1 not sure if a higher number is worth it.

apps/services/search-indexer/src/migrate/migrateElastic.ts
```diff
@@ -66,6 +66,28 @@ class App {
...
+
+            if (oldIndexName) {
+              try {
+                await elastic.migratePopularityScores(
+                  oldIndexName,
+                  newIndexName,
+                )
+                logger.info('Popularity scores from previous index migrated', {
+                  oldIndexName,
+                })
```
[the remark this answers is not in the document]
Ingimarsson: I agree, moved it to the try block above.

apps/services/search-indexer/src/migrate/lib/elastic.ts
```diff
@@ -165,3 +169,92 @@ export const removeIndexIfExists = async (indexName: string) => {
...
+export const migratePopularityScores = async (
+  oldIndex: string,
+  newIndex: string,
+) => {
+  const query = {
+    query: {
+      // eslint-disable-next-line @typescript-eslint/camelcase
+      match_all: {},
+    },
+    size: 10000,
```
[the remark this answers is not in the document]
Ingimarsson: I added a scroll query so that the state of the index is preserved while we are scrolling through. If we ever upgrade ES to >=7.10 then we can use a point in time instead.

libs/api/domains/cms/src/lib/cms.elasticsearch.service.ts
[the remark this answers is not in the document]
stefanhar: Object property order is not guaranteed in js although in this case it should be.
I think a more failsafe approach would be to use an array as we do in:
`libs/content-search-toolkit/src/queries/documentByMetaData.ts`
Using an array also seems to be the [documented](https://www.elastic.co/guide/en/elasticsearch/reference/7.11/sort-search-results.html) approach to multi field sorting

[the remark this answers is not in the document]
Ingimarsson: The sort property in `documentByMetaData.ts` is an Object, not an array. It turns out that `getDocumentsByMetaData` expects the sort field to be an Object but then maps it to an array before query-ing.

I changed `getDocumentsByMetaData` to accept an array instead (type `sortRule[]` that I created) it makes more sense to me.

apps/services/search-indexer/src/migrate/migrateElastic.ts
```diff
@@ -55,6 +55,17 @@ class App {
...
+              const oldIndexName = await elastic.getPreviousIndex(locale)
+              if (oldIndexName) {
+                await elastic.migratePopularityScores(
+                  oldIndexName,
+                  newIndexName,
+                )
+                logger.info('Popularity scores from previous index migrated', {
+                  oldIndexName,
+                })
+              }
```
stefanhar, line 68: Maybe add a info log in else here to make it clear this did not run maybe something like:
`logger.info('No older index found, skipping popularity score migration', {locale})`

apps/services/search-indexer/src/migrate/lib/elastic.ts
```diff
@@ -165,3 +169,99 @@ export const removeIndexIfExists = async (indexName: string) => {
...
+export const migratePopularityScores = async (
+  oldIndex: string,
+  newIndex: string,
+) => {
+  const oldScores = await getAllPopularityScores(oldIndex)
+
+  const requests = []
+
+  oldScores.forEach((x) => {
+    if (x.score > 0) {
```
stefanhar, line 230: Added suggestion to query making this redundant

apps/services/search-indexer/src/migrate/lib/elastic.ts
```diff
@@ -165,3 +169,99 @@ export const removeIndexIfExists = async (indexName: string) => {
...
+  const client = await esService.getClient()
+  const scores = []
+
+  // do a scroll query to get a snapshot of the index
+  let data = await client.search<SearchResponse<MappedData>, unknown>({
+    index,
+    body: {
+      query: {
+        // eslint-disable-next-line @typescript-eslint/camelcase
+        match_all: {},
```
stefanhar, line 188: We might be able to replace `match_all: {}` with something like:
```JSON
"range": {
  "popularityScore": {
    "gt": 0
  }
}
```
This would reduce payload and remove need for checks in code

apps/services/search-indexer/src/migrate/lib/elastic.ts
```diff
@@ -165,3 +169,99 @@ export const removeIndexIfExists = async (indexName: string) => {
...
+      _source: ['popularityScore'],
+    },
+    scroll: '1m',
+  })
+
+  while (data.body.hits?.hits.length) {
+    data.body.hits?.hits.forEach((x) => {
+      scores.push({
+        id: x._id,
+        score: x._source.popularityScore ?? 0,
```
stefanhar, line 201: With query change above `popularityScore` should always be a positive number

stefanhar (changes requested):

stefanhar (approved):
Nice

PR #3091, opened 2021-03-02
Adding delegations fields to db, api and form

unnarb:
#Adding delegations fields to db, api and form

## What

## Why

## Screenshots / Gifs

## Checklist:


libs/auth-api-lib/src/lib/entities/dto/base/client-base.dto.ts
```diff
@@ -280,4 +280,32 @@ export abstract class ClientBaseDTO {
...
+    example: true,
+  })
+  readonly supportsDelegation!: boolean
+
+  @IsBoolean()
+  @IsNotEmpty()
+  @ApiProperty({
+    example: true,
+  })
+  readonly supportsLegalGuardians!: boolean
```
valurefugl, line 296: Shouldn't these be nullable?

unnarb: This is the dto. The thought being that this would be turned to non nullable in the DB and in the models after updating clients that already exist

valurefugl (approved):

PR #3101, opened 2021-03-03
Api cms domain/strict typescript

kksteini:
# \<Description\>

I turned on strict in the tsconfig for TypeScript and fixed, accordingly, all of the following

- `libs/api/domains/cms/src/lib/models/*`
- `libs/api/domains/cms/src/lib/dto/*`
- `libs/api/domains/cms/src/lib/cms.resolver.ts`

## What

Utilizing the power of TypeScript fully.

## Why

By setting TypeScript to strict we are made better aware of variable typing as it flows through the system. This forces us to make certain assertions and have certain understanding about our code which in turn makes it more robust.

F.x, turning strict on allowed for identifying a couple of places where assumptions of `null` as a return were made but `undefined` was a potential instead.

### Input models
Checked resolvers/service and made decisions on assertions based on checks done in the layer (if any).

### Models
Those models who are directly made from content models in contentful: Not null assertions (! assertion) for required fields, (? assertion) otherwise.
Those models who were different were either made to match the mapper (if mapper had no safeguards then (! assertion), (? assertion) otherwise) or their usage traced in the system and decisions made from there

### Resolver
Added undefined/null safeguards where strict complained

### Services
The autogenerated `contentfulTypes.d.ts` file defines the `sys.contentTypes.sys.id` as a constant whereas the `getLocalizedEntries()` returns an `EntryCollection` wherein is contained `Entry` which defines `sys.contentTypes.sys.id` as a string. We know that the constant is correct in the service scope since we pass the content type in the parameter.

Therefore it is safe to assert that the items in the result are of a compatible, more specific, type.

## Checklist:

- [ x ] I have commented my code, particularly in hard-to-understand areas
- [ x ] My changes generate no new warnings


libs/api/domains/cms/src/lib/cms.resolver.ts
```diff
@@ -368,7 +368,7 @@ export class CmsResolver {
     @Args('input') { lang, slug }: GetSingleNewsInput,
   ): Promise<News | null> {
     return this.cmsElasticsearchService.getSingleDocumentTypeBySlug<News>(
-      getElasticsearchIndex(lang),
+      getElasticsearchIndex(lang ?? 'is'),
```
stefanhar, line 371: It looks like these fields have default values in dto but are marked as optional in typescript.
We should be able to remove the optional mark in TS since they should always be set, that would negate the need for checks here

libs/api/domains/cms/src/lib/cms.resolver.ts
```diff
@@ -377,7 +377,7 @@ export class CmsResolver {
   @Query(() => [String])
   getNewsDates(@Args('input') input: GetNewsDatesInput): Promise<string[]> {
     return this.cmsElasticsearchService.getNewsDates(
-      getElasticsearchIndex(input.lang),
+      getElasticsearchIndex(input.lang ?? 'is'),
```
stefanhar, line 380: ☝️

libs/api/domains/cms/src/lib/cms.resolver.ts
```diff
@@ -386,7 +386,7 @@ export class CmsResolver {
   @Query(() => NewsList)
   getNews(@Args('input') input: GetNewsInput): Promise<NewsList> {
     return this.cmsElasticsearchService.getNews(
-      getElasticsearchIndex(input.lang),
+      getElasticsearchIndex(input.lang ?? 'is'),
```
stefanhar, line 389: ☝️ ☝️

libs/api/domains/cms/src/lib/cms.resolver.ts
```diff
@@ -427,7 +427,7 @@ export class LatestNewsSliceResolver {
   @ResolveField(() => [News])
   async news(@Parent() { news: input }: LatestNewsSlice): Promise<News[]> {
     const newsList = await this.cmsElasticsearchService.getNews(
-      getElasticsearchIndex(input.lang),
+      getElasticsearchIndex(input.lang ?? 'is'),
```
stefanhar, line 430: ☝️ ☝️ ☝️

stefanhar (changes requested):

libs/api/domains/cms/.eslintrc
```diff
@@ -1,5 +1,7 @@
 {
   "extends": "../../../../.eslintrc",
-  "rules": {},
+  "rules": {
+    "@typescript-eslint/no-inferrable-types": 0
+  },
```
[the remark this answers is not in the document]
kksteini: Weird. This caused errors at first.

libs/api/domains/cms/src/lib/cms.resolver.ts
kksteini: Good point

libs/api/domains/cms/src/lib/cms.health.ts
```diff
@@ -38,14 +38,20 @@ export class CmsHealthIndicator extends HealthIndicator {
   }
 
   async isHealthy(): Promise<HealthIndicatorResult> {
+    if (!process.env.ELASTIC_NODE) {
+      throw new HealthCheckError(
+        'Cms check failed',
+        `ELASTIC_NODE environment variable not provided, is ${process.env.ELASTIC_NODE}`,
+      )
+    }
```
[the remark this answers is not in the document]
kksteini: Good point. Supplied it from `environment.ts`.

libs/api/domains/cms/src/lib/dto/getAlertBanner.input.ts
```diff
@@ -5,9 +5,9 @@ import { IsString } from 'class-validator'
 export class GetAlertBannerInput {
   @Field()
   @IsString()
-  id: string
+  id!: string // Used in the main layout with an explicit id in the input
```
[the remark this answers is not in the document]
stefanhar: Got this turned around absence of `@IsOptional` should ensure this is always set 😅

libs/api/domains/cms/src/lib/cms.health.ts
```diff
@@ -38,14 +38,20 @@ export class CmsHealthIndicator extends HealthIndicator {
...
     const requiredUrls = {
       contentful: process.env.CONTENTFUL_HOST,
       elasticsearch: new URL(process.env.ELASTIC_NODE).hostname,
     }
 
     const response = await this.health.check(
       Object.entries(requiredUrls).map(([key, url]) => () =>
-        this.canUrlBeResolved(key, url),
+        this.canUrlBeResolved(key, url ?? ''),
```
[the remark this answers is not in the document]
stefanhar: The `??`check here should not be needed

libs/api/domains/cms/src/lib/cms.resolver.ts
```diff
@@ -298,7 +298,7 @@ export class CmsResolver {
   ): Promise<LifeEventPage[]> {
     return this.cmsContentfulService.getLifeEventsInCategory(
       input.lang,
-      input.slug,
+      input.slug ?? '',
```
stefanhar, line 301: I think slug is not allowed to be `undefined` here via `class-calidator` in model. Maybe we should add a non-null assertion operator(!) to the model to reflect this, then there should be no need for check here

stefanhar (approved):

PR #3108, opened 2021-03-03
Feature/dummy test

unnarb:
#Altering fields to non nullable and adding a temporary dummy test since it runs into unknown problems

## What

## Why

## Screenshots / Gifs

## Checklist:

### DIFF
--- a/apps/services/auth-api/src/app/modules/case/e2e/auth-api.spec.ts
+++ b/apps/services/auth-api/src/app/modules/case/e2e/auth-api.spec.ts
@@ -1,216 +1,222 @@
-import { GrantDto, UserIdentityDto } from '@island.is/auth-api-lib'
-import { INestApplication } from '@nestjs/common'
-import { setup } from '../../../../../test/setup'
-import * as request from 'supertest'
-
-let app: INestApplication
-
-beforeAll(async () => {
-  app = await setup()
-})
-
-const userIdentity: UserIdentityDto = {
-  subjectId: '0123456789',
-  name: 'Sigurdur Sturla Bjarnason',
-  providerName: 'Audkenni',
-  active: false,
-  providerSubjectId: 'audkenni_provider',
-  claims: [
-    {
-      type: 'nat',
-      value: '1234560123',
-      valueType: 'natreg',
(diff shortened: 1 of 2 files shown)


valurefugl (approved):

dabbeg (commented):
I don't understand the reasons for commenting out all these tests, can you clarify that?

unnarb:
> I don't understand the reasons for commenting out all these tests, can you clarify that?

A seed file, setting all values for new delegation fields to false, ran on the dev db. We wanted to change the fields to NOT NULL since it is in the same structures as other Boolean fields in the table. The test fails and the only logical reason for that, we feel, is that it is running these tests to a temporary db, that the build pipeline builds. Something is wrong in that pipeline. The test should be fine if the db in the build flow is setup correctly. We don't know the solution to that problem

dabbeg:
> > I don't understand the reasons for commenting out all these tests, can you clarify that?
> 
> A seed file, setting all values for new delegation fields to false, ran on the dev db. We wanted to change the fields to NOT NULL since it is in the same structures as other Boolean fields in the table. The test fails and the only logical reason for that, we feel, is that it is running these tests to a temporary db, that the build pipeline builds. Something is wrong in that pipeline. The test should be fine if the db in the build flow is setup correctly. We don't know the solution to that problem

If I remember correctly the test database runs a code first migration with the sequelize models. Did you change the Sequelize models to be notnull as well?

unnarb:
> > > I don't understand the reasons for commenting out all these tests, can you clarify that?
> > 
> > 
> > A seed file, setting all values for new delegation fields to false, ran on the dev db. We wanted to change the fields to NOT NULL since it is in the same structures as other Boolean fields in the table. The test fails and the only logical reason for that, we feel, is that it is running these tests to a temporary db, that the build pipeline builds. Something is wrong in that pipeline. The test should be fine if the db in the build flow is setup correctly. We don't know the solution to that problem
> 
> If I remember correctly the test database runs a code first migration with the sequelize models. Did you change the Sequelize models to be notnull as well?

AHA... ok, that's the reason. We were going to do this in separate PR. Thank you

dabbeg (approved):

PR #3110, opened 2021-03-03
Sjukratryggingar/Fix condition render of text area

dd-sdang:
# Sjukratryggingar/Fix condition render of text area

## What

Removed check of previous answer so that it is always displayed if the user does not require wating period.

## Why

Because it caused infinite re-rendering as the value was set to undefined for some reason.

## Checklist:


tthornvik (approved):
Looks good :)

PR #3040, opened 2021-02-26
regression(core): make production bundle much smaller by removing mock code

eirikurn:
## What

Make sure there is no mocking code in production bundles.

## Why

Since #2446 fixed a bug in our "sideEffects" configuration to get mocking code in E2E builds, we've been bundling mocking code in our production builds as well.

This happened for two reasons:

* In NextJS apps, the process.env.API_MOCKS was undefined, so it was not defined correctly in the bundle. This meant webpack was unable to remove code that depended on it.
* In React apps, the @nrwl/web webpack config defines `process.env` to be an object containing all valid environment variables. Unfortunately, webpack is unable to statically analyze the object syntax to mark unused code. We need to define each `process.env.*` variable on their own so webpack can tree shake all of the mocking code. (this took **a long time** to figure out!)

**TODO:**

* Create an issue in @nrwl/nx.

## Checklist:

- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [x] Formatting passes locally with my changes
- [x] I have rebased against main before asking for a review

PR #3113, opened 2021-03-04
feat(judicial-system): Migrate to NextJS

oddsson:
# Move Judicial-system from React to Nextjs

https://app.asana.com/0/1199153462262248/1200017647359051

## What

Move Judicial-system from React to Nextjs

## Why

We began having issues after #3040. We don't know exactly what was causing issues, but regardless, we decided to migrate to NextJS because we believe there is value in being like everyone else (most web projects in this monorepo use NextJS and some components in island-ui are directly dependant on it). 

## Checklist:


addi (approved):

eirikurn (approved):
LGTM! I'll have to upgrade this in the NX11 PR, but that's my problem :)

PR #3123, opened 2021-03-04
chore: Add pkg.engines mandates + require node@14

maranomynet:
## What / Why

Making the system requirements [specified in the README](https://github.com/island-is/island.is#prerequisites) explicit

## Checklist:


package.json
```diff
@@ -2,6 +2,10 @@
   "name": "island.is",
   "version": "0.0.1",
   "license": "MIT",
+  "engines": {
+    "node": ">=12",
+    "yarn": ">=1.20"
```
maranomynet, line 7: Should we maybe pin down a major version here?

```suggestion
    "node": "^12.20",
    "yarn": "^1.20"
```

maranomynet: @eirikurn has been considering pinning the repo to node@^14 at some point in the near future — so this sort of thing should just wait until then.

maranomynet:
These files also need to be updated to use node@14:

* [scripts/ci/Dockerfile](/island-is/island.is/blob/main/scripts/ci/Dockerfile)
* [scripts/ci/Dockerfile.test](/island-is/island.is/blob/main/scripts/ci/Dockerfile.test)

eirikurn:
Let's change both Dockerfiles to use `node:14.16.0-alpine3.11`

maranomynet:
Done!

eirikurn (approved):
LGTM

PR #3148, opened 2021-03-05
Add resize query parameter to image URL in topic cards

Ingimarsson:
# Add image resize query parameter

## What

Add `?w=400` to image URL's for topic cards.

This reduces the request to https://island.is/stafraent-island/teymid-okkar by ~20 MB


libs/island-ui/core/src/lib/ProfileCard/ProfileCard.tsx
```diff
@@ -46,7 +46,7 @@ export const ProfileCard: FC<ProfileCardProps> = ({
       {image && (
         <Box
           className={styles.image}
-          style={{ backgroundImage: `url(${image})` }}
+          style={{ backgroundImage: `url(${image}?w=400)` }}
```
barabrian, line 49: I'm not sure if we should use project specific query params inside the core component, but this is definitely something we should be using when defining the image url when using the component. Maybe add a TS comment to the image prop that the recommended image width is 400px?
> to clarify "project specific" there is a chance projects are using different image hosting services and they might have a different api for custom image sizes if they do offer custom image sizes

Ingimarsson: I moved it to the `TeamList` slice instead.

barabrian (approved):
LGTM

PR #3150, opened 2021-03-05
fix(service-portal): nationalId should be included in ff request

dabbeg:
## What

NationalId should be included in configcat request

## Why

So we will be able to use user segments for special users until we get the delegation system

## Checklist:


dabbeg:
Need to modify this a little bit, closing until I fix it on monday

PR #3161, opened 2021-03-08
Fix(parental-leave-application): Show async options in review screen, clean up hardcoded data

madebynoam:
## What
Fixes not showing async options in review (https://app.asana.com/0/1182378413629561/1199940349790924/f)
Remove placeholder data for the other parent's periods.

## Why

Prep for QA.

## Checklist:


libs/application/templates/parental-leave/src/types.ts
```diff
@@ -20,3 +20,20 @@ export interface Payment {
     paid: boolean
   }
 }
+
```
jeremybarbet, line 23: I believe we shouldn't have to create these types and we can use the types generated out of the vmst-client through openapi

libs/application/templates/parental-leave/src/fields/InReviewSteps/index.tsx
```diff
@@ -141,61 +109,6 @@ const InReviewSteps: FC<FieldBaseProps> = ({ application, refetch }) => {
                 )}
             </Button>
           </Box>
-          <Box display="inlineBlock">
```
jeremybarbet, line 144: Why all these codes have been removed?

libs/application/templates/parental-leave/src/types.ts
madebynoam: I moved these over from `ParentalLeaveForm.ts` so that I could also use them in the `Review` screen as well. Do you know how we can try to see if they are generated? Was just sitting there before.

libs/application/templates/parental-leave/src/fields/InReviewSteps/index.tsx
madebynoam: I took these out since Edit flow is not yet ready, and I didn't want to leave so much commented code hanging around. So I left it in a branch on my machine, and tried to make main a bit cleaner until the functionality is fully ready. What do you think?

jeremybarbet: Okay, I understand, that's sound good, I would probably add a TODO comment and a link to the github branch so we know where it is and what happened

Alternatively, would be to create a `rfc.index.tsx` or something like this, and keep the review code there, up to you on this 👍

libs/application/templates/parental-leave/src/types.ts
jeremybarbet: I would say it has been created before Bæring creates the VMST client so we probably added it manually at first. Unfortunately, I don't see any automatic way to find that. I guess the easier would be to run the swagger from vmst and see what it returns and check if it's what you need. It's very manual steps. Maybe @baering knows better way

libs/application/templates/parental-leave/src/fields/InReviewSteps/index.tsx
madebynoam: Nice, I like the separate file idea (less branches to keep track off and then we can delete it when ready). I've added that file and a comment to `index.tsx`

libs/application/templates/parental-leave/src/types.ts
madebynoam: Sounds good, I added TODO comment for us to investigate for the next PR.

libs/application/templates/parental-leave/src/fields/InReviewSteps/rfc.index.tsx
```diff
@@ -0,0 +1,224 @@
...
+  )
+}
+
+type stateMapEntry = { [key: string]: reviewSectionState }
+type statesMap = {
+  otherParent: stateMapEntry
+  employer: stateMapEntry
+  vinnumalastofnun: stateMapEntry
+}
+const statesMap: statesMap = {
```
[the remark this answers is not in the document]
madebynoam: Good catch thanks, should be fixed now

baering (approved):

PR #3164, opened 2021-03-08
Don't try to enforce null constraints yet.

valurefugl:
# \<Description\>

## What

## Why

## Screenshots / Gifs

## Checklist:

### DIFF
--- a/libs/auth-api-lib/migrations/20210303105229-client-delegations-non-nullable.js
+++ b/libs/auth-api-lib/migrations/20210303105229-client-delegations-non-nullable.js
@@ -4,11 +4,6 @@ module.exports = {
   up: (queryInterface, Sequelize) => {
     return queryInterface.sequelize.query(`
     BEGIN;
-      ALTER TABLE client
-      ALTER COLUMN supports_delegation SET NOT NULL,
-      ALTER COLUMN supports_legal_guardians SET NOT NULL,
-      ALTER COLUMN supports_procuring_holders SET NOT NULL,
-      ALTER COLUMN prompt_delegations SET NOT NULL;
     COMMIT;
   `)
   },
@@ -16,11 +11,6 @@ module.exports = {
   down: (queryInterface, Sequelize) => {
     return queryInterface.sequelize.query(`
     BEGIN;
-      ALTER TABLE client
-      ALTER COLUMN supports_delegation DROP NOT NULL,
-      ALTER COLUMN supports_legal_guardians DROP NOT NULL,
-      ALTER COLUMN supports_procuring_holders DROP NOT NULL,
-      ALTER COLUMN prompt_delegations DROP NOT NULL;
     COMMIT;
   `)
   },
--- a/package.json
+++ b/package.json
@@ -371,6 +371,7 @@
     "webpack": "4.44.1",
     "xml-crypto": "2.0.0",
     "prismjs": "1.23.0",
-    "urijs": "1.19.6"
+    "urijs": "1.19.6",
+    "elliptic": "6.5.4"
   }
 }


stullifugl (dismissed):

libs/auth-api-lib/migrations/20210303105229-client-delegations-non-nullable.js
```diff
@@ -4,23 +4,13 @@ module.exports = {
   up: (queryInterface, Sequelize) => {
     return queryInterface.sequelize.query(`
     BEGIN;
-      ALTER TABLE client
-      ALTER COLUMN supports_delegation SET NOT NULL,
-      ALTER COLUMN supports_legal_guardians SET NOT NULL,
-      ALTER COLUMN supports_procuring_holders SET NOT NULL,
-      ALTER COLUMN prompt_delegations SET NOT NULL;
```
darrikonn, line 11: This won't have the desired effect for applications that have already been deployed, since we use version controlled migrations, meaning they are executed automatically as a part of the deployment. This version `20210303105229` will already have been executed (at least for dev), so this migration won't change.
Rather have to create a new migration with the desired effect.

darrikonn (changes requested):

libs/auth-api-lib/migrations/20210303105229-client-delegations-non-nullable.js
valurefugl: The desired effect is to have this migration do nothing (because it fails).

valurefugl: Can we resolve this then?

pshomov: @darrikonn the migration in question is not working on Staging but it was just fine on Dev. Adding a new migration is not going to help us with that. I don't like it since it will mean we have to drop the db on Dev but don't see many options. Alternative would be to "fix" the data on Staging and Prod and that is even more work. Will need to discuss how can we have avoided this situation.

darrikonn: Yeah I see.. in that case, dev should be fine, since this migration won't run, and staging will just commit an empty transaction -> but sync the sequelize meta version :+1: .. so yeah, this will work. Good to get the dev vs. staging context :+1:

valurefugl: Agree it shouldn't be a problem on dev.

valurefugl: Not sure about the etiquette for resolving, á ég að gera það?

darrikonn: Buuuut still an issue since dev and staging will have different table structure.

valurefugl: Will be fixed in an upcoming PR.

pshomov: We will run this for dev:

```
npx sequelize-cli db:migrate:undo
```

Or something along those lines.

valurefugl: Why is that necessary?

pshomov: @valurefugl because the migration has already been executed on Dev to completion

valurefugl: But we have prepared a new migration that first sets nulls to default values, and then sets the not null constraints. That should align the environments.
Should actually have done that here, but was in too much of a rush.

PR #3175, opened 2021-03-09
feat(crc): parentB contact info

VilliGunn:
# \<Description\>

https://app.asana.com/0/1199733748699731/1200017649877378

## What

- Added a Contact info screen for parent B
- The info entered on the new screen is used on the overview screen and to sign the PDF digitally
- The contact info will not be prefilled (for now) since we don't have the data for the signed in user. For parent A we get data from the UserProfile data provider but parent B never goes through the external dataprovider screen.

### Note
We probably want to find another solution for the checks we do `application.state === 'draft'` seems like it's not the way we want to go. We should probably be using the assigne or something to decide what data is displayed.

## Why

We need the contact info for parent B to sign the PDF

## Screenshots / Gifs

Screenshot 2021-03-09 at 13 21 58

## Checklist:


libs/application/templates/children-residence-change/src/fields/Overview/index.tsx
```diff
@@ -140,6 +158,9 @@ const Overview = ({ application, setBeforeSubmitCallback }: FieldBaseProps) => {
 
   const controlCode =
     requestFileSignatureData?.requestFileSignature?.controlCode
+  // TODO: Look into if we want to do this in a different way - using the application state seems wrong
+  const contactInfoKey = application.state === 'draft' ? 'parentA' : 'parentB'
```
Rikkiabb, line 162: Totally agree.
We could compare the id of the application (which is parents a ssn) to the ssn of the logged in user. 
It would of course be ideal that we know which role the user is acting in, but don't think we have access to that :/

Rikkiabb (approved):

PR #3181, opened 2021-03-09
Fix(Judicial-system): Fixed update / pdf generation race condition

addi:
https://app.asana.com/0/1108390964845040/1199999639914923

## Checklist:


oddsson (approved):

apps/judicial-system/web/src/routes/Judge/Confirmation/Confirmation.tsx
```diff
@@ -606,6 +609,7 @@ export const Confirmation: React.FC = () => {
             <PdfButton
               caseId={workingCase.id}
               title="Opna PDF þingbók og úrskurð"
+              disabled={isUpdating}
```
gudjong, line 612: Disabling the PDF button is a good idea, but does not resolve the issue reported. That had to do with the signature button.

It is possible that the issue was incorrectly reported. If not, however, then the signature button needs to be disabled as well.

gudjong (changes requested):

gudjong (approved):

PR #3198, opened 2021-03-10
fix(health-insurance) - fix getting bucket files

qdd1:


libs/application/template-api-modules/src/lib/modules/templates/health-insurance/health-insurance.utils.ts
```diff
@@ -87,21 +86,8 @@ export const transformApplicationToHealthInsuranceDTO = (
...
-        const { region, bucket, key } = AmazonS3URI(arrUrl[i])
-        bucketName = bucket
-        attachmentNames.push(key)
-      }
-    } catch (err) {
-      logger.error(`Failed to obtain bucket's name`)
-      throw new Error(`Failed to obtain bucket's name`)
-    }
-  }
+  const attachmentNames = Object.values(application.attachments) ?? []
```
eirikurn, line 89: Fyi, this property is mainly to resolve attachments to absolute S3 URLs.

It's generally better to iterate the files in the answer field, and then look up the S3 URL here. Especially if there is more than one upload field in the form. The answer object also includes the original filename from the user, if you'd want that.

It doesn't make a huge difference in this application. But you should verify that "deleting" an attachment in the form also syncs to the `application.attachments`. Otherwise you might upload files that the user did not intend.

eirikurn (approved):
This should fix the multiple bucket issue.

dd-sdang (approved):

PR #3203, opened 2021-03-10
fix(crc): fix apos

VilliGunn:
https://app.asana.com/0/1199733748699731/1200030056847438


libs/application/templates/children-residence-change/src/fields/components/DescriptionText/DescriptionText.tsx
```diff
@@ -25,12 +25,11 @@ const textOverride = {
 
 const DescriptionText = ({ text, format }: Props) => {
   const { formatMessage } = useIntl()
+  const markdown = formatMessage(text, format)
+  const formattedMarkdown = markdown.replace(/&#39;/g, '&apos;')
```
Rikkiabb, line 29: We should add a comment why we need to handle this special case, right?

Rikkiabb (approved):

PR #3211, opened 2021-03-11
fix(air-discount-scheme): Should not count connecting flights

dabbeg:
## What

The discount quota should not be count connecting flights

## Why

Because connecting flights should have their own quota

## Checklist:


kksteini (approved):

darrikonn (approved):
nice

PR #3212, opened 2021-03-11
feature(crc): upload to district magistrate

Rikkiabb:
# \<Description\>
Uploads signed contract and data to sýslumenn starfskerfi when both parents have signed contract.
Uses application template api to call action to upload data when application is submitted. 
Uses already created syslumenn service to call starfskerfi.

## Checklist:


libs/api/domains/syslumenn/src/lib/client/syslumenn.client.ts
```diff
@@ -20,30 +26,57 @@ export class SyslumennClient {
...
-  private async getToken() {
+  private async login() {
     const config = {
       notandi: this.clientConfig.username,
       lykilord: this.clientConfig.password,
     }
 
     const response: { data: ILogin } = await this.httpService
-      .post(`${this.clientConfig.url}/dev/v1/Innskraning`, config)
+      .post(`${this.clientConfig.url}/v1/Innskraning`, config)
```
Rikkiabb, line 36: I removed `/dev` from this url, since we are reusing the login funcitonality. 
This means we will have to update the value in helm. Both I think we would have to do that in the end anyways.

libs/api/domains/syslumenn/src/lib/client/syslumenn.client.ts
```diff
@@ -20,30 +26,57 @@ export class SyslumennClient {
...
       .toPromise()
 
-    this.accessToken = response.data.audkenni
+    this.id = response.data.audkenni
+    this.accessToken = response.data.accessToken
```
Rikkiabb, line 40: FYI: Since the login response returns a value called `accessToken` which I have to use to upload data then I changed the variable name of the variable formerly known as `accessToken` in this scope to `id`

libs/application/template-api-modules/src/lib/modules/templates/children-residence-change/children-residence-change.service.ts
```diff
@@ -0,0 +1,104 @@
...
+
+    const attachment: Attachment = {
+      name: `Lögheimilisbreyting-barns-${nationalRegistry.nationalId}.pdf`,
+      content: fileContent,
+    }
+
+    const parentA: Person = {
+      name: nationalRegistry.fullName,
+      ssn: nationalRegistry.nationalId,
+      phoneNumber: application.answers.parentA.phoneNumber,
```
VilliGunn, line 61: This is just nitpicking but we could deconstruct like this:
```
const { parentNationalRegistry } = externalData
```

And save a few characters 😄

libs/application/templates/children-residence-change/src/lib/ChildrenResidenceChangeTemplate.ts
```diff
@@ -88,6 +96,36 @@ const ChildrenResidenceChangeTemplate: ApplicationTemplate<
...
+      [ApplicationStates.SUBMITTED]: {
+        meta: {
+          name: applicationName,
+          progress: 0.75,
+          onEntry: {
+            apiModuleAction: 'submitApplication',
+          },
+          roles: [
+            {
+              id: Roles.ParentB,
```
VilliGunn, line 114: Is it not possible to do `id: Roles.ParentA || Roles.ParentB` here?

libs/api/domains/syslumenn/src/lib/client/syslumenn.client.ts
```diff
@@ -20,30 +26,57 @@ export class SyslumennClient {
...
       .get(url)
       .toPromise()
 
     return response.data
   }
+
+  async uploadData(persons: Person[], attachment: Attachment): Promise<string> {
+    await this.login()
+
+    const url = `${this.clientConfig.url}/api/v1/SyslMottakaGogn`
```
Ingimarsson, line 60: Should it not be like this?

```suggestion
    const url = `${this.clientConfig.url}/v1/SyslMottakaGogn`
```

libs/api/domains/syslumenn/src/lib/client/syslumenn.client.ts
Ingimarsson: Nice. Can you make the changes in the helm repo? I guess you'll need to change the host in `portal` and add all the variables in `application-system`

Ingimarsson (changes requested):

libs/api/domains/syslumenn/src/lib/client/syslumenn.client.ts
Rikkiabb: No but it should be ☹️ But as you can see here: https://api.syslumenn.is/dev/swagger/index.html#/ then for whatever reason then this endpoint has and extra `/api/` in it's path.

Ingimarsson: Interesting :thinking:

libs/api/domains/syslumenn/src/lib/client/syslumenn.client.ts
Rikkiabb: Of course 😄  I was thinking of creating a single pr in helm where I update it after this one get's merged. How does that sound? :)

Ingimarsson (approved):
LGTM (syslumenn client)

libs/api/domains/syslumenn/src/lib/client/syslumenn.client.ts
Ingimarsson: Great!

dabbeg (approved):
Looks good!

VilliGunn (approved):
💯

apps/application-system/api/src/environments/environment.ts
```diff
@@ -30,6 +30,12 @@ export default {
     jwtSecret: 'supersecret',
     xRoadBasePathWithEnv: process.env.XROAD_BASE_PATH_WITH_ENV ?? '',
     baseApiUrl,
+    syslumenn: {
+      url: 'https://api.syslumenn.is/dev',
+      username: process.env.SYSLUMENN_USERNAME,
+      password: process.env.SYSLUMENN_PASSWORD,
+    },
```
baering, line 37: Same as above.

libs/api/domains/syslumenn/src/lib/models/dataUpload.ts
```diff
@@ -0,0 +1,32 @@
+export type DataUpload = {
+  id: string
+  data: {
+    id: string
+    name: string
+    person: [Person]
+    attachments: [Attachment]
```
baering, line 7: Is this intentional? Array of one `Person` and one `Attachment`.

baering (commented):
Nice, LGTM 👍   Just have some minor comments/questions before approval.

apps/application-system/api/src/environments/environment.prod.ts
```diff
@@ -25,6 +25,12 @@ export default {
     jwtSecret: process.env.AUTH_JWT_SECRET,
     xRoadBasePathWithEnv: process.env.XROAD_BASE_PATH_WITH_ENV ?? '',
     baseApiUrl,
+    syslumenn: {
+      url: 'https://api.syslumenn.is/api',
+      username: process.env.SYSLUMENN_USERNAME,
+      password: process.env.SYSLUMENN_PASSWORD,
+    },
```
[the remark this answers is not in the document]
Rikkiabb: This should be used for all api's from syslumenn within the application system. 

This is a valid point since this is pretty specific and currently I don't know how many endpoints we will use from syslumenn.

libs/api/domains/syslumenn/src/lib/models/dataUpload.ts
Rikkiabb: 🤦  nice catch 👏

baering (approved):

PR #3214, opened 2021-03-11
fix(air-discount-scheme): fix admin page norlandair is now a regular airline

dabbeg:
## What

Removing all special rules about norlandair on the admin page

## Why

Because norlandair is no longer being booked through Icelandair but instead through their own booking system

## Checklist:


apps/air-discount-scheme/backend/src/app/modules/flight/flight.service.ts
```diff
@@ -114,7 +114,6 @@ export class FlightService {
     return this.flightLegModel.findAll({
       where: {
         ...(body.airline ? { airline: body.airline } : {}),
-        ...(body.cooperation ? { cooperation: body.cooperation } : {}),
```
darrikonn, line 117: I believe we still need this for legacy reasons. Although future flights won't have a cooperation, the previous ones did.

apps/air-discount-scheme/web/screens/Admin/Admin.tsx
```diff
@@ -76,14 +75,7 @@ const Admin: Screen = ({}) => {
...
   const input = {
     ...filters,
-    airline:
-      filters.airline?.value === Airlines.norlandair
-        ? Airlines.icelandair
-        : filters.airline?.value,
-    cooperation:
-      filters.airline?.value === Airlines.norlandair
-        ? Airlines.norlandair
-        : undefined,
```
darrikonn, line 86: Then this needs to be reverted and maybe just add a Selection field that contains a cooperation. Doesn't need the airline trick

apps/air-discount-scheme/web/screens/Admin/components/Summary/Summary.tsx
```diff
@@ -19,22 +19,9 @@ function Summary({ flightLegs, airline: filteredAirline }: PropTypes) {
     arr.reduce((acc, item) => acc + item[key], 0)
 
   const airlines = Object.values(Airlines).filter(
-    (airline) =>
-      airline !== Airlines.norlandair &&
-      (!filteredAirline ||
-        airline ===
-          (filteredAirline === Airlines.norlandair
-            ? Airlines.icelandair
-            : filteredAirline)),
```
darrikonn, line 28: ditto

apps/air-discount-scheme/web/screens/Admin/utils.ts
```diff
@@ -7,9 +7,7 @@ import { FilterInput, financialStateOptions } from './consts'
 
 const getAirlinesAvailableForCSVDownload = (filters: FilterInput) =>
```
darrikonn, line 8: ditto on this file.... since e.g. the first month will differ, could be some flights that still have a cooperation.

darrikonn (changes requested):

apps/air-discount-scheme/backend/src/app/modules/flight/flight.service.ts
dabbeg: If I am not mistaken, this and the things you dittoed are so that Vegagerðin can filter for the `Norlandair` airline and get all the flights that icelandair did in cooperation with Norlandair.

I know we can add the cooperation select box and use that. But we don't really need that feature for Vegagerðin because they only need to send to filter after the airlines that are in direct connection with them.

dabbeg: I am essentially saying that it was an unused feature to filter for Norlandair airline and I don't see the point in adding the select box to be able to do it for legacy data.

darrikonn: mm I see, good point!

darrikonn (approved):

PR #3218, opened 2021-03-11
feature(crc): copy url to clipboard

VilliGunn:
# \<Description\>

https://app.asana.com/0/1199733748699731/1200023303772998

## What

Added a component so users will be able to copy the current url. Only used on the confirmation screen for now.

## Why

So users can copy the application url and share with the other parent

## Screenshots / Gifs

Screenshot 2021-03-11 at 13 30 28

## Checklist:

### DIFF
--- a/libs/application/templates/children-residence-change/src/fields/Confirmation/index.tsx
+++ b/libs/application/templates/children-residence-change/src/fields/Confirmation/index.tsx
@@ -1,16 +1,28 @@
 import React from 'react'
+import { useIntl } from 'react-intl'
 import { Box } from '@island.is/island-ui/core'
-import { confirmation } from '../../lib/messages'
-import { DescriptionText } from '../components'
+import { confirmation, copyUrl } from '../../lib/messages'
+import { DescriptionText, CopyUrl } from '../components'
 
 const Confirmation = () => {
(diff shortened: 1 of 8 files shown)


libs/application/templates/children-residence-change/src/fields/Confirmation/index.tsx
```diff
@@ -1,16 +1,28 @@
...
-      <DescriptionText
-        text={confirmation.general.description}
-        format={{ applicationNumber: '12345' }}
-      />
-    </Box>
+    <>
+      <Box marginTop={3}>
+        <DescriptionText
+          text={confirmation.general.description}
+          format={{ applicationNumber: '12345' }}
```
Rikkiabb, line 14: Not in scope for this PR, but will we have this for parent a? Or will this only be availble when both parents have submitted?

Rikkiabb (approved):

libs/application/templates/children-residence-change/src/fields/Confirmation/index.tsx
VilliGunn: No we probably won't have this for parent A since we don't submit the application to the 'Starfskerfi' until parent B has signed.

PR #3226, opened 2021-03-11
feature(services-search-indexer): Add frontpage and organizations to ES

Ingimarsson:
# Add frontpage and organizations to ES

## What

This PR adds importers for `frontpage`, `organizationPage` and `organizationSubpage`.

This also changes the Contentful sync so that nested entries of depth `<=2` will trigger a sync. Before it was `<=1` i.e. only direct children could trigger a sync.

## Why

This makes these pages load much faster as the request doesn't go to Contentful but is instead cached in ES. We may also want to implement search on organization pages in the future.

## Checklist:


libs/api/domains/cms/src/lib/cms.elasticsearch.service.ts
```diff
@@ -175,6 +177,27 @@ export class CmsElasticsearchService {
     return response ? JSON.parse(response) : null
   }
 
+  async getOrganizationSubpage(
+    index: string,
+    { slug, organizationSlug }: GetOrganizationSubpageInput,
+  ): Promise<OrganizationSubpage | null> {
+    // return a single news item by slug
```
stefanhar, line 184: Is this maybe a wrong comment?

libs/api/domains/cms/src/lib/cms.elasticsearch.service.ts
```diff
@@ -175,6 +177,27 @@ export class CmsElasticsearchService {
...
+    const query = {
+      types: ['webOrganizationSubpage'],
+      tags: [
+        { type: 'slug', key: slug },
+        { type: 'organization', key: organizationSlug },
+      ],
+    }
+    const menuResponse = await this.elasticService.getDocumentsByMetaData(
```
stefanhar, line 192: Is this potentially a copy/paste error? Should this be `menuResponse`?
Should this also maybe use `getSingleDocumentByMetaData`

libs/api/domains/cms/src/lib/cms.elasticsearch.service.ts
```diff
@@ -175,6 +177,27 @@ export class CmsElasticsearchService {
     return response ? JSON.parse(response) : null
   }
 
+  async getOrganizationSubpage(
```
stefanhar, line 180: Should this maybe be called `getSingleOrganizationSubpage` to follow convention?

libs/api/domains/cms/src/lib/search/importers/organizationSubpage.service.ts
```diff
@@ -0,0 +1,56 @@
...
+import { CmsSyncProvider, processSyncDataInput } from '../cmsSync.service'
+import { mapOrganizationSubpage } from '../../models/organizationSubpage.model'
+
+@Injectable()
+export class OrganizationSubpageSyncService
+  implements CmsSyncProvider<IOrganizationSubpage> {
+  processSyncData(entries: processSyncDataInput<IOrganizationSubpage>) {
+    return entries.filter(
+      (entry: Entry<any>): entry is IOrganizationSubpage =>
+        entry.sys.contentType.sys.id === 'organizationSubpage',
```
stefanhar, line 15: We probably want to check if `title` field is set here else elasticsearch will import empty pages for all set locales.
We consider empty title fields as empty documents this is not a problem for Icelandic as field validation is respected when publishing the main locale.

You might also want to check if the entry is a circular structure and take appropriate action such as pruning the data.
This can happen if this entry has related entries as reference and one of the related entries has this entry as reference or in other similar cases.
A circular structure would throw an error in `JSON.stringify` this would cause the entry to not be imported and would log a warning in the importer.

This applies to all importers

libs/api/domains/cms/src/lib/cms.elasticsearch.service.ts
```diff
@@ -175,6 +177,27 @@ export class CmsElasticsearchService {
...
+    index: string,
+    { slug, organizationSlug }: GetOrganizationSubpageInput,
+  ): Promise<OrganizationSubpage | null> {
+    // return a single news item by slug
+    const query = {
+      types: ['webOrganizationSubpage'],
+      tags: [
+        { type: 'slug', key: slug },
+        { type: 'organization', key: organizationSlug },
```
stefanhar, line 189: I understand this as `parentSlug` and `childSlug` ignore if this is not the case.
This approach might not be scalable since you would need one field for each level of the page hierarchy.
We experimented with a custom UI extension called parentSlug to better tackle this type of issue the extension concatenates the parent and child slug allowing for a single non-conflicting field as a page identifier.
This is currently implemented on `aboutPage`(page) and `aboutSubPage`.
We plan to implement this UI extension to articles in a upcoming sprint, It would be useful to get your take on this approach and if it might serve your use case

stefanhar (changes requested):

libs/api/domains/cms/src/lib/cms.elasticsearch.service.ts
Ingimarsson: Yes, sounds good. Then we can also start using ReferenceLink instead of hardcoded URLs.

I tested this UI extension some time ago and it didn't work correctly when the slug of the parent was changed. If we can get that working then it's a good enough for us.

libs/api/domains/cms/src/lib/search/importers/organizationSubpage.service.ts
Ingimarsson: I've added null check for the title fields. I think circular checks are unnecessary because the Contentful schema for these types doesn't have circles.

libs/api/domains/cms/src/lib/search/contentful.service.ts
```diff
@@ -233,22 +233,33 @@ export class ContentfulService {
...
-          if (
-            !alreadyProcessedIds.includes(entry.sys.id) &&
-            environment.indexableTypes.includes(entry.sys.contentType.sys.id)
-          ) {
-            items.push(entry)
-          }
-        })
+
+      // For now we will look for linked entries up to depth 2
+      for (let i = 1; i <= 2; i++) {
```
stefanhar, line 238: It would be optimal to have this go to the same depth as the max include depth in our Contentful service `contentful.service.ts` (this is 10 atm), to prevent the partial sync issues completely.
Did you look into this in regards to added time/pressure when running the sync?

stefanhar (changes requested):

libs/api/domains/cms/src/lib/search/contentful.service.ts
Ingimarsson: The sync endpoint gets really slow as the number of nested entries to look up increases. I tested with ~60 entries and the sync took around 30 seconds. I think it's fine because there is usually only 1 entry when Contentful calls the webhook.

But increasing the depth might be risky because we don't know how fast the number grows with depth, best case would be `1 -> 2 -> 3 -> 4` but there's no guarantee it won't be `1 -> 10 -> 100 -> 1000`

The queries for these pages only go to depth 2 so I think this is safe for now, although not the best solution.

libs/api/domains/cms/src/lib/cms.elasticsearch.service.ts
```diff
@@ -175,6 +177,27 @@ export class CmsElasticsearchService {
...
+  ): Promise<OrganizationSubpage | null> {
+    // return an organization page by organization slug and subpage slug
+    const query = {
+      types: ['webOrganizationSubpage'],
+      tags: [
+        { type: 'slug', key: slug },
+        { type: 'organization', key: organizationSlug },
+      ],
+    }
+    const subpageResponse = await this.elasticService.getDocumentsByMetaData(
```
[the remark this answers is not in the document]
Ingimarsson: I'd like to make a separate PR with this :) it needs to be changed in more places AFAIK, and I'd like to change `getSingleDocumentByMetaData` so that it returns a single document (currently returns the exact same thing as `getDocuments`)

libs/api/domains/cms/src/lib/search/contentful.service.ts
stefanhar: I see your point, true the safe approach is to limit this to current max query depth.
I think this is 3 in our queries atm set by menu: `Grouped menu->menus->menuLinks->childLinks`
That should do until we can apply a more permanent solution to this issue

Ingimarsson: I've changed it from 2 to 3.

stefanhar (approved):
LGTM

PR #3234, opened 2021-03-12
feat: Move all organization slices to RichText

Ingimarsson:
# Move all organization slices to RichText

## What

## Why

## Screenshots / Gifs

## Checklist:


Ingimarsson:
@barabrian not ready for review, but would like your opinion on this :smile: 

Also two questions:

1. What would be the best way to get translations from namespaces into slices?
2. Can I make the margins on the Boxes returned by richText adjustable? Our pages are designed with smaller margins. It's currently hardcoded like this in `RichText.tsx`

```
<Box
  key={slice.id}
  id={slice.id}
  marginBottom={[5, 5, 5, 6]}
  marginTop={[5, 5, 5, 6]}
>
  {renderComponent[slice.__typename]?.(slice, locale)}
</Box>
```

libs/island-ui/contentful/src/lib/OneColumnText/OneColumnText.tsx
```diff
@@ -0,0 +1,45 @@
...
+          <Button
+            icon="arrowForward"
+            iconType="filled"
+            type="button"
+            variant="text"
+          >
+            {link.text}
+          </Button>
+        </Link>
+      )}
```
barabrian, line 42: Not sure if this is intentional, but shouldn't this be using the url?
```suggestion
      {link && (
        <Link href={link.url}>
          <Button
            icon="arrowForward"
            iconType="filled"
            type="button"
            variant="text"
          >
            {link.text}
          </Button>
        </Link>
      )}
```

barabrian (commented):
I missed the part about "not ready for review" 😄   but the comments are minor and might be valid :)

1. Translations and namespaces are not my strong suite maybe @stefanhar could have some insight
2.  At the moment of implementation the lead designer was set on those fixed margins, but it makes sense to make it more flexible. Maybe we should add some kind of mapper to the options object with a default value and based on contentful typename

this is just some pseudo code, not sure if it works but I was thinking something like this

```jsx
export const richText: RichText = (
  documents,
  opt = { renderNode: {}, renderMark: {}, renderComponent: {}, componentMargins: {}},
  locale = 'is',
) => {
```
`...`
```jsx
const componentMargins = {
  DEFAULT: {
    marginBottom: [5, 5, 5, 6],
    marginTop: [5, 5, 5, 6]
  },
  ...opt.componentMargins
}
```
`...`
```jsx
<Box
  key={slice.id}
  id={slice.id}
  marginBottom={
    componentMargins[slice.__typename]
      ? componentMargins[slice.__typename].marginBottom
      : componentMargins.DEFAULT.marginBottom
  }
  marginTop={
    componentMargins[slice.__typename]
      ? componentMargins[slice.__typename].marginTop
      : componentMargins.DEFAULT.marginTop
  }
>
  {renderComponent[slice.__typename]?.(slice, locale)}
</Box>
```

Ingimarsson:
@barabrian @stefanhar maybe I should just add another option for the namespace? So that I would call `richText(slice as SliceType[], namespace)` and then the translations would be available inside the slices.

I'm also having a similar problem with the `linkResolver`. I can't import it inside slices as it creates a circular dependency (`web -> island-ui -> web`. That could be passed through `richText()` as well?

Maybe you have some better ideas? :smiley:

stefanhar:
1. I'm not sure I see the use case for namespaces inside slices. We view slices as dynamic sections that can technically be inserted into any page type of the web, I feel namespaces inside slices might compromise that.
I believe our approach has been to make most texts inside the slice types editable, hence we might not have encountered this issue...yet.
If this model does not fit your use case plz contact me on slack and we can find some alternative approach for this case.

> @barabrian @stefanhar maybe I should just add another option for the namespace? So that I would call `richText(slice as SliceType[], namespace)` and then the translations would be available inside the slices.
> 
> I'm also having a similar problem with the `linkResolver`. I can't import it inside slices as it creates a circular dependency (`web -> island-ui -> web`. That could be passed through `richText()` as well?
> 
> Maybe you have some better ideas? 😃

It seems to me that putting web specific code like the link resolver inside `island-ui` might reduce compatibility of the components.
Maybe `island-ui` should implement its own generic link resolver and give the option to overwrite it with our own (e.g. web resolver) any thoughts on this @barabrian?

Ingimarsson:
That sounds good, I'll put all the texts that need translation into the slice types. I can also store links inside the slice types until a better solution is found.

There is just one translation that is hard to fit inside Contentul, the "Umsókn" tag on articles. I guess I'll just add an `applicationLabel` field to the slice type in Contentful.
image

barabrian (approved):
LGTM

PR #3242, opened 2021-03-15
feat(air-discount-scheme): Connecting flights 

kksteini:
# \<Description\>

Loftbrú needs a quick fix for the incoming changes to the airlines' flight preservation system.

## What

Rewrote the application to work within the confines of the connecting flight issue. The cache now contains connectingDiscountCodes, the cache is more managed instead of deleted at every step.

## Why

Breaking changes for Countrymen who need connecting flights in the middle

## Checklist:

- [ x ] I have commented my code, particularly in hard-to-understand areas


apps/air-discount-scheme/api/src/app/modules/discount/discount.resolver.ts
```diff
@@ -31,7 +31,11 @@ export class DiscountResolver {
           let discount: TDiscount = await backendApi.getDiscount(
             relation.nationalId,
           )
-          if (!discount || discount.expiresIn <= TWO_HOURS) {
+          if (
+            !discount ||
+            discount.expiresIn <= TWO_HOURS ||
+            discount.discountCode.length !== 8
```
dabbeg, line 37: What is this?

kksteini: Ah, good catch.
I currently make the `discountCode` equal to an empty string (and then force regeneration with the length check). 
The cache lives differently now that it has two types of discount codes in it but I have since fixed an issue with `connectionDiscountCodes` being regenerated every time the regular `discountCode` is used. I forgot to change this "hack" as well.

What should be done instead is simply regenerate a `discountCode`, make that code point to the cache and delete the old `discountCode` reference.

Reference:
```
type ConnectionDiscountCodes = {
  code: string
  flightId: string
  flightDesc: string
  validUntil: string
}[]

interface Discount {
  discountCode: string
  connectionDiscountCodes: ConnectionDiscountCodes
  nationalId: string
  expiresIn: number
}
```

libs/island-ui/core/src/lib/Toast/Toast.tsx
```diff
@@ -74,7 +74,7 @@ export const ToastContainer: React.FC<ToastProps> = ({
         pauseOnHover
         transition={Slide}
       />
-      {useKeyframeStyles && <style jsx>{toastKeyframes}</style>}
+      {useKeyframeStyles && <style>{toastKeyframes}</style>}
```
oddsson, line 77: Why are you removing this? There is code in this file (line 13) that explicitly adds the ability for this to be set on the style tag.

oddsson (changes requested):
One thing regarding the island-ui change.

apps/air-discount-scheme/api/src/app/modules/discount/discount.resolver.ts
kksteini: Rewrote this retriggering 'hack' away. Retrigger is now done by invalidating cache pointers

apps/air-discount-scheme/backend/src/app/modules/discount/discount.controller.ts
```diff
@@ -8,12 +16,15 @@ import {
...
 import { DiscountService } from './discount.service'
 import { NationalRegistryService } from '../nationalRegistry'
+import { FlightService } from '../flight'
 
 @Controller('api/private')
 export class PrivateDiscountController {
   constructor(
     private readonly discountService: DiscountService,
     private readonly nationalRegistryService: NationalRegistryService,
+    @Inject(forwardRef(() => FlightService))
```
darrikonn, line 26: Import the module instead. Property based injection should only be used in very rare and specific cases

apps/air-discount-scheme/backend/src/app/modules/discount/discount.module.ts
```diff
@@ -1,12 +1,17 @@
...
 import { CacheModule } from '../cache'
 import { NationalRegistryModule } from '../nationalRegistry'
+import { FlightModule } from '../flight'
 
 @Module({
-  imports: [CacheModule, NationalRegistryModule],
+  imports: [
+    CacheModule,
+    NationalRegistryModule,
+    forwardRef(() => FlightModule),
```
darrikonn, line 13: Where I commented above, import he module instead. I see that you're already doing it.. so you don't need the `Inject` decorator if the flightService provider is correctly exported.

apps/air-discount-scheme/backend/src/app/modules/flight/dto/viewModel.dto.ts
```diff
@@ -30,7 +30,22 @@ export class FlightLegViewModel {
...
+    } else {
+      // Should not end here, CheckFlight should throw an exception instead
+      this.message = 'Something went wrong'
+      this.statusCode = '500'
+    }
+  }
 
+  @ApiProperty()
+  readonly message: string
+  readonly statusCode: string
```
darrikonn, line 47: isn't the statusCode an ApiProperty as well?

apps/air-discount-scheme/backend/src/app/modules/flight/flight.controller.ts
```diff
@@ -47,10 +54,77 @@ import { HttpRequest } from '../../app.types'
 export class PublicFlightController {
   constructor(
     private readonly flightService: FlightService,
+    @Inject(forwardRef(() => DiscountService))
```
darrikonn, line 57: ditto with the injection

dabbeg:
When I try to use swagger I get `TypeError: Window.fetch: HEAD or GET Request cannot have a body.`
I assume because you are using a request body for the GET request, can we put this in query params instead?

apps/air-discount-scheme/backend/src/app/modules/flight/flight.controller.ts
```diff
@@ -47,10 +54,77 @@ import { HttpRequest } from '../../app.types'
...
+  })
+  @ApiResponse({
+    status: 403,
+    description:
+      'The provided discount code is not intended for connecting flights',
+  })
+  @ApiResponse({ type: CheckFlightViewModel })
+  async checkFlightStatus(
+    @Param() params: CheckFlightParams,
+    @Body() flight: CheckFlightBody,
```
dabbeg, line 80: When I try to use swagger I get TypeError: Window.fetch: HEAD or GET Request cannot have a body.
I assume because you are using a request body for the GET request, can we put this in query params instead?

darrikonn (changes requested):
Didn't review the business logic... so stopping here

apps/air-discount-scheme/backend/src/app/modules/flight/flight.controller.ts
```diff
@@ -47,10 +54,77 @@ import { HttpRequest } from '../../app.types'
 export class PublicFlightController {
   constructor(
     private readonly flightService: FlightService,
+    @Inject(forwardRef(() => DiscountService))
     private readonly discountService: DiscountService,
     private readonly nationalRegistryService: NationalRegistryService,
   ) {}
 
+  @Get('discounts/:discountCode/checkFlightStatus')
```
darrikonn, line 62: Needs a more descriptive name for connection flights.

libs/island-ui/core/src/lib/Toast/Toast.tsx
barabrian: This was causing warnings in the console, I did not notice any breaking changes after removing this. If this has a special meaning for a specific framework maybe we should look into moving this to treat? or we could try to find a better way to remove the warning 😄 

image

apps/air-discount-scheme/backend/src/app/modules/flight/flight.controller.ts
kksteini: Moved everything into path parameters

apps/air-discount-scheme/api/src/app/modules/discount/discount.model.ts
```diff
@@ -10,6 +12,8 @@ export class Discount {
   @Field()
   discountCode: string
 
+  @Field((type) => [ConnectionDiscountCode])
+  connectionDiscountCodes: ConnectionDiscountCodes
```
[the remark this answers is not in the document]
kksteini: :+1:

apps/air-discount-scheme/backend/src/app/modules/flight/flight.controller.ts
kksteini: Oh you're right.

What about instead of checkFlightStatus:

- isValidConnectionFlight
- isValid
- canConnect

apps/air-discount-scheme/backend/src/app/modules/flight/dto/viewModel.dto.ts
kksteini: :+1:

apps/air-discount-scheme/backend/src/app/modules/flight/flight.controller.ts
kksteini: I chose `isValidConnectionFlight`

apps/air-discount-scheme/api/src/app/modules/discount/connectionDiscountCode.model.ts
```diff
@@ -0,0 +1,16 @@
+import { Field, ObjectType } from '@nestjs/graphql'
+
+@ObjectType()
+export class ConnectionDiscountCode {
+  @Field()
+  code: string
```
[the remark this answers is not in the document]
darrikonn: Yup.. yeah this is for caching and merging/constructing the objects in apollo

apps/air-discount-scheme/backend/src/app/modules/flight/flight.controller.ts
kksteini: This was proving difficult to circumvent

apps/air-discount-scheme/backend/src/app/modules/flight/flight.service.ts
```diff
@@ -62,6 +66,186 @@ export class FlightService {
...
+        this.hasConnectingFlightPotentialFromFlightLegs(flightLeg, incomingLeg)
+      ) {
+        return true
+      }
+    }
+
+    return false
+  }
+
+  async countThisYearsConnectedFlightsByNationalId(
```
dabbeg, line 151: Same as in the userService, do we need this?

apps/air-discount-scheme/backend/src/app/modules/user/user.service.ts
```diff
@@ -27,6 +27,9 @@ export class UserService {
     } = await this.flightService.countThisYearsFlightLegsByNationalId(
       user.nationalId,
     )
+    const availableConnectionFlights = await this.flightService.countThisYearsConnectedFlightsByNationalId(
```
dabbeg, line 30: Why do we need this, I don't see this used in the frontend and I don't see any reason why the airlines would need this?

libs/island-ui/core/src/lib/Toast/Toast.tsx
barabrian: I'll revert this for now and move the transitions to treat in a separate PR

apps/air-discount-scheme/backend/src/app/modules/flight/flight.service.ts
```diff
@@ -62,6 +66,186 @@ export class FlightService {
...
+
+    return noConnectableFlightLegs - noConnectedFlightLegs
+  }
+
+  async findThisYearsConnectableFlightsByNationalId(
+    nationalId: string,
+  ): Promise<Flight[]> {
+    const currentYear = new Date(Date.now()).getFullYear().toString()
+
+    let flights = await this.flightModel.findAll({
```
dabbeg, line 213: To make the code more concise we could use `findThisYearsFlightsByNationalId` and just use a filter for the `connectable` part.

apps/air-discount-scheme/backend/src/app/modules/flight/flight.service.ts
```diff
@@ -62,6 +66,186 @@ export class FlightService {
...
+            financialState: availableFinancialStates,
+          },
+        },
+      ],
+    })
+
+    // Filter out non-Reykjavík flights
+    flights = flights.filter((flight) => {
+      return (
+        flight.flightLegs.length === 1 &&
```
dabbeg, line 239: Again why `flightlegs.length === 1`?

dabbeg (changes requested):
Take a look at my comments :)

libs/island-ui/core/src/lib/Toast/Toast.tsx
barabrian: @oddsson https://github.com/island-is/island.is/pull/3264

apps/air-discount-scheme/backend/migrations/20210311225505-add-marked-flight-column.js
```diff
@@ -0,0 +1,19 @@
+'use strict'
+
+module.exports = {
+  up: (queryInterface) => {
+    return queryInterface.sequelize.query(`
+      BEGIN;
+
+        ALTER TABLE flight ADD COLUMN connectable BOOLEAN NOT NULL DEFAULT TRUE;
```
dabbeg, line 8: I am wondering if storing something like `connectingFlightId` instead of `connectable` would be better.
We could use it the same by checking if it is null and we would also get the connection between the flights stored in the db

apps/air-discount-scheme/backend/src/app/modules/flight/flight.controller.ts
```diff
@@ -47,10 +53,80 @@ import { HttpRequest } from '../../app.types'
...
+      destination: params.destination,
+      date: new Date(Date.parse(params.date.toString())),
+    }
+
+    if (!discount) {
+      throw new BadRequestException('Discount code is invalid')
+    }
+
+    const user = await this.nationalRegistryService.getUser(discount.nationalId)
+    if (!user) {
```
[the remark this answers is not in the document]
kksteini: :+1:

apps/air-discount-scheme/backend/src/app/modules/user/user.service.ts
kksteini: Ah, that's residual code from an approach that I didn't end up using. Good catch :+1:

apps/air-discount-scheme/backend/src/app/modules/flight/flight.controller.ts
```diff
@@ -47,10 +53,80 @@ import { HttpRequest } from '../../app.types'
...
+    status: 403,
+    description:
+      'The provided discount code is not intended for connecting flights',
+  })
+  @ApiResponse({ type: CheckFlightViewModel })
+  async checkFlightStatus(
+    @Param() params: CheckFlightParams,
+    @Req() request: HttpRequest,
+  ): Promise<CheckFlightViewModel> {
+    const discount = await this.discountService.getDiscountByDiscountCode(
```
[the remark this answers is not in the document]
kksteini: True. Much cleaner that way :+1:

apps/air-discount-scheme/backend/src/app/modules/flight/flight.service.ts
kksteini: Residual code. No we don't :+1:

apps/air-discount-scheme/backend/src/app/modules/flight/flight.service.ts
kksteini: Good point :+1:

apps/air-discount-scheme/backend/src/app/modules/flight/flight.controller.ts
```diff
@@ -94,15 +171,66 @@ export class PublicFlightController {
     if (flightLegsLeft < flight.flightLegs.length) {
       throw new ForbiddenException('Flight leg quota is exceeded')
     }
+
+    let connectingFlight = false
+    let connectingId = undefined
+
+    if (flight.flightLegs.length === 1) {
```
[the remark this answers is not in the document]
kksteini: Whoops. I remember having gotten it in my head from previous meetings that they were going to restrict flights to a single flightleg per request. In retrospect that's absurd.

Nice catch :+1: 
Fixed

apps/air-discount-scheme/backend/migrations/20210311225505-add-marked-flight-column.js
kksteini: Yeah it would make sense. Is this something we want to do for phase 2?

apps/air-discount-scheme/backend/src/app/modules/flight/dto/body.dto.ts
```diff
@@ -91,4 +90,18 @@ export class GetFlightLegsBody implements FlightLegsInput {
...
+  @ApiProperty()
+  readonly origin!: string
+
+  @IsString()
+  @ApiProperty()
+  readonly destination!: string
+
+  @IsISO8601()
+  @ApiProperty()
+  readonly date!: string
```
dabbeg, line 104: The ts type should be a Date

apps/air-discount-scheme/backend/src/app/modules/discount/discount.service.ts
```diff
@@ -63,13 +71,90 @@ export class DiscountService {
...
+    const previousCache = await this.getCache<CachedDiscount>(previousCacheId)
+
+    let previousFlightIds: string[] = []
+
+    if (previousCache?.connectionDiscountCodes) {
+      connectionDiscountCodes = previousCache.connectionDiscountCodes
+      previousFlightIds = connectionDiscountCodes.map((cdc) => cdc.flightId)
+    }
+
+    const connectableFlightCounts = connectableFlights.length
```
dabbeg, line 93: If I am not mistaken, this could be simplified a lot by creating the connection discountCode when the flight is created.
So the POST /flights endpoint would check if the flight is connectable and if so it would call the discountService and create the connectionDiscountCode for that flight.

Then we would need all the checks here

apps/air-discount-scheme/backend/src/app/modules/flight/dto/body.dto.ts
kksteini: :+1:

apps/air-discount-scheme/backend/src/app/modules/flight/flight.service.ts
```diff
@@ -36,6 +36,10 @@ const availableFinancialStates = [
   financialStateMachine.states[States.sentDebit].key,
 ]
 
+export const CONNECTING_FLIGHT_GRACE_PERIOD = 48 * (1000 * 60 * 60) // 48 hours in milliseconds
+export const REYKJAVIK_FLIGHT_CODES = ['RVK', 'REK']
+export const AKUREYRI_FLIGHT_CODES = ['AK']
+
```
dabbeg, line 42: One thing we need to add here is `ALLOWED_CONNECTING_FLIGHT_CODES = ['VPN', 'GRY', 'THO']`

Connecting flights are only allowed to these places

apps/air-discount-scheme/backend/src/app/modules/flight/flight.service.ts
```diff
@@ -36,6 +36,10 @@ const availableFinancialStates = [
   financialStateMachine.states[States.sentDebit].key,
 ]
 
+export const CONNECTING_FLIGHT_GRACE_PERIOD = 48 * (1000 * 60 * 60) // 48 hours in milliseconds
+export const REYKJAVIK_FLIGHT_CODES = ['RVK', 'REK']
```
dabbeg, line 40: It is 'RKV' not 'RVK'

apps/air-discount-scheme/backend/src/app/modules/discount/discount.service.ts
dabbeg: This is something that should be looked at in phase 2

apps/air-discount-scheme/backend/src/app/modules/flight/flight.controller.ts
```diff
@@ -94,15 +164,74 @@ export class PublicFlightController {
...
     }
+
+    let connectingFlight = false
+    let connectingId = undefined
+
+    const flightLegCount = flight.flightLegs.length
+
+    const incomingLeg = {
+      origin: flight.flightLegs[0].origin,
+      destination: flight.flightLegs[flightLegCount - 1].destination,
```
dabbeg, line 175: It is probably better to take the first flightleg here also, so we can check in the flightservice if the destination matches the ALLOWED_CONNECTING_FLIGHT_CODES

kksteini: It is assumed that flightLegs for a single flight are given chronologically, right? 
That's why the destination is pulled from `flightLegs[flightLegs.length - 1].destination`
Does the intermediate airport matter in the cases where a flight has more than one flightLeg?

apps/air-discount-scheme/backend/src/app/modules/flight/flight.service.ts
kksteini: Vó, vel spottað

apps/air-discount-scheme/backend/src/app/modules/flight/flight.controller.ts
kksteini: Re-orders flightLegs chronologically. 
Checks whether firstLeg origin is the same as lastLeg destination to detect round-trips.

Functionality remains the same if no round-trip detected.

If roundtrip detected:
The flightLeg check was going through every leg of the connectedId (Reykjavík flight)  so we can send the lastLeg to that too when roundtrips are detected.
That way we check that when your returnTrip flight from norlandair returns to Akureyri, whether it has a corresponding return flightLeg from icelandair.

apps/air-discount-scheme/backend/src/app/modules/flight/flight.controller.ts
```diff
@@ -86,23 +223,63 @@ export class PublicFlightController {
...
+      }
     }
+
+    if (!hasReykjavik) {
+      connectingFlight = await this.validateConnectionFlights(
+        discount,
+        params.discountCode,
+        flight.flightLegs as FlightLeg[],
+      )
+      if (connectingFlight) {
```
dabbeg, line 252: You can simplify this if `this.validateConnectionFlights` would return `connectingId` instead of a boolean

apps/air-discount-scheme/backend/src/app/modules/flight/flight.controller.ts
```diff
@@ -86,23 +223,63 @@ export class PublicFlightController {
...
-      discount.nationalId,
-    )
-    if (flightLegsLeft < flight.flightLegs.length) {
-      throw new ForbiddenException('Flight leg quota is exceeded')
+    let connectingFlight = false
+    let connectingId = undefined
+
+    let hasReykjavik = false
+    let hasAkureyri = false
+    for (const flightLeg of flight.flightLegs) {
```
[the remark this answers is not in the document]
kksteini: Oh that's neat. Thank you for that suggestion.

apps/air-discount-scheme/backend/src/app/modules/flight/flight.controller.ts
kksteini: A good point :+1:

apps/air-discount-scheme/backend/src/app/modules/flight/flight.controller.ts
```diff
@@ -86,23 +223,63 @@ export class PublicFlightController {
...
+      }
+      if (
+        AKUREYRI_FLIGHT_CODES.includes(flightLeg.origin) ||
+        AKUREYRI_FLIGHT_CODES.includes(flightLeg.destination)
+      ) {
+        hasAkureyri = true
+      }
     }
+
+    if (!hasReykjavik) {
```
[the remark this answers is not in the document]
kksteini: That is true. Fixed :+1:

apps/air-discount-scheme/backend/src/app/modules/flight/flight.controller.ts
```diff
@@ -86,23 +222,54 @@ export class PublicFlightController {
...
+
+    const hasAkureyri = flight.flightLegs.some(
+      (flightLeg) =>
+        AKUREYRI_FLIGHT_CODES.includes(flightLeg.origin) ||
+        AKUREYRI_FLIGHT_CODES.includes(flightLeg.destination),
     )
-    if (flightLegsLeft < flight.flightLegs.length) {
-      throw new ForbiddenException('Flight leg quota is exceeded')
+
+    if (!hasReykjavik && hasAkureyri) {
```
dabbeg, line 240: It would probably be best to do
```
if (hasReykjavik) {
 countThisYears....
} else if (hasAkureyri) {
 isConnectingFlight
} else {
throw error
}
```

I was testing this and saw that it is not a requirement for non connecting flights to be from reykjavik

apps/air-discount-scheme/backend/src/app/modules/flight/flight.service.ts
```diff
@@ -62,6 +67,115 @@ export class FlightService {
...
+
+  async findThisYearsConnectableFlightsByNationalId(
+    nationalId: string,
+  ): Promise<Flight[]> {
+    const flights = await this.findThisYearsFlightsByNationalId(nationalId)
+    // Filter out non-Reykjavík flights
+    return flights.filter((flight) => {
+      return (
+        flight.connectable &&
+        (REYKJAVIK_FLIGHT_CODES.includes(flight.flightLegs[0].origin) ||
```
dabbeg, line 171: You should check here also if flightlegs connect to akureyri

apps/air-discount-scheme/backend/src/app/modules/flight/flight.controller.ts
kksteini: I did it somewhat similarly for it wasn't as simple as that. I changed a parameter for how flights are created so that we are discoupling `discountService.useDiscount()` and `flightService.create()` semantically in the `flightController`.

`isConnectingFlight` was being used also for marking whether a flight is connectable. I replaced it with `isConnectable` for the `flightService.create()` in order to increase clarity and differentiate between intended functionality.

apps/air-discount-scheme/backend/src/app/modules/flight/flight.service.ts
kksteini: Yuss :+1:
Good point

dabbeg (approved):
Approved, this is good enough for now. We will need to refactor some of this in phase2 though.

oddsson (approved):

PR #3246, opened 2021-03-15
fix: codegen issue for '[object Object]'

jeremybarbet:
## What

wrong types generated


scripts/schemas.js
```diff
@@ -66,5 +66,4 @@ const main = async () => {
   }
 }
 
-// this is a main function
```
sindrig, line 69: so this isn't a main function anymore? :joy:

sindrig (approved):
LGTM

scripts/schemas.js
jeremybarbet: Not anymore. But next time we need to force the rerun of schema generation, it will definitely be the main function again

PR #3249, opened 2021-03-15
feature(application-system-api): Allow template developer to configure template api module actions

baering:
This PR refactors how template API actions are defined and how they are invoked.

## What

The goal of this refactor is to
- be able to (optionally) persist successful and unsuccessful responses from these actions to the application
- move the responsibility of how to configure and/or handle actions to the developers of applications

Developers now have access to both `onEntry` and `onExit` template API actions to use on states.

See the [updated README for template API modules](https://github.com/island-is/island.is/blob/c0f60b94ae9108a01222d4ef992d2ee992c7c9b6/libs/application/template-api-modules/README.md#5-invoke-your-new-module-actions-from-the-template-state-machine) for more information on the new functionality.

## Why

To allow application developers to react to data received by actions:
- Display data from these actions to the user
- Block state transitions on errors
- Revert state changes on errors

## Screenshots / Gifs

Here is an example where an application is submitted and it receives back a dynamic ID that should be displayed to the user:

https://user-images.githubusercontent.com/6312838/111297570-a855e900-8645-11eb-8c5a-ea93f360f351.mov

The dynamic id can be fetched from `application.externalData` which now looks like this:
```json
{
    "createApplication": {
        "data": {
            "id": 1337
        },
        "date": "2021-03-16T10:51:31.285Z",
        "status": "success"
    }
}
```

Should the action fail the changes to `answers` / `state` / `assignees` will be (optionally) reverted and the error will be (optionally) persisted into `application.externalData.createApplication.data`

## Checklist:


apps/application-system/api/src/app/modules/application/application.controller.ts
```diff
@@ -404,86 +426,203 @@ export class ApplicationController {
...
+        updatedApplication.id,
+        updatedApplication.externalData,
+        newExternalDataEntry,
+      )
+
+      updatedApplication = {
+        ...updatedApplication,
+        externalData: {
+          ...updatedApplication.externalData,
+          ...withExternalData.externalData,
```
jeremybarbet, line 499: I might be wrong on this, but there is no way the API action is returning an object it shouldn't handle and overwrite it by mistake? if API module should return objA and is returning objA and objB, won't objB overwritten?

apps/application-system/api/src/app/modules/application/application.controller.ts
```diff
@@ -404,86 +426,203 @@ export class ApplicationController {
...
+      return {
+        updatedApplication,
+        hasError: true,
+        error: actionResult.error,
+      }
+    }
 
-    const updatedApplication = update.updatedApplication as BaseApplication
-    const newStateOnEntry = helper.getStateOnEntry(newState)
+    return {
```
jeremybarbet, line 512: If `actionResult.success` is false, but `throwOnError` is false, we then return hasError: false?

jeremybarbet (commented):
Great job! I just have some minor comment

apps/application-system/api/src/app/modules/application/application.controller.ts
baering: Yes, since the developer specifically chose that this action should not throw an error we ignore all errors. I just included this config variable to try and support anything that may come up, but I doubt it will be used much since you likely only want to continue if your actions were successful.

jeremybarbet (approved):
LGTM

apps/application-system/api/src/app/modules/application/application.controller.ts
baering: This happens after updating the application's external data in the database and is only updating the `updatedApplication` externalData with whatever it confirmed to be updated in the database.

There are two cases where unrelated external data would get updated:
- The developer named the action the same as the id of an existing DataProvider
- The developer chose a custom `externalDataId` with the same id as an existing DataProvider

PR #3258, opened 2021-03-16
fix(island-ui): Expand props signature to also accept ReadonlyArrays…

maranomynet:
# \<Description\>

`ReadonlyArray` is generally preferable to mutable `Array` in signatures for component props and function parameters (because its more permissive), unless they actually intend to mutate the arrays passed to them.

## What

This updates a couple of cases I ran into, but is not an exhaustive list of potential places this sort of change might be beneficial.

## Why

Because if I am working with `ReadonlyArray`s and try and pass them into these components, I get a type error – even though these components don't mutate their inputs and have no intention of ever doing so.

## Checklist:


oddsson (approved):
Nice! Didn't know this was a thing 🙏

PR #3260, opened 2021-03-16
fix(web): Add missing export of `Image` to web/components

maranomynet:
I noticed what looked like an accidental omission.

## Checklist:


apps/web/components/real.ts
```diff
@@ -65,3 +65,4 @@ export * from './InstitutionPanel/InstitutionPanel'
 export * from './InstitutionsPanel/InstitutionsPanel'
 export * from './Organization/'
 export * from './ApiCatalogueFilter/ApiCatalogueFilter'
+export * from './Image'
```
jeremybarbet, line 68: Looks like the component is never imported in the web project, not sure if it has been deprecated in favor of the contentful Image component or just not used anymore but will be in the future

maranomynet: FWIW I'm using it on a local feature branch, ATM.
I can downgrade to a plain `` if needed, though.

jeremybarbet (approved):

PR #3261, opened 2021-03-16
feat(web): Add type for components to signal they accept `NoChildren`

maranomynet:
## What

This introduces an optional prop type to add to components (and a couple of examples of it being used) to signal that they don't accept `children`.

## Why

This way we leverage TypeScript to help developers avoid simple JSX/React mistakes while using our components.

It's nice to get a TS warning when you pass children to a component that has no intention of using/displaying them.

## Checklist:


maranomynet:
Better just not use `FC<Props>` typing, I guess.

https://github.com/facebook/create-react-app/pull/8177#issue-353062710

PR #3266, opened 2021-03-16
fix(crc): update syslumenn data contract

Rikkiabb:
# \<Description\>
Sends extra data to syslumenn.

## Checklist:


libs/api/domains/syslumenn/src/lib/client/models/dataUpload.ts
```diff
@@ -14,6 +14,7 @@ export interface IDataUpload {
     skeytaHeiti: string
     adilar: ChildrenTransferPerson[]
     attachments: File[]
+    gagnaMengi: object
```
VilliGunn, line 17: According to the documentation from Þjónustugátt starfskerfa this is a key: string object, should we type it in the same way instead of just `object`?

``` 
{ [key: string]: string }
```

libs/api/domains/syslumenn/src/lib/client/models/dataUpload.ts
```diff
@@ -45,6 +46,7 @@ export function constructUploadDataObject(
   id: string,
   persons: Person[],
   attachment: Attachment,
+  extraData: object,
```
Ingimarsson, line 49: Should this maybe be an optional argument?

libs/api/domains/syslumenn/src/lib/client/syslumenn.client.ts
```diff
@@ -54,7 +54,11 @@ export class SyslumennClient {
     return response.data
   }
 
-  async uploadData(persons: Person[], attachment: Attachment): Promise<string> {
+  async uploadData(
+    persons: Person[],
+    attachment: Attachment,
+    extraData: object,
```
Ingimarsson, line 60: Same as last comment.

libs/api/domains/syslumenn/src/lib/client/models/dataUpload.ts
Ingimarsson: A more specific type would also be nice if that's possible :)

Ingimarsson (changes requested):

libs/api/domains/syslumenn/src/lib/client/models/dataUpload.ts
Rikkiabb: True, this was lazy of me 🙈 will fix

libs/api/domains/syslumenn/src/lib/client/models/dataUpload.ts
Rikkiabb: 1. Let's change it to an optional argument if someone uses this endpoint in the feature without using the extra data. Since all callers are using extra data now then I would like to have it as mandatory. 
2. Fixed 😄

VilliGunn (approved):
💯

Ingimarsson (approved):
LGTM

PR #3270, opened 2021-03-16
Zendesk instead of email for contact form

stjanilofts:
Today the contact form sends a regular email to the island.is support email address. This instead creates a Zendesk ticket and submits it.

First checks if the user with the email exists. If the user does not exist it creates it. Then it submits a ticket with that user as the requester.

## Checklist:

### DIFF
--- a/libs/api/domains/communications/src/lib/communications.service.ts
+++ b/libs/api/domains/communications/src/lib/communications.service.ts
@@ -90,7 +90,7 @@ export class CommunicationsService {
 
     // If we did not find an existing user we will create a new one
     if (!user) {
-      let identities = []
+      const identities = []
 
       if (input.phone) {
         identities.push({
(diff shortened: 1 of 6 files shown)


libs/api/domains/communications/src/lib/communications.service.ts
```diff
@@ -44,4 +49,112 @@ export class CommunicationsService {
...
+      )
+    }
+
+    let user =
+      searchUserResponse.data.results.length > 0 &&
+      searchUserResponse.data.results[0]
+
+    // If we did not find an existing user we will create a new one
+    if (!user) {
+      let identities = []
```
barabrian, line 93:
```suggestion
      const identities = []
```

barabrian (approved):
LGTM

PR #3279, opened 2021-03-17
fix(application-system): conditional sections and subsections

VilliGunn:
# \<Description\>

https://www.notion.so/93316a1505b14dfc9c1feffb4795b73b?v=82697e5acbec44de8d9710f21d0c02c7&p=6d983c292fc44f88bd8a4e227d10538d

## What

Filtering sections and subSections using the `shouldShowFormItem` that is currenly only used for form fields.

## Why

According to the specs, conditional rendering of sections and subSections should be possible. 

## Checklist:


libs/application/core/src/lib/formUtils.ts
```diff
@@ -101,27 +102,39 @@ export const getFormNodeLeaves = (node: FormNode): FormLeaf[] => {
...
-export function getSubSectionsInSection(section: Section): SubSection[] {
+export function getSubSectionsInSection(
+  section: Section,
+  answers: FormValue,
+  externalData: ExternalData,
+): SubSection[] {
   const subSections: SubSection[] = []
-  section.children.forEach((child) => {
-    if (child.type === FormItemTypes.SUB_SECTION) {
+  section?.children.forEach((child) => {
```
jeremybarbet, line 125: if `section` can be undefined, then `forEach` will fail because of not being an array. Add extra check `(section?.children ?? [])`, isn't it?

jeremybarbet (approved):
LGTM

PR #3284, opened 2021-03-17
feat(application-system): change base paths

jeremybarbet:
## What

- Configure paths for umsoknir
- Add slugs for each application templates
- Quick design to follow the service portal design for the cards

Screenshot 2021-03-17 at 16 01 55


apps/application-system/form/src/hooks/useAuth.ts
```diff
@@ -1,28 +1,30 @@
...
+
         dispatch({
           type: ActionType.SET_USER_FULFILLED,
           payload: user,
         })
       }
     } catch {
       await userManager.signinRedirect({
-        state: { redirect: `${pathname}${search}` },
+        state: window.location.pathname,
```
baering, line 27: I'm not sure why the `search` was in there, but I guess it is fine to remove it?

baering: It was added in [#2291](https://github.com/island-is/island.is/pull/2291/)

baering, line 27: `search` was added in [#2291](https://github.com/island-is/island.is/pull/2291).

Are you sure we can remove it and it won't affect the sign-in flow?

libs/application/core/src/types/ApplicationTypes.ts
```diff
@@ -12,16 +12,49 @@ export enum ApplicationTypes {
...
+  [ApplicationTypes.PASSPORT]: {
+    slug: 'vegabref',
+    translation: 'pp.application',
+  },
+  [ApplicationTypes.DRIVING_LESSONS]: {
+    slug: 'okunam',
+    translation: 'dl.application',
+  },
+  [ApplicationTypes.DRIVING_LICENSE]: {
+    slug: 'okuskilriki',
```
baering, line 29: `okuskirteini`

apps/application-system/form/src/routes/Applications.tsx
```diff
@@ -14,24 +14,27 @@ import {
...
 import useAuth from '../hooks/useAuth'
 
 export const Applications: FC = () => {
-  const { type } = useParams<{ type: ApplicationTypes }>()
+  const { slug } = useParams<{ slug: string }>()
   const history = useHistory()
   const { userInfo } = useAuth()
   const { formatMessage } = useLocale()
   const nationalRegistryId = userInfo?.profile?.nationalId
+  const type = getTypeFromSlug(slug)
```
baering, line 37: `getTypeFromSlug` can return undefined, in that case we should probably have something like:
* `hasValidType = typeof type !== 'undefined'`
* pass `skip: !hasValidType` into `useQuery`
* `if (!hasValidType) return <NotFound />`

libs/application/core/src/lib/slugUtil.ts
```diff
@@ -0,0 +1,14 @@
+import {
+  ApplicationConfigurations,
+  ApplicationTypes,
+} from '../types/ApplicationTypes'
+
+export const getTypeFromSlug = (slug: string) =>
+  Object.entries(ApplicationConfigurations).find(
+    ([_, item]) => item.slug === slug,
+  )?.[0] as ApplicationTypes | undefined
```
baering, line 9: I had to look at this code for a while before realising why you were using Object.entries not Object.values 😄 

I'm wondering if we should just simplify it to be:
```typescript
for (const [key, value] of Object.entries(ApplicationConfigurations) {
  if (value.slug === slug) {
    return key
  }
}

return undefined
```

So that it is more readable

libs/application/core/src/lib/slugUtil.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export const getSlugFromType = (type: ApplicationTypes) =>
+  Object.entries(ApplicationConfigurations).find(([id]) => id === type)?.[0] as
+    | string
+    | undefined
```
baering, line 14: There is a bug here, you're returning `[0]` when you would want to return `[1].slug`

This could also use a similar refactor as the above function to end up like:
```typescript
for (const [key, value] of Object.entries(ApplicationConfigurations) {
  if (type === key) {
    return value.slug
  }
}

return undefined
```

If you find the type you're returning `[0]` which is the type, right?

libs/localization/src/lib/useApplicationNamespaces.tsx
```diff
@@ -1,12 +1,16 @@
...
 } from '@island.is/application/core'
 
 import { useNamespaces } from './useNamespaces'
 
-export const useApplicationNamespaces = (type: ApplicationTypes) => {
-  const array = ['application.system', ApplicationTranslations?.[type] ?? null]
+export const useApplicationNamespaces = (type?: ApplicationTypes) => {
+  const array = [
+    'application.system',
+    type ? ApplicationConfigurations?.[type].translation ?? null : null,
```
baering, line 11: In theory this could end up as `undefined` if `ApplicationConfigurations?.[type]` is not defined.

baering (commented):
Nice, well done. I added a few comments.

libs/application/core/src/types/ApplicationTypes.ts
jeremybarbet: I checked with the team who worked on it, they told me `okuskilriki` was correct

baering: [https://island.is/okuskirteini](https://island.is/okuskirteini)

🤷

apps/application-system/form/src/index.html
```diff
@@ -2,8 +2,29 @@
...
+    <meta
+      property="og:image"
+      content="https://island.is/island-fb-1200x630.png"
+    />
+    <meta property="og:image:width" content="1200" />
+    <meta property="og:image:height" content="630" />
+    <meta property="twitter:card" content="summary_large_image" />
+    <meta property="twitter:url" content="http://island.is/umsoknir/" />
+    <meta property="twitter:title" content="Umsóknir - Ísland.is" />
+    <meta property="twitter:description" content="Umsóknir Ísland.is - TODO." />
```
baering, line 23: We'll tackle this in a separate PR.

baering (approved):
LGTM 🚀 

Let's notify other teams that the urls have changed when this is merged.

dd-sdang (approved):
health insurance looks good :)

obmagnusson (approved):
LGTM

VilliGunn (approved):

PR #3286, opened 2021-03-17
feat(web): zendesk ticket service

stjanilofts:
Today the contact form sends a regular email to the island.is support email address. This instead creates a Zendesk ticket and submits it.

First checks if the user with the email exists. If the user does not exist it creates it. Then it submits a ticket with that user as the requester.

## Checklist:


libs/zendesk-service/src/lib/zendesk.module.ts
```diff
@@ -0,0 +1,23 @@
...
+} from './zendesk.service'
+
+export class ZendeskModule {
+  static register(options: ZendeskServiceOptions): DynamicModule {
+    return {
+      module: ZendeskModule,
+      providers: [
+        {
+          provide: ZENDESK_OPTIONS,
+          useFactory: () => options,
```
stefanhar, line 16: Maybe useValue is better suited bere?

stefanhar (commented):
we should update `.github/CODEOWNERS` file for these new libraries

.github/CODEOWNERS
```diff
@@ -17,6 +17,7 @@
 /libs/api/domains/cms/                                                                        @island-is/kosmos-kaos @island-is/aranja
 /libs/application/templates/party-letter/                                                     @island-is/kosmos-kaos
 /libs/clients/rsk/                                                                            @island-is/kosmos-kaos
+/libs/zendesk-service/                                                                        @island-is/kosmos-kaos
```
stefanhar, line 20: Maybe we should add the communications domain to here as well?

libs/clients/zendesk/src/lib/zendesk.service.ts
```diff
@@ -0,0 +1,148 @@
...
+}
+
+export interface ZendeskServiceOptions {
+  email: string
+  token: string
+  subdomain: string
+}
+
+export class ZendeskService {
+  #api: string
```
jeremybarbet, line 27: Is the `#` doing something special?

stjanilofts: not really, I will remove the #'s..

stefanhar (approved):
LGTM

jeremybarbet (approved):
LGTM

PR #3288, opened 2021-03-18
feat(service-portal): configure feature flag for applications module

jeremybarbet:
## What

- Configure feature flag for Application modules on the service portal
- Added `isServicePortalApplicationsModuleEnabled` to ConfigCat with ON on dev and staging


obmagnusson (approved):
LGTM

PR #3307, opened 2021-03-19
feat(education): Move mms client to lib/clients and make ready for grade api

darrikonn:


libs/api/domains/education/src/lib/education.service.ts
```diff
@@ -24,7 +45,105 @@ export class EducationService {
...
+
+    return this.s3Service.uploadFileFromStream(responseStream, {
+      fileName: uuid(),
+      bucket: this.config.fileDownloadBucket,
+    })
+  }
+
+  async getFamily(nationalId: string) {
+    const family = await this.nationalRegistryApi.getMyFamily(nationalId)
+    return family.filter(
```
dabbeg, line 62: The nationalregistry api domain has this logic already, but we need to move it to the nationalregistry client so it can be reused here. I want to do that in another slice.

libs/api/domains/education/src/lib/education.service.ts
```diff
@@ -24,7 +45,105 @@ export class EducationService {
...
+      (familyMember) =>
+        nationalId === familyMember.Kennitala ||
+        (!['1', '2'].includes(familyMember.Kyn) &&
```
darrikonn, line 65: There are more variations of `Kyn`. Probably missing `7`???

libs/api/domains/education/src/lib/education.service.ts
```diff
@@ -24,7 +45,105 @@ export class EducationService {
...
+
+  mapLanguageGrade(grade: LanguageGrade) {
+    return {
+      ...this.mapBaseGrade(grade),
+      readingGrade: this.mapGrade(grade.lesskilningur),
+      grammarGrade: this.mapGrade(grade.malnotkun),
+    }
+  }
+
+  mapMathGrade(grade: MathGrade) {
```
darrikonn, line 126: Your mappings and `getFamily` are missing the return type. Thought we had strict typing.

libs/api/domains/education/src/lib/education.utils.ts
```diff
@@ -0,0 +1,8 @@
+export const getYearInterval = (dates: string[]) => {
```
darrikonn, line 1: ditto on return type

libs/api/domains/education/src/lib/graphql/main.resolver.ts
```diff
@@ -38,18 +33,33 @@ export class MainResolver {
...
-      bucket: this.config.fileDownloadBucket,
-    })
     if (url === null) {
       throw new ApolloError('Could not create a download link')
     }
     return { url }
   }
+
+  @Query(() => [ExamOverview])
+  educationExamOverviews(@CurrentUser() user: User): Promise<ExamOverview[]> {
```
darrikonn, line 47: I believe the naming should incline that the user's family member's exam overviews are also fetched.

Otherwise agree that the family handling should be done in the service :+1:

darrikonn: Then it's the same for the name of the query.

darrikonn (commented):
Nice one. Great work.

Couple of comments, otherwise the business logic is solid and the code is solid as well.

libs/api/domains/education/src/lib/graphql/main.resolver.ts
dabbeg: Not sure I understand, are you saying we should use the naming `ExamFamilyOverview` instead of `ExamOverview`?

darrikonn: Yeah something like that.

libs/api/domains/education/src/lib/education.utils.ts
dabbeg: Typescript should be able to determine the return type here by itself

libs/api/domains/education/src/lib/graphql/main.resolver.ts
darrikonn: Both for the query name and the service method name

libs/api/domains/education/src/lib/education.service.ts
dabbeg: Good point, I will fix and check if I can turn the strict on

libs/api/domains/education/src/lib/education.service.ts
dabbeg: Ah, yes 7 is missing, I will add it here.
It is not a part of the nationaregistry api domain either, we will need to fix that in the next slice as well.

libs/api/domains/education/src/lib/education.service.ts
dabbeg: On second thought, the type in `getFamily` is derived from the nationalRegistryApi as I never recreate the object.
And for the mapping functions, the `getExamResults` has a return type and there is no real gain from putting return types on the mapping functions. I just adds verbosity.

darrikonn: aiit :+1:

dabbeg (approved):
Nice!

Thorrik58 (approved):
Everything related to the national registry LGTM 🎉

PR #3308, opened 2021-03-21
feat(web): Add SidebarCard and fix broken links

Ingimarsson:
# Add SidebarCard and fix broken links

## What

Sidebar cards are now fetched from Contentful. This PR also fixes some broken links on organization pages.

## Screenshots / Gifs

image

## Checklist:


apps/web/hooks/useLinkResolver/useLinkResolver.tsx
```diff
@@ -65,6 +65,10 @@ export const routesTemplate = {
     is: '/stofnanir/[slug]/[subSlug]',
     en: '/en/organizations/[slug]/[subSlug]',
   },
+  organizationservices: {
+    is: '/stofnanir/[slug]/thjonusta',
+    en: '/en/organizations/[slug]/services',
+  },
```
stefanhar, line 71: The order of url template string should be from most specific to least specific for correct type resolution (used e.g. in link resolver).
The order of these seems wrong to me, I think it should be:
`organizationservices`
`organizationsubpage`
`organizationpage`
`organizations`

libs/api/domains/cms/src/lib/models/organizationPage.model.ts
```diff
@@ -35,6 +36,9 @@ export class OrganizationPage {
 
   @Field(() => [FooterItem], { nullable: true })
   footerItems?: Array<FooterItem>
+
+  @Field(() => [SidebarCard], { nullable: true })
+  sidebarCards?: Array<SidebarCard>
```
stefanhar, line 41: Seems like these always return array, maybe the nullable and optional are not needed?

libs/api/domains/cms/src/lib/models/sidebarCard.model.ts
```diff
@@ -0,0 +1,29 @@
...
+  type!: string
+
+  @Field()
+  title!: string
+
+  @Field()
+  content!: string
+
+  @Field(() => Link, { nullable: true })
+  link?: Link | null
```
stefanhar, line 20: The optional here might not be needed since mapper ensures this is always set

apps/web/components/Organization/SidebarCard/SidebarCard.tsx
```diff
@@ -0,0 +1,57 @@
...
+        <GridColumn span={['3/12', '3/12', '12/12']}>
+          <Box
+            display="flex"
+            justifyContent={'center'}
+            alignItems={'center'}
+            paddingBottom={3}
+          >
+            <img
+              src={
+                'https://images.ctfassets.net/8k0h54kbe6bj/2c9RXjtApgqkUWeKh0s1xP/820afa7b351e638473bff013277929ea/Vector.svg'
```
stefanhar, line 30: This approach might be fragile since there is no indication this asset is in use inside Contentful.
Did you consider using the Icon component from island-ui?
It seems to already have a similar warning icon with a red variant:
https://ui.devland.is/?path=/docs/core-icon--all-icons

stefanhar (changes requested):

stefanhar (approved):
LGTM

ingolfurs (approved):

PR #3315, opened 2021-03-22
feat: dpo add admin guard

fjandakornid:
# Add admin guard to DocumentProvider api

## What

Add guard to DocumentProvider api that checks if current user is included in list of national ids of admins.

## Why

Protect the methods that are used by the admin client. This is only the first half of the authentication restiction. The methods that are used by the application needs to be restricted also, but are not included in this PR.

## Checklist:


ssh16 (approved):

libs/api/domains/document-provider/src/lib/document-provider.module.ts
```diff
@@ -9,9 +9,11 @@ import {
   DOCUMENT_PROVIDER_CLIENT_CONFIG_TEST,
 } from './client/documentProviderClientConfig'
 import { DocumentProviderClientTest } from './client/documentProviderClientTest'
+//import { AdminGuard, DOCUMENT_PROVIDER_ADMINS } from './admin.guard'
```
dabbeg, line 12: were you going to use DOCUMENT_PROVIDER_ADMINS below?

fjandakornid: fixed

dabbeg (approved):

PR #3323, opened 2021-03-23
fix: secret service macos bug

busla:


scripts/create-secret.sh
```diff
@@ -1,100 +1,87 @@
 #!/bin/bash
-set -euo pipefail
+set -e
 
-#---------------------------------------------------------------------------------------------------------------#
-#--- Script for creating secrets (SecureString) in the AWS parameter store using the path /k8s/[secret-name]----#
-#---------------------------------------------------------------------------------------------------------------#
+export AWS_PAGER=""
```
sindrig, line 4: unused?

scripts/create-secret.sh
```diff
@@ -104,69 +91,74 @@ function validate_length() {
...
+
+  if [ -n "$TAGS" ]
+  then
+    CMD="$CMD --tags $TAGS"
+  fi
+  eval "$CMD"
+  printf "%sDone!%s" "$GREEN" "$RESET"
 }
 
+case $1 in
```
sindrig, line 156: I don't get this case. Is this for testing purposes? Maybe add a comment about it then?

scripts/create-secret.sh
busla: yes, disable the output from going to the pager

pshomov (approved):
LGTM I think

scripts/create-secret.sh
busla: used in [shellspec](https://shellspec.info/) but threw shellspec out. Forgot to clean up this line.

scripts/create-secret.sh
busla: so, no... its used 😄

sindrig (approved):

scripts/create-secret.sh
sindrig: yep got it :)

PR #3330, opened 2021-03-23
feat(judicial-system): Case File Upload

gudjong:
# Case File Upload

https://app.asana.com/0/1199153462262248/1199986469006831

## What

☝️ 

## Why

## Screenshots / Gifs

image

## Checklist:


apps/judicial-system/api/src/app/modules/auth/auth.service.ts
```diff
@@ -23,7 +23,7 @@ export class AuthService {
     return await res.json()
   }
 
-  validateUser(user?: User): boolean {
-    return Boolean(user?.active)
+  validateUser(user: User): boolean {
+    return Boolean(user.active)
```
addi, line 27: do we still need the boolean here?

gudjong: It is not necessary for correctness, but using it is one way to deal with a "strict typescript" warning we get because user can be undefined.

barabrian (approved):
LGTM

oddsson (approved):

PR #3335, opened 2021-03-23
feat(crc): national registry data contract and mock providers

VilliGunn:
# \<Description\>

https://app.asana.com/0/1199733748699731/1200065515355085

## What

- Updated the initial data contract from NationalRegistry
  - Removed parentDataProvider and childrenDataProvider
- Updated the flow according to the new data contract
- Updated mock form and flow with the new contract in mind

## Why

- We needed to update the contract according to new specifications in the flow
- This contract is simpler than the previous one

## Checklist:


apps/application-system/api/src/app/modules/application/files/file.service.spec.ts
```diff
@@ -25,75 +25,70 @@ describe('FileService', () => {
...
-    id: 'id',
-    name: 'child',
-    ssn: '123456-7890',
-    postalCode: '101',
-    address: 'Borgartún',
-    city: 'Reykjavík',
-    phoneNumber: '222-3333',
-    email: 'email3@email3.is',
+    fullName: 'child',
+    nationalId: '123456-7890',
```
Rikkiabb, line 47: Should we remove the dash in the nationalId here, since we don't have it in other fake nationalId's in these tests?

apps/application-system/api/src/app/modules/application/files/file.service.ts
```diff
@@ -71,14 +73,18 @@ export class FileService {
...
+    const { nationalRegistry } = externalData
     const isParentA = state === 'draft'
-
-    const parentBName = parentNationalRegistry.data.name
+    const applicant = nationalRegistry?.data
+    const selectedChildren = getSelectedChildrenFromExternalData(
+      applicant.children,
+      answers.selectChild,
+    )
+    const parentB = selectedChildren[0].otherParent
```
Rikkiabb, line 83: Should we guard here against an empty array? 
Know it should not be possible with the current flow, but a thought.

apps/application-system/api/src/app/modules/application/files/file.service.spec.ts
VilliGunn: Sure, I'm pretty sure this is copied from older code though, thought there was some reason 😃 
```suggestion
    nationalId: '1234567890',
```

apps/application-system/api/src/app/modules/application/files/file.service.ts
```diff
@@ -71,14 +73,18 @@ export class FileService {
   ): Promise<SigningServiceResponse> {
     this.validateApplicationType(application.typeId)
     const { answers, externalData, id, state } = application as CRCApplication
-    const { nationalRegistry, parentNationalRegistry } = externalData
+    const { nationalRegistry } = externalData
     const isParentA = state === 'draft'
```
Rikkiabb, line 77: Not in this PR but I always think when I see this that we should change this check.

I am thinking if we should send in the authenticated user national id to the function and check if that is the same as the `application.applicant` national id. Any thoughts about this?

apps/application-system/api/src/app/modules/application/files/file.service.ts
```diff
@@ -71,14 +73,18 @@ export class FileService {
...
+    )
+    const parentB = selectedChildren[0].otherParent
 
     switch (pdfType) {
       case PdfTypes.CHILDREN_RESIDENCE_CHANGE: {
-        const name = isParentA ? nationalRegistry.data.fullName : parentBName
+        const name = isParentA ? applicant.fullName : parentB.fullName
         const phoneNumber = isParentA
           ? answers.parentA.phoneNumber
           : answers.parentB.phoneNumber
```
Rikkiabb, line 90: Should we join this ternary and return an object with parentinfo instead of two strings?

apps/application-system/api/src/app/modules/application/files/file.service.ts
```diff
@@ -105,17 +111,22 @@ export class FileService {
...
 
+    const parentB = selectedChildren[0].otherParent
+
     const pdfBuffer = await generateResidenceChangePdf(
       selectedChildren,
-      (application.externalData.nationalRegistry.data as unknown) as User,
-      application.externalData.parentNationalRegistry.data,
-      application.answers.selectDuration,
-      application.answers.residenceChangeReason,
+      applicant,
```
Rikkiabb, line 126: Nice 🤩 Love that we don't have to use two types for the user any more

apps/application-system/api/src/app/modules/application/files/utils/pdf.ts
```diff
@@ -101,7 +103,7 @@ export async function generateResidenceChangePdf(
     PdfConstants.NORMAL_FONT,
     PdfConstants.VALUE_FONT_SIZE,
     PdfConstants.LARGE_LINE_GAP,
-    `Heimilisfang: ${parentA.address?.streetAddress}, ${parentA.address?.postalCode} ${parentA.address?.city}`,
+    `Heimilisfang: ${parentA.address.streetName}, ${parentA.address.postalCode} ${parentA.address.city}`,
```
Rikkiabb, line 106: If we want then we can use the utils function here.

libs/application/template-api-modules/src/lib/modules/templates/children-residence-change/children-residence-change.service.ts
```diff
@@ -37,59 +37,62 @@ export class ChildrenResidenceChangeService {
 
   async submitApplication({ application }: props) {
     const { answers, externalData } = application
-    const { parentNationalRegistry } = externalData
-    const nationalRegistry = (externalData.nationalRegistry
-      .data as unknown) as User
+    const { nationalRegistry } = externalData
+    const applicant = nationalRegistry?.data
```
Rikkiabb, line 41: Do we need the `?` here?

apps/application-system/api/src/app/modules/application/files/file.service.ts
VilliGunn: Yeah, that is definitely nicer than this check 💯

libs/application/template-api-modules/src/lib/modules/templates/children-residence-change/children-residence-change.service.ts
```diff
@@ -37,59 +37,62 @@ export class ChildrenResidenceChangeService {
...
     const fileContent = file.Body as Buffer
 
-    const selectedChildren = application.externalData.childrenNationalRegistry.data.filter(
-      (c) => application.answers.selectChild.includes(c.name),
+    const selectedChildren = getSelectedChildrenFromExternalData(
+      applicant.children,
+      answers.selectChild,
     )
 
+    const otherParent = selectedChildren?.[0]?.otherParent
```
Rikkiabb, line 53: Should we check array length and do we need to null check here?

libs/application/template-api-modules/src/lib/modules/templates/children-residence-change/children-residence-change.service.ts
```diff
@@ -37,59 +37,62 @@ export class ChildrenResidenceChangeService {
...
     const participants: Array<Person> = selectedChildren.map((child) => {
       return {
-        name: child.name,
-        ssn: child.ssn,
-        homeAddress: child.address,
-        postalCode: child.postalCode,
-        city: child.city,
+        name: child.fullName,
+        ssn: child.nationalId,
+        // Not sure if this should be the current address or the future address
```
Rikkiabb, line 92: As I understand it then this should be the current address.

libs/application/templates/children-residence-change/src/dataProviders/MockNationalRegistryProvider.ts
```diff
@@ -0,0 +1,84 @@
...
+            code
+            postalCode
+            city
+            streetAddress
+            lastUpdated
+          }
+        }
+      }
+    `
+    const crcApplication = (application as unknown) as CRCApplication
```
Rikkiabb, line 30: Do want to use typed props and the `Override` function from `libs/application/templates/children-residence-change/src/types/index.ts` instead of throwing to unknown?

libs/application/templates/children-residence-change/src/dataProviders/MockNationalRegistryProvider.ts
```diff
@@ -0,0 +1,84 @@
...
+    const {
+      answers: {
+        mockData: { parents, children },
+      },
+    } = crcApplication
+    if (!children) {
+      throw new Error('Engin börn fundust í þinni forsjá')
+    }
+    const childrenArray: NationalRegistry['children'] = []
+    children?.map((child) => {
```
Rikkiabb, line 40: Do we need the null check here?

libs/application/templates/children-residence-change/src/dataProviders/MockNationalRegistryProvider.ts
```diff
@@ -0,0 +1,84 @@
...
+      childrenArray.push({
+        ...child,
+        livesWithApplicant: child?.livesWithApplicant?.includes('yes') || false,
+        otherParent: parents[child.otherParent],
+      })
+    })
```
Rikkiabb, line 46: Instead of setting the `childrenArray` to an empty array and pushing to that array we could do this: 
```    const childrenArray: NationalRegistry['children'] = children.map(
      (child) => {
        return {
          ...child,
          livesWithApplicant:
            child.livesWithApplicant.includes('yes') || false,
          otherParent: parents[child.otherParent],
        }
      },
    )

File: libs/application/templates/children-residence-change/src/dataProviders/MockNationalRegistryProvider.ts [hunk 13]
```diff
@@ -0,0 +1,84 @@
...
+      query NationalRegistryUserQuery {
+        nationalRegistryUser {
+          nationalId
+          fullName
+          address {
+            code
+            postalCode
+            city
+            streetAddress
+            lastUpdated
```
Rikkiabb, line 25: Even though it is not ready don't we want the query to reflect what data we expect?

libs/application/templates/children-residence-change/src/dataProviders/MockNationalRegistryProvider.ts
```diff
@@ -0,0 +1,84 @@
...
+        ...child,
+        livesWithApplicant: child?.livesWithApplicant?.includes('yes') || false,
+        otherParent: parents[child.otherParent],
+      })
+    })
+    return this.useGraphqlGateway(query)
+      .then(async (res: Response) => {
+        const response = await res.json()
+        if (response.errors) {
+          throw new Error('Error from NationalRegistry')
```
Rikkiabb, line 51: Don't we always end up here since the endpoint is not ready? Or am I missing something? 😅

libs/application/templates/children-residence-change/src/dataProviders/index.ts
```diff
@@ -1,8 +1,3 @@
...
-  NationalRegistryProvider,
-  UserProfileProvider,
-} from '@island.is/application/data-providers'
-export { ChildrenNationalRegistryProvider } from './ChildrenNationalRegistryProvider'
-export { MockChildrenNationalRegistryProvider } from './MockChildrenNationalRegistryProvider'
-export { ParentNationalRegistryProvider } from './ParentNationalRegistryProvider'
-export { MockParentNationalRegistryProvider } from './MockParentNationalRegistryProvider'
+export { UserProfileProvider } from '@island.is/application/data-providers'
+export { MockNationalRegistryProvider } from './MockNationalRegistryProvider'
+export { NationalRegistryProvider } from './NationalRegistryProvider'
```
Rikkiabb, line 3: 👏

libs/application/templates/children-residence-change/src/fields/ChangeInformation/index.tsx
```diff
@@ -2,7 +2,7 @@ import React from 'react'
 import { useIntl } from 'react-intl'
 import { Box, Text } from '@island.is/island-ui/core'
 import { CheckboxController } from '@island.is/shared/form-fields'
-import { constructParentAddressString } from '../../lib/utils'
+import { constructAddressString } from '../../lib/utils'
```
Rikkiabb, line 5: Wondering if we should rename this function. What do you think? 
I am thinking about something like `formatHomeAddress`

libs/application/templates/children-residence-change/src/fields/MockData/parent.tsx
```diff
@@ -0,0 +1,121 @@
...
+const MockParents = ({ application, field }: CRCFieldBaseProps) => {
+  const { id } = field
+  const { answers } = application
+  const parents = answers.mockData?.parents
+
+  const { fields, append, remove } = useFieldArray({ name: id })
+
+  useEffect(() => {
+    // The repeater should include one line by default
+    if (fields.length === 0) handleAddPerson()
```
Rikkiabb, line 23: Maybe just me....... but...... I prefer: 
```
if (fields.length === 0)  {
    handleAddPerson()
}
```
🙈

libs/application/templates/children-residence-change/src/fields/Overview/index.tsx
```diff
@@ -32,9 +35,12 @@ const Overview = ({
...
   )
   const applicant = externalData.nationalRegistry.data
-  const parent = externalData.parentNationalRegistry.data
-  const parentAddress = constructParentAddressString(parent)
-  const children = answers.selectChild
+  const children = getSelectedChildrenFromExternalData(
+    applicant.children,
+    answers.selectChild,
+  )
+  const parentB = children?.[0].otherParent
```
Rikkiabb, line 42: Do we need the null check here?

libs/application/templates/children-residence-change/src/fields/Overview/index.tsx
```diff
@@ -203,15 +211,15 @@ const Overview = ({
           })}
         </Text>
         <Text>{applicant?.fullName}</Text>
-        <Text>{applicant?.legalResidence}</Text>
+        <Text>{constructAddressString(applicant?.address)}</Text>
```
Rikkiabb, line 214: Do we need the null check here?

libs/application/templates/children-residence-change/src/lib/utils.ts
```diff
@@ -1,8 +1,15 @@
...
+import { Address, Child } from '../types'
 
-export const constructParentAddressString = (parent: PersonResidenceChange) => {
-  if (!parent) {
+export const constructAddressString = (address: Address) => {
+  if (!address) {
     return null
   }
-  return `${parent?.address}, ${parent?.postalCode} ${parent?.city}`
+  return `${address?.streetName}, ${address?.postalCode} ${address?.city}`
```
Rikkiabb, line 7: We have already null checked on line 4

libs/application/templates/children-residence-change/src/lib/utils.ts
```diff
@@ -1,8 +1,15 @@
...
+}
+
+export const getSelectedChildrenFromExternalData = (
+  children: Child[],
+  selectedChildren: string[],
+): Child[] => {
+  return children.filter((child) => selectedChildren.includes(child.nationalId))
```
Rikkiabb, line 14: Nice

Rikkiabb (changes requested):
Awesome work, added some comments 😄

libs/application/template-api-modules/src/lib/modules/templates/children-residence-change/children-residence-change.service.ts
VilliGunn: I don't think we need a check here, neither for length or the null check.

libs/application/templates/children-residence-change/src/dataProviders/MockNationalRegistryProvider.ts
VilliGunn: No, we are using the old endpoint from national registry to populate the applicant info.

libs/application/templates/children-residence-change/src/fields/ChangeInformation/index.tsx
VilliGunn: Yeah, I just removed the `parent` bit because that didn't make sense but I agree with `formatHomeAddress` or even just `formatAddress` since it can be whatever address really.

libs/application/templates/children-residence-change/src/fields/MockData/parent.tsx
VilliGunn: True, C/P 😅

libs/application/templates/children-residence-change/src/dataProviders/MockNationalRegistryProvider.ts
VilliGunn: I can't figure out how to make this work 🤷

apps/application-system/api/src/app/modules/application/files/file.service.ts
VilliGunn: Well I don't think we need to guard aginst this. You shouldn't be able to get to this step without this being in place. You will be stopped earlier in the flow. I have also removed other unneccessary guards.

Rikkiabb (approved):
💯

PR #3359, opened 2021-03-25
feat(auth-api): Add company registry client.

valurefugl:
Client for RSK company registry REST service.

## What

Create a reusable client library generated from the current openapi yaml file from RSK.

Note: The openapi file was hand edited so that the endpoint we need (/{Kennitala}/companies) would work.
We probably need to get an updated file and regenerate the client.
Also, there currently is an error in the generated code if no companies are found.

## Why

Needed to fetch companies that an individual holds a procuration for (for delegations in IDS).

## Checklist:


darrikonn (commented):
Looks good, but mainly wondering: What is the reason for splitting this up from the already defined `rsk` client in `libs/clients`?

valurefugl:
Can you please elaborate?

darrikonn:
> Can you please elaborate?

This is implementing a rsk client.
Another such client has been implemented at `libs/clients/rsk` in our monorepo.
So I was wondering why the logic in this PR wasn't just added to the existing `rsk` client instead of generating a new library called company-registry?

valurefugl:
I see. Why is that preferrable?

darrikonn:
> I see. Why is that preferrable?

Since it's implementing the same client.
Isn't it just some extra endpoints for the same client?

That was actually my question to you, why isn't this PR just adding to the existing client, was there a reason or just you didn't know of the existing client?

darrikonn:
The way I look at it is:
Future projects that might need to interact with the company registry part of the RSK client will most likely go through this process:
1. We need to interact with `rsk`
2. Has the `rsk` client been implemented?
3. Yes, but it's missing the company registry part, so we'll add it to the existing client

Now we have two clients implementing the same logic.

valurefugl:
My understanding was that the existing client was interacting with a soap service, while this client is using a rest service over xroad. Therefore not obvious, to me at least, that they should be combined.

darrikonn:
> My understanding was that the existing client was interacting with a soap service, while this client is using a rest service over xroad. Therefore not obvious, to me at least, that they should be combined.

Yeah makes sense to split that in some way. I was just fishing for a reason for the split. Didn't know they differed in SOAP vs. REST.
Then it's an architectural decision where the split should happen, i.e. in different libraries or if they should live in the same library, makes sense where there are multiple versions of the same client, so it can be versioned?
Good to make a decision for future clients. We should probably been taken up at dev_faghopur on Slack.

libs/clients/company-registry/src/index.ts
```diff
@@ -0,0 +1,5 @@
+export {
+  CompanyRegistryModule,
+  CompanyRegistryModuleConfig,
+} from './lib/company-registry.module'
+export * from '../gen/fetch'
```
eirikurn, line 5: We should re-export some of the generated exports with better names. E.g.

```
export * from './lib/company-registry.api'
```

Where `company-registry.api.ts` looks something like this:

```
export {
  DefaultApi as CompanyRegistryApi,
  // Other exports/renames?
} from '../../gen/fetch'
```

We can also import CompanyRegistryApi from this module in `company-registry.module.ts`.

eirikurn (commented):
This module should ideally replace `libs/clients/rsk`, which currently serves the same purpose (or subset) but talks to the old REST API instead of XRoad, and is not using any OpenAPI definitions.

This client is also more of an RSK client then a CompanyRegistry client.

Ideally we can do a clean replace, but if that's unsafe at the moment we might want to rename the old client lib to `libs/clients/rsk-legacy` or something.

darrikonn:
> This module should ideally replace `libs/clients/rsk`, which currently serves the same purpose (or subset) but talks to the old REST API instead of XRoad, and is not using any OpenAPI definitions.
> 
> This client is also more of an RSK client then a CompanyRegistry client since the API includes a bunch of RSK (financial) endpoints.
> 
> Ideally we can do a clean replace, but if that's unsafe at the moment we might want to rename the old client lib to `libs/clients/rsk-legacy` or something.

Agree here. I actually believe the clients in `libs/clients` should be versioned. What if there is yet another client for `rsk`, `libs/clients/rsk-legacy2` 🙈 . 
There are e.g. 3+ clients being used in our monorepo that is implementing a some specific version of the nationalRegistry client. For future pains, I think this will solve that:
```
- libs
  |- clients
     |- rsk
        |- v1
        |- v2
```
Then these libraries will be named `rsk-v1`/`rsk-v2` in `workspace.json`.

valurefugl:
The old and new rsk clients are now under /rsk in folders /v1 and /v2.

libs/clients/company-registry/src/index.ts
valurefugl: Fixed.

eirikurn (approved):

libs/clients/rsk/v2/src/companyregistryapi_openapi3.yml
```diff
@@ -0,0 +1,345 @@
...
+          type: string
+          description: Hlutverkið sem innsend kennitala gegnir innan félagsins.
+            T.d. "Stjórnarformaður".
+        Rekstarform:
+          type: string
+          description: Rekstrarform félagsins sem innsend kennitala tilheyrir. T.d.
+            "E1".
+        ErProkuruhafi:
+          type: string
+          description: Rofi sem segir til um hvort hlutverk innsendrar kennitölu prókúruhafi
```
kksteini, line 340:
```suggestion
          description: Rofi sem segir til um hvort hlutverk innsendrar kennitölu sé prókúruhafi
```

kksteini (approved):

PR #3362, opened 2021-03-26
feat(crc): both parents can apply

VilliGunn:
# \<Description\>

https://app.asana.com/0/1199733748699731/1200047583274103

## What

- Added a functionality so both parents can start an application.
- Added a helper function to get the current and future children residence according to the parents info
- Changed the answer name `selectChild` to `selectedChildren`

## Why

- Both parents can start an application regardless of where the children currently live
- The applicant will always be `Parent A` and the second signer `Parent B`

### Note

- This PR does not handle children living with seperate parents - this will be handled in a follow up PR
  - One child lives with Parent A and the other with Parent B
  - https://app.asana.com/0/1199733748699731/1199584418237686

## Checklist:


libs/application/templates/children-residence-change/src/lib/utils.ts
```diff
@@ -13,3 +19,50 @@ export const getSelectedChildrenFromExternalData = (
 ): Child[] => {
   return children.filter((child) => selectedChildren.includes(child.nationalId))
 }
+
+interface ChildrenResidenceInfo {
+  parent: {
+    letter: 'A' | 'B'
```
Rikkiabb, line 25: Do we want to change to an enum?

Rikkiabb (approved):

PR #3378, opened 2021-03-29
feat(crc): children selection condition

VilliGunn:
# \<Description\>

https://app.asana.com/0/1199733748699731/1199584418237686

## What

- Added a `disabled` props to the `CheckboxController` options to be able to disable single checkbox instead of all checkboxes rendered by the controller.
- Added an `onSelect` prop to the `CheckboxController` to be able to get the selection state when using the component.
- Render the `CheckBox` label as `<span>` instead of `<p>`
  - the disabled checkbox label style was overridden by the `<p>` styles.
  - According to mdn only phrasing content is allowed as a child element of `<label>`.
  - Bug introduced here: #3213 
- Conditionally disable options according to the first option selected.

## Why

It is not possible to apply for a residence change for two or more children that are currently not registered at the same residence or do not have the same two parents.

## Screenshots / Gifs

Screenshot 2021-03-29 at 10 02 18

Disabled states:
Before fix
Screenshot 2021-03-29 at 10 25 10
After fix
Screenshot 2021-03-29 at 10 23 20

## Checklist:


libs/application/templates/children-residence-change/src/fields/SelectChildren/index.tsx
```diff
@@ -25,16 +52,22 @@ const SelectChildren = ({ field, application, error }: CRCFieldBaseProps) => {
         id={id}
         disabled={disabled}
         name={`${id}`}
-        defaultValue={[]}
+        defaultValue={
+          selectedChildrenState !== undefined
+            ? selectedChildrenState
+            : undefined
```
Rikkiabb, line 58: Is this ternarny needed? Can't we always send in selectedChildrenState?

libs/application/templates/children-residence-change/src/fields/SelectChildren/index.tsx
```diff
@@ -1,18 +1,45 @@
...
   const { formatMessage } = useIntl()
   const {
     externalData: { nationalRegistry },
+    answers,
   } = application
   const children = nationalRegistry.data.children
+  const currentAnswer = answers.selectedChildren
+  const [selectedChildrenState, setSelectedChildrenState] = useState<string[]>(
+    currentAnswer,
+  )
```
Rikkiabb, line 41: Since `currentAnswer` can be undefined isn't the type here `string[] | undefined` ?

Rikkiabb (commented):
Two questions from me

Rikkiabb (approved):
🚢

oddsson (approved):

DavidBachmann (approved):
LGTM

PR #3382, opened 2021-03-29
fix(crc): Date format in contract

Rikkiabb:
# \<Description\>
Fixes formatting for date in crc.

## Why
It is hardcoded to an American format in the application system which we don't want.

## Screenshots / Gifs

## Checklist:


libs/application/templates/children-residence-change/src/lib/dataSchema.ts
```diff
@@ -28,13 +28,11 @@ export const dataSchema = z.object({
...
-  //   .optional()
-  //   .refine((v) => v, {
-  //     message: 'Velja þarf valmöguleika',
-  //   }),
-  selectDuration: z.array(z.string()),
+  selectDuration: z
+    .array(z.string())
+    .refine((v) =>
+      v.length && v[0] === 'permanent' ? v.length === 1 : v.length === 2,
+    ),
```
VilliGunn, line 35: Just a question. Is there no way to validate the first item of the array as the enum `'temporary' | 'permanent'`?
Not a showstopper but I'm just wondering if it's possible

Rikkiabb: I talked about this in the meeting we had about validation in the application system and if I remember correctly then it is possible in zod but it is not possible in the application system because of the validation middleware. But I can test it out in the validation ticket we have on our backlog.

VilliGunn: Let's keep it as is for now.

VilliGunn (approved):
:shipit:

PR #3391, opened 2021-03-29
feat(web): Organization themes

Ingimarsson:
# Organization themes

## What

This PR includes three different themes for organization pages.

## Why

Organizations will have their own branding.

## Screenshots / Gifs

image

image

image

## Checklist:


Ingimarsson:
@barabrian do you have a better idea of how we could change the header at runtime?

I added this to `layouts/main.tsx` so that the screens could change the color scheme through their `getInitialProps` but maybe it's not ideal. See [d515195](https://github.com/island-is/island.is/pull/3391/commits/d5151959fe4728954426a97d82a3111038356a54) for context :grin: 

```
const themeConfig: Partial<LayoutProps> =
  'darkTheme' in componentProps
    ? { headerColorScheme: 'white', headerButtonColorScheme: 'negative' }
    : {}

return {
  layoutProps: { ...layoutProps, ...layoutConfig, ...themeConfig },
  componentProps,
}
```

libs/api/domains/cms/src/lib/models/sidebarCard.model.ts
```diff
@@ -0,0 +1,29 @@
...
+  type!: string
+
+  @Field()
+  title!: string
+
+  @Field()
+  content!: string
+
+  @Field(() => Link, { nullable: true })
+  link!: Link | null
```
stefanhar, line 20: Not sure what the use case is here but if this link field includes internal links we should consider using ReferenceLink here to allow for dynamic link resolution.
ReferenceLink also supports external link resolution

libs/api/domains/cms/src/lib/models/organizationPage.model.ts
```diff
@@ -45,11 +55,16 @@ export const mapOrganizationPage = ({
   title: fields.title ?? '',
   slug: fields.slug ?? '',
   description: fields.description ?? '',
+  theme: fields.theme ?? 'default',
+  themeProperties: fields.themeProperties
+    ? JSON.stringify(fields.themeProperties)
+    : '',
```
stefanhar, line 61: We should avoid using JSON if able since it breaks type safety between the api and web.
It seems like `themeProperties` contains a set of known properties:
```
gradientStartColor: string
gradientEndColor: string
```
Can we move this to its own model with a mapper?

Ingimarsson: Good idea. See [a35c446](https://github.com/island-is/island.is/pull/3391/commits/a35c446825a4436f6397d3e860c2d248622618c3)

libs/api/domains/cms/src/lib/models/sidebarCard.model.ts
Ingimarsson: Yes that's in the backlog. We ran into some problems with `ReferenceLink` with a slug and subslug. One solution was to use your UI extension but it didn't always seem to update the slug correctly.

Maybe this deserves its own PR? :smile:

libs/api/domains/cms/src/lib/models/organizationPage.model.ts
```diff
@@ -19,22 +20,31 @@ export class OrganizationPage {
...
+  @Field()
+  themeProperties!: string
 
   @Field(() => [SliceUnion])
-  slices?: Array<typeof SliceUnion | null>
+  slices!: Array<typeof SliceUnion | null>
 
   @Field(() => [LinkGroup])
-  menuLinks?: Array<LinkGroup>
+  menuLinks!: Array<LinkGroup>
```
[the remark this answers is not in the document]
Ingimarsson: Same as above, `ReferenceLink` related.

libs/api/domains/cms/src/lib/models/sidebarCard.model.ts
stefanhar: Yes, also seems not to belong to this PR anymore 😄

stefanhar (commented):
Backend LGTM

stjanilofts (approved):

ingolfurs (approved):

PR #3399, opened 2021-03-30
feat(crc): transfer duration validation

VilliGunn:
# \<Description\>

https://app.asana.com/0/1199733748699731/1199910145131639

## What

- Added `minDate` and `maxDate` to the `DatePickerController`
- Split `selectDuration` up into two fields `durationType` and `durationDate`
  - Updated all references to `selectDuration`
- `durationDate` has a `minDate` 6 months ahead of the current date
- Added validation and translation strings 

## Why

Temporary transfers should be at least 6 months.
This screen was missing validation.

## Screenshots / Gifs

Screenshot 2021-03-30 at 11 41 54
Screenshot 2021-03-30 at 11 42 06

## Checklist:


libs/application/template-api-modules/src/lib/modules/templates/children-residence-change/children-residence-change.service.ts
```diff
@@ -104,13 +104,14 @@ export class ChildrenResidenceChangeService {
 
     participants.push(parentA, parentB)
 
+    const durationType = answers.durationType as string
```
VilliGunn, line 107: I don't like this but it's needed because `durationType` is set as optional in the dataschema..

libs/application/templates/children-residence-change/src/lib/dataSchema.ts
```diff
@@ -28,11 +28,14 @@ export const dataSchema = z.object({
...
     .length(1, error.validation.approveChildrenResidenceChange.defaultMessage),
-  selectDuration: z
-    .array(z.string())
-    .refine((v) =>
-      v.length && v[0] === 'permanent' ? v.length === 1 : v.length === 2,
-    ),
+  durationType: z
+    .enum(['permanent', 'temporary'])
+    .optional()
+    .refine((v) => v, error.validation.durationType.defaultMessage),
```
VilliGunn, line 34: Set as optional and with refinement to control the error message string

libs/application/templates/children-residence-change/src/fields/Duration/index.tsx
```diff
@@ -10,26 +11,31 @@ import {
...
+export const selectDurationInputs = [typeInput, dateInput]
+
 export type ValidAnswers = 'permanent' | 'temporary' | undefined
-const Duration = ({ field, application, error }: CRCFieldBaseProps) => {
-  // TODO: Fix when we fix the schema validation for this field
-  const currentAnswer = application.answers.selectDuration
-    ? (application.answers.selectDuration[0] as ValidAnswers)
-    : undefined
+const Duration = ({ application, errors }: CRCFieldBaseProps) => {
+  const currentAnswer = application.answers?.durationType || undefined
```
Rikkiabb, line 21: Isn't this equal to:
`  const currentAnswer = application.answers.durationType` ?

libs/application/template-api-modules/src/lib/modules/templates/children-residence-change/children-residence-change.service.ts
Rikkiabb: Agree, but don't see a better solution currently 😞 
if we would mix this logic in on line 112 then the code would become pretty messy I think.

libs/application/template-api-modules/src/lib/modules/templates/children-residence-change/children-residence-change.service.ts
```diff
@@ -104,13 +104,14 @@ export class ChildrenResidenceChangeService {
...
     const extraData = {
       interviewRequested: answers.interview,
       reasonForChildrenResidenceChange: answers.residenceChangeReason ?? '',
       transferExpirationDate:
-        answers.selectDuration[0] === 'permanent'
-          ? answers.selectDuration[0]
-          : formatDate(answers.selectDuration[1]),
+        durationType === 'temporary' && answers.durationDate
```
Rikkiabb, line 112: Since we are doing this check in a couple of places currently then we could add a helper function for it. Something like `IsDurationTemporary' which would return a boolean. Just a thought not a must.

Rikkiabb (approved):
Just nits from me, no stoppers, approved ✅

libs/application/template-api-modules/src/lib/modules/templates/children-residence-change/children-residence-change.service.ts
VilliGunn: Keeping as is for now.

baering (approved):
LGTM 👍 

I don't like having code to generate a specific PDF for a specific application inside the application-system API source code and it would be nice to refactor that into the template api module at some point.

PR #3404, opened 2021-03-30
build(all): better caching for assets in static-rendered apps

### LINKED ISSUE #3314: Configure caching headers for static apps
Our static apps are hosted in an nginx container which has no cache control set. The [default behaviour](https://webmasters.stackexchange.com/questions/111298/what-happens-if-you-dont-set-cache-control-header) will cause issues.

For index.html requests, we should set something like `cache-control: public, max-age=0`

For most other requests, we should set something like `cache-control: public, max-age=31536000`.

My only concern are favicon and some static files in the /assets folder which are not versioned currently.

Ideally, if we can detect requests that have `/\.[0-9a-f]{20}\./` hashes in their filename. we can safely configure far future caching with `immutable` (`cache-control: public, max-age=31536000, immutable`), which skips revalidation requests even on page refresh.

pshomov:
# \<Description\>

Fixes #3314

## What

Cache in the browser more assets than fewer

## Why

We like things snappy

## Screenshots / Gifs

## Checklist:


eirikurn (commented):
This looks like a great fix for static file caching, but we also need to disable caching for all index.html paths, somehow.

pshomov:
Yes, was just interested in the webpack assets section only at the time. Now added the explicit cache control for the routes as well.

eirikurn (approved):
Sweet!!

sindrig (approved):
LGTM

PR #3409, opened 2021-03-30
feat(crc): national registry date and timestamp in pdf

VilliGunn:
# \<Description\>

https://app.asana.com/0/1199733748699731/1199968000253747

## What

Added a date and timestamp of when the national registry data is fetched

## Why

It was missing from the pdf contract

## Checklist:


apps/application-system/api/src/app/modules/application/files/utils/pdf.ts
```diff
@@ -11,13 +14,28 @@ import {
 import { PdfConstants } from './constants'
 import { DistrictCommissionerLogo } from './districtCommissionerLogo'
 
+const formatDays = (date: string): string => {
```
VilliGunn, line 17: At first I wasn't going to do this but decided to humor the content creators. Instead of displaying 
"Fyrirspurn og uppfletting í gögnum Þjóðskrár fór fram mánudagur 25. janúar 2021 kl. 13:36."
Then we have 
"Fyrirspurn og uppfletting í gögnum Þjóðskrár fór fram mánudaginn 25. janúar 2021 kl. 13:36."

apps/application-system/api/src/app/modules/application/files/file.service.ts
```diff
@@ -117,9 +117,11 @@ export class FileService {
   private async createChildrenResidencePdf(application: CRCApplication) {
     const bucket = this.getBucketName()
     const { answers, externalData } = application
-    const applicant = externalData.nationalRegistry.data
 
-    const pdfBuffer = await generateResidenceChangePdf(applicant, answers)
+    const pdfBuffer = await generateResidenceChangePdf(
+      externalData.nationalRegistry,
+      answers,
```
Rikkiabb, line 123: We should maybe change this interface and send in the whole application?

libs/application/templates/children-residence-change/src/types/index.ts
```diff
@@ -36,9 +36,11 @@ export interface UserInfo {
 export interface ExternalData {
   nationalRegistry: {
     data: NationalRegistry
+    date: string
   }
   userProfile: {
     data: UserInfo
+    date: string
```
Rikkiabb, line 43: How does the response date look like? In what format is the date?

Rikkiabb (commented):
Two questions from me 😊

libs/application/templates/children-residence-change/src/types/index.ts
VilliGunn: It is a string and that is the reason I need to use the `parseISO` when formatting the date..

PR #3423, opened 2021-04-06
feat(crc): Dokobit error handling

Rikkiabb:
# \<Description\>
Shows custom errors based on what we get from Dokobit. There is one 'bug' in the error handling which is dokobit throws 'User cancelled' excepetion before the timeout happens, I have synched with Guðjón who is the owner of the signing service and we are working out actionables. 

Error codes from dokobit are based of: https://support.dokobit.com/article/545-error-handling-for-documents-api-and-identification-api
## Checklist:


apps/application-system/api/src/app/modules/application/files/utils/constants.ts
```diff
@@ -26,3 +26,10 @@ export const BucketTypePrefix: KeyMapping<PdfTypes, string> = {
 export const DokobitFileName: KeyMapping<PdfTypes, string> = {
   ChildrenResidenceChange: 'Lögheimilisbreyting-barns.pdf',
 }
+
+export const DokobitErrorCodes: { [key: string]: number } = {
```
VilliGunn, line 30: Do we need to define the type here? 
```suggestion
export const DokobitErrorCodes = {
```

libs/application/templates/children-residence-change/src/fields/Overview/SignatureModal/ErrorMessage.tsx
```diff
@@ -1,21 +1,41 @@
-import React from 'react'
-import { useIntl } from 'react-intl'
+import React, { useState, useEffect } from 'react'
```
VilliGunn, line 1: Just a suggestion, different style really and without useEffect or useState. I've not tested this implementation though.
```
import React from 'react'
import { Box, AlertMessage, Button } from '@island.is/island-ui/core'
import { signatureModal } from '../../../lib/messages'

const errorMessages = {
  404: signatureModal.error.noElectronicId,
  400: signatureModal.error.userCancelled,
  409: signatureModal.error.timeOut,
}

interface ErrorMessageProps {
  onClose: () => void
  errorCode?: keyof typeof errorMessages
}

const ErrorMessage = ({ onClose, errorCode }: ErrorMessageProps) => {
  return (
    <>
      <AlertMessage
        type="error"
        title={formatMessage(signatureModal.defaultError.title)}
        message={formatMessage(
          errorCode
            ? errorMessages[errorCode]
            : signatureModal.defaultError.message,
        )}
      />
      <Box marginTop={3} justifyContent="center">
        <Button onClick={onClose} variant="primary">
          {formatMessage(signatureModal.general.closeButtonLabel)}
        </Button>
      </Box>
    </>
  )
}

export default ErrorMessage
```

libs/application/templates/children-residence-change/src/fields/Overview/SignatureModal/ErrorMessage.tsx
```diff
@@ -1,21 +1,41 @@
...
+        break
+      default:
+        setErrorMessage(signatureModal.defaultError.message)
+    }
+  }, [errorCode])
+
   return (
     <>
       <AlertMessage
         type="error"
```
VilliGunn, line 36: Wondering if we should be using a different type here if for example the user aborts. 
https://ui.devland.is/?path=/docs/alerts-alertmessage--info - we could use warning or info in that case

libs/application/templates/children-residence-change/src/fields/Overview/fileSignatureReducer.tsx
```diff
@@ -21,11 +21,19 @@ export enum FileSignatureActionTypes {
 
 type Action =
   | { type: Exclude<FileSignatureActionTypes, FileSignatureActionTypes.ERROR> }
-  | { type: FileSignatureActionTypes.ERROR; error: string; status: ErrorStatus }
+  | {
+      type: FileSignatureActionTypes.ERROR
+      error?: number
```
VilliGunn, line 26: Should this be optional? Just thinking if we should pass a default error code if there isn't one

libs/application/templates/children-residence-change/src/fields/Overview/fileSignatureReducer.tsx
```diff
@@ -21,11 +21,19 @@ export enum FileSignatureActionTypes {
...
+      status: ErrorStatus
+    }
 
 export type ReducerState =
   | { status: Exclude<FileSignatureStatus, ErrorStatus>; modalOpen: boolean }
-  | { status: ErrorStatus; errorCode: string; modalOpen: boolean }
+  | {
+      status: ErrorStatus
+      errorCode?: number
```
VilliGunn, line 34: Same here, it seems strange to maybe not have an errorCode when we are in an error state

libs/application/templates/children-residence-change/src/fields/Overview/SignatureModal/ErrorMessage.tsx
Rikkiabb: This didn't quite work but did a similar approach.

VilliGunn (approved):

PR #3424, opened 2021-04-06
feat(cms): subarticle search

kksteini:
# \<Description\>

Subarticles (Baby Articles) weren't showing up in search results

https://user-images.githubusercontent.com/77672665/116001348-93c12500-a5e3-11eb-8440-93004499f8c8.mp4

## What

Contentful models were changed so that now; subarticles are aware of their parents. Changes made to the search so that subarticles can appear in results. 

## Why

Subarticles need to know their parent and/or their parent's slug so that the subarticles can be accessed from within the search.

## Checklist:


apps/web/components/SearchInput/SearchInput.tsx
```diff
@@ -424,29 +426,46 @@ const Results: FC<{
...
                         onClick(e)
                         onRouting()
                       }}
                     >
-                      <Link {...linkResolver(__typename as LinkType, [slug])}>
-                        <Text variant="h5" color="blue400">
-                          {title}
-                        </Text>
-                      </Link>
+                      {item.parentSlug ? (
```
kksteini, line 446: Note: Here I am testing how the data gets delivered to the front.
I highly suspect this will be changed by someone who knows the frontend well

apps/web/screens/Search/Search.tsx
```diff
@@ -187,11 +188,20 @@ const Search: Screen<CategoryProps> = ({
   }
 
   const searchResultsItems = (searchResults.items as Array<
-    Article & LifeEventPage & AboutPage & News & AdgerdirPage
+    Article & LifeEventPage & AboutPage & News & AdgerdirPage & SubArticle
   >).map((item) => ({
-    title: item.title,
-    description: item.intro ?? item.seoDescription ?? item.description,
-    link: linkResolver(typenameResolver(item.__typename), [item.slug]),
+    title: item.parent ? item.parent.title + '>' + item.title : item.title,
```
kksteini, line 193: NOTE: Here, again, I am testing how the data gets delivered to the front.
This might be unnecessary and not according to the accepted design document.

libs/api/domains/cms/src/lib/models/articleReference.ts
```diff
@@ -0,0 +1,29 @@
+import { Field, ObjectType, ID } from '@nestjs/graphql'
```
kksteini, line 1: I had to create this file. If we refactor `subArticles` out of the `article` model then we can get rid of this. Webpack failed to resolve the circularity introduced by attaching a parent `article` to the `subArticle` model.
This acted as a patch or "plaster" for that.

apps/web/screens/Search/Search.tsx
```diff
@@ -194,15 +195,33 @@ const Search: Screen<CategoryProps> = ({
...
+    parentTitle: item.parent?.title,
+    description:
+      item.intro ??
+      item.seoDescription ??
+      item.description ??
+      item.parent.intro,
+    link:
+      item.slug.split('/').length > 1
+        ? linkResolver(typenameResolver(item.__typename), item.slug.split('/'))
+        : linkResolver(typenameResolver(item.__typename), [item.slug]),
```
stefanhar, line 224: The check here might not be needed, split would result in an array with the slug even if slash is missing:
```suggestion
    link: linkResolver(typenameResolver(item.__typename), item.slug.split('/'))
```

libs/api/domains/cms/src/lib/models/articleReference.ts
stefanhar: I see, maybe add a comment to both models to indicate this.
We should also mark the other model as depreciated (e.g. via comment) and set a date for deletion

libs/api/domains/cms/src/lib/search/importers/article.service.ts
```diff
@@ -50,6 +64,9 @@ export class ArticleSyncService implements CmsSyncProvider<IArticle> {
             relatedArticles: (relatedArticles.length
               ? relatedArticles
               : undefined) as IArticleFields['relatedArticles'],
+            subArticles: (subArticles.length
+              ? subArticles
+              : undefined) as IArticleFields['subArticles'],
```
stefanhar, line 69: This might cause the article document to contain alot of unused data since articles only use title in display, consider implementing a more aggressive prune here to patch this until nested resolvers or other permanent solutions are implemented

stefanhar (changes requested):

apps/web/components/SearchInput/SearchInput.tsx
```diff
@@ -427,24 +429,33 @@ const Results: FC<{
               {(search.results.items as Article[] &
                 LifeEventPage[] &
                 AboutPage[] &
-                News[])
+                News[] &
+                SubArticle[])
                 .slice(0, 5)
-                .map(({ id, title, slug, __typename }) => {
+                .map((item) => {
+                  console.log(item)
```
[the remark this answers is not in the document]
kksteini: Whoops, thanks

apps/web/screens/Search/Search.tsx
kksteini: Oh, you are right. Nice catch :+1:

libs/api/domains/cms/src/lib/search/importers/article.service.ts
kksteini: You are right. It only needs the slug/url and the title.

libs/api/domains/cms/src/lib/models/articleReference.ts
kksteini: I have added an explanation. The article reference will be removed once subArticles have been successfully removed from articles.

apps/web/components/SearchInput/SearchInput.tsx
```diff
@@ -427,24 +429,32 @@ const Results: FC<{
...
                         onRouting()
                       }}
                     >
-                      <Link {...linkResolver(__typename as LinkType, [slug])}>
+                      <Link
+                        {...linkResolver(
+                          item.__typename as LinkType,
+                          item.slug.split('/').length > 1
+                            ? item.slug.split('/')
+                            : [item.slug],
```
stefanhar, line 453: Missed this earlier, same as below:
```suggestion
                          item.slug.split('/')
```

stefanhar (approved):

apps/web/components/SearchInput/SearchInput.tsx
kksteini: I should've known as well

eirikurn (approved):
Only reviewed the mock. Feel free to codeown the mock domain.

PR #3427, opened 2021-04-06
feat(dpo): check if current user is last modifier of organisation

fjandakornid:
# Check if current user is last modifier of organisation

## What

Check if current user is last modifier of organisation before creating provider.

## Why

For security reason. Since the createProvider method is in public api, we have to check if the current user created the organisation (since it is a former step of the application where the createProvider method is used), and if so, then he is allowed to create provider.

## Checklist:


hinriksteinar (approved):
LGTM

PR #3430, opened 2021-04-06
feat(web): Add organization secondary menu

Ingimarsson:
# Add organization secondary menu

## What

Adds option for secondary purple menu on organization pages. Also fixes header color scheme for subpages.

## Why

This is according to design.

## Screenshots / Gifs

Screenshot from 2021-04-06 14-33-22

## Checklist:

### DIFF
--- a/libs/api/domains/cms/src/lib/models/organizationPage.model.ts
+++ b/libs/api/domains/cms/src/lib/models/organizationPage.model.ts
@@ -38,6 +38,9 @@ export class OrganizationPage {
   @Field(() => [LinkGroup])
   menuLinks!: Array<LinkGroup>
 
+  @Field(() => LinkGroup, { nullable: true })
+  secondaryMenu!: LinkGroup | null
+
   @Field(() => Organization)
   organization!: Organization | null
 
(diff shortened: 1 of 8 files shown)


libs/api/domains/cms/src/lib/models/organizationPage.model.ts
```diff
@@ -38,6 +38,9 @@ export class OrganizationPage {
   @Field(() => [LinkGroup])
   menuLinks!: Array<LinkGroup>
 
+  @Field(() => LinkGroup, { nullable: true })
+  secondaryMenu!: LinkGroup | null
+
```
stefanhar, line 43: If this is not exclusively external links we should consider making this an actual menu to support reference links for dynamic link resolution

Ingimarsson: They are mostly internal links to organization subpages, those have two slugs `/stofnanir/[pageSlug]/[subpageSlug]`. We need to change Reference Links or the UI extension to make that work well. We're gonna do that in a separate PR as we discussed :)

stefanhar: Yes, supporting this type in ReferenceLink would be nice for consistency until we finish the UI extension work.
We should look into that ASAP in a separate PR, shouldn't be to complicated since OrganizationPage already includes the parent ref, could do similar logic to typeMap but with custom link resolvers to support special cases

ingolfurs (approved):
LGTM

stefanhar (approved):

PR #3431, opened 2021-04-06
feat(service-portal): Remove education degree page

barabrian:
# Remove education degree page

## What

Remove education degree page

## Why

Education degree page is not production ready

## Checklist:

### DIFF
--- a/apps/service-portal/src/store/modules.ts
+++ b/apps/service-portal/src/store/modules.ts
@@ -29,7 +29,6 @@ export type ModuleKeys =
   | 'finance'
   | 'settings'
   | 'education'
-  | 'educationDegree'
   | 'educationLicense'
   | 'educationCareer'
   | 'educationStudentAssessment'
@@ -38,7 +37,6 @@ export type ModuleKeys =
 export const featureFlaggedModules: ModuleKeys[] = [
   'documentProvider',
   'education',
-  'educationDegree',
   'educationLicense',
   'educationCareer',
   'educationStudentAssessment',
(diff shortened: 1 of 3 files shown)


libs/service-portal/core/src/lib/navigation/masterNavigation.ts
```diff
@@ -73,6 +73,35 @@ export const servicePortalMasterNavigation: ServicePortalNavigationItem[] = [
           icon: 'people',
         },
       },
+
+      // Menntun
```
arnorhs, line 77: Is this comment helpful?

arnorhs (changes requested):

atligudl (approved):

apps/service-portal/src/store/modules.ts
```diff
@@ -38,7 +37,6 @@ export type ModuleKeys =
 export const featureFlaggedModules: ModuleKeys[] = [
   'documentProvider',
   'education',
-  'educationDegree',
   'educationLicense',
```
atligudl, line 40: Just wondering, why declare separate modules for each of these education sections instead of just including them all under the education module?

barabrian: That's a good question! I'm not sure what the preferred method of working with modules is for the service portal, this is how I was told how it's done and it worked so I didn't question it 😄

libs/service-portal/core/src/lib/navigation/masterNavigation.ts
barabrian: That's up for debate, it's consistent to all the other defined navigation items so it might help you skim through the list and if you are used to collapse code you are not working with and collapse the 4. level you will have a nice overview of the available navigation items

stefanhar (approved):
LGTM

apps/service-portal/src/store/modules.ts
atligudl: Each module can declare multiple routes (and even internal routing) so the idea was that elements like these should all be grouped together under education, this might be a good topic for the next discipline meeting so lets bring it up then 😀

PR #3436, opened 2021-04-07
fix(car-recycling): fix redirect after logout

dd-sdang:
# Fix redirect after logout

## What

Redirect user to /skilavottord after logout instead of /app/skilavottord

## Why

Because the old logout link does not exist. 

## Checklist:


tommigitgaur (approved):
LFTM

PR #3444, opened 2021-04-07
feat(service-portal): Education overview feature flags

barabrian:
# Education overview feature flags

## What

Use Configcat to filter navigation

## Checklist:

### DIFF
--- a/libs/service-portal/education/src/environments/environment.prod.ts
+++ b/libs/service-portal/education/src/environments/environment.prod.ts
@@ -1,6 +1,0 @@
-import { getStaticEnv } from '@island.is/utils/environment'
-
-export default {
-  production: true,
-  featureFlagSdkKey: getStaticEnv('SI_PUBLIC_CONFIGCAT_SDK_KEY'),
-}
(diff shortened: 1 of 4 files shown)


libs/service-portal/education/src/environments/environment.prod.ts
```diff
@@ -0,0 +1,6 @@
+import { getStaticEnv } from '@island.is/utils/environment'
+
```
dabbeg, line 2: libs do not use the filereplacements as apps do.
So only apps have `environment.prod.ts` and `environment.ts`.

I would only use the environment.ts file and use getStaticEnv there

libs/service-portal/education/src/environments/environment.ts
```diff
@@ -0,0 +1,7 @@
+// This file can be replaced during build by using the `fileReplacements` array.
+// When building for production, this file is replaced with `environment.prod.ts`.
+
+export default {
+  production: false,
```
dabbeg, line 5: no need for production variable

dabbeg (changes requested):
Two comments, main thing that libs do not have file replacements

dabbeg (approved):

PR #3459, opened 2021-04-09
feat(content-search-toolkit): strictly type content-search-toolkit

jeremybarbet:
## What

- Imported the CMS module inside the application-system-api which is `strict: true` and throw errors because of `content-search-toolkit` not being strictly typed. Now it is


stefanhar (approved):
Nice

libs/content-search-toolkit/src/services/elastic.service.ts
```diff
@@ -187,10 +190,17 @@ export class ElasticService {
     })
 
     const results = await Promise.all(requests)
-    return results.reduce((groupedResults, result, index) => {
-      groupedResults[metrics[index]] = result
-      return groupedResults
-    }, {})
+    return results.reduce(
```
eirikurn, line 193: I think you can also put the result type as generic argument, which types the handler and second argument.

```
results.reduce<Record<string, RankEvaluationResponse<searchTermUnion>>>(...)
```

Also helps to define long types before, 

```
type RankResultMap = Record<string, RankEvaluationResponse<searchTermUnion>>
```

eirikurn (approved):
LGTM

madebynoam (approved):

ingolfurs (approved):

libs/content-search-toolkit/src/services/elastic.service.ts
jeremybarbet: Good point, thanks!

PR #3460, opened 2021-04-09
fix(crc): stop throwing an error when signature fails

VilliGunn:


libs/application/templates/children-residence-change/src/fields/Overview/index.tsx
```diff
@@ -109,12 +109,12 @@ const Overview = ({
           return response.data?.requestFileSignature?.documentToken
         })
         .catch((error: ApolloError) => {
+          console.log('error', error.graphQLErrors[0].extensions?.code)
```
Rikkiabb, line 112: 👀

Rikkiabb (changes requested):
No way!

Rikkiabb (approved):

PR #3462, opened 2021-04-09
feat(judicial-system): Prosecutor files list in overview

oddsson:
# Prosecutor files list in Overview

https://app.asana.com/0/1199153462262248/1200058740837141

## What

Display a list of files in the prosecutor's overview.

## Why

The Overview screen is a summary of the case and the files are a part of the case.

## Screenshots / Gifs

Screen Shot 2021-04-09 at 14 01 30

## Checklist:


apps/judicial-system/web/src/routes/Prosecutor/Overview/Overview.tsx
```diff
@@ -326,8 +327,26 @@ export const Overview: React.FC = () => {
...
+                    id="id_5"
+                    label={`Rannsóknargögn ${`(${workingCase.files.length})`}`}
+                    labelVariant="h3"
+                  >
+                    {workingCase.files?.map((file, index) => (
+                      <Box marginBottom={3}>
+                        <CaseFile
+                          name={`${index + 1}. ${file.name}`}
+                          size={file.size}
+                          uploadedAt={file.modified}
```
gudjong, line 341: I suggest you use "created" instead of "modified". Currently, they are the same, but later we will be updating the entries in the "file" table and then "modified" will not reflect the upload time.

gudjong (changes requested):

apps/judicial-system/web/src/routes/Prosecutor/Overview/Overview.tsx
oddsson: Cool, thanks, fixed 🙏

gudjong (approved):

PR #3476, opened 2021-04-12
fix(api-domains-cms): Enabled strict and fixed a remaining typescript error

kksteini:
# \<Description\>
Forgot to enable strict in the tsConfig. Fixed an error introduced since first error pruning.


stefanhar (approved):
LGTM

PR #3491, opened 2021-04-13
feat(endorsement-system): Endorsement metadata

stefanhar:
# \<Description\>

Validating and displaying endorsements requires different metadata based on validation and display rules for each list.

## What

- Added metadata service that fetches as resolves only required data for each list validation
- Added option to persist metadata in meta field on each endorsement
- Updated tests
- Changed listId uuid validation to use build in UUIDPipe
- Updated delete code
- Added custom validation rule to handle different error messages
- Removed unused config from tsconfig

## Why

- We want to allow for extendability to ease implementation into other systems
- We want to allow for flexibility when persisting user data by assigning metadata per list
- We want to have the ability to add new metadata to endorsements easily e.g. to extend validation rules

## Checklist:


kksteini (approved):
LGTM

siggakogk (approved):
❇️ nice!

PR #3496, opened 2021-04-14
feat(contentful-translation-extension): hide the 'removed messages' behind a toggle button

jeremybarbet:
## What

- I was updating namespaces and a bunch of messages were deleted here and there. Added a simple toggle button to hide them by default


eirikurn (approved):
👍

PR #3503, opened 2021-04-14
feat(crc): Sends sms to counterparty on parent a submission

Rikkiabb:
# \<Description\>
Sends sms to parent b when parent a submits application if parent b phone number is registered.

TODO: Update copy of sms text (yet to be designed) and add variables to helm.

## Checklist:

### DIFF
--- a/libs/application/template-api-modules/src/lib/modules/templates/children-residence-change/children-residence-change.service.ts
+++ b/libs/application/template-api-modules/src/lib/modules/templates/children-residence-change/children-residence-change.service.ts
@@ -20,6 +20,7 @@ import {
   transferRequestedEmail,
 } from './emailGenerators'
 import { Application } from '@island.is/application/core'
+import { SmsService } from '@island.is/nova-sms'
 
 export const PRESIGNED_BUCKET = 'PRESIGNED_BUCKET'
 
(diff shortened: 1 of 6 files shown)


libs/application/template-api-modules/src/lib/modules/templates/children-residence-change/children-residence-change.service.ts
```diff
@@ -153,5 +155,12 @@ export class ChildrenResidenceChangeService {
         (application as unknown) as Application,
       )
     }
+
+    if (counterParty.phoneNumber) {
+      await this.smsService.sendSms(
+        counterParty.phoneNumber,
+        'Borist hefur umsókn um breytt lögheimili barns.',
```
VilliGunn, line 162: Would it be possible to send a link to the application? Also the copy here should probably be revisited.

Rikkiabb: Yes noted it in the description. I decided to have this rather bad than good so that we would defentely revisit it 😅 And yes we could send a link :)

VilliGunn (approved):

baering (approved):

PR #3504, opened 2021-04-14
feat(crc): update overview screen according to new design

VilliGunn:
# \<Description\>

https://app.asana.com/0/1199733748699731/1200184411501103

## What

- Updated overview screen and translations
- Pushed translations to Contentful

## Why

New design

## Screenshots / Gifs

Screenshot 2021-04-14 at 17 46 51

## Checklist:

### DIFF
--- a/libs/application/templates/children-residence-change/src/fields/Overview/Overview.treat.ts
+++ b/libs/application/templates/children-residence-change/src/fields/Overview/Overview.treat.ts
@@ -1,0 +1,6 @@
+import { theme } from '@island.is/island-ui/theme'
+import { style } from 'treat'
+
+export const container = style({
+  marginTop: '-24px',
+})
--- a/libs/application/templates/children-residence-change/src/fields/Overview/Overview.treat.ts
+++ b/libs/application/templates/children-residence-change/src/fields/Overview/Overview.treat.ts
@@ -1,4 +1,3 @@
-import { theme } from '@island.is/island-ui/theme'
 import { style } from 'treat'
 
 export const container = style({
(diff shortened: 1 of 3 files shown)


libs/application/templates/children-residence-change/src/fields/Overview/Overview.treat.ts
```diff
@@ -0,0 +1,6 @@
+import { theme } from '@island.is/island-ui/theme'
+import { style } from 'treat'
+
+export const container = style({
```
VilliGunn, line 4: 🙈 - needed because of ghost div rendered by `FormMultiField` component.

libs/application/templates/children-residence-change/src/lib/messages/contract.ts
```diff
@@ -53,6 +40,18 @@ export const contract = {
...
   }),
+  childBenefit: defineMessages({
+    label: {
+      id: 'crc.application:section.contract.overview.childBenefit.label',
+      defaultMessage: 'Meðlag',
+      description: 'Label for child benefit',
+    },
+    text: {
+      id: 'crc.application:section.contract.overview.childBenefit.text',
+      defaultMessage: '{otherParent} fær einfalt meðlag mánaðarlega.',
```
Rikkiabb, line 51: Does this line make sense if the parent has changing residence for multiple children?

VilliGunn: Yes it does because all the children will change to the same residence and therefore the child benefits will change for both children. After a team discussion we have decided to iterate on this later but this should not be a blocker for now.

Rikkiabb (approved):

PR #3507, opened 2021-04-15
feat(island-ui): Replace Animejs with Framer-Motion

oddsson:
# Replace Animejs with Framer-Motion

## What

Replace Animejs with Framer-Motion

## Why

Animejs is limited in that it's not a "React" library and therefore cannot do things in React that other libraries can, f.x. animating components on mount/unmount.

Why Framer-motion and not something else? I believe that Framer-motion is the gold standard of animation libraries. It provides an animation hook that is really useful.

## Screenshots / Gifs

https://user-images.githubusercontent.com/3789875/114852799-a4e38800-9dd2-11eb-9b8a-5e95ff2ccad6.mov

## Checklist:


eirikurn (commented):
I approve adopting framer motion instead of animejs. But I would reaaaally like if someone rewrote this specific animation as a css animation somehow and get rid of [95KB minified JS](https://bundlephobia.com/result?p=framer-motion@4.1.3) to render the first loading animation ... :)

eirikurn (approved):

oddsson:
Thank you for the feedback @eirikurn 🙏 I agree but I don't really know how to write this animation as a css animation tbh. I was able to make this a bit better by using _m, LazyMotion and domAnimation_ (5kb added to bundle) instead of motion (25kb added to bundle (Note that framer-motion is tree-shakable)). If someone knows how to rewrite this animation as a css animation, please show me how, I'd really like to learn 👨‍🎓

eirikurn:
Hmm. I was imagining this reimplemented as rounded div boxes, with a keyframe animation that animates an extra box with translate, width and height. But the gradient fill variant ruins it. No easy way to replicate that with a moving div.

So there's probably no way to animate this with css only. Maybe this can be done leaner with SMIL (discontinued?), Web Animations (supports css properties only?), vanillajs (messy?) or apng/gif (tricky to compress/maintain without visual artifacts?). But let's not ... 😃

Especially since this tree shakes to 5KB (gzipped surely?).

eirikurn (approved):

atligudl (approved):
LGTM

PR #3517, opened 2021-04-15
refactor(api): JSON animations retrieved from assets

kksteini:
# \<Description\>

Lazy loading of JSON animations

## What

Animations are uploaded to contentful assets and the animation data loaded via the asset URL

## Why

Decrease initial payload size. 

## Checklist:


apps/web/components/FrontpageSlider/FrontpageSlider.tsx
```diff
@@ -114,11 +115,19 @@ export const FrontpageSlider: FC<FrontpageSliderProps> = ({
   }
 
   useEffect(() => {
+    const requests = slides.reduce((all, slide) => {
+      if (slide.animationJsonAsset && slide.animationJsonAsset.url) {
+        all.push(axios.get(slide.animationJsonAsset.url))
```
eirikurn, line 120: Should we use `fetch` instead? Not sure if Axios is already bundled or not, but ideally we don't need it.

libs/api/domains/cms/src/lib/models/frontpageSlider.model.ts
```diff
@@ -35,4 +39,9 @@ export const mapFrontpageSlider = ({
   animationJson: fields.animationJson
     ? JSON.stringify(fields.animationJson)
     : '',
```
stefanhar, line 41: Consider removing this since it should no longer be needed

libs/api/domains/cms/src/lib/models/frontpageSlider.model.ts
```diff
@@ -21,6 +22,9 @@ export class FrontpageSlider {
 
   @Field({ nullable: true })
   animationJson?: string
```
stefanhar, line 24: Consider removing this since it should no longer be needed

stefanhar (changes requested):

apps/web/components/FrontpageSlider/FrontpageSlider.tsx
kksteini: Axios was already bundled

libs/api/domains/cms/src/lib/models/frontpageSlider.model.ts
kksteini: Good catch, thank you

stefanhar (approved):
LGTM

PR #3419, opened 2021-03-31
feat(core): implement audit module

eirikurn:
## What

Nest module to audit user actions in our backends.

Includes:

* New `@island.is/nest/audit` module.
* Refactored auth-nest-utils, to better handle different kinds of auth guards, and forwarding of credentials.
* Change from IdsAuthGuard to IdsUserGuard in most places.
* Audit integration in Application System

## Why

So we can audit who did what. This is especially important when we launch the delegation system which allows users to sign in on behalf of other users and companies.

## Checklist:

- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] Formatting passes locally with my changes
- [x] I have rebased against main before asking for a review

PR #3524, opened 2021-04-16
fix(application-system): check if user nationalId has access to application id

jeremybarbet:
## What

- Even with the mapUserToRole util we have in templates, the application payload is returned to the form. This PR checks if the national ID is part of the applicant or assignee otherwise it throws an error


valurefugl (approved):

apps/application-system/api/src/app/modules/application/tools/application.serializer.ts
```diff
@@ -25,22 +27,39 @@ export class ApplicationSerializer
...
+        }
+
+        const application = res as Application
+
+        if (
+          application.applicant !== user.nationalId &&
+          !application.assignees.includes(user.nationalId)
+        ) {
+          throw new NotFoundException('No application found for this user')
+        }
```
eirikurn, line 60: This check happens to late. It needs to be at the top of the controller handlers instead, so we block access to any update logic.

Maybe we should update `ApplicationController#findOne`, remove `ApplicationByIdPipe` and instead call a helper that fetches and checks access: `this._findOneByIdAndNationalId(id, user.nationalId)`.

libs/auth-nest-tools/src/lib/current-user.decorator.ts
```diff
@@ -1,26 +1,35 @@
...
 
+export const getCurrentUser = (context: ExecutionContext): User => {
+  const request = context.switchToHttp().getRequest()
+
+  if (!request) {
+    throw new UnauthorizedException('You are not authenticated')
+  }
+
+  return request.user
+}
```
eirikurn, line 31: Great change. :)

eirikurn (changes requested):

apps/application-system/api/src/app/modules/application/tools/application.serializer.ts
jeremybarbet: I did that first. But didn’t know how to pass the nationalID down to the pipeTransform so change to this. But sounds good. I will do that 👍

jeremybarbet: I created a small service for it that contains the function. Wasn't sure if it should live at the root of the controller or into it's own file

apps/application-system/api/src/app/modules/application/application.controller.ts
```diff
@@ -257,14 +256,11 @@ export class ApplicationController {
       throw new BadRequestException('Invalid token')
     }
 
-    const existingApplication = await this.applicationService.findOneById(
+    const existingApplication = await this.applicationAccessService.findOneByIdAndNationalId(
```
eirikurn, line 259: In the assign endpoint, the application is not yet assigned to the authenticated user, so it should just findById.

eirikurn (changes requested):
Sweeet! Just one issue.

apps/application-system/api/src/app/modules/application/application.controller.ts
jeremybarbet: I made the nationalId optional in the `findOneById` function so it can be used here for the assign endpoint

jeremybarbet:
Tests are failing because the context should contain the user object with the nationalId but I'm not sure how to mock the `getRequest()` in the tests 😰

jeremybarbet:
Will wait https://github.com/island-is/island.is/pull/3419 to be merged to use the tests there for the mockAuthGuard

eirikurn (approved):

eirikurn (approved):
LGTM

PR #3526, opened 2021-04-16
feat(application-system): validation improvements

VilliGunn:
# \<Description\>

- Changes the `partialSchemaValidation` return value from being a flat object to a nested object that reflects the actual data schema defined with Zod. Also removes redundant errors from the error object where nested properties are validated.

Example return values
```
// Zod schema
person: z
    .object({
      nationalId: string(),
      age: z.string(),
      email: z.email().nonEmpty(),
    })

// Old after submitting without touching the form
{
  'person': 'Invalid value', // Redundant
  'person.nationalId': 'Invalid value',
  'person.age': 'Invalid value',
  'person.email': 'Required',
}

// Old after focusing all fields and blurring
{
  'person': {
    'nationalId': 'Invalid value',
    'age': 'Invalid value',
    'email': 'Invalid email',
  },
  'nationalId': 'Invalid value', // redundant/wrong
  'age': 'Invalid value', // redundant/wrong
  'email': 'Invalid email', // redundant/wrong
}

// New  - Now we always return the same object
{
  'person': {
    'nationalId': 'Invalid value',
    'age': 'Invalid value',
    'email': 'Invalid email',
  },
}
```

- Adds partial object validation support.
```
selectDuration: z
    .object({
      type: z.enum([Duration.Permanent, Duration.Temporary]),
      date: z.string().optional(),
    })
    .refine((v) => (v.type === Duration.Temporary ? v.date : true), {
      message: error.validation.durationDate.defaultMessage,
      path: ['date'],
    }),
```

- Fixes application templates according to new error handling

## What

- Updated the `partialSchemaValidation` and `populateError` functions
- Updated the tests
- Changed data schema for Children residence change application
- Added a type for the Children residence change error object

## Why

- There are inconsistencies in the return error value
- It is better to reflect the error return value and return a nested object instead of flatten the error object with dot separated string as key
- It is currently not possible to clear errors after submitting without blurring the field
- Support partial/conditional object validation

## Checklist:


libs/application/core/src/validation/validators.ts
```diff
@@ -10,18 +14,17 @@ interface SchemaValidationError {
...
-): SchemaValidationError | undefined {
-  if (newError === undefined) {
-    return currentError
-  }
-
-  if (!currentError) {
-    return { [pathToError]: newError.errors[0].message }
+  newError: ZodSuberror[] | undefined,
+  pathToError?: string,
+) {
```
Rikkiabb, line 19: Could we write this function like this: 

```
function populateError(
  currentError?: SchemaValidationError = {},
  newError: ZodSuberror[] | undefined,
  pathToError?: string,
) {
  let errorObject = {}
  newError?.forEach((element) => {
    errorObject = set(errorObject, pathToError || element.path, element.message)
  })
   return merge(currentError, errorObject)
}
```

libs/application/core/src/validation/validators.ts
```diff
@@ -10,18 +14,17 @@ interface SchemaValidationError {
...
-  newError: ZodError | undefined,
-  pathToError: string,
-): SchemaValidationError | undefined {
-  if (newError === undefined) {
-    return currentError
-  }
-
-  if (!currentError) {
-    return { [pathToError]: newError.errors[0].message }
+  newError: ZodSuberror[] | undefined,
```
Rikkiabb, line 17: Is this at any time undefined?

libs/application/core/src/validation/validators.ts
```diff
@@ -49,37 +53,45 @@ function partialSchemaValidation(
     try {
       trimmedSchema.parse({ [key]: answer })
     } catch (e) {
-      error = populateError(error, e, newPath)
-
+      const zodErrors: ZodSuberror[] = e.errors
```
Rikkiabb, line 56: Should we think about commenting this function? 😅

libs/application/core/src/validation/validators.ts
```diff
@@ -49,37 +53,45 @@ function partialSchemaValidation(
...
+      }
       if (Array.isArray(answer)) {
         const arrayElements = answer as Answer[]
         arrayElements.forEach((el, index) => {
           try {
             trimmedSchema.parse({ [key]: [el] })
           } catch (e) {
-            const elementPath = `${newPath}[${index}]`
-            error = populateError(error, e, elementPath)
+            const elementPath = `${constructedErrorPath}[${index}]`
```
Rikkiabb, line 73: Is `indexedPath` a better name here? Since that is the only thing we are adding to the element path?

Rikkiabb (commented):
Awesome stuff, minor considerations from me 🥇

libs/application/core/src/validation/validators.spec.ts
```diff
@@ -317,9 +322,7 @@ describe('validateAnswers', () => {
       } as FormValue
       const secondError = validateAnswers(schemaWithArray, anotherBadFormValue)
       expect(secondError).toEqual({
-        person: invalidInput,
-        'person[1]': invalidInput,
-        'person[1].age': invalidInput,
+        person: [undefined, { age: invalidInput }],
```
Rikkiabb, line 325: Should we add tests for this changed functionality? Maybe especially some regarding arrays?

libs/application/core/src/validation/validators.ts
Rikkiabb: Another solution would be to change the order of variables into `merge` and have it like this `return merge(errorObject, currentError)`  
I think it would result in the same function and then we could skip the if check and returning error at the bottom of the function.

libs/application/core/src/validation/validators.spec.ts
VilliGunn: I think this is sufficient for arrays with objects.

libs/application/core/src/validation/validators.ts
```diff
@@ -49,37 +49,44 @@ function partialSchemaValidation(
     try {
       trimmedSchema.parse({ [key]: answer })
     } catch (e) {
-      error = populateError(error, e, newPath)
-
+      const zodErrors: ZodSuberror[] = e.errors
+      const keyIsIncludedInErrors = zodErrors.some((err) =>
+        err.path.includes(key),
```
eirikurn, line 54: Is this a good enough check, eg if there are two `name` fields in different places? Can we check that the path starts with `constructedErrorPath`?

eirikurn (approved):
Nice! Just one question from me.

libs/application/core/src/validation/validators.ts
VilliGunn: Good question, this code was actually some leftover since debugging and is not needed at all 😄

Thorrik58 (approved):
LGTM

Rikkiabb (approved):
🚀

jeremybarbet (approved):

fjandakornid (approved):
LGTM

PR #3527, opened 2021-04-16
feat(service-portal): Icelandic names registry

stjanilofts:
## What

A service-portal module for managing the Icelandic names registry DB.

## Why

This is part of closing down https://vefur.island.is/mannanofn/. Once this is up and running there will be a new name searcher added on https://island.is/mannanofn/ which will be used instead.

## Screenshots / Gifs

image

## Checklist:


eirikurn:
I don't like duplicating the service portal code just for this one admin. The future is for these admin interfaces to be access controlled inside the service portal or a standalone admin portal.

So can we add this as a feature-flagged module in the service portal that only specific users see? Seeing how the APIs are already locked down to specific users. Can either use an environment variable or ConfigCat to control who can see the module.

hrefnalind:
I agree with Eiki. We should not duplicate the code for this project only. I know we have plan on creating admin portal in near future but first we need to create the architecture and review the structure before we go into implementation.  Should we meet up next week to discuss? @stjanilofts ?

libs/service-portal/icelandic-names-registry/src/components/EditForm/EditForm.tsx
```diff
@@ -0,0 +1,364 @@
...
+                        required
+                      />
+                    )
+                  }}
+                />
+              </GridColumn>
+            </GridRow>
+
+            <GridRow marginTop={3}>
+              <GridColumn span={'12/12'} paddingBottom={spacing}>
```
barabrian, line 222: unnecessary grid row and column? maybe just use box with margin padding?

barabrian: this applies to all cases where there is a single column with span 1

barabrian (approved):
Looks fine, some minor optional comments

libs/service-portal/icelandic-names-registry/src/components/EditForm/EditForm.tsx
stjanilofts: I need to do changes to the layout. Will change this as well.

libs/service-portal/icelandic-names-registry/codegen.yml
```diff
@@ -0,0 +1,18 @@
+schema:
+  - apps/api/src/api.graphql
+documents: libs/service-portal/icelandic-names-registry/src/queries/*.{ts,tsx}
+generates:
+  libs/service-portal/icelandic-names-registry/src/queries/schema.ts:
+    plugins:
+      - typescript
+      - typescript-operations
```
eirikurn, line 8: This might be copy pasta, but I recommend checking out the typescript-react-apollo plugin: https://www.graphql-code-generator.com/docs/plugins/typescript-react-apollo

It creates hooks for you to use and skip some more boilerplate.

libs/service-portal/icelandic-names-registry/src/components/Editor/Editor.tsx
```diff
@@ -0,0 +1,266 @@
...
+        variables: { input: { id, body } },
+      })
+        .then(() => {
+          toast.success(`Nafn var uppfært`)
+          setIsVisible(false)
+          doSearch()
+        })
+        .catch(() => {
+          toast.error(`Villa kom upp`)
+        })
```
eirikurn, line 134: Why mix `await` and `then`?

eirikurn (approved):
Should assign a codeowner to the service portal module.

Otherwise LGTM. Just minor nitpick. Also curious if any admin modules will be properly translated.

obmagnusson (approved):

PR #3536, opened 2021-04-19
fix(parental-leave): Fixes to duration slider screen

madebynoam:
This PR fixes 2 issues and updates 1 value:
1. Allow the user to move the duration slider beyond 6 months
2. Resolve an error if the user refreshes while on the duration slider screen.
3. Change the maxMonths value to 24

Notes:
On the duration slider screen, we can't determine if the user will spread their leave beyond 6 months with a percent ratio lower than 100. So we will take out the check `"if the endDate is inside the allowed range of days/months"` for now.

## Checklist:

### DIFF
--- a/libs/application/core/src/lib/formUtils.ts
+++ b/libs/application/core/src/lib/formUtils.ts
@@ -41,7 +41,7 @@ export function getValueViaPath(
 ): unknown | undefined {
   // Errors from dataSchema with array of object looks like e.g. `{ 'periods[1].startDate': 'error message' }`
   if (path.match(/.\[\d\]\../g) && !containsArray(obj)) {
-    return obj?.[path]
+    return obj?.[path] || defaultValue
   }
 
   // For the rest of the case, we are into e.g. `personalAllowance.usePersonalAllowance`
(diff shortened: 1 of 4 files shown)


libs/application/core/src/lib/formUtils.ts
```diff
@@ -41,7 +41,7 @@ export function getValueViaPath(
 ): unknown | undefined {
   // Errors from dataSchema with array of object looks like e.g. `{ 'periods[1].startDate': 'error message' }`
   if (path.match(/.\[\d\]\../g) && !containsArray(obj)) {
-    return obj?.[path]
+    return obj?.[path] || defaultValue
```
jeremybarbet, line 44:
```suggestion
    return obj?.[path] ?? defaultValue
```

jeremybarbet (commented):
Just a minor comment

libs/application/core/src/lib/formUtils.ts
madebynoam: Nice!

jeremybarbet (approved):
LGTM

PR #3542, opened 2021-04-19
fix(core): Fixes security audit

oddsson:
# Fix security audit error

## What

Running `yarn run security-audit` on main results in an error.

## Why

No new PR's can go through the pipe while this happens.

## Screenshots / Gifs

Screen Shot 2021-04-19 at 15 55 24

## Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -389,7 +389,6 @@
     "elliptic": "6.5.4",
     "serialize-to-js": "3.1.1",
     "netmask": "2.0.1",
-    "ssri": "8.0.1",
-    "jose": "1.28.1"
+    "ssri": "8.0.1"
   }
 }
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -389,6 +389,7 @@
     "elliptic": "6.5.4",
     "serialize-to-js": "3.1.1",
     "netmask": "2.0.1",
-    "ssri": "8.0.1"
+    "ssri": "8.0.1",
+    "jose": "1.28.1"
```
eirikurn, line 393: Let's remove this resolution before merging. It's not needed since this is a safe upgrade.

eirikurn (approved):
I removed automerge so you can address my comment and then add the label back after pushing.

PR #3554, opened 2021-04-20
feat(crc): parent-b-landing-page

Rikkiabb:
# \<Description\>
Adds landing page for parent b.

## Screenshots / Gifs
Screenshot 2021-04-20 at 13 02 17

## Checklist:


libs/application/templates/children-residence-change/src/fields/ParentBIntro/index.tsx
```diff
@@ -0,0 +1,94 @@
...
+  const { externalData, answers } = application
+  const { formatMessage } = useIntl()
+  const applicant = externalData.nationalRegistry.data
+  const childResidenceInfo = childrenResidenceInfo(applicant, answers)
+  const children = getSelectedChildrenFromExternalData(
+    applicant.children,
+    answers.selectedChildren,
+  )
+  return (
+    <>
```
Rikkiabb, line 24: Some stuff here could be extracted into components, not sure if we want to, what do you think @VilliGunn ?

VilliGunn: Yes we could extract some of this to a component to use on the overview screen and success screens. I suggest we do it when working on the success screen task.

libs/application/templates/children-residence-change/src/fields/ParentBIntro/index.tsx
```diff
@@ -0,0 +1,94 @@
...
+      <Box marginTop={3} marginBottom={5}>
+        <Text>
+          {formatMessage(parentBIntro.general.description, {
+            otherParentName: applicant.fullName,
+          })}
+        </Text>
+      </Box>
+      <Box marginTop={4} marginBottom={4}>
```
VilliGunn, line 32: It seems strange to have both margin top and bottom on these. I would prefer if we stick to only bottom or top here.

libs/application/templates/children-residence-change/src/fields/ParentBIntro/index.tsx
```diff
@@ -0,0 +1,94 @@
...
+        </Text>
+        <Text variant="h4" color="blue400">
+          {childResidenceInfo.future.parentName}
+        </Text>
+        <Text fontWeight="light">
+          {formatAddress(childResidenceInfo.future.address)}
+        </Text>
+      </Box>
+      <Text variant="h4">{formatMessage(parentBIntro.disagreement.title)}</Text>
+      <Box marginTop={1} marginBottom={2}>
```
VilliGunn, line 65: Same here, I would prefer to have `marginBottom` on the `<Text>` component and only `marginBottom` on Box

VilliGunn (commented):
Nitpick from me and future suggestions 😃

libs/application/templates/children-residence-change/src/fields/ParentBIntro/index.tsx
VilliGunn: I also think the margin bottom should be set to 5 instead of 2 here? 40px instead of 16px

libs/application/templates/children-residence-change/src/fields/ParentBIntro/index.tsx
```diff
@@ -0,0 +1,94 @@
...
+        <DescriptionText
+          text={parentBIntro.disagreement.description}
+        ></DescriptionText>
+      </Box>
+      <Box marginBottom={5}>
```
VilliGunn, line 70:
```suggestion
      <Box marginBottom={4}>
```

VilliGunn (approved):
Nice 🚀

PR #3558, opened 2021-04-20
fix(auth): IDS silent renew should update graphql client

eirikurn:
## What

Fix silent renew so GraphQL requests continue working after access token expires.

## Why

We are lowering the lifetime of access token to 5 minutes for security reasons, so silent renew needs to work for smooth UX.

Note: I'm working on refactoring this stuff into a shared library, where I plan to make a couple of extra improvements (monitor IDS session and redirect to login page, only silently renew the access token if the user is active).

## Checklist:


atligudl (approved):

jeremybarbet (approved):
Nice!

PR #3564, opened 2021-04-21
fix(web): ARIA landmarks

ingolfurs:
# Fix ARIA landmarks usage

## What
Scan with [axe DevTools](https://chrome.google.com/webstore/detail/axe-devtools-web-accessib/lhdoppojpmngadmnindnejefpokejbdd) (web accessibility testing) detects accessibility issues related to usage of aria landmarks.

- [Document does not have a main landmark](https://dequeuniversity.com/rules/axe/4.1/landmark-one-main)
- [All page content must be contained by landmarks](https://dequeuniversity.com/rules/axe/4.1/region)

The [Main](https://github.com/island-is/island.is/blob/dd4fcc3d3a7f0248d6389731933050757fa78a50/apps/web/components/Main/Main.tsx) component was probably a bit too complicated. It was supposed to detect itself if it should add the `<main>` landmark html tag, by checking if child element with the id main-content existed and then not add `<main>` tag to parent components.

This was broken by accident in [#2587](https://github.com/island-is/island.is/pull/2587) when the id `main-content` was moved from Main to suitable components to fix the skip to main content button. It led to that `<main>` tag is never added to the document, causing this accessibility issue.

Instead of trying this _magic_ detection inside the Main component, now each screen controls how the `<main>` tag is used. It perhaps becomes a bit complicated when pages have sidebar, because we want sidebars to be inside `<aside>`, but [complementary landmarks and/or asides must be top level](https://dequeuniversity.com/rules/axe/4.1/landmark-complementary-is-top-level) may not be located inside `<main>`. That's why the main html tag is located differently on pages with sidebar versus full width pages.

## Why
We've already received award for being accessible website and we need to continue make great things great. Using aria landmarks correctly gives screen reader users better navigation between page sections.

## Checklist:

### DIFF
--- a/apps/web/components/Main/Main.tsx
+++ b/apps/web/components/Main/Main.tsx
@@ -18,11 +18,12 @@ export const Main: React.FC<MainProps> = ({
         outline: 'none',
       }
     : {}
-  const boxId = addId
-    ? {
-        id: 'main-content',
-      }
-    : {}
+  const boxId =
+    addLandmark && addId
+      ? {
+          id: 'main-content',
+        }
+      : {}
   return (
     <Box {...boxProps} {...boxId}>
       {children}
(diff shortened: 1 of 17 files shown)


Ingimarsson (approved):

ingolfurs:
Will be rebased and requested for review later

PR #3567, opened 2021-04-21
feat(application-templates-party-letter): Collect endorsements

stefanhar:
# \<Description\>

Collect endorsements for given list

## What

- We want to be able to collect endorsements for a given party letter

## Checklist:

### DIFF
--- a/workspace.json
+++ b/workspace.json
@@ -970,6 +970,43 @@
         }
       }
     },
+    "api-domains-endorsement-system": {
+      "root": "libs/api/domains/endorsement-system",
+      "sourceRoot": "libs/api/domains/endorsement-system/src",
+      "projectType": "library",
+      "schematics": {},
+      "architect": {
+        "lint": {
+          "builder": "@nrwl/linter:lint",
+          "options": {
+            "linter": "eslint",
+            "tsConfig": [
+              "libs/api/domains/endorsement-system/tsconfig.lib.json",
+              "libs/api/domains/endorsement-system/tsconfig.spec.json"
+            ],
+            "exclude": [
+              "**/node_modules/**",
+              "!libs/api/domains/endorsement-system/**/*"
+            ]
+          }
+        },
+        "test": {
+          "builder": "@nrwl/jest:jest",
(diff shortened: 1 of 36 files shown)


workspace.json
```diff
@@ -6035,7 +6072,11 @@
           "builder": "@nrwl/workspace:run-commands",
           "options": {
             "outputPath": "apps/services/endorsements/api/src/openapi.yaml",
-            "command": "yarn ts-node -P apps/services/endorsements/api/tsconfig.app.json apps/services/endorsements/api/src/buildOpenApi.ts"
+            "commands": [
+              "nx run services-endorsements-api:build",
+              "yarn ts-node -P apps/services/endorsements/api/tsconfig.app.json dist/apps/services/endorsements/api/buildOpenApi.js"
```
jeremybarbet, line 6077: shouldn't it be `.ts`?

jeremybarbet (commented):
Just 1 comment for me

workspace.json
stefanhar: No, we are building the file since it uses the NestJS root app module to generate the schema and when not build it does not have access to the swagger plugin since that is supplied via webpack config applied in build step.
There might be better ways around this such as running `nest-cli` to do this generation, we did not look into this further.
But this should probably not be using `ts-node`, will change to `node`.

jeremybarbet (approved):

libs/application/templates/party-letter/src/lib/PartyLetterApplicationTemplate.ts
```diff
@@ -68,6 +68,9 @@ const PartyLetterApplicationTemplate: ApplicationTemplate<
           name: 'In Review',
           progress: 0.75,
           lifecycle: DefaultStateLifeCycle,
+          onEntry: {
+            apiModuleAction: API_MODULE_ACTIONS.CreateEndorsementList,
```
siggakogk, line 72: 👌

siggakogk (approved):
LGTM

PR #3575, opened 2021-04-21
chore(crc): support markdown for all description translation strings

VilliGunn:
https://app.asana.com/0/1199733748699731/1200194415618675


libs/application/templates/children-residence-change/src/fields/ContactInfo/index.tsx
```diff
@@ -43,9 +44,9 @@ const ContactInfo = ({ errors, application }: CRCFieldBaseProps) => {
 
   return (
     <>
-      <Text marginTop={3}>
-        {formatMessage(contactInfo.general.description)}
-      </Text>
+      <Box marginTop={3}>
+        <DescriptionText text={contactInfo.general.description} />
```
Rikkiabb, line 48: Don't we need to tag this a markdown in the translation string definition?

libs/application/templates/children-residence-change/src/fields/Reason/index.tsx
```diff
@@ -12,7 +13,9 @@ const Reason = ({ application, field }: CRCFieldBaseProps) => {
 
   return (
     <Box>
-      <Text marginTop={3}>{formatMessage(reason.general.description)}</Text>
+      <Box marginTop={3}>
+        <DescriptionText text={reason.general.description} />
```
Rikkiabb, line 17: Same as above

Rikkiabb (approved):

PR #3578, opened 2021-04-23
chore(application-system): add contentful to env

jeremybarbet:
## What

- Specify needed token for contentful for the application-system-api


baering (approved):
LGTM, do we need to add documentation for this new env variable?

jeremybarbet:
I will add documentation on another PR 👍

PR #3582, opened 2021-04-23
fix(icelandic-names-registry): default to string for auth values

stjanilofts:
again, hopefully a final fix
simply return empty string as fallback
https://github.com/island-is/island.is/runs/2419387239#step:10:1289


barabrian (approved):
lgtm

PR #3595, opened 2021-04-26
fix(parental-leave): simple valid email util

jeremybarbet:
## What

- fix email regex for parental leave answer validators


baering (approved):
🚀

libs/shared/constants/src/lib/email-regex.ts
```diff
@@ -0,0 +1 @@
+export const emailRegex = /^[\w!#$%&'*+/=?`{|}~^-]+(?:\.[\w!#$%&'*+/=?`{|}~^-]+)*@(?:[A-Z0-9-]+\.)+[A-Z]{2,6}$/i
```
baering, line 1: Personally I don't like regexes that take a lot of mental capacity, I do prefer a util function that explains the thought process in a simpler way:

```typescript
const isValidEmail = (value: string) => {
  if (!value.includes('@') {
    return false
  }

  const parts = value.split('@')

  if (parts.length !== 2) {
    return false
  }

  const [id, domain] = parts

  const domainParts = domain.split('.')

  if (domainParts.length !== 2) {
    return false
  }

  const [hostname, topLevelDomain] = domainParts

  if (id.length === 0 || hostname.length === 0 || topLevelDomain.length < 2) {
    return false
  }

  // TODO: then verify only legal characters - this could maybe still be a regex?

  return true
}
```

VilliGunn (approved):
LGTM 🚀

libs/shared/constants/src/lib/email-regex.ts
baering: Or `/\S+@\S+\.\S+/` 😆

PR #3598, opened 2021-04-26
fix(icelandic-names-registry): Add migration assets to icelandic-names-registry

sindrig:


sindrig:
Closing as duplicate of #3603

PR #3627, opened 2021-04-27
fix(island-ui/core): CategoryCard Tags can't be disabled

maranomynet:
## What

Add missing `disabled` prop to `CategoryCardProps['Tags'][number]` type

The `<Tags/>` component supports `disabled` but `<CategoryCard/>` omits that as an option for its tags.

## Why

We need this for our use of the `CategoryCard` component and this omission feels unintentional.

## Checklist:


barabrian (approved):
LGTM!

PR #3640, opened 2021-04-27
chore(web): html markup validity

stjanilofts:
A number of mostly minor changes to fix the html markup validity. Fixes most of the errors/warnings listed here https://validator.w3.org/nu/?doc=https%3A%2F%2Fisland.is%2F

Adds "As" prop from Reakit props to Button which removes "type" from the button when used. Can still be used with text button variant. This is so we can use Button inside anchor elements.

## Checklist:

### DIFF
--- a/apps/web/components/FrontpageSlider/FrontpageSlider.treat.ts
+++ b/apps/web/components/FrontpageSlider/FrontpageSlider.treat.ts
@@ -62,6 +62,7 @@ export const textItemVisible = style({
 })
 
 export const tabBullet = style({
+  display: 'inline-block',
   width: 8,
   height: 8,
   backgroundColor: theme.color.red200,
--- a/apps/web/screens/Home.tsx
+++ b/apps/web/screens/Home.tsx
@@ -32,7 +32,6 @@ import { GlobalContext } from '@island.is/web/context'
 import { QueryGetNewsArgs } from '@island.is/api/schema'
 import { LinkType, useLinkResolver } from '../hooks/useLinkResolver'
 import { FRONTPAGE_NEWS_TAG_ID } from '@island.is/web/constants'
-import { useRouter } from 'next/router'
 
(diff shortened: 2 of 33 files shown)


oddsson (approved):

apps/web/components/FrontpageSlider/FrontpageSlider.treat.ts
```diff
@@ -62,6 +62,7 @@ export const textItemVisible = style({
 })
 
 export const tabBullet = style({
+  display: 'inline-block',
```
kksteini, line 65: Was this implicit before?
Does this rule change behaviour in any way?

stjanilofts: The element using the class was a div but changed to span and needed to be block/inline-block

apps/web/screens/Home.tsx
```diff
@@ -42,6 +43,7 @@ interface HomeProps {
 const Home: Screen<HomeProps> = ({ categories, news, page }) => {
   const namespace = JSON.parse(page.namespace.fields)
   const { activeLocale, t } = useI18n()
+  const Router = useRouter()
```
kksteini, line 46: Do you need to import and use the Router here?

stjanilofts: Ah, not used. I will remove.

apps/web/components/FrontpageSlider/FrontpageSlider.treat.ts
kksteini: :+1:

apps/web/screens/Home.tsx
kksteini: :+1:

kksteini (approved):
LGTM

PR #3649, opened 2021-04-28
fix: set readyForProduction flag on collaboration application

Thorrik58:
Set ready for application flag as true


pshomov (approved):
LGTM

PR #3652, opened 2021-04-28
fix(air-discount-scheme): Fix specific customer record

kksteini:
# \<Description\>

## What

An airline company inserted wrong departure dates on a specific entry.

## Why

An airline company, using Loftbru, had a bug which they have now fixed. The entry within this migration is being amended so that the customer gets their connecting flight codes


dabbeg (approved):
LGTM

PR #3655, opened 2021-04-28
fix(application-system): use env variables when sending emails

VilliGunn:
# \<Description\>

Currently emails are not sending because `development@island.is` was hardcoded as a sender.
Fixing this issue for applications that are already in production.

We have a better solution for this that we will implement and release later.

## Why

- Emails are failing on production


Rikkiabb (approved):

baering (approved):

pshomov (approved):
LGTM

PR #3656, opened 2021-04-28
feat(island-ui): Add secondary CTA option to Action Card

Ingimarsson:
# Add secondary CTA option to Action Card

## What

This adds the option for a secondary CTA on Action Cards.

## Why

We need something like this for a list of applications that have both a digital version (primary) and a PDF (secondary).

Designers think this should be a part of the core library.

## Screenshots / Gifs

image

image

## Checklist:

### DIFF
--- a/libs/island-ui/core/src/lib/ActionCard/ActionCard.tsx
+++ b/libs/island-ui/core/src/lib/ActionCard/ActionCard.tsx
@@ -13,7 +13,7 @@ import { Icon } from '../IconRC/Icon'
 
 type ActionCardProps = {
   date?: string
-  heading: string
+  heading?: string
   text?: string
   tag?: {
     label: string
@@ -24,6 +24,11 @@ type ActionCardProps = {
     label: string
     variant?: ButtonTypes['variant']
     size?: ButtonSizes
+    icon?: 'arrowForward'
+    onClick?: () => void
+  }
+  secondaryCta?: {
+    label: string
     icon?: 'arrowForward'
     onClick?: () => void
   }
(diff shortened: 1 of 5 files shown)


libs/island-ui/core/src/lib/ActionCard/ActionCard.tsx
```diff
@@ -130,16 +136,37 @@ export const ActionCard: React.FC<ActionCardProps> = ({
...
+            flexDirection="row"
+          >
+            {hasSecondaryCTA && (
+              <Box paddingRight={4} paddingLeft={2}>
+                <Button
+                  variant="text"
+                  onClick={secondaryCta?.onClick}
+                  icon={'document'}
+                >
+                  {secondaryCta?.label}
```
oddsson, line 161: I think you should check if secondaryCta.label is present in line 154, i.e. `hasSecondaryCTA && secondaryCta.label &&...` because right now you could be rendering a button with no text.

oddsson (changes requested):
Looks good! One small comment 🙂

libs/island-ui/core/src/lib/ActionCard/ActionCard.tsx
Ingimarsson: `hasSecondaryCTA` is created like this `const hasSecondaryCTA = hasCTA && secondaryCta?.label && !progressMeter.active` in line 139 so it is not rendered if label is empty. I tried setting `label: ''` and it works as expected :)

oddsson (approved):

PR #3657, opened 2021-04-28
fix(as): Reads email from env values

Rikkiabb:
# \<Description\>
Reads email sender name and email address from environment variables instead of hard coding them.

## Why
There is a different value white listed for sending email's on production then on dev and staging. This means that the current implementation with hard coding the email address would always be blocked on production.

## Checklist:


baering (approved):
LGTM, thanks!

VilliGunn (approved):
:shipit:

PR #3658, opened 2021-04-28
build(endorsement-system-api): Renaming the service and adding docker image

pshomov:
# \<Description\>

Supports island-is/infrastructure#664

## What

Renaming endorsement-system => endorsement-system-api to match a bit better the naming scheme

## Why

Consistency


CHANGELOG.md
```diff
@@ -0,0 +1,891 @@
+# Changelog
+
+All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
+
```
eirikurn, line 4: Was this accidental?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "island.is",
-  "version": "0.0.1",
+  "version": "0.0.2",
```
eirikurn, line 3: Was this accidental?

eirikurn (approved):

eirikurn:
Since we're renaming this for consistency. Should this be `apps/services/endorsements` instead?

The top-level app folders are meant for larger/standalone projects, or when a project has both frontend and backend and we want to group the two. If this is an internal service that's only exposed through other public projects (like graphql API) I think it should be under services.

TBF, this should be documented somewhere but it's probably not ...

stefanhar (approved):
Renaming LGTM

stefanhar (changes requested):
On closer inspection we are missing changes to paths inside some files, I'll add those

pshomov:
> Since we're renaming this for consistency. Should this be `apps/services/endorsements` instead?
> 
> 
> 
> The top-level app folders are meant for larger/standalone projects, or when a project has both frontend and backend and we want to group the two. If this is an internal service that's only exposed through other public projects (like graphql API) I think it should be under services.
> 
> 
> 
> TBF, this should be documented somewhere but it's probably not ...

What do say @stefanhar - move the service as @eirikurn is suggesting?

stefanhar:
> > Since we're renaming this for consistency. Should this be `apps/services/endorsements` instead?
> > The top-level app folders are meant for larger/standalone projects, or when a project has both frontend and backend and we want to group the two. If this is an internal service that's only exposed through other public projects (like graphql API) I think it should be under services.
> > TBF, this should be documented somewhere but it's probably not ...
> 
> What do say @stefanhar - move the service as @eirikurn is suggesting?

Fine by me

package.json
pshomov: yes, sorry will clean up. I think the Changelog experiments have committed locally something.

stefanhar (approved):

PR #3659, opened 2021-04-29
feat(judicial-system): Case file not found

oddsson:
# Handle when users try to open a file that does not exist

https://app.asana.com/0/322488613152836/1200181907151990

## What

Display a modal that tells the user that a file is no longer available if the file is no longer in S3 for any reason. 

## Why

Better UX

## Screenshots / Gifs

https://user-images.githubusercontent.com/3789875/116530872-0c6afe80-a8ce-11eb-8fef-ea8096b47122.mov

## Checklist:


gudjong (commented):
Brilliant. Will make a simple change to server side code, that I wrote and then apporve.

gudjong (approved):

PR #3662, opened 2021-04-29
feat(judicial-system): Send a copy of prison email to prosecutors

oddsson:
# Send a copy of prison email to prosecutors

https://app.asana.com/0/1199153462262248/1200089982514663

## What

Send a copy of prison email to prosecutors

## Why

This is something that prosecutors are asking a lot for.

## Checklist:

### DIFF
--- a/apps/judicial-system/backend/src/app/modules/notification/notification.service.ts
+++ b/apps/judicial-system/backend/src/app/modules/notification/notification.service.ts
@@ -413,8 +413,8 @@ export class NotificationService {
     )
 
     await this.sendEmail(
-      'Gæsluvarðhaldsfangelsi',
-      existingCase.prosecutor.email,
+      existingCase.prosecutor?.name,
+      existingCase.prosecutor?.email,
       subject,
       html,
     )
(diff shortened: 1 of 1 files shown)


apps/judicial-system/backend/src/app/modules/notification/notification.service.ts
```diff
@@ -412,6 +412,13 @@ export class NotificationService {
       existingCase.additionToConclusion,
     )
 
+    await this.sendEmail(
+      'Gæsluvarðhaldsfangelsi',
+      existingCase.prosecutor.email,
+      subject,
+      html,
+    )
```
gudjong, line 420: The first argument to sendEmail is the recipient's name, i.e. in this case it should be the name of the prosecutor. The '?' is to protect agains null pointer errors. By the time this email is sent, the prosecutor should have been set, but we have sometimes seen null pointers in the past. I would rather get an error from the email server than getting an unhandled null pointer.

```suggestion
    await this.sendEmail(
      existingCase.prosecutor?.name,
      existingCase.prosecutor?.email,
      subject,
      html,
    )
```

gudjong (changes requested):

PR #3674, opened 2021-04-29
chore(web): Web performance tweaks

stjanilofts:
- Only load frontpage slider json/svg animation on desktop
- Add lazy load to BackgroundImage and add option for showing blurred out thumbnail while loading image

## Checklist:


apps/web/components/BackgroundImage/BackgroundImage.tsx
```diff
@@ -59,7 +65,22 @@ export const BackgroundImage = ({
       }
     : {}
 
-  const imageLoaded = useImageLoader(src)
+  useLayoutEffect(() => {
+    if (backgroundImageRef?.current && 'IntersectionObserver' in window) {
+      const lazyBackgroundObserver = new IntersectionObserver((entries) => {
```
arnorhs, line 70: what is the minimum browser requirement? is this already polyfilled?

apps/web/components/BackgroundImage/BackgroundImage.tsx
```diff
@@ -14,21 +14,24 @@ export type BackgroundImageProps = {
...
 
   useEffect(() => {
-    const img = new window.Image(100)
-    img.onload = img.onerror = () => {
-      if (isMounted) {
-        setLoaded(true)
+    if (shouldLoad) {
+      const img = new window.Image(100)
+      img.onload = img.onerror = () => {
+        if (isMounted) {
```
arnorhs, line 28: Not really related to this change in particular, but..

I'm not sure this mounted check would work and/or be accurate if this fires post-mounting.. since the scope of this function will still be seeing the older version of isMounted.. it would be much better if you'd return a function to simply unsubscribe from this event (unassign the error handlers)

apps/web/components/BackgroundImage/BackgroundImage.tsx
```diff
@@ -59,7 +65,22 @@ export const BackgroundImage = ({
...
+      const lazyBackgroundObserver = new IntersectionObserver((entries) => {
+        const entry = entries[0]
+
+        if (entry.isIntersecting) {
+          setShouldLoad(true)
+          lazyBackgroundObserver.unobserve(entry.target)
+        }
+      })
+
+      lazyBackgroundObserver.observe(backgroundImageRef.current)
```
arnorhs, line 79: don't you need to unobserve this on unmounting?

apps/web/components/BackgroundImage/BackgroundImage.tsx
stjanilofts: I found a similar tool already in the package.json (react-use, https://github.com/streamich/react-use/blob/master/docs/useIntersection.md).

kksteini (approved):
LGTM

arnorhs (approved):
múchó betteró

PR #3680, opened 2021-04-30
feat(judicial-system): Create Court Case

gudjong:
# Create Court Case

https://app.asana.com/0/1199153462262248/1200011538011589

## What

☝️ 

## Why

We want to be able to create court cases for custody, extended custody, travel ban and extended travel ban.

## Checklist:


apps/judicial-system/api/src/app/modules/court/court.service.ts
```diff
@@ -1,10 +1,14 @@
+import { formatISO } from 'date-fns' // eslint-disable-line no-restricted-imports
```
oddsson, line 1: Can you write `import formatISO from 'date-fns/formatISO'` and remove the eslint thing?

apps/judicial-system/api/src/app/modules/court/court.spec.ts
```diff
@@ -0,0 +1,200 @@
+import { uuid } from 'uuidv4'
+import { formatISO } from 'date-fns' // eslint-disable-line no-restricted-imports
```
oddsson, line 2: Same as ☝️

apps/judicial-system/api/src/app/modules/court/court.spec.ts
```diff
@@ -0,0 +1,200 @@
...
+import { BackendAPI } from '../../../services'
+import { environment } from '../../../environments'
+import { CourtResolver } from './court.resolver'
+import { CourtService } from './court.service'
+
+function expectReceivalDate(receivalDate: string) {
+  const date = new Date()
+  const today = formatISO(date, { representation: 'date' })
+  date.setDate(date.getDate() - 1)
+  const yesterday = formatISO(new Date(), { representation: 'date' })
```
oddsson, line 24: Isn't the `new Date()` here supposed to be `date`? Otherwise, I think this is a confusing variable name in that the value is actually today 🤔

oddsson (changes requested):

apps/judicial-system/api/src/app/modules/court/court.service.ts
gudjong: No, that results in failing tests and I don't know a way to fix that.

apps/judicial-system/api/src/app/modules/court/court.spec.ts
gudjong: Same answer

apps/judicial-system/api/src/app/modules/court/court.spec.ts
gudjong: You are right. Will fix.

gudjong: Well spotted.

oddsson (approved):

PR #3688, opened 2021-04-30
fix(judicial-system): Removes logging of email addresses and mobile numbers on error

gudjong:
# Removes logging of email addresses and mobile numbers on error

https://app.asana.com/0/1199153462262248/1200214011389030

## What

☝️ 

## Why

## Checklist:


oddsson (changes requested):
There are a few more that I think we need to remove

128: auth.controller.ts

75: case.service.ts

97: notification.service.ts
146: notification.service.ts

50: user.service.ts
80: user.service.ts

gudjong:
Other changes may be need, but I was just addressing specific recommendations from Syndis. I will take a look.

gudjong:
What I did:

128: auth.controller.ts - removed the nationalId

75: case.service.ts - removed the email address

97: notification.service.ts - this had already been address in the original PR
146: notification.service.ts - this had already been address in the original PR

50: user.service.ts - removed the national id
80: user.service.ts - removed the national id

oddsson (approved):

PR #3707, opened 2021-05-03
feat(crc): child support screen

VilliGunn:
# \<Description\>

https://app.asana.com/0/1199733748699731/1200266567269680

## What

- Added a child support screen
- Remove child support from terms screen

### Note
- This is not exactly like in figma because we do not yet support the bulletin styles for headings and text.
- I've pushed translations to Contentful
- Styles of links in the `DescriptionText` are broken and will be fixed in another task (https://app.asana.com/0/1199733748699731/1200259148760500)

## Why

- To help the user understand what effect this contract/application will have on child support

## Screenshots / Gifs

Screenshot 2021-05-03 at 18 35 48

## Checklist:


libs/application/templates/children-residence-change/src/lib/messages/childSupport.ts
```diff
@@ -0,0 +1,31 @@
...
+      description: 'Child support page description',
+    },
+  }),
+  childBenefitCheckbox: defineMessages({
+    label: {
+      id: 'crc.application:section.effect.childSupport.label',
+      defaultMessage:
+        'Samkomulag er um að meðlagsgreiðslur flytjast með barninu til nýs lögheimilisforeldris',
```
Rikkiabb, line 27: Should be fixed in contentful but shouldn't this be `Ég samþykki að....` instead og `Samkomulag er um að....`?

Rikkiabb (approved):
🚢

libs/application/templates/children-residence-change/src/lib/messages/childSupport.ts
VilliGunn: This is what was in Figma so I went with that 🤷

PR #3716, opened 2021-05-04
chore(contentful): Added the source code for the already hosted extensions on Contentful

kksteini:
# \<Description\>

I added the parent-slug-extension and the mideind-extension for contentful.
I created a new folder to collect Contentful extensions and moved the already included `contentful-translations-extension` in there.
These extensions were developed for SÍ and therefore they'll live in the monorepo.
They are currently hosted/served by Contentful, not by us.

## Checklist:


libs/contentful-extensions/translation/src/index.html
```diff
@@ -0,0 +1,10 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8" />
+  </head>
+  <body>
+    <div id="root"></div>
+    <script type="text/javascript" src="./index.tsx"></script>
+  </body>
+</html>
```
kksteini, line 10: This file is exactly as it was in the old `contentful-translation-extension/src`. Not sure why github is not picking up on it being simply moved

workspace.json
```diff
@@ -2862,31 +2862,39 @@
         }
       }
     },
```
kksteini, line 2864: I added the bare minimum since I am only adding these extensions as they are written. They are served from within Contentful itself but this code is owned by SÍ.

Please let me know if this is inadvisable.

kksteini:
Deleted a bunch of "please run `yarn gitbook` messages as they were cluttering up the history". I have resolved it.

SUMMARY.md
```diff
@@ -152,7 +152,9 @@
 - [Content Search Indexer](libs/content-search-indexer/README.md)
 - [Content Search Metrics](libs/content-search-metrics/README.md)
 - [Content Search Toolkit](libs/content-search-toolkit/README.md)
-- [Contentful Translation Extension](libs/contentful-translation-extension/README.md)
+  - [Contentful Extension for Miðeind Machine Translations](libs/contentful-extensions/mideind-translation/README.md)
```
jeremybarbet, line 155: You need a README at the root of the contentful-extensions folder to get the correct gitbook hierarchy https://docs.devland.is/repository/gitbook-contribute#root-readme-md-within-handbook-directory

libs/contentful-extensions/mideind-translation/package.json
```diff
@@ -0,0 +1,48 @@
...
+    "contentful-cli": "1.4.48",
+    "eslint": "7.12.1",
+    "typescript": "4.0.5",
+    "@types/jest": "26.0.15",
+    "@types/react": "16.9.56",
+    "@types/react-dom": "16.9.9",
+    "@types/webpack-env": "1.15.3"
+  },
+  "scripts": {
+    "start": "contentful-extension-scripts start --output build/ --id mideind-translation --name mideind-translation --space-id 8k0h54kbe6bj --field-types Object",
```
jeremybarbet, line 23: Are the two packages.json required? You could move these scripts inside the workspace.json so you could start/build and so on from the root

SUMMARY.md
kksteini: Thank you, I missed that :+1:

libs/contentful-extensions/mideind-translation/package.json
kksteini: Hmm.

The start and build scripts could be moved into the workspace. The remaining issue would be the deployment script.
The deploy script which we need in order to update the extension hosted on Contentful.
The deploy script relies on the `contentful-cli`.
If we were to move the deployment into the workspace then we would need to add this package or fork it and do some minimalization work.

I tested adding the `contentful-cli` package to our repo and that results in 19 new packages.

Deployments are going to be so infrequent that I'd consider keeping the package.json independent in this case, since these extensions, in particular, are not being deployed from within our system.

Is there perhaps something I am not seeing?

jeremybarbet: Good point, better to keep it local to the extensions 👍

jeremybarbet (approved):
LGTM

PR #3724, opened 2021-05-05
fix(web): Show related links even if there are no related articles

Ingimarsson:
# Show related links even if there are no related articles

## What

The related content section is currently hidden if there are no related articles, but the section contains both related articles and related links (external links). Some articles only have related links, and then the section should not be hidden.


apps/web/screens/Article.tsx
```diff
@@ -263,13 +263,11 @@ const ArticleSidebar: FC<ArticleSidebarProps> = ({
...
         <ArticleNavigation article={article} activeSlug={activeSlug} n={n} />
       )}
-      {article.relatedArticles.length > 0 && (
-        <RelatedContent
-          title={n('relatedMaterial')}
-          articles={article.relatedArticles}
-          otherContent={article.relatedContent}
-        />
-      )}
+      <RelatedContent
```
stjanilofts, line 266: Maybe we should check for `relatedContent.length` or `relatedArticle.length`? Just wondering what happens if there is neither?

stjanilofts (approved):

apps/web/screens/Article.tsx
Ingimarsson: it is done inside `RelatedContent` so I  thought it would be redundant to have it here too :)

PR #3742, opened 2021-05-06
feat('auth-public-api): Delegations

unnarb:
#Delegations

CRUD of Delegations and Delegation Scopes

## What

## Why

## Screenshots / Gifs

## Checklist:


apps/services/auth-public-api/src/app/modules/delegations/delegations.controller.ts
```diff
@@ -0,0 +1,87 @@
...
+@Controller('delegations')
+export class DelegationsController {
+  constructor(private readonly delegationsService: DelegationsService) {}
+
+  @Scopes('@island.is/auth/delegations:read')
+  @Get()
+  @ApiOkResponse({ isArray: true })
+  async findAllTo(@CurrentUser() user: User): Promise<IDelegation[]> {
+    return await this.delegationsService.findAllTo(user.nationalId)
+  }
```
eirikurn, line 37: This is an endpoint that we want to be able to call with the currently "authenticated" user rather than the delegated user, i.e. `user.act?.nationalId ?? user.nationalId`.

We probably want to define an API naming convention for that (/me/delegations vs /delegations or /subject/delegations? Or just /users/:nationalId/delegations?) as well as nest helpers.

apps/services/auth-public-api/src/app/modules/delegations/delegations.controller.ts
```diff
@@ -0,0 +1,87 @@
...
+  Param,
+  Post,
+  Put,
+  UseGuards,
+} from '@nestjs/common'
+import { ApiCreatedResponse, ApiOkResponse, ApiTags } from '@nestjs/swagger'
+
+@UseGuards(IdsUserGuard, ScopesGuard)
+@ApiTags('delegations')
+@Controller('delegations')
```
eirikurn, line 28: It would be mighty nice if these endpoints created/updated/returned scopes as well, as these will generally be used together, and there might be a lot of scopes.

If you'd rather keep these separate, I'd at least like a scopeBulkUpdate endpoint that overrides the list of scopes for a delegation.

apps/services/auth-public-api/src/app/modules/delegation-scopes/delegation-scope.controller.ts
```diff
@@ -0,0 +1,56 @@
...
+  @Delete()
+  @ApiCreatedResponse()
+  async delete(
+    @Body() delegationScope: DelegationScopeDeleteDTO,
+  ): Promise<number> {
+    return await this.delegationsScopeService.delete(
+      delegationScope.delegationId,
+      delegationScope.scopeName,
+    )
+  }
```
eirikurn, line 55: These need to be protected based on the "fromNationalId" of the delegation.

Also, we should filter any scopes queried/created/deleted here using the list of scopes in the access token. We may take that logic further in the future, but I think it's a good start.

apps/services/auth-public-api/src/app/modules/delegation-scopes/delegation-scope.module.ts
```diff
@@ -0,0 +1,14 @@
...
+  DelegationScope,
+} from '@island.is/auth-api-lib'
+import { DelegationScopeController } from './delegation-scope.controller'
+
+@Module({
+  imports: [SequelizeModule.forFeature([DelegationScope])],
+  controllers: [DelegationScopeController],
+  providers: [DelegationScopeService],
+})
+export class DelegationScopeModule {}
```
eirikurn, line 14: Are you adding "scope metadata" endpoints elsewhere? I.e. to query the list of scopes, with translations, that the user can pick when creating custom delegations?

libs/auth-api-lib/src/lib/entities/dto/delegation.dto.ts
```diff
@@ -0,0 +1,33 @@
...
+  IsDateString,
+} from 'class-validator'
+// For use with deleting Inserting and updating Delegation
+export class DelegationDTO {
+  @IsString()
+  @ApiProperty()
+  fromNationalId!: string
+  @IsString()
+  @ApiProperty()
+  fromDisplayName!: string
```
eirikurn, line 16: Do we want the client to provide this? I don't think we have a plan for it now. The plan is to generally show the person's real name from Þjóðskrá. Not sure how/where we query/cache that though. Maybe the graphql layer.

libs/auth-api-lib/src/lib/entities/dto/delegation.dto.ts
```diff
@@ -0,0 +1,33 @@
...
+  @IsBoolean()
+  @ApiProperty()
+  isFromCompany!: boolean
```
eirikurn, line 19: What is this for?

libs/auth-api-lib/src/lib/entities/dto/delegation.dto.ts
```diff
@@ -0,0 +1,33 @@
...
+  @IsString()
+  @ApiProperty()
+  toNationalId!: string
```
eirikurn, line 22: This should always be the CurrentUser. Otherwise, I can create a delegation for anyone/anything.

libs/auth-api-lib/src/lib/services/delegations.service.ts
```diff
@@ -92,6 +97,65 @@ export class DelegationsService {
         },
     )
   }
+
+  async create(delegation: DelegationDTO): Promise<Delegation | null> {
+    this.logger.debug('Creating a new delegation')
+    const id = uuid()
+    return await this.delegationModel.create({ id: id, ...delegation })
+  }
```
eirikurn, line 105: Should we validate the toNationalId further? In natreg?

Or filter that it's not a company since we do not have a systems to handle delegations where toNationalId is a company.

eirikurn (changes requested):

libs/auth-api-lib/src/lib/services/delegations.service.ts
eirikurn: It might be nice if the IDS included a claim indicating if the current user is a company. So we're not doing "kennitala" checks all over the place.

apps/services/auth-public-api/src/app/modules/delegations/delegations.controller.ts
eirikurn: Fyi, the design has a screen that lists "all" delegations, grouped by toNationalId, sorted by person's name, and listing which scopes each has.

This will be tricky to paginate without having a copy of user's names. But we should try not to do N+1 requests to show what each delegated user has access to.

apps/services/auth-public-api/src/app/modules/delegations/delegations.controller.ts
eirikurn: cc @valurefugl

libs/auth-api-lib/src/lib/services/delegations.service.ts
valurefugl: https://github.com/island-is/identity-server.web/issues/274.

libs/auth-api-lib/src/lib/entities/dto/delegation.dto.ts
valurefugl: Good point, not sure what i was thinking when i added this to our table. We should drop it.

libs/auth-api-lib/src/lib/services/delegations.service.ts
```diff
@@ -92,6 +97,65 @@ export class DelegationsService {
...
+    const now = new Date()
+
+    this.logger.debug(`Finding a delegation for nationalId ${nationalId}`)
+    return await this.delegationModel.findAll({
+      where: {
+        [Op.and]: [
+          { toNationalId: nationalId },
+          { validFrom: { [Op.lt]: now } },
+          { validTo: { [Op.or]: [{ [Op.eq]: null }, { [Op.gt]: now }] } },
+          { validCount: { [Op.or]: [{ [Op.eq]: null }, { [Op.gt]: 0 }] } },
```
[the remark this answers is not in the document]
valurefugl: The idea was that we could support like a "one time" or "several times" style delegations, this had come up in discussions. But think we should drop it until there is a clear requirement.

apps/services/auth-public-api/src/app/modules/delegations/delegations.controller.ts
valurefugl: async findAllTo(@CurrentActor() user: User) ?

valurefugl: https://github.com/island-is/identity-server.web/issues/275

libs/auth-api-lib/src/lib/entities/dto/delegation.dto.ts
valurefugl: fromNationalId should always be the CurrentUser, right?

libs/auth-api-lib/src/lib/entities/dto/delegation.dto.ts
valurefugl: Was thinking the client would provide this (to save calls to Þjóðskrá). Could default to the real name when registering, with the option of overriding with any alias (for display in dropdown and on registration page, not used in tokens)?
But your call.

apps/services/auth-public-api/src/app/modules/delegation-scopes/delegation-scope.controller.ts
unnarb: I think we will delete this controller

libs/auth-api-lib/src/lib/services/delegations.service.ts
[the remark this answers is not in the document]
eirikurn: ahh, interesting!

apps/services/auth-public-api/src/app/modules/delegations/delegations.controller.ts
eirikurn: Yeah, CurrentActor LGTM.

apps/services/auth-public-api/src/app/modules/delegation-scopes/delegation-scope.controller.ts
eirikurn: Sounds good

libs/auth-api-lib/src/lib/entities/dto/delegation.dto.ts
eirikurn: Ok, makes sense. Let's keep this around for now. 👍

libs/auth-api-lib/src/lib/entities/dto/delegation.dto.ts
eirikurn: Arrg 🤯😂

So the delegation goes "from" a company "to" a user? Then I log in as that user, and query all delegations going "to" that user, to show in the delegation dropdown.

Makes sense. And then "fromNationalId should always be the CurrentUser, but passing in "fromDisplayName" is a bit weird. "I'm giving you access to me, but I want me to have a custom display name in your dropdown". I imagined the use case the other way around, me adding an alias for the list of people that have access to me, in findAllFrom.

If we want it for caching reasons, maybe it's better to fetch it from the ID token when creating the delegation. But as I said elsewhere, we'll need to solve this also somehow when querying findAllFrom.

libs/auth-nest-tools/src/lib/current-actor.decorator.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  const user = request.user
+
+  if (!user) {
+    logger.warn(
+      'No user authentication found. Did you forget to add IdsUserGuard?',
+    )
+    throw new UnauthorizedException()
+  }
+  return user.actor ?? user
```
eirikurn, line 22: Should maybe add some of the `user` properties to the `user.actor`, so it fulfils the interface and can be used in the audit service and auth middleware.

```
return !user.actor ? user : {
  ...user.actor,
  client: user.client,
  authorization: user.authorization,
  ip: user.ip,
  userAgent: user.userAgent,
}
```

libs/auth-api-lib/src/lib/services/delegations.service.ts
```diff
@@ -91,78 +90,119 @@ export class DelegationsService {
...
     id: string,
   ): Promise<Delegation | null> {
     this.logger.debug(`Updating a delegation with id ${id}`)
-    await this.delegationModel.update({ ...delegation }, { where: { id: id } })
-    return this.findByPk(id)
+    await this.delegationModel.update(
+      { ...delegation },
+      { where: { id: id, fromNationalId: nationalId } },
+    )
+    await this.delegationScopeService.delete(id)
```
eirikurn, line 126: Should only delete scopes if a new scope array is passed?

libs/auth-nest-tools/src/lib/current-actor.decorator.ts
unnarb: cc @valurefugl

libs/auth-api-lib/src/lib/services/delegations.service.ts
unnarb: Depends on how the client would behave. The simplest way would if those scope which are selected are sent into the request. Then we delete the existing ones and add in all those which are selected

eirikurn: Right. I guess I was expecting PATCH behaviour rather than PUT behaviour. That I could pass scopes: [] to remove all scopes, and skip the scopes parameter to not modify it.

But since this is PUT, I'm happy to keep it like it is.

libs/auth-nest-tools/src/lib/current-actor.decorator.ts
valurefugl: Good point.

PR #3755, opened 2021-05-06
fix(icelandic-names-registry/web): finalizing changes/fixes

stjanilofts:
# \<Description\>

A few changes to mannanafnaskrá: https://app.asana.com/0/1199123945323117/1200247320558955

## What

1. Added links to the verdict text if url available
2. Fixed bug with deleting names from DB
3. Added the gender neutral name type
4. Make the names searcher more usable in mobile
5. Remove the hardcoded /mannanafnaskra page and create a normal article instead that gets the name searcher component appended to the content

## Why

1. Url field wasn't being used at all, now it can link to the verdict page
2. Deleting names didn't work
3. Needed the gender neutral name type
4. Filter section was below results in table making it pretty useless/hard to use in mobile
5. The current page was made before realizing a normal article page would do. Added a check for article id wether to display the component or not.

## Checklist:


apps/web/components/AppendedArticleComponents/AppendedArticleComponents.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+type Article = GetSingleArticleQuery['getSingleArticle']
+
+interface Props {
+  article: Article
+}
+
+export const AppendedArticleComponents = ({ article }: Props) => {
+  switch (article?.id) {
+    // Leit í mannanafnaskrá
```
kksteini, line 20: Keep comments in English
https://docs.devland.is/technical-overview/code-standards#language

apps/web/components/IcelandicNamesSearcher/IcelandicNamesSearcher.tsx
```diff
@@ -205,228 +215,264 @@ export const IcelandicNamesSearcher: FC = () => {
...
+                >
+                  Hreinsa val
+                </ResetButton>
+              )}
+            </Stack>
+          </SidebarAccordion>
+        </Box>
+        {!!someFilterSelected && (
+          <ResetButton onClick={() => dispatch({ type: 'clearAll' })}>
+            Hreinsa síu
```
kksteini, line 453: A little above you had the text `Hreinsa síu` but changed it to `Hreinsa val`.
Is this inconsistent?

kksteini (approved):
LGTM

apps/web/components/IcelandicNamesSearcher/IcelandicNamesSearcher.tsx
stjanilofts: Good point.

libs/service-portal/icelandic-names-registry/src/components/Editor/Editor.tsx
```diff
@@ -150,7 +150,10 @@ const Editor: FC = () => {
       }
 
       try {
-        await deleteName({ variables: { input: { id: body.id } } })
+        await deleteName({
+          variables: { input: { id: body.id } },
+        })
```
darrikonn, line 155: Would like to see a gif of this, in terms of seeing the update.

There are some implications in this component, most likely due to the fact that the AST is split from the component so it's hard to know what is going on.

The lazy query doesn't rely on the cache, so a full round-trip is needed every time a mutation occurs. Since I don't have a gif I don't know if the query is slow and if it needs to be speed up. By querying the correct fields after mutating an object, the Apollo cache should handle updating its store based on the unique ID that should be fetched after a mutation (e.g. the update name mutation).
Where the store doesn't already have the ID, the store could be updated based on the objects returned from the mutation (e.g. create name mutation). Similar approach can be used for the delete name mutation, along with optimistically updating the store if wanted for better UX.
This would both reduce the round-trips needed for the search query, as well as being able to use the cache.

libs/service-portal/icelandic-names-registry/src/queries/deleteIcelandicName.ts
```diff
@@ -3,8 +3,7 @@ import gql from 'graphql-tag'
 export const DELETE_ICELANDIC_NAME_MUTATION = gql`
   mutation DeleteIcelandicNameMutation($input: DeleteIcelandicNameByIdInput!) {
     deleteIcelandicNameById(input: $input) {
-      id
-      icelandicName
+      success
```
darrikonn, line 6: This is a mutation inside the queries directory.

darrikonn (changes requested):

libs/service-portal/icelandic-names-registry/src/components/Editor/Editor.tsx
stjanilofts: I made some changes to it, I think it should be better now. Now it updates the cache instead of making a full round-trip on mutations.

image

apps/web/components/IcelandicNamesSearcher/IcelandicNamesSearcher.tsx
```diff
@@ -255,7 +255,7 @@ export const IcelandicNamesSearcher = () => {
   const tableSection = (
     <GridColumn
       span={['12/12', '12/12', '12/12', '12/12', '3/5']}
-      paddingTop={shouldStack ? 2 : 0}
+      paddingTop={[2, 2, 2, 2, 0]}
```
darrikonn, line 258: Forgot to delete the `useWindowSize` hook ... the point was to completely get rid of it :).
Thus you can delete the `shouldStack` variable as well.

libs/service-portal/icelandic-names-registry/src/components/Editor/Editor.tsx
darrikonn: Nice, good job!

darrikonn (commented):
One small changes as well. I saw that the IcelandicName models doesn't have a specified ID column:
```typescript
@Field(() => ID)
```
This is crucial to let the apollo store update its cache accordingly. I'm not sure how `nestjs/graphql` works in regards if the name of the column is `id`, then it automatically sets this as ID type (if that is the case, then you can ignore this comment 👍). However, if the `id` is `kennitala`, then it should most likely be renamed to `nationalId`, since that is the common name for all the other models.

Otherwise, looks good :+1:

darrikonn (approved):

PR #3758, opened 2021-05-07
fix(party-application): Adding "In review" state

siggakogk:
# Adding "In review" state to Party Application

## What

Note: the Party Application is still a very much a WIP 
After endorsements have been submitted the supreme court will get an email notification with a link to the "In Review" state where they can reject or approve the application.

* Added a new state to application
* Added a new form to application
* (wip) Added assignment to supreme court

## Checklist:


libs/application/templates/party-application/src/assets/CoatOfArms.tsx
```diff
@@ -0,0 +1,1152 @@
+import React from 'react'
+
+const CoatOfArms = () => (
+  <svg
```
kksteini, line 4: Can this be moved to Contentful as an asset?

libs/application/templates/party-application/src/fields/EndorsementList/index.tsx
```diff
@@ -1,75 +1,87 @@
...
+  const namesCountString = formatMessage(
+    endorsements && endorsements.length > 1
+      ? m.endorsementList.namesCount
+      : m.endorsementList.nameCount,
+  )
   return (
     <Box marginBottom={8}>
       <CopyLink
-        linkUrl="www.island.is/listabókstafur/128877634/"
+        linkUrl={window.location.href}
```
kksteini, line 79: Is this supposed to copy the current URL?

libs/application/templates/party-application/src/fields/PartyApplicationApproved/index.tsx
```diff
@@ -5,8 +5,9 @@ import { useLocale } from '@island.is/localization'
 import { Approved } from '@island.is/application/ui-components'
 import { m } from '../../lib/messages'
 
-const PartyApplicationApproved: FC<FieldBaseProps> = () => {
+const PartyApplicationApproved: FC<FieldBaseProps> = ({ application }) => {
   const { formatMessage } = useLocale()
+  console.log(application)
```
kksteini, line 10: Console log

libs/application/templates/party-application/src/fields/EndorsementList/index.tsx
siggakogk: yes

libs/application/templates/party-application/src/assets/CoatOfArms.tsx
siggakogk: no :( i'm 97.2% positive this is the current way to display logos in the application system.. someone correct me if i'm wrong

libs/application/templates/party-application/src/fields/PartyApplicationApproved/index.tsx
siggakogk: i remove yes yes yes

libs/application/templates/party-application/src/fields/EndorsementList/index.tsx
kksteini: :+1:

libs/application/templates/party-application/src/assets/CoatOfArms.tsx
stefanhar: Currently we get our Contentful strings from the `Translation Namespace`.
I believe we only support strings atm, we could get a string that represents a url to a svg image that way.
But I feel that approach might be fragile for several reasons, so currently I think this an acceptable approach especially since we don't expect this image to change often

PR #3759, opened 2021-05-07
fix(security): Security audit of lodash and underscore

sindrig:


sindrig:
Closing as duplicate of https://github.com/island-is/island.is/pull/3760

PR #3765, opened 2021-05-07
feat(core): react authentication library with security fixes

eirikurn:
## What

Refactored React SPA authentication code from service-portal and application-form into a shared library. Also addresses a few reported security issues.

* Handles oidc-client and callback routes.
* Handles authentication flow with a loading screen.
* Manages user context.
* New: Renews access tokens on demand (when calling APIs) instead of continuously. This helps us support an (eg) 1-8 hour IDS session, depending on how long the user is active.
* New: Preloads a new access token sometime before it expires (when calling APIs).
* New: Monitor the IDS session and redirect to the login screen if the user is not logged in anymore.

## Why

This is security-critical code was being copied in multiple projects. We need to make sure future fixes and improvements benefit all our (non-next) react SPA projects.

## Checklist:


libs/auth/react/src/lib/Authenticator/Authenticator.tsx
```diff
@@ -0,0 +1,158 @@
...
+        payload: user,
+      })
+    }
+
+    // This is raised when the user is signed out of the IDP.
+    const userSignedOut = () => {
+      dispatch({
+        type: ActionType.LOG_OUT,
+      })
+      if (autoLogin) {
```
atligudl, line 114: Calling signIn here to redirect the user back to the IDS after logout?

atligudl (approved):
LGTM

libs/auth/react/src/lib/Authenticator/Authenticator.tsx
eirikurn: Yeah. If the user initiates a logout themself, they are sent (by the IDS) to the `post_logout_redirect_uri` (usually island.is front page).

But if the IDS session expires (1 hour) or if the user logs out in another tab, we call signin here to redirect this tab to the IDS login screen. This means:

* any sensitive information is hidden from the screen.
* the user is aware that the session is over, and doesn't try to submit changes with an expired session.
* the user can finish logging in and end up back where they were.

We'll see how this works and "feels" in practice, e.g. with delegations.

This is similar to google btw, though they take things a step further, so if I have 10 inactive tabs, all showing the login screen, it's enough for me to login to one of them, and all will then go back to where they were. That might be a cool addition here.

apps/application-system/form/src/main.tsx
```diff
@@ -3,8 +3,10 @@ import React from 'react'
 import ReactDOM from 'react-dom'
 import * as Sentry from '@sentry/react'
 import { Integrations } from '@sentry/tracing'
-import { environment } from './environments'
+import { configure } from '@island.is/auth/react'
```
jeremybarbet, line 6: Not used?

apps/application-system/form/src/main.tsx
```diff
@@ -3,8 +3,10 @@ import React from 'react'
...
+import { ApplicationIdentityServerScope } from '@island.is/application/core'
```
jeremybarbet, line 7: Not used neither?

libs/auth/react/src/lib/Authenticator/AuthenticatorLoadingScreen.tsx
```diff
@@ -1,14 +1,16 @@
...
-const AuthenticatorLoadingScreen: FC<{}> = () => {
+const AuthenticatorLoadingScreen = () => {
   return (
     <Box
       display="flex"
       justifyContent="center"
       alignItems="center"
       className={styles.wrapper}
+      role="progressbar"
+      aria-valuetext="Er að vinna í innskráningu"
```
jeremybarbet, line 13: Translation for that copy?

libs/auth/react/src/lib/Authenticator/OidcSilentSignIn.tsx
```diff
@@ -0,0 +1,18 @@
...
+import { getUserManager } from '../userManager'
+
+// Note: This route gets mounted as an iframe to silently log the user in
+// For further optimization it can be minimized into a barebones html file
+export const OidcSilentSignIn = (): null => {
+  useEffect(() => {
+    const userManager = getUserManager()
+    userManager.signinSilentCallback().catch(function (error) {
+      // TODO: Handle error
+      console.log(error)
```
jeremybarbet, line 11: Is it using console.log on purpose rather than the logger?

jeremybarbet (commented):
Really nice, I just added a few comments

apps/service-portal/src/Main.tsx
```diff
@@ -3,8 +3,9 @@ import React from 'react'
 import ReactDOM from 'react-dom'
 import * as Sentry from '@sentry/react'
 import { Integrations } from '@sentry/tracing'
-import { environment } from './environments'
 
+import './auth'
```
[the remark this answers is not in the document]
eirikurn: Good catch. This is a mistake. Forgot to import it in application-system and remove tye old imports.

libs/auth/react/src/lib/Authenticator/AuthenticatorLoadingScreen.tsx
eirikurn: That's tricky :/ 

We haven't loaded any translations at this point, while loading/authenticating.

jeremybarbet (approved):
LGTM

PR #3766, opened 2021-05-07
fix(judicial-system): Court start date is now the same as courtDate, not the current day

oddsson:
# Court start date is now the same as courtDate, not the current day

https://app.asana.com/0/1199153462262248/1200295410102943

## What

Court start date is now the same as courtDate, not the current day

## Checklist:


apps/judicial-system/web/src/routes/Judge/CourtRecord/CourtRecord.tsx
```diff
@@ -193,7 +193,7 @@ export const CourtRecord: React.FC = () => {
                         onChange={(evt) =>
                           validateAndSetTime(
                             'courtStartTime',
-                            new Date().toString(),
+                            workingCase.courtDate,
```
gudjong, line 196: This looks good and should take care of the case when we are setting the court start and end times.

I wonder about the court date itself. What if the registrar prepares the start and perhaps event the end times, but then it is decided to postpone the court session for a day. The registrar would go back and set the court date, but that would not affect the start and end times. Perhaps we should modify the start and end times when setting the court date, given that they have been set. What do you think?

gudjong (approved):

PR #3772, opened 2021-05-07
feat(web): Don't open chat panel automatically on load

Ingimarsson:
# Don't open chat panel automatically on load

## What

The chat bot currently always pops up on page load. Instead we want it to be hidden and only show the button to open it.

## Why

Requested by Digital Iceland.


kksteini (approved):
LGTM

PR #3773, opened 2021-05-07
chore(application-system): Improve logging and refactor parental leave template

baering:
# \<Description\>

- Persist response when saving an application
- Throw the error when sending an application fails
- Create util to validate current environment
- Improve logging for vmst module when not running on production
- Refactor getExpectedDateOfBirth to use selectedChild
- Remove isDev from template
- Refactor how application is transformed to VMST application dto
- Catch all errors in application service in GraphQL API
- Update tests to reflect new changes

## Checklist:


libs/application/template-api-modules/src/lib/modules/templates/parental-leave/parental-leave.service.ts
```diff
@@ -56,8 +56,11 @@ export class ParentalLeaveService {
       } else {
         throw new Error('Could not send application')
       }
+
+      return response
     } catch (e) {
       console.log('Failed to send application', e)
```
jeremybarbet, line 62: Shouldn't it be using the logger here?

libs/application/templates/parental-leave/src/lib/answerValidators.spec.ts
```diff
@@ -6,26 +6,33 @@ import { minPeriodDays, usageMaxMonths } from '../config'
...
+              parentalRelation: 'primary',
+            },
+          ],
         },
       },
     },
     id: '',
     modified: new Date(),
     state: '',
+    status: 'inprogress' as Application['status'],
```
jeremybarbet, line 35: It could probably be `ApplicationStatus.IN_PROGRESS` couldn't it?

libs/utils/api/.eslintrc
```diff
@@ -1 +1,3 @@
-{ "extends": "../../../.eslintrc.json", "rules": {}, "ignorePatterns": ["!**/*"] }
+{ "extends": "../../../.eslintrc.json", "rules": {
```
jeremybarbet, line 1: Look like some wrong prettier here

jeremybarbet (commented):
Looks good, just a few comments 👍

libs/utils/api/.eslintrc
baering: 🤷  I don't even remember [touching this file](https://github.com/island-is/island.is/commit/650c66c0f37c3513cb365f94ff5dd6750bc5f7bb#diff-dab8984f2c6d62a1a30064e49788515551d4f4b0278dce04746e4ea094259fc1), but there are no lint or formatting errors

PR #3801, opened 2021-05-11
feat('auth-public-api'): Delegation features

unnarb:
#Delegation features

## What

## Why

## Screenshots / Gifs

## Checklist:


valurefugl (approved):

apps/services/auth-public-api/src/app/modules/delegations/rsk.config.ts
```diff
@@ -0,0 +1,26 @@
...
+import { createXRoadAPIPath, XRoadMemberClass } from '@island.is/utils/api'
+import { Base64 } from 'js-base64'
+
+const XROAD_BASE_PATH = process.env.XROAD_BASE_PATH_WITH_ENV ?? ''
+const XROAD_MEMBER_CLASS = XRoadMemberClass.GovernmentInstitution
+const XROAD_MEMBER_CODE = process.env.XROAD_RSK_MEMBER_CODE ?? ''
+const XROAD_API_PATH = process.env.XROAD_RSK_API_PATH ?? ''
+const XROAD_CLIENT_ID = process.env.XROAD_RSK_CLIENT_ID ?? ''
+const USERNAME = process.env.RSK_USERNAME ?? ''
+const PASSWORD = process.env.RSK_PASSWORD ?? ''
```
darrikonn, line 11: These should come from the environments file.

apps/services/auth-public-api/src/app/modules/delegations/rsk.config.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export class RskConfig {
+  static get(): RskModuleConfig {
+    return <RskModuleConfig>{
```
darrikonn, line 15: Better readability to just return a constant object here. No need to create the class with a static method.

libs/auth-api-lib/src/lib/entities/models/delegation-scope.model.ts
```diff
@@ -0,0 +1,49 @@
...
+  ForeignKey,
+} from 'sequelize-typescript'
+import { ApiProperty } from '@nestjs/swagger'
+import { Delegation } from './delegation.model'
+
+@Table({
+  tableName: 'delegation_scope',
+  indexes: [
+    {
+      fields: ['delegation_id, scope_name'],
```
darrikonn, line 18: Primary keys are already indexed. This is not needed, and I'm not sure how the ORM will try to map to a index.

darrikonn: I see this is done in the other models as well. Causes a confusion to the reader

libs/auth-api-lib/src/lib/services/delegation-scope.service.ts
```diff
@@ -0,0 +1,60 @@
...
+  ): Promise<DelegationScope | null> {
+    this.logger.debug('Creating new delegation scope')
+    return await this.delegationScopeModel.create({ ...delegationScope })
+  }
+
+  async createMany(delegationId: string, scopes: string[]): Promise<any> {
+    const arr: DelegationScopeDTO[] = []
+    for (let i = 0; i < scopes.length; i++) {
+      arr.push({ delegationId: delegationId, scopeName: scopes[i] })
+    }
```
darrikonn, line 27: Just a food for thought: using a `map` is most often preferred due to its immutability and readability (the code becomes more predictable):
```typescript
const delegationScopes: DelegationScopeDTO[] = scopes.map(scope => ({delegationId, scopeName: scope}))
```

libs/auth-api-lib/src/lib/services/delegations.service.ts
```diff
@@ -86,19 +90,119 @@ export class DelegationsService {
...
+
+    return response
+  }
+
+  async deleteTo(nationalId: string, id: string): Promise<number> {
+    this.logger.debug(`Deleting Delegation for Id ${id}`)
+
+    const delegation = await this.delegationModel.findByPk(id)
+
+    if (!delegation || delegation?.toNationalId !== nationalId) {
```
darrikonn, line 187: This should most likely be handled in the controller, changing the url to:
```typescript
/users/:nationalId/delegations/:id
```
Then return `404` when not found for that user.

libs/auth-api-lib/src/lib/services/delegation-scope.service.ts
```diff
@@ -0,0 +1,60 @@
...
+    private delegationScopeModel: typeof DelegationScope,
+    @Inject(LOGGER_PROVIDER)
+    private logger: Logger,
+  ) {}
+
+  async create(
+    delegationScope: DelegationScopeDTO,
+  ): Promise<DelegationScope | null> {
+    this.logger.debug('Creating new delegation scope')
+    return await this.delegationScopeModel.create({ ...delegationScope })
```
darrikonn, line 20: This goes for the whole PR > You can just return the promise: skipping `async` + `await` ... not requesting changes, but it makes the code more readable.

libs/auth-api-lib/src/index.ts
```diff
@@ -68,6 +71,7 @@ export * from './lib/services/sequelizeConfig.service'
 export * from './lib/services/idp-provider.service'
 export * from './lib/services/translation.service'
 export * from './lib/services/delegations.service'
+export * from './lib/services/delegation-scope.service'
```
darrikonn, line 74: Would be better to actually specify "WHAT" this library is exporting, instead of just export everything. It makes the abstraction more clear, and it'll be easier to use the API of the library.

e.g. `'./lib/services/sequelizeConfig.service'` this should not be exported and should not be used anywhere else than in this library.

libs/auth-api-lib/src/lib/entities/dto/delegation-scope.dto.ts
```diff
@@ -0,0 +1,12 @@
+import { ApiProperty } from '@nestjs/swagger'
+import { IsString } from 'class-validator'
+
+export class DelegationScopeDTO {
+  @IsString()
+  @ApiProperty()
+  delegationId!: string
+
+  @IsString()
+  @ApiProperty()
```
darrikonn, line 10: `ApiPropertyOptional`

darrikonn: The swagger documentation is most likely only for island.is devs, right? ... so it doesn't really matter.

libs/auth-api-lib/src/lib/services/delegation-scope.service.ts
```diff
@@ -0,0 +1,60 @@
...
+  ): Promise<number> {
+    if (scopeName) {
+      return await this.delegationScopeModel.destroy({
+        where: { delegationId: delegationId, scopeName: scopeName },
+      })
+    }
+
+    return await this.delegationScopeModel.destroy({
+      where: { delegationId: delegationId },
+    })
```
darrikonn, line 58: Don't need the if statement. The `scopeName` will be null and thus finds the correct record.

libs/auth-api-lib/src/lib/services/delegation-scope.service.ts
```diff
@@ -0,0 +1,60 @@
...
+  ): Promise<DelegationScope[] | null> {
+    if (scopeName) {
+      const response = await this.delegationScopeModel.findOne({
+        where: { delegationId: delegationId, scopeName: scopeName },
+      })
+      return <DelegationScope[]>[response]
+    }
+    return await this.delegationScopeModel.findAll({
+      where: { delegationId: delegationId },
+    })
```
darrikonn, line 43: Same here, `findOne` can be `findAll` since you return it as an array.

libs/auth-api-lib/src/lib/services/delegations.service.ts
```diff
@@ -86,19 +90,119 @@ export class DelegationsService {
...
+    const id = uuid()
+    await this.delegationModel.create({
+      id: id,
+      fromNationalId: nationalId,
+      ...delegation,
+    })
+    if (delegation.scopes) {
+      this.delegationScopeService.createMany(id, delegation.scopes)
+    }
+    return this.findOne(nationalId, id)
```
darrikonn, line 113: `this.delegationModel.create` already returns the instance.. you can just return that.

libs/auth-api-lib/src/lib/services/delegations.service.ts
```diff
@@ -86,19 +90,119 @@ export class DelegationsService {
...
+
+  async update(
+    nationalId: string,
+    delegation: DelegationDTO,
+    id: string,
+  ): Promise<Delegation | null> {
+    this.logger.debug(`Updating a delegation with id ${id}`)
+    await this.delegationModel.update(
+      { ...delegation },
+      { where: { id: id, fromNationalId: nationalId } },
```
darrikonn, line 124: If the id is found, but doesn't belong to the user ... a `404` should be returned. That logic should be implemented in the controller.

darrikonn (changes requested):
My main concern here is the request validation.

libs/auth-api-lib/src/lib/services/delegations.service.ts
unnarb: Shouldn't at least also be checked in the service so the service is not trusting the controller to take care of this?

libs/auth-api-lib/src/lib/services/delegation-scope.service.ts
unnarb: I would think the sql would try to delete a record where delagationId == X and scopeName == null which would not be found.

libs/auth-api-lib/src/lib/services/delegations.service.ts
unnarb: Should we maybe throw an Exception like UnAuthorized rather than 404?

libs/auth-api-lib/src/index.ts
unnarb: The API's import this for the SequalizeModule in their app.module class
SequelizeModule.forRootAsync({
      useClass: SequelizeConfigService,
    }),

libs/auth-api-lib/src/lib/entities/models/delegation-scope.model.ts
unnarb: I have seen bugs when this is not done.. but I will need to test this without defining this.

libs/auth-api-lib/src/lib/entities/dto/delegation-scope.dto.ts
unnarb: We use this for testing the service for instance

libs/auth-api-lib/src/lib/services/delegations.service.ts
unnarb: We are throwing Unauthorized exception now

darrikonn: Yeah both are valid,.. I usually go with 404 in these cases. The interpretation behind that is:
There is no delegation with that id that you can view... thus 404. In addition, I don't want the owner to know that there exists a record with that ID but it's owned by someone else. Therefore as far as the user (with nationalId) is concerned, a record with this ID doesn't exist.

But again, you can counter with arguments in favour of Unauthorized. So it's up to you how you interpret it, and then to try to enforce that rule to be consistent.

libs/auth-api-lib/src/lib/entities/models/delegation-scope.model.ts
darrikonn: hmm yeah ok, interesting 🤔

libs/auth-api-lib/src/lib/services/delegation-scope.service.ts
darrikonn: Mm I see, might be that I misinterpreted this. Is it supposed to delete all delegation scopes if no specific scope name is applied?

darrikonn: If so, you can ignore my comment.

libs/auth-api-lib/src/lib/services/delegation-scope.service.ts
darrikonn: Then this should also be ignored.
I originally thought that the scope name could be null.

libs/auth-api-lib/src/index.ts
valurefugl: We will take a look at this during refactor week.

libs/auth-api-lib/src/lib/services/delegations.service.ts
valurefugl: Is this a showstopper?

valurefugl:
@darrikonn we really would like to move forward with this, can you please review the resolutions?

libs/auth-api-lib/src/lib/services/delegations.service.ts
darrikonn: IMO yes. The consumer of this API (wether it's the GraphQL API, or the client) needs to take some actions based on the HTTP status codes. As it's now, only 200/201 is returned, even though the action wasn't successful.

unnarb: The Api is returning 401 status when the delegation is not owned by the current user. This could easily be 404 or 403 (Forbidden). Whatever is preferred. 
The API therefor returns 401 in this case before trying to delete the record(s).

darrikonn: Where is that happening?
e.g.
```typescript
// controller
Promise<number> {
  return await this.delegationsService.deleteFrom(user.nationalId, id)
}
```
```typescript
// service
if (!delegation || delegation?.fromNationalId !== nationalId) {
  this.logger.debug('Delegation does is not assigned to user')
  return 0
}
```

unnarb: I think you are looking at an older version. This is how we are doing this now:

```
if (!delegation || delegation?.fromNationalId !== nationalId) {
      this.logger.debug('Delegation is not assigned to user')
      throw new NotFoundException()
    }
```

darrikonn: Geez, sorry. I always clicked on the file link in this conversation where I'm routed to (as you say) the old commit :/.

darrikonn (approved):

PR #3806, opened 2021-05-11
feat(service-portal): Initialize access control module

darrikonn:
# Initialize access control module

## Todos (draft reason)

- ~[x] Navigation should check on enabled modules (feature flags) before rendering (fixed in https://github.com/island-is/island.is/pull/3815)~
- [x] Setup flag for access control on ConfigCat (enabled on dev)

## What

* Move `settings` to `settings/personal-information`
* Initialize access control module

## Screenshots / Gifs

accesscon

## Checklist:


SUMMARY.md
```diff
@@ -177,6 +177,7 @@
   - [Service Portal Education Student Assessment](libs/service-portal/education-student-assessment/README.md)
   - [Service Portal Eligibility](libs/service-portal/eligibility/README.md)
   - [Icelandic Names Registry](libs/service-portal/icelandic-names-registry/README.md)
+    - [Service Portal Settings Access Control](libs/service-portal/settings/access-control/README.md)
```
jeremybarbet, line 180: It's missing a README.md at the root of the settings folder cf:
https://docs.devland.is/repository/gitbook-contribute#root-readme-md-within-handbook-directory

darrikonn: Is that the general rule for parent directories?

Since the settings folder isn't a library, but a parent folder for multiple libraries. That way, we need to `nx` ignore it as well.

libs/service-portal/core/src/lib/navigation/paths.ts
```diff
@@ -13,6 +13,12 @@ export enum ServicePortalPath {
 
   // Settings
   SettingsRoot = '/stillingar',
+  SettingsAccessControl = '/stillingar/adgangsstyring',
+  SettingsPersonalInformation = '/stillingar/personuupplysingar',
+  SettingsPersonalInformationEditPhoneNumber = '/stillingar/personuupplysingar/breyta-simanumeri',
+  SettingsPersonalInformationEditEmail = '/stillingar/personuupplysingar/breyta-netfangi',
+  SettingsPersonalInformationEditLanguage = '/stillingar/personuupplysingar/breyta-tungumali',
+  SettingsPersonalInformationEmailConfirmation = '/stillingar/personuupplysingar/stadfesta-netfang/:hash',
```
atligudl, line 21: The user profile service generates an email that has a hardcoded link to the old path, that will need to be updated as well. See: https://github.com/island-is/island.is/blob/main/apps/services/user-profile/src/app/user-profile/verification.service.ts

@obmagnusson Might have some more insights

atligudl (changes requested):
One comment, rest looks good

libs/service-portal/core/src/lib/navigation/paths.ts
darrikonn: Great spot! That email should be using the navigation paths to prevent these errors from happening 👍 .. fixing it!

darrikonn: Fixed.

atligudl (approved):

SUMMARY.md
jeremybarbet: Yes, it's the general rule. If the folder contains only libraries/apps then we add the root README to the .nxignore. It's afaik the only way for Gitbook to successfully create the default root page to have a link to each sub-README within, for example: https://docs.devland.is/technical-overview/x-road

darrikonn: Makes sense 👍

jeremybarbet (approved):

dd-sdang (approved):
LGTM

PR #3814, opened 2021-05-11
feat(application-templates-parental-leave): Fetch contact info from external data and be able to mock children data provider

baering:
# \<Description\>

## What

- Fetch contact info from external info if it is not found in answers
  - There is a bug in the AS navigation so that it skips the contact details screen for secondary parents (that conditionally render choosing the other parent)
- Be able to mock children data provider

## Why

- To be able to send the application
- To be able to create an application when not expecting a child, also to be able to create multiple applications

## Screenshots / Gifs
By answering 'yes' to using mock data we skip the queries in the data provider and make it return the data from these inputs as a child

image

Screen capture:

https://user-images.githubusercontent.com/6312838/117894721-82973a00-b2ac-11eb-9725-f97ae3a77667.mov

## Checklist:


libs/application/templates/parental-leave/src/forms/Prerequisites.ts
```diff
@@ -26,6 +29,65 @@ export const PrerequisitesForm: Form = buildForm({
...
+                  condition: (answers) =>
+                    answers.useMockData === YES &&
+                    answers.useMockedParentalRelation === 'secondary',
+                  placeholder: '1234567-7890',
+                  format: '######-####',
+                }),
+              ],
+            }),
+          ],
+        }),
```
baering, line 90: I intentionally skipped translations for this, as it is temporary™.

Do we need it?

jeremybarbet: Agree, it's only temporary™. However, I would probably add a condition on the whole section to not show up on production, just in case™.

madebynoam (approved):
Lgtm!

jeremybarbet (approved):
LGTM

PR #3815, opened 2021-05-12
fix(service-portal): Fix override of enabled modules when system route is defined

darrikonn:
# Fix override of enabled modules when system route is defined

## What

Fix navigation when a system route is defined

## Why

This was overriding the cases when a feature was disabled from Config Cat.

## Screenshots / Gifs

fixnavroutes

## Checklist:

### DIFF
--- a/apps/service-portal/src/hooks/useNavigation/useNavigation.ts
+++ b/apps/service-portal/src/hooks/useNavigation/useNavigation.ts
@@ -21,7 +21,7 @@ const filterNavigationTree = (
         (Array.isArray(route.path) &&
           item.path &&
           route.path.includes(item.path)),
-    ) !== undefined || item.systemRoute === true
+    ) !== undefined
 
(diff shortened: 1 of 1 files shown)


atligudl (changes requested):
Is /stillingar/adgangsstyring defined as a system route?
The only functionality of this flag is: "render this item into the nav even though you can't find the module for it"

You can see that "Fjármál" and "Fasteignir" in the purple nav disappear once you apply the change in the gif

darrikonn:
> Is /stillingar/adgangsstyring defined as a system route?
> The only functionality of this flag is: "render this item into the nav even though you can't find the module for it"
> 
> You can see that "Fjármál" and "Fasteignir" in the purple nav disappear once you apply the change in the gif

Yeah, this was me misinterpreting the API (i.e. the system route prop).
Thanks for clarifying!

PR #3149, opened 2021-03-05
fix(web): Strict typechecking errors

maranomynet:
## What

Flips the hooks folder to `"strict": true`  
Fixes to various typing bugs that become visible when `"strict": false` is removed.  
Add some tests and an optional-to-use `useNamespaceStrict()` hook with generics (and blackjack).

## Why

To make it easier to eventually return to strict mode for the `web` app ... and reduce the number of runtime bugs meanwhile.

## Checklist:

- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [x] ~~I have made corresponding changes to the documentation~~
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] Formatting passes locally with my changes
- [ ] I have rebased against main before asking for a review

PR #3819, opened 2021-05-12
fix(download-service): hotfix-download service auth and timeout extend

Thorrik58:
# Download service auth and timeout remove

## What

Increasing the timeout to 30s , added exeption handling on the document client.
Body token was incorrect in controller after auth changes from [#3149](https://github.com/island-is/island.is/pull/3419)

## Why

Some document requests are taking longer than 10s to respond and thus returning an unhandled error which was causing problems for the app. 

## Screenshots / Gifs

## Checklist:


obmagnusson (approved):
LGTM

PR #3839, opened 2021-05-14
fix: Aafgp - set ready for prod flag

Thorrik58:
Application for funding government projects is missing the ready for production flag.
Also translation added for application name.


obmagnusson (approved):
LGTM

PR #3846, opened 2021-05-17
fix(icelandic-names-registry/web): finalizing changes/fixes (#3755)

stjanilofts:
https://github.com/island-is/island.is/pull/3755


darrikonn (approved):

kksteini (approved):
Approved previously on a PR

PR #3849, opened 2021-05-17
feat(user-profile): Adding audit logs and authentication to the user-profile api

saevarma:
# Adding audit logs and authentication to the user-profile api

https://www.notion.so/Integrate-audit-log-into-our-services-8fca3f2f49484f1f96a3837fc62f1cfd

## What

Update the user-profile api to audit log operation and require user authentication.

## Why

As part of implementing the delegation system we need to audit log our services to track who did what for whom.

## Checklist:


apps/services/user-profile/src/app/user-profile/userProfile.controller.ts
```diff
@@ -153,12 +220,20 @@ export class UserProfileController {
...
+    nationalId: string,
+    @CurrentUser()
+    user: User,
+  ): Promise<EmailVerification> {
+    // findOneByNationalId must be first as it implictly checks if the
+    // route param matches the authenticated user.
+    const profile = await this.findOneByNationalId(nationalId, user)
     if (!profile.email) {
-      return null
+      throw new BadRequestException('Profile have no configured email address.')
```
obmagnusson, line 236: 'Profile has no ...' or '... does not have a configured ...'

saevarma: Takk!

obmagnusson (approved):
LGTM

libs/auth/scopes/src/lib/application.scope.ts
```diff
@@ -0,0 +1,4 @@
+export enum ApplicationScope {
+  read = '@island.is/applications:read',
```
pshomov, line 2: Should we consider shortening the scopes' names? Is there a limit on the JWT token in IDS we should consider @valurefugl? I recall having problems with JWT size in Auth0 due to being overly verbose in claims.

eirikurn: Yeah, good question. I did some testing a while back, and with our current production servers and proxies, we support up to 4KB header values. This allows us to use around 100 scopes of this length. It "should" be enough, but I'm definitely open to better naming conventions.

We want to namespaces to avoid conflicts between gov orgs. I like readability, but I know some OAuth2 providers use fairly obfuscated scope names. "island/app:r"?

eirikurn (approved):
Looks great! I love the scope library. 👍

libs/auth/scopes/src/lib/application.scope.ts
valurefugl: Am not aware of any limits in the IDS itself.

PR #3855, opened 2021-05-17
refactor(parental-leave): review screen to show data non editable by default

jeremybarbet:
## What

- https://app.asana.com/0/1182378413629561/1200214178491343

https://user-images.githubusercontent.com/937328/118646527-d388c000-b7cf-11eb-9ad8-4b909a94b1ec.mov


libs/application/core/src/lib/messages.ts
```diff
@@ -134,6 +139,16 @@ export const coreMessages = defineMessages({
     description:
       'To continue the application process, you will need to sign in.',
   },
+  radioYes: {
+    id: 'application.system:radio.option.yes',
+    defaultMessage: 'Já',
+    description: 'Yes option value',
+  },
+  radiosNo: {
```
baering, line 147: Very minor, but you have plural `radios` here.

libs/application/templates/parental-leave/src/fields/Review/index.tsx
```diff
@@ -171,231 +148,456 @@ const Review: FC<ReviewScreenProps> = ({
...
-            <Box paddingY={4}>
-              <GridRow>
-                <GridColumn span="6/12">
-                  {editable ? (
-                    <Input
-                      id="payments.bank"
-                      name="payments.bank"
+    <>
+      <ReviewGroup
+        isEditable={editable}
```
baering, line 154: Here we need to keep in mind that only a primary parent can edit the other parent

libs/application/templates/parental-leave/src/fields/Review/index.tsx
```diff
@@ -411,87 +613,205 @@ const Review: FC<ReviewScreenProps> = ({
...
+              <DataValue
+                label={formatMessage(parentalLeaveFormMessages.employer.email)}
+                value={employerEmail}
+              />
+            )}
+          </GridColumn>
+        </GridRow>
+      </ReviewGroup>
+
+      <ReviewGroup isEditable={editable}>
```
baering, line 749: I assume this will not be editable so can probably just be `false`

baering (commented):
Looks very good 💯  We'll need to keep in mind which parent is applying to know what they can edit. For example only a primary parent can change the other parent. Also if at some point we want to make rights editable that will also just be for primary parent. I'm not sure how we want to have it for personal allowance, if that should be chosen by primary parent like rights. But we'll discuss it later.

libs/application/templates/parental-leave/src/fields/Review/index.tsx
jeremybarbet: Good point 👍

baering (approved):

PR #3872, opened 2021-05-18
feat(services-party-letter-registry-api): First version

stefanhar:
# \<Description\>

We want to keep a registry of party letters for applications that require a party letter to be assigned.
We also want to be able to assign new party letters to national ids in the application system.

What
- Added database config for party letter registry
-  Added tests
- Added api

## Checklist:


apps/services/party-letter-registry-api/src/app/modules/partyLetterRegistry/partyLetterRegistry.controller.ts
```diff
@@ -0,0 +1,37 @@
...
+    const resource = await this.partyLetterRegistryService.findByOwner(owner)
+
+    if (!resource) {
+      throw new NotFoundException("This resource doesn't exist")
+    }
+
+    return resource
+  }
+
+  @Post()
```
eirikurn, line 33: These endpoints should have OAuth2 authentication with some scopes.

If the client is another service like the application system (and not a user with these scopes), then it would need client credentials to get an access token with these scopes.

apps/services/party-letter-registry-api/README.md
```diff
@@ -0,0 +1,8 @@
+# Party Letter Registry
+
+## About
+
+**This service needs to be accessed by users through another service providing appropriate authentication for getting this data**  
+The consuming application should validate that the requesting user is in managers  
+This service contains the political party letter registry.  
+The source of the data contained in this service is the Ministry of Justice and the party letter application currently contained within the application system.
```
eirikurn, line 8: How do you plan to prefill the database with existing party letters?

apps/services/party-letter-registry-api/migrations/20210514135724-party-letter-registry.js
```diff
@@ -0,0 +1,53 @@
...
+          allowNull: false,
+        },
+        created: {
+          type: Sequelize.DATE,
+          allowNull: false,
+        },
+        modified: {
+          type: Sequelize.DATE,
+          allowNull: false,
+        },
```
eirikurn, line 40: Should this be based on election terms, or should I be able to use some application to "edit" this data for future elections?

eirikurn (commented):
Note, as a core review, I'm mainly skimming the core files (CODEOWNERS), and checking a couple of best practices we're trying to enforce. You should always get a team member to review the files you plan to own.

That said, this looks like a solid service implementation.

Would like to see OAuth2 authorization. Do you think that's possible in your timeframe?

apps/services/party-letter-registry-api/README.md
stefanhar: We plan to implement a import script using the existing seeder functionality in sequelize.
The database would then be prefilled with data from a xlsx file, allowing us to keep that out of the public repository.
This will also serve as the method of cose to update data in the database until we implement an admin panel.

The plan is to implement a admin panel to manage these letters and make it accessible to the Ministry of justice to enable them to maintain this data.
We decided to postpone the admin panel until the agency system (umboðskerfi) has been implemented.
The agency system should also allow us to omit the `managers` field from this database.

apps/services/party-letter-registry-api/migrations/20210514135724-party-letter-registry.js
stefanhar: We plan for this to be a database containing the current state of party letters.
We consider this a source of truth when managing applications for party letter related activities

apps/services/party-letter-registry-api/src/app/modules/partyLetterRegistry/partyLetterRegistry.controller.ts
stefanhar: I don't think the current timeframe allows for implementation of OAuth2, but I see where you are coming from.
Our goal is to present a working system by thursday.
After that deadline the user flows and functionality need to be tested probably accompanied by some iterations.
I'm confident we can implement auth into this service during that timeframe

eirikurn (approved):

apps/services/party-letter-registry-api/migrations/20210514135724-party-letter-registry.js
```diff
@@ -0,0 +1,53 @@
...
+        party_letter: {
+          type: Sequelize.CHAR(2),
+          allowNull: false,
+          unique: true,
+        },
+        party_name: {
+          type: Sequelize.STRING,
+          allowNull: false,
+        },
+        owner: {
```
arnorhs, line 24: Is owner a single person? And is it not possible for a single person to be the owner of more than one party?

arnorhs (approved):

apps/services/party-letter-registry-api/migrations/20210514135724-party-letter-registry.js
stefanhar: Owner is a single person or company.
Single national id per party letter covers the required conditions for our current project.
It might be theoretically possible to have more than on political party assigned to a single national id

PR #3874, opened 2021-05-18
chore(api-domains-cms): Removed deprecated resolver and service functions and models from cms

kksteini:
# \<Description\>

## What

#2982 simplified the homepage by reducing queries. This was accomplished by combining unique content into a more encompassing content model on contentful. To avoid issues the old types weren't deleted straight away but now it is time.

## Why

Removes unused files

## Checklist:


libs/api/domains/cms/src/lib/dto/getFrontpageSliderList.input.ts
```diff
@@ -1,10 +0,0 @@
-import { Field, InputType } from '@nestjs/graphql'
```
kksteini, line 1: These gotta go too. I forgot to put a deprecation warning on but they are to be deleted since they are unused

libs/api/domains/cms/src/lib/dto/getHomepage.input.ts
```diff
@@ -1,9 +0,0 @@
-import { Field, InputType } from '@nestjs/graphql'
```
kksteini, line 1: These gotta go too. I forgot to put a deprecation warning on but they are to be deleted since they are unused

kksteini:
Still a draft because TODO:
Check what to remove from lib/api/mocks if necessary

libs/api/domains/cms/src/lib/cms.resolver.ts
```diff
@@ -254,17 +250,6 @@ export class CmsResolver {
...
-  @Directive(cacheControlDirective())
-  @Query(() => FrontpageSliderList, { nullable: true })
-  getFrontpageSliderList(
-    @Args('input') input: GetFrontpageSliderListInput,
-  ): Promise<FrontpageSliderList | null> {
-    return this.cmsContentfulService.getFrontpageSliderList(
-      input?.lang ?? 'is-IS',
-    )
-  }
-
```
stefanhar, line 267: Seems like this might be in use in the `service-portal` here:
`libs/service-portal/graphql/src/lib/queries/getFrontpageSlides.ts`
We should probably sync with the and move them to the new way of fetching this data

kksteini: Ah very good point, I missed this.
Seems like that query is also outdated in regards to the "animationJsonAsset".
Yes, we should sync with the service-portal-team.

kksteini: The queries should be mostly compatible with the new frontpage query:
https://github.com/island-is/island.is/blob/main/apps/web/screens/queries/Frontpage.ts#L35-L52

kksteini: Thank you :+1:

stefanhar (approved):

saevarma (approved):
LGTM

PR #3879, opened 2021-05-18
feat(parental-leave): allow attachment upload for self-employed

jeremybarbet:
## What

- Attachment upload for self employed


libs/application/template-api-modules/src/lib/modules/templates/parental-leave/parental-leave.service.ts
```diff
@@ -34,12 +42,43 @@ export class ParentalLeaveService {
...
+      (application.answers.employer as { isSelfEmployed: 'yes' | 'no' })
+        .isSelfEmployed === 'no'
+    ) {
+      return undefined
+    }
+
+    try {
+      const filename = (application.answers.employer as {
+        selfEmployed: { file: { key: string }[] }
+      }).selfEmployed.file?.[0].key
```
baering, line 56: Extract this to a function?

libs/application/template-api-modules/src/lib/modules/templates/parental-leave/parental-leave.service.ts
```diff
@@ -34,12 +42,43 @@ export class ParentalLeaveService {
...
+      return fileContent.toString('base64')
+    } catch (e) {
+      this.logger.error('Cannot get self employed attachment', { e })
+      return undefined
+    }
+  }
+
   async sendApplication({ application }: TemplateApiModuleActionProps) {
     const nationalRegistryId = application.applicant
+    const attachment = await this.getSelfEmployedPdf(application)
```
baering, line 76: Shouldn't we move the self employment check outside of the function

```typescript
let attachment

if (selfEmployed) {
  attachment = await this.getSelfEmployedPdf(application)
}
```

And by doing that, we can change lines 68-70 to not fail silently:

```typescript
catch (e) {
  this.logger.error('Cannot get self employed attachment', { e })
  return undefined
}
```

libs/application/templates/parental-leave/src/forms/ParentalLeaveForm.ts
```diff
@@ -414,6 +415,31 @@ export const ParentalLeaveForm: Form = buildForm({
...
+              id: 'employer.selfEmployed.attachment',
+              title: parentalLeaveFormMessages.selfEmployed.attachmentTitle,
+              description:
+                parentalLeaveFormMessages.selfEmployed.attachmentDescription,
+              condition: (answers) =>
+                (answers as {
+                  employer: {
+                    isSelfEmployed: string
+                  }
+                })?.employer?.isSelfEmployed === YES,
```
baering, line 428: I was going to suggest that you would:

```typescript
import { SchemaFormValues } from '../lib/dataSchema'
```

And then:
```typescript
{
  // ...
  condition: (answers: SchemaFormValues) => answers.employer?.isSelfEmployed === YES
}
```

But then I noticed `employer` does not exist in the `dataSchema` so you may want to add that as well and then you can try this out.

baering (commented):
Great work 🚀  left some comments

libs/application/template-api-modules/src/lib/modules/templates/parental-leave/parental-leave.service.ts
jeremybarbet: Sure that sounds good to me 👍

jeremybarbet: However, we still need the try/catch around the S3 function, but I will throw an error instead

libs/application/templates/parental-leave/src/forms/ParentalLeaveForm.ts
jeremybarbet: We discussed about it together on Slack. It will need the rework on the schema validators function in order to add it back to the data schema

PR #3887, opened 2021-05-19
feat(core): New national registry service

Rikkiabb:
# \<Description\>
Sets up connection with new national registry service through x-road. 
Sets  up an endpoint that will be used in children residence change application and uses new national registry service.
Helm variables have been merged see [here](https://github.com/island-is/helm/pull/491)

Here is the documentation about the service from national-registry 

Documentation about this service is [here](https://api-dev.skra.is)

Documentation about testing the service [here](https://www.notion.so/5831b7853df34aeeb763d789a1c81585?v=cf8edae088c34c4e9721c3343740a815)

TODO:
The new national registry service is currently only on dev. 
Some endpoints in national registry service are not working and are being fix as this is written.
Connect national registry to data provider to consume data, this will be done once the service is more hardened.

## Checklist:


pshomov:
What is this component "as"?

Rikkiabb:
@pshomov Named it as (abbreviation for Application system) since I was going to use the service there, but that was wrong, updated it to core 😊

apps/api/src/app/environments/environment.ts
```diff
@@ -91,4 +91,10 @@ export default {
   endorsementSystem: {
     baseApiUrl: 'http://localhost:4246',
   },
+  nationalRegistryXRoad: {
+    url: process.env.XROAD_BASE_PATH ?? 'http://localhost:8080/r1/IS-DEV',
+    memberCode: process.env.XROAD_TJODSKRA_MEMBER_CODE ?? '',
+    apiPath: process.env.XROAD_TJODSKRA_API_PATH ?? '',
+    clientId: process.env.XROAD_CLIENT_ID ?? '',
```
darrikonn, line 98: can't most of these be hard coded, since probably all of these aren't secrets?

libs/api/domains/national-registry-x-road/src/lib/national-registry-x-road.module.ts
```diff
@@ -0,0 +1,42 @@
+import { DynamicModule, Module } from '@nestjs/common'
+import { NationalRegistryModule } from '@island.is/clients/national-registry-v2'
+import { createXRoadAPIPath, XRoadMemberClass } from '@island.is/utils/api'
+
+import { NationalRegistryXRoadResolver } from './national-registry-x-road.resolver'
+import { NationalRegistryXRoadService } from './national-registry-x-road.service'
```
darrikonn, line 6: The filenames aren't following the file naming pattern for island.is.

It's correct to use `kebab-case` for library names, but the filenames + directories within should use `camel-case`. (`national-registry-x-road.resolver` => `nationalRegistryXRoad.resolver`).

darrikonn (changes requested):
I'm going to stop here since the whole PR is about to change due to the file naming pattern, and the code review might get funky.

libs/api/domains/national-registry-x-road/src/lib/nationalRegistryXRoad.service.ts
```diff
@@ -0,0 +1,124 @@
+import { logger } from '@island.is/logging'
+import { Inject, Injectable } from '@nestjs/common'
+import { ApolloError } from 'apollo-server-express'
+import { EinstaklingarApi } from '@island.is/clients/national-registry-v2'
+import { NationalRegistryXRoadConfig } from './nationalRegistryXRoad.module'
+import { ChildrenCustodyResponse } from '../models/childrenCustodyResponse.model'
+import { ChildrenCustodyChild } from '../models/childrenCustodyChild.model'
+
+const handleError = (error: any) => {
+  logger.error(error)
```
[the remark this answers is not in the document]
Rikkiabb: Great point will fix, just an fyi that most services in `api/domains` do logging this way, that is without injecting the logger 😱 So hopefully you are wrong 🤞

libs/api/domains/national-registry-x-road/src/models/nationalRegistryPerson.model.ts
```diff
@@ -0,0 +1,26 @@
+import { Field, ObjectType, ID } from '@nestjs/graphql'
+import { NationalRegistryAddress } from './nationalRegistryAddress.model'
+
+@ObjectType()
+export class NationalRegistryPerson {
```
darrikonn, line 5: :+1:

libs/api/domains/national-registry-x-road/src/lib/nationalRegistryXRoad.resolver.ts
```diff
@@ -0,0 +1,29 @@
...
+
+@UseGuards(IdsAuthGuard, IdsUserGuard, ScopesGuard)
+@Resolver()
+export class NationalRegistryXRoadResolver {
+  constructor(
+    private nationalRegistryXRoadService: NationalRegistryXRoadService,
+  ) {}
+
+  @Query(() => NationalRegistryPerson, { nullable: true })
+  async getChildrenCustodyAndParents(
```
darrikonn, line 21: This is looking a lot better!

The only thing that's missing is using the power of GraphQL to only fetch specific fields when requested... here if you make the resolver a resolver for the `nationalRegistryPerson`, you can use decorators such as ResolveField to explicitly fetch the desired fields.
Steps to do that:
1. @Resolver(() => NationalRegistryPerson)
2. Rename query from `getChildrenCustodyAndParents` to `nationalRegistryPersons` (don't use get.* to follow the graphql naming conventions)
3. Rename `this.nationalRegistryXRoadService.getCustodyChildrenAndParents` to `this.nationalRegistryXRoadService.getNationalRegistryPerson` and only return `einstaklingarGetEinstaklingur`
4. Make `children` a `ResolveField` which explicitly calls `this.nationalRegistryXRoadService.getCustodyChildren` where the rest of the API calls are made


Here you save a ton of requests when it's not needed. Let's say a client only wants to fetch the address, or the fullName:
```graphql
query nationalRegistryUser {
  nationalId
  fullName
}
```
now only a single API call is made, since `children` wasn't explicitly asked for => which I call the power of GraphQL :D

darrikonn (changes requested):
Looking a lot better... minor improvements to make it 100%

libs/api/domains/national-registry-x-road/src/lib/nationalRegistryXRoad.resolver.ts
Rikkiabb: Niceeee

darrikonn (approved):
Nice

stefanhar (approved):

Thorrik58 (approved):
LGTM

PR #3902, opened 2021-05-20
feat(web): Add new slice 'Overview Links'

Ingimarsson:
# Add new slice 'Overview Links'

## What

This adds a new slice. It's layout is taken from https://island.is/throun

## Why

It's been requested that we move the content of `/throun` to the Stafrænt Ísland organization page.

## Screenshots / Gifs

The slice on a subpage

island brynjar_stofnanir_stafraent-island_throun

## Checklist:


ingolfurs (approved):

libs/api/domains/cms/src/lib/models/overviewLinks.model.ts
```diff
@@ -0,0 +1,27 @@
...
+@ObjectType()
+export class OverviewLinks {
+  @Field(() => ID)
+  id!: string
+
+  @Field(() => [IntroLinkImage])
+  overviewLinks?: Array<IntroLinkImage>
+
+  @Field(() => Link, { nullable: true })
+  link?: Link | null
```
stefanhar, line 16: According to mapper these fields should never be undefined as they default to empty array and null.
Having them optional might not be wanted

stefanhar (commented):
Backend LGTM

libs/api/domains/cms/src/lib/models/overviewLinks.model.ts
Ingimarsson: Right, fixed that :)

stefanhar (approved):

PR #3905, opened 2021-05-20
feat(judicial-system): Court case facts and arguments

oddsson:
# Add fields courtCaseFacts and courtLegalArguments to Case.

https://app.asana.com/0/1199153462262248/1200221937662924

## What

Add fields courtCaseFacts and courtLegalArguments to the ruling screen.

## Why

The court needs to be able to change the case facts and legal arguments that are set by the prosectutor.

## Screenshots / Gifs

screencapture-localhost-4200-domur-urskurdur-ee5b7533-7a68-436c-9d8e-330c4b2203b2-2021-05-20-13_26_43

## Checklist:


apps/judicial-system/backend/src/app/formatters/rulingPdf.ts
```diff
@@ -215,7 +215,7 @@ export async function getRulingPdfAsString(
     .text('Málsatvik')
     .font('Helvetica')
     .fontSize(12)
-    .text(existingCase.caseFacts, {
+    .text(existingCase.courtCaseFacts, {
```
gudjong, line 218: This will cause a problem is someone retrieves the ruling pdf for an older case.

We have two ways to address this:

1.  use existingCase.courtCaseFacts || existingCase.caseFacts when generating the pdf
2. copy the value from case_facts to court_case_facts during the db migration

apps/judicial-system/backend/src/app/formatters/rulingPdf.ts
```diff
@@ -226,7 +226,7 @@ export async function getRulingPdfAsString(
     .text('Lagarök')
     .font('Helvetica')
     .fontSize(12)
-    .text(existingCase.legalArguments, {
+    .text(existingCase.courtLegalArguments, {
```
gudjong, line 229: Same problem as above.

apps/judicial-system/web/src/routes/Judge/Ruling/StepOne/RulingStepOne.spec.tsx
```diff
@@ -31,6 +31,18 @@ describe('/domari-krafa/urskurdur', () => {
...
+              courtCaseFacts: 'Court Case Facts',
+            } as UpdateCase,
+            {
+              id: 'test_id_3',
+              courtLegalArguments: 'Court Legal Arguments',
+            } as UpdateCase,
+            {
+              id: 'test_id_3',
+              ruling: 'Court Legal Arguments',
+            } as UpdateCase,
```
gudjong, line 45: Why this?

gudjong (changes requested):

apps/judicial-system/backend/src/app/formatters/rulingPdf.ts
oddsson: Good catch 👀

I think way nr 2 is the way to go because caseFacts should never be under málsatvik anymore so `existingCase.courtCaseFacts || existingCase.caseFacts` is incorrect.

gudjong: Implemented in the migration script.

gudjong (approved):

PR #3909, opened 2021-05-20
chore(ui-core): Add arrow down to icon map

atligudl:
# \<Add arrow down to icon map\>

## What

Arrow down added to iconMap Icon type

## Why

The icon exists as an asset but was missing in the type

## Checklist:


DavidBachmann (approved):
🤯

PR #3912, opened 2021-05-20
fix(core): Avoid commenting

darrikonn:
Various reasons why this should be removed (since devs should not be encouraged to go this way):

- A good code should explain itself with good variable naming.
- Comments tend to get outdated when code gets updated.
- Comments should be treated as second class citizens, and used in rare cases where the code does not fall under the criteria of self explanation.


sindrig (approved):
LGTM

PR #3918, opened 2021-05-21
fix(parental-leave): Fix cases where we allowed undefined dates

madebynoam:
Adds more validation for undefined dates.

## Why

We were allowing to Continue with undefined dates, and it caused periods with missing `endDates` or `startDates`.

## Checklist:


jeremybarbet (approved):
LGTM 🚀

PR #3924, opened 2021-05-21
chore(core): split up aranja codeowner teams

eirikurn:
## What

Split up aranja github/codeowner teams.

## Why

To optimize review requests.


darrikonn (approved):
Nice one!

PR #3933, opened 2021-05-21
feat(application-templates-driving-license): application for driving license (full) & driving assessment by teacher

arnorhs:
# Driver's license application

- Implementation of new API calls
- Updated driver's license application to be strictly for "fullnaðarskírteini"
- Added a new application for teachers to submit driving assessments for their students (may be performed irrespective of whether or not a student applies digitally etc

## Checklist:


libs/api/domains/driving-license/src/lib/graphql/applicationEligibilityRequirement.model.ts
```diff
@@ -0,0 +1,11 @@
+import { Field, ObjectType, ID } from '@nestjs/graphql'
+
+@ObjectType()
+export class ApplicationEligibilityRequirement {
+  @Field(() => ID)
+  key!: string
+
+  @Field(() => Boolean)
```
stefanhar, line 8: Graphql generation should resolve basetypes automagically.
There is no need for a type resolution function in these cases

jGeir (approved):

libs/api/domains/driving-license/src/lib/graphql/newDrivingLicenseResult.model.ts
```diff
@@ -0,0 +1,14 @@
...
+export class NewDrivingLicenseResult {
+  @Field()
+  @IsBoolean()
+  success!: boolean
+
+  @Field({ nullable: true })
+  @IsOptional()
+  @IsString()
+  errorMessage!: string | null
+}
```
stefanhar, line 14: Are the class validators on the return type possibly a mistake? Or am I missing something?

libs/api/domains/driving-license/src/lib/graphql/studentInformationResult.model.ts
```diff
@@ -0,0 +1,17 @@
...
+  @IsString()
+  @IsNotEmpty()
+  name!: string
+}
+
+@ObjectType()
+export class StudentInformationResult {
+  @Field(() => StudentInformation, { nullable: true })
+  @IsOptional()
+  student!: StudentInformation | null
```
stefanhar, line 16: Are the class validators on the return type possibly a mistake? Or am I missing something? (repeat)

libs/api/domains/driving-license/src/lib/graphql/main.resolver.ts
```diff
@@ -49,4 +58,71 @@ export class MainResolver {
   drivingLicensePenaltyPointStatus(@CurrentUser() user: User) {
     return this.drivingLicenseService.getPenaltyPointStatus(user.nationalId)
   }
+
+  @Query(() => HasTeachingRights)
+  drivingLicenseTeachingRights(@CurrentUser() user: User) {
```
stefanhar, line 63: TS return types are inconsistently defined in this resolver.
Consider adding return types to all resolve functions as this would ensure this resolver return correct type if changes are made to the underling service

libs/application/core/src/types/ApplicationTypes.ts
```diff
@@ -35,6 +36,10 @@ export const ApplicationConfigurations = {
     slug: 'okuskirteini',
     translation: 'dl.application',
   },
+  [ApplicationTypes.DRIVING_ASSESSMENT_APPROVAL]: {
+    slug: 'akstursmat',
+    translation: 'dl.application',
```
stefanhar, line 41: `DRIVING_LESSONS`, `DRIVING_LICENSE ` and `DRIVING_ASSESSMENT_APPROVAL` all belong to the same namespace `dl`, is this intentional since this might cause messages to conflict in Contentful if any messages fields have the same keys.  
I'm not sure how conflicts are handled inside the export strings function, this might fail safe.

libs/application/templates/driving-assessment-approval/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+This library is a reference how all application template libraries can be.
+
+## Requirements
+
+There are multiple requirements needed for a new template to be usable by the application system:
+
+1. Add a unique application type to `application/core/src/types/ApplicationType.ts`
+2. Build a new library similar to this one under the folder `application/templates/`
+3. The default export of this library has to be an object that extends the `ApplicationTemplate` interface
```
stefanhar, line 11: This `README` seems to contain text from the reference template, consider updating or removing it

libs/application/templates/driving-assessment-approval/jest.config.js
```diff
@@ -0,0 +1,13 @@
...
+  transform: {
+    '^.+\\.[tj]sx?$': [
+      'babel-jest',
+      { cwd: __dirname, configFile: './babel-jest.config.json' },
+    ],
+  },
+  moduleFileExtensions: ['ts', 'tsx', 'js', 'jsx'],
+  coverageDirectory:
+    '../../../../coverage/libs/application/templates/reference-template',
+  displayName: 'application-templates-reference-template',
```
stefanhar, line 12:
```suggestion
  displayName: 'application-template-driving-assessment-approval',
```

libs/application/templates/driving-assessment-approval/jest.config.js
```diff
@@ -0,0 +1,13 @@
...
+  preset: '../../../../jest.preset.js',
+  transform: {
+    '^.+\\.[tj]sx?$': [
+      'babel-jest',
+      { cwd: __dirname, configFile: './babel-jest.config.json' },
+    ],
+  },
+  moduleFileExtensions: ['ts', 'tsx', 'js', 'jsx'],
+  coverageDirectory:
+    '../../../../coverage/libs/application/templates/reference-template',
```
stefanhar, line 11:
```suggestion
    '../../../../coverage/libs/application/templates/driving-assessment-approval',
```

libs/application/templates/driving-assessment-approval/src/fields/CustomRepeater.tsx
```diff
@@ -0,0 +1,19 @@
...
+        easily implementing repetitive flows. The `expandRepeater` prop is
+        important as it initiates that subflow to generate more nested items.
+      </Text>
+
+      <Button onClick={expandRepeater}>Expand repeater</Button>
+    </>
+  )
+}
+
+export default CustomRepeater
```
stefanhar, line 19: This file seem not to be in use, consider removing it

libs/application/templates/driving-assessment-approval/src/fields/StudentLookupField.tsx
```diff
@@ -0,0 +1,67 @@
...
+  const { data = {}, error: queryError, loading } = useQuery(QUERY, {
+    skip:
+      !studentNationalId || !kennitala.isPerson(studentNationalId as string),
+    variables: {
+      nationalId: studentNationalId,
+    },
+  })
+
+  if (queryError) {
+    return <Text>Villa kom upp við að sækja upplýsingar um nemanda</Text>
```
stefanhar, line 36: Consider moving this string to messages to support multiple languages.
This applies to all unlocalized strings

libs/application/templates/driving-assessment-approval/src/lib/DrivingAssessmentApprovalTemplate.ts
```diff
@@ -0,0 +1,116 @@
...
+                ),
+              read: 'all',
+            },
+          ],
+        },
+        type: 'final' as const,
+      },
+    },
+  },
+  mapUserToRole: () => Roles.TEACHER,
```
stefanhar, line 113: This seems to cause everyone who opens this application to be mapped to the teacher role.
Is access control handled in some other way to prevent non teachers from filling out this application?

libs/application/templates/driving-license/src/dataProviders/EligibilityProvider.ts
```diff
@@ -0,0 +1,75 @@
...
+  SuccessfulDataProviderResult,
+  FailedDataProviderResult,
+} from '@island.is/application/core'
+
+const extractReason = (
+  requirements: ApplicationEligibilityRequirement[],
+): string => {
+  return requirements
+    .filter(({ requirementMet }) => !requirementMet)
+    .map(({ key }) => key)
```
stefanhar, line 14: If these reasons are displayed to the user we should consider mapping them to values in messages to support multiple languages

libs/application/templates/driving-license/src/fields/Congratulations.tsx
```diff
@@ -24,23 +20,39 @@ function Congratulations({
...
+              <Text>
+                Næsta sem gerist er að við ákveðum hvað á að standa í þessu
+                boxi.
+              </Text>
+            </AccordionItem>
+          </Accordion>
+        </Box>
+
+        <Box marginTop={4}>
+          <img src="/assets/images/movingTruck.svg" alt="Skrautmynd" />
```
stefanhar, line 54: Consider removing text from `alt` as the image is only decorative also adding `role="presentation"` would make this more clear to e-readers

yarn.lock
```diff
@@ -21418,7 +21418,7 @@ postcss-value-parser@^4.0.0, postcss-value-parser@^4.0.2, postcss-value-parser@^
   resolved "https://registry.yarnpkg.com/postcss-value-parser/-/postcss-value-parser-4.1.0.tgz#443f6a20ced6481a2bda4fa8532a6e55d789a2cb"
   integrity sha512-97DXOFbQJhk71ne5/Mt6cOu6yxsSfM0QGQyl0L25Gca4yGWEGJaig7l7gbCX623VqTBNGLRLaVUCnNkcedlRSQ==
 
-postcss@6.0.1, postcss@7.0.21, postcss@7.0.32, postcss@^6.0.1, postcss@^7.0.0, postcss@^7.0.1, postcss@^7.0.11, postcss@^7.0.14, postcss@^7.0.17, postcss@^7.0.18, postcss@^7.0.23, postcss@^7.0.26, postcss@^7.0.27, postcss@^7.0.32, postcss@^7.0.5, postcss@^7.0.6, "postcss@npm:@island.is/postcss@107.0.37":
+postcss@6.0.1, postcss@7.0.21, postcss@7.0.32, postcss@^6.0.1, postcss@^7.0.1, postcss@^7.0.11, postcss@^7.0.17, postcss@^7.0.18, postcss@^7.0.23, postcss@^7.0.26, postcss@^7.0.27, postcss@^7.0.32, postcss@^7.0.5, postcss@^7.0.6, "postcss@npm:@island.is/postcss@107.0.37":
```
stefanhar, line 21421: Was the update to this file intentional since i see no changes to `package.json`?

stefanhar (changes requested):

libs/api/domains/driving-license/src/lib/drivingLicense.service.ts
```diff
@@ -1,21 +1,32 @@
...
+import {
+  ApplicationEligibilityResponse,
+  DrivingAssessmentDto,
+  NewDrivingAssessmentResponse,
+  NewDrivingLicenseInput,
+  NewDrivingLicenseResponse,
+} from './client/drivingLicense.type'
+import { DAYS, DRIVING_ASSESSMENT_MAX_AGE } from './util/constants'
+
+const DEV_ONLY_REMOVE_ME_SKIP_CHECKS = true
```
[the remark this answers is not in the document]
arnorhs: úps..

stefanhar (approved):

libs/api/domains/driving-license/src/lib/graphql/main.resolver.ts
arnorhs: yes, this needs a closer look as a followup

libs/api/domains/driving-license/src/lib/graphql/main.resolver.ts
```diff
@@ -49,4 +48,71 @@ export class MainResolver {
...
+  @Query(() => [Juristiction])
+  drivingLicenseListOfJuristictions() {
+    return this.drivingLicenseService.getListOfJuristictions()
+  }
+
+  @Mutation(() => NewDrivingAssessmentResult)
+  async drivingLicenseNewDrivingAssessment(
+    @Args('input') input: NewDrivingAssessmentInput,
+    @CurrentUser() user: User,
+  ): Promise<NewDrivingAssessmentResult> {
```
eirikurn, line 86: Should all of these queries and mutations be publicly accessible?

Unfortunately the current data provider setup in the application system "encourages" adding everything to the graphql API (since dataProviders and template modules both have a utility to send graphql requests). But sensitive/internal APIs should not be in the graphql API. Instead, we should refactor integrations/clients to `libs/clients` and expose some stuff in graphql, but call other stuff directly from the application system template-module.

Template modules can get other NestJS modules injected. We have on the roadmap to support the same in dataProviders.

libs/api/domains/driving-license/src/lib/graphql/main.resolver.ts
```diff
@@ -49,4 +48,71 @@ export class MainResolver {
...
+    return {
+      success: response.ok,
+    }
+  }
+
+  @Mutation(() => NewDrivingLicenseResult)
+  async drivingLicenseNewDrivingLicense(
+    @Args('input') input: NewDrivingLicenseInput,
+    @CurrentUser() user: User,
+  ): Promise<NewDrivingLicenseResult> {
```
eirikurn, line 101: Same here.

libs/api/domains/driving-license/src/lib/client/drivingLicense.api.ts
```diff
@@ -23,13 +32,29 @@ export class DrivingLicenseApi {
...
-        SECRET: this.secret,
-        Accept: 'application/json',
+        ...this.headers(),
+        'Content-Type': 'application/json',
       },
+      method: 'POST',
+      body: JSON.stringify(body),
+    })
+    return res.json()
+  }
```
eirikurn, line 53: Are there OpenAPI/Swagger docs for this api? Would be nice to use a generated client for consistency.

libs/application/core/src/types/ApplicationTypes.ts
eirikurn: I think a single "extract-strings" command needs to find all strings belonging to a namespace, so depending on how they're, it might be more appropriate to have separate namespaces for each template.

jeremybarbet: Regarding messages with the same key, if the message has already been synced once with Contentful, it won't override the Icelandic and English translations. However it will change the defaultMessage and Description. Also, these two objects might expect different messages for the 2 different templates, but only one of them will live, which is the one with the latest changes on Contentful.

I don't know how much each of these templates are sharing messages, but it's easier to have 1 namespace by the template to avoid conflict and some surprise when checking the app and finding the wrong copy from another template

libs/application/core/src/types/StateMachine.ts
```diff
@@ -19,6 +19,7 @@ export enum DefaultEvents {
   SUBMIT = 'SUBMIT',
   ABORT = 'ABORT',
   EDIT = 'EDIT',
+  PAYMENT = 'PAYMENT',
```
jeremybarbet, line 22: Not a big deal, but not sure it should be part of the default state but rather a custom state for your template instead

eirikurn (changes requested):
Probably needs to refactor the driving license domain/client to protect sensitive endpoints.

Otherwise LGTM.

libs/application/templates/driving-license/src/fields/HealthDeclaration.tsx
```diff
@@ -46,6 +46,7 @@ function HealthDeclaration({
           <RadioController
             id={field.id}
             split="1/2"
+            largeButtons={false}
```
jeremybarbet, line 49: Based on the application design system, it should always be true cf: https://www.figma.com/file/ine6cGn7cnrJJK43fzUZTF/Templates-%2F-h%C3%B6nnunarkerfi-fyrir-ums%C3%B3knir?node-id=2%3A6045

libs/application/templates/driving-license/src/lib/messages.ts
```diff
@@ -39,281 +44,288 @@ export const m = defineMessages({
       'Information from the National Registry will be used to prefill the data in the application',
   },
   userProfileInformationTitle: {
-    id: 'pl.application:userprofile.title',
+    id: 'dl.application:userprofile.title',
```
jeremybarbet, line 47: thanks for fixing that 👍

libs/api/domains/driving-license/src/lib/graphql/main.resolver.ts
arnorhs: ah, i see. yes, we definitely don't want them publicly accessible.

ok, so you think i should add the driver's license nest module as a dependency of the api template, instead of calling the graphql API?

And then think through whether or not to expose all those endopints to graphql. Makes sense, if I understand correctly

libs/api/domains/driving-license/src/lib/client/drivingLicense.api.ts
arnorhs: there is - i haven't looked into how to generate the client from those.. i had an attempt at it but since parts of it were also already implemented, and i was stilling getting the hang of things i abandoned it.

but i'd be happy to do so as a followup as well

libs/api/domains/driving-license/src/lib/graphql/main.resolver.ts
eirikurn: Yes exactly.

libs/application/core/src/types/StateMachine.ts
arnorhs: yes, sorry, that's actually leftovers. good catch

libs/application/templates/driving-license/src/fields/HealthDeclaration.tsx
arnorhs: i see.. it looked super weird when i didn't change it after your change to the default - perhaps i was missing some props.

so this is based on a design that needs to be updated then.

arnorhs: Somebody pointed out that those are not very suitable for yes/no questions.. you are talking about these, correct?

image

libs/api/domains/driving-license/src/lib/graphql/main.resolver.ts
arnorhs: ok, i updated the PR with this change, if I understood :)

libs/application/templates/driving-license/src/dataProviders/EligibilityProvider.ts
```diff
@@ -4,29 +4,36 @@ import {
...
-    .map((key) => requirementKeyToMessage(key))
+    .map((key) => requirementKeyToMessageId(key))
     .join(', ')
 }
 
-const requirementKeyToMessage = (key: string) => {
+// TODO: we need a better way of getting the translated string in here, outside
+// of react. Possibly we should just make a more flexible results screen.
+// This string ends up being used as the paramejter displayed as the error message
+// for the failed dataprovider
```
jeremybarbet, line 26: I didn't look where it was used but according to the description it's regarding the data provider reject reason, and it takes a StaticText argument as seen here https://github.com/island-is/island.is/blob/main/libs/application/templates/parental-leave/src/dataProviders/Children/Children.ts#L214-L216

arnorhs: yes, however, i'm returning multiple messages / reasons for this check.

libs/application/templates/driving-license/src/fields/HealthDeclaration.tsx
arnorhs: yes, this feels like something for the designers to argue about. I'll tell Orvar and he can update the design or bring it up with the other designers.

Thanks for pointing this out, but Arndis, my skilastjóri doesn't want to have this changed right now.

libs/application/templates/driving-license/src/dataProviders/EligibilityProvider.ts
arnorhs: So i'd need to have the display logic handle multiple reasons.

it would be nice to add that as a feature of the data provider UI - perhaps even style with bullet points. I could do that as a followup

eirikurn (approved):
Thanks for fixing this in record time 🙏👍🏻

Generally api domains should only be imported by the api app. We can refactor reused integration code into separate client libs. But we can do that later in a follow up.

PR #3942, opened 2021-05-24
fix(parental-leave): Show timeline in employer approval state

madebynoam:
Fixes the timeline not appearing in employer-approval state:
https://app.asana.com/0/1182378413629561/1200340285569551/f

## Checklist:


libs/application/templates/parental-leave/src/lib/ParentalLeaveTemplate.ts
```diff
@@ -245,6 +245,7 @@ const ParentalLeaveTemplate: ApplicationTemplate<
                   Promise.resolve(val.EmployerApproval),
                 ),
               read: { answers: ['periods'], externalData: ['pregnancyStatus'] },
+              write: 'all',
```
baering, line 248: We should not give `ASSIGNEE` a `write: 'all'` permission.

baering, line 248: We should not give `ASSIGNEE` a `write: 'all'` permission.

baering (changes requested):

libs/application/templates/parental-leave/src/lib/ParentalLeaveTemplate.ts
madebynoam: True, was using it as hotfix 🙃. Will investigate further.

madebynoam: I investigated it further, and it was related to the new structure for `selectedChild` and `children` externalData. 
The latest commit should fix it more cleanly. I'll take a look if we can have a test for this, otherwise should be ready for re-review.

jeremybarbet (approved):
Nice, good catch, LGTM

baering (approved):

PR #3944, opened 2021-05-25
feat(service-portal): Implement delegation switcher

eirikurn:
## What

Allow user to switch between delegations after signing in.

This PR also:

- Makes libs/feature-flags more enjoyable to use in react projects, with `useFeatureFlag` hook and default user logic.
- Makes TopicCard clickable without having a link.
- Tweaks auth-public-api endpoints and scopes.
- Creates a auth-public-api client library.
- Adds a function to `useAuth` to switch to another user.
- Adds react-apollo codegen to `service-portal/graphql`.

## Why

To support a better UX for delegations.

## Screenshots / Gifs

https://user-images.githubusercontent.com/115094/119511597-1c0d2400-bd62-11eb-870d-78e946a80d5c.mov

## Checklist:

Other dependencies and follow-up tasks:

- [x] Update helm environment for API so it can find auth-public-api.
- [x] Create new auth scopes in IDS environments.
- [ ] Fix "create scope" regression in IDS around the new scope group feature.
- [ ] Add actor claims to id token and/or userinfo endpoint.
- [ ] Look into client "prompt delegation" setting. Does it work as it should?
- [ ] Fix IDS so the actor-delegations scope stays around when using a delegation.


apps/service-portal/src/components/UserDelegations/UserDelegations.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+  let delegations = [] as ActorDelegationsQuery['authActorDelegations']
+  if (data) {
+    delegations.push(...data.authActorDelegations)
+  }
+  if (isDelegation) {
+    delegations = delegations.filter(
+      (delegation) => delegation.fromNationalId !== currentNationalId,
+    )
+    delegations.unshift({ fromName: 'Ég', fromNationalId: actorNationalId })
```
darrikonn, line 56: Don't we want translations for `Ég`?

apps/service-portal/src/components/UserMenu/UserMenu.tsx
```diff
@@ -45,57 +51,60 @@ const UserMenu = ({ state, onClose, onRouteChange }: UserMenuProps) => {
...
+          <Box>
+            <Box
+              display="flex"
+              alignItems="center"
+              marginBottom={1}
+              className={styles.breakWord}
+            >
+              <Box display="flex" alignItems="center" marginRight={2}>
+                <Icon type="outline" icon="mail" color="blue300" />
+              </Box>
```
darrikonn, line 64: FYI: the `<Inline />` component is good for this.

libs/api/domains/auth/src/lib/models/delegation.model.ts
```diff
@@ -0,0 +1,70 @@
...
+      case DelegationType.LegalGuardian:
+        return LegalGuardianDelegation
+      case DelegationType.ProcurationHolder:
+        return ProcuringHolderDelegation
+      case DelegationType.Custom:
+        return CustomDelegation
+    }
+  },
+})
+export abstract class Delegation {
```
darrikonn, line 28: Would like to see the id fields here as well (which exists on the delegation)

libs/auth-api-lib/src/lib/entities/models/delegation.model.ts
```diff
@@ -56,13 +55,27 @@ export class Delegation extends Model<Delegation> {
...
   readonly created!: Date
 
   @UpdatedAt
-  @ApiProperty()
   readonly modified?: Date
 
   @HasMany(() => DelegationScope)
   delegationScopes?: DelegationScope[]
+
+  toDTO(): DelegationDTO {
```
darrikonn, line 66: Like this approach :+1:

libs/auth-api-lib/src/lib/services/delegations.service.ts
```diff
@@ -82,24 +87,29 @@ export class DelegationsService {
           { validTo: { [Op.or]: [{ [Op.eq]: null }, { [Op.gt]: now }] } },
         ],
       },
+      include: [DelegationScope],
     })
 
     return result.map(
       (d) =>
```
darrikonn, line 94: You can just return `result.map(d => d.toDTO())`

libs/auth-api-lib/src/lib/services/delegations.service.ts
```diff
@@ -137,35 +147,38 @@ export class DelegationsService {
...
+  async findAllCustomFrom(nationalId: string): Promise<DelegationDTO[] | null> {
     this.logger.debug(`Finding a delegation for nationalId ${nationalId}`)
-    return await this.delegationModel.findAll({
+    const delegations = await this.delegationModel.findAll({
       where: {
         fromNationalId: nationalId,
       },
       include: [DelegationScope],
     })
+    return delegations.map((delegation) => delegation.toDTO())
```
darrikonn, line 181: Shouldn't the controller be the one that maps the entity (and strip unwanted data) to DTO, since internally (i.e. between controller to service) you can just work with the original object?

libs/service-portal/graphql/src/lib/queries/getActorDelegations.ts
```diff
@@ -0,0 +1,10 @@
+import { gql } from '@apollo/client'
+
+export const ACTOR_DELEGATIONS = gql`
+  query ActorDelegations {
+    authActorDelegations {
```
darrikonn, line 5: Query the `id` as well for the apollo store.

We should just add the
```
graphql/required-fields
```
eslint rule where it's required to query the ID of a type.
If the ID isn't queried, the apollo store doesn't know how to construct its objects based on the data already in the cache.

darrikonn (changes requested):
Nice work. Minor comments.

apps/service-portal/src/components/UserDelegations/UserDelegations.tsx
eirikurn: Ah, forgot to explain this.

This is temporary until we get the name of the real user in the id token or userinfo endpoint.

apps/service-portal/src/components/UserMenu/UserMenu.tsx
eirikurn: Just indenting. 👍

libs/api/domains/auth/src/lib/models/delegation.model.ts
eirikurn: We only have an id field for database (custom) delegations. The legal guardian and procuring holder delegations don't have an id, unless we make one up.

libs/auth-api-lib/src/lib/services/delegations.service.ts
eirikurn: Good catch, forgot this one.

libs/auth-api-lib/src/lib/services/delegations.service.ts
eirikurn: Good question. Opinions differ on how far up DTOs should go and how far down connected ORM objects should go. E.g. we commonly pass input DTO's to the service.

In this case, we're sharing this service between two different APIs (and corresponding controllers) and want to reuse as much implementation as possible between them. The DTO object provides a fine abstraction for the three different delegation sources returned by this service.

apps/service-portal/src/components/MobileMenu/MobileMenu.tsx
```diff
@@ -25,6 +28,10 @@ const MobileMenu: FC<{}> = () => {
   const navigation = useNavigation()
   const { changeLanguage } = useNamespaces()
   const { signOut } = useAuth()
+  const showDelegations = useFeatureFlag(
+    'identityserverDelegationsEnabled',
```
atligudl, line 32: Maybe we should set up an Enum for these flags

unnarb (approved):

libs/service-portal/graphql/src/lib/queries/getActorDelegations.ts
eirikurn: Yeah, the access control pages use different auth-public-api endpoints that only manage custom delegations. The GraphQL layer should return CustomDelegations that have the non-null id field.

atligudl (approved):
Looks good, I feel like the UI in the user menu can be improved upon in the future

libs/service-portal/graphql/src/lib/queries/getActorDelegations.ts
eirikurn: Also, this queries "actor delegations" while the access control queries "subject delegations". These should not intersect.

eirikurn:
@atligudl

Yeah I agree. This is the "design system"-following compromise between the designers (though the TopicCard should be a bit smaller on desktop). I preferred a more google-like custom design, but I assume we'll iterate on this.

darrikonn (approved):

PR #3945, opened 2021-05-25
fix(core): security audit multicast-dns and browserslist

VilliGunn:
# Security audit fix

This should get rid of the build error but I don't know if it will impact `webpack-dev-server` in any apps or libraries. I have only verified that the `application-system` works fine.

https://github.com/webpack/webpack-dev-server/issues/3340#issuecomment-847659624

## What

Fix security issues with `multicast-dns`


sindrig (approved):
There are literally no release notes I can find for multicast-dns so I have no idea if this breaks anything.

PR #3951, opened 2021-05-25
fix(parental-leave): switch to radio buttons without days select for now

jeremybarbet:
## What

Everything in the title

Screenshot 2021-05-25 at 10 25 28
Screenshot 2021-05-25 at 10 25 22
Screenshot 2021-05-25 at 10 25 15
Screenshot 2021-05-25 at 10 25 07


baering (approved):
We'll need to make sure that only a primary parent receives these questions, can do that in another PR.

PR #3968, opened 2021-05-26
feat(parental-leave): plug all select and radio controller and refactor stateful logic

jeremybarbet:
## What

- Bunch of fields were not plugged to be editable.
- Refactor stateful logic to make it on all fields retrieved from answers by default
- Fix mobile layout


libs/application/templates/parental-leave/src/hooks/useEstimatedPayments.ts
```diff
@@ -0,0 +1,24 @@
...
+            ratio: 100,
+            approved: true,
+            paid: true,
+          },
+        ],
+      },
+    },
+  })
+  */
+}
```
saevarma, line 24: Maybe this file should be removed completely while not in use.
Than added later when this is implemented again.

saevarma (changes requested):
Looks good, just one minor feedback.

libs/application/templates/parental-leave/src/hooks/useEstimatedPayments.ts
jeremybarbet: Yes I was not sure, but you are right, and the API might might be different as well so I will do that 👍

saevarma (approved):
Nice!

libs/application/templates/parental-leave/src/hooks/useStatefulAnswers.ts
```diff
@@ -0,0 +1,16 @@
...
+} from './useApplicationAnswers'
+
+export const useStatefulAnswers = (
+  application: Application,
+): [ApplicationAnswers, Dispatch<SetStateAction<ApplicationAnswers>>] => {
+  const answers = useApplicationAnswers(application)
+  const [stateful, setStateful] = useState(answers)
+
+  return [stateful, setStateful]
+}
```
baering, line 16: Not used anywhere?

baering (commented):
LGTM.

`useStatefulAnswers` does not seem to be used anywhere.

libs/application/templates/parental-leave/src/hooks/useStatefulAnswers.ts
jeremybarbet: Yes, it's used in the big file diff that's being minimized :D 
Screenshot 2021-05-31 at 10 15 06

madebynoam (approved):

libs/application/templates/parental-leave/src/hooks/useStatefulAnswers.ts
baering: 🙈

PR #3970, opened 2021-05-26
feat(regulations): various tweaks to regulations ui

Valur:
## What

Various changes and tweaks to regulations ui after design and client review
See: https://regulationsrelease-beta.dev01.devland.is/reglugerdir

## Checklist:


package.json
```diff
@@ -64,6 +64,7 @@
     "@formatjs/intl-numberformat": "6.2.1",
     "@formatjs/intl-utils": "3.8.4",
     "@hookform/error-message": "0.0.5",
+    "@hugsmidjan/regulations-editor": "0.2.0",
```
eirikurn, line 67: Shouldn't this be open source under the same terms as other Digital Iceland work? I couldn't find the source.

I assume it will be continuously updated, which doesn't fit great with our workflow and CI. Should we try to get it into the monorepo?

yarn.lock
```diff
@@ -2924,6 +2924,20 @@
...
+  version "0.2.0"
+  resolved "https://registry.yarnpkg.com/@hugsmidjan/regulations-editor/-/regulations-editor-0.2.0.tgz#92e4ad834bf047206e9d5a8b4a6975e572502262"
+  integrity sha512-LhUnJH8+g30YE8lVCsXdI20opol9rPQfl7bT//PtLpP8ycWzwam+F0Gna0Dr76+khElAx9WMLkUfchPh9HXNzw==
+  dependencies:
+    "@tinymce/tinymce-react" "^3.8.4"
+    cleanup-prettier "npm:prettier@2.2.0"
+    htmldiff-js "^1.0.5"
+    jsdom "^16.4.0"
+    lodash "^4.17.19"
+    qj "npm:@hugsmidjan/qj@^4.6.1"
```
eirikurn, line 2937: Why is `@hugsmidjan/qj` aliased as `qj`?

yarn.lock
```diff
@@ -2924,6 +2924,20 @@
   resolved "https://registry.yarnpkg.com/@hookform/error-message/-/error-message-0.0.5.tgz#606912b506513669980c440f383ff6c78a7aca3a"
   integrity sha512-es7eLLFA3SXNYAT8aUjvf7Gok1eMHK+9DMILtJA7ZEwYZlCCCPifhpoZmY+5SOopEtF7e+qxFOjX+MJnI3jOcg==
 
+"@hugsmidjan/regulations-editor@0.2.0":
+  version "0.2.0"
+  resolved "https://registry.yarnpkg.com/@hugsmidjan/regulations-editor/-/regulations-editor-0.2.0.tgz#92e4ad834bf047206e9d5a8b4a6975e572502262"
+  integrity sha512-LhUnJH8+g30YE8lVCsXdI20opol9rPQfl7bT//PtLpP8ycWzwam+F0Gna0Dr76+khElAx9WMLkUfchPh9HXNzw==
+  dependencies:
+    "@tinymce/tinymce-react" "^3.8.4"
+    cleanup-prettier "npm:prettier@2.2.0"
```
eirikurn, line 2933: If we were to include the regulations-editor in the monorepo we should probably alias this as `@island.is/regulations-prettier`.

apps/web/components/Regulations/RegulationDisplay.treat.ts
```diff
@@ -26,193 +80,7 @@ export const upcomingWarning = style({
...
-  {
-    marginTop: '0',
-  },
-)
-global('.article__name', {
-  display: 'block',
-  fontStyle: 'italic',
-})
+regulationContentStyling(bodyText)
+diffStyling(bodyText)
```
eirikurn, line 86: Why refactor just the styles into a library? I prefer colocating styles with components.

Maybe refactor the whole component into shared components?

Also, to create scoped selectors I prefer including them as part of the root style instead of using `globalStyle`:

```ts
export const bodyText = style({
  lineHeight: 28 / 18 + 'em',

  selectors: {
    '& .chapter__title': {
      marginTop: '2em',
      // ...
    },
    // Sure, a bit longer.
    '& p, & ul, & ol, & table, & blockquote': {
    },
    // Or use a helper here.
    [childSelector('p,ul,ol,table,blockquote')]: {
    },
  },
})
```

eirikurn (commented):
LGTM, minor comments.

apps/web/components/Regulations/RegulationDisplay.treat.ts
eirikurn: Sadly this does not have good enough browser support, and is not even supported in treat: `& :is(p, ul, ol, table, blockquote)`

maranomynet: hmm... Treat screams at me when I try this, and demands that the `&` be at the end of the selector (unless I'm sticking attribute-selectors up against the `&`).

It seemed to me that globalStyle was the only way to go.

image

yarn.lock
maranomynet: That's a mistake. Should have removed the alias before refactoring into a published lib.

apps/web/components/Regulations/RegulationDisplay.treat.ts
maranomynet: The `@hugsmidjan/regulations-editor` components are code-shared with another project (the regulation database migration, text cleanup and change-history insertion system) and that project does not use treat.

Thus we de-couple the styling library from the component, and define the available/required class-names as a TS interface.

package.json
maranomynet: No secrecy or closed-sourceness intended.

This code resides in a repo for a standalone tool we've already developed for cleaning up the texts of the old Regulation database. That tool has a very limited lifetime and it would thus not be productive try and bring that whole mess into the monorepo.

However, there are significant bits of code/logic that we **need** to code-share between that tool and the regulation-admin system, at the moment.

...and code-sharing seems easier in this direction, rather than trying to make the cleanup-tool somehow reach into the monorepo for the bits it needs.

My plan was to, once things have settled down a bit more, to split this module into its own public repo, and publish it under the @island.is npm scope.

Alternatively, we _could_ inline it in the monorepo when the time comes. I'm completely neutral on that, but I'd prefer to keep this code outside the monorepo for now.

apps/web/components/Regulations/RegulationDisplay.treat.ts
maranomynet: Rant: Why! oh why! does Treat not support defining CSS custom properties!?

maranomynet (approved):

yarn.lock
Valur: Fixed, alias removed

libs/regulations/README.md
```diff
@@ -0,0 +1,7 @@
+# Regulations
+
+This library was generated with [Nx](https://nx.dev).
+
+## Running unit tests
+
+Run `nx test regulations` to execute the unit tests via [Jest](https://jestjs.io).
```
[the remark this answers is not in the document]
Valur: Description added

Valur:
This was merged here: https://github.com/island-is/island.is/pull/4674

PR #3971, opened 2021-05-26
fix(parental-leave): don't show other parent discount screen when not applicable

jeremybarbet:
## What

- Only show the discount section when the other parent is the spouse or if manual with approvable to share information about the parental leave


libs/application/templates/parental-leave/src/parentalLeaveUtils.ts
```diff
@@ -143,6 +143,8 @@ export const getAvailableRightsInMonths = (application: Application) => {
 }
 
 export const getOtherParentOptions = (application: Application) => {
+  console.log('-application.externalData', application.externalData)
```
baering, line 146: 💩

libs/application/templates/parental-leave/src/forms/ParentalLeaveForm.ts
```diff
@@ -339,10 +340,10 @@ export const ParentalLeaveForm: Form = buildForm({
...
               id: 'usePersonalAllowanceFromSpouse',
               title: parentalLeaveFormMessages.personalAllowance.useFromSpouse,
-              condition: (answers) => {
-                // TODO add check if this person has a spouse...
-                return true
-              },
+              condition: (answers) =>
+                answers.otherParent === SPOUSE ||
+                (answers.otherParent === MANUAL &&
+                  answers.otherParentRightOfAccess === YES),
```
baering, line 346: We'll need to give `usePersonalAllowanceFromSpouse` a default value in the cases where it is skipped, since undefined will [result in an error when transforming/serializing the application](https://github.com/island-is/island.is/blob/main/libs/application/template-api-modules/src/lib/modules/templates/parental-leave/parental-leave.utils.ts#L23) to API payload, or refactor the [transformation/serialization code](https://github.com/island-is/island.is/blob/main/libs/application/template-api-modules/src/lib/modules/templates/parental-leave/parental-leave.utils.ts#L47).

baering (commented):
Looks good.

I realised we _may_ run into issues when the answer to the question being skipped is undefined.

libs/application/templates/parental-leave/src/forms/ParentalLeaveForm.ts
jeremybarbet: Good catch, I will fix that, thanks 👍

baering (approved):

PR #3975, opened 2021-05-26
fix(web): Add missing TOC field to article importer

Ingimarsson:
# Add missing TOC field to article importer

## What

This fixes table of contents not appearing on baby articles. The article importer was not including the `showTableOfContents` field so it defaulted to false.

## Why

It's a bug.

## Checklist:


kksteini (approved):
Good catch :+1:

PR #3978, opened 2021-05-26
fix(web): Stafraent Island page - minor fixes

Ingimarsson:
# Stafraent Island page - minor fixes

## What

Small fixes on Stafrænt Ísland page

- Added intro field to timeline
- Links on `Overview Links` slice now open in a new tab (reqested by SÍ)
- Green focus border in menu doesn't get cut off under main content
- Make timeline arrows disabled when it reaches either end

## Screenshots / Gifs

Only visual change is the heading and intro on the timeline slice. 

image

## Checklist:


apps/web/components/Organization/Slice/OverviewLinks/OverviewLinks.tsx
```diff
@@ -76,16 +77,17 @@ export const OverviewLinksSlice: React.FC<SliceProps> = ({ slice }) => {
                               )}{' '}
                             </Box>
                           )}
-                          <Link
+                          <a
```
kksteini, line 80: Why the need for switching to an `a` tag instead of Link?
Are these internal island.is links that are supposed to open as a new window?

If not then they should open up externally automatically from the else statement of the `isInternal` block:
https://github.com/island-is/island.is/blob/71eb2514ef3bb2d012d0050e852ae2dd42dd0e65/libs/island-ui/core/src/lib/Link/Link.tsx#L88-L92

Ingimarsson: Yes, we got a request from Digital Iceland to make them open in a new tab. The `Link` component does not take a `target` property so I thought this was the simplest method.

kksteini: :+1:

kksteini: Could, and should, we create a prop on the Link component `libs/island-ui/core/src/lib/Link/Link.tsx` called `newTab`?

Then in the relevant lines for internal:
https://github.com/island-is/island.is/blob/71eb2514ef3bb2d012d0050e852ae2dd42dd0e65/libs/island-ui/core/src/lib/Link/Link.tsx#L78-L82

Adding:
```diff
            className={classNames}
            {...linkProps}
+           {...(newTab && {target: ‘_blank’})}
            tabIndex={skipTab ? -1 : undefined}
```

kksteini (approved):
LGTM - suggest modifying LINK in the component library

apps/web/components/Organization/Slice/OverviewLinks/OverviewLinks.tsx
Ingimarsson: :+1:

libs/island-ui/core/src/lib/Link/Link.tsx
```diff
@@ -78,6 +80,7 @@ export const Link: React.FC<LinkProps> = ({
           <a
             className={classNames}
             {...linkProps}
+            {...(newTab && { target: '_blank' })}
```
jeremybarbet, line 83: I would use it along `rel="noopener noreferrer"`

Ingimarsson: Is it useful for internal links that open in a new tab? This link is within a `  if (isInternal) ...` block. External links already get `rel="noopener noreferrer"`.

jeremybarbet: Good point, didn't see it was for internal links 👍

jeremybarbet (approved):

PR #3692, opened 2021-05-03
fix(parental-leave): Fix duration slider dates don't reflect half months

madebynoam:
Fixes the duration slider to reflect half months and to use 30 day months instead of calendar months.
* https://app.asana.com/0/1182378413629561/1200216819967902/f
* https://app.asana.com/0/1182378413629561/1200214178491309/f

## Screenshots / Gifs
image

## Checklist:

- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] Formatting passes locally with my changes
- [x] I have rebased against main before asking for a review

PR #3980, opened 2021-05-26
fix(parental-leave): plug job ratio to use max from periods choose, fix actual dob logic

jeremybarbet:
## What

- use percentage calculated on the period to define the max percentage usage possible
- hide periods when using actual dob and display a duration in months instead

Actual DOB:
Screenshot 2021-05-25 at 15 04 23
Screenshot 2021-05-25 at 15 04 30

Expect DOB/Specifc Date:
Screenshot 2021-05-25 at 15 05 03
Screenshot 2021-05-25 at 15 04 56

Video showing job ratio usage under:

https://user-images.githubusercontent.com/937328/119524009-c38f5400-bd6c-11eb-9ae2-488242e388f6.mov


libs/application/templates/parental-leave/src/fields/Duration/index.tsx
```diff
@@ -146,29 +142,36 @@ const Duration: FC<FieldBaseProps> = ({ field, application }) => {
...
+                }
                 currentIndex={chosenDuration}
                 onChange={(selectedMonths: number) => {
                   clearErrors(id)
-                  const newEndDate = addMonths(
-                    parseISO(currentStartDateAnswer),
-                    selectedMonths,
+
+                  const newEndDate = formatISO(
+                    addMonths(parseISO(currentStartDateAnswer), selectedMonths),
```
madebynoam, line 170: For future PR: 
I think we might still have a bug here, where it won't reflect half months, right? 
Since `addMonths` floors the value `Positive decimals will be rounded using Math.floor`
Related:
https://github.com/island-is/island.is/pull/3692/files

jeremybarbet: Yes, we will update everything at once when VMST has finalized the API to calculate all the dates

jeremybarbet: It's part of this ticket and all the tickets that depends on it https://app.asana.com/0/1182378413629561/1200358543222708

madebynoam (approved):
Just one small comment, otherwise lgtm!

PR #3997, opened 2021-05-27
feat(application-system): application list

VilliGunn:
# ...

## What

## Why

## Screenshots / Gifs

## Checklist:

### DIFF
--- a/apps/application-system/form/src/routes/Applications.tsx
+++ b/apps/application-system/form/src/routes/Applications.tsx
@@ -117,10 +117,9 @@ export const Applications: FC = () => {
               {(data?.applicationApplications ?? []).map(
                 (application: Application, index: number) => {
                   const isCompleted =
-                    application.status === ApplicationStatus.COMPLETED
+                    application.state === ApplicationStatus.COMPLETED
                   const isRejected =
-                    application.status === ApplicationStatus.REJECTED
-
+                    application.state === ApplicationStatus.REJECTED
                   return (
                     <ActionCard
                       key={`${application.id}-${index}`}
@@ -154,7 +153,7 @@ export const Applications: FC = () => {
                           history.push(`../${slug}/${application.id}`),
                       }}
                       progressMeter={{
-                        active: true,
+                        active: Boolean(application.progress),
                         progress: application.progress,
                         variant: isRejected
                           ? 'red'
--- a/apps/application-system/form/src/routes/Applications.tsx
+++ b/apps/application-system/form/src/routes/Applications.tsx
@@ -170,7 +170,6 @@ export const Applications: FC = () => {
             <Stack space={2}>
               {(data?.applicationApplications ?? []).map(
                 (application: Application, index: number) => {
-                  console.log('appli', application)
                   const stateMetaData = application.stateMetaData
                   const stateDefaultData =
                     ApplicationStateDisplayedData[application.status] ||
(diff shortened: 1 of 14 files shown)


libs/application/templates/family-matters/children-residence-change/src/lib/ChildrenResidenceChangeTemplate.ts
```diff
@@ -158,11 +160,15 @@ const ChildrenResidenceChangeTemplate: ApplicationTemplate<
           ],
         },
       },
-      [ApplicationStates.REJECTED]: {
+      [ApplicationStates.REJECTEDBYPARENTB]: {
         meta: {
           name: applicationName,
-          progress: 1,
+          description: stateDescriptions.rejectedByParentB,
```
jeremybarbet, line 166: To follow the same example as the lifecycle object, I would use an object to define theses fields. We are already in the meta object, so not sure what the name should be but something like

```
obj(state?,info?): {
  title: '',
  description: '',
  tag: {
    variant: '',
    label: '',
  }
}
```

apps/application-system/form/src/routes/Applications.tsx
```diff
@@ -32,6 +34,58 @@ import { dateFormat } from '@island.is/shared/constants'
...
+  }
+  progress: {
+    variant: 'blue' | 'red' | 'rose' | 'mint'
+  }
+  cta: {
+    label: MessageDescriptor
+  }
+}
+
+const ApplicationStateDisplayedData: Record<ApplicationStatus, StateData> = {
```
jeremybarbet, line 50: Would be nice to move that within the core library so it can be exported only once and imported in the form and service portal

VilliGunn: Yeah I just extracted this to a component that is used both here and in the service-portal

Thorrik58 (approved):

libs/application/ui-components/src/components/ApplicationList/ApplicationList.tsx
```diff
@@ -0,0 +1,123 @@
...
+        if (!slug) {
+          return null
+        }
+        return (
+          <ActionCard
+            key={`${application.id}-${index}`}
+            date={format(new Date(application.modified), formattedDate)}
+            tag={{
+              label: actionCard?.tag?.label
+                ? formatMessage(actionCard?.tag?.label)
```
jeremybarbet, line 95: `actionCard?.tag?.label` should be defined

```suggestion
                ? formatMessage(actionCard.tag.label)
```

libs/application/ui-components/src/components/ApplicationList/ApplicationList.tsx
```diff
@@ -0,0 +1,123 @@
...
+const ApplicationList = ({ applications, onClick }: Props) => {
+  const { lang: locale, formatMessage } = useLocale()
+  const formattedDate = locale === 'is' ? dateFormat.is : dateFormat.en
+
+  return (
+    <Stack space={2}>
+      {applications.map((application, index: number) => {
+        const actionCard = application.actionCard
+        const stateDefaultData =
+          DefaultData[application.status] ||
```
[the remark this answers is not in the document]
VilliGunn: I actually think the `||` operator is better here since we don't have any real scenario where the value of `DefaultData[application.status]` will be `0` or `''` 

It's definitely better in this example
`actionCard?.title ?? application.name`
Where the `actionCard?.title` could be set to an empty string intentionally and therefore we would not like to fallback to the `application.name`

I might be missing some case that is the reason for you going for `??` instead?

@jeremybarbet

jeremybarbet (approved):

baldur-kolibri (approved):
lgtm

PR #4002, opened 2021-05-28
chore(parental-leave): update to new API payload

jeremybarbet:
## What

- Update to new payload


libs/application/template-api-modules/src/lib/modules/templates/parental-leave/parental-leave.service.ts
```diff
@@ -65,18 +66,33 @@ export class ParentalLeaveService {
...
+    const isSelfEmployed =
+      getValueViaPath(application.answers, 'employer.isSelfEmployed') ===
+      formConstants.boolean.true
+
+    if (isSelfEmployed) {
+      const pdf = await this.getSelfEmployedPdf(application)
+
+      return [
+        {
+          attachmentType: 'selfEmployed',
```
baering, line 79: Shouldn't we store this in a enum/constant?

baering (commented):
LGTM, just one minor comment.

libs/application/template-api-modules/src/lib/modules/templates/parental-leave/parental-leave.service.ts
jeremybarbet: That's what I did first, but the VMST API has only a type string and when trying to use enum vs string it was conflicting. I could have it as an enum on our side, but still send a string to the API, I will change that

jeremybarbet: Reused the api constants object we already have for others VMST values [`7903a53` (#4002)](https://github.com/island-is/island.is/pull/4002/commits/7903a53a1a30eaccb3fe1b257432bdfaadb25d2c#diff-f4c7486e6c153b6ac589499b8728a5678955365f3c7a5610f3572a3223c42a4bR22)

baering (approved):

libs/application/template-api-modules/src/lib/modules/templates/parental-leave/parental-leave.service.ts
```diff
@@ -65,18 +66,33 @@ export class ParentalLeaveService {
     }
   }
 
+  async getAttachments(application: Application): Promise<Attachment[] | null> {
```
saevarma, line 69: I find it cleaner to have an empty array instead of null when there are no items. But that is just my preference :)

saevarma (approved):
LGTM

libs/application/template-api-modules/src/lib/modules/templates/parental-leave/parental-leave.service.ts
jeremybarbet: I followed your recommendation, looks better, thanks 👍

PR #4015, opened 2021-05-31
feat(gjafakort): using configcat to configure ferdagjof 2 on/off

arnorhs:
Using confgicat directly to toggle whether or not ferdagjof2 is enabled or not.

## Checklist:


apps/gjafakort/api/src/services/versionConfiguration.ts
```diff
@@ -17,9 +16,24 @@ const v2: Version = {
...
+const IS_FERDAGJOF_2_ENABLED = 'isferdagjof2enabled'
+
+const sdkKey = environment.configCat.sdkKey
+if (!sdkKey) {
+  throw new Error(
+    'Trying to initialize configcat client without configCat.sdkKey environment variable',
+  )
+}
+
+const configCat = eval('require')('configcat-node').createClient(sdkKey, {
```
pshomov, line 28: This could be problematic down the line with JS bundlers but since in this case is pretty short-lived, should be ok

pshomov (approved):
LGTM

apps/gjafakort/api/src/services/versionConfiguration.ts
arnorhs: @pshomov I agree

stjanilofts (approved):

PR #4017, opened 2021-05-31
feat(services-documents): Add audit logs

saevarma:
# Adding audit logs to the `services-documents`

## What

Add audit logs for the `services-documents` service.

## Why

To improve monitoring of access and user activity of our services.

## Checklist:


apps/services/documents/src/app/modules/document-provider/controllers/provider.controller.ts
```diff
@@ -6,38 +9,50 @@ import {
...
+  })
+  async getAllProviders(): Promise<Provider[]> {
+    return this.documentProviderService.getProviders()
   }
 
   @Get(':id')
   @ApiOkResponse({ type: Provider })
+  @Audit<Provider>({
+    resources: (provider) => provider?.id,
+  })
```
eirikurn, line 55: Can `provider` be null/undefined? (also below)

eirikurn (approved):
Awesome. Nice cleanup!

apps/services/documents/src/app/modules/document-provider/controllers/provider.controller.ts
saevarma: No, as these two actions check for a falsy object, and throw an error if it is not found. 
So I removed the optional operator.

PR #4021, opened 2021-05-31
fix(parental-leave): don't use title on action card for now

jeremybarbet:
## What

Keep having "Parental leave application" on all action card


madebynoam (approved):
lgtm!

baering (approved):

PR #4029, opened 2021-06-01
fix(serve): Temporary fix for nx serve for express apps

sindrig:


eirikurn (approved):

sindrig:
There are missing changes from a feature branch of esbuildnx missing here - need a new version

PR #4035, opened 2021-06-01
refactor(application-system): provider for header information

jeremybarbet:
## What

- Better way to pass the provider and fix warning


madebynoam (approved):
lgtm!

PR #4036, opened 2021-06-01
fix(gjafakort-application): Add sequelize assets

sindrig:
## What

Fix glob match in workspace.json

## Why

gjafakort-application migrations are failing

## Checklist:

### DIFF
--- a/workspace.json
+++ b/workspace.json
@@ -3403,6 +3403,11 @@
                 "glob": "migrations/**",
                 "input": "apps/gjafakort/application",
                 "output": "./"
+              },
+              {
+                "glob": ".sequelizerc",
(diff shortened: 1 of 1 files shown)


workspace.json
```diff
@@ -3403,6 +3403,11 @@
                 "glob": "migrations/**",
                 "input": "apps/gjafakort/application",
                 "output": "./"
+              },
+              {
+                "glob": ".sequelizerc",
```
darrikonn, line 3408: The glob above should cover this: `"*sequelize*"`???

sindrig: It doesn't catch hidden files. Might be something that could be fixed in the esbuildnx package but this is an ok fix for now IMO.

PR #4037, opened 2021-06-01
fix(auth-public-api): Move to esbuild

sindrig:
## What

Build auth-public-api like all other service

## Why

So it runs

## Checklist:


valurefugl (approved):

pshomov (approved):
LGTM

PR #4039, opened 2021-06-01
fix(parental-leave): avoid duplicated periods and define default value of spouse allowance

jeremybarbet:
## What

👆


baering (approved):

madebynoam (approved):
lgtm!

PR #4044, opened 2021-06-01
chore(core): Improving e2e in CI

saevarma:
# Improving e2e in CI 

Improving e2e-ci by using only spawn for child process, and piping otput to stdout for better debugging in CI.

## What

e2e test are sometimes flaky and recently we saw e2e test for `web` hanging for 25min without any useful logs.

## Why

To be able to debug flaky e2e tests in CI with more output logs.

## Checklist:

### DIFF
--- a/scripts/e2e-ci.js
+++ b/scripts/e2e-ci.js
@@ -1,8 +1,6 @@
 const yargs = require('yargs')
-const { spawn, exec } = require('child_process')
-const { promisify } = require('util')
-
-const pexec = promisify(exec)
+const { spawn } = require('child_process')
+const { exec } = require('./utils')
 
 // Setup command line args
 // prettier-ignore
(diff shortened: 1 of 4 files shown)


scripts/static-serve.js
```diff
@@ -33,14 +33,6 @@ const argv = yargs
 // Set the path for static file serve
 app.use(argv['base-path'], express.static(path.join(process.cwd(), argv.dist)))
 
-// Setup server shutdown signal from parent
-process.on('SIGTERM', () => {
```
sindrig, line 37: not needed anymore?

scripts/e2e-ci.js
```diff
@@ -108,69 +91,86 @@ const extractEnv = async () => {
...
 
   try {
-    const testResult = await pexec(CMD.TEST)
-    console.log(testResult.stdout)
-    console.log(testResult.stderr)
-
-    // Tests can fail without the command exiting with code 1
-    // So to be sure that all tests where successful we check
-    // for the string 'All specs passed!'
-    if (!testResult.stdout.includes('All specs passed!')) {
```
sindrig, line 159: not needed anymore?

saevarma: Yes this is still needed, but implemented a bit differently to allow the test process to stream the output to stdout. This was moved to the `test()` function and the test for `'All specs passed!'` is still there.

scripts/static-serve.js
saevarma: Seems like this event is never triggered when the main process uses `kill()` on the child process. So I  removed it.

scripts/e2e-ci.js
sindrig: Right. Missed that.

sindrig (approved):
LGTM

jeremybarbet (approved):
LGTM

PR #4046, opened 2021-06-02
fix(crc): children legal residence check

VilliGunn:
We now use the `fjolskylda` endpoint to determine if children live with the applicant, parentB or both.


baldur-kolibri (approved):
lgtm

PR #4047, opened 2021-06-02
feat(access-control): Setup skeleton for editing an access

darrikonn:
# Setup skeleton for editing an access

## What

Skeleton for editing an access

## Screenshots / Gifs

image

## Checklist:


libs/service-portal/settings/access-control/src/screens/Access/Access.tsx
```diff
@@ -21,6 +59,76 @@ function Access() {
...
+                      defaultMessage: 'Útskýring',
+                    })}
+                  </Text>
+                </T.HeadData>
+                <T.HeadData>
+                  <Text variant="small" color="blue600">
+                    {formatMessage({
+                      id: 'service.portal:access-control-access-created',
+                      defaultMessage: 'Í gildi frá',
+                    })}
```
eirikurn, line 111: These should be namespaced separately, eg under service portal settings or access-control namespaces.

libs/service-portal/settings/access-control/src/screens/Access/Access.tsx
```diff
@@ -1,8 +1,36 @@
...
+    scopes: [
+      {
+        id: '1337',
+        displayName: 'Undiraðgangur',
+        description: 'Útskýring',
+        validTo: '',
+      },
+    ],
+  },
+]
```
eirikurn, line 33: I assume we'll fetch groups/scopes to list, separately from the delegation which lists which scopes the user has, and how long they're valid.

darrikonn: Yeah, I just decided to stop there and work on getting the actual data in.

oddsson (approved):

eirikurn (approved):
Great!

PR #4060, opened 2021-06-03
test: Running tests sequentially to avoid potential concurrency issues in the CI

pshomov:
# Description

Run the jest-based tests sequentially in the CI

## What

This is an experiment to determine if the failures in most of the flaky tests are due to concurrency execution

## Why

We decided to try this approach in a meeting with @hrefnalind, @eirikurn and @GunnarIngiReykjalin 

## Checklist:


sindrig:
Could we rather fix the concurrency issues please?

pshomov:
> Could we rather fix the concurrency issues please?

This is a hypothesis for now. Due to a lack of time we cannot go into the real issue for now but the summer is looking really good time for that 🌞 .

eirikurn:
We already have a lot of parallelism by testing multiple workspaces at the same time. For integration tests, it could fill up the RAM if each workspace is also spinning up multiple Postgres containers and application servers in parallel.

But maybe we could differentiate between unit tests and integration tests. Not sure if it would affect the total duration though to run unit tests in parallel.

eirikurn (approved):

pshomov:
I forgot about us running test suites in parallel. Let's see how bad the test time gets.

sindrig (approved):
I want it noted that I'm not in favor of this, but I yield because I don't have time to look into this properly now. How will we improve stuff if we keep doing stuff like this, I don't know. :shrug:

apps/financial-aid/backend/test/setup.ts
```diff
@@ -11,12 +11,12 @@ import { AppModule } from '../src/app'
...
 let sequelize: Sequelize
 
-const truncate = () => {
+const truncate = async () => {
   if (!sequelize) {
     return
   }
 
-  Promise.all(
+  await Promise.all(
```
eirikurn, line 19: Good catch 😳

.github/workflows/pullrequest.yml
```diff
@@ -136,7 +136,7 @@ jobs:
     timeout-minutes: 25
     env:
       AFFECTED_PROJECTS: ${{ matrix.projects }}
-      MAX_JOBS: 4
+      MAX_JOBS: 1
```
eirikurn, line 139: Why this?

eirikurn (approved):

.github/workflows/pullrequest.yml
pshomov: To serialize the tests at the process level. I think we mentioned this during our meeting that seemed like some some tests were using a different, parallel running tests' DB. This is to exclude that possibility.

I have not measure this but it seems to me this parallelism was not buying us anything significant anyways.

valurefugl (approved):

thordurhhh (approved):

gudjong (approved):

PR #4063, opened 2021-06-03
feat(access-control): create resolvers for auth-public-api

darrikonn:
# Create resolvers for auth-public-api

## Checklist:


libs/api/domains/auth/src/lib/auth.service.ts
```diff
@@ -16,4 +19,39 @@ export class AuthService {
   getActorDelegations(user: User): Promise<Array<DelegationDTO>> {
     return this.delegationsApiWithAuth(user).delegationsControllerFindAllTo()
   }
+
+  getAuthScopes(): Promise<any> {
+    // XXX: waiting on auth-public-api implementation
```
unnarb, line 24: Not requesting a change. Marking with TODO: makes it visible in the TODO tree in VS Code.

unnarb (approved):

libs/api/domains/auth/src/lib/auth.service.ts
darrikonn: Yeah.. both XXX and TODO are highlighted in most IDEs and editors. 
But this will be implemented in a following  PR

libs/infra-nest-server/src/lib/bootstrap.ts
```diff
@@ -72,7 +73,10 @@ export const createApp = async (options: RunServerOptions) => {
 
   // Enable validation of request DTOs globally.
   app.useGlobalPipes(
-    new ValidationPipe({ whitelist: true, forbidNonWhitelisted: true }),
```
darrikonn, line 75: whitelisting does not interact well with GraphQL:
```typescript
import { Field, InputType } from '@nestjs/graphql'
@InputType()
export class DelegationInput {
  @Field((_) => String)
  name!: string
  
  @Field((_) => String)
  toNationalId!: string
}
```
this for instance won't work, since it'll be stripped in the API even though the client provides the correct input, so a 400 bad request will be thrown.

By whitelisting, class-validator enforces GraphQL to decorate every field with any class-validator decorator in order to pick it up, just for the sake of it... so the previous example needs to be changed to:
```typescript
import { Field, InputType } from '@nestjs/graphql'
import { Allow } from 'class-validator'
@InputType()
export class DelegationInput {
  @Allow()
  @Field((_) => String)
  name!: string
  
  @Allow()
  @Field((_) => String)
  toNationalId!: string
}
```
in order to prevent nestjs's validator from stripping it.

eirikurn: So GraphQL and Class Validator both solve this same thing:

* Validate data types and structure.
* Trim extraneous inputs.

Additionally, the class validator has many ways to validate input data, including string length, regex, etc. Arguably all fields should have some class validator decorator, as it's important to validate all user input. But that's probably the only reason to have `whitelist: true` in our GraphQL project, to force the developer to give this some thought.

Your call. We can disable whitelist in the graphql API, but I want to keep it in the REST APIs.

eirikurn: Actually, probably most of the graphql DTOs are directly transferred to internal services. Arguably proper validation should live there instead, so we don't have out of sync validation in our backend layers.

So let's disable the whitelist, but make sure that all user input is validated.

libs/api/domains/auth/src/lib/models/delegation.model.ts
```diff
@@ -68,6 +62,12 @@ export class CustomDelegation extends Delegation {
   @Field((type) => ID)
   id!: string
 
+  @Field((type) => Date, { nullable: true })
+  validTo?: Date
+
+  @Field((type) => [String])
+  scopesNames!: string[]
```
eirikurn, line 69: scopeNames?

eirikurn: Do we need this?

libs/api/domains/auth/src/lib/dto/updateDelegation.input.ts
```diff
@@ -0,0 +1,19 @@
...
+@InputType('AuthDelegationScopeInput')
+class Scope {
+  @Field((_) => String)
+  name!: string
+
+  @Field((_) => Date)
+  validTo!: Date
+}
+
+@InputType('UpdateAuthDelegationInput')
```
eirikurn, line 12: AuthUpdateDelegationInput?

cc @saevarma

libs/api/domains/auth/src/lib/dto/delegation.input.ts
```diff
@@ -0,0 +1,10 @@
+import { Field, InputType } from '@nestjs/graphql'
+
+@InputType('AuthDelegationInput')
+export class DelegationInput {
+  @Field((_) => String)
+  name!: string
+
+  @Field((_) => String)
+  toNationalId!: string
+}
```
eirikurn, line 10: Do we want to create empty delegations? Could be a source of confusion.

libs/api/domains/auth/src/lib/resolvers/main.resolver.ts
```diff
@@ -0,0 +1,50 @@
...
+@UseGuards(IdsUserGuard)
+@Resolver()
+export class MainResolver {
+  constructor(private authService: AuthService) {}
+
+  @Query(() => [Delegation], { name: 'authActorDelegations' })
+  getDelegations(@CurrentUser() user: User): Promise<Delegation[]> {
+    return this.authService.getActorDelegations(user)
+  }
+
```
eirikurn, line 26: actorDelegations is a completely separate query which we should not call from the access control page. We should create a `delegations` query which returns results for the CurrentUser instead of CurrentActor.

So if you are logged in to a company and open the access control page, you are managing delegations for the company, but the dropdown shows the (actor) delegations for you as a person.

libs/api/domains/auth/src/lib/dto/delegation.input.ts
eirikurn: I would prefer to create the delegation after finishing the scope screen.

libs/infra-nest-server/src/lib/bootstrap.ts
```diff
@@ -72,7 +73,10 @@ export const createApp = async (options: RunServerOptions) => {
...
+    new ValidationPipe({
+      forbidNonWhitelisted: true,
+      exceptionFactory: (errors: ValidationError[]) => errors.join('\n'),
```
eirikurn, line 78: What is this doing? Joining error objects?

eirikurn (approved):
LGTM. Mostly minor comments. Feel free to address here or later.

libs/api/domains/auth/src/lib/dto/updateDelegation.input.ts
darrikonn: Mutations should be named verb first. This also applies for our naming conventions regarding module names:
```Mutations should still follow the verb first rule. Then the module name, following the object, or “noun,” if applicable.```

https://docs.devland.is/technical-overview/api-design-guide/graphql-naming-conventions#solution

libs/api/domains/auth/src/lib/resolvers/main.resolver.ts
darrikonn: Yeah, this is a result of renaming the resolver from `auth.resolver` to `main.resolver`, since resolvers can be generic (thus `main`) or individual type resolvers (e.g. `delegation.resolver`).
So this is not connected to the access control page and IMO doesn't need any further actions.
Please raise your concern if you disagree :)

libs/infra-nest-server/src/lib/bootstrap.ts
darrikonn: Yeah, the error handler wasn't displaying any messages so I couldn't see why the class-validator was failing. After stringifying the errors, I could take actions based on the error message.

libs/api/domains/auth/src/lib/dto/delegation.input.ts
darrikonn: Yeah, that was the thought. Create delegations with no scopes. Then work with that delegation to apply/remove scopes based on api-scopes... similar to editing a delegation, still just working with a single delegation.

"User created a delegation but didn't apply any scopes"

Could you further elaborate what possible confusions might occur?

libs/api/domains/auth/src/lib/models/delegation.model.ts
darrikonn: Yeah, I can delete this! The original model typed the scope as string array, but I can see the field isn't queried, so I'll delete it 👍

libs/infra-nest-server/src/lib/bootstrap.ts
darrikonn: Yeah I agree, opting out would be a better approach.
Implemented in: https://github.com/island-is/island.is/pull/4063/commits/7b8c4f225b986526700936990544fa316141e140

libs/api/domains/auth/src/lib/resolvers/main.resolver.ts
eirikurn: Ok. Just saw that this resolver is named "getDelegations" and you implemented a "getDelegation" resolver, and we also need to get a list of delegation for the access control page.

So these two should be named "getActorDelegations" (authActorDelegations) and "getDelegation" (authDelegation). Then we should create a "getDelegations" (authDelegations) resolver.

libs/api/domains/auth/src/lib/dto/delegation.input.ts
eirikurn: Just that the user may not be aware that something is saved after the "first step" of the wizard. Need to check if a delegation with no scopes would/should show up on the front page list and in login dropdowns?

If so, it could make some people surprised (thought they had cancelled when they saw the scope list), and others might think "that's it! you have access." even though they have no access.

libs/api/domains/auth/src/lib/dto/updateDelegation.input.ts
eirikurn: Yeah, that documents how to namespace the mutation operation, but the docs don't explicitly address how to name input and payload object types that are used for mutations.

I expected the namespace to come first, but don't mind either way. Would be good to tweak the docs to make this clear.

darrikonn: Ahh, didn't even think about that 🙈 .. yeah, the docs need to be tweaked.
IMO `input` names should follow the name of their consumer + `Input`. But yeah, let's take it up in some other thread.

libs/api/domains/auth/src/lib/resolvers/main.resolver.ts
darrikonn: Agree!

libs/api/domains/auth/src/lib/dto/delegation.input.ts
darrikonn: Yeah I see.
We previously talked about `delegation.isActive` which checks if there are any active scopes, so IMO we should not display inactive delegations in the dropdown

eirikurn (approved):
LGTM!

Valur (approved):
LGTM

pshomov (approved):
LGTM

PR #4068, opened 2021-06-04
feat(island-ui): Add zIndex to useBoxStyles

oddsson:
# Add zIndex to useBoxStyles

## What

Add the ability to use zIndex: 1 in useBoxStyles

## Checklist:

### DIFF
--- a/libs/island-ui/core/src/lib/Box/useBoxStyles.treat.ts
+++ b/libs/island-ui/core/src/lib/Box/useBoxStyles.treat.ts
@@ -589,7 +589,15 @@ export const opacity = styleMap({
 })
 
 export const zIndex = styleMap({
-  1: { zIndex: 1 },
+  10: { zIndex: 10 },
+  20: { zIndex: 20 },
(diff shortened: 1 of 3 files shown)


libs/island-ui/core/src/lib/Box/useBoxStyles.treat.ts
```diff
@@ -588,6 +588,10 @@ export const opacity = styleMap({
   1: { opacity: 1 },
 })
 
+export const zIndex = styleMap({
+  1: { zIndex: 1 },
```
jeremybarbet, line 592: Sounds very restrictive to only have `1` here. I can think of having 3 boxes with different z-index to pass on top of each other. It's probably good to not allow everything so we can keep big z-indexes for overlay/modal and so on, but maybe allowing a few more values could be nice? `10 - 90`?

gudjong (approved):
LGTM

Supporting multiple indices could be done as a future improvement when the need arises.

jeremybarbet (approved):

PR #3845, opened 2021-05-17
fix(web): Include current day in upcoming auctions list

Ingimarsson:
# Include current day in upcoming auctions list

## What

This fixes the [auctions page](https://island.is/stofnanir/syslumenn/uppbod) so that auctions for the current day are also shown in the list of upcoming auctions.

## Why

It's confusing if you can't see today's auctions

## Checklist:

- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] Formatting passes locally with my changes
- [x] I have rebased against main before asking for a review

PR #4071, opened 2021-06-04
fix(web): Organization web hot-fixes

Ingimarsson:
# Organization web hot-fixes

## What

Cherry-picked #3845 by request from Syslumenn.

Also fixes a grid offset issue with subpage dropdowns (see screenshot) - not yet merged on main

## Screenshots / Gifs

This has been fixed.

image

## Checklist:

### DIFF
--- a/libs/api/domains/cms/src/lib/cms.contentful.service.ts
+++ b/libs/api/domains/cms/src/lib/cms.contentful.service.ts
@@ -286,8 +286,9 @@ export class CmsContentfulService {
       year !== undefined && month !== undefined
         ? new Date(year, month, 1)
         : new Date()
-    const toDate = new Date(fromDate.getTime())
+    fromDate.setDate(fromDate.getDate() - 1)
 
+    const toDate = new Date(fromDate.getTime())
     toDate.setMonth(toDate.getMonth() + 1)
 
     const params = {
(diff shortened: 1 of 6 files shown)


libs/api/domains/cms/src/lib/cms.contentful.service.ts
```diff
@@ -286,8 +286,9 @@ export class CmsContentfulService {
       year !== undefined && month !== undefined
         ? new Date(year, month, 1)
         : new Date()
-    const toDate = new Date(fromDate.getTime())
+    fromDate.setDate(fromDate.getDate() - 1)
 
+    const toDate = new Date(fromDate.getTime())
```
stefanhar, line 291: It's not obvious to me what we are doing here, mind explaining why this is needed?
And maybe add it as a comment?

Ingimarsson: The "upcoming auctions" used to show everything from the current datetime, so an auction at 12:00 would disappear at 12:01.

This makes the from date one day before.

I can add a comment in a fix PR I am working on for `main`. But this is the release branch so it would not be very useful :p

stefanhar (approved):

jeremybarbet (approved):

ingolfurs (approved):

PR #4074, opened 2021-06-04
fix(parental-leave): missing tiles and wrong ids

jeremybarbet:
## What

- https://app.asana.com/0/1182378413629561/1200409061590810

## Why

- So we have nice titles in the browser tabs while the step is active


madebynoam (approved):
lgtm!

PR #4076, opened 2021-06-04
docs: Explain the `extract-strings` script, improve examples

maranomynet:
Does what the title says

### DIFF
--- a/libs/localization/README.md
+++ b/libs/localization/README.md
@@ -205,7 +205,9 @@ The first time a namespace is encountered a new Contentful entry is created (eve
 **Notes on the `extract-strings` script's behavior**:
 
 - Adding a new id automatically creates it in Contentful.
-- Removing an id from the project will automatically remove that id from the Contentful Namespace entry.
+- Removing an id from the project will automatically remove\* that id from the Contentful Namespace entry.
 - `defaultMessage` will be used as the initial "Icelandic" translation.
 - Updating or removing a `defaultMessage` **does not** affect the "Icelandic" translation that was initially based on it.
 - If multiple instances of the same id are found, but with different defaults/description, the script throws an error.
+
+\*) Technically, missing ids are not removed from the Contentful entry but marked as deprecated. If the id is used again, the deprecation flag is removed.
--- a/libs/localization/README.md
+++ b/libs/localization/README.md
@@ -82,7 +82,8 @@ export default Home
 
 ## Message Declaration
 
-To declare a message you have to provide an `id` and a `defaultMessage`, `description` is optional but recommended.
+To declare a message you have to provide a unique `id`. The `defaultMessage` and `description` fields are optional, but recommended.
+
 You can declare a message by:
 
 - Using `formatMessage` method from the `useLocale` hook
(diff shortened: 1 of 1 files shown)


libs/localization/README.md
```diff
@@ -124,28 +125,35 @@ return (
...
-  id: 'global:title',
-  defaultMessage: 'Default title',
-  description: 'This is a default title in the global namespace',
+const greeting = defineMessage({
+  id: 'global:greeting',
+  defaultMessage: 'Default greeting',
+  description: 'This is a default greeting in the global namespace',
 })
 
 const messages = defineMessages([
```
jeremybarbet, line 134: Should be an object here

```suggestion
const messages = defineMessages({
```

libs/localization/README.md
```diff
@@ -124,28 +125,35 @@ return (
...
-  {
+  header: {
     id: 'global:title',
```
jeremybarbet, line 136: Just so it follows your convention, `global:header` and same below

libs/localization/README.md
```diff
@@ -189,3 +197,13 @@ Add the `extract-strings` script to `workspace.json`. Running this script will e
   }
 }
 ```
+
+Running `yarn nx run your-project:extract-strings` will extract messages from the project and create or update a Namespace entry in Contentful, if the namespace did not exist it will need to be published in Contentful before the client can query the entry.
```
jeremybarbet, line 201: Just to follow global notation when using template within blockcode

```suggestion
Running `yarn nx run <project>:extract-strings` will extract messages from the project and create or update a Namespace entry in Contentful, if the namespace did not exist it will need to be published in Contentful before the client can query the entry.
```

Regarding "if the namespace did not exist it will need to be published in Contentful before the client can query the entry" it will be the case as well when the entry already exists in Contentful. You will need to go there and publish each time

libs/localization/README.md
```diff
@@ -189,3 +197,13 @@ Add the `extract-strings` script to `workspace.json`. Running this script will e
...
+
+**Notes**:
+
+- Adding a new id automatically creates it in Contentful.
```
jeremybarbet, line 205: "automatically" when the extract strings script is executed

libs/localization/README.md
```diff
@@ -189,3 +197,13 @@ Add the `extract-strings` script to `workspace.json`. Running this script will e
...
+- Removing an id from the project will automatically remove that id from the Contentful Namespace entry.
```
jeremybarbet, line 206: At the moment it doesn't remove it from the Contentful entry but marks it as deprecated. If the id is used again, it will be unmarked from being deprecated

jeremybarbet (commented):
Just a few comments

libs/localization/README.md
maranomynet: Hmm...

I manually created and published an empty Namespace entry in Contentful before I ran the script.
The first time I ran the script, it generated a new Contentful entry with the exact same name, and I had to delete my original entry and publish the new one.

However, I've re-run the `extract-strings` script multiple times now and it has not generated any new entries. It seems to consistently update the one it created on the first run. (I'm guessing it has something to do with authorship, etc.)

libs/localization/README.md
maranomynet: I was intentionally making the `key` different from the `id` to emphasize they're independent of each other.

But I notice I used it as `messages.title` which I've now changed to `messages.header` to match with this one.

maranomynet:
@jeremybarbet f501191 contains my attempt at incorporating your feedback.

Let me know if you want more/other changes.

libs/localization/README.md
maranomynet: ...added that detail as a footnote.

libs/localization/README.md
jeremybarbet: Sorry, I did not explain myself clearly. I meant each time you extract the strings, you will need to go to Contentful and publish the entry to get the new strings through the API, not only when the entry does not exist.

Regarding the creation process, when you create an entry through Contentful it gives it a uid and you cannot manually change it through the UI, however through the extract-strings script, we give it the namespace id as uid, which is why it doesn't find the entry once you run the script again afterward

jeremybarbet (approved):

PR #4091, opened 2021-06-07
fix(application-templates-health-insurance): Patch flaky test

sindrig:
https://www.notion.so/Health-insurance-utils-Check-draft-applications-should-return-true-if-there-is-more-than-one-dra-8fab77bf042f4f2b8a35db5d5b41e2cd

## What

This test doesn't seem to be testing what the tester thinks it is.
The `oldestDraftId` will always be the last item created due to all
created attributes being the same, unless a clock tick occurs between
initializing the first external application and the second one - then it
will be the first external application, which for some reason has the
same UUID as the "current" application.

## Why

This test fails randomly. Whether or not it's testing what it should be should be reviewed asap.

## Checklist:


dd-sdang (approved):
This is fine to unblock failures in the pipeline. Will look into fixing the test properly to make sure it does what is intended in separate PR

PR #4092, opened 2021-06-07
feat(party-application): Bulk upload in party application

albinagu:
Bulk Upload in party application

on success:
Screenshot 2021-06-07 at 16 20 41

on error:
Screenshot 2021-06-07 at 16 21 03

## Checklist:


libs/application/templates/party-application/src/fields/BulkUpload/index.tsx
```diff
@@ -0,0 +1,141 @@
+import React, { FC, useState } from 'react'
+import {
+  Box,
+  Text,
+  Button,
+  LoadingIcon,
```
siggakogk, line 6: LoadingIcon has been deprecated, you should use LoadingDots instead :D

libs/application/templates/party-application/src/forms/CollectEndorsementsForm.ts
```diff
@@ -32,7 +32,7 @@ export const CollectEndorsementsForm: Form = buildForm({
               title: m.endorsementList.title,
               component: 'EndorsementList',
             }),
-            buildCheckboxField({
```
siggakogk, line 35: remove commented code

libs/application/templates/party-application/src/lib/messages.ts
```diff
@@ -196,6 +196,33 @@ export const m: MessageDir = {
...
+    fileFormatText: {
+      id: 'pa.application:uploadButton.fileFormatText',
+      defaultMessage: 'Tekið er við skjölum með endingu: .xlsx',
+      description: 'File format text',
+    },
+    uploadSuccess: {
+      id: 'pa.application:uploadButton.uploadSuccess',
+      defaultMessage: 'Pappírsmeðmælum hefur verið hlaðið upp!',
+      description: 'Upload succeeded text',
+    },
```
siggakogk, line 220: these strings already exist under collectEndorsements, maybe move the "uploadFail" string to collectEndorsements section and delete this section one.. or vice versa, just avoid duplication 😛

siggakogk (approved):

eirikurn (approved):

eirikurn (changes requested):
Wait, there's already a node-xlsx package in the repo. I'd like to avoid having two packages solving the same problem. Please evaluate.

eirikurn (approved):
Looking further into this, I see you're loading `xlsx` in the browser, and `node-xlsx` uses `xlsx` behind the scenes. Please be aware that `xlsx` is [900kb of minified js](https://bundlephobia.com/package/xlsx@0.17.0), so we need to make sure to not load it until needed. Afaik the applications and/or custom components are lazily loaded.

Regarding `node-xlsx`, I can only see it used in one place in the `voter-registry`, so we should get rid of it and just use `xlsx` everywhere instead, if possible.

cc @stefanhar

stefanhar:
> Looking further into this, I see you're loading `xlsx` in the browser, and `node-xlsx` uses `xlsx` behind the scenes. Please be aware that `xlsx` is [900kb of minified js](https://bundlephobia.com/package/xlsx@0.17.0), so we need to make sure to not load it until needed. Afaik the applications and/or custom fields are lazily loaded in separate bundles.
> 
> Regarding `node-xlsx`, I can only see it used in one place in the `voter-registry`, so we should get rid of it and just use `xlsx` everywhere instead, if possible.
> 
> cc @stefanhar

It's used in both the `voter-registry` and `party-letter-registry`.
I'll convert those services to use `xlsx` and remove the `node-xlsx` dependency in a separate PR

PR #4107, opened 2021-06-08
fix: rename License to EducationLicense in api

Valur:
## Why

Education license service was exporting a generic "Licence" type to api schema

## Checklist:


darrikonn (approved):

darrikonn (changes requested):
~Where are the front-end changes?~

darrikonn (commented):
BTW... an easier change would be to just change:
```
@ObjectType() -> @ObjectType('EducationLicense')
```

darrikonn (approved):

PR #4115, opened 2021-06-09
chore(deps): Enable dependabot pull requests

sindrig:
This might be too much. Objections to trying it out?


jeremybarbet (approved):
LGMT, let's try that 👍

PR #4140, opened 2021-06-09
feat(parental-leave): add 30 days VMST api + create periods DTO

jeremybarbet:
## What

- Transform users periods to VMST periods with rightPeriodCode and so on
- Add new VMST API to calculate dates based on 30 days rules


libs/application/template-api-modules/src/lib/modules/templates/parental-leave/parental-leave.service.ts
```diff
@@ -99,13 +105,113 @@ export class ParentalLeaveService {
...
+      )
+      const periodLength = Number(getPeriodLength?.periodLength ?? 0)
+
+      if (daysAllowed > periodLength) {
+        periods.push({
+          from: period.startDate,
+          to: period.endDate,
+          ratio: Number(period.ratio),
+          approved: false,
+          paid: false,
```
baering, line 140: Unrelated but I would like to get rid of these from our payloads if that makes sense (need to ask VMST).

baering (approved):

PR #4148, opened 2021-06-09
fix(feature-deployment): trim dots in cleanup

sindrig:
The feature deployment already does this, I forgot to make the same change in the cleanup phase


pshomov (approved):
LGTM

PR #4153, opened 2021-06-10
fix(parental-leave): add codeowner on template api module

jeremybarbet:


.github/CODEOWNERS
```diff
@@ -53,6 +53,7 @@
 /libs/application/ui-fields/                                                                  @island-is/aranja-applications
 /libs/application/ui-shell/                                                                   @island-is/aranja-applications
 /libs/application/template-api-modules/                                                       @island-is/aranja-applications
+/libs/application/template-api-modules/src/lib/modules/templates/parental-leave/              @island-is/aranja-applications
```
sindrig, line 56: isn't this caught by the line above? why would this one be needed as well...?

PR #4170, opened 2021-06-10
fix(judicial-system): Update investigation case state

oddsson:
# Update investigation case state

https://app.asana.com/0/1199153462262248/1200446333396000

## What

Update investigation case state from NEW to DRAFT after the HearingArrangements screen. 

## Why

This is expected behaviour

## Screenshots / Gifs

https://user-images.githubusercontent.com/3789875/121551849-816a4180-c9ff-11eb-9491-330b72567770.mov

## Checklist:


apps/judicial-system/web/src/utils/hooks/useCase.tsx
```diff
@@ -121,6 +129,49 @@ const useCase = () => {
     }
   }
 
+  const transitionCase = async (
```
gudjong, line 132: Good start. Later transitionCase needs to handle other transitions.

gudjong (approved):

PR #4178, opened 2021-06-11
chore(skilavottord): Adding dependency that cannot be js-bundled

pshomov:
## What

Fix skilavottord login process


tommigitgaur (commented):
LFTM

tommigitgaur (approved):
LFTM

PR #4211, opened 2021-06-15
fix(party-application): Adding validation rules for siganturee

siggakogk:
# ...

Adding validation rules to display a warning message for the signaturee if he is
a) not existing in the temp voter registry
b) not in the correct constituency

Some minor cleanup

## Checklist:


eirikurn (approved):
Would be good to assign codeowner for the template api.

stefanhar:
> Would be good to assign codeowner for the template api.

The api templates for `party-application` and `party-letter` seem to be defined already could the definition for `/libs/application/template-api-modules/` assigned to Aranja be taking precedence since it's lower in the CODEOWNERS file?

stefanhar (approved):

PR #4213, opened 2021-06-15
feat(DPO): Remove Administrative Contact Email and other deprecated fields

atligudl:
# Remove Administrative Contact Email and other deprecated fields

## What

Administrative Contact email field removed
Applicant address and zip code removed from overview screen and review form

## Why

Request from PO

## Checklist:


libs/application/templates/document-provider-onboarding/src/forms/DocumentProviderApplication.ts
```diff
@@ -83,12 +83,6 @@ export const DocumentProviderOnboarding: Form = buildForm({
               title: m.administrativeContactName,
               placeholder: m.administrativeContactNamePlaceholder,
             }),
-            buildTextField({
-              id: 'administrativeContact.email',
-              title: m.administrativeContactEmail,
```
fjandakornid, line 88: Isn't this a required field in the API, and would therefor break the application?

Thorrik58: Could you elaborate on what you mean by the API? 
According to the data schema the administrativeContact only needs a phoneNumber and a name. 
See here: https://github.com/island-is/island.is/pull/4213/files/ad1c42b949643f335115ba5a667e48a5a18921aa#diff-78e8283282abb55432d9560707ba386fbe965a62996cbce0cb9ea52d3e640159R39

PR #4232, opened 2021-06-16
feat(ia): Email Sender Field

atligudl:
# Institution Application - Email Sender Field

## What

contact name and email set as emailFrom when sending application submission emails

## Why

Request from PO

## Checklist:

### DIFF
--- a/libs/application/template-api-modules/src/lib/modules/templates/institution-collaboration/emailGenerators/applicationEmail.ts
+++ b/libs/application/template-api-modules/src/lib/modules/templates/institution-collaboration/emailGenerators/applicationEmail.ts
@@ -44,13 +44,19 @@ export const generateApplicationEmail: ApplicationEmail = (
       )
     : []
 
+  const contactName = getValueViaPath(application.answers, 'contact.name') as
+    | string
+    | undefined
+  const contactEmail = getValueViaPath(application.answers, 'contact.email') as
+    | string
+    | undefined
   const overview = applicationOverviewTemplate(application)
   const body = dedent(`<h2>Yfirlit umsóknar</h2> ${overview}`)
 
   return {
     from: {
-      name: applicationSenderName,
-      address: applicationSenderEmail,
+      name: contactName ?? applicationSenderName,
+      address: contactEmail ?? applicationSenderEmail,
     },
     to: [
       {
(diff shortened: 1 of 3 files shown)


eirikurn (approved):

libs/application/template-api-modules/src/lib/modules/templates/institution-collaboration/emailGenerators/applicationEmail.ts
```diff
@@ -44,13 +44,19 @@ export const generateApplicationEmail: ApplicationEmail = (
...
+    | undefined
   const overview = applicationOverviewTemplate(application)
   const body = dedent(`<h2>Yfirlit umsóknar</h2> ${overview}`)
 
   return {
     from: {
-      name: applicationSenderName,
-      address: applicationSenderEmail,
+      name: contactName ?? applicationSenderName,
+      address: contactEmail ?? applicationSenderEmail,
```
eirikurn, line 59: Hm, what's the use case for this? I assume spam filters will block us if we start sending emails from arbitrary email addresses.

eirikurn (changes requested):

libs/application/template-api-modules/src/lib/modules/templates/institution-collaboration/emailGenerators/applicationEmail.ts
atligudl: Good point, I'll raise the issue with our PO

GunnarIngiReykjalin: The usecase is that there is a ticketing system on the other end (Zendesk) where communication with the applicant takes place. In the current setup, the sender is island@island.is so notifications from the ticketing system never reach the applicant.
That being said, we do indeed risk being flagged as SPAM if we introduce this functionality.
I'm discussing this with our BO

GunnarIngiReykjalin: Update. Lets scrap this functionality.
We will take a look at using an API with the ticketing system after the summer.

GunnarIngiReykjalin: @atligudl FYI

PR #4245, opened 2021-06-17
feat(auth-public-api): Use token exchange for national registry service

valurefugl:
Enable token exchange in auth middleware. Use token exchange in national registry service.

## Checklist:

- [ x] I have performed a self-review of my own code

### DIFF
--- a/apps/services/auth-public-api/src/environments/environment.ts
+++ b/apps/services/auth-public-api/src/environments/environment.ts
@@ -29,6 +29,15 @@ const devConfig = {
       clientId: 'IS-DEV/GOV/10000/island-is-client',
       memberClass: XRoadMemberClass.GovernmentInstitution,
     },
+    authMiddlewareOptions: {
+      forwardUserInfo: false,
+      tokenExchangeOptions: {
+        issuer: 'https://identity-server.dev01.devland.is',
+        clientId: '@island.is/clients/national-registry',
+        clientSecret: process.env.NATIONAL_REGISTRY_IDS_CLIENT_SECRET,
+        scope: '@skra.is/individuals',
+      },
+    },
   },
 }
 
@@ -62,6 +71,15 @@ const prodConfig = {
       memberCode: process.env.XROAD_NATIONAL_REGISTRY_MEMBER_CODE,
       apiPath: process.env.XROAD_NATIONAL_REGISTRY_API_PATH,
       clientId: process.env.XROAD_NATIONAL_REGISTRY_CLIENT_ID,
+    },
+    authMiddlewareOptions: {
+      forwardUserInfo: false,
+      tokenExchangeOptions: {
+        issuer: process.env.IDS_ISSUER,
+        clientId: process.env.NATIONAL_REGISTRY_IDS_CLIENT_ID,
+        clientSecret: process.env.NATIONAL_REGISTRY_IDS_CLIENT_SECRET,
+        scope: process.env.NATIONAL_REGISTRY_IDS_SCOPE,
+      },
     },
   },
 }
--- a/apps/services/auth-public-api/src/environments/environment.ts
+++ b/apps/services/auth-public-api/src/environments/environment.ts
@@ -35,7 +35,7 @@ const devConfig = {
         issuer: 'https://identity-server.dev01.devland.is',
         clientId: '@island.is/clients/national-registry',
         clientSecret: process.env.NATIONAL_REGISTRY_IDS_CLIENT_SECRET,
-        scope: '@skra.is/individuals',
+        scope: '@skra.is/individuals api_resource.scope', // TODO: remove api_resource.scope
       },
     },
   },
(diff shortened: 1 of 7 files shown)


libs/auth-nest-tools/src/lib/auth.middleware.ts
```diff
@@ -18,22 +19,111 @@ interface Middleware {
...
+      return ''
+    }
+  }
+
+  private logError(errorMessage: string) {
+    if (this.logger) {
+      this.logger.error(errorMessage)
+    } else {
+      console.log(errorMessage)
+    }
```
eirikurn, line 127: It´s possible to import a logger object from @island.is/logging directly.

libs/auth-nest-tools/src/lib/auth.middleware.ts
```diff
@@ -18,22 +19,111 @@ interface Middleware {
...
+          client_secret: options.clientSecret,
+          subject_token: accessToken,
+          subject_token_type: 'urn:ietf:params:oauth:token-type:access_token',
+        }),
+      })
+
+      if (!response.ok) {
+        this.logError(`Token exchange failed, ${await response.text()}`)
+        return ''
+      }
```
eirikurn, line 93: Should throw an error instead if token exchange fails?

apps/services/auth-public-api/src/environments/environment.ts
```diff
@@ -26,6 +26,15 @@ const devConfig = {
       clientId: 'IS-DEV/GOV/10000/island-is-client',
       memberClass: XRoadMemberClass.GovernmentInstitution,
     },
+    authMiddlewareOptions: {
+      forwardUserInfo: false,
+      tokenExchangeOptions: {
+        issuer: 'https://identity-server.dev01.devland.is',
+        clientId: '@island.is/clients/national-registry',
+        clientSecret: process.env.NATIONAL_REGISTRY_IDS_CLIENT_SECRET,
+        scope: '@skra.is/individuals api_resource.scope', // TODO: remove api_resource.scope
```
eirikurn, line 35: Why is api_resource.scope needed? What's the plan? Just curious.

libs/auth-nest-tools/src/lib/auth.middleware.ts
eirikurn: Better than having to pass in a logger and sometimes forgetting and ending with console.log

libs/auth-nest-tools/src/lib/auth.middleware.ts
```diff
@@ -18,22 +19,111 @@ interface Middleware {
...
       context.init.headers = Object.assign({}, context.init.headers, {
         'User-Agent': this.auth.userAgent,
         'X-Real-IP': this.auth.ip,
       })
     }
   }
+
+  private async exchangeToken(
+    accessToken: string,
+    options: TokenExchangeOptions,
```
eirikurn, line 72: Should access the options from `this` directly?

libs/auth-nest-tools/src/lib/auth.middleware.ts
```diff
@@ -18,22 +19,111 @@ interface Middleware {
...
+  forwardUserInfo: boolean
+  tokenExchangeOptions?: TokenExchangeOptions
+}
+
+export interface TokenExchangeOptions {
+  issuer: string
+  clientId: string
+  clientSecret: string
+  scope: string
+}
```
eirikurn, line 32: Would be nice if the token exchange could be short-circuited if the authentication already contains the necessary scopes.

libs/auth-nest-tools/src/lib/auth.middleware.ts
eirikurn: Might be good to include the requested URL from the context in the error/log.

eirikurn (changes requested):
Nice to build this into the AuthMiddleware, at least for now.

libs/auth-nest-tools/src/lib/auth.middleware.ts
valurefugl: Throwing errors now, no logging in middleware.

libs/auth-nest-tools/src/lib/auth.middleware.ts
valurefugl: Fixed.

libs/auth-nest-tools/src/lib/auth.middleware.ts
valurefugl: Implemented

apps/services/auth-public-api/src/environments/environment.ts
valurefugl: This is what Þjóðskrá used in the beginning and changing it has not been a priority for them. Should be removed soon.

unnarb (approved):

valurefugl:
Abandoning this, continued here, https://github.com/island-is/island.is/pull/4302

PR #4247, opened 2021-06-18
fix(api): move driving license xroad path into env

osk:
# Move driving license xroad path into env

## What

Move xroad path for RLS driving license service into env var `DRIVING_LICENSE_XROAD_PATH`.

## Why

Path was mistakenly left hard-coded in first version.

## Checklist:

### DIFF
--- a/libs/api/domains/license-service/src/lib/client/driving-license-client/drivingLicenseService.api.ts
+++ b/libs/api/domains/license-service/src/lib/client/driving-license-client/drivingLicenseService.api.ts
@@ -12,7 +12,12 @@ export class LicenseServiceApi {
   private readonly xroadPath: string
   private readonly secret: string
 
-  constructor(xroadBaseUrl: string, xroadClientId: string, xroadPath: string, secret: string) {
+  constructor(
+    xroadBaseUrl: string,
+    xroadClientId: string,
+    xroadPath: string,
+    secret: string,
+  ) {
     this.xroadApiUrl = xroadBaseUrl
     this.xroadClientId = xroadClientId
     this.xroadPath = xroadPath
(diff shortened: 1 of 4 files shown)


libs/api/domains/license-service/src/lib/client/driving-license-client/drivingLicenseService.api.ts
```diff
@@ -71,12 +78,8 @@ export class LicenseServiceApi {
   async getGenericDrivingLicense(
     nationalId: User['nationalId'],
   ): Promise<GenericDrivingLicenseResponse[] | null> {
-    // TODO(osk) Move into env
-    const xroadDrivingLicensePath =
-      'r1/IS-DEV/GOV/10005/Logreglan-Protected/RafraentOkuskirteini-v1'
-
     const response = await this.requestApi(
-      `${xroadDrivingLicensePath}/api/Okuskirteini/${nationalId}`,
+      `${this.xroadPath}/api/Okuskirteini/${nationalId}`,
```
eirikurn, line 82: For future clients, I recommend moving all xroad path logic to a lower (requestApi in this case), so this logic can call endpoints just with "/api/Okuskirteini/...".

Also rename the variable/env: xroadProviderId, and hardcode the r1 segment.

That way, ClientID and ProviderID have the same structure and (with the baseUrl) can be used to call SOAP and REST endpoints on x-road.

For REST endpoints, the request is `${xroadBaseUrl}/r1/${xroadProviderId}${endpoint}`, with client id in a header.

For SOAP endpoints, the request is `${xroadBaseUrl}` with client id AND provider id included in the SOAP request body (split up into multiple XML elements).

eirikurn (approved):
Happy to get this out! 👏

libs/api/domains/license-service/src/lib/client/driving-license-client/drivingLicenseService.api.ts
osk: Yes, makes sense, I'll add to our list for v2 to refactor this

eirikurn (approved):

PR #4257, opened 2021-06-18
feat(core): enhanced fetch

eirikurn:
## What

A new library providing an createEnhancedFetch function.
 
- Includes circuit breaker logic. By default, if more than 50% of the requests from the last 10 seconds are misbehaving, we'll open the circuit. All future requests will be stopped to lower pressure on the remote server. Every 30 seconds we'll allow one request through. If it's successful, we'll close the circuit and let requests through again.
- Includes request timeout logic. By default, throws an error if there is no response in 10 seconds.
- Throws an error for non-200 responses. The error object includes details from the response, including a `problem` property if the response implements the [Problem Spec](https://datatracker.ietf.org/doc/html/rfc7807).
- Logs circuit breaker events and failing requests
- Optionally opens the circuit for 400 responses.
- Optionally parses and logs error response bodies.

This function (and it's error logic) is mostly compatible with "OpenAPI Generator" clients. The only difference revolve around non-200 responses. It throws an Error object instead of the response (all response properties are copied to the Error object), and since these errors are thrown "inside" the fetch call, any "post" middlewares will not be invoked for non-200 responses.

This PR also updates the delegation API to calls RSK and Thjodskra in parallel using EnhancedFetch.

## Why

To improve API error handling. Back off and fail early if there are upstream API issues.

## Checklist:


libs/clients/middlewares/README.md
```diff
@@ -0,0 +1,88 @@
...
+  } catch (error) {
+    // For non-200 responses:
+    error.name // FetchError
+    error.response // the response object
+    error.problem // if server returned a standard problem response.
+    error.body // if configured with `logErrorResponseBody`.
+    error.url // the requested url
+    error.status // response status code
+    `error.statusText` // response status text
+    error.headers // response headers
```
darrikonn, line 67: What's going on here?

libs/clients/middlewares/src/lib/enhanced-fetch.ts
```diff
@@ -0,0 +1,184 @@
+import CircuitBreaker from 'opossum'
```
darrikonn, line 1: nit-picky, but the file name breaks the IslandIs file naming conventions:
`enhanced-fetch.ts` -> `enhancedFetch.ts`

libs/clients/middlewares/src/lib/enhanced-fetch.ts
```diff
@@ -0,0 +1,184 @@
...
+  return error
+}
+
+/**
+ * Creates a fetch function for resilient ops:
+ *
+ * - Includes circuit breaker logic. By default, if more than 50% of the
+ *   requests from the last 10 seconds are misbehaving, we'll open the circuit.
+ *   All future requests will be stopped to lower pressure on the remote server.
+ *   Every 30 seconds we'll allow one request through. If it's successful, we'll
```
darrikonn, line 89: Curious why every 30 seconds is used?
Is 30 seconds some "known" sweet spot to handle a circuit breaker?

darrikonn (commented):
Will enhancedFetch only be used for 3rd party clients? (thinking about the location of the lib)

libs/clients/middlewares/README.md
eirikurn: Ah. I wrote this to document how response errors look, but then wrote better docs further up. I'll remove it.

libs/clients/middlewares/src/lib/enhanced-fetch.ts
eirikurn: I love nit-picks. Guess I forgot what the naming convention was, as I've gotten so used to the nx/nest style `some-stuff.service.ts` file names. Will fix.

libs/clients/middlewares/src/lib/enhanced-fetch.ts
eirikurn: I decided to keep (and document) all of opossum's default options (except volumeThreshold), since we don't have the experience to know better defaults.

eirikurn:
@darrikonn, in the beginning I meant to split this up into multiple opt-in "OpenAPI Generator" middlewares with one that includes all of the "best practice" ones. I then figured out that the middleware system wouldn't be able to do half the things I'm doing.

We're also not using OpenAPI Generator for all apis, eg we can plug in custom `fetch` function in RESTDataSource.

But later we can add more "middlewares" here, eg x-road logic, auth-middleware and more.

To answer your question, I think most clients should use this. We should also move more of our own clients to this folder imo.

valurefugl (approved):

darrikonn (approved):
Nice :+1:

VilliGunn (approved):

stefanhar (approved):
LGTM

PR #4259, opened 2021-06-18
feat(core): mask national ids from logs

eirikurn:
## What

Mask national ids from our logs.

## Why

To avoid logging personally identifiable information.

## Checklist:


saevarma (approved):
LGTM.
Only one question:
Is it possible that the nationalId could also be in the form: `123456 7890`?

eirikurn:
I think that's exceedingly unlikely in logs. I'm not even sure we have a format function for that at the moment.

pshomov (approved):
Triple BEAM FTW 🌈 🦄

darrikonn (approved):
Nice

VilliGunn (approved):
LGTM 💯

PR #4263, opened 2021-06-21
fix(judicial-system): Fixes isolation to date in prison notifications

gudjong:
# Isolation to fix

https://app.asana.com/0/1199153462262248/1200480969662510/f

## What

☝️ 

## Why

## Screenshots / Gifs

image
image
image

## Checklist:

### DIFF
--- a/apps/judicial-system/web/src/routes/Court/Confirmation/Confirmation.tsx
+++ b/apps/judicial-system/web/src/routes/Court/Confirmation/Confirmation.tsx
@@ -404,6 +404,8 @@ export const Confirmation: React.FC = () => {
                       {formatCustodyRestrictions(
                         workingCase.accusedGender,
                         workingCase.custodyRestrictions,
+                        workingCase.validToDate,
+                        workingCase.isolationToDate,
                       )}
                     </Text>
                   </Box>
--- a/apps/judicial-system/web/src/routes/Court/Confirmation/Confirmation.tsx
+++ b/apps/judicial-system/web/src/routes/Court/Confirmation/Confirmation.tsx
@@ -81,6 +81,7 @@ const SigningModal: React.FC<SigningModalProps> = ({
     },
     fetchPolicy: 'no-cache',
   })
+
   // TODO: Handle case when resSignatureConfirmationResponse is never set
   const resSignatureConfirmationResponse = data?.signatureConfirmation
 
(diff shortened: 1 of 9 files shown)


apps/judicial-system/web/src/routes/Court/Confirmation/Confirmation.tsx
```diff
@@ -89,20 +90,26 @@ const SigningModal: React.FC<SigningModalProps> = ({
...
+                )
+              : workingCase.state === CaseState.REJECTED ||
+                workingCase.state === CaseState.ACCEPTED
 
-        if (caseCompleted) {
-          await sendNotification(workingCase.id, NotificationType.RULING)
+          if (caseCompleted) {
+            await sendNotification(workingCase.id, NotificationType.RULING)
+          } else {
+            // TODO: Handle error
```
VilliGunn, line 109: Should we create a ticket in Asana for these TODOs?

libs/judicial-system/formatters/src/lib/formatters.ts
```diff
@@ -160,12 +162,29 @@ export function formatCustodyRestrictions(
     return 'Sækjandi tekur fram að gæsluvarðhaldið sé án takmarkana.'
   }
 
+  const formattedValidToDateAndTime = `${formatDate(
+    validToDate,
+    'PPPp',
+  )?.replace(' kl.', ', kl.')}`
+
+  const formattedIsolationToDateAndTime = `${formatDate(
```
VilliGunn, line 170: Why is there a mismatch in the wording and date format here and here: https://github.com/island-is/island.is/pull/4263/files#diff-b2788662cb5317437f7ba480f353c0d321e3181497eb8d1accfed00b066ca25eR111

Day and time formatting
```skal sæta einangrun ekki lengur en til {date.replace('dagur,', 'dagsins')?.replace(' kl.', ', kl.')}```
vs
Only time formatting
```skuli sæta einangrun til {date.replace(' kl.', ', kl.')}```

apps/judicial-system/web/src/routes/Court/Confirmation/Confirmation.tsx
gudjong: I think there already is a ticket for client side error handling. At least it has been discussed multiple times.
There is a general lack of api call error handling in the client.

libs/judicial-system/formatters/src/lib/formatters.ts
gudjong: No particular reason, but there is no reason they should be the same either. These are two different sentences.

The general pattern has been to use a short version in SMS and long version else where. I will modify this text to follow that convention.

VilliGunn (approved):

PR #4267, opened 2021-06-21
feat(web): Add project pages

Ingimarsson:
# Web: Add project pages

## What

Project pages are customizable pages for presenting government projects and initiatives. They have the following features.

- Can be either a single page (without sidebar) or include a menu with subpages
- Custom header can be specified
- Includes a news section

**Thought:** It might be nice to combine everything from this and organization pages into a single "Generic Page" type. But this is mostly for demo purposes right now :) 

## Why

This was requested by Digital Iceland

## Screenshots / Gifs

localhost_v_ferdathjonusta

## Checklist:


apps/web/screens/Project/Project.tsx
```diff
@@ -0,0 +1,212 @@
...
+        <title>{projectPage.title} | Ísland.is</title>
+      </Head>
+      <DefaultProjectHeader projectPage={projectPage} />
+      <ProjectWrapper
+        withSidebar={projectPage.sidebar}
+        sidebarContent={
+          <>
+            <Navigation
+              baseId="pageNav"
+              items={projectPage.sidebarLinks?.map((x) => ({
```
arnorhs, line 94: it's usually cleaner to destructure the variable rather than using a bad placeholder name if possible, eg.

libs/api/domains/cms/src/lib/cms.contentful.service.ts
```diff
@@ -296,6 +298,19 @@ export class CmsContentfulService {
     return (result.items as types.IAuction[]).map(mapAuction)[0]
   }
 
+  async getProjectPage(slug: string, lang: string): Promise<ProjectPage> {
+    const params = {
+      ['content_type']: 'projectPage',
```
arnorhs, line 303: aren't the square brackets unnecessary here?

apps/web/screens/Project/Project.tsx
arnorhs, line 94: it's usually cleaner to destructure the variable rather than using a placeholder name if possible, eg.
```
.map(({ text, url }) => ({
```

ps. this may be a copy paste of code that was already in place, in which case, sorry about that

libs/api/domains/cms/src/lib/models/projectPage.model.ts
```diff
@@ -0,0 +1,67 @@
...
+  @Field(() => GenericTag, { nullable: true })
+  newsTag!: GenericTag | null
+
+  @Field(() => [ProjectSubpage])
+  projectSubpages!: Array<ProjectSubpage>
+}
+
+export const mapProjectPage = ({ sys, fields }: IProjectPage): ProjectPage => ({
+  id: sys.id,
+  title: fields.title ?? '',
```
arnorhs, line 54: are these fields ever undefined/null? if this is mostly to appease the type system, you should be able to get away with doing `xxx.yy as string` instead of assigning a default

arnorhs (commented):
looking good.. i mostly have minor comments, but to be fair, i'm not so versed in the treat stuff yet or the way we are generally querying contentful, but this all seems pretty reasonable.

apps/web/screens/Project/Project.tsx
Ingimarsson: Good advice, changed it.

libs/api/domains/cms/src/lib/models/projectPage.model.ts
Ingimarsson: This is how all the other CMS models do it, I think it's best to follow the trend :stuck_out_tongue: 

When the Contentful preview API is used (for example on the beta), all fields can be null. The defaults are provided so that WIP content can be previewed.

arnorhs (approved):

PR #4268, opened 2021-06-21
fix(ci): Bust cypress cache since we're on self hosted now

sindrig:
https://github.com/island-is/island.is/runs/2876016028?check_suite_focus=true


pshomov (approved):
LGTM

PR #4278, opened 2021-06-22
fix: Fix length check in create secret

sindrig:
You've got a problem. You decide to use regex. Now you've got two problems.


busla (approved):
LGTM

PR #4293, opened 2021-06-23
feat(auth-public-api): Fixing security issues

unnarb:
# ...

## What

## Why

## Screenshots / Gifs

## Checklist:


libs/auth-api-lib/src/lib/services/resources.service.ts
```diff
@@ -346,6 +346,47 @@ export class ResourcesService {
...
+        allowExplicitDelegationGrant: true,
+        isAccessControlled: false,
+      },
+    })
+
+    // const allowedScopes = userScopes.filter(
+    //   (x) => x.allowExplicitDelegationGrant && !x.isAccessControlled,
+    // )
+    // return allowedScopes
+  }
```
valurefugl, line 366: can remove this, right?

libs/auth-api-lib/src/lib/services/resources.service.ts
```diff
@@ -346,6 +346,47 @@ export class ResourcesService {
...
+        },
+        allowExplicitDelegationGrant: true,
+      },
+    })
+
+    // const allowedIdentityResources = userIdentityResources.filter(
+    //   (x) => x.allowExplicitDelegationGrant
+    // )
+    // return allowedIdentityResources
+  }
```
valurefugl, line 388: also this

libs/auth-api-lib/src/lib/services/resources.service.ts
```diff
@@ -346,6 +346,47 @@ export class ResourcesService {
...
+  /** Filters out scopes that don't have delegation grant and are access controlled */
+  async findAllowedDelegationApiScopeListForUser(scopes: string[]) {
+    this.logger.debug(`Finding allowed api scopes for scopes ${scopes}`)
+    return this.apiScopeModel.findAll({
+      where: {
+        name: {
+          [Op.in]: scopes,
+        },
+        allowExplicitDelegationGrant: true,
+        isAccessControlled: false,
```
valurefugl, line 358: we should probably also exclude "alsoForDelegatedUser" scopes

libs/auth-api-lib/src/lib/services/resources.service.ts
valurefugl: same with "alsoForDelegatedUser"

valurefugl (changes requested):

valurefugl (approved):

PR #4303, opened 2021-06-24
fix(auth): Improve signin error handling

eirikurn:
## What

* Render an error screen if the OIDC callback has an error from the IDS.
* Avoid race conditions in redirects during error callbacks.
* Never pass the OIDC callback url as a return url for IDS.

## Why

Better error handling. Spotted when an IDS regression caused strange client behaviour.

## Screenshots / Gifs

In case of an error during IDS login:

image

## Checklist:


saevarma (approved):
LGTM

PR #4317, opened 2021-06-24
feat(core): Add className props to Text and Button, and margins* props to Divider

maranomynet:
Attempt to make meaningless div/Box wrappers needed less often to do simple custom layout/styling tweaks.

Also vertical-margin-collapse fix for the Divider component.

## Checklist:


jeremybarbet (changes requested):
It was a design decision to not define any `classname` props for these components. We don't want unwanted visual changes by passing custom classnames that are not part of the design system. All the design variants should be already defined within the component itself. If it's more a matter of spacing/layout, for example, wrapping the component within Grid/Box is the way to go

PR #4318, opened 2021-06-24
feat(identity): query national registry user with masking

darrikonn:
# Identity

## What

Add mask out middleware to mask out fields that user doesn't have permission to view

## Why

To be able to query external users using national registry without being able to view personal informations.

## Screenshots / Gifs

masker

### What is happening?

The gender is being masked out since the user being queried isn't the logged in user and `gender` is not opted in in `validFields`

## Checklist:


libs/api/domains/identity/src/lib/models/identity.model.ts
```diff
@@ -0,0 +1,68 @@
...
+@Extensions({
+  validateIf,
+  validFields: [
+    'address',
+    'age',
+    'fullName',
+    'birthday',
+    'type',
+    'name',
+    'nationalId',
```
darrikonn, line 57: Opt-in is a safer choice for masking out fields, since as new fields come in, they could potentially slip out here.

libs/api/domains/identity/src/lib/models/address.model.ts
```diff
@@ -0,0 +1,19 @@
+import { Field, ObjectType, ID } from '@nestjs/graphql'
+
+@ObjectType('IdentityAddress')
+export class Address {
+  @Field(() => ID)
+  code!: string
```
eirikurn, line 6: What is this? Should add a description.

libs/api/domains/identity/src/lib/models/address.model.ts
```diff
@@ -0,0 +1,19 @@
...
+
+  @Field(() => String, { nullable: true })
+  lastUpdated?: string
```
eirikurn, line 9: What is this? Should add a description.

libs/api/domains/identity/src/lib/models/identity.model.ts
```diff
@@ -0,0 +1,68 @@
...
+
+  @Field(() => Address, { nullable: true })
+  address?: Address
+
+  @Field(() => IdentityType)
+  type!: IdentityType
+}
+
+const validateIf = ({ context, source }: MiddlewareContext): boolean => {
+  return (context as any).req.user.nationalId !== source.nationalId
```
eirikurn, line 45: Would be nice to create our own graphql context type. I think we have somewhere in auth-nest-tools a type of `request` with our user/auth attributes.

Should conditionally read `user` as it can be null?

libs/api/domains/identity/src/lib/models/identity.model.ts
```diff
@@ -0,0 +1,68 @@
...
+}
+
+@Extensions({
+  validateIf,
+  validFields: [
```
eirikurn, line 50: Should be one extension? Naming suggestion:

```
filterFields: {
  condition?: (context: MiddlewareContext) => boolean // Defaults to true.
  fields: Array<string>
}
```

libs/api/domains/identity/src/lib/identity.resolver.ts
```diff
@@ -0,0 +1,56 @@
...
+@UseGuards(IdsUserGuard)
+@Resolver(() => IdentityPerson)
+@Resolver(() => IdentityCompany)
+export class IdentityResolver {
+  constructor(private nationalRegistryService: NationalRegistryService) {}
+
+  @Query(() => Identity, { name: 'identity', nullable: true })
+  async getIdentity(
+    @CurrentUser() user: User,
+    @Args('nationalId', { nullable: true }) optionalNationalId?: string,
```
eirikurn, line 32: Should have an input type according to the guidelines.

apps/api/src/app/graphql.middleware.ts
```diff
@@ -0,0 +1,23 @@
+import { FieldMiddleware, NextFn, MiddlewareContext } from '@nestjs/graphql'
+
+export const maskOutFieldsMiddleware: FieldMiddleware = async (
```
eirikurn, line 3: Would be nice to get tests for this. At least a unit test.

Integration test for IdentityPerson field filtering would also be nice.

eirikurn (approved):
Really clean implementation. IMO this is good enough for the next release, but check out the comments and see if we can address them soon.

I'll write down a task to update the graphql naming conventions.

eirikurn:
Also, the new national registry service should already support our use case. Happy to join for further tests tomorrow.

libs/api/domains/identity/src/lib/models/identity.model.ts
darrikonn: I assumed this wouldn't get called since its resolvers are protected with auth and thus at this stage, the req.user will always be set.


But I agree with our own graphql context type. Will add it.

valurefugl (approved):

osk (approved):

PR #4321, opened 2021-06-25
fix(e2e): Allow e2e reruns

### LINKED ISSUE #2343: Re-runs for e2e fail
Re-running any action from the github-actions UI fails because cypress recognized that the commit has run before.

sindrig:
This closes #2343


.github/workflows/pullrequest.yml
```diff
@@ -117,6 +118,7 @@ jobs:
           if [[ $CHUNKS != "[]" ]]; then
             echo "::set-output name=CHUNKS::{\"projects\":$CHUNKS}"
           fi
+          echo "::set-output name=BUILD_ID::$GITHUB_RUN_ID-$(uuidgen)"
```
eirikurn, line 121: Should this be E2E_BUILD_ID?

sindrig: No - the output of the `e2e_projects` step is named BUILD_ID, but the output of the `prepare` job is named E2E_BUILD_ID (see https://github.com/island-is/island.is/pull/4321/files#diff-ce9a2f9036cc2a5aaf1f78839c4f65c7a12808822e0dd0f3552471921ac5e007R19)

sindrig: I can rename it to E2E_BUILD_ID if it's confusing. I was going for consistency with the `*_CHUNKS` stuff - but maybe that's not necessary.

pshomov (approved):
LGTM

PR #4323, opened 2021-06-25
fix(parental-leave): don't show dob when choosing actual date of birth

jeremybarbet:
## What

- don't show dob when choosing actual date of birth

Before:
Screenshot 2021-06-25 at 08 45 55

After:
Screenshot 2021-06-25 at 09 21 33


madebynoam (approved):
lgtm

PR #4325, opened 2021-06-25
chore(core): enable eslint in vscode

jeremybarbet:
## What

Not sure why it has been disabled in the first place? Any reasons why it shouldn't be enable?

## Why

- Improve DX to catch error straight from their IDEs

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,1 @@
-{
-  "eslint.enable": false
-}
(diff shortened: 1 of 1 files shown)


eirikurn (approved):
It might have been causing issues because eslint was taking too much RAM and CPU and killing the IDE.

Recently we improved the resource consumption by 75%, so worth trying this again.

PR #4338, opened 2021-06-25
feat(parental-leave): allow searchable and full width select fields

jeremybarbet:
## What

- Better UX to allow search within select of unions/pension funds/private pension fund

https://user-images.githubusercontent.com/937328/123428084-f3708800-d5b4-11eb-9905-b64859b1672f.mov


madebynoam (approved):
lgtm!

PR #4342, opened 2021-06-25
fix(parental-leave): Fix giveDays not appearing in Review screen

madebynoam:
Fixes an issue where the secondary parent does not see the amount given to the other parent.
https://app.asana.com/0/1182378413629561/1200470778570815/f

## Checklist:


libs/application/templates/parental-leave/src/lib/directorateOfLabour.utils.ts
```diff
@@ -16,7 +16,8 @@ export const DAYS_IN_MONTH = 30
  * Convert a number of days into a floored number of months
  */
 export const daysToMonths = (days: number) => {
-  const fullMonths = Math.floor(days / DAYS_IN_MONTH)
+  const ceilOrFloor = days < 0 ? Math.ceil : Math.floor
```
jeremybarbet, line 19: I wouldn't touch this function directly and handle this case in the SummaryRights component instead where the function is used: `const given = daysToMonths(giveDays)`.

jeremybarbet (changes requested):

libs/application/templates/parental-leave/src/lib/directorateOfLabour.utils.ts
madebynoam: Sounds good, just updated it now. 
I've set it to `const given = daysToMonths(Math.abs(giveDays))`. 
It's using `Math.abs` because `daysToMonths` will floor in the wrong direction that we want:
`daysToMonths(-45) === -2.5` when it should be `1.5`

libs/application/templates/parental-leave/src/fields/Rights/SummaryRights.tsx
```diff
@@ -68,7 +68,7 @@ export const SummaryRights = ({ application }: SummaryRightsProps) => {
               </>
             )}
 
-            {isGivingRights === YES && giveDays > 0 && (
```
jeremybarbet, line 71: Should be back to `>` right?

libs/application/templates/parental-leave/src/lib/directorateOfLabour.utils.spec.ts
```diff
@@ -41,6 +41,13 @@ describe('monthsToDays', () => {
 
     expect(daysToMonths(sum)).toBe(7.5)
   })
+
+  it('should return -1.5 months for the maximum period given to the other parent', () => {
```
jeremybarbet, line 45: This test is probably not necessary anymore

madebynoam: Agree, should be out now.

libs/application/templates/parental-leave/src/fields/Rights/SummaryRights.tsx
madebynoam: `const { giveDays} = useApplicationAnswers(application)` returns a negative value in this scenario:
1. Primary parent applies first
2. Secondary parent approves giving the days
3. Secondary parent applies and reaches the Summary screen.

I think we set it to a negative value because VMLST requested it that way...

So the question is, is the negative value a bug? If not, then we need to keep it `< 0` to show this block...

What do you think?

jeremybarbet: Ah yeah, I read it too fast last time. I thought it was using the `given` variable instead, which now use `Math.abs` hence my comment above.

I believe we need another case then. If the primary parent decides to gives his days, the value will be positive, right? We might just want to check if `giveDays !== 0`

jeremybarbet (approved):
LGTM 👍

libs/application/templates/parental-leave/src/fields/Rights/SummaryRights.tsx
madebynoam: Sounds good, thanks for taking another look. I've changed it to `giveDays !== 0`. I was doing some testing around it, and have a hard time reaching the give rights screen on dev with mock data (but that's a separate issue that I'll continue to investigate for a future PR).

PR #4354, opened 2021-06-27
fix(service-portal): Service portal/UI tweaks

thordurhhh:
## What

Some tweaks to the service portal after a few comments from the island.is product owners.

## Checklist:


libs/service-portal/documents/src/utils/getOrganizationLogoUrl.ts
```diff
@@ -0,0 +1,32 @@
...
+): string => {
+  if (orgs.length > 0) {
+    const qs = String(forName).trim().toLocaleLowerCase()
+    const match =
+      orgs.find((o: Organization) => o.title?.toLocaleLowerCase() === qs) ||
+      orgs.find((o: Organization) => o.logo?.title === 'Skjaldarmerki')
+    const c = match?.logo?.url
+    const url =
+      c ??
+      '//images.ctfassets.net/8k0h54kbe6bj/6XhCz5Ss17OVLxpXNVDxAO/d3d6716bdb9ecdc5041e6baf68b92ba6/coat_of_arms.svg'
```
osk, line 25: Need to revisit, this shouldn't be hard coded here, but we're making the implementation match the apps for v1:
https://github.com/island-is/island.is/blob/a8af0e669c49008df542d2ed4cf00da829fcf0e4/apps/native/app/src/stores/organizations-store.ts#L67

libs/service-portal/graphql/src/lib/queries/getOrganizations.ts
```diff
@@ -0,0 +1,24 @@
...
+      id
+      url
+      title
+    }
+  }
+`
+
+export const GET_ORGANIZATIONS_QUERY = gql`
+  query getOrganizations {
+    getOrganizations(input: { lang: "is-IS", perPage: 200 }) {
```
osk, line 17: Should revisit, getting all orgs when we'll only display a fraction isn't ideal.
But, we're getting all documents and fake-paging.. so we know the whole list of document providers.
Cache in redis?

osk (approved):

PR #4355, opened 2021-06-27
chore(parental-leave): Persist other parent id to answers when spouse is selected

baering:
## What

When the primary parent is creating an application and selects their spouse as the other parent, then the national registry of the other parent is never persisted into answers.otherParentId, which is then used as a search parameter when other parents create their application. Resulting in them getting the error message when starting an application since no application with them registered as the other parent can be found (cannot use complex search parameters until we refactor the state machine into template api modules).

## Why

So `secondary parents` can start their application process.

## Screenshots / Gifs

## Checklist:


jeremybarbet (approved):
LGTM 👍

PR #4357, opened 2021-06-27
fix(party-application): Stop hardcoding assignee ids

siggakogk:
# ...

Use ids for assignees through environment variables 

## Checklist:


libs/application/templates/party-application/src/lib/PartyApplicationTemplate.ts
```diff
@@ -275,12 +275,14 @@ const PartyApplicationTemplate: ApplicationTemplate<
   stateMachineOptions: {
     actions: {
       assignToSupremeCourt: assign((context) => {
+        const supremeCourtIds = process.env.SUPREME_COURT_ASSIGNEES?.split(
```
stefanhar, line 278: Maybe rename `supremeCourt` to `electoralCommission` since `supremeCourt` is `Hæðstiréttur`

libs/application/templates/party-application/src/lib/PartyApplicationTemplate.ts
```diff
@@ -275,12 +275,14 @@ const PartyApplicationTemplate: ApplicationTemplate<
...
+          ',',
+        ) ?? ['']
```
stefanhar, line 280: This should probably default to empty array

libs/application/templates/party-letter/src/lib/PartyLetterApplicationTemplate.ts
```diff
@@ -236,12 +236,15 @@ const PartyLetterApplicationTemplate: ApplicationTemplate<
   stateMachineOptions: {
     actions: {
       assignToMinistryOfJustice: assign((context) => {
+        const ministryOfJusticeIds = process.env.MINISTRY_OF_JUSTICE_ASSIGNEES?.split(
+          ',',
+        ) ?? ['']
```
stefanhar, line 241: This should probably default to empty array

libs/application/templates/party-application/src/lib/PartyApplicationTemplate.ts
stefanhar, line 280: This needs to account for multiple voter regions

libs/application/templates/party-application/src/lib/PartyApplicationTemplate.ts
```diff
@@ -10,7 +10,34 @@ import {
 } from '@island.is/application/core'
 import { dataSchema } from './dataSchema'
 import { assign } from 'xstate'
-import { API_MODULE_ACTIONS, States, Roles } from '../constants'
+import { ApiModuleActions, States, Roles } from '../constants'
+import { EndorsementListTags } from '../constants'
+
+const assignees = (constituency: EndorsementListTags) => {
+  switch (constituency) {
+    case 'partyApplicationReykjavikurkjordaemiSudur2021': {
```
stefanhar, line 18: Maybe use enums for these case values

libs/application/templates/party-application/src/lib/PartyApplicationTemplate.ts
```diff
@@ -279,8 +306,9 @@ const PartyApplicationTemplate: ApplicationTemplate<
           ...context,
           application: {
             ...context.application,
-            // todo: get list of supreme court national ids
-            assignees: ['0000000000'],
+            assignees: assignees(
```
stefanhar, line 309: Maybe rename the `assignees` function here to make it more clear this is a custom function maybe `getConstituencyAssignees`

stefanhar (approved):

apps/application-system/form/infra/application-system-form.ts
```diff
@@ -14,6 +13,22 @@ export const serviceSetup = (services: {}): ServiceBuilder<'application-system-f
         prod: 'https://innskra.island.is',
       },
     })
+    .secrets({
```
sindrig, line 16: This creates 7 new lists to manage, in 3 environments. That's 21 lists that we'll have to maintain.

Is there no way to give users credentials/scopes for this using ids?

stefanhar: We are basing this of existing admin setups.
I belive the most correct way of dong this is using the delegation system (umboðskerfið), but I don't think that's ready yet.
Unless I'm missing something

sindrig: I know this is based off prior config - I'm only pointing out that the current state of things is crazy :(

stefanhar: Agree this is suboptimal but I don't see a better solution here. 
Do you want us to change our approach here to something else?

sindrig: How often do you think these lists will change?

I don't have a better solution right now and it bothers me that I have to approve stuff because we can't do stuff properly, just good enough. Oh well.

sindrig (approved):

apps/application-system/form/infra/application-system-form.ts
stefanhar: Thank you.
Once these list have been set up to include the correct nationalIds, I expect these lists to change every 4 years

sindrig: Hopefully we'll have a better way to do this in 4 years then 😅

stefanhar: 🤞

libs/application/templates/party-application/src/lib/PartyApplicationTemplate.ts
```diff
@@ -10,8 +10,50 @@ import {
...
 import { assign } from 'xstate'
-import { API_MODULE_ACTIONS, States, Roles } from '../constants'
+import { ApiModuleActions, States, Roles } from '../constants'
+import { EndorsementListTags } from '../constants'
 
+const assignees = (constituency: EndorsementListTags) => {
+  switch (constituency) {
+    case 'partyApplicationReykjavikurkjordaemiSudur2021': {
+      return (
+        process.env.PARTY_APPLICATION_RVK_SOUTH_ASSIGNED_ADMINS?.split(',') ??
```
arnorhs, line 20: everytime i read these, i think this is an application for having a party

should we be reading these directly from env variables? would it be better to put them in the environments config file, where they are read with other env variables?

PR #4358, opened 2021-06-27
fix(payment-application): fix ARK_ENDPOINT => ARK_BASEURL

arnorhs:
We had another env variable wrongly assigned. Reading ARK_ENDPOINT, but setting ARK_BASE_URL

ARK_BASE_URL is probably a better name for it, so I created a new secure param in the param store, and used that everywhere.

## Checklist:


jGeir (approved):

apps/application-system/api/infra/application-system-api.ts
```diff
@@ -146,7 +146,7 @@ export const serviceSetup = (services: {
         '/k8s/application-system-api/PAYMENT_BASE_CALLBACK_URL',
       PAYMENT_ADDITION_CALLBACK_URL:
         '/k8s/application-system-api/PAYMENT_ADDITION_CALLBACK_URL',
-      ARK_BASE_URL: '/k8s/application-system-api/ARK_ENDPOINT',
```
sindrig, line 149: So is this parameter not used anymore? Can we delete it?

sindrig (approved):

apps/application-system/api/infra/application-system-api.ts
arnorhs: yes, we probably can. i didn't know if something would break, so i just started by leaving a comment about removing it after today. i'll delete it after this goes through

sindrig: you don't have permissions to delete it, I just created https://github.com/island-is/infrastructure/issues/802

jeremybarbet (approved):

PR #4333, opened 2021-06-25
fix(party-applications): Add error handling before assigning applications

siggakogk:
# ...

Before we assign party letter and party application to ministry of justice and supreme court we try to open/close the  current endorsement list, we should not be sending out the emails if that mutation failed.

## Checklist:

- [ ] I have performed a self-review of my own code
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] Formatting passes locally with my changes
- [ ] I have rebased against main before asking for a review

PR #4362, opened 2021-06-28
chore(core): move aranja codeowners list to not overwrite precedence

jeremybarbet:
## What

- This PR https://github.com/island-is/island.is/pull/4333 was requesting code review from aranja-application while it was already defining its own code owner. I believe it's because of the order precedence.
- Move aranja codeowners on top (the important one here: `/libs/application/template-api-modules/`) so lines like `/libs/application/template-api-modules/src/lib/modules/templates/party-application/` or `/libs/application/template-api-modules/src/lib/modules/templates/party-letter/` are not "overwritten"

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -12,8 +12,33 @@
 # Make sure that team names are correct or everything stops working.
 # Organisation name is @island-is, not @island.is
 
-
 /libs/api/domains/identity/                                                                   @island-is/core
+
+/apps/reference-next-app/                                                                     @island-is/aranja-core
+/libs/api/domains/translations/                                                               @island-is/aranja-core
+/libs/contentful-extensions/                                                                  @island-is/aranja-core @island-is/kosmos-kaos
+/libs/contentful-extensions/translation/                                                      @island-is/aranja-core
+/libs/localization/                                                                           @island-is/aranja-core
+/libs/utils/api/                                                                              @island-is/aranja-core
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -12,9 +12,34 @@
...
 /libs/api/domains/identity/                                                                   @island-is/core
 
+/apps/reference-next-app/                                                                     @island-is/aranja-core
+/libs/api/domains/translations/                                                               @island-is/aranja-core
+/libs/contentful-extensions/                                                                  @island-is/aranja-core @island-is/kosmos-kaos
+/libs/contentful-extensions/translation/                                                      @island-is/aranja-core
+/libs/localization/                                                                           @island-is/aranja-core
+/libs/utils/api/                                                                              @island-is/aranja-core
+/libs/utils/shared/                                                                           @island-is/aranja-core
+/libs/service-portal/settings/access-control                                                  @island-is/aranja-core @island-is/vice-versa
```
eirikurn, line 24: I guess we lose significance here then instead. Should maybe have "top level" project ownership first (service-portal, application-system, web and api), then all the rest grouped by teams?

jeremybarbet: Sounds good to me. I wasn't sure of the best way to go with this. Will update the PR 👍

jeremybarbet: Take 2 is here https://github.com/island-is/island.is/pull/4367

PR #4364, opened 2021-06-28
fix(auth): configure prefix to prevent token conflicts between apps

eirikurn:
## What

Add prefix to the AuthManager in service-portal and application-system-form so they can better coexist when switching back and forth.

## Why

Service portal needs to be able to request a different set of scopes than the application-system and other apps. So they should each have their own token.

For example, Application System Form requests read and write access to applications, while service-portal only requests read access.

These apps should also have separate clients. That's a bit larger change that we'll do later.

## Screenshots / Gifs

## Checklist:

- [x] ~My changes generate no new warnings~
- [x] ~Formatting passes locally with my changes~
- [x] ~I have rebased against main before asking for a review~


jeremybarbet (approved):
Nice 👍

thordurhhh (approved):
lgtm 👍

PR #4376, opened 2021-06-28
fix(ci): helm docker build depends on docker build

sindrig:
We don't want to announce feature deployment endpoint before docker
images have been pushed.
We might even want to postpone it even further, after `deploy`, but
lets see if this isn't enough for now.


pshomov (approved):
LGTM

PR #4377, opened 2021-06-28
chore(core): allow to pass an array of scope 

jeremybarbet:
## What

- Allow passing scope as an array.

## Why

- Better DX to avoid all the string interpolation for enums scope names
- Better readability


libs/auth/react/src/lib/utils/toStringScope.spec.ts
```diff
@@ -0,0 +1,12 @@
+import { toStringScope } from './toStringScope'
+
+describe('toStringScope', () => {
+  it('should return a string for an array of scope', () => {
+    expect(toStringScope(['scope1', 'scope2'])).toBe('scope1 scope2')
+  })
+
+  it('should return an empty string when no scope is defined', () => {
+    expect(toStringScope()).toBe('')
+    expect(toStringScope([])).toBe('')
```
sindrig, line 10: might want an extra case for the empty scope - not sure how ids or other clients might handle a scope like `['something', '', 'something']`? throw error? maybe not the correct place for a check like that...?

libs/auth/react/src/lib/AuthSettings.ts
```diff
@@ -38,6 +38,13 @@ export interface AuthSettings extends UserManagerSettings {
   userStorePrefix?: string
 }
 
+export interface AuthSettingsWithScopes extends AuthSettings {
+  /**
+   * Allow to pass an array of scope needed for the client.
+   */
+  scopes: string[]
```
sindrig, line 45: `scopes` feels kind of strange to me. i know it's a list and all that, but you're just so used to always using `scope` - scope is a collection of permissions, it's not a collection of scopes. right?

jeremybarbet: Yes you are right, I will change that 👍

libs/auth/react/src/lib/utils/toStringScope.spec.ts
jeremybarbet: It would mean in the first place is has been defined as an empty string. That should not happen, right?

sindrig: probably not, just a minor think about it, if it's not a concern just disregard my comment

sindrig (approved):
LGTM

libs/auth/react/src/lib/userManager.ts
```diff
@@ -18,9 +24,12 @@ export const getAuthSettings = (): AuthSettings => {
   return authSettings
 }
 
-export const configure = (settings: AuthSettings) => {
-  authSettings = mergeAuthSettings(settings)
+export const configure = (settings: AuthSettingsWithArrayOfScope) => {
+  const scope = toStringScope(settings.scope)
+
+  authSettings = mergeAuthSettings({ ...settings, scope })
```
eirikurn, line 30: Not a big fan of `AuthSettingsWithArrayOfScope`. 

Would it work to modify AuthSettings to have a scope array, then just pass it to UserManager like this:

```
userManager = new UserManager({
  ...authSettings,
  scope: toStringScope(authSettings.scope),
})
```

Either that, or have the AuthSettings interface (with string array) separate from the UserManager argument, eg:

```
userManager = new UserManager(toUserManagerOptions(authSettings))
```

eirikurn (changes requested):
Nice! Wanted to do this for some time now. Great job!

libs/auth/react/src/lib/AuthSettings.ts
```diff
@@ -12,12 +12,6 @@ export interface AuthSettings extends UserManagerSettings {
    */
   baseUrl?: string
 
-  /**
-   * Base path
-   * Default: Path from baseUrl.
-   */
-  basePath?: string
```
jeremybarbet, line 19: Removed because not used anywhere

libs/auth/react/src/lib/userManager.ts
jeremybarbet: I went for your first solution 👍

eirikurn (approved):

thordurhhh (approved):

PR #4380, opened 2021-06-28
feat(application-system-api): application payment price integrity

jGeir:
## What

To get and list the price(along with other information since I'm doing this) without exposing the request to client.

## Why

Robustness and security

## Checklist:


apps/application-system/api/src/app/modules/payment/payment.controller.ts
```diff
@@ -59,18 +61,39 @@ export class PaymentController {
...
       application_id: applicationId,
       fulfilled: false,
-      amount: 8000,
+      amount: catalog.priceAmount,
+      definition: {
+        chargeItemName: catalog.chargeItemName,
+        chargeItemCode: catalog.chargeItemCode,
+        performingOrganiationID: catalog.performingOrgID,
+        chargeType: catalog.chargeType,
+      } as any,
```
jGeir, line 96: unsure what else I could use instead of any. Currently the definition is listed like this in the model
`  @Column({
    type: DataType.JSONB,
    defaultValue: {},
  })
  @ApiPropertyOptional()
  definition?: string`

apps/application-system/api/src/app/modules/payment/payment.controller.ts
```diff
@@ -59,18 +61,39 @@ export class PaymentController {
   async createCharge(
     @CurrentUser() user: User,
     @Param('application_id', new ParseUUIDPipe()) applicationId: string,
+    @Body() payload: CreateChargeInput,
   ): Promise<CreatePaymentResponseDto> {
     if (!isUuid(applicationId)) {
       throw new BadRequestException(`ApplicationId is on wrong format.`)
     }
+    const allCatalogs = await this.paymentAPI.getCatalogByPerformingOrg(
```
jGeir, line 69: Maybe its better to call the API DOMAINS resolver instead of clients paymentAPI.... ?

apps/application-system/api/src/app/modules/payment/payment.controller.ts
```diff
@@ -59,18 +61,39 @@ export class PaymentController {
...
+      '6509142520',
```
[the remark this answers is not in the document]
jGeir: For now - yes, we want the national ID hard coded.

apps/application-system/api/src/app/modules/payment/payment.service.ts
```diff
@@ -44,18 +67,21 @@ export class PaymentService {
...
+    const paymentDefinition = JSON.parse(payment.definition as string)
     const charge: Charge = {
       // TODO: this needs to be unique, but can only handle 22 or 23 chars
       // should probably be an id or token from the DB charge once implemented
       chargeItemSubject: payment.id.substring(0, 22),
       systemID: 'ISL',
-      performingOrgID: '6509142520',
+      // The OR values can be removed later when the system will be more robust.
+      performingOrgID:
+        paymentDefinition?.performingOrganizationID || '6509142520',
```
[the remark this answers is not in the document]
jGeir: We could put it into a variable - this is a temporary input that will be removed in later updates. So it is up for debate.

apps/application-system/api/src/app/modules/payment/payment.service.ts
```diff
@@ -44,18 +67,21 @@ export class PaymentService {
...
       payeeNationalID: user.nationalId,
-      chargeType: 'AY1',
+      chargeType: paymentDefinition?.chargeType || 'AY1',
       performerNationalID: user.nationalId,
       charges: [
         {
-          chargeItemCode: 'AY110',
+          chargeItemCode: paymentDefinition?.chargeItemCode || 'AY110',
```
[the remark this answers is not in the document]
jGeir: Great observation, again it is up for debate - this is only temporary hard coding, which I would hope to be solved within next two weeks.
Right now the payment is hard coded for Driving Licence's, but it will become generic soon - which means no hard coding.
The debate lies with putting work & crowding env variables temporary or letting it be and perhaps sticking to comments?

libs/application/template-api-modules/src/lib/modules/templates/payable-dummy-template/payable-dummy-template.service.ts
```diff
@@ -16,6 +16,7 @@ export class PayableDummyTemplateService {
     const result = await this.sharedTemplateAPIService.createCharge(
       authorization,
       id,
+      'AY110',
```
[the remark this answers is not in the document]
jGeir: The dummy template service is also temporary - only to allow external partners to test their service. When they finish their testing we will remove it from island.is.
Although - Fair point about the magic string that could be a env var :D

apps/application-system/api/src/app/modules/payment/utils/findItemType.ts
```diff
@@ -0,0 +1,74 @@
+//import { DrivingLicenseType } from '../types/schema'
+
+export const FindItemType = (itemCode: string) => {
```
baering, line 3: Very minor, but shouldn't this use a lower case `f`?

apps/application-system/api/src/app/modules/payment/utils/findItemType.ts
```diff
@@ -0,0 +1,74 @@
...
+      typeId = 'Umsókn um ríkisborgararétt'
+      break
+    case 'L2101':
+      typeId = 'Búsforræðisvottorð'
+      break
+    case 'l3101':
+      typeId = 'Staðfesting áritana'
+      break
+    default:
+      typeId = 'Ökuskírteini'
```
baering, line 71: Should we default to `Ökuskírteini` if there was no matching type ID?

apps/application-system/api/src/app/modules/payment/utils/findItemType.ts
```diff
@@ -0,0 +1,74 @@
...
+    case 'FO146':
+      typeId = 'Endurmat tryggingarfjárhæðar'
+      break
+    case 'L1101':
+      typeId = 'Umsókn um ríkisborgararétt'
+      break
+    case 'L2101':
+      typeId = 'Búsforræðisvottorð'
+      break
+    case 'l3101':
```
baering, line 67: I guess this should be a capital `L`

apps/application-system/api/src/app/modules/payment/payment.controller.ts
baering: I see you're parsing this in another place

```typescript
const parsedDefinition = JSON.parse(payment.definition as string)
```

Shouldn't this then be stringified in the `DTO` and then you can lose the `any`?

```typescript
definition: JSON.stringify({
  chargeItemName: catalog.chargeItemName,
  chargeItemCode: catalog.chargeItemCode,
  performingOrganiationID: catalog.performingOrgID,
  chargeType: catalog.chargeType,
  amount: catalog.priceAmount,
})
```

jGeir: I changed model definition which brought on some other changes.

apps/application-system/api/src/app/modules/payment/payment.model.ts
```diff
@@ -69,8 +69,8 @@ export class Payment extends Model<Application> {
   user4?: string
 
   @Column({
-    type: DataType.JSONB,
-    defaultValue: {},
+    type: DataType.STRING,
+    defaultValue: '',
   })
   @ApiPropertyOptional()
   definition?: string
```
baering, line 76: Let's revert the changes here and change `string` to `object`.

See [application.model](https://github.com/island-is/island.is/blob/main/apps/application-system/api/src/app/modules/application/application.model.ts#L82) as reference.

apps/application-system/api/src/app/modules/payment/payment.controller.ts
```diff
@@ -56,18 +61,58 @@ export class PaymentController {
...
     const paymentDto: Pick<
       BasePayment,
-      'application_id' | 'fulfilled' | 'amount' | 'expires_at'
+      'application_id' | 'fulfilled' | 'amount' | 'definition' | 'expires_at'
     > = {
       application_id: applicationId,
       fulfilled: false,
-      amount: 8000,
+      amount: catalog.priceAmount,
+      definition: `{"chargeItemName":"${catalog.chargeItemName}","chargeItemCode":"${catalog.chargeItemCode}","performingOrganiationID":"${catalog.performingOrgID}","chargeType":"${catalog.chargeType}","amount":"${catalog.priceAmount}"}`,
```
baering, line 115: By changing the model you can revert this to be an object like it was earlier.

apps/application-system/api/src/app/modules/payment/payment-callback.controller.ts
```diff
@@ -16,20 +16,22 @@ export class PaymentCallbackController {
...
       {
         fulfilled: true,
+        reference_id: parsedCallback.callback.receptionID,
       },
       {
         where: {
           id,
           application_id: applicationId,
         },
+        returning: true,
```
arnorhs, line 34: is this necessary?

apps/application-system/api/src/app/modules/payment/payment-callback.controller.ts
```diff
@@ -16,20 +16,22 @@ export class PaymentCallbackController {
     @Body() callback: Callback,
     @Param('id', new ParseUUIDPipe()) id: string,
   ): Promise<void> {
-    if (callback.status !== 'paid') {
+    const parsedCallback = await JSON.parse(JSON.stringify(callback))
```
arnorhs, line 19: i'm not fully understanding the reason for this

apps/application-system/api/src/app/modules/payment/payment.controller.ts
```diff
@@ -56,18 +61,64 @@ export class PaymentController {
...
+        ? await this.paymentAPI.getCatalogByPerformingOrg(
+            DISTRICT_COMMISSIONER_OF_REYKJAVIK,
+          )
+        : await this.paymentAPI.getCatalog()
+
+    // Sort through all catalogs to find the correct one.
+    const catalog = await this.paymentService
+      .searchCorrectCatalog(
+        payload.chargeItemCode,
+        JSON.stringify(allCatalogs.item),
```
arnorhs, line 100: why stringify?

apps/application-system/api/src/app/modules/payment/payment.service.ts
```diff
@@ -44,34 +66,86 @@ export class PaymentService {
...
-    return {
-      ...result,
-      paymentUrl: this.makePaymentUrl(result.user4),
+  async findApplicationById(
+    applicationId: string,
+    nationalId: string,
+    applicationType: string,
+  ): Promise<ApplicationModel> {
+    const application = await ApplicationModel.findAll({
+      raw: true,
```
arnorhs, line 136: why raw?

apps/application-system/api/src/app/modules/payment/utils/findItemType.ts
```diff
@@ -0,0 +1,74 @@
+//import { DrivingLicenseType } from '../types/schema'
+
+export const findItemType = (itemCode: string) => {
+  let typeId = itemCode
+  // AYXXX PayableDummyTemplate should be deleted when Dummy testing has finished.
```
arnorhs, line 5: it's good practice to prefix such comments with `// TODO: ` - a lot of editors can present those in its own panel etc and there's some additional tooling around such things

apps/application-system/api/src/app/modules/payment/utils/findItemType.ts
```diff
@@ -0,0 +1,74 @@
+//import { DrivingLicenseType } from '../types/schema'
```
arnorhs, line 1: delete?

apps/application-system/api/src/app/modules/payment/utils/findItemType.ts
```diff
@@ -0,0 +1,74 @@
...
+
+export const findItemType = (itemCode: string) => {
+  let typeId = itemCode
```
arnorhs, line 4: this is probably useless?

apps/application-system/api/src/app/modules/payment/utils/findItemType.ts
```diff
@@ -0,0 +1,74 @@
...
+  // AYXXX PayableDummyTemplate should be deleted when Dummy testing has finished.
+  switch (typeId) {
+    case 'AYXXX':
+      typeId = 'PayableDummyTemplate'
```
arnorhs, line 8: it might be more elegant to just return the string directly, instead of assigning to a variable name

libs/application/templates/driving-license/src/forms/application.tsx
```diff
@@ -206,7 +206,7 @@ export const application: Form = buildForm({
             ),
             buildCustomField(
               {
-                id: 'healthDeclaration.hasDiabetes',
+                id: 'healthDeclaration.isAlcoholic',
```
arnorhs, line 209: is this a related change?

libs/application/templates/driving-license/src/lib/messages.ts
```diff
@@ -179,54 +179,61 @@ export const m = defineMessages({
   healthDeclaration2: {
     id: 'dl.application:healthDeclaration.2',
     defaultMessage:
-      '2. Hefur þú verið flogaveik(ur) eða orðið fyrir alvarlegri truflun á meðvitund og stjórn hreyfinga?',
+      '2. Hefur þú skert sjónsvið til annarrar hliðar eða beggja?',
```
arnorhs, line 182: seems also unrelated

arnorhs (changes requested):

apps/application-system/api/src/app/modules/payment/utils/findItemType.ts
jGeir: one-entry-one-exit vs elegance.
Both can be argued for.

apps/application-system/api/src/app/modules/payment/payment.controller.ts
jGeir: Otherwise typescript throws this error in the service when trying to parse the catalog..
> Argument of type 'Item[]' is not assignable to parameter of type 'string'.ts(2345)
I could stringify inside the service, but I prefer to do it this way, it is cleaner.

libs/application/template-api-modules/src/lib/modules/templates/driving-license-submission/driving-license-submission.service.ts
```diff
@@ -28,10 +28,17 @@ export class DrivingLicenseSubmissionService {
...
 
   async createCharge({
-    application: { id },
+    application: { id, externalData },
     authorization,
   }: TemplateApiModuleActionProps) {
-    return this.sharedTemplateAPIService.createCharge(authorization, id)
+    const parsedPaymentData = JSON.parse(
+      JSON.stringify(externalData.payment.data),
+    )
```
baering, line 36: What is going on here:

```typescript
JSON.parse(JSON.stringify(...))
```

apps/application-system/api/src/app/modules/payment/payment.controller.ts
baering: Maybe I am lacking some context, but why is the argument of type `string` in the first place? Looks like it is parsed straight away.

```typescript
const resultCatalog = JSON.parse(searchJSON)
```

jGeir: I have it in string format so it can easily be parsed. The service returns a [object, object] thus
I parse&for-loop the object so I can locate the correct catalog.
Is there a better way to go through [object, object] without json parsing?

libs/application/template-api-modules/src/lib/modules/templates/driving-license-submission/driving-license-submission.service.ts
jGeir: I was getting undefined when trying to send the item specifically through externalData.payment.data.chargeItemCode.
Is there a better way to access such data point without risking undefined?

jGeir: f.x
`console.log(externalData.payment.data?.chargeItemCode)`
gives error :

> Property 'chargeItemCode' does not exist on type 'string | number | boolean | object'.
  Property 'chargeItemCode' does not exist on type 'string'.ts(2339)

baering: Yeah so this is because `externalData.data` is of type `string | number | boolean | object`.

I believe you'll just need to tell the typescript compiler what to expect here, do a type cast into the type you expect this to be (since this can be of virtually any type).

Something like:

```typescript
interface PaymentData {
  chargeItemCode: ....
}

// ...

const paymentData = externalData.payment.data as PaymentData
```

apps/application-system/api/src/app/modules/payment/payment.controller.ts
arnorhs: I think the problem is stringifying.. you definitely don't want that.. if you just want to search one of the items, and you are sure you are only getting a single item in allCatalogs, then you probably want something like:

```
const [ itemToSearchFor ] = allCatalogs.item

const catalog = await this.paymentService
      .searchCorrectCatalog(
        payload.chargeItemCode,
        itemToSearchFor,
```

arnorhs (approved):

apps/application-system/api/src/app/modules/payment/utils/findItemType.ts
arnorhs: one-entry-one-exit is an anti-pattern, imo, unless you are dealing with a situation where you need to clean up before exiting a function. readability should be the primary concern.

>“Programs must be written for people to read, and only incidentally for machines to execute.”

😃

apps/application-system/api/src/app/modules/payment/payment.controller.ts
arnorhs: agree with @baering that the argument to the function should probably just be an item, if that's all you are ever searching for.

baering (approved):

PR #4384, opened 2021-06-28
chore(service-portal): Move all general service.portal: prefixed strings to service-portal/core

thordurhhh:
## What

Add all general service.portal translation strings to `service-portal/core` and remove said string prefixes from all other libs within the service portal.

## Why

I'm open for discussions on this, but we definitely need to make some changes in how we handle the strings within the service-portal.

When running `extract-strings` on any service portal module you will deprecate all strings with the `service.portal:` prefix not contained in said module. So it would be better to keep the correct scope of the namespaces.

This way we would contain all strings prefixed with `service.portal:*` in the same place (`/core` in this case) and we would avoid conflicts of importing or deprecating strings when using `extract-strings` on certain modules within the service-portal.

## Checklist:


arnorhs (approved):

maranomynet:
Worse yet, if someone ran `extract-strings` on a single service portal module, it instantly deprecated every `service.portal:*` message, except those few defined by that module 💥

maranomynet (approved):

libs/service-portal/document-provider/src/lib/messages.ts
```diff
@@ -7,27 +7,27 @@ export const m = defineMessages({
     defaultMessage: 'Skjalaveita',
   },
   documentProviders: {
-    id: 'service.portal:document-provider-document-providers',
+    id: 'sp.document-provider:document-provider-document-providers',
     defaultMessage: 'Skjalaveitur',
   },
   documentProviderSingle: {
-    id: 'service.portal:document-provider-document-provider-single',
+    id: 'sp.document-provider:document-provider-document-provider-single',
```
maranomynet, line 14: Since you're changing those ids, doesn't it make sense to remove the duplications?
```suggestion
    id: 'sp.document-provider:document-provider-single',
```

libs/service-portal/driving-license/src/screens/DrivingLicense/DrivingLicense.tsx
```diff
@@ -49,11 +49,11 @@ function DrivingLicense(): JSX.Element {
         {(nationalRegistryUser?.age || 0) >= ELIGABLE_FOR_RENEWAL_AGE && (
           <Application
             heading={defineMessage({
-              id: 'service.portal:driving-license-renewal-heading',
+              id: 'sp.driving-license:driving-license-renewal-heading',
```
maranomynet, line 52: Same here: Since ids are changing, let's de-duplicate the string.

```suggestion
              id: 'sp.driving-license:renewal-heading',
```

Thorrik58 (approved):

thordurhhh:
@island-is/advania - any chance of reviewing this one?

fjandakornid (approved):
LGTM

PR #4385, opened 2021-06-28
feat(services-endorsements-api): Redefined authentication scopes

stefanhar:
#  Redefined authentication scopes

The endorsement system and underlying systems don't use the correct approach to scope implementation.

## What

- Redefined scopes to align with a more preferable scopes approach
- Removed scopes from server side only endpoints
- Removed server side only endpoints from central api
- Updated tests
- Added clients to application api endpoints
- Added a auth header proxy middleware
- Updated environment variables

## Checklist:


libs/auth/scopes/src/lib/endorsements.scope.ts
```diff
@@ -0,0 +1,3 @@
+export enum EndorsementsScope {
+  actor = '@island.is/endorsements',
```
eirikurn, line 2: Let's change this to `main`. We did the same in another PR, and `actor` may cause misunderstandings since the delegation system uses the word `actor` a lot.

eirikurn (changes requested):
Looks good. Hopefully your IDE can do a simple rename for my comment :)

eirikurn (approved):

siggakogk (approved):

pshomov (approved):
LGTM

osk (approved):

PR #4389, opened 2021-06-28
feat(service-portal): Disabling nav item if user does not have access

saevarma:
# Disabling nav item if user does not have access

## What

Disabling nav item in the sidebar if the user is missing the required scope.
Making the nav item grayed out and removed from ReactRouter.

## Why

So the user won't try to navigate to sites which the data fetch is denied because he his missing
the required scopes.

## Screenshots / Gifs

Screen Shot 2021-06-28 at 10 06 08 PM

## Checklist:


eirikurn (approved):
nice!

thordurhhh (approved):
lgtm 👍

PR #4392, opened 2021-06-29
feat(clients-driving-license): driving license client

arnorhs:
Generate driving license API client from openApi doc, and configuring
for environments

## Note re: replacement from/to

The production service has a different api path that in dev, but I don't have an openApi document for the production version, so this is really the only way I found to force the generated client to use the lowercased `api/okuskirteini` in the path.

## Checklist:


libs/api/domains/driving-license/src/lib/drivingLicense.service.ts
```diff
@@ -10,153 +10,207 @@ import {
...
     type: DrivingLicenseType['id'],
   ): Promise<ApplicationEligibility> {
-    const assessmentResult = await this.drivingLicenseApi.getDrivingAssessment(
-      nationalId,
-    )
-    const hasFinishedSchoolResult = await this.drivingLicenseApi.hasFinishedSchool(
-      nationalId,
+    const assessmentResult = await this.getDrivingAssessmentResult(nationalId)
+
+    const hasFinishedSchoolResult = await this.drivingLicenseApi.apiOkuskirteiniKennitalaFinishedokugerdiGet(
```
jGeir, line 203: is there a capital O missing in the end of the function name?
FinishedOkugerdi ?
> Finishedokugerdi

libs/api/domains/driving-license/src/lib/drivingLicense.service.ts
```diff
@@ -191,10 +245,12 @@ export class DrivingLicenseService {
     studentNationalId: string,
     teacherNationalId: User['nationalId'],
   ): Promise<NewDrivingAssessmentResult> {
-    await this.drivingLicenseApi.newDrivingAssessment({
-      kennitala: studentNationalId,
-      kennitalaOkukennara: teacherNationalId,
-      dagsetningMats: new Date(),
+    await this.drivingLicenseApi.apiOkuskirteiniNewDrivingassesmentPost({
```
jGeir, line 248: Should assessment be lower case ?

libs/api/domains/driving-license/src/lib/drivingLicense.service.ts
arnorhs: this is a generated client from the api docs.. if it's like this in the webservice, it's lice that in the function name

libs/api/domains/driving-license/src/lib/drivingLicense.service.ts
arnorhs: generated client.. not really for me to say

jGeir (approved):
LGTM

libs/api/domains/driving-license/src/lib/drivingLicense.service.ts
```diff
@@ -207,23 +263,29 @@ export class DrivingLicenseService {
...
-    const response = await this.drivingLicenseApi.newDrivingLicense({
-      authorityNumber: input.juristictionId,
-      needsToPresentHealthCertificate: input.needsToPresentHealthCertificate
-        ? 1
-        : 0,
-      personIdNumber: nationalId,
-    })
+    const response: unknown = await this.drivingLicenseApi.apiOkuskirteiniApplicationsNewCategoryPost(
+      {
+        category: 'B',
```
jeremybarbet, line 268: I would move that in a configuration file to define the meaning of `B` and be able to reuse it easily if needed

libs/api/domains/driving-license/src/lib/drivingLicense.service.ts
```diff
@@ -207,23 +263,29 @@ export class DrivingLicenseService {
...
+        postNewFinalLicense: {
+          authorityNumber: input.juristictionId,
+          needsToPresentHealthCertificate: input.needsToPresentHealthCertificate
+            ? 1
```
jeremybarbet, line 272: Same here, I guess it's a boolean, but it's easier sometimes to have something like this: (just an example)

```ts
enum HealhCertificate {
  TRUE = 1,
  FALSE = 0,
}
```

libs/api/domains/driving-license/src/lib/drivingLicense.service.ts
```diff
@@ -207,23 +263,29 @@ export class DrivingLicenseService {
...
+          personIdNumber: nationalId,
+        },
+      },
+    )
 
     // Service returns string on error, number on successful/not successful
-    const responseIsString = typeof response !== 'string'
-    const success = responseIsString && response === 1
+    const responseIsString = typeof response === 'string'
+    const success = !responseIsString || response !== 1
```
jeremybarbet, line 281: I would also move the `1` under a meaningful variable

libs/api/domains/driving-license/src/lib/graphql/main.resolver.ts
```diff
@@ -103,11 +107,11 @@ export class MainResolver {
   }
 
   @Query(() => ApplicationEligibility)
-  drivingLicenseApplicationEligibility(
+  async drivingLicenseApplicationEligibility(
```
jeremybarbet, line 110: You don't need the async/await in this case, when the Promise function is directly returned

workspace.json
```diff
@@ -938,6 +938,14 @@
             "passWithNoTests": true
           },
           "outputs": ["coverage/libs/api/domains/driving-license"]
+        },
+        "schemas/external-openapi-generator": {
+          "builder": "@nrwl/workspace:run-commands",
+          "options": {
+            "commands": [
```
jeremybarbet, line 945: It can be simplify to 

```
"command": "nx run clients-driving-license:schemas/openapi-generator"
```

workspace.json
```diff
@@ -2079,6 +2087,14 @@
           "options": {
             "command": "graphql-codegen --config libs/application/templates/driving-assessment-approval/codegen.yml"
           }
+        },
+        "schemas/external-openapi-generator": {
+          "builder": "@nrwl/workspace:run-commands",
+          "options": {
+            "commands": [
```
jeremybarbet, line 2094: same as above

workspace.json
```diff
@@ -2157,6 +2173,14 @@
           "options": {
             "command": "graphql-codegen --config libs/application/templates/driving-license/codegen.yml"
           }
+        },
+        "schemas/external-openapi-generator": {
+          "builder": "@nrwl/workspace:run-commands",
+          "options": {
+            "commands": [
```
jeremybarbet, line 2180: same as above

workspace.json
jeremybarbet: Actually, the following 3 commands are not necessary if you rename 

```
        "schemas/openapi-generator": {
          "options": {
            "command": "yarn openapi-generator -o libs/clients/driving-license/gen/fetch -i libs/clients/driving-license/src/openapi_doc.json"
          }
        }
```

to 

```
        "schemas/external-openapi-generator": {
          "options": {
            "command": "yarn openapi-generator -o libs/clients/driving-license/gen/fetch -i libs/clients/driving-license/src/openapi_doc.json"
          }
        }
```

then it will be executed at the first step of the schema generation and it will be available for the 3 API domains later on

jeremybarbet (changes requested):

libs/api/domains/driving-license/src/lib/drivingLicense.module.ts
```diff
@@ -2,12 +2,38 @@ import { Module, DynamicModule } from '@nestjs/common'
...
   xroadBaseUrl: string
+  xroadPath: string
   xroadClientId: string
   secret: string
+  replaceInPath?: PathReplacement
+}
+
+// This is maybe a bit of a hacky fix for the API being different in production
+// than it is on dev - apparently this is due for an update
+const replacePathMiddleware = ({ from, to }: PathReplacement) => {
```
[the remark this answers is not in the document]
arnorhs: I don't think I have access, but I could get Andes to pull in the openapi doc for production, if it's available there as well. I agree that that would be a much nicer approach

libs/api/domains/driving-license/src/lib/drivingLicense.service.ts
```diff
@@ -10,153 +10,207 @@ import {
...
   NewDrivingLicenseResult,
   NewDrivingAssessmentResult,
+  DeprevationType,
+  RemarkType,
+  RequirementKey,
+  ApplicationEligibility,
 } from './drivingLicense.type'
-import { DrivingLicenseApi, DrivingLicenseResponse } from './client'
+import {
+  AkstursmatDto,
```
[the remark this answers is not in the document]
arnorhs: it seems to be 70/30 .. most in icelandic, some in english

libs/clients/driving-license/src/openapi_doc.json
```diff
@@ -0,0 +1,1712 @@
+{
```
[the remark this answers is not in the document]
arnorhs: ah, ok.. thanks

workspace.json
arnorhs: ah, ok, i'll try that

libs/api/domains/driving-license/src/lib/drivingLicense.module.ts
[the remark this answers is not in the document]
arnorhs: unfortunately it seems the openapi definition is not available in production, I had Petar run the command for me :/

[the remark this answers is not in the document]
arnorhs: I could, of course, manually change the openapi schema.. but that seems like a shaky approach.. esp. when updating the schema

libs/clients/driving-license/src/lib/utils.ts
```diff
@@ -0,0 +1,28 @@
+import fetch from 'isomorphic-fetch'
```
jeremybarbet, line 1: There is a more advanced library for that https://docs.devland.is/libs/clients/middlewares. Unless you really want the success response to be log, it's probably better to use this one

jeremybarbet (approved):
LGTM

apps/api/infra/api.ts
```diff
@@ -132,6 +132,16 @@ export const serviceSetup = (services: {
       PARTY_LETTER_REGISTRY_BASE_API_URL: ref(
         (h) => `http://${h.svc(services.servicesPartyLetterRegistryApi)}`,
       ),
+      DRIVING_LICENSE_PATH_REPLACEMENT_FROM: {
+        prod: '/RafraentOkuskirteini-v1/api/Okuskirteini/',
+        staging: '',
```
pshomov, line 137: Use the constant `MissingSetting` to be more semantic

apps/application-system/api/infra/application-system-api.ts
```diff
@@ -121,6 +121,16 @@ export const serviceSetup = (services: {
       PARTY_LETTER_REGISTRY_API_BASE_PATH: ref(
         (h) => `http://${h.svc(services.servicesPartyLetterRegistryApi)}`,
       ),
+      DRIVING_LICENSE_PATH_REPLACEMENT_FROM: {
+        prod: '/RafraentOkuskirteini-v1/api/Okuskirteini/',
+        staging: '',
```
pshomov, line 126: Shouldn't these be also for all envs?

libs/clients/driving-license/src/lib/utils.ts
```diff
@@ -0,0 +1,28 @@
...
+import { logger } from '@island.is/logging'
+
+// TODO: Copy paste from VMST client - a really helpful helper
+
+export const createWrappedFetchWithLogging = (
+  input: RequestInfo,
+  init?: RequestInit,
+): Promise<Response> => {
+  return new Promise((resolve, reject) => {
+    fetch(input, init)
```
pshomov, line 11: Isn't it strange to wrap `fetch` in a Promise since it is already returning a pretty useful promise?

pshomov (changes requested):
Some env variables are not defined for Staging/prod. They should be or marked as MissingSetting.

apps/api/infra/api.ts
arnorhs: ah ok, thanks

apps/application-system/api/infra/application-system-api.ts
arnorhs: well, there's no staging environment for this api, and prod is the only place we need to do this horrible hack ..

libs/clients/driving-license/src/lib/utils.ts
arnorhs: yes, this is a copy paste from vmst client, without much thought.. @jeremybarbet already pointed me to a shared utils thing that is basically the same thing but better

apps/application-system/api/infra/application-system-api.ts
arnorhs: perhaps, i should have just changed this to use the `isRunningOnEnvironment` helper.. and not bother with any env vars.. this should also be a short-term change, apparently

libs/clients/driving-license/src/lib/utils.ts
arnorhs: i added that one, btw, instead

apps/api/infra/api.ts
pshomov: After talking to @arnorhs turns out empty string is the intended values here.

apps/application-system/api/infra/application-system-api.ts
pshomov: No, please do not use that helper.

pshomov (approved):
LGTM

PR #4394, opened 2021-06-29
fix(ci): Avoid sending message on feature deploy

sindrig:
Feature deployments should not send a message about failed builds to slack


pshomov (approved):
LGTM

PR #4400, opened 2021-06-29
feat(auth-public-api): validTo added to Delegation model

unnarb:
# ...

## What

## Why

## Screenshots / Gifs

## Checklist:


libs/auth-api-lib/src/lib/entities/models/delegation.model.ts
```diff
@@ -48,6 +48,40 @@ export class Delegation extends Model<Delegation> {
...
 
+  get validTo(): Date | null | undefined {
+    // 1. Find a value with null as validTo. Null means that delegation scope set valid not to a specific time period
+    const withNullValue = this.delegationScopes?.find((x) => x.validTo === null)
+    if (withNullValue) {
+      return null
+    }
+
+    // 2. Find the max value
+    try {
```
valurefugl, line 59: why the try/catch?

unnarb: I'm mapping an array that could return null. I'm not sure if that will produce an error or not. I could map it after I see it has some elements
const arrDates = this.delegationScopes
        ?.filter((x) => x.validTo !== null && x.validTo !== undefined)
        .map((x) => x.validTo)

valurefugl (approved):

PR #4407, opened 2021-06-29
feat(judicial-system): laws and restrictions text

VilliGunn:
# Strings for laws and restrictions info bubbles

https://app.asana.com/0/1199153462262248/1200501166205908

## What

- Added strings for laws and restrictions

### Note

I'm not sure this is how we want to define the strings but the thinking behind it is to have it type safe and consistent. All comments are well appreciated 😄 

## Why

- To be able to control through Contentful

## Checklist:


oddsson (approved):
AWESOME 👊

PR #4409, opened 2021-06-29
fix(health-insurance): support single digit month and date

eirikurn:
## What

Fix a date parsing issue in the health-insurance application. Single-digit months and days (`1.1.2021`) in the relocation date weren't parsed correctly, so the date was missing in the upstream API request, possibly resulting in errors.

Made phone and email fields editable, since they can't be edited on my pages yet.

## Why

To fix issues that have been reported in production.

## Checklist:


libs/application/templates/health-insurance/src/forms/HealthInsuranceForm.ts
```diff
@@ -201,19 +200,13 @@ export const HealthInsuranceForm: Form = buildForm({
...
               variant: 'tel',
-              disabled: true,
               format: '###-####',
               placeholder: '000-0000',
               defaultValue: (application: Application) =>
                 (application.externalData.userProfile?.data as {
                   mobilePhoneNumber?: string
                 })?.mobilePhoneNumber,
             }),
-            buildDescriptionField({
```
jeremybarbet, line 212: Why did you delete this description?

eirikurn: It tells the user that the (only) way to change the email/phone number is to go to the Service Portal and change it there.

jeremybarbet (approved):

libs/application/template-api-modules/src/lib/modules/templates/health-insurance/health-insurance.utils.ts
```diff
@@ -23,6 +25,27 @@ const extractAnswerFromJson = (object: unknown, key: string) => {
...
 }
 
+export const parseNationalRegistryDate = (dateString: string | null) => {
+  if (typeof dateString !== 'string') {
+    return undefined
+  }
+
+  // Icelandic format
+  let parsedDate = parse(dateString, 'd.M.yyyy HH:mm:ss', new Date())
+  if (!isNaN(parsedDate.getTime())) {
```
baering, line 35: Will this ever not be true? Since we pass in a fallback date initialised to now.

Thinking if we should rather extract the default date to a variable `const defaultDate = new Date(0)`

And in both of these checks look if `parsedDate.getTime() !== 0`?

PR #4415, opened 2021-06-30
chore(service-portal): Add service-portal WIP module

thordurhhh:
## What

Add a Work In Progress module to display upcoming modules screens.
For example now we have the Finance module behind a feature flag, and there is a request to have the Finance module displayed in the lower sidebar marked and displayed as an incoming feature. And there are more modules who would fit into that category on the horizon.

## Why

To be able to display "work in progress pages" while still being able to have those features as featureFlagged modules.

## Checklist:


libs/service-portal/wip/README.md
```diff
@@ -0,0 +1,7 @@
+# Service Portal Wip
```
sindrig, line 1: consider putting some information about the library here

sindrig (approved):
LGTM

libs/service-portal/wip/README.md
thordurhhh: Absolutely!

Valur (approved):

libs/service-portal/finance/src/screens/PaymentOverview/PaymentOverview.tsx
```diff
@@ -13,7 +13,7 @@ export const PaymentOverview: ServicePortalModuleComponent = () => {
         defaultMessage: 'Greiðslur',
       })}
       intro={defineMessage({
-        id: 'sp.assets:intro',
+        id: 'sp.payments:intro',
```
maranomynet, line 16: `useNamespaces('sp.payments')` is never run.

shouldn't this be `sp.finance:*` ??

libs/service-portal/assets/src/screens/AssetsOverview/AssetsOverview.tsx
```diff
@@ -20,27 +20,6 @@ export const AssetsOverview: ServicePortalModuleComponent = () => {
...
-          defineMessage({
-            id: 'sp.assets:inc-2',
-            defaultMessage: 'Ganga frá skattskýrsla og sjá eldi skattskýrslur',
-          }),
-          defineMessage({
-            id: 'sp.assets:inc-3',
-            defaultMessage: 'Sjá yfirlit og ráðstafa séreignarsparnaði',
-          }),
-        ],
-      }}
```
maranomynet, line 43: What's happening here? How does this relate the WIP module?

maranomynet: ...but while we're at it: Why then is `libs/service-portal/finance/src/screens/PaymentOverview/PaymentOverview.tsx` still full of references to the `sp.assets` without ever running `useNamespaces('sp.asssets')`?

thordurhhh: This relates to the WIP assets module.
But you are right about the namespace. That needs to be changed.

libs/service-portal/finance/src/screens/PaymentOverview/PaymentOverview.tsx
thordurhhh: This module is actually not being used for now so it will default to the defaultMessage.
Not sure if this should be under sp.finance, but it certainly does not fit under sp.assets so I changed it.. 

This will need to be fixed once this module gets back into the project.

PR #4417, opened 2021-06-30
fix(api): RSK environment variables

stefanhar:
# RSK environment variables

The RSK domain relies on these variables being set.

## What

- Added environment variables

## Why

- Fix domain not working in environments

## Checklist:


pshomov (approved):
LGTM

PR #4431, opened 2021-06-30
chore(parental-leave): Only allow one active application per child for secondary parent

baering:
## What

☝️ 

## Why

So that there cannot be many applications for the same child

## Screenshots / Gifs

Also updated the copy for existing application to match new applications

### Before
image

### After
image

## Checklist:


jeremybarbet (approved):
LGTM

PR #4438, opened 2021-06-30
Don't run generate schemas parallel

sindrig:
condi race tion


pshomov (approved):
LGTM

sindrig:
doing https://github.com/island-is/island.is/pull/4439 instead

PR #4133, opened 2021-06-09
feat(service-portal): login and layout

thordurhhh:
## What

New design layout for launch of minar-sidur.

## Screenshots / Gifs

https://www.figma.com/proto/sMnUuOHOGuill3O3hxEICG/H%C3%B6nnun-m%C3%ADnar-s%C3%AD%C3%B0ur?page-id=7259%3A0&node-id=7307%3A1207&viewport=297%2C873%2C0.16715258359909058&scaling=min-zoom

## Checklist:

- [x] I have performed a self-review of my own code
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] Formatting passes locally with my changes
- [x] I have rebased against main before asking for a review

PR #4449, opened 2021-07-01
fix(web): All users should go to Postholf on login

thordurhhh:
## What

All users logging in should go directly to Postholf

## Why

A decision has been made to not go the mobile only way as in https://github.com/island-is/island.is/pull/4133

## Checklist:


stefanhar (approved):
Super 👍

PR #4450, opened 2021-07-01
chore(core): Single registration pattern

darrikonn:
- Identity gql field instead of from / to fields
- Return tmp data until rsk gets implemented
- Make name field readonly
- Query identity instead of from / to
- Lint
- Use single registration pattern

# ...

## What

## Why

## Screenshots / Gifs

## Checklist:


apps/api/src/app/app.module.ts
```diff
@@ -44,6 +44,157 @@ const autoSchemaFile = environment.production
...
+  },
+})
+const nationalRegistryXRoadModule = NationalRegistryXRoadModule.register({
+  xRoadBasePathWithEnv: environment.nationalRegistryXRoad.url,
+  xRoadTjodskraMemberCode: environment.nationalRegistryXRoad.memberCode,
+  xRoadTjodskraApiPath: environment.nationalRegistryXRoad.apiPath,
+  xRoadClientId: environment.nationalRegistryXRoad.clientId,
+})
+const identityModule = IdentityModule.register({
+  modules: [nationalRegistryModule],
```
darrikonn, line 63: Note the changes here.
Just pass in the registered module instead of passing down the environment variables and reregister it in the dependant module:
```typescript
IdentityModule.register({
  nationalRegistry: {
    baseSoapUrl: environment.nationalRegistry.baseSoapUrl,
    user: environment.nationalRegistry.user,
    password: environment.nationalRegistry.password,
    host: environment.nationalRegistry.host,
  },
}),
```

PR #4453, opened 2021-07-01
feat(application-templates-party-application): Email per constituency

stefanhar:
# Email per constituency

Currently we can only send party application submissions to one email.

## What

- Added multiple admin emails
- Fixed admin secret names

## Why

- Allows us to have one email address per constituency

## Checklist:


apps/application-system/api/infra/application-system-api.ts
```diff
@@ -154,20 +154,34 @@ export const serviceSetup = (services: {
...
-        '/k8s/application-system-form/PARTY_APPLICATION_NORTH_ASSIGNED_ADMINS',
+        '/k8s/application-system/api/PARTY_APPLICATION_NORTH_ASSIGNED_ADMINS',
       PARTY_APPLICATION_SOUTH_ASSIGNED_ADMINS:
-        '/k8s/application-system-form/PARTY_APPLICATION_SOUTH_ASSIGNED_ADMINS',
+        '/k8s/application-system/api/PARTY_APPLICATION_SOUTH_ASSIGNED_ADMINS',
       PARTY_LETTER_ASSIGNED_ADMINS:
-        '/k8s/application-system-form/PARTY_LETTER_ASSIGNED_ADMINS',
+        '/k8s/application-system/api/PARTY_LETTER_ASSIGNED_ADMINS',
+
+      PARTY_APPLICATION_RVK_SOUTH_ADMIN_EMAIL:
```
pshomov, line 173: how often are these gonna change 😅?

pshomov (approved):
lgtm

apps/application-system/api/infra/application-system-api.ts
stefanhar: Once when we insert the correct values then every 4 years.
This is a temporary solution some of these variables will be replaced with the delegation system.
Others will be moved to a admin system that should be part of the next iteration of the system

albinagu (approved):
sgtm

libs/application/template-api-modules/src/lib/modules/templates/party-application/party-application.service.ts
```diff
@@ -34,6 +41,10 @@ type EndorsementListResponse =
     }
   | ErrorResponse
 
+export type PartyApplicationServiceOptions = {
```
jeremybarbet, line 44: An interface would be better to describe this type. In general `interface` have better perfs than `type`

stefanhar: Fixed

jeremybarbet (approved):
LGTM

PR #4456, opened 2021-07-01
fix(access-control): fix datepicker

darrikonn:
# Fix datepicker

## What

Allow overflow in table.
Set min width on datepicker in order to render properly

## Screenshots / Gifs

datp

## Checklist:


libs/service-portal/settings/access-control/src/screens/Access/Access.treat.ts
```diff
@@ -28,3 +28,8 @@ export const closeButton = style({
   right: theme.spacing['1'],
   zIndex: 2,
 })
+
+export const table = style({
+  marginBottom: '200px',
```
darrikonn, line 33: Adding some margin to reduce the UI flickering. The needed margin is 350px, but it didn't look good in regards of the footer.

atligudl (approved):

libs/island-ui/core/src/lib/Table/Table.tsx
```diff
@@ -19,11 +19,13 @@ type DataField = {
 type Table = {
   children?: ReactNode
   box?: Omit<UseBoxStylesProps, 'component'>
+  className?: string
 }
 
 export const Table = ({
   children,
   box,
+  className,
```
eirikurn, line 28: We'd like to avoid the className prop as much as possible in `island-ui/core` as it becomes much harder to maintain. Rather add explicit features in the components using typed props.

libs/service-portal/settings/access-control/src/screens/Access/Access.treat.ts
eirikurn: I'd like to find a better solution, as this one adds unnecessary margin and painful scroll issues.

Do you know why the table needs overflow hidden/scroll at all? If so, we probably need to implement a way to render the open datepicker modal outside the table or in a portal, with correct absolute position.

eirikurn (changes requested):

libs/service-portal/settings/access-control/src/screens/Access/Access.treat.ts
darrikonn: Yeah I agree, from our conversation, this was supposed to be a temp solution (because of the summer vacations) until a better solution would be implemented:
```
þarf að vera leyst á meiri global hátt, eins og t.d. í mobile verði þetta modal og þannig. En fínt í bili
```

I didn't want to change the overflow since I didn't know why it was set in the first place and might affect other applications.
I'll remove the className and try to find a better solution.

eirikurn: That's fair. Would be good to git blame and ask the component author sometime.

eirikurn: I'd even prefer a "temporaryDropdownSupport" prop on Table. 💩

darrikonn: Contacted the "author" and as far as he could remember, this was supposed to be for horizontal scrolling; so we can "safely" remove the overflow hidden.

eirikurn: Ahhh makes sense. Horizontal scrolling is actually quite important for responsive tables, e.g. so we can fit this table on mobile. Not that it's ideal UX in any way.

But maybe we can cheat for now with a `disableHorizontalScrolling` prop? :/

darrikonn: The table is wrapped in a div which has: `overflow: auto`. In addition to that, the table inside the div had `overflow: hidden` which I believe is not needed, thus we can remove that.

darrikonn: https://github.com/island-is/island.is/pull/4456/files#diff-bb7eb0ed44e43426c91ec7e387240ed091adba21dd5cef938cd1558b6ec620e9R31

libs/service-portal/settings/access-control/src/screens/Access/Access.treat.ts
```diff
@@ -28,3 +28,9 @@ export const closeButton = style({
   right: theme.spacing['1'],
   zIndex: 2,
 })
+
+export const tempTableDatePickerFix = style({})
+
+globalStyle(`${tempTableDatePickerFix} > div > table`, {
```
eirikurn, line 34: Fyi: (I think) it's also possible to use "selectors:" in `style()` to select related elements, rather than using globalStyles.

eirikurn (approved):

libs/service-portal/settings/access-control/src/screens/Access/Access.treat.ts
darrikonn: Yeah, apparently it's not supported (or at least not the last time I checked).
Can't use the ampersand on the left side in a CSS child combination with treat.
I asked them about it as a side question to an issue I submitted to treat: https://github.com/seek-oss/treat/issues/137

PR #4459, opened 2021-07-02
fix(file-storage): get key from S3 uri and escape special chars from files

jeremybarbet:
## What

- Instead of hardcoding the path to the file we get the key from the S3 uri
- Escape special characters from files URL to avoid any upload issue with S3

File upload field:
Screenshot 2021-07-02 at 09 26 50

`island-is-dev-upload-api` bucket:
Screenshot 2021-07-02 at 09 26 57

Review screen:
Screenshot 2021-07-02 at 09 27 10

`island-is-dev-storage-application-system` bucket:
Screenshot 2021-07-02 at 09 27 32


libs/island-ui/core/src/lib/InputFileUpload/InputFileUpload.tsx
```diff
@@ -28,7 +29,7 @@ export const fileToObject = (
   status?: UploadFileStatus,
 ): UploadFile => {
   return {
-    name: file.name,
+    name: kebabCase(file.name),
```
eirikurn, line 32: Let's store the filename as is (shown to the user) just modify the key we store it in s3 (server side, when creating the presigned post fields).

eirikurn (changes requested):
Lgtm, just move the escape logic

eirikurn (approved):
Great!

darrikonn (approved):

obmagnusson (approved):

PR #4462, opened 2021-07-02
feat(access-control): Error notifying in grant access

darrikonn:
# Error notifying in grant access

## What

Error notifying in grant access

## Why

IMO temporary handling since I believe this should be handled by the GraphQL error link.

## Screenshots / Gifs

error

## Checklist:


libs/service-portal/settings/access-control/src/screens/GrantAccess/GrantAccess.tsx
```diff
@@ -69,12 +70,22 @@ function GrantAccess() {
   }, [authDelegation, setValue])
 
   const onSubmit = handleSubmit(async ({ toNationalId, name }) => {
-    const { data, errors } = await createAuthDelegation({
```
darrikonn, line 72: The error link will throw an error so errors will always be undefined here, thus handled by try catch.

eirikurn (approved):
LGTM

PR #4468, opened 2021-07-02
fix(service-portal): Undefined userInfo error fix

thordurhhh:
## What

Fixes an error in the service-portal when userInfo.scopes is unavailable for mock user.

## Why

When running the service-portal with a mock user you get this error:
`Uncaught (in promise) TypeError: Cannot read property 'includes' of undefined`

## Checklist:


libs/service-portal/applications/src/index.ts
```diff
@@ -15,7 +15,7 @@ export const applicationsModule: ServicePortalModule = {
     {
       name: m.heading,
       path: ServicePortalPath.ApplicationRoot,
-      enabled: userInfo.scopes.includes(ApplicationScope.read),
+      enabled: userInfo?.scopes?.includes(ApplicationScope.read),
```
maranomynet, line 18: the type signature of the `routes` method states that `userInfo` is required.

If it isn't always present in the props, then there's a bug elsewhere and re responsible thing to do is to trace the origin of the bug and fix it.

maranomynet (changes requested):

libs/service-portal/applications/src/index.ts
maranomynet: The bug may very likely be caused by a `strict: false` setting in some up-stream module's `tsconfig`.

Look out for that.

maranomynet: ...or there might be a nasty type assertion somewhere forcing an incorrectly shaped/half-formed object into a tsc state that then gets an undeserved trust thereafter.

thordurhhh: The userInfo is actually there, it's just that the scopes are not available for a mock user.
I'll just change that and undo this change.

libs/auth/react/src/lib/userManager.ts
```diff
@@ -44,6 +44,7 @@ export const configureMock = (user?: MockUser) => {
     profile: { name: 'Mock', locale: 'is', nationalId: '0000000000' },
     expired: false,
     expires_in: 9999,
+    scopes: [],
```
maranomynet, line 47: The larger problems here are all these type assertions that totally mask the possibility of all sorts of other problems.

https://github.com/island-is/island.is/blob/ee4c0520fb05c178e93c98075013c8498474276f/libs/auth/react/src/lib/userManager.ts#L43-L68

`userInfo` should be made to be a `User` straight away, and the issue with the profile property should be fixed.

That then obviates the two `as User` assertions further down.

Finally, the trailing `as unknown) as UserManager` is just super smelly, and ought to go.

maranomynet (changes requested):

libs/auth/react/src/lib/userManager.ts
maranomynet: At the very(!) least we should push any icky type-assertions as far down the property tree as possible, to make it more explicit what were lying about.

thordurhhh: That's a valid point. 
My plan here was to just prevent the error when using the service-portal as a mock user.
Those other changes might be revisited by the code owners at a later time.

maranomynet: Makes sense.
Your changes LGTM!

maranomynet (approved):

eirikurn (approved):

libs/auth/react/src/lib/userManager.ts
eirikurn: UserManager is a huge interface, but it's mostly wrapped (and unit tested) inside this library.

Elsewhere I've started adding our User types to express `request.user` and oicd-client `user.profile`. Hopefully we can apply that everywhere.

Maybe it would be better to completely wrap (hide) oidc-client, but it has so many features that I didn't dare yet.

PR #4473, opened 2021-07-02
fix(parental-leave): Fix how Duration component initialises duration

baering:
## What

Since calculated period duration has to be calculated on the API with VMST month logic we cannot initialise this variable using normal date functions.

## Why

So that when you navigate back and forth between selecting period length and period ratio, the length of the period remains the same.

### Before

https://user-images.githubusercontent.com/6312838/124296945-0e5d7200-db4a-11eb-8b4b-1701ef27768e.mov

### After

https://user-images.githubusercontent.com/6312838/124296756-d7875c00-db49-11eb-8dae-738d3a772a0a.mov

## Checklist:


jeremybarbet (approved):
LGTM

PR #4478, opened 2021-07-02
fix(application-template-party-application): Seeder packages

stefanhar:
# Seeder packages

Seeders are missing some packages to run


pshomov (approved):
LGTM

PR #4479, opened 2021-07-02
fix(application-template-party-application): Seeder packages

stefanhar:
# Seeder packages

Merges:
fix(application-template-party-application): Seeder packages #4478


arnorhs (commented):
lgtm

PR #4480, opened 2021-07-03
fix(islandis-login): fixing all island.is login validations relying on islandis-login

arnorhs:
## Checklist:


pshomov (approved):
lgtm

apps/air-discount-scheme/backend/src/app/modules/flight/flight.controller.ts
```diff
@@ -25,7 +25,6 @@ import {
   ApiResponse,
 } from '@nestjs/swagger'
 
-import IslandisLogin, { VerifyResult } from 'islandis-login'
```
eirikurn, line 28: Unused?

eirikurn (approved):

apps/air-discount-scheme/backend/src/app/modules/flight/flight.controller.ts
eirikurn: Yup

arnorhs: apparently, yes :)

PR #4485, opened 2021-07-05
fix(crc): hide mock data section

VilliGunn:
Hotfix hiding mock data section so we don't display the `Continue` and `Back` buttons in places where they shouldn't be present.


baldur-kolibri (approved):
lgtm

PR #4487, opened 2021-07-05
chore(parental-leave): Send empty union value to API if no union was selected

baering:
## What

Union is optional, in the case when no union is selected, send the default 'no union selected' (`F000`)

## Why

To not get an error when sending an application because it is required in the API

## Checklist:

### DIFF
--- a/libs/application/templates/parental-leave/src/fields/Review/index.tsx
+++ b/libs/application/templates/parental-leave/src/fields/Review/index.tsx
@@ -1,3 +1,5 @@
+/* eslint-disable @typescript-eslint/ban-types */
+
 import React, { FC, useMemo } from 'react'
 
 import {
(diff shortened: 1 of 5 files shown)


libs/application/templates/parental-leave/src/fields/Review/index.tsx
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable @typescript-eslint/ban-types */
```
baering, line 1: Prettier tries to change `Boolean` to `boolean`

jeremybarbet: Arf, yeah, Boolean is also probably understood by Typescript internally, so it probably can't find our type defintion

jeremybarbet (approved):
LGTM 👍

PR #4495, opened 2021-07-06
fix(parental-leave): use percentage calculated by VMST api

jeremybarbet:
## What

- Use percentage calculated by VMST API
- Refactor `calculatePercentage` functions to depend on API as much as possible
- Loading state for buttons when loading VMST API calls
- Persisting percentage on periods to be reused within select

https://user-images.githubusercontent.com/937328/124768308-16415b80-df28-11eb-9d3e-924167231137.mov


libs/application/templates/parental-leave/src/fields/Duration/index.tsx
```diff
@@ -18,21 +17,22 @@ import { FieldDescription } from '@island.is/shared/form-fields'
...
 import { parentalLeaveFormMessages } from '../../lib/messages'
 import { usageMaxMonths, usageMinMonths } from '../../config'
 import { StartDateOptions } from '../../constants'
 import { monthsToDays } from '../../lib/directorateOfLabour.utils'
 import { useGetOrRequestEndDates } from '../../hooks/useGetOrRequestEndDates'
 
 const df = 'yyyy-MM-dd'
-
 const DEFAULT_PERIOD_LENGTH = 1
+const DEFAULT_PERIOD_PERCENTAGE = 100
```
jeremybarbet, line 29: I know, it might be not perfect, if someone is creating an application and have less than 1 month of rights. We could move that into a VMST API madness hooks to get startDate/endDate/percentage from multiple calls

baering (approved):
LGTM

PR #4496, opened 2021-07-06
feat(service-portal): Add tracking to login buttons on /innskraning

thordurhhh:
## What

Add tracking to login buttons on /innskraning to trigger a button press registered in [Plausible](https://plausible.io/)

## Why

Island.is team wants to see the difference between how many people press the button for the old login vs the beta login.

I have created a custom event to track these buttons.
  I am still a little unsure if this is the best way to go about the tracking for the Beta button. With a little added complexion in case of Adblockers.

   Looking at the changes in Login.tsx on the link to the beta site (line [#102](https://github.com/island-is/island.is/pull/4496/commits/715e84855212e22f17f0d2ff7e152b305e8d4c49#diff-10d1f168bc7b5a4066e3cc74df852dc70b80e15ad5f54b5f99eabc55eb2ab389R102) & [#62](https://github.com/island-is/island.is/pull/4496/commits/715e84855212e22f17f0d2ff7e152b305e8d4c49#diff-10d1f168bc7b5a4066e3cc74df852dc70b80e15ad5f54b5f99eabc55eb2ab389R62)):  
   
   There are obviously other ways to go about this, I will happily change this if the other options would be more viable.
 Other options for example: 

- Using the outbound-link https://plausible.io/docs/outbound-link-click-tracking instead of a custom event. The drawback I saw from that was that the /minarsidur link is not really an outbound link for the user and should probably not belong with the outbound link data in Plausible. Plus we would need to add a different (little bit bigger) plausible .js file to the web than we have today.
- Opens in new tab. We could open the /minarsidur page in a new tab and therefore remove the need for the callback function and custom window.location. With the obvious drawback of opening a new tab when that’s not how the button functions today, would be a bit strange to change that behaviour just for this.
- Keep it as is .. possibly with some modifications.

If anyone has a cleaner way to go about this, then please, I’m all ears.

## Checklist:


arnorhs (approved):

thordurhhh:
Just a little heads up @arnorhs I have significant changes since your Approval.

thordurhhh:
We kinda need to get this functionality out so a review would be highly appreciated 🙏

jeremybarbet (approved):
LGTM

PR #4505, opened 2021-07-07
fix(judicial-system): Show default message for defender name in request PDF

oddsson:
# Show default message for defender name in request PDF

https://app.asana.com/0/1199153462262248/1200565003571741

## What

The default message "Hefur ekki verið skráður" was not showing in the request PDF when a defender had not been set. This happened because we were using the `??` operator. 

**Example**
`const a = "" ?? "PRINT"`

In this example ☝️ `a` should be "PRINT", right? Apparently not 🤷‍♂️  It seems like ?? uses a non-strict comparison because `"" === false` is definitely `true`.  Anyway, I think we should steer clear of the ?? operator from now on.

## Why

It's a bug.

## Screenshots / Gifs

Screen Shot 2021-07-07 at 08 47 02

## Checklist:


baldur-kolibri (approved):
lgtm

PR #4508, opened 2021-07-07
chore(parental-leave): rename boolean type

jeremybarbet:
## What

- rename boolean type name


baering (approved):
Nice, thanks for this.

PR #4524, opened 2021-07-08
fix(parental-leave): Send an empty true date of birth

baering:
## What

The API only supports `dateOfBirth` when children have been born. We do not have this data at the moment and will revist when we do.

Will revisit this field in another PR where integrating custody API from Þjóðskrá as well as connecting estimated dates of birth to national registry ids of born children.

## Why

So that we do not get an error when sending applications for unborn children.

## Checklist:

### DIFF
--- a/libs/application/template-api-modules/src/lib/modules/templates/parental-leave/parental-leave.utils.ts
+++ b/libs/application/template-api-modules/src/lib/modules/templates/parental-leave/parental-leave.utils.ts
@@ -230,7 +230,7 @@ export const transformApplicationToParentalLeaveDTO = (
     expectedDateOfBirth: selectedChild.expectedDateOfBirth,
     // TODO: get true date of birth, not expected
     // will get it from a new Þjóðskrá API (returns children in custody of a national registry id)
-    dateOfBirth: selectedChild.expectedDateOfBirth,
+    dateOfBirth: '',
     email,
     phoneNumber,
     paymentInfo: {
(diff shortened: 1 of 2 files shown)


libs/clients/vmst/src/clientConfig.yaml
```diff
@@ -491,8 +489,8 @@ components:
           type: integer
           format: int32
         transferableMonths:
-          type: integer
-          format: int32
+          type: number
+          format: float
```
baering, line 493: I updated the client config and this was in the new config, although unrelated to this PR.

libs/application/template-api-modules/src/lib/modules/templates/parental-leave/parental-leave.utils.ts
```diff
@@ -230,7 +230,7 @@ export const transformApplicationToParentalLeaveDTO = (
     expectedDateOfBirth: selectedChild.expectedDateOfBirth,
     // TODO: get true date of birth, not expected
     // will get it from a new Þjóðskrá API (returns children in custody of a national registry id)
-    dateOfBirth: selectedChild.expectedDateOfBirth,
+    dateOfBirth: '',
```
eirikurn, line 233: Should this be undefined instead? Or does that break the validation?

baering: The C# API that receives this uses `isNullOrEmpty` so I decided to have it empty.

eirikurn (approved):

PR #4541, opened 2021-07-10
fix: Schema generation more reliable

### LINKED ISSUE #4533: Fixing unreliable schema generation
Schema generation has been flaky failing for no good reason. Needs to be not that.

pshomov:
Fixes #4533 

## What

Schema generation was sometimes failing unpredictably which I think was a combination of two things which are fixed with this PR:
  * Default NX cache is stored under `node_modules/.cache` which gets stored in the GitHub Actions cache as part of the node_modules cache we create
  * The flag `--skip-nx-cache` was not propagated in nested `nx run ...` executions
I suspect these two things in combination created situations where parts of the schema were retrieved from a cache where they should have been generated.

## Why

Consistent, clean schema generation.

## Screenshots / Gifs

## Checklist:


kksteini (approved):
LGTM

jeremybarbet (approved):

eirikurn:
Not a big fan of this. We should embrace the nx cache to speed up schema generation and builds, both locally and in ci. I thought we (Sindri) had switched to NX Cloud though instead of storing it in node_modules.

If the cache is breaking, there's probably an issue with the nx config or cache invalidation (hard problems, I know).

pshomov:
> Not a big fan of this. We should embrace the nx cache to speed up schema generation and builds, both locally and in ci. I thought we (Sindri) had switched to NX Cloud though instead of storing it in node_modules.
> 
> If the cache is breaking, there's probably an issue with the nx config or cache invalidation (hard problems, I know).

We were using the local NX cache in an unreliable way, that's what I am trying to fix here. Using the cache in the CI should happen using their cloud cache, not the builtin one. I do not recall what was the reason for not embracing it at the time. Also as is it is slowing down the docker image building since the cache is part of the node_modules cache.

busla (approved):
LGTM

PR #4553, opened 2021-07-14
fix(clients-driving-license): updating driving license endpoints

arnorhs:
The service has been updated to use the same casing for dev/prod, so this should address that change.

I'm not going to add automerge until RLS has allowed access to our client ID, and then I'll add it

Edit: It's open now, and I am able to call some of the endpoints now, so I believe it should be fixed

## Checklist:


libs/api/domains/driving-license/src/lib/drivingLicense.service.ts
```diff
@@ -166,6 +166,7 @@ export class DrivingLicenseService {
         kennitala: nationalId,
       },
     )) as unknown) as string
+    console.log({ statusStr })
```
pshomov, line 169: Remove? or use logging lib

pshomov (changes requested):
LGTM, small cleanup needed

albinagu (approved):
lgtm

libs/api/domains/driving-license/src/lib/drivingLicense.service.ts
arnorhs: oops, ty

arnorhs: should be fixed

jeremybarbet (approved):
Nice! LGTM

pshomov (approved):
LGTM

PR #4555, opened 2021-07-14
chore(core): add polyfill for intl-locale

jeremybarbet:
## What

- Polyfill for Intl Locale to fix `Intl.Locale is not a constructor` https://app.asana.com/0/1182378413629561/1200596650522421
- Update @formatjs package. The breaking changes are not impacting us


libs/localization/src/lib/polyfills.ts
```diff
@@ -35,35 +33,33 @@ const localeDataModules = {
...
+const maybePolyfillNumberFormat = async () => {
+  await import(
+    /* webpackChunkName: "intl-numberformat" */ '@formatjs/intl-numberformat/polyfill'
+  )
+}
+
+const maybePolyfillDateTimeFormat = async () => {
+  await import(
+    /* webpackChunkName: "intl-datetimeformat" */ '@formatjs/intl-datetimeformat/polyfill'
+  )
```
eirikurn, line 51: A few issues:

1. Major: It is common for browsers to support Intl and uncommon for them to support Icelandic. For us to polyfill Icelandic, we need to force-load the polyfill first (with polyfill-force). I.e. this may break the Icelandic locale on Chrome.
2. Minor: We would like to shave of some KBs if the browser supports Intl AND the chosen locale.
3. Minor: If we're always going to load all the polyfills, we should not dynamically import them from multiple different chunks. Either just import them normally, or configure the same chunk name for all of them.

eirikurn (changes requested):
Need to address some things.

Do you think we could revert to how it was, add the intl-locale polyfill and simplify the conditional logic so we either load all or nothing?

jeremybarbet:
I see, makes sense. I just wanted to remove intl-locales-supported because it has be migrated to the intl packages. I will check this out

jeremybarbet:
- I reverted back to the initial code
- Use the same chunkName for all polyfills
- Added the IntlLocale polyfill, and checked the supported locale
- I'm not sure it's answering your comment about "simplify the conditional logic so we either load all or nothing?". I'm not sure I get it, if we test individually each polyfill if they are supported, can we load all or nothing?

eirikurn (approved):
LGTM.

I was just wondering if we should in fact conditionally load each polyfill or all. Maybe it depends on if you can actually get the intl-datetimeformat polyfill-force without first getting intl-locale polyfill ... for example.

PR #4560, opened 2021-07-15
fix(application-system): configure expiry by application for the assignEmail token

jeremybarbet:
## What

- Configure the assignEmail expiry by application

### DIFF
--- a/libs/application/template-api-modules/src/lib/modules/shared/shared.service.ts
+++ b/libs/application/template-api-modules/src/lib/modules/shared/shared.service.ts
@@ -54,11 +54,13 @@ export class SharedTemplateApiService {
   async assignApplicationThroughEmail(
     templateGenerator: AssignmentEmailTemplateGenerator,
     application: Application,
+    expiresIn: number,
     locale = 'is',
   ) {
     const token = createAssignToken(
       application,
       getConfigValue(this.configService, 'jwtSecret'),
+      expiresIn,
     )
 
     const clientLocationOrigin = getConfigValue(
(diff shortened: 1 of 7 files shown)


pshomov (commented):
Can I get a bit of explanation what is this token for and why do we want it not to expire. Non-expiring tokens sounds like a time-unbound access.

jeremybarbet:
Sure, when we create an application, depending on the state you end up in, it might need a review from someone else. In this case, we assign the application to someone else through email.

For example, in the parental leave, the applicant adds their employer's email address and asks for 6 months off. It then sends an email to the employer to approves these periods. However, we don't know when the employer will open this email to review the application, hence the no expiry on the token. 

With the previous code, if the employer didn't open the email on time, the token was expired and the application was in a "broken" state where the employer couldn't access and give a review on the application.

I was also thinking to regenerate a new token if we catch an expired token and resend an email to the assigned person. I'm not sure what's the standard for this kind of behavior in general.

If you see a better solution from a security standpoint, just let me know.

kksteini (approved):
LGTM

apps/application-system/api/infra/application-system-api.ts
```diff
@@ -3,6 +3,7 @@ import { ref, service, ServiceBuilder } from '../../../../infra/src/dsl/dsl'
 const postgresInfo = {
   passwordSecret: '/k8s/application-system/api/DB_PASSWORD',
 }
+
```
pshomov, line 6: revert please

pshomov (approved):
LGTM

PR #4566, opened 2021-07-16
feat(application-system): [RFC] new api action to retry assigning application to email

jeremybarbet:
## What

- New api action when assignEmail token expires

## Why

Through the application system we have the ability to assign the application to an email address. It's useful when external people (employer, family member, etc) have to do an action to continue the process.

For instance, in the parental leave application, if the applicant has an employer, they will need to add its email address, once submitting the application it will send an email to the employer to open the application and review the periods chosen by the applicant for its parental leave.

The link for the assignment to the employer contains a token that has an expiry date:
- When the person clicks on the link it will redirect to the application id and execute a specific state machine action to change any state defined by the application. In the case of the parental leave, once the link is clicked, the employer will be added as an "assignee" and the state will change from "employerWaitingToAssign" to "employerApproval".
- If the employer doesn't open the link on time they won't be able to access the application anymore. In this case, we are stuck in an "employerWaitingToAssign" state and we don't have any mechanism in place to unblock the situation, apart from manual DB update.

This PR adds the ability to define a new API action to re-send an email if the token expired when the employer (in our example case) open the link.

## Comment

- Let's say the link has been found by someone that shouldn't have it:
	- If the token is still valid, a check is made against the nationalId to verify it has access to the application otherwise it returns a 403
	- If the token is not valid anymore, I use it to query the application and get the email of the person that should be assigned (from the answers DTO), then I send again an email to this address. Does this have an impact to use a user token to query the data in this case?


apps/application-system/api/src/app/modules/application/application.controller.ts
```diff
@@ -301,30 +304,91 @@ export class ApplicationController {
...
+          baseApplication,
+          template,
+          user.authorization,
+          retryOnAssignApplicationErrorAction,
+        )
+
+        throw new BadRequestException('Token expired', 'retry')
+      }
+
+      this.logger.error(
```
eirikurn, line 350: User errors should be logged as warning.

apps/application-system/api/src/app/modules/application/application.controller.ts
```diff
@@ -301,30 +304,91 @@ export class ApplicationController {
...
+      throw new BadRequestException('Token expired')
+    }
+
+    // We get another error from the token, we return a BadRequestException and logs some info about the application concerned
+    if (verifiedToken === null) {
+      const decodedToken = decodeToken<DecodedAssignmentToken>(
+        assignApplicationDto.token,
+      )
+
+      this.logger.error('Error when trying to verify the token', {
```
eirikurn, line 368: User errors should be logged as warning.

apps/application-system/api/src/app/modules/application/utils/tokenUtils.ts
```diff
@@ -1,16 +1,32 @@
...
     ) as unknown) as T
+  } catch (e) {
+    const errors = e as VerifyErrors | null
+
+    if (errors?.name === 'TokenExpiredError') {
+      return 'TokenExpiredError'
+    }
 
-    return decoded
+    return null
```
eirikurn, line 20: Not a fan of swallowing errors, losing information and returning a union type. It's better to throw error objects.

Maybe in this case you could extend the error with the decoded token (though in that case we might need to log and swallow a decode error).

```
type TokenError<T> = VerifyErrors & { decoded?: T }

// in verifyToken
catch (e) {
  const verifyError = e as TokenError<T>
  try {
    verifyError.decoded = jwt.decode(token) as unknown as T
  } catch (decodeError) {
    logger.warn('Cannot decode token', e)
  }
  throw verifyError
}
```

I don't know 🤷‍♂️ It's at least good to somewhere log the name/message of the jwt verify error.

eirikurn (commented):
I feel like it's a bit weird that the system sends an email automatically when opening an expired link. It can cause spam if someone refreshes the link.

It's also weird to have to implement this retry action but I guess you could always point to the same api action or `assignEmployer` in this case.

The common UX is for the sender to see the status of the "invitation" and have an option to resend the email. They might also want to go back to the previous state to fix some mistake, eg with the employer email.

So I would suggest taking a step back and see if we can do this in the form somehow, ideally with shared fields that can be reused between projects. Does that make sense? What do you think @baering?

Maybe a longer expiry is enough for now?

jeremybarbet:
> I feel like it's a bit weird that the system sends an email automatically when opening an expired link. It can cause spam if someone refreshes the link.

Yeah, it's the case at the moment, it will spam refresh email. I believe there is a way to prevent that but I didn't want to focus on that too much, and rather wanted to show the idea.

> It's also weird to have to implement this retry action but I guess you could always point to the same api action or `assignEmployer` in this case.

Yes, that's correct, would probably be better in this case, it would make more sense.

> The common UX is for the sender to see the status of the "invitation" and have an option to resend the email. They might also want to go back to the previous state to fix some mistakes, eg with the employer email.

We are seeing it with the current test group, the applicants are not aware of the status view. I would suggest fixing this by showing the status of the application straight after it has been sent. At the moment once the application is done, we have a success message, but we need a page to reload to show the status view. If the applicant sees straight away the status view then having an "invitation pending" would make a lot of sense. They will also be more aware of how to follow the status of their application. I believe once we have a notification system in place, we could also send a notification that the invitation expired and has to be resent by the applicant.
 
> So I would suggest taking a step back and see if we can do this in the form somehow, ideally with shared fields that can be reused between projects. Does that make sense? What do you think @baering?
> 
> Maybe a longer expiry is enough for now?

Yeah, I already implemented a longer expiry token as a first solution, it now lasts 6 months. I was just trying to think of a better/future-proof solution. I like the idea of moving this into the status of the application and can be controlled manually by the applicant.

PR #4580, opened 2021-07-23
feat(an): Location and purpose screen added

Thorrik58:
# Location and purpose screen added

https://app.clickup.com/t/nz7xw2

## What

Adding the location and purpose screen.

## Why

User needs to be able to specify location and purpose of his/her accident.

## Screenshots / Gifs

image

## Checklist:


libs/application/templates/accident-notification/src/lib/dataSchema.ts
```diff
@@ -88,6 +88,12 @@ export const AccidentNotificationSchema = z.object({
     email: z.string().optional(),
     phoneNumber: z.string().optional(),
   }),
+  locationAndPurpose: z.object({
+    location: z.string().nonempty(),
```
sigruntg, line 92: Maybe we should start using something else than nonempty since it's deprecated.
image

sigruntg (changes requested):
Maybe we should start using something else than nonempty now since we know that it's deprecated.

libs/application/templates/accident-notification/src/lib/dataSchema.ts
Thorrik58: Fixed and replaced all usage of nonempty on accident notification

sigruntg (approved):
Looks cool!!

PR #4582, opened 2021-07-26
fix(crc): change prefix for correct template's name

jeremybarbet:
## What

- Use another title to differentiate logs


Rikkiabb (approved):
Thanks 👏

PR #4589, opened 2021-07-27
feat(pdpp): overview screen

sigruntg:
# Overview screen

https://app.clickup.com/t/j59jeq

## What

Overview screen that you can edit from.

## Why

So that the user can have an overview over what he/she is submittig.

## Screenshots / Gifs

image

## Checklist:


libs/application/templates/public-debt-payment-plan/src/lib/dataSchema.ts
```diff
@@ -86,6 +106,20 @@ export const PublicDebtPaymentPlanSchema = z.object({
   }),
 })
 
+export type PaymentPlanBuildIndex = 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9
```
Thorrik58, line 109: Should this be in the types folder?

libs/application/templates/public-debt-payment-plan/src/fields/Overview/Overview.tsx
```diff
@@ -1,78 +1,147 @@
...
+                <Text>{employerInfo.name}</Text>
+              </Box>
+            </GridColumn>
+          )}
+          {employerInfo?.nationalId && (
+            <GridColumn span={['6/12', '5/12']}>
+              <Box>
+                <Label>{formatMessage(overview.employerSsn)}</Label>
+                <Text>
+                  {employer?.isCorrectInfo === YES
```
Thorrik58, line 102: We have been exporting the answers to a constant file where they are defined like:
export const YES = 'yes'
export const NO = 'no'
And then importing that constant in the fields needed to have a defined constant instead of hardcoded string

libs/application/templates/public-debt-payment-plan/src/fields/Overview/Overview.tsx
```diff
@@ -1,78 +1,147 @@
...
-              <Text>Krónan ehf.</Text>
-            </Box>
-          </GridColumn>
-          <GridColumn span={['6/12', '5/12']}>
-            <Box>
-              <Label>Kennitala vinnuveitanda</Label>
-              <Text>711298-2239</Text>
-            </Box>
-          </GridColumn>
+          {employer?.isCorrectInfo === YES && (
```
Thorrik58, line 89: Same as below

libs/application/templates/public-debt-payment-plan/src/lib/paymentPlanUtils.ts
```diff
@@ -23,3 +29,68 @@ export const prerequisitesFailed = (data: ExternalData) => {
...
+  useEffect(() => {
+    setIsLoading(true)
+    getDistributionCallback({
+      monthAmount,
+      monthCount,
+      totalAmount,
+      scheduleType,
+    })
+      .then((response) => {
+        if (!response) return
```
Thorrik58, line 77: Should this be handled somehow to show the user that no response was received?

libs/application/templates/public-debt-payment-plan/src/fields/Overview/DistributionTabel.tsx
```diff
@@ -0,0 +1,26 @@
...
+  totalAmount,
+  scheduleType,
+}: PaymentDistribution) => {
+  const { isLoading, distributionData } = useDistributionTable({
+    monthAmount,
+    monthCount,
+    totalAmount,
+    scheduleType,
+  })
+  if (!distributionData) return null
```
Thorrik58, line 18: Do we need to display something to the client in this case?

Thorrik58 (changes requested):
Minor comments, looks good tho

libs/application/templates/public-debt-payment-plan/src/lib/paymentPlanUtils.ts
sigruntg: I'm going to remove this. I don't think this should ever fail. But there is a lot of rethinking that needs to take place when people come from summer break when it comes to the payment plan.

libs/application/templates/public-debt-payment-plan/src/fields/Overview/DistributionTabel.tsx
sigruntg: Maybe, but I think we need to rethink errors later. There are so many questions I have regarding this form, that I think many things will change later regarding errors.

Thorrik58 (approved):

PR #4591, opened 2021-07-28
fix(core): use european date format

jeremybarbet:
## What

Dates were using American format `MM/dd/yyyy`, move to `dd/MM/yyyy` for European standard


eirikurn (approved):
Ugh, so confusing. :( Wish we could make this better somehow.

PR #4608, opened 2021-08-03
refactor(core): move cms and translations libraries

jeremybarbet:
## What

- The cms and translations libraries coming from the API domains folder are now used in other projects, e.g. application-system-api, judicial-system-api. Taking it out at the root of the libs folder


oddsson (approved):

sindrig (approved):
LGTM

barabrian (approved):
LGTM

eirikurn (approved):
Personally I would split these modules so the graphql resolvers are still in domain libs, while the services and models can be in shared libs.

But fine to get this merged in as is.

PR #4617, opened 2021-08-05
feat(an): Application confirmation email

Toti91:
# Accident Notification - Confirmation email

https://app.clickup.com/t/pzdj6p

## What

Set up template API module for the application and added functionality that sends confirmation email when form is submitted.

## Why

So that the applicant receives a confirmation when he submits the form.

## Checklist:


libs/application/template-api-modules/src/lib/modules/templates/accident-notification/accident-notification.service.ts
```diff
@@ -0,0 +1,84 @@
...
+
+        const signedUrl = await this.fileStorageService.generateSignedUrl(url)
+
+        return { name, url: signedUrl }
+      }),
+    )
+  }
+
+  private getFilesFromAnswers(answers: FormValue, id: string) {
+    return getValueViaPath(answers, id) as Array<{ key: string; name: string }>
```
eirikurn, line 82: Can fallback to empty array.

libs/application/template-api-modules/src/lib/modules/templates/accident-notification/accident-notification.service.ts
```diff
@@ -0,0 +1,84 @@
...
+    const injuryCertificateFiles = this.getFilesFromAnswers(
+      application.answers,
+      'attachments.injuryCertificateFile',
+    )
+
+    const attachments = []
+
+    if (powerOfAttorneyFiles) attachments.push(...powerOfAttorneyFiles)
+    if (deathCertificateFiles) attachments.push(...deathCertificateFiles)
+    if (injuryCertificateFiles) attachments.push(...injuryCertificateFiles)
```
eirikurn, line 60: Can simplify if all variables are (maybe empty) arrays.

libs/application/template-api-modules/src/lib/modules/templates/accident-notification/accident-notification.service.ts
```diff
@@ -0,0 +1,84 @@
...
+
+    const hasAttachments = attachments && attachments?.length > 0
+
+    if (!hasAttachments) {
+      return []
+    }
```
eirikurn, line 66: No need to do this. `Promise.all` supports empty arrays.

libs/application/template-api-modules/src/lib/modules/templates/accident-notification/types/index.ts
```diff
@@ -0,0 +1,9 @@
+export interface FileAttachment {
+  name: string
+  url: string
+}
+
+export interface NodemailAttachment {
+  filename: string
+  href: string
+}
```
eirikurn, line 9: Does nodemailer include a type definition for attachments that we can use?

Would be best to skip these interfaces and prepare the attachments once using the `nodemailer` attachment type instead of mapping the attachments twice with different types used.

libs/application/template-api-modules/src/lib/modules/templates/accident-notification/accident-notification.module.ts
```diff
@@ -0,0 +1,42 @@
...
+
+const applicationRecipientEmail =
+  process.env.ACCIDENT_NOTIFICATION_APPLICATION_RECIPIENT_EMAIL_ADDRESS ??
+  'island@island.is'
+
+const applicationSenderName = process.env.EMAIL_FROM_NAME ?? ''
+
+const applicationSenderEmail = process.env.EMAIL_FROM ?? 'development@island.is'
+
+export class AccidentNotificationModule {
```
eirikurn, line 19: Should add this template api module to codeowners so we don't need to review every change.

eirikurn (changes requested):
I know you probably based this code on another api module, but I'd like to suggest these minor code improvements before it is copied even more.

But I look forward to a bigger refactor of these api modules to reduce the boilerplate. 

Looks good otherwise 👍

sigruntg (approved):
Looks good

eirikurn (approved):
Perfect! Thanks! 👏🙏🎉

PR #4619, opened 2021-08-09
chore: Allow all white space characters in strings and comments

maranomynet:
## What

Expand ESLint recommended config's default white space rules to not error on fancy spaces inside template strings and comments.

## Why

Custom white spaces have valid use-cases and are perfectly safe to display as part of the on-screen UI

The default rule already allows custom white space characters inside old-school strings, but as soon as we either turn those strings into template literals or move them into a code-comment, ESLint throws a  fit.

...which is surprising, and annoying.

## Checklist:

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -31,6 +31,14 @@
             ]
           }
         ],
+        "no-irregular-whitespace": [
+          "error",
+          {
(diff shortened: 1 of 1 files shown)


sindrig (approved):
LGTM

PR #4621, opened 2021-08-09
style: added run-psql.sh script for SQL queries

busla:
# ...

## What

## Why

## Screenshots / Gifs

## Checklist:


arnorhs (commented):
My bash is rusty, but this seems all pretty reasonable.

Did you consider skip having the SQL query part of this script, and maybe have the script only open the tunnel? Perhaps the reason is that you don't want to clog up long running db connections too long?

busla:
the script runs db-proxy.sh in the background to create the tunnel, which can be used for tunnel-only purposes.

The query is optional. The allowed options (mutually exclusive)  are `-q [string]` and `-p` (psql console).


Did I understand your question correctly 😄

pshomov (approved):
LGTM

gudjong (approved):
Not sure why Kolibri Justice League needs to review this, but it looks fine.

PR #4622, opened 2021-08-09
fix(web)/functional download csv on open data sub page

alexandradilja:
## What

Download CSV

## Why

Users can now download csv data for graphs displayed on open data subpages

## Screenshots / Gifs

Screenshot 2021-08-09 at 12 05 15

## Checklist:

### DIFF
--- a/apps/web/components/Charts/ChartsCard/ChartsCard.treat.ts
+++ b/apps/web/components/Charts/ChartsCard/ChartsCard.treat.ts
@@ -10,12 +10,9 @@ export const frameWrapper = style({
   minHeight: 124,
   textDecoration: 'none',
   position: 'relative',
-  borderRadius: 'large',
   overflow: 'visible',
   background: 'transparent',
   outline: 'none',
-  borderColor: 'purple100',
-  borderWidth: 'standard',
   ':hover': {
     textDecoration: 'none',
   },
--- a/apps/web/components/Charts/ChartsCard/ChartsCard.treat.ts
+++ b/apps/web/components/Charts/ChartsCard/ChartsCard.treat.ts
@@ -1,7 +1,7 @@
 import { style, styleMap } from 'treat'
 import { theme } from '@island.is/island-ui/theme'
 
-export const card = style({
+export const frameWrapper = style({
   display: 'flex',
   height: '646px',
   width: '100%',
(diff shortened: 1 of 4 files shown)


apps/web/components/Charts/ChartsCard/ChartsCard.tsx
```diff
@@ -56,34 +56,15 @@ export const ChartsCard: React.FC<ChartsCardsProps> = ({ data, blue }) => {
...
-      display="flex"
-      style={{ flexDirection: 'column' }}
-      flexGrow={1}
-      flexDirection={shouldStack ? 'columnReverse' : 'row'}
-      alignItems="stretch"
-      justifyContent="flexStart"
-    >
+    <Box ref={ref} className={cn(styles.card)}>
       <Box
+        className={cn(styles.outerWrapper)}
```
barabrian, line 61:
```suggestion
        className={styles.outerWrapper}
```

apps/web/components/Charts/ChartsCard/ChartsCard.tsx
```diff
@@ -56,34 +56,15 @@ export const ChartsCard: React.FC<ChartsCardsProps> = ({ data, blue }) => {
...
+        background={subPage ? 'blue100' : 'purple100'}
         style={{
-          width: '100%',
-          minHeight: '156px',
-          borderTopLeftRadius: '8px',
-          borderTopRightRadius: '8px',
+          width: graph.type === 'Pie' ? '100%' : '889px',
```
barabrian, line 64: we should try to avoid inline styles, in this case it looks like outerWrapper style has width 100% that would always be overwritten.
Probably better to define className for graph variants if they differ

apps/web/components/Charts/ChartsCard/ChartsCard.treat.ts
```diff
@@ -1,16 +1,61 @@
...
+  // overflow: 'visible',
+  background: 'transparent',
+  outline: 'none',
   ':hover': {
     textDecoration: 'none',
   },
 })
+
+export const card = style({
+  display: 'flex',
```
barabrian, line 22: if we are using the Box component we should try to use the internal styles from the Box component for most of our styles

ex.
```jsx
<Box
  display="flex"
  flexDirection="column"
  flexGrow={1}
  alignItems="stretch"
  justifyContent="flexStart"
></Box>
```
this minimizes generated css as it will reuse the Box css

barabrian: this applies to all styles that will be using the box component

apps/web/components/Charts/ChartsCard/ChartsCard.tsx
```diff
@@ -56,34 +56,15 @@ export const ChartsCard: React.FC<ChartsCardsProps> = ({ data, blue }) => {
...
       >
-        <Box
-          style={{
-            minHeight: '156px',
-            borderTopLeftRadius: '8px',
-            borderTopRightRadius: '8px',
-            paddingBottom: '24px',
-          }}
-        >
+        <Box className={cn(styles.innerWrapper)}>
```
barabrian, line 67:
```suggestion
        <Box className={styles.innerWrapper}>
```

apps/web/components/Charts/ChartsCard/ChartsCard.tsx
```diff
@@ -97,23 +78,24 @@ export const ChartsCard: React.FC<ChartsCardsProps> = ({ data, blue }) => {
...
-        )}
       </Box>
-
       <Box
         display="flex"
         justifyContent="center"
         alignItems="center"
-        style={{ width: '100%', height: '100%' }}
+        style={{
+          width: graph.type === 'Pie' || !shouldStack ? '100%' : '889px',
```
barabrian, line 94: we should try to avoid inline styles

apps/web/components/Charts/ChartsCard/ChartsCard.tsx
```diff
@@ -97,23 +78,24 @@ export const ChartsCard: React.FC<ChartsCardsProps> = ({ data, blue }) => {
...
+          height: '100%',
+        }}
       >
-        <Box justifyContent="center" style={{ width: '80%', height: '408px' }}>
-          {children}
-        </Box>
+        <Box className={cn(styles.graphParent)}>{children}</Box>
```
barabrian, line 98: only use classnames library when working with multiple classNames
```suggestion
        <Box className={styles.graphParent}>{children}</Box>
```

barabrian: this applies to all instances

apps/web/components/DataLinkSection/DataLinkSection.tsx
```diff
@@ -66,6 +66,7 @@ export const DataLinkSection = ({
               </GridColumn>
             )
           }
+          return
```
barabrian, line 69: it might be better to filter cards before mapping rather then returning undefined

barabrian (commented):
Looks good just some minor comments

PR #3621, opened 2021-04-26
feat: Add ToggleSwitch components

maranomynet:
Implementation of the [ToggleSwitch](https://www.figma.com/file/pDczqgdlWxgn3YugWZfe1v/UI-Library-%E2%80%93-%F0%9F%96%A5%EF%B8%8F-Desktop?node-id=2682%3A0) UI element

Although the Figma designs don't show any evidence of labels being considered, these components display a visible label by default, to ensure accessibility, but offer an optional `hideLabel` prop.

(Note: The storybook stories are a bit of a shot in the dark, as we don't really know how to run the storybook server locally – yet.)

## What

This implementation provides support for three different markup patterns, applicable to different use-cases: Checkbox. Button, Link

## Why

Instead of providing a single ToggleSwitch component with a super-complex props type signature, we provide three simple components each focusing on a single markup-/usage-pattern.

* `ToggleSwitchLink` for cases where a route change is neccessary part of the state change involved.
* `ToggleSwitchButton` for cases where the toggler should influence other areas/elements on screen (via `aria-controls=""`, etc.)
* `ToggleSwitchCheckbox` for all other cases where the toggling is really just a fancy-looking checkbox (includes support for old-school `name` and `value` attributes.

## Screenshot

image

## Checklist:

- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] Formatting passes locally with my changes
- [x] I have rebased against main before asking for a review

PR #4625, opened 2021-08-10
feat(an): Illustrated radio button

Thorrik58:
# Adding prop and styling to radio button to add an illustration according to design.

## What

Adding prop and styling to radio button to add an illustration according to design.
Also fix to error message for cases when tooltip and/or illustration is present on radio button.

## Why

Needed for the accident notification application.

## Screenshots / Gifs

**Default state** 

### Desktop:
image

### Mobile:
image

### Half width Desktop:
image

### Half width Mobile:
image

---

**Error Message state**

**Before** changes to error message, messages will appear like so:
### Large:
image

### Small:
image

---

**After** changes to error message and illustration added will appear so:
### Large Desktop:
image

### Large Mobile:
image

### Small Desktop:
image

### Small Mobile:
image

### Half width Large Desktop:
image

### Half width Large Mobile:
image

### Half width Small Desktop:
image

### Half width Small Mobile:
image

---

This is another approach which I liked a little bit more but didnt not implement as its further away from the original, would love some input.
### Large:
image

### Small:
image

## Checklist:


andriandresson (commented):
The illustration addition to the radioButton looks good. 
I have few  thoughts about the error state, maybe it would be enough to have a single error state for the whole radioButton instead of error state on every single option in the radioButton? 
**For example:**
image

eirikurn (approved):
Approving Aranja files.

Would like to get Island UI input on this though. At least tweak the spacing in the small variant so the error message aligns with ... something :D

Thorrik58:
Good point, I fixed the spacing so it aligns on the small version.

libs/island-ui/core/src/lib/RadioButton/RadioButton.tsx
```diff
@@ -109,25 +114,28 @@ export const RadioButton = ({
...
+            <Illustration />
+          </Box>
+        )}
         {tooltip && (
           <div
             className={cn(styles.tooltipContainer, {
               [styles.tooltipLargeContainer]: large,
+              [styles.toolTipLargeContainerWithIllustration]:
+                large && Illustration,
             })}
```
barabrian, line 128: this should probably be separated so no styles are overwritten, if large is true and/or Illustration is true marginLeft will conflict between all classnames, this makes the styles more fragile

libs/application/templates/accident-notification/src/assets/WorkTypeIllustration.tsx
```diff
@@ -0,0 +1,194 @@
+import React, { FC } from 'react'
+
+export const WorkTypeIllustration: FC = () => (
```
barabrian, line 3: We should probably avoid unnecessary use of RC https://github.com/island-is/island.is/pull/3621#discussion_r620690786

barabrian (commented):
LGTM, just some minor comments

barabrian (approved):
LGTM

Toti91 (approved):
LGTM 👍

PR #4632, opened 2021-08-11
fix(core): 08-11-2021 yarn audit

jeremybarbet:


eirikurn (approved):
These both fall within valid upgrade ranges so we should remove the resolution entries.

PR #4636, opened 2021-08-11
chore(parental-leave): improve sentry logs

jeremybarbet:
## What

- Better errors loggings


apps/application-system/form/src/routes/AssignApplication.tsx
```diff
@@ -1,77 +1,98 @@
...
-    if (isMissingToken) {
-      return
-    }
+    const init = async () => {
+      if (isMissingToken) {
+        Sentry.captureException(
+          new Error(
+            `Missing token, cannot assign the application ${location.search}`,
+          ),
+        )
```
eirikurn, line 52: This can be a user error which can add to non-fixable noise in Sentry.

User errors can be logged as a warning somewhere to help with debugging. Though not sure if/how we want to log it in the browser though.

eirikurn (commented):
One comment 👍

apps/application-system/form/src/routes/AssignApplication.tsx
jeremybarbet: I replaced it with a `captureMessage` instead of an exception. It might still be noisy but would help the debugging. I don't see other alternatives at the moment than going with Sentry for logging this

eirikurn (approved):
Great!

PR #4648, opened 2021-08-12
feat(an): in review

sigruntg:
# In review state

https://app.clickup.com/t/pzcu0u

## What

This is an in review screen.
Its display depends on which state the user is in.
* NEEDS_DOCUMENT
* NEEDS_REVIEW
* NEEDS_DOCUMENT_AND_REVIEW
* IN_FINAL_REVIEW

## Why

So the user can see what needs to be done for the application to be accepted after submission.

## Screenshots / Gifs

image

## Checklist:


Toti91 (approved):
Nice

PR #4673, opened 2021-08-19
fix(parental-leave): make the expectedDateOfBirth optional

jeremybarbet:
## What

- the `expectedDateOfBirth` return by VMST can be optional like the `dateOfBirth`


baering (approved):
LGTM.

Would be great to keep the `clientConfig.yaml` in sync with the schema, but we'll have to wait for an update.

PR #4679, opened 2021-08-20
feat(service-portal): allow dynamic routes in service-portal

thordurhhh:
## What

A service is called to determine which tabs the user can see in the finance section on the service-portal.
A little bit extra extending the disabled navItems to subnavItems as well.

## Why

The service returning the response for which navigation items should be available could be a little bit slow, and if that is the case, then with the current set up the whole service portal menu will be hidden until all promises are resolved.
With this change, all the original navigation items are loaded from their modules, and then the dynamic routes will be called and added when they are ready. They will only be held up on their own.

## Screenshots / Gifs
Screenshot 2021-08-20 at 10 29 30

## Checklist:


Valur (approved):

Valur:
LGTM

PR #4696, opened 2021-08-23
fix(parental-leave): allow one month before dob to create a period

jeremybarbet:
## What

- A parental leave period can be added 1 month before the expected dob


baering (approved):
LGTM 👍  let's remember to update `errorMessages.periodsStartDateBeforeDob` to match this change.

jeremybarbet:
Yes I already updated on contentful, but didn't publish the changes yet, waiting for the release train to be done for that

PR #4704, opened 2021-08-23
feat(services-endorsements-api): Access control update

stefanhar:
# Access control update

Currently DMR can't access the data in party list, we don't want these lists open to the public so we validate the against lists of allowed national ids.
This is a temporary solution until a more permanent system for access control is in place.

## What

- Added access guard for na tional ids
- Updated submit button texts

## Why

- Allows a limited set of individuals to access endorsement lists
- Allows a more granular access control

## Checklist:


maggisk (approved):

sindrig (commented):
LGTM
"This is a temporary solution until a more permanent system for access control is in place." I feel like I've heard this before :sweat_smile:

sindrig (approved):

PR #4711, opened 2021-08-24
fix(parental-leave): remove the percentage from the slider duration

jeremybarbet:
## What

- Creates confusion, we remove the percentage from the duration slider
- Rename file from `index.tsx` to `Duration.tsx` and use named import to follow convention

Screenshot 2021-08-24 at 08 41 42


baering (approved):
LGTM 👍

PR #4741, opened 2021-08-25
fix(services-search-indexer): Reduce bulk import chunk size

stefanhar:
# Reduce bulk import chunk size

Bulk import is failing due to large payload size

## What

- Reduce chunk size to a value that is unlikely to cause further issues

## Why

- Allow search indexer to continue to work normally

## Checklist:


kksteini (approved):
LGTM

PR #4797, opened 2021-08-30
fix(driving-license-assessment-submission): fixing missing env vars and error checking

arnorhs:
- Making 4xx responses result in an actual error for the driving license service
- Adding appropriate env vars for the shared template api service
- Changing the lifecycle of the assessment applications after reviews by Syslumenn

Turns out the submission never worked in prod/dev properly after I added the driving license client the other day - wasn't discovered because the service was returning a 404 and those did not result in actual errors in dev, and since the data transparency isn't very great nobody really realised since things were ending up in the right place in local testing

Things on the dev cluster were not ending up in the right place. This should address that issue.

Weird that we need to have secret values duplicated but that's probably for a good reason.

## Checklist:


darrikonn (approved):
LGTM

libs/clients/driving-license/src/lib/drivingLicense.module.ts
```diff
@@ -21,6 +21,7 @@ export class DrivingLicenseApiModule {
               new Configuration({
                 fetchApi: createEnhancedFetch({
                   name: 'clients-driving-license',
+                  treat400ResponsesAsErrors: true,
```
eirikurn, line 24: Hm. Not sure if I understand. 400 responses are always logged and thrown as an Error that should be dealt with by the callee. The only thing this flag does is change the logLevel for 400 responses from warn to error, and cause the circuit breaker to fail for too many 400 responses. Sorry if the interface is not clear, it was pretty hard to name this :)

The reason why this is set to false by default, is that 400 responses are "request issues". With this configuration an end-user might be able to cause DOS if they can trigger 400 responses easily, eg some validation error response.

Those responses would spam the logs with "errors", raising incidents and causing noise for the ops team. They would also cause the circuit to break which causes DOS for everyone else.

Regarding the original issue, I'm curious what happened to the thrown 400 errors. We should also monitor warnings when testing new integrations.

eirikurn: So this flag should only be set to true when we're fairly confident that end-users can't intentionally cause 400 responses, and we want to trigger devops incidents for 400 responses.

pshomov (approved):
LGTM

libs/clients/driving-license/src/lib/drivingLicense.module.ts
arnorhs: hm, interesting. I might need to dig into this enhanced fetch a bit better

barabrian (approved):
LGTM

libs/clients/driving-license/src/lib/drivingLicense.module.ts
arnorhs: Really good catch... that was completely unrelated. Turns out if you send a post request to some random API URL on x-road with no configured service, even if it's complaining in the xml response that it's not a handled content type, it's still always returning a 200 response.. eg:

```
* upload completely sent off: 122 out of 122 bytes
< HTTP/1.1 200 OK
< Date: Tue, 31 Aug 2021 10:30:52 GMT
< Content-Type: text/xml;charset=utf-8
< SOAPAction: 
< Content-Length: 427
< 
* Connection #0 to host localhost left intact
<?xml version="1.0" encoding="UTF-8"?><SOAP-ENV:Envelope xmlns:SOAP-ENV="http://schemas.xmlsoap.org/soap/envelope/"><SOAP-ENV:Body><SOAP-ENV:Fault><faultcode>Client.InvalidContentType</faultcode><faultstring>Invalid content type: application/json</faultstring><faultactor></faultactor><detail><faultDetail xmlns="">xxx</faultDetail></detail></SOAP-ENV:Fault></SOAP-ENV:Body></SOAP-ENV:Envelope>
```

so I got those two mixed up.. I updated the PR, in any case the PR addresses the issue.

eirikurn (approved):

PR #4798, opened 2021-08-31
feat(delegation): better UX for expired delegations

darrikonn:
# Better UX for expired delegations

https://www.notion.so/Expired-delegations-8535b0dcc7d84c25a5e1911a849e32fe
https://www.notion.so/Functionality-before-user-has-selected-any-delegations-38d21c6cd4eb4c7b817a861859bb451d
https://www.notion.so/Lets-revisit-the-date-of-scope-on-the-A-gangst-ring-panel-1bcf64c0e5b847aa90b9378cea21bf47

## What

- Only show validTo if all the scopes validTo are mutual
- Don't allow deleting a delegation when there are no scopes behind it
- Filter out expired and unused delegations

## Screenshots / Gifs

access

## Further improvements?
- can use optimisticUpdate for better UX

## Checklist:


libs/api/domains/auth/src/lib/resolvers/delegation.resolver.ts
```diff
@@ -49,8 +49,14 @@ export class DelegationResolver {
...
 
   @Query(() => [Delegation], { name: 'authDelegations' })
-  getDelegations(@CurrentUser() user: User): Promise<DelegationDTO[]> {
-    return this.authService.getDelegations(user)
+  async getDelegations(@CurrentUser() user: User): Promise<DelegationDTO[]> {
+    const delegations = await this.authService.getDelegations(user)
+    return delegations.filter(
+      (delegation) =>
+        delegation.scopes &&
+        delegation.scopes.length > 0 &&
```
eirikurn, line 57: We also need to somewhere filter out individual scopes that are not valid anymore. Do you plan to fix that in another PR?

libs/api/domains/auth/src/lib/resolvers/delegation.resolver.ts
```diff
@@ -49,8 +49,14 @@ export class DelegationResolver {
...
+        (delegation.validTo === undefined || delegation.validTo >= new Date()),
```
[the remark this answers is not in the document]
darrikonn: You're right, at least auth-public-api filters like this:
```typescript
{ validTo: { [Op.or]: [{ [Op.eq]: null }, { [Op.gt]: now }] } }
```

so today is exclusive.

libs/api/domains/auth/src/lib/resolvers/delegation.resolver.ts
darrikonn: I really feel like this shouldn't be handled by the resolver. To me, the resolver is just a consumer of the business logic that is implemented in `auth-(public-?)api`.

Same applies to the filtering that I did.

So IMO, this should be handled in auth-api.

eirikurn: You're right. Can we move this there? :)

libs/auth-api-lib/src/lib/services/delegations.service.ts
```diff
@@ -156,8 +158,10 @@ export class DelegationsService {
           model: DelegationScope,
           where: {
             [Op.and]: [
-              { validFrom: { [Op.lt]: now } },
-              { validTo: { [Op.or]: [{ [Op.eq]: null }, { [Op.gt]: now }] } },
+              { validFrom: { [Op.lte]: today } },
+              {
+                validTo: { [Op.or]: [{ [Op.eq]: null }, { [Op.gte]: today }] },
```
darrikonn, line 163: Note that today is **inclusive**.
Would be good to verify other places that this should indeed be inclusive

eirikurn (approved):
Great!

valurefugl (approved):

PR #4817, opened 2021-09-01
fix(fa): button import

Rikkiabb:
# ...
https://app.asana.com/0/1200826552987384/1200590454737643

## What

Uses our button for import 

## Why

So it won't break on prod


MargretFinnboga (commented):
🐘 a

linavidars (approved):

PR #4818, opened 2021-09-01
fix(financial-aid): Redirect to status page on web-osk upgraded

MargretFinnboga:
# Redirect to status page on web-osk upgraded

https://app.asana.com/0/1200826552987384/1200368932625373

## What

Added return url enum to user 
Get application added to status page

## Why

Better login flow for web-osk

## Checklist:


libs/financial-aid/shared/src/lib/types.ts
```diff
@@ -27,6 +27,12 @@ export enum RolesRule {
   VEITA = 'veita',
 }
 
+export enum ReturnUrl {
+  UMSOKN = '/umsokn',
+  STADA = '/stada',
+  VEITA = '/nymal',
```
Rikkiabb, line 33: Should we change to something like:
```
 APPLICATION = '/umsokn',
 MYPAGE = '/stada',
 ADMIN = '/nymal',
```
?

apps/financial-aid/api/src/app/modules/auth/auth.controller.ts
```diff
@@ -172,7 +174,17 @@ export class AuthController {
...
+      user.returnUrl = ReturnUrl.VEITA
+    }
+    if (res.req?.query.applicationId) {
+      user.returnUrl = ReturnUrl.STADA
+    }
+
+    const returnURl =
+      user.returnUrl === ReturnUrl.STADA
+        ? ReturnUrl.STADA + '/' + res.req?.query.applicationId
+        : user.returnUrl
```
Rikkiabb, line 187: You could add the applicationId param on line 181 and then you don't have to have this check here.

apps/financial-aid/web-osk/graphql/sharedGql.ts
```diff
@@ -62,3 +62,16 @@ export const CurrentUserQuery = gql`
     }
   }
 `
+
+export const GetApplicationQuery = gql`
+  query GetApplicantyQuery($input: ApplicationInput!) {
```
Rikkiabb, line 67: `GetApplicantyQuery` 🤔 should this be `GetApplicationQuery`

apps/financial-aid/web-osk/src/components/Login/login.tsx
```diff
@@ -11,53 +12,59 @@ interface Props {
 const Login = ({ headline, about }: Props) => {
   const router = useRouter()
 
-  const apiLoginRoute = '/api/auth/login'
+  const apiLoginRoute = router.query.id
+    ? `/api/auth/login?applicationId=${router.query.id}&nationalId=`
+    : '/api/auth/login?nationalId='
```
Rikkiabb, line 17: Should we concat `nationald=` where use them below?

And should we have these routes as const in this file?

apps/financial-aid/web-osk/src/components/Status/InProgress.tsx
```diff
@@ -41,7 +41,7 @@ const InProgress = ({ currentApplication }: Props) => {
             cta={{
               label: 'Hlaða upp gögnum',
               onClick: () => {
-                router.push(`${router.query.id}/gogn`)
+                router.push(`stada/${router.query.id}/gogn`)
```
Rikkiabb, line 44: Should we maybe have all routes in a const file? Then in the future if we add something like `stada` then we only change that file.

apps/financial-aid/web-osk/src/routes/status/MainPage/mainPage.tsx
```diff
@@ -15,22 +16,40 @@ import {
...
+      variables: { input: { id: router.query.id } },
+      fetchPolicy: 'no-cache',
+      errorPolicy: 'all',
+    },
+  )
   const currentApplication = useMemo(() => {
-    if (user?.currentApplication) {
-      return user.currentApplication
+    if (data?.application) {
+      return data?.application
```
Rikkiabb, line 50: is the question mark needed on line 50?

Rikkiabb (changes requested):
🔴

apps/financial-aid/web-osk/src/components/Login/login.tsx
MargretFinnboga: Changed to two different ones for the time being, while we are still using fake users. What about that?

apps/financial-aid/api/src/app/modules/auth/auth.controller.ts
MargretFinnboga: true, but I cant set the user.returnurl with application id it only accepts enum, so I will always have to add another variable anyways?

Rikkiabb: So what I am thinking is basically you could drop lines 177-187 and instead do something like: 
```
        const returnUrl = `${user.returnUrl}${
          res.req?.query.applicationId ? `/${res.req.query.applicationId}` : ''
        }`
```
Or you could even do it in the redirect on line 198 or am I misunderstanding something? 👼

apps/financial-aid/web-osk/src/components/Login/login.tsx
```diff
@@ -11,53 +12,63 @@ interface Props {
 const Login = ({ headline, about }: Props) => {
   const router = useRouter()
 
-  const apiLoginRoute = '/api/auth/login'
+  const apiLoginRouteForFake = router.query.id
+    ? `/api/auth/login?applicationId=${router.query.id}&nationalId=`
+    : '/api/auth/login?'
```
Rikkiabb, line 17: Shouldn't line 17 also include `nationalId=`?

apps/financial-aid/web-osk/src/routes/status/MainPage/mainPage.tsx
```diff
@@ -16,22 +17,40 @@ import {
...
+import {
+  Application,
+  CurrentApplication,
+  getActiveTypeForStatus,
+} from '@island.is/financial-aid/shared'
+
 import { useLogOut } from '@island.is/financial-aid-web/osk/src/utils/useLogOut'
+import { useQuery } from '@apollo/client'
+import { useRouter } from 'next/router'
+import { GetApplicationQuery } from '@island.is/financial-aid-web/oskgraphql'
```
Rikkiabb, line 31: Some imports here not used

libs/financial-aid/shared/src/lib/const.ts
```diff
@@ -11,3 +11,5 @@ export const AllowedFakeUsers = [
   '0000000002', // Veita user
   '0000000003', // User with an application which needs data
 ]
+
+export const LoginStatusPage = '/stada'
```
Rikkiabb, line 15: So what I was thinking here is something like this: 

```
export const Routes = {
  status: '/stada',
  statusFileUpload: (id: string) => `/stada/${id}/gogn`,
}
```

Etc....

Then we would never have hardcoded values in our components but would only refer to the `Routes` obj. Does this make sense? It could also be a separate PR.

Rikkiabb (commented):
Awesome, couple of comments + what we talked about IRL

apps/financial-aid/api/src/app/modules/auth/auth.controller.ts
```diff
@@ -172,12 +174,26 @@ export class AuthController {
...
+  getReturnUrl = (service: string, applicationId: string) => {
+    switch (true) {
+      case service === 'veita':
+        return ReturnUrl.ADMIN
+      case applicationId != undefined:
+        return `${ReturnUrl.MYPAGE}/${applicationId}`
+      default:
+        return ReturnUrl.APPLICATION
+    }
+  }
```
Rikkiabb, line 186: How do you feel about?
```
  getReturnUrl = (service: RolesRule, applicationId: string | undefined) => {
    switch (true) {
      case service === RolesRule.OSK && applicationId !== undefined:
        return `${ReturnUrl.MYPAGE}/${applicationId}`
      case service === RolesRule.VEITA:
        return ReturnUrl.ADMIN
      default:
        return ReturnUrl.APPLICATION
    }
  }
```

apps/financial-aid/web-osk/src/components/HasApplied/hasApplied.tsx
```diff
@@ -18,7 +19,7 @@ const HasApplied = () => {
 
       <Button
         onClick={() => {
-          router.push(`/${user?.currentApplication?.id}`)
+          router.push(`${Routes.status}/${user?.currentApplication?.id}`)
```
Rikkiabb, line 22: This could be in the Routes object instead of interpolating

Rikkiabb (approved):
Some comments, you can look at but feel free to merge

PR #4819, opened 2021-09-01
feat(delegation): ux fix for nationalId input in grant access

darrikonn:
# UX fix for nationalId input in grant access 

https://www.notion.so/No-action-when-kennitala-is-wrong-5a15ea7d34924bb6aa48fb8a5a7762ee
https://www.notion.so/A-gangshafi-field-seems-to-be-editable-0e7780b0bbe34fc5949674ab449c10f5

## What

* Better error handling for incorrect nationalId
* Fix UX when fetching and displaying user's name based on nationalId

### v1
natid

### v2
v7

## Checklist:

### DIFF
--- a/libs/service-portal/settings/access-control/src/screens/GrantAccess/GrantAccess.tsx
+++ b/libs/service-portal/settings/access-control/src/screens/GrantAccess/GrantAccess.tsx
@@ -213,7 +213,6 @@ function GrantAccess() {
               {queryLoading ? (
                 <span
                   className={cn(styles.icon, styles.loadingIcon)}
-                  aria-hidden="false"
                   aria-label="Loading"
                 >
                   <Icon icon="reload" size="large" color="blue400" />
--- a/libs/service-portal/settings/access-control/src/screens/GrantAccess/GrantAccess.tsx
+++ b/libs/service-portal/settings/access-control/src/screens/GrantAccess/GrantAccess.tsx
@@ -199,15 +199,14 @@ function GrantAccess() {
                 loading={loading}
               >
                 {name
-                  ? `${formatMessage({
-                      id:
-                        'service.portal.settings.accessControl:grant-form-pre-submit-active',
-                      defaultMessage: 'Veita',
-                    })} ${name} ${formatMessage({
-                      id:
-                        'service.portal.settings.accessControl:grant-form-post-submit-active',
-                      defaultMessage: 'aðgang',
-                    })}`
+                  ? formatMessage(
+                      {
+                        id:
+                          'service.portal.settings.accessControl:grant-form-submit-active',
+                        defaultMessage: 'Veita {name} aðgang',
+                      },
+                      { name },
+                    )
                   : formatMessage({
                       id:
                         'service.portal.settings.accessControl:grant-form-submit-disabled',
--- a/libs/service-portal/settings/access-control/src/screens/GrantAccess/GrantAccess.tsx
+++ b/libs/service-portal/settings/access-control/src/screens/GrantAccess/GrantAccess.tsx
@@ -47,14 +47,26 @@ const IdentityQuery = gql`
 `
 
 function GrantAccess() {
-  const { handleSubmit, control, errors, setValue, watch } = useForm()
-  const [createAuthDelegation, { loading }] = useMutation<Mutation>(
-    CreateAuthDelegationMutation,
+  const { handleSubmit, control, errors, watch } = useForm({ mode: 'onChange' })
+  const [
+    createAuthDelegation,
+    { loading: mutationLoading },
+  ] = useMutation<Mutation>(CreateAuthDelegationMutation, {
+    refetchQueries: [{ query: AuthDelegationsQuery }],
+  })
+  const [getIdentity, { data, loading: queryLoading }] = useLazyQuery<Query>(
+    IdentityQuery,
     {
-      refetchQueries: [{ query: AuthDelegationsQuery }],
+      onError: (error) => {
+        toast.error(
+          formatMessage({
+            id: 'service.portal.settings.accessControl:grant-identity-error',
+            defaultMessage: 'Enginn notandi fannst með þessa kennitölu.',
+          }),
+        )
+      },
     },
   )
-  const [getIdentity, { data }] = useLazyQuery<Query>(IdentityQuery)
   const { identity } = data || {}
   const { formatMessage } = useLocale()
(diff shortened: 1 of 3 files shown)


libs/service-portal/settings/access-control/src/screens/GrantAccess/GrantAccess.tsx
```diff
@@ -160,52 +181,41 @@ function GrantAccess() {
...
+                    })} ${name} ${formatMessage({
+                      id:
+                        'service.portal.settings.accessControl:grant-form-post-submit-active',
+                      defaultMessage: 'aðgang',
+                    })}`
+                  : formatMessage({
+                      id:
+                        'service.portal.settings.accessControl:grant-form-submit-disabled',
+                      defaultMessage: 'Sláðu inn kennitölu aðgangshafa',
+                    })}
```
eirikurn, line 214: Should use interpolated translations to help with i18n, but these labels aren't great (long, name not correct "fall") and think we should keep "Áfram".

We should either keep the disabled input (and wait for disabled design to be improved), or just render the name standalone below or to right of the ssn input.

libs/service-portal/settings/access-control/src/screens/GrantAccess/GrantAccess.tsx
```diff
@@ -43,41 +43,56 @@ const IdentityQuery = gql`
...
+  })
+  const [getIdentity, { data, loading: queryLoading }] = useLazyQuery<Query>(
+    IdentityQuery,
     {
-      refetchQueries: [{ query: AuthDelegationsQuery }],
+      onError: (error) => {
+        toast.error(
+          formatMessage({
+            id: 'service.portal.settings.accessControl:grant-identity-error',
+            defaultMessage: 'Enginn notandi fannst með þessa kennitölu.',
```
eirikurn, line 61: "Fann ekki einstakling með þessa kennitölu."?

Meikar sense að þetta sé onError? Finnst eins og að queryið ætti að skila null ef kennitalan er ekki skráð.

Getum við hent inn graphql query til að filter-a einstaklinga. Svo það gleymist ekki þegar við tengjum fyrirtækjaskrá hinum megin.

eirikurn (changes requested):

libs/service-portal/settings/access-control/src/screens/GrantAccess/GrantAccess.tsx
eirikurn: *graphql input argument

darrikonn: You're correct, this should be `null`.
But it isn't handled by the national registry xroad service; so fixing that would've needed refactoring in bunch of places. I'll create a ticket for it and we should tackle that as a separate PR.

darrikonn: https://github.com/orgs/island-is/projects/9#card-67950820

libs/service-portal/settings/access-control/src/screens/GrantAccess/GrantAccess.tsx
darrikonn: Ahh nice, didn't know interpolated translations were supported.

Yeah, rendering something based on user interaction is always difficult in regards to UX.
At first I just rendered the name as a standalone, both below and to the right of the input; and IMO it was a bad UX and just didn't look good. The focus of the user would either be on the button or the input, so I went with the button. I'm open for other ideas, but as I said, the plain text (or key/value text) was not a good option.
Agree that "fall" breaks that, not sure if rephrasing the text makes it better:

libs/service-portal/settings/access-control/src/screens/GrantAccess/GrantAccess.tsx
eirikurn: Sounds good.

libs/service-portal/settings/access-control/src/screens/GrantAccess/GrantAccess.tsx
eirikurn: Yeah, sorry, I just don't think this change is much better than how it is now. Having duplicate instructions in the inactive button, and an Icelandic issue in the active button.

I think we should split these changes up and get more eyes on this or a new design proposal.

darrikonn: Yeah I agree

libs/service-portal/settings/access-control/src/screens/GrantAccess/GrantAccess.tsx
```diff
@@ -116,97 +140,111 @@ function GrantAccess() {
...
+                    requestDelegation(value)
+                  }}
                 />
-              )}
-            />
+              </Box>
+              {queryLoading ? (
+                <span
+                  className={cn(styles.icon, styles.loadingIcon)}
+                  aria-hidden="false"
```
eirikurn, line 216: Remove?

libs/service-portal/settings/access-control/src/screens/GrantAccess/GrantAccess.tsx
```diff
@@ -116,97 +140,111 @@ function GrantAccess() {
...
                     id:
-                      'service.portal.settings.accessControl:grant-label-user',
-                    defaultMessage: 'Aðfangshafi',
+                      'service.portal.settings.accessControl:grant-form-access-holder',
+                    defaultMessage: 'Aðgangshafi',
+                  })}
+                  disabled
+                />
+              )}
+              <Box display={name ? 'none' : 'block'}>
```
eirikurn, line 164: Ah. I guess it's pretty tricky to implement this behavior with good a11y. I wonder if this implementation loses the focus position. Should create a follow up task for island-ui.

eirikurn (approved):
Love the UX but question the a11y. Let's get this in and follow up later.

libs/service-portal/settings/access-control/src/screens/GrantAccess/GrantAccess.tsx
eirikurn: It should also announce the change somehow, eg with [live regions](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions).

PR #4822, opened 2021-09-01
feat(api): Syslumenn API Auctions

siggifv:
## What

Adding a method to Syslumenn API client to fetch auctions.

## Why

This new method will be utilised in the Syslumenn Auctions view in the Web to fetch Auctions directly from Syslumenn instead of from Contentful, coming up in a separate PR for the Web project.

The term `getAuctions` is already used by the Contentful Auctions service; therefor I've named this new method `getSyslumennAuctions`, to avoid conflicts in the GraphQL schema. 

## Checklist:


libs/api/domains/syslumenn/src/lib/syslumenn.resolver.ts
```diff
@@ -16,4 +17,11 @@ export class SyslumennResolver {
   getHomestays(@Args('input') input: GetHomestaysInput): Promise<Homestay[]> {
     return this.syslumennService.getHomestays(input.year)
   }
+
+  // Note: We need to be careful when caching the Auction data, as it's prone to changes only minutes before the auction takes place.
+  // TODO: Check if the CACHE_TIME in production. If it's only a few seconds, we should consider caching the data.
```
siggifv, line 22: Raising this TODO in a thread.

As stated in the code comment/todo, we need to be careful when caching the Auction data, as it's prone to changes only minutes before the auction takes place.

We need to find out what value `CACHE_TIME` is set to in production and after that assess how we'll handle caching of Auction data.

siggifv: We've decided not to add cache to this method, at least for now. cc. @Ingimarsson

Ingimarsson (approved):
LGTM

PR #4829, opened 2021-09-02
fix(island-ui): fix random issues with style order

eirikurn:
## What

Tweak ordering of island-ui imports to reduce chances of style order issues.

## Why

The financial-aid team reported issues where button styles were being overridden by (box) reset styles. It turns out that mini-css-extract-plugin does not guarantee order of css, and there is some strange re-ordering behaviour with next, nx and our styles. 

Long story short, it helped in this case to order the imports of island-ui correctly. Hopefully this won't be a recurring issue.

## Screenshots / Gifs

Before:

image

After:

image

## How I got there

* Reproduced issue on my machine.
* Found the mis-ordered reset style in reset.treat.ts, used by Box. The issue was not in globalReset as thought.
* Try logging out order of button.treat.ts and reset.treat.ts. Reset treat files are correctly compiled before Button, but still show up after Button styles in the CSS file.
* Try to reproduce in Osk. Osk works, but stops working if I only use Button. Some imported component seems to “fix” the button.
* Narrow down all imports until I arrive at ActionCard which “fixes” Button.
* ActionCard also imports Box but ActionCard is imported before Button in island-ui index.ts.
* Moving Box import above Button import fixed the issue.
* Thesis: Treat files are correctly compiled depth first (in dependency order, Box before Button), but appended to the CSS incorrectly breadth-first (Button before Box).
* Dug into the treat-webpack-plugin but did not find the issue there.
* Dug into mini-css-extract-plugin and see that it sorts css based on webpack module index, where reset.css has a higher index than button.css.
* Try to reproduce in a clean next project, but the issue does not occur there. Module indexes are "correct".
* Try the working setup in island.is monorepo, figure out that having the exact same files inside the Next folder and in the Island-UI folder changes (breaks) the module indices.
* Was also able to reproduce this with normal CSS modules, so Treat is not a suspect anymore.
* End thesis: Something something NextJS something monorepo, something webpack something.
* Current workaround: Make sure to import Box before Button in island-ui.

## Checklist:


oddsson (approved):

darrikonn (approved):
Nice debugging!

darrikonn:
Btw. shouldn't reset.treat only be globalStyles and thus wouldn't affect class stylings due to its precedence?

eirikurn:
Yeah, that's a good point. Reset could be globalStyles with lower precedence.

Other useBoxStyles may still conflict in rarer cases, but then that's a misuse of the Box component. I.e. you should configure layout styles with Box props only, but it's possible to add other styles on top.

PR #4841, opened 2021-09-02
fix(financial-aid):  Fixing application filters count

MargretFinnboga:
# Application filters count

https://app.asana.com/0/1200826552987384/1200590454737658

## What

Fixed when reducing  the arrays and adding the numbers together check
Moved the fadeout line, so if the history doesn't appear the line doesn't stand alone
Moved Timeline on status page, to main page so it always is shown and made less margin between sections

## Why

Want to have right numbers and design choices added

## Checklist:


apps/financial-aid/web-veita/src/components/History/History.tsx
```diff
@@ -40,12 +40,12 @@ const History = ({ className }: Props) => {
...
-        Saga umsóknar
-      </Text>
-
-      {data?.applicationEvents && (
+      {data?.applicationEvents && data?.applicationEvents.length > 0 && (
         <>
+          <Text as="h2" variant="h3" color="dark300" marginBottom={3}>
+            Saga umsóknar
+          </Text>
+
```
Rikkiabb, line 48: You can skip the null check on line 49

Rikkiabb (approved):
🗄️

PR #4856, opened 2021-09-06
fix(application-template-driving-license): updating language string

arnorhs:
making it so the driving license messages does not contain any mention of 'skilríki'

## Checklist:


libs/application/templates/driving-license/src/lib/drivingLicenseTemplate.ts
```diff
@@ -56,15 +56,15 @@ const template: ApplicationTemplate<
...
+  name: m.applicationForDrivingLicense,
   dataSchema,
   readyForProduction: true,
   stateMachineConfig: {
     initial: States.DRAFT,
     states: {
       [States.DRAFT]: {
         meta: {
-          name: 'Umsókn um ökuskilríki',
+          name: 'Umsókn um ökuskírteini',
```
stefanhar, line 67: Why not move all strings to messages.ts file?

arnorhs: this is actually a string type, but I could refer to the title in the messages, will do,

kksteini (approved):
LGTM

PR #4865, opened 2021-09-06
feat(fa): Flow after file upload

Rikkiabb:
# ...
https://app.asana.com/0/1200826552987384/1200908596880396

## What
Makes mocking users better, handles more cases
Handles success and error flow from file upload on status page
Adds shared for web-clients, up for debate if we want to go that way. Waited with creating alias for it until we decide, will do in a later PR if we go that route

For core: Adds shared web component library for components which are shared between financial aid clients but are not in UI Core. 

## Why
Flow from figma

## Screenshots / Gifs
Screenshot 2021-09-05 at 22 01 00
Screenshot 2021-09-05 at 22 08 30

## Checklist:


apps/financial-aid/web-osk/src/routes/status/FileUpload/fileUpload.treat.ts
```diff
@@ -0,0 +1,11 @@
+import { theme } from '@island.is/island-ui/theme'
+import { style } from 'treat'
+
+export const widthAlmostFull = style({
```
Rikkiabb, line 4: This is declared in a couple of places, not sure if we should re-use it or not, for now it is c/p

apps/financial-aid/web-shared/tsconfig.app.json
```diff
@@ -0,0 +1,9 @@
+{
```
Rikkiabb, line 1: Not sure if we want to got this router, but added it as a suggestion, if we want to go this route then I'll add a alias for the import in another PR.

apps/financial-aid/web-osk/src/routes/status/FileUpload/fileUpload.tsx
```diff
@@ -18,45 +17,42 @@ import { useMutation } from '@apollo/client'
...
-      setNextButtonText('Engar skrár til staðar')
+    if (form?.otherFiles.length <= 0 || router.query.id === undefined) {
       return
     }
 
+    setIsLoading(true)
+
     try {
-      await uploadFiles(currentApplication.id, FileType.OTHER).then(
+      await uploadFiles(router.query.id as string, FileType.OTHER).then(
```
MargretFinnboga, line 37: as string check is this necessary? 🤓

apps/financial-aid/web-osk/src/components/FileUploadResult/FileUploadResult.tsx
```diff
@@ -0,0 +1,56 @@
...
+  Footer,
+} from '@island.is/financial-aid-web/osk/src/components'
+import { FileList } from '../../../../web-shared/src/components'
+import { FormContext } from '@island.is/financial-aid-web/osksrc/components/FormProvider/FormProvider'
+import { Box, Text } from '@island.is/island-ui/core'
+import * as styles from './fileUpload.treat'
+import { Colors } from '@island.is/island-ui/theme'
+
+interface Props {
+  subtitle: string
```
MargretFinnboga, line 14: Maybe camelcasing here subTitle?

Rikkiabb: Subtitle is a single 🙈  see: https://dictionary.cambridge.org/dictionary/english/subtitle

apps/financial-aid/web-osk/src/routes/status/FileUpload/fileUpload.tsx
Rikkiabb: Yes because the type is `string | string[]` 😢

dabbeg (approved):
Core files LGTM

eirikurn:
Why are we creating a new file upload design when there is one already in the design system and island-ui? I think the current one might be a bit coupled to the application system, but we should probably refactor and share it.

SUMMARY.md
```diff
@@ -85,6 +85,7 @@
 - [API](apps/api/README.md)
 - [Application System](apps/application-system/README.md)
 - [Financial Aid System for Samband Sveitarfélaga](apps/financial-aid/README.md)
+  - [Financial Aid Web Shared](apps/financial-aid/web-shared/README.md)
```
eirikurn, line 88: For consistency, this should be a library under libs. App projects should not import stuff from other app projects. 

If this contains stuff that should only be shared between financial-aid (web) projects you can move it to for example `libs/financial-aid/components`.

eirikurn (changes requested):
Nice refactor. Should be a library and wonder about new file upload designs.

MargretFinnboga (approved):
👑

SUMMARY.md
Rikkiabb: Very true, fixed, will was at it then I also changed the alias for other files in the financial-aid system.

Rikkiabb:
> Why are we creating a new file upload design when there is one already in the design system and island-ui? I think the current one might be a bit coupled to the application system, but we should probably refactor and share it.

We are using the UI-core file upload component. 
Here we have a page where the user can add files after submitting the application if the employee who is working the application asks for extra data. 
Then the user uploads the x files, presses submit (which triggers a call to link those files to the application) and depending on that result the user goes to a success/error screen which you see implemented in the screenshots above. 
This is a fair point, I'll talk with our designer if the file list component should be in core or not.

apps/financial-aid/web-osk/src/components/Files/Files.tsx
```diff
@@ -7,7 +7,7 @@ import {
 } from '@island.is/financial-aid-web/osk/src/components'
 
 import { FormContext } from '@island.is/financial-aid-web/osk/src/components/FormProvider/FormProvider'
-import { useFileUpload } from '@island.is/financial-aid-web/osksrc/utils/useFileUpload'
+import { useFileUpload } from '@island.is/financial-aid-web/osk/src/utils/useFileUpload'
```
MargretFinnboga, line 10: 😻

apps/financial-aid/web-veita/src/components/FilesListWithHeader/FilesListWithHeader.tsx
```diff
@@ -1,8 +1,8 @@
 import React from 'react'
 import { Text, Box } from '@island.is/island-ui/core'
 import cn from 'classnames'
-import { ApplicationFile } from '@island.is/financial-aid/shared'
-import { FileList } from '@island.is/financial-aid-web/shared/src/components'
+import { ApplicationFile } from '@island.is/financial-aid/shared/index'
+import FileList from '@island.is/financial-aid/shared/components/FileList/FileList'
```
eirikurn, line 5: Strange to get some stuff from `shared/index` and some stuff from `shared/deep/path`.

We should generally encapsulate libs and avoid deep importing from them. Better to create more libs (eg `@island.is/financial-aid/components`) and import stuff from each lib's root index.ts module.

Then we also don't need this `/index` part.

Fyi, the next version of NX adds a lint error when using absolute imports instead of relative imports for modules in the same workspace. I've also been suggesting for the `web` project to move away from absolute deep imports from `web` to splitting it up into libs.

libs/financial-aid/shared/src/components/FileList/FileList.tsx
```diff
@@ -7,7 +7,7 @@ import {
   ApplicationFile,
   getFileSizeInKilo,
   getFileType,
-} from '@island.is/financial-aid/shared'
+} from '@island.is/financial-aid/shared/index'
```
eirikurn, line 10: I can imagine you did not export this component in `index` to avoid circular imports. Splitting things into more libs can help with this.

eirikurn (approved):
Great! I feel like the `/index` changes are a bit ugly, but your call if you want to tweak this further here or later. 👍

PR #4872, opened 2021-09-06
fix: Suggestion towards a fix for file extension being removed in s3

Thorrik58:
# Fix for file storage name

## What

Format filenames without the extension of the filename to prevent breaking application with attachments.

## Why

This was breaking the health insurance application due to the filename not having an extension.

## Screenshots / Gifs

Before:
image
After:
image

## Checklist:


libs/file-storage/src/lib/file-storage.service.ts
```diff
@@ -24,8 +24,15 @@ export class FileStorageService {
       throw new Error('Upload bucket not configured.')
     }
 
+    // To make sure we dont format the file extension when its present. If its not present return the filename
+    const splitFileName = filename.split('.')
```
saevarma, line 28: If `filename` is null or undefined then this will throw error. Do you want to test for that case before using `split()`?

Thorrik58: It doesnt hurt to check for it at least. Even though im not sure if you can upload an file with no filename. 
Belti og axlabönd!

saevarma (commented):
One minor comment.

libs/file-storage/src/lib/file-storage.service.ts
Thorrik58: Throwing an error when filename is missing, we could also define the filename as 'unknown-file' or something to make sure that the upload doesnt break.

saevarma (approved):
LGTM

PR #4864, opened 2021-09-06
feat(web): Syslumenn auctions via Syslumenn API

siggifv:
# Syslumenn auctions via Syslumenn API

## What

Cherry-picking the merge commits from the following PRs from `main`:
- #4822 
- #4823 

## Why

Requested to be deployed asap to production, as this eliminates extensive manual labor for syslumenn staff related to auctions.

## Checklist:

- [x] I have performed a self-review of my own code
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] Formatting passes locally with my changes
- [x] I have rebased against main before asking for a review

PR #4873, opened 2021-09-07
feat(web): Revert: Syslumenn auctions via Syslumenn API

siggifv:
Reverts island-is/island.is#4864

Unfortunately, it turns out that the Auctions endpoint in the Production deployment of the Syslumenn API was not production ready. Therefor we need to revert this and hotfix ASAP.


ingolfurs (approved):

kksteini (approved):
LGTM

PR #4875, opened 2021-09-07
fix(delegation): Notify screen-reader of changes

darrikonn:
# Notify screen-reader of changes

https://www.notion.so/Go-over-a11y-regarding-GrantAccess-We-should-probably-add-assertive-aria-live-to-the-Inputs-b335e1d8860941248f7b13f5aeafb8cf

## What

`aria-live="assertive"` Any updates made to this region are important enough to be announced to the user as soon as possible.

## Checklist:


saevarma (approved):
LGTM

PR #4888, opened 2021-09-07
fix(service-porta-education): preventing siblings from being able to see each others' grades

arnorhs:
## Checklist:

### DIFF
--- a/libs/api/domains/education/src/lib/education.service.ts
+++ b/libs/api/domains/education/src/lib/education.service.ts
@@ -67,8 +67,10 @@ export class EducationService {
   }
 
   private isChild(familyMember: ISLFjolskyldan): boolean {
-    return !['1', '2', '7'].includes(familyMember.Kyn) &&
+    return (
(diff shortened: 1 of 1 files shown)


libs/api/domains/education/src/lib/education.service.ts
```diff
@@ -66,13 +66,25 @@ export class EducationService {
     })
   }
 
-  async getFamily(nationalId: string) {
+  private isChild(familyMember: ISLFjolskyldan): boolean {
+    return (
+      !['1', '2', '7'].includes(familyMember.Kyn) &&
```
eirikurn, line 71: Would be nice to use enum values here for readability.

eirikurn (approved):

pshomov (approved):
LGTM, please add tests for these scenarios

libs/api/domains/education/src/lib/education.service.ts
arnorhs: agree.. or honestly, just using the new national registry service instead.

PR #4922, opened 2021-09-09
fix(application-data-providers): making the national registry provider actually fail on errors

arnorhs:
## Checklist:


libs/application/data-providers/src/providers/NationalRegistryProvider.ts
```diff
@@ -32,31 +33,25 @@ export class NationalRegistryProvider extends BasicDataProvider {
...
+      return Promise.resolve(response.data.nationalRegistryUser)
+    })
   }
 
-  onProvideError(result: string): FailedDataProviderResult {
+  onProvideError(result: unknown): FailedDataProviderResult {
     return {
       date: new Date(),
-      reason: result,
+      reason: coreErrorMessages.errorDataProvider,
```
baering, line 52: Would it not be good that this message indicates where it went wrong? This will just say that something went wrong fetching your data, not that it is connected to the national registry.

arnorhs: good point.. it does show up below the national registry provider:

image

Perhaps I should just add inline translation strings or at least along with the provider - do we have a translation namespace for global providers such as this one?

baering (approved):

PR #4948, opened 2021-09-10
feat(application-system): WIP Allow organisations to make requests to application system

baering:
## What

In order for an institution to make requests to the application system we need to:
- Create a connection between their x-road server (handled by the institution) and our x-road server (handled by Andes)
- When this connection has been established the institution can make a request to the application-system

When the request hits the application system:
- A new client is created for the institution by an IDP admin and the client is given the applications:read and/or :write scopes
- Headers are forwarded so we can make use of `X-Road-Client` to know which institution is making the request
- We modify the code where the user object is generated to also read national id from `X-Road-Client` when the request is coming through x-road
- The request hits the application just like any other authenticated request coming from a user through GraphQL
- An application template can map the institution national id to a role depending on the current state and the institution can fire events and/or edit data

If we see that we need more advanced request handlers we can implement them in the future.

## Why

For institution to be able to make requests to the application system (for example to advance from a state where the user is waiting for their approval or rejection)

## Checklist:

### DIFF
--- a/apps/application-system/api/src/environments/environment.ts
+++ b/apps/application-system/api/src/environments/environment.ts
@@ -21,6 +21,7 @@ const devConfig = {
   auth: {
     issuer: 'https://identity-server.dev01.devland.is',
     audience: '@island.is',
+    allowXRoadClientHeaderAuthentication: true,
   },
   templateApi: {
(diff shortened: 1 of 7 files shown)


andes-it:

andes-it:

PR #4977, opened 2021-09-14
feat(fa): File download

Rikkiabb:
# ...

https://app.asana.com/0/1200897217304888/1200385618296210

## What

Allows downloading of files in veita.
Makes filelist not have hover state nor clickable in Ósk.

## Why
Needed for handling of applications

## Checklist:


libs/financial-aid/shared/src/components/FileList/FileList.tsx
```diff
@@ -9,6 +8,17 @@ import {
...
+import { gql, useLazyQuery } from '@apollo/client'
+
+export const GetSignedUrlQuery = gql`
+  query GetSignedUrlQuery($input: GetSignedUrlForIdInput!) {
+    getSignedUrlForId(input: $input) {
+      url
+      key
+    }
+  }
+`
```
Rikkiabb, line 20: Know this should be in sharedGql, but if this is there then we have a circular dependency 😭  We could add a sharedGql file into our shared lib, but this is the only on there so putting it on hold for now 🧑‍🤝‍🧑

libs/financial-aid/shared/src/components/FileList/FileList.tsx
```diff
@@ -18,18 +28,35 @@ const FileList = ({ className, files }: Props) => {
...
+    },
+  })
+
   return (
     <Box className={cn({ [`${className}`]: true })} marginBottom={2}>
       <>
         {files.map((item, index) => {
           return (
-            <a
+            <Box
```
MargretFinnboga, line 48: div with an onclick event, isnt that bad practice? why not have it a tag ? Just curious👀

apps/financial-aid/backend/src/environments/environment.ts
```diff
@@ -40,7 +41,8 @@ const devConfig = {
   files: {
     cloudFrontPublicKeyId: process.env.CLOUDFRONT_PUBLIC_KEY_ID ?? '',
     cloudFrontPrivateKey: process.env.CLOUDFRONT_PRIVATE_KEY ?? '',
-    fileBaseUrl: 'https://fjarhagsadstod.dev.sveitarfelog.net/files',
+    fileBaseUrlOsk: 'https://fjarhagsadstod.dev.sveitarfelog.net/files',
+    fileBaseUrlVeita: 'https://fjarhagsadstod.dev.sveitarfelog.net/files',
```
MargretFinnboga, line 45: Should this be the same links? Then why have them twice? 👯‍♀️

Rikkiabb: This a change that was here when I took over the PR, but great question, will change, this doesn't make sense

libs/financial-aid/shared/src/components/FileList/FileList.tsx
Rikkiabb: True 💯 will change

MargretFinnboga (approved):
😎

PR #4982, opened 2021-09-14
fix(fa): Encodes filename

Rikkiabb:
# ...
https://app.asana.com/0/1200897217304888/1200951215441482

## What
Encode filename before uploading to s3

## Why
We get an error if we don't and upload files with icelandic characters

## Checklist:


MargretFinnboga (approved):
🏆  🎖️

PR #5000, opened 2021-09-14
fix(api-domains-education): Prevent the national IDs from showing up in URLs

arnorhs:
The national IDs are showing up in analytics applications, and possibly even some other places since they are presented as part of the URLs and used in navigation.

This change should make it so we access them by index within your visible family members, when you can see more than a single one. This affects both the showing of URLs in the service portal's education part, as well as the way these things are presented.

I also added a few tests for the case of visibility between different family members and for the ordering

## Checklist:


libs/api/domains/education/src/lib/education.service.ts
```diff
@@ -80,19 +79,37 @@ export class EducationService {
...
+    // family memember, so that indexes are consistant no matter how they
+    // are displayed or fetched.
+    // They also don't need to be absolute indexes, only indexes that are
+    // unique from the point of view of each viewer.
+
+    return family
+      .filter(
+        (familyMember) =>
+          familyMember === myself ||
+          (this.isChild(familyMember) && !this.isChild(myself)),
```
osk, line 93: This is kinda hard to read, filter out all other adults in the family and don't show it if I'm <18?

osk (approved):
Looks good, although the index feels brittle. Backseat implementing here, but what about hashing the nationalId?

libs/api/domains/education/src/lib/education.service.ts
arnorhs: yes, you are right.

if the family member we are viewing is a child, and we are an adult, we show allow the family member through.

I updated the code to make it more clear to read, and moved the logic to its own function, and added extra tests

arnorhs:
@osk 
> Looks good, although the index feels brittle. Backseat implementing here, but what about hashing the nationalId?

yes, the problem there is that is a form of deanonymization, since this hashed ID would be either a) absolutely unique or b) have possible collisions within the same family (probably unlikely) - there is also some index attribute within the national id query results, in the form of some meta information, but it all falls on one side of this line, and this is really the only thing I could come up with that's both consistent, unidentifiable and not too complicated.

but I'm also being maximally careful/particular with this thing, since a) INAL, b) this implementation is the result of an incident that came up and c) this does affect a lot of people.

libs/api/domains/education/src/lib/education.service.ts
```diff
@@ -144,10 +169,11 @@ export class EducationService {
     }
   }
 
-  private mapCourseGrade(grade: Grade) {
+  private mapCourseGrade(grade: Grade): any {
```
eirikurn, line 172: Can we use an interface here to make sure its use is type checked?

libs/api/domains/education/src/lib/mock-data/my-family.ts
```diff
@@ -0,0 +1,77 @@
...
+// national IDs have been changed to protect the identity of said family.
+// Extra note: it would be nice to have more family cases to test.
+export const MyFamilyMock = [
+  {
+    attributes: {
+      'diffgr:id': 'ISLFjolskyldan2',
+      'msdata:rowOrder': '1',
+      'diffgr:hasChanges': 'inserted',
+    },
+    Kennitala: '0910819979',
```
eirikurn, line 13: Let's make sure all of these kt's are invalid since we don't want real people's kts to be here (even by accident) if we can avoid it. There's a lint warning for this btw.

eirikurn (changes requested):

Thorrik58 (approved):
LGTM

libs/api/domains/education/src/lib/mock-data/my-family.ts
arnorhs: yup, i know. Though those are not real national IDs, even though they are valid.

The problem is that the service uses the kennitala package to determine people's age, and the age doesn't work for invalid kennitala.

But I'll mock the package and ping you when it's done. Then I can get rid of this stuff.

albinagu (approved):

libs/api/domains/education/src/lib/mock-data/my-family.ts
arnorhs: yeah, it was a bit of a headache to mock the package, mostly with having it play nicely with typescript, but the result is not __too__ bad 😮 maybe.. anyways.. i'd appreciate a second glance.

agree re: whitelisted national IDs, possibly for other stuff as well.. but generating the non-real-but-valid national IDs was messy too, and being able to have the mock handle all cases in our tests as well is kind of nice.

eirikurn (approved):
Better!

I would probably have mocked `kennitala.info` closer to the mock data with a table of kt -> info instead of dealing with `__setAge` typescript party. But I like that these "test kennitalas" won't grow too old with time :)

Fyi, we'll need to update `kennitala` (and all of these domains/modules) to work with [kerfiskennitölur](https://www.skra.is/folk/eg-i-thjodskra/um-kennitolur/um-kerfiskennitolur/) soon, which don't have any included birthday information or even vartölutékk. 😬

We can only assume that these are people (not companies), but we need to fetch any other information (including birth date) from the kerfiskennitölu DB.

We'll also need to support eIDAS at some point (from other countries). :)

PR #5002, opened 2021-09-14
feat(auth): improve delegation name handling

eirikurn:
## What

* Get `delegation.fromName` from the ID token.
* Validate that `delegation.toNationalId` is a valid person by calling the national registry.
* Stop passing toName from UI to GraphQL to Auth Public API.

## Why

* To support delegations from companies.
* To validate the toNationalId.
* To stop passing down names from the UI.

## Checklist:


libs/auth-api-lib/src/lib/services/delegations.service.ts
```diff
@@ -169,45 +184,76 @@ export class DelegationsService {
...
     this.logger.debug('Creating a new delegation')
     const id = uuid()
     await this.delegationModel.create({
       id: id,
       fromNationalId: user.nationalId,
       toNationalId: delegation.toNationalId,
-      fromDisplayName: person.fulltNafn || person.nafn,
-      toName: delegation.toName,
+      fromDisplayName,
+      toName,
```
darrikonn, line 253: Should we not skip saving the names altogether. Just save the nationalIds and rely on the Identity resolver to fetch the correct names when appropriate?
Would fix cases where the user has e.g. changed their name etc.

darrikonn (approved):
Nice work. One comment, otherwise approved.

libs/auth-api-lib/src/lib/services/delegations.service.ts
eirikurn: Great comment! That's absolutely the plan, but it depends on us getting a company registry api, implementing that fancy caching logic everywhere and performing name lookups in the IDS delegation picker.

This PR is a temporary improvement for the beta test.

valurefugl (approved):

PR #5009, opened 2021-09-15
feat(license-service): add Scopes

osk:
# License service: add Scopes

Adds scopes to the generic license service:

* `@island.is/internal`, shared generic scope for APIs until an explicit scope is needed
* `@island.is/licenses:verify`, can verify licenses

Scopes have only been created in **dev environment** and attched to clients:

* `@island.is/internal` => `island-is-1`, `@island.is/app`, and `@island.is/apps/scanner`
* `@island.is/licenses:verify` => `@island.is/app`, `@island.is/apps/scanner`

## Checklist:


libs/auth/scopes/src/lib/licenses.scope.ts
```diff
@@ -0,0 +1,10 @@
+export enum LicenseScope {
+  // Client can read all licenses
+  read = '@island.is/licenses:read',
+
+  // Client can generate licenses to display (e.g. pkpass for drivers license)
+  generate = '@island.is/licenses:generate',
```
eirikurn, line 6: The plan is to start with a shared `@island.is/internal` scope for most APIs until we have a use case for explicit scopes (either delegations or external clients). 

I don't know ... `GenericScope.internal`? :)

eirikurn (changes requested):

libs/auth/scopes/src/lib/licenses.scope.ts
osk: OK, will add, the generic scope, name sounds fine.

eirikurn (approved):
Great! Let's get these scopes created and added to the web and app clients.

osk:
Old and outdated

PR #5016, opened 2021-09-15
chore(cms): Regenerating contentful-types and fixing type errors

kksteini:
## What

The generated types file for contentful had become desynced from what is actually on contentful today, [see generator](https://contentful-type-generator.shared.devland.is/)

The schema file was updated and deprecated types, and associated uses, were removed from the project.

A way forward should be to ensure synchronization between the `libs/cms/src/lib/generated/contentfulTypes.d.ts` and contentful so that it does not de-sync again.

## Checklist:


libs/cms/src/lib/cms.module.ts
```diff
@@ -5,7 +5,6 @@ import {
   CmsResolver,
   ArticleResolver,
   LatestNewsSliceResolver,
-  AboutSubPageResolver,
```
kksteini, line 8: The AboutSubPageResolver utilizes `AboutPage` which is no longer in the content model. I was aiming to refactor AboutSubPage out of the project in a separate pull request but this deletion has to happen here.

siggifv (approved):
LGTM

apps/web/components/Organization/Slice/TabSection/TabSection.tsx
```diff
@@ -1,7 +1,7 @@
 import React from 'react'
 import { TabSection } from '@island.is/web/graphql/schema'
 import { Box, GridColumn, GridRow, Tabs, Text } from '@island.is/island-ui/core'
-import * as styles from '@island.is/web/screens/AboutPage/AboutPage.treat'
+import * as styles from '@island.is/web/screens/Organization/Organization.treat'
```
arnorhs, line 4: will this affect the style of the tab section?

kksteini: I created a file called Organization.treat since the TabSection was under Organization.
I copied and pasted the code from AboutPage.treat into Organization.treat, a new file.
I assumed this would not change the styles. Am I missing something?

arnorhs (approved):

Issue #5021, opened 2021-09-16
`AccordionItem`'s `startExpanded` overrides local `expanded` state


maranomynet:
My expectation was that the `startExpanded` would only take effect when the component mounts, and thereafter local state would maintain control — similar to how `defaultValue` works on `<input />`s.

https://github.com/island-is/island.is/blob/be79b5cf3413e8a18e60b04b3c1ccdbcc458672a/libs/island-ui/core/src/lib/Accordion/AccordionItem/AccordionItem.tsx#L119-L125

However, the dependency array on the above `useEffect` hook triggers an aggressive reset every time `startExpanded` changes.

Is this really the intended behavior?

My suggestion would be to empty out the dependencies array and only apply the "startExpanded" on mount, like so:

```ts
    useEffect(() => {
      if (startExpanded) {
        onHandleOpen()
      }
    }, [])
```

(NOTE: triggering `handleClose` would be rendundant since closed is the default state)

maranomynet:
Fun fact: This code renders a **closed** `<AccordionItem />`

```jsx
<AccordionItem
  id="foo"
  label="I am a label"
  onToggle={() => undefined}
  expanded={true}
  startExpanded={true}
>
  I should start visible, but I don't
</AccordionItem>
```

maranomynet:
Thus I'd suggested change would be this:

```ts
    useEffect(() => {
      if (startExpanded && expandedProp == null) {
        onHandleOpen()
      }
    }, [])
```

PR #5024, opened 2021-09-16
fix(island-ui): Wonky `AccordionItem` state props behavior

maranomynet:
See #5021 for more details on what and why.

This PR fixes that issue and does a bit of additional house-keeping.

The commit messages also include more clarifications

## Checklist:


libs/island-ui/core/src/lib/Accordion/AccordionItem/AccordionItem.tsx
```diff
@@ -101,28 +106,15 @@ export const AccordionItem = forwardRef<HTMLButtonElement, AccordionItemProps>(
...
-    useEffect(() => {
-      if (startExpanded) {
-        onHandleOpen()
-      } else {
-        handleClose()
-      }
-    }, [onHandleOpen, startExpanded])
+    useEffect(
+      () => {
+        if (startExpanded && expandedProp == null) {
```
oddsson, line 111:
```suggestion
        if (startExpanded && expandedProp === null) {
```

maranomynet: Nope, this is checking if `expandedProp` is either `null` or `undefined`.

The options are either using `==` or going with fully expaneded double-check, like so:

```ts
        if (startExpanded && (expandedProp === null || expandedProp === undefined)) {
```

which is IMO excessively verbose.

maranomynet: I guess this would be sufficient, since the type signature does not allow `null`.

```ts
        if (startExpanded && expandedProp === undefined) {
```

eirikurn: I'm a fan of generally allowing `== null` and `!= null` to catch both undefined and null cases but I'm an old paranoid geezer. 😄

eirikurn (approved):

libs/island-ui/core/src/lib/Accordion/AccordionItem/AccordionItem.tsx
maranomynet: Yeah, for me it's a habit to use `== null` almost like a shorthand for `=== undefined` and the fact it also captures any stray `null`s that might hypothetically float through (perhaps via slightly inaccurate upstream typings (often API related)) just feels like a minor bonus.

oddsson (approved):

PR #5034, opened 2021-09-16
feat(rsk): Implement rsk company info search and company info

Thorrik58:
# Implement rsk company info search and company info

## What

Add company info and company search for rsk api.

## Why

In order to lookup companies by national id or name

## Screenshots / Gifs

## Checklist:


libs/api/domains/rsk-company-info/src/lib/graphql/rsk-company-info.service.ts
```diff
@@ -0,0 +1,108 @@
...
+                  href: link.href,
+                } as RskCompanyLink),
+            ),
+          } as RskCompanySearchItem),
+      ),
+      hasMore: searchResults.hasMore,
+      limit: searchResults.limit,
+      offset: searchResults.offset,
+      count: searchResults.count,
+      links: searchResults.links,
```
eirikurn, line 105: I assume these link objects are for paginating the REST api. We should wrap this in the GraphQL API.

Please update the GraphQL schema to match our [API design guide](https://docs.devland.is/technical-overview/api-design-guide/graphql-naming-conventions#fetching-array-of-items). Hint: it's common to hide implementation details in `cursor` values with`base64`, e.g. we could base64 the whole link url, or some part of it if we assume the link structure is stable.

libs/api/domains/rsk-company-info/src/lib/graphql/rsk-company-info.service.ts
```diff
@@ -0,0 +1,108 @@
...
+            status: item.stada,
+            vat: item.vskNumer,
+            lastUpdated: item.sidastUppfaert,
+            links: item.links?.map(
+              (link) =>
+                ({
+                  rel: link.rel,
+                  href: link.href,
+                } as RskCompanyLink),
+            ),
```
eirikurn, line 98: If this is just to provide a link to the details REST api, we should skip this field. Clients can manually call the other GraphQL query to get extra details. We can also provide a field resolver on `RskCompanySearchItem` that returns the full `RskCompany` from the API.

libs/api/domains/rsk-company-info/src/lib/graphql/rsk-company-info.service.ts
```diff
@@ -0,0 +1,108 @@
...
+    )
+    return {
+      items: searchResults.items?.map(
+        (item) =>
+          ({
+            nationalIdCompany: item.kennitala,
+            name: item.nafn,
+            dateOfRegistration: item.skrad,
+            status: item.stada,
+            vat: item.vskNumer,
```
eirikurn, line 90: Should be `vatNumber`?

libs/api/domains/rsk-company-info/src/lib/graphql/rsk-company-info.service.ts
```diff
@@ -0,0 +1,108 @@
...
+    const searchResults = await this.rskCompanyInfoApi.searchCompanyInformation(
+      searchTerm,
+    )
+    return {
+      items: searchResults.items?.map(
+        (item) =>
+          ({
+            nationalIdCompany: item.kennitala,
+            name: item.nafn,
+            dateOfRegistration: item.skrad,
```
eirikurn, line 88: Are the REST API dates ISO formatted?

libs/api/domains/rsk-company-info/src/lib/graphql/rsk-company-info.service.ts
```diff
@@ -0,0 +1,108 @@
...
+    searchTerm: string,
+  ): Promise<RskCompanySearchItems> {
+    const searchResults = await this.rskCompanyInfoApi.searchCompanyInformation(
+      searchTerm,
+    )
+    return {
+      items: searchResults.items?.map(
+        (item) =>
+          ({
+            nationalIdCompany: item.kennitala,
```
eirikurn, line 86: Why `nationalIdCompany`? I'd prefer `nationalId`.

libs/api/domains/rsk-company-info/src/lib/graphql/rsk-company-info.service.ts
```diff
@@ -0,0 +1,108 @@
...
+          } as RskCompanyAddress),
+      ),
+      relatedParty: company.tengdirAdilar?.map(
+        (item) =>
+          ({
+            type: item.tegund,
+            nationalId: item.kennitala,
+            name: item.nafn,
+          } as RskCompanyRelatedParty),
+      ),
```
eirikurn, line 70: Do we know what this is?

I'd be happy if this includes a list of "Prókúruhafi" among other roles, but I can't find that information on their public website so we should ask if it can be exposed in our external API.

libs/api/domains/rsk-company-info/src/lib/graphql/rsk-company-info.service.ts
```diff
@@ -0,0 +1,108 @@
...
+        (item) =>
+          ({
+            address: item.heimilisfang1,
+            address2: item.heimilisfang2,
+            postNumber: item.postnumer,
+            municipality: item.sveitarfelag,
+            municipalityNumber: item.sveitarfelagsnumer,
+            country: item.land,
+          } as RskCompanyAddress),
+      ),
```
eirikurn, line 62: Ah great! A new integration, a new address schema :)

eirikurn: We don't want to lose any extra fields here, but I'd like it if the naming can be consistent with other places. E.g. there is an `Address` ObjectType already which includes `streetAddress`, `postalCode` and `city`.

libs/api/domains/rsk-company-info/src/lib/graphql/rsk-company-info.service.ts
```diff
@@ -0,0 +1,108 @@
...
+        (item) =>
+          ({
+            vatNumber: item.vskNumer,
+            dateOfRegistration: item.skrad,
+            status: item.stada,
+            deregistration: item.afskraning,
+            classification: item.flokkun?.map(
+              (classification) =>
+                ({
+                  type: classification.gerd,
```
eirikurn, line 44: I wish we had enums for all of these type fields. :(

libs/api/domains/rsk-company-info/src/lib/graphql/rsk-company-info.service.ts
```diff
@@ -0,0 +1,108 @@
...
+            type: item.tegund,
+            name: item.heiti,
+            suffix: item.vidskeyti,
+          } as RskCompanyFormOfOperation),
+      ),
+      vat: company.virdisaukaskattur?.map(
+        (item) =>
+          ({
+            vatNumber: item.vskNumer,
+            dateOfRegistration: item.skrad,
```
eirikurn, line 38: Date format?

libs/api/domains/rsk-company-info/src/lib/graphql/rsk-company-info.service.ts
```diff
@@ -0,0 +1,108 @@
...
+            status: item.stada,
+            deregistration: item.afskraning,
```
eirikurn, line 40: Date format?

libs/api/domains/rsk-company-info/src/lib/graphql/rsk-company-info.service.ts
```diff
@@ -0,0 +1,108 @@
...
+  constructor(private rskCompanyInfoApi: RskCompanyInfoAPI) {}
+
+  async getCompanyInformation(nationalId: string): Promise<RskCompany> {
+    const company = await this.rskCompanyInfoApi.getCompanyInformation(
+      nationalId,
+    )
+    return {
+      nationalIdCompany: company.kennitala,
+      name: company.nafn,
+      dateOfRegistration: company.skrad,
```
eirikurn, line 24: Date format?

libs/api/domains/rsk-company-info/src/lib/graphql/rsk-company-info.service.ts
```diff
@@ -0,0 +1,108 @@
...
+@Injectable()
+export class RskCompanyInfoService {
+  constructor(private rskCompanyInfoApi: RskCompanyInfoAPI) {}
+
+  async getCompanyInformation(nationalId: string): Promise<RskCompany> {
+    const company = await this.rskCompanyInfoApi.getCompanyInformation(
+      nationalId,
+    )
+    return {
+      nationalIdCompany: company.kennitala,
```
eirikurn, line 22: `nationalId`?

eirikurn (changes requested):
Love that we're getting this API! 🎉🎉🎉 

We need to integrate this into the `Identity` domain so we can have a single query that works for both people and companies by nationalId. Fine to do this in a follow up. We had some ideas on maybe sharing schema types between Identity and other domains, and use graphql union/interface types for Person/Company. Can reach out to @darrikonn for more details 👍

eirikurn:
It's unfortunate that there are now two APIs in X-Road called "Company Registry" from "Skatturinn" with different endpoints. Hope they merge/rename them. :)

libs/api/domains/rsk-company-info/src/lib/graphql/rsk-company-info.service.ts
Thorrik58: You are correct, will fix 👍

libs/api/domains/rsk-company-info/src/lib/graphql/rsk-company-info.service.ts
Thorrik58: Sure thing, will fix

libs/api/domains/rsk-company-info/src/lib/graphql/dto/getRskCompanyInfoInput.ts
```diff
@@ -0,0 +1,9 @@
+import { Field, InputType } from '@nestjs/graphql'
+import { IsString } from 'class-validator'
+
+@InputType()
+export class GetRskCompanyInfoInput {
```
darrikonn, line 5: Both the filename and the GraphQL Input are not following the filenaming conventions of Island.is:
https://docs.devland.is/technical-overview/api-design-guide/graphql-naming-conventions

Prefer `RskCompanyInfoInput` over `Get.*`.
The filename would then be `*.input.ts`.

This goes for the other inputs as well.

libs/api/domains/rsk-company-info/src/lib/graphql/models/rskCompanySearchItems.model.ts
```diff
@@ -0,0 +1,24 @@
+import { Field, ObjectType, ID } from '@nestjs/graphql'
+import { RskCompanyLink } from './rskCompanyLink.model'
+import { RskCompanySearchItem } from './rskCompanySearchItem.model'
+
+@ObjectType()
+export class RskCompanySearchItems {
```
darrikonn, line 6: Should follow a simplified version of Relays connection pattern:
https://docs.devland.is/technical-overview/api-design-guide/graphql-naming-conventions#fetching-array-of-items

libs/api/domains/rsk-company-info/src/lib/graphql/models/rskCompanySearchItems.model.ts
```diff
@@ -0,0 +1,24 @@
...
+  @Field(() => [RskCompanySearchItem])
+  items?: RskCompanySearchItem[]
```
darrikonn, line 8: Should this not return `RskCompany`. This would help with updating the Apollo cache in the front-end.

libs/api/domains/rsk-company-info/src/lib/graphql/rsk-company-info.resolver.ts
```diff
@@ -0,0 +1,43 @@
...
+import { GetRskCompanyInfoSearchInput } from './dto/getRskCompanyInfoSearchInput'
+import { RskCompanySearchItems } from './models/rskCompanySearchItems.model'
+
+@UseGuards(IdsUserGuard, ScopesGuard)
+@Resolver()
+export class RskCompanyInfoResolver {
+  constructor(private rskCompanyInfoService: RskCompanyInfoService) {}
+
+  @Query(() => RskCompany, {
+    name: 'rskCompanyInfo',
```
darrikonn, line 17: Name: `rskCompany`?

libs/api/domains/rsk-company-info/src/lib/graphql/rsk-company-info.resolver.ts
```diff
@@ -0,0 +1,43 @@
...
+    @Args('input', { type: () => GetRskCompanyInfoInput })
+    input: GetRskCompanyInfoInput,
+  ): Promise<RskCompany> {
+    return await this.rskCompanyInfoService.getCompanyInformation(
+      input.nationalId,
+    )
+  }
+
+  @Query(() => RskCompanySearchItems, {
+    name: 'rskCompanyInfoSearch',
```
darrikonn, line 31: And this then the plural of `rskCompany`:
`rskCompanies`

darrikonn (changes requested):
Will be nice to get this into the identity domain :+1:

libs/api/domains/rsk-company-info/src/lib/graphql/models/rskCompanySearchItems.model.ts
eirikurn: I was wondering about this too.

Should we do N+1 queries to get full company information in search results? Or should we have a `RskCompanySearchItem` with basic details and a `searchItem.company` field resolver which returns the full company details?

darrikonn: We could extract the difference into a new field with its own field resolver (info).
This means that fetching the default search items only uses a single query, but if more detailed result is needed for the array, then n+1 calls will be made.

libs/api/domains/rsk-company-info/src/lib/graphql/models/rskCompany.model.ts
```diff
@@ -0,0 +1,45 @@
...
+  formOfOperation?: RskCompanyFormOfOperation[]
+
+  @Field(() => [RskCompanyAddress], { nullable: true })
+  address?: RskCompanyAddress[]
+
+  @Field(() => [RskCompanyRelatedParty], { nullable: true })
+  relatedParty?: RskCompanyRelatedParty[]
+
+  @Field(() => [RskCompanyVat], { nullable: true })
+  vat?: RskCompanyVat[]
```
eirikurn, line 20: I generally prefer defaulting these kind of fields to empty arrays instead of nullable, unless there's a reason to differentiate the two.

libs/api/domains/rsk-company-info/src/lib/graphql/models/rskCompanyAddress.model.ts
```diff
@@ -0,0 +1,22 @@
+import { Field, ObjectType } from '@nestjs/graphql'
+
+@ObjectType()
+export class RskCompanyAddress {
+  @Field(() => String, { nullable: true })
```
eirikurn, line 5: so. many. nullables.

Guess that's the real world :)

libs/api/domains/rsk-company-info/src/lib/graphql/models/rskCompanySearchItem.model.ts
```diff
@@ -0,0 +1,31 @@
+import { Field, ObjectType, ID } from '@nestjs/graphql'
+import { RskCompany } from './rskCompany.model'
+import { RskCompanyLink } from './rskCompanyLink.model'
+import { RskCompanyVat } from './rskCompanyVat.model'
+
+@ObjectType()
+export class RskCompanySearchItem {
```
eirikurn, line 7: Do we need this now that we have RskCompany?

libs/api/domains/rsk-company-info/src/lib/graphql/models/rskCompanySearchItems.model.ts
```diff
@@ -0,0 +1,23 @@
...
+export class RskCompanySearchItems {
+  @Field(() => [RskCompany], { nullable: true })
+  items?: RskCompany[]
+
+  @Field(() => RskCompanyPageInfo, { nullable: true })
+  pageInfo?: RskCompanyPageInfo
+
+  @Field(() => Int, { nullable: true })
+  count?: number
+}
```
eirikurn, line 23: Any input on this @saevarma after our recent pagination discussions?

libs/api/domains/rsk-company-info/src/lib/graphql/rsk-company-info.resolver.ts
```diff
@@ -0,0 +1,77 @@
...
+    private rskCompanyInfoService: RskCompanyInfoService,
+    @Inject(LOGGER_PROVIDER)
+    private logger: Logger,
+  ) {}
+
+  @Query(() => RskCompany, {
+    name: 'rskCompany',
+    nullable: true,
+  })
+  // @Audit()
```
eirikurn, line 26: Not sure if we need audit in this domain (company registry lookups). @hrefnalind?

libs/api/domains/rsk-company-info/src/lib/graphql/rsk-company-info.resolver.ts
```diff
@@ -0,0 +1,77 @@
...
+  ): Promise<RskCompany[]> {
+    this.logger.debug('RskCompanySearchItems')
+    return parent.items ?? []
+  }
+
+  @Audit()
+  @ResolveField(() => RskCompany)
+  async companyInfo(
+    @Parent() rskCompanyItem: RskCompany,
+  ): Promise<RskCompany | null> {
```
eirikurn, line 71: Should this return RskCompanyInfo?

libs/api/domains/rsk-company-info/src/lib/graphql/rsk-company-info.resolver.ts
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  @Audit()
+  @ResolveField(() => RskCompany)
+  async companySearchItems(
+    @Parent() parent: RskCompanySearchItems,
+  ): Promise<RskCompany[]> {
+    this.logger.debug('RskCompanySearchItems')
+    return parent.items ?? []
+  }
```
eirikurn, line 65: Does this do anything? Since there's no companySearchItems field on RskCompany.

libs/api/domains/rsk-company-info/src/lib/graphql/rsk-company-info.resolver.ts
```diff
@@ -0,0 +1,77 @@
...
+
+  @Audit()
+  @ResolveField(() => RskCompany)
+  async companyInfo(
+    @Parent() rskCompanyItem: RskCompany,
+  ): Promise<RskCompany | null> {
+    this.logger.debug('Resolving companyInfo')
+    return await this.rskCompanyInfoService.getCompanyInformationWithExtra(
+      rskCompanyItem.nationalId ?? '',
+    )
```
eirikurn, line 75: Should maybe check rskCompanyItem.companyInfo first?

libs/api/domains/rsk-company-info/src/lib/graphql/rsk-company-info.service.ts
```diff
@@ -0,0 +1,164 @@
...
+      )
+      const start = index + 1
+
+      return {
+        items: searchResult.slice(start, start + first),
+        hasNextPage: start + first < searchResult.length,
+      }
+    }
+    return { items: [], hasNextPage: false }
+  }
```
eirikurn, line 155: This is only doing pagination in our graphql layer.

Does the original API not have any pagination setup? I assumed it used "links" for this.

libs/api/domains/rsk-company-info/src/lib/graphql/models/rskCompany.model.ts
```diff
@@ -0,0 +1,45 @@
...
+  dateOfRegistration!: Date
+
+  @Field(() => String)
+  status!: string
+
+  @Field(() => String)
+  vatNumber?: string
+
+  @Field(() => Date)
+  lastUpdated?: Date
```
eirikurn, line 41: Looks like this needs to be nullable.

libs/api/domains/rsk-company-info/src/lib/graphql/rsk-company-info.service.ts
```diff
@@ -0,0 +1,164 @@
...
+            nationalId: item.kennitala,
+            name: item.nafn,
+          } as RskCompanyRelatedParty
+        }),
+        vat: company.virdisaukaskattur?.map((item) => {
+          return {
+            vatNumber: item.vskNumer,
+            dateOfRegistration: new Date(item.skrad),
+            status: item.stada,
+            dateOfDeregistration: new Date(item.afskraning),
```
eirikurn, line 66: Are these Date fields always set? Otherwise we may end up with some weird dates.

libs/api/domains/rsk-company-info/src/lib/rsk-company-info-module.ts
```diff
@@ -0,0 +1,37 @@
...
+      module: RskCompanyInfoModule,
+      providers: [
+        RskCompanyInfoResolver,
+        RskCompanyInfoService,
+        {
+          provide: CompanyApi,
+          useFactory: () =>
+            new CompanyApi(
+              new RestConfiguration({
+                fetchApi: fetch,
```
eirikurn, line 25: Please use [createEnhancedFetch](https://github.com/island-is/island.is/blob/2fdf9c5f2aaac1607a0e8a1a1ab363f2e326e1e0/libs/clients/middlewares/README.md)

libs/clients/rsk-company-info/src/lib/rsk-company-info.ts
```diff
@@ -0,0 +1,36 @@
+import { Inject } from '@nestjs/common'
+import { DataSourceConfig } from 'apollo-datasource'
+import { RequestOptions, RESTDataSource } from 'apollo-datasource-rest'
+import { CompanyItem, SearchCompanyItem, SearchCompanyItems } from './gen/fetch'
+import { RskCompanyInfoServiceOptions, RSK_COMPANY_INFO_OPTIONS } from './types'
+
+export class RskCompanyInfoAPI extends RESTDataSource {
```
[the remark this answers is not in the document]
eirikurn: Should this file be removed?

libs/api/domains/rsk-company-info/src/lib/rsk-company-info-module.ts
```diff
@@ -0,0 +1,37 @@
...
+    return {
+      module: RskCompanyInfoModule,
+      providers: [
+        RskCompanyInfoResolver,
+        RskCompanyInfoService,
+        {
+          provide: CompanyApi,
+          useFactory: () =>
+            new CompanyApi(
+              new RestConfiguration({
```
eirikurn, line 24: This stuff can be done in a RskCompanyInfoModule which is part of the RskCompanyInfo client library. Check out VMST as an example, [client module](https://github.com/island-is/island.is/blob/a737ca6b15e902fe9d1e214d87d3e585774fc7e7/libs/clients/vmst/src/lib/vmst.module.ts#L24) vs [domain](https://github.com/island-is/island.is/blob/a737ca6b15e902fe9d1e214d87d3e585774fc7e7/libs/api/domains/directorate-of-labour/src/lib/directorate-of-labour.module.ts#L30-L41).

libs/api/domains/rsk-company-info/src/lib/graphql/rsk-company-info.service.ts
eirikurn: Great, let's pipe pagination directly to the backend. The cursor can be the offset for the next page, maybe base64 encoded. cc @saevarma

libs/api/domains/rsk-company-info/src/lib/graphql/rsk-company-info.resolver.ts
```diff
@@ -10,7 +9,7 @@ import { RskCompanySearchItems } from './models/rskCompanySearchItems.model'
 import { LOGGER_PROVIDER } from '@island.is/logging'
 import type { Logger } from '@island.is/logging'
 
-@UseGuards(IdsUserGuard, ScopesGuard)
+// @UseGuards(IdsUserGuard, ScopesGuard)
```
eirikurn, line 12: Should these resolvers and all returned data be publicly accessible?

libs/api/domains/rsk-company-info/src/lib/graphql/rsk-company-info.service.ts
```diff
@@ -111,54 +121,21 @@ export class RskCompanyInfoService {
...
-      items: pageResult.items,
-      count: totalLength,
+      data: formattedSearchResults,
+      totalCount: totalLength,
       pageInfo: {
         endCursor: cursor,
-        hasNextPage: pageResult.hasNextPage,
+        hasNextPage: !!searchResults.hasMore,
+        hasPreviousPage: resultOffset > 0,
+        startCursor: toBase64(resultOffset.toString()),
```
eirikurn, line 137: @saevarma This API currently doesn't support paginating with "before". Did we decide that should be required for all APIs? :)

If not, we should get rid of "hasPreviousPage" and "startCursor" since they're not really useful if the API doesn't support "before".

libs/api/domains/rsk-company-info/src/lib/graphql/rsk-company-info.service.ts
```diff
@@ -84,11 +90,15 @@ export class RskCompanyInfoService {
 
   async companyInformationSearch(
     searchTerm: string,
-    first: number,
-    after = '',
+    limit: number,
+    offset = '0',
```
eirikurn, line 94: Is this a valid base64 encoded offset? Probably better to call this `cursor?: string` and do `const offset = cursor ? decodeBase64(cursor) : 0`

libs/api/domains/rsk-company-info/src/lib/rsk-company-info-module.ts
```diff
@@ -1,36 +1,23 @@
...
-              }),
-            ),
-        },
+      providers: [RskCompanyInfoResolver, RskCompanyInfoService],
+      imports: [
+        RskCompanyInfoClientModule.register({
+          xRoadBaseUrl: config.xRoadBaseUrl,
+          xRoadProviderId: config.xRoadProviderId,
+          xRoadClientId: config.xRoadClientId,
+        }),
```
eirikurn, line 20: Alternatively, you can **now** use our `defineConfig` and skip the `register` function :D 

Sorry, I'll stop pushing the goal post for this PR :D

libs/clients/rsk-company-info/src/openapi.yml
```diff
@@ -20,6 +20,21 @@ paths:
...
             type: string
             default: Miracle%20ehf
+        # TODO: Limit and offset were manually added in. Update accordingly after RSK has updated their openapi to include limit and offset
+        - name: limit
+          in: query
+          required: false
+          description: Limit the number of results
+          schema:
+            type: string
+            default: '100'
```
eirikurn, line 30: You can see some examples of scripted modifications in workspace.json "update-openapi-document" commands using `jq` black magic. But ideally they fix this on their side.

Don't need to stress over this comment. Just an FYI.

libs/api/domains/rsk-company-info/src/lib/graphql/rsk-company-info.service.ts
saevarma: We did not make it optional fields in the documentation. But I think we should allow for the `hasPreviousPage` and `startCursor` to be optional for use cases like this. I will update the pagination docs.

libs/api/domains/rsk-company-info/src/lib/graphql/rsk-company-info.resolver.ts
bensijons: @eirikurn nope my bad here. Was commented out for testing locally, should not have been included in the commit

libs/api/domains/rsk-company-info/src/lib/graphql/rsk-company-info.service.ts
bensijons: @eirikurn good point, fixed

eirikurn (approved):
Great job! 😍

dabbeg (approved):
DevOps files look good :+1:

darrikonn (changes requested):

.github/CODEOWNERS
```diff
@@ -114,6 +114,8 @@
 /libs/application/template-api-modules/src/lib/modules/templates/public-debt-payment-plan     @island-is/sendiradid
 /libs/api/domains/health-insurance/                                                           @island-is/sendiradid
 /libs/application/templates/health-insurance/                                                 @island-is/sendiradid
+/libs/clients/rsk-company-info/                                                               @island-is/sendiradid
+/libs/api/domains/rsk-company-info/                                                           @island-is/sendiradid
```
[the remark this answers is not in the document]
bensijons: 1. Unfortunately there are two APIs from skatturinn for company information, so there are two corresponding clients, rsk and rsk-company-info.

2. I don't think there is no particular reason why they are being kept seperate, except maybe since there were two clients there were made two domains. They have just now been merged into one domain rsk

libs/api/domains/rsk/src/lib/rsk-company-info.service.ts
```diff
@@ -0,0 +1,146 @@
...
+  ): Promise<RskCompany> {
+    this.logger.debug(`Service getting company by nationalId ${nationalId}`)
+    const company = await this.rskCompanyInfoApi.v1FyrirtaekjaskraSsidGet({
+      ssid: nationalId,
+    })
+    this.logger.debug(`Company in service ${company.toString()}`)
+    return {
+      nationalId: company.kennitala,
+      name: company.nafn,
+      dateOfRegistration: new Date(company.skrad),
```
kksteini, line 33: Seeing checks for dates in some places. Should they be in all?

For example line 122:
```javascript
item.skrad ? new Date(item.skrad) : undefined
```

kksteini (approved):

.github/CODEOWNERS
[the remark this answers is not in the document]
darrikonn: Regarding 1. Not sure if I understand you correctly.
You're saying there are two APIs from skatturinn for company information. This PR is implementing one of them.
Is the other version inside `clients/rsk/v(1|2)`?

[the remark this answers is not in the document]
bensijons: @darrikonn yes I believe so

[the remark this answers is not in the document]
darrikonn: So IMO it makes a total sense that this is just another version of the `rsk` client, thus it should live inside `clients/rsk/v3`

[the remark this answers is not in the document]
bensijons: I took a closer look at what the other API and they don't exactly provide the same information. One is for fetching all companies that a kennitala has access to, and the other provide information about a company kennitala (companyName, registraitonDate and more). They also have separate base URLs.  Personally I feel like this isn't really a new version unless there exists a new API from Skatturinn that includes the previous functions (since the function from v2 is not being deprecated, or at least now that I know of).

@darrikonn Thoughts?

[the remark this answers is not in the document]
eirikurn: Tó complicate this further, some of the existing clients are for APIs which should not be used anymore (Vidspyrna related). RSK has created yet another API for getting all companies that a kennitala has access to, which I need to integrate.

So how should all of these be named?

IMO, we should have a separate client lib for each published API we can integrate with. It should be named according to the published API name, but they can be grouped by provider (eg libs/clients/rsk/company-registry). If a publisher publishes two versions of the same API, I would prefer to keep the clients in one library if we can make sure the interface is backwards compatible, but this can sometimes be tricky.

However, when it comes to GraphQL domains, we should try to “simplify” the world. Clients should not map to domains 1-to-1. A domain might pull data from multiple APIs when that makes sense. Basically, domains should be domain-oriented. This is of course very subjective and it’s tricky to design the GraphQL schema in a distributed and consistent way. I don’t have any solutions yet, but would appreciate a schema design sessions in each new integration. Unfortunately many integrations are quite focused (eg for a one-off application) which complicates things further.

IMHO, this specific integration might be represented in some focused CompanyRegistry domain, but it should also be pulled in to the Identity domain.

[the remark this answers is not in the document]
darrikonn: Yeah I see where you're going with this. This is really a question of how we look at this.

1. `v1` | `v2` | `v3` are just three different versions (not necessarily related) of a client, and should not be thought of as the increased number being an improved version. So `v2` is just another version and might contain other endpoints that provide different informations than `v3`.
2. `v1` | `v2` | `v3` are all versions of the exact same client and can be thought of as the higher number is an improved version.

I originally interpreted it as `1`, but seems like you're going with `2`; if I understand you correctly.

It's probably best to generate a general discussion about this on Slack in #dev_faghopur.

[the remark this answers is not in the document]
darrikonn: Yeah I agree @eirikurn .. like to group it by the provider and name them according to the published API name (so stop using `v1`-`vx`, unless we can't integrate and have to go for e.g. `/libs/clients/rsk/company-registry-v1`).

"Clients should not map to domains 1-to-1." -> Definitely!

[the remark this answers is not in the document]
eirikurn: So we’re saying that the Client should be: `/libs/clients/rsk/company-registry

And the domain should be? CompanyRegistry, with companyRegistryCompany query? :)

darrikonn (approved):
Nice work!

PR #5040, opened 2021-09-17
fix(crc): Await tests

Rikkiabb:
## What
Adds await to file tests. 

## Why
To get correct result in tests

## Checklist:


apps/application-system/api/src/app/modules/application/files/file.service.spec.ts
```diff
@@ -346,6 +326,6 @@ describe('FileService', () => {
         PdfTypes.CHILDREN_RESIDENCE_CHANGE,
       )
 
-    expect(act).rejects.toEqual(BadRequestException)
+    expect(act).resolves.toEqual('url')
```
saevarma, line 329: `.resolves` needs also to be awaited.

VilliGunn (approved):

PR #5050, opened 2021-09-20
feat(license-service): refactor smartsolution client; add auth retries

osk:
Refactor smartsolution stuff into its own client (but needs to become an *actual* client) and add support for retrying calls if un authenticated. Move config into env:

* `PKPASS_CACHE_KEY` for the name of token cache in cacheManager
* `PKPASS_CACHE_TOKEN_EXPIRY_DELTA` for how much we shave of the expiry token datetime, so we're not caching up until it expires
* `PKPASS_AUTH_RETRIES`, how many retries to do if 401 returned from service

Keys created dev, staging, prod.

Also adds a `category` to all logs from the licenseService with values of `drivinglicense-service` and `pkpass`.

## Checklist:


apps/api/src/app/environments/environment.ts
```diff
@@ -131,6 +131,10 @@ const devConfig = {
     apiKey: process.env.PKPASS_API_KEY,
     apiUrl: process.env.PKPASS_API_URL,
     secretKey: process.env.PKPASS_SECRET_KEY,
+    cacheKey: process.env.PKPASS_CACHE_KEY ?? 'smartsolution:apitoken',
```
maranomynet, line 134: I worry that `??` might be overly specific. Is there no chance of the env variable being present but empty (`""`)?

thordurhhh (approved):
Lgtm 👍

apps/api/src/app/environments/environment.ts
osk: technically it could happen, but I'm following a pattern in the file and this is not done for prod settings, only as convenience in dev

maranomynet (approved):
LGTM

eirikurn (approved):

PR #5052, opened 2021-09-20
fix(financial-aid): Preventing user from submitting a comment without files

MargretFinnboga:
# Preventing user from submitting a comment without files

https://app.asana.com/0/1201017282841966/1200876741986742

## What

Setting up error state in case the user submits a comment without files

## Checklist:


apps/financial-aid/web-osk/src/routes/status/FileUpload/fileUpload.tsx
```diff
@@ -40,11 +43,13 @@ const FileUpload = () => {
...
   )
 
-  const sendFiles = async () => {
-    if (form?.otherFiles.length <= 0 || router.query.id === undefined) {
-      return
+  useEffect(() => {
+    if (error) {
+      setError(false)
     }
+  }, [form?.otherFiles])
```
Rikkiabb, line 50: I could be wrong..... but is this needed? 😄

Rikkiabb (commented):
On question

apps/financial-aid/web-osk/src/routes/status/FileUpload/fileUpload.tsx
MargretFinnboga: No not really, but kinda weird that the error message is still showing when you have already uploaded a file? And if you cant upload file you have two error messages showing, I put this just in case, what do you think unnecessary?

Rikkiabb: No, my bad, this makes sense 😃

Rikkiabb (approved):
🙆‍♂️

PR #5059, opened 2021-09-20
feat(fa): Assign case.

Rikkiabb:
# ...
https://app.asana.com/0/1201017282841966/1200474172678027

## What
Adds foreign key from application to staff table.
Adds functionality to assign case.
Display employee name in list and application if the name is existing.
Displays button to assign case if staff name is undefined.
Adds a applicationStateHook to change the application state. 
Adds staff id to application table when application is assigned. 
Adds row to table for staff.

## Why
Feature for case assignment.

## Checklist:


apps/financial-aid/backend/migrations/20210920105246-create-staff.js
```diff
@@ -0,0 +1,50 @@
...
+          name: {
+            type: Sequelize.STRING,
+            allowNull: false,
+          },
+          municipalityId: {
+            type: Sequelize.UUID,
+            allowNull: false,
+          },
+          role: {
+            type: Sequelize.ENUM('employee', 'admin'),
```
MargretFinnboga, line 28: Most of our enum have first letter higher case, should it not be here? 🤓

apps/financial-aid/backend/src/app/modules/staff/staff.controller.ts
```diff
@@ -26,17 +17,7 @@ export class StaffController {
...
-
-    if (!staff) {
-      throw new NotFoundException(
-        `Staff with nationalId ${nationalId} not found`,
-      )
-    }
-
-    return staff
+  async getStaffByNationalId(@Param('nationalId') nationalId: string) {
+    return await this.staffService.findByNationalId(nationalId)
```
MargretFinnboga, line 21: Why are you throwing away the NotFoundException? 🤔

MargretFinnboga (commented):
🦕

apps/financial-aid/backend/migrations/20210920105246-create-staff.js
Rikkiabb: This actually a type, but did some manipulation here to match that, but will change the base type to enum instead 😃

apps/financial-aid/backend/src/app/modules/staff/staff.controller.ts
Rikkiabb: The reason is because now this is matched with a foreign key both in finding all applications and fetching an application by id. See here: https://github.com/island-is/island.is/pull/5059/files#diff-f3a07a64c5f74919ec24c99d9eeab8d431534a2eed62b62d6b952c66d1378fceR64

It would be fine to throw the error for getting all applications but not with finding an application by id because that is used by Osk as well, what we need todo in the near future is to add another guard or add to the application guard to check if veita members have access to this application depending on the municipality id.

Commented here: https://github.com/island-is/island.is/pull/5059/files#diff-aace4f8af6f227d904d3e9b152ba253fe376bc9b338a7fc02e4cce7f15da28cdR13

apps/financial-aid/web-veita/src/components/StateModal/StateModal.tsx
```diff
@@ -47,51 +44,22 @@ const StateModal = ({
...
 
-  const { applicationFilters, setApplicationFilters } = useContext(
-    ApplicationFiltersContext,
-  )
-
-  const [
-    updateApplicationMutation,
-    { loading: saveLoading },
-  ] = useMutation<SaveData>(UpdateApplicationMutation)
+  const changeApplicationState = useApplicationState()
```
MargretFinnboga, line 47: This is very nice 👍

apps/financial-aid/web-veita/src/components/TableBody/TableBody.tsx
```diff
@@ -21,13 +21,20 @@ import {
   calcDifferenceInDate,
   getTagByState,
 } from '@island.is/financial-aid-web/veita/src/utils/formHelper'
+import { useApplicationState } from '../../utils/useApplicationState'
```
MargretFinnboga, line 24: @island.is/financial-aid-web/veita/src/utils/useApplicationState, maybe this instead?

MargretFinnboga (approved):
🪄

PR #5061, opened 2021-09-20
feat(license-service): refactor smartsolution client; add auth retrie…

osk:
…s (#5050)

# ...

## What

## Why

## Screenshots / Gifs

## Checklist:


dabbeg (changes requested):
Something is not working correctly, I haven't found the exact problem but though the retries are exceeded it does not refresh the token.

https://app.datadoghq.eu/logs?query=kube_namespace%3Afeature-tempload-testing-exercise

libs/api/domains/license-service/src/lib/client/driving-license-client/pkpass.client.ts
```diff
@@ -0,0 +1,445 @@
...
+
+      if (res?.status !== 401) {
+        return res
+      }
+
+      count = count + 1
+      this.logger.info(`401 from service, retry number ${count}`, {
+        category: LOG_CATEGORY,
+      })
+    } while (count < actualRetries)
```
dabbeg, line 158: I suspect it is because `actualRetries` is by default 1. Then this do while loop will only run once.
Because the loop runs only once, it will always use the token from the cache and never refresh it when it gets 401.

I am trying out changing the environment variable that controls actualRetries to more than 1. Then it should refresh the token when it is invalid.

dabbeg: I works when I set `PKPASS_AUTH_RETRIES=3`

maranomynet (approved):

PR #5067, opened 2021-09-21
feat(metrics): add metrics module; add metrics to license-service

osk:
Adds a metrics module that allows API to collect metrics via StatsD.

WIP with some open questions:

1. Should we pass the API through [`hot-shots`](https://github.com/brightcove/hot-shots) or wrap individually?
2. How does this fit in with datadog, and are we exposing enough config for it to work as is?
3. How should v1 of "how we use metrics" look like? I.e. naming of counters/tracked things, how we differentiate between success/failure.
4. How do we test this locally?

## Checklist:

### DIFF
--- a/SUMMARY.md
+++ b/SUMMARY.md
@@ -197,6 +197,7 @@
 - [Localization](libs/localization/README.md)
 - [Nest Modules](libs/nest/README.md)
   - [Audit Module](libs/nest/audit/README.md)
+  - [Metrics Module](libs/nest/metrics/README.md)
 - [NOVA SMS](libs/nova-sms/README.md)
 - [Plausible](libs/plausible/README.md)
 - [Regulations](libs/regulations/README.md)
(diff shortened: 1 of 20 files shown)


eirikurn:
1. Whenever possible we should wrap external service interfaces with the stuff we need and shared logic.
2. I think we should expose the minimal amount of config needed to integrate with datadog ... until we need more. E.g. so we only need to configure one or two environment variables?
3. This reminds me a bit about the audit service. Makes sense to document a naming convention, eg `islandis.{serviceName}.{componentSegments}[.{result}]`. But it can be tricky to know "what to count". It might make sense to implement an interceptor and decorator, at least for devops counters. For business counters it should be one of the "definition of done" for each project, and defined in collaboration with Digital Iceland.
4. Same as audit. Have some kind of local mode which logs counters without connecting to statsd. To avoid noise it might make sense to have an optional DEBUG_METRICS variable that turns on logging. Same for the audit module really.

osk:
Old and outdated

PR #5079, opened 2021-09-22
feat(driving-license): info screen with driving instructor select

stjanilofts:
Add step for entering email and choosing driving instructor.

## Screenshots / Gifs

image

## Checklist:


arnorhs (changes requested):
We can't merge it like this, since this application is in production - we have to make it so it still works normally without this. Hide behind feature flag etc

PR #5081, opened 2021-09-22
fix(parental-leave): Request rights were missing proper initialization

saevarma:
## What

Initialize requestDays with existing answer.
Add error to the field if either radio option is not selected.

## Why

If applicant left the application and returned later. Navigated to previous steps and back again through RequestRights, without changing it value, it resulted in the error state `requestDays: 0, isRequesting: "yes"`.

If the applicant was going through new application and no value is selected in RequestRights and he clicked "Áfram" nothing happened and no error was shown. Now the forms highlights the fields with error message.

## Screenshots / Gifs

image

## Checklist:

### DIFF
--- a/libs/application/templates/parental-leave/src/fields/Rights/RequestRightsRadio.tsx
+++ b/libs/application/templates/parental-leave/src/fields/Rights/RequestRightsRadio.tsx
@@ -9,14 +9,12 @@ import {
   RadioField,
 } from '@island.is/application/core'
 import { RadioFormField } from '@island.is/application/ui-fields'
-import {
-  NO,
-  YES,
-  YesOrNo,
-  getApplicationAnswers,
-} from '@island.is/application/templates/parental-leave'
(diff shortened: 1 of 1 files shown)


libs/application/templates/parental-leave/src/fields/Rights/RequestRightsRadio.tsx
```diff
@@ -5,14 +5,18 @@ import {
...
   RadioField,
 } from '@island.is/application/core'
 import { RadioFormField } from '@island.is/application/ui-fields'
-
+import {
+  NO,
+  YES,
+  YesOrNo,
+  getApplicationAnswers,
+} from '@island.is/application/templates/parental-leave'
```
eirikurn, line 17: It's not recommended to import from self like this, as it can cause circular dependency errors.

eirikurn (approved):

PR #5088, opened 2021-09-23
feat(judicial-system): Add court location field

oddsson:
# Add court location field

https://app.asana.com/0/1199153462262248/1200978846073179/f

## What

Add court location field

## Why

Traditionally, court users specify where the court session took place.

## Screenshots / Gifs

Screen Shot 2021-09-23 at 13 48 40

Screen Shot 2021-09-23 at 13 50 21

## Checklist:


apps/judicial-system/backend/src/app/formatters/rulingPdf.ts
```diff
@@ -64,7 +65,9 @@ function constructRestrictionRulingPdf(
     .text(
       `Þann ${formatDate(existingCase.courtStartDate, 'PPP')} heldur ${
         existingCase.judge?.name ?? '?'
-      } ${existingCase.judge?.title ?? '?'} dómþing. Fyrir er tekið mál nr. ${
+      } ${existingCase.judge?.title ?? '?'} dómþing ${lowercase(
+        existingCase.courtLocation?.replace('.', ''),
```
gudjong, line 69: What is the intent with replace here? Are you trying to remove a period at the end? I think this will remove the first period.

gudjong (changes requested):
Consider addressing the single comment.

apps/judicial-system/backend/src/app/formatters/rulingPdf.ts
oddsson: Yep, you are right, this is not the correct approach. 

I've now changed this so the functionality is as follows

If the users inputs "í Héraðsdómi Reykjavíkur", we display "í Héraðsdómi Reykjavíkur."
If the users inputs "í Héraðsdómi Reykjavíkur**.**", we display "í Héraðsdómi Reykjavíkur."
If the users inputs "í Héraðsdómi Reykjavíkur**..**", we display "í Héraðsdómi Reykjavíkur.."

That is we display exactly as many periods as the user inputs and we don't touch other periods in the string, only the ones at the end. This is an acceptable solution to Gudny.

gudjong (approved):

PR #5095, opened 2021-09-23
fix(Financial aid): Layout and logo fix

MargretFinnboga:
# Refactoring layout and fixing logo

https://app.asana.com/0/1201017282841966/1200989775580144

## What

- Moved adminlayout to _app
- Moved where logosvg is getting logo from so it will always get logo
- Made Login a component
- removed index from pages

## Why

- Better usage , render and non flickering logo 
- So logo will always appear and not dependent on route
- If admin not logged in login component will render
- it was useless 

## Checklist:


Rikkiabb (approved):
💯

PR #5097, opened 2021-09-24
feat(application-system-api): Pass auth to template-api-modules

stefanhar:
# Pass auth to template-api-modules

Currently we only pass the authentication token as string to template api endpoints.
Our tooling expects authentication objects to be passed, this makes it difficult to work with auth inside template api modules

## What

- Changed `authorization` to `auth`
- Changed types to reflect change
- Updated endpoints to reflect the change

## Why

- Makes it easier to work with auth in application endpoints
- Variable name of `auth` is more fitting and makes more sense e.g. when accessing properties `auth.authorization` vs `authorization.authorization`

## Checklist:


Thorrik58 (approved):

kksteini (approved):
LGTM

jeremybarbet (approved):

PR #5107, opened 2021-09-27
fix(cd): Retag helm config latest

sindrig:
Supports island-is/infrastructure#868


sindrig:
Supports island-is/infrastructure#868

pshomov (approved):
LGTM

PR #5114, opened 2021-09-27
fix(cd): Preserve auth cookie for e2e tests

sindrig:
Also upgrade cypress.

This closes island-is/infrastructure#856 (although it was never a problem apparently)


apps/system-e2e/src/support/index.ts
```diff
@@ -14,5 +14,10 @@
 // ***********************************************************
 
 // Import commands.js using ES2015 syntax:
+Cypress.Cookies.debug(true)
```
darrikonn, line 17: Do we wanna keep the Cookies.debug in?
Wasn't it just beneficial while working on this PR (and will become noise to others)?

darrikonn (approved):

apps/system-e2e/src/support/index.ts
sindrig: It's really low volume, and might become beneficial when and if we need to debug failed logins.

PR #5117, opened 2021-09-27
fix(ops): Use internal endpoints for rds

sindrig:
As defined in island-is/infrastructure#872

Do not merge until island-is/infrastructure#872 is merged and applied


pshomov (approved):
LGTM

PR #5118, opened 2021-09-27
chore(judicial-system): Court record accordion fixes

oddsson:
# Court record accordion fixes

https://app.asana.com/0/1199153462262248/1201014062889027/f

## What

See Asana task ☝️

## Why

Better UX

## Checklist:


libs/judicial-system/formatters/src/lib/formatters.ts
```diff
@@ -304,7 +304,7 @@ export const areAccusedRightsHidden = (
   isAccusedRightsHidden?: boolean,
   sessionArrangements?: SessionArrangements,
 ): boolean => {
-  return isAccusedRightsHidden
+  return isAccusedRightsHidden !== undefined
```
gudjong, line 307: isAccuesedRightsHidden is never undefined.

gudjong (changes requested):

gudjong (approved):

PR #5120, opened 2021-09-27
feat(financial-aid): email confirmation for application

MargretFinnboga:
# Email confirmation for application

https://app.asana.com/0/1201017282841966/1201017282842009

## What

When received application, send email to user. Using email service. 

## Checklist:


apps/financial-aid/backend/src/app/modules/application/application.service.ts
```diff
@@ -124,6 +133,15 @@ export class ApplicationService {
       await Promise.all(promises)
     }
 
+    await this.sendEmail(
+      {
+        name: user.name,
+        address: application.email,
+      },
+      appModel.id,
```
Rikkiabb, line 141: Is there any reason why we are sometimes using the appModel and sometimes the application?

apps/financial-aid/backend/src/environments/environment.ts
```diff
@@ -29,6 +29,18 @@ const prodConfig = {
     postTimeToLiveMinutes: 5,
     getTimeToLiveMinutes: 5,
   },
+  email: {
+    fromEmail: process.env.EMAIL_FROM ?? '',
+    fromName: process.env.EMAIL_FROM_NAME ?? '',
+    replyToEmail: process.env.EMAIL_REPLY_TO ?? '',
+    replyToName: process.env.EMAIL_REPLY_TO_NAME ?? '',
```
Rikkiabb, line 36: What will be the values here on different envs?

apps/financial-aid/backend/src/app/modules/application/application.service.ts
```diff
@@ -124,6 +133,15 @@ export class ApplicationService {
...
+      `Umsókn þín er móttekin og er nú í vinnslu.`,
```
Rikkiabb, line 142: Maybe not a part of this PR, but I think that it would be super nice if the next step in emails is adding a link to status page.

apps/financial-aid/backend/src/environments/environment.ts
Rikkiabb: I am also wondering if this should be in environment. Because the end product will have different values depending on municipalities which we will read from a table (probably) so maybe this should just be hardcoded until we get there. Any thoughts?

Rikkiabb (commented):
Some questions ❔

apps/financial-aid/backend/src/environments/environment.ts
MargretFinnboga: Ahh good point! I didn't realize, we have different email for municipality well spotted, I will hardcoded until we get there 🤖

apps/financial-aid/backend/src/app/modules/application/application.service.ts
```diff
@@ -136,10 +136,10 @@ export class ApplicationService {
     await this.sendEmail(
       {
         name: user.name,
-        address: application.email,
+        address: appModel.email,
       },
       appModel.id,
-      `Umsókn þín er móttekin og er nú í vinnslu.`,
+      `Umsókn þín er móttekin og er nú í vinnslu. <a href="https://fjarhagsadstod.dev.sveitarfelog.net/stada/${appModel.id}" target="_blank"> Getur kíkt á stöðu síðuna þína hér</a>`,
```
Rikkiabb, line 142: The base url here should be a env variable in the future, not needed right now just an fyi

apps/financial-aid/backend/src/app/modules/application/application.service.ts
```diff
@@ -181,12 +181,12 @@ export class ApplicationService {
     try {
       await this.emailService.sendEmail({
         from: {
-          name: environment.email.fromName,
-          address: environment.email.fromEmail,
+          name: 'rikki@kolibri.is',
+          address: 'Rikka',
```
Rikkiabb, line 185: Like the idea but lets have this no-reply@svg.is 😄

apps/financial-aid/backend/src/app/modules/application/application.service.ts
```diff
@@ -181,12 +181,12 @@ export class ApplicationService {
...
         },
         replyTo: {
-          name: environment.email.replyToName,
-          address: environment.email.replyToEmail,
+          name: 'rikki@kolibri.is',
+          address: 'Rikka',
```
Rikkiabb, line 189: Same here 😄

apps/financial-aid/backend/src/environments/environment.ts
```diff
@@ -30,10 +30,10 @@ const prodConfig = {
...
   },
   email: {
-    fromEmail: process.env.EMAIL_FROM ?? '',
-    fromName: process.env.EMAIL_FROM_NAME ?? '',
-    replyToEmail: process.env.EMAIL_REPLY_TO ?? '',
-    replyToName: process.env.EMAIL_REPLY_TO_NAME ?? '',
+    fromEmail: 'rikki@kolibri.is',
+    fromName: 'Rikka',
+    replyToEmail: 'rikki@kolibri.is',
+    replyToName: 'Rikka',
```
Rikkiabb, line 36: And we should remove this since we are not using env

Rikkiabb (approved):
🥅

PR #5133, opened 2021-09-28
fix(Financial aid): Gets application when state is changed

MargretFinnboga:
# Gets application when state is changed

https://app.asana.com/0/1201017282841966/1201062859132695

## What

Added better error message to index file 
Added setFilesToApplication function to backend application.server 
When state is changed gets the whole application and reloads the application state

## Why

We want the database to be the only source of truth 

## Checklist:


apps/financial-aid/web-veita/src/components/StateModal/StateModal.tsx
```diff
@@ -52,9 +46,17 @@ const StateModal = ({
...
+    const updatedApplication = await changeApplicationState(
+      application,
+      state,
+      amount,
+      rejection,
+    )
+    if (updatedApplication) {
+      setApplication(updatedApplication)
+    }
+
```
Rikkiabb, line 58: Are we doing this in many places? Just wondering if it makes sense to extract this or not

apps/financial-aid/web-veita/src/utils/useApplicationState.ts
```diff
@@ -43,14 +43,18 @@ export const useApplicationState = () => {
...
-        [state]: applicationFilters[state] + 1,
-      }))
+      if (data) {
+        if (applicationFilters && setApplicationFilters) {
+          setApplicationFilters((preState) => ({
+            ...preState,
+            [prevState]: applicationFilters[prevState] - 1,
+            [state]: applicationFilters[state] + 1,
+          }))
+        }
```
Rikkiabb, line 54: This will be removed in the forseeable future right?

Rikkiabb (approved):
Awesome have couple of points if you want to check them out.

PR #5137, opened 2021-09-29
fix(service-portal): no finance transactions

thordurhhh:
## What

* Fix when to show the empty state for finances.
* Minor additions to the filter.

## Why

The check was wrong.

## Checklist:


libs/service-portal/finance/src/screens/FinanceStatus/FinanceStatus.tsx
```diff
@@ -114,33 +115,7 @@ const FinanceStatus: ServicePortalModuleComponent = () => {
...
-                        { value: '' },
-                        { value: getChargeTypeTotal(), align: 'right' },
-                      ]}
-                    />
-                  </T.Body>
-                </T.Table>
-              </Box>
-            )}
-          {financeStatusData?.organizations?.length > 0 ? (
+          {financeStatusData?.organizations?.length > 0 || financeStatusZero ? (
```
osk, line 118: should pull this check out and resuse here and on line 196?
https://github.com/island-is/island.is/pull/5137/files#diff-822997789905148344f1668b55b36d844e1ba0830f38297858835b3a72415d22R196

osk (approved):

libs/service-portal/finance/src/screens/FinanceStatus/FinanceStatus.tsx
thordurhhh: I believe it will not be needed there.

Because the _Total Row_ is outside of that check and it will show a row containing:

Samtals - 0kr. with the help of `getChargeTypeTotal` if there are no organizations present.

PR #5142, opened 2021-09-29
feat(api-domains-national-registry-x-road): buseta check

arnorhs:
A feature for extracting a person's country of residency according to thjodskra - will be used to evaluate whether or not they have spent more than 185 days as a resident in the country for the driver's license application.

## Checklist:

### DIFF
--- a/libs/api/domains/national-registry-x-road/src/lib/nationalRegistryXRoad.resolver.ts
+++ b/libs/api/domains/national-registry-x-road/src/lib/nationalRegistryXRoad.resolver.ts
@@ -11,6 +11,7 @@ import { Audit } from '@island.is/nest/audit'
 
 import { NationalRegistryPerson } from '../models/nationalRegistryPerson.model'
 import { NationalRegistryXRoadService } from './nationalRegistryXRoad.service'
+import { NationalRegistryResidence } from '../models/nationalRegistryResidence.model'
 
 @UseGuards(IdsAuthGuard, IdsUserGuard, ScopesGuard)
 @Resolver(() => NationalRegistryPerson)
@@ -40,6 +41,18 @@ export class NationalRegistryXRoadResolver {
     @Parent() person: NationalRegistryPerson,
   ): Promise<NationalRegistryPerson[] | undefined> {
     return await this.nationalRegistryXRoadService.getChildrenCustodyInformation(
+      person.nationalId,
+      user.authorization,
+    )
+  }
+
+  @ResolveField('residenceHistory', () => [NationalRegistryResidence])
+  @Audit()
+  async resolveResidenceHistory(
+    @Context('req') { user }: { user: User },
+    @Parent() person: NationalRegistryPerson,
+  ): Promise<NationalRegistryResidence[] | undefined> {
+    return await this.nationalRegistryXRoadService.getNationalRegistryResidenceHistory(
       person.nationalId,
       user.authorization,
     )
--- a/libs/api/domains/national-registry-x-road/src/lib/nationalRegistryXRoad.service.ts
+++ b/libs/api/domains/national-registry-x-road/src/lib/nationalRegistryXRoad.service.ts
@@ -3,11 +3,14 @@ import { ApolloError } from 'apollo-server-express'
 import {
   Einstaklingsupplysingar,
   Fjolskylda,
+  Heimili,
 } from '@island.is/clients/national-registry-v2'
 import type { NationalRegistryXRoadConfig } from './nationalRegistryXRoad.module'
 import { NationalRegistryPerson } from '../models/nationalRegistryPerson.model'
 import type { Logger } from '@island.is/logging'
 import { LOGGER_PROVIDER } from '@island.is/logging'
+import { NationalRegistryResidence } from '../models/nationalRegistryResidence.model'
+import { NationalRegistryAddress } from '../models/nationalRegistryAddress.model'
 
 @Injectable()
 export class NationalRegistryXRoadService {
@@ -44,6 +47,36 @@ export class NationalRegistryXRoadService {
     } catch (error) {
       throw this.handleError(error)
     }
+  }
+
+  async getNationalRegistryResidenceHistory(
+    nationalId: string,
+    authToken: string,
+  ): Promise<NationalRegistryResidence[]> {
+    const historyList = await this.nationalRegistryFetch<Array<Heimili>>(
+      `/${nationalId}/buseta`,
+      authToken,
+    )
+
+    const history = historyList.map((heimili: Heimili) => {
+      if (!heimili.breytt) {
+        throw new Error('All history entries have a modified date')
+      }
+
+      const date = new Date((heimili.breytt as unknown) as string)
+
+      return {
+        address: {
+          city: heimili.stadur,
+          postalCode: heimili.postnumer,
+          streetName: heimili.heimilisfang,
+        } as NationalRegistryAddress,
+        country: heimili.landakodi,
+        dateOfChange: date,
+      } as NationalRegistryResidence
+    })
+
+    return history
   }
 
   async getNationalRegistryPerson(
--- a/libs/api/domains/national-registry-x-road/src/models/nationalRegistryResidence.model.ts
+++ b/libs/api/domains/national-registry-x-road/src/models/nationalRegistryResidence.model.ts
@@ -1,0 +1,14 @@
+import { Field, ObjectType, ID } from '@nestjs/graphql'
+import { NationalRegistryAddress } from './nationalRegistryAddress.model'
+
+@ObjectType()
+export class NationalRegistryResidence {
+  @Field(() => NationalRegistryAddress)
+  address!: NationalRegistryAddress
+
+  @Field()
+  country!: string
+
+  @Field(() => Date)
+  dateOfChange!: Date
+}
--- a/libs/api/domains/national-registry-x-road/src/models/nationalRegistryAddress.model.ts
+++ b/libs/api/domains/national-registry-x-road/src/models/nationalRegistryAddress.model.ts
@@ -5,7 +5,7 @@ export class NationalRegistryAddress {
   @Field(() => String)
   streetName?: string
 
-  @Field(() => String)
+  @Field(() => String, { nullable: true })
   postalCode?: string
 
   @Field(() => String)
--- a/libs/api/domains/national-registry-x-road/src/models/nationalRegistryPerson.model.ts
+++ b/libs/api/domains/national-registry-x-road/src/models/nationalRegistryPerson.model.ts
@@ -1,5 +1,6 @@
 import { Field, ObjectType, ID } from '@nestjs/graphql'
 import { NationalRegistryAddress } from './nationalRegistryAddress.model'
+import { NationalRegistryResidence } from './nationalRegistryResidence.model'
 
 @ObjectType()
 export class NationalRegistryPerson {
@@ -23,4 +24,7 @@ export class NationalRegistryPerson {
 
   @Field(() => NationalRegistryPerson, { nullable: true })
   otherParent?: NationalRegistryPerson
+
+  @Field(() => [NationalRegistryResidence], { nullable: true })
+  residenceHistory?: NationalRegistryResidence[]
 }


libs/api/domains/national-registry-x-road/src/lib/computeCountryResidence.spec.ts
```diff
@@ -0,0 +1,99 @@
+import { computeCountryResidence } from './computeCountryResidence'
+
+jest
+  .useFakeTimers('modern')
+  .setSystemTime(new Date('2021-09-18T00:00:00').getTime())
+
+const d = new Date().getTime()
+
+// helper to return a date object with subtracted days from "today"
+const makeDate = (addDays: number): Date => new Date(d + addDays * 86400 * 1000)
```
Rikkiabb, line 10: Nitpick, but would have the constants 86400 and 1000 as varibles for readability

Rikkiabb: I think you could also do it like this: 

```
const d = new Date().getDate()

const makeDate = (addDays: number): Date => new Date(d + addDays)
```

libs/api/domains/national-registry-x-road/src/lib/computeCountryResidence.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  const currentDateTime = new Date()
+
+  const now = new Date(
+    currentDateTime.getFullYear(),
+    currentDateTime.getMonth(),
+    currentDateTime.getDate(),
+  ).getTime()
+
+  const yearFromNow = new Date(
```
Rikkiabb, line 36: This should be named last year, right?

libs/api/domains/national-registry-x-road/src/lib/computeCountryResidence.ts
```diff
@@ -0,0 +1,56 @@
...
+  let current = Number.MAX_SAFE_INTEGER
+  const timeByCountry: Record<string, number> = {}
+  while (current > yearFromNow && simplified[i]) {
+    const { time, country } = simplified[i]
+    current = Math.max(time, yearFromNow)
+    const period = Math.round((lastTime - current) / (86400 * 1000))
+    timeByCountry[country] = (timeByCountry[country] || 0) + period
+    lastTime = current
+    i++
+  }
```
Rikkiabb, line 53: Could this help instead of doing it manually:
https://date-fns.org/v2.24.0/docs/differenceInDays ? 

Is there any reason with working with milliseconds instead of days?

libs/api/domains/national-registry-x-road/src/models/nationalRegistryResidence.model.ts
```diff
@@ -0,0 +1,14 @@
...
+@ObjectType()
+export class NationalRegistryResidence {
+  @Field(() => NationalRegistryAddress)
+  address!: NationalRegistryAddress
+
+  @Field()
+  country!: string
+
+  @Field(() => Date, { nullable: true })
+  dateOfChange!: Date
```
Rikkiabb, line 13: Since this is nullable, shouldn't this be: 

```
 @Field(() => Date, { nullable: true })
   dateOfChange!: Date
```

libs/api/domains/national-registry-x-road/src/lib/nationalRegistryXRoad.service.ts
```diff
@@ -46,6 +51,40 @@ export class NationalRegistryXRoadService {
...
+          city: heimili.stadur,
+          postalCode: heimili.postnumer,
+          streetName: heimili.heimilisfang,
+        } as NationalRegistryAddress,
+        country: heimili.landakodi,
+        dateOfChange: date,
+      } as NationalRegistryResidence
+    })
+
+    computeCountryResidence(history)
```
Rikkiabb, line 80: This function returns the history, but you are not using the outcome from it, why is that?

libs/api/domains/national-registry-x-road/src/lib/nationalRegistryXRoad.service.ts
```diff
@@ -46,6 +51,40 @@ export class NationalRegistryXRoadService {
...
+      `/${nationalId}/buseta`,
+      authToken,
+    )
+
+    const history = historyList.map((heimili: Heimili) => {
+      // API says Date, but is string -- fallback in case that changes in the future
+      const date =
+        typeof heimili.breytt === 'string'
+          ? new Date(heimili.breytt)
+          : heimili.breytt
```
Rikkiabb, line 68: Won't this always be sent as a string, since the api stringifies the values before sending them? 

If so then the typeof check is not needed.

Rikkiabb (changes requested):
Some questions 😄

libs/api/domains/national-registry-x-road/src/lib/computeCountryResidence.spec.ts
arnorhs: I think that wouldn't work, but I might be misunderstanding :)

But I'll add constants, thanks

libs/api/domains/national-registry-x-road/src/lib/nationalRegistryXRoad.service.ts
arnorhs: I might be doing something wrong, but as far as I could tell, this is a bug or at the very least a very weird feature.

The type is registered as a Date according to the openapi => typescript client generator. however, the data is indeed a string, so, eg. calling anything such as .getTime() on it will always fail, even though the typescript checker does not complain.

I didn't dig deep enough to figure out whether the bug is in the generator, in the deserializer or what, but this is what I was observing

libs/api/domains/national-registry-x-road/src/lib/computeCountryResidence.ts
arnorhs: milliseconds is simple - re: date-fns i'm not sure you'd be gaining a whole lot, you'd still need to diff the time period etc.. however it might be simpler to reason about.

libs/api/domains/national-registry-x-road/src/lib/utils/computeCountryResidence.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  const currentDateTime = new Date()
+
+  const now = new Date(
+    currentDateTime.getFullYear(),
+    currentDateTime.getMonth(),
+    currentDateTime.getDate(),
+  ).getTime()
+
+  const yearFromNow = new Date(
```
Rikkiabb, line 37: isn't this last year?

libs/api/domains/national-registry-x-road/src/lib/utils/computeCountryResidence.ts
```diff
@@ -0,0 +1,59 @@
...
+      return b - a
+    })
+
+  const currentDateTime = new Date()
+
+  const now = new Date(
+    currentDateTime.getFullYear(),
+    currentDateTime.getMonth(),
+    currentDateTime.getDate(),
+  ).getTime()
```
Rikkiabb, line 35: A general question I am wondering about, is there any reason why you don't use days instead of ms?

libs/api/domains/national-registry-x-road/src/lib/nationalRegistryXRoad.service.ts
```diff
@@ -46,6 +51,47 @@ export class NationalRegistryXRoadService {
...
+      } as NationalRegistryResidence
+    })
+
+    const daysPerCountry = computeCountryResidence(history)
+    const countryResidenceDays = daysPerCountry
+      ? Object.entries(daysPerCountry).map(([countryCode, days]) => ({
+          countryCode,
+          days,
+        }))
+      : null
```
Rikkiabb, line 86: What is happening here?

libs/api/domains/national-registry-x-road/src/lib/utils/computeCountryResidence.ts
```diff
@@ -0,0 +1,59 @@
...
+    currentDateTime.getFullYear() - 1,
+    currentDateTime.getMonth(),
+    currentDateTime.getDate(),
+  ).getTime()
+
+  let lastTime = now
+  let i = 0
+  let current = Number.MAX_SAFE_INTEGER
+
+  const timeByCountry: Record<string, number> = {}
```
Rikkiabb, line 47: Should this be `List<{country: string, time: number}>`

libs/api/domains/national-registry-x-road/src/lib/computeCountryResidence.ts
arnorhs: yes!

libs/api/domains/national-registry-x-road/src/lib/nationalRegistryXRoad.service.ts
arnorhs: number? if that would be the case, then we have bigger problems. in that case, as the code is written, the value of it would be a number when calling getTime() etc on it, which would result in an error/exception, and that would be discoverable.

however, if you put a number into the date constructor, you will get a date.. but since we don't really know what the format of that number would be (Date-compatible? unix timestamp w/o milliseconds? something else?) .. it would mean we'd just introduce a bug/wrong information, resulting in bad data that probably wouldn't be discovered until it is reported by a user

libs/api/domains/national-registry-x-road/src/models/nationalRegistryResidenceHistory.model.ts
```diff
@@ -0,0 +1,15 @@
...
+@ObjectType()
+export class NationalRegistryResidenceHistory {
+  @Field(() => ID)
+  nationalId!: string
+
+  @Field(() => [NationalRegistryResidence])
+  history!: NationalRegistryResidence[]
+
+  @Field(() => [NationalRegistryCountryResidenceDays], { nullable: true })
+  countryResidenceDays!: NationalRegistryCountryResidenceDays[] | null
```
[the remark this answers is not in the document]
arnorhs: yes, fixed

libs/api/domains/national-registry-x-road/src/lib/nationalRegistryXRoad.service.ts
arnorhs: the calculation functions returns a Record<string, number> and I don't really know how to express that as a type in graphql, so i'm converting it to more normal `NationalRegistryCountryResidenceDays` type.

this function converts it between those two formats

libs/api/domains/national-registry-x-road/src/lib/utils/computeCountryResidence.ts
arnorhs: well.. it's simpler to read/write/understand by using a map instead of an array/list... unless I'm misunderstanding something

libs/api/domains/national-registry-x-road/src/lib/utils/computeCountryResidence.ts
```diff
@@ -0,0 +1,59 @@
...
+import { DAY } from './constants'
+
+/**
+ * Compute a summary of how many days you've spent as a resident of each country in
+ * the last 365 days
+ *
+ * @param history list from national id registry of residences
+ * @returns map of country code to days spent in country in the last 365 days
+ */
+export const computeCountryResidence = (
```
darrikonn, line 11: This function feels overly complex for a shared library and will be hard for other devs to contribute and understand.

This could be simplified a lot with an implementation similar to this:
```typescript
const last354days = xxx

return history.reduce((acc, item) =? {
  const numberOfDaysLivedInCountry = Math.max(0, differenceInDays(last365days, item.dateOfChange))
  return {
    ...acc,
    [item.country]: numberOfDaysLivedInCountry + (acc[item.country] || 0)
  }
}, {})
```

darrikonn (changes requested):

libs/api/domains/national-registry-x-road/src/lib/utils/computeCountryResidence.spec.ts
```diff
@@ -0,0 +1,100 @@
+import { computeCountryResidence } from './computeCountryResidence'
```
darrikonn, line 1: Nice to see tests 🙏

libs/api/domains/national-registry-x-road/src/lib/utils/computeCountryResidence.ts
arnorhs: @darrikonn I agree, that reads a lot better. Nut sure it captures the nuance of each case.. 

i agree about the premise of this being probably too involved to live in this library, in many cases.. I could move it to someplace else

libs/api/domains/national-registry-x-road/src/lib/nationalRegistryXRoad.service.ts
Rikkiabb: I understand, then the `computeCountryResidence` function should return that type. Instead of that returning another type that always need to be transformed.

libs/api/domains/national-registry-x-road/src/lib/utils/computeCountryResidence.ts
Rikkiabb: I think @darrikonn answer catches everything and if it doesn't then your tests should catch it. I would try it out.

libs/api/domains/national-registry-x-road/src/models/nationalRegistryResidence.model.ts
```diff
@@ -0,0 +1,14 @@
...
+}
```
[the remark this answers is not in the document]
arnorhs: i don't want this field to be nullable

libs/api/domains/national-registry-x-road/src/lib/nationalRegistryXRoad.resolver.ts
```diff
@@ -44,4 +45,17 @@ export class NationalRegistryXRoadResolver {
       user.authorization,
     )
   }
+
+  @Query(() => NationalRegistryResidenceHistory, {
```
darrikonn, line 49: Would like to see this field as a field resolver. This means that we could query this field on the nationalRegistryPerson field:
```graphql
query {
  nationalRegistryPerson {
    id
    residence {
	  address {
	    postalCode
	  }
	  country
	  history {
	    address {
	      postalCode
	    }
	    country
	  }
    }
  }
}
```

where `history` of the residence field would be resolved in a field resolver. Leads to a better structured object.

arnorhs: great idea

libs/api/domains/national-registry-x-road/src/lib/nationalRegistryXRoad.resolver.ts
```diff
@@ -44,4 +45,16 @@ export class NationalRegistryXRoadResolver {
       user.authorization,
     )
   }
+
+  @ResolveField('history', () => [NationalRegistryResidence])
```
darrikonn, line 49: I believe you have a mismatch here... nestjs/graphql should flag this.
The field is called `residenceHistory`, not `history`

darrikonn (changes requested):

Rikkiabb (approved):
LGTM :shipit:

darrikonn (approved):
Nice

PR #5147, opened 2021-09-29
chore(core): Adding type to testServer override

### LINKED ISSUE #5146: Fix testServer to stop allowing double compilation of the app
Currently few test specs are calling compile on the builder when overriding, causing the testServer to compile the application two times.

saevarma:
## What

Adding type to override in testServer to stop test specs from being able to call `.compile()` and instead need to return the builder in override.

## Why

To stop specs implementation from double compiling the app.
Fixes #5146 

## Screenshots / Gifs

## Checklist:

### DIFF
--- a/libs/infra-nest-server/src/lib/testServer.ts
+++ b/libs/infra-nest-server/src/lib/testServer.ts
@@ -5,7 +5,7 @@ import {
 } from '@island.is/auth-nest-tools'
 import { Type, ValidationPipe } from '@nestjs/common'
 import { InfraModule } from './infra/infra.module'
-import { Test } from '@nestjs/testing'
+import { Test, TestingModule } from '@nestjs/testing'
 import { TestingModuleBuilder } from '@nestjs/testing/testing-module.builder'
 
 export type TestServerOptions = {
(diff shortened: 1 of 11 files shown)


libs/infra-nest-server/src/lib/testServer.ts
```diff
@@ -18,15 +18,15 @@ export type TestServerOptions = {
...
 }
 
 export const testServer = async (options: TestServerOptions) => {
-  const builder = Test.createTestingModule({
+  let builder = Test.createTestingModule({
     imports: [InfraModule.forRoot(options.appModule)],
   })
   if (options.override) {
-    options.override(builder)
+    builder = options.override(builder)
```
darrikonn, line 29: good spot, how did the other one work 🤔

darrikonn (approved):

libs/infra-nest-server/src/lib/testServer.ts
saevarma: I guess it works because it is passed the instance of the builder so modifies it in-place instead of overriding.

So this assignment is probably unnecessary, but is just logical to have it as the override method now returns a `TestingModuleBuilder` so technically the specs could create a completely new builder instance and return it.

obmagnusson (approved):

baering (approved):

Valur (approved):

PR #5148, opened 2021-09-29
fix(financial-aid): Moving filtering of applications from client to backend

MargretFinnboga:
# Moving filtering of applications from client to backend

https://app.asana.com/0/1201017282841966/1201079906615276

## What

Added a enum ApplicationStateUrl 
When getting all applications, takes input that is ApplicationStateUrl 
Filters in backend instead of client 
When routing between tables, we get the newest

## Why

Put the filter rather on the backend than in client

## Checklist:

### DIFF
--- a/apps/financial-aid/api/src/app/modules/application/dto/applications.input.ts
+++ b/apps/financial-aid/api/src/app/modules/application/dto/applications.input.ts
@@ -1,0 +1,11 @@
+import { Allow } from 'class-validator'
+
+import { Field, InputType } from '@nestjs/graphql'
+import { ApplicationStateUrl } from '@island.is/financial-aid/shared/lib'
+
+@InputType()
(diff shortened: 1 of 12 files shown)


apps/financial-aid/api/src/app/modules/application/dto/applications.input.ts
```diff
@@ -0,0 +1,11 @@
+import { Allow } from 'class-validator'
+
+import { Field, InputType } from '@nestjs/graphql'
+import { ApplicationStateUrl } from '@island.is/financial-aid/shared/lib'
+
+@InputType()
+export class ApplicationsInput {
```
Rikkiabb, line 7: Should we change this to AllApplicationInput, so we don't mix it with ApplicationInput

Rikkiabb (approved):
🥇

PR #5152, opened 2021-09-30
chore(judicial-system): Cypress test data refactoring

oddsson:
# Refactor Cypress test data

https://app.asana.com/0/1199153462262248/1201093995861204/f

## What

Before this change our test data in Cypress was stored in JSON files. This was a bad pattern because

- There was no type safety
- If I needed to test something where the case type was CUSTODY and another thing where the case type TRAVEL_BAN, I would need two different JSON files. This quickly became a problem and we had way too many files with JSON in our project.

Now we create a base case that we can change on the fly, which is so much cleaner.

## Why

See ☝️

## Checklist:


gudjong (approved):
Nice

PR #5159, opened 2021-09-30
fix(build): Preventing collision of config keys from different sections

pshomov:
# ...

Secrets and env variables all end up as environment variables and thus share the same name space. Since these are coming from different parts of the code, added a bit of checks to validate there are no collisions.

## What

Avoid weird deployment errors.

## Screenshots / Gifs

## Checklist:

### DIFF
--- a/infra/src/dsl/map-to-values.ts
+++ b/infra/src/dsl/map-to-values.ts
@@ -33,6 +33,13 @@ export const serializeService: SerializeMethod = (
   uberChart: UberChartType,
 ) => {
   let allErrors: string[] = []
+  const mergeObject = (target: any, source: any) => {
+    const targetKeys = Object.keys(target)
+    return (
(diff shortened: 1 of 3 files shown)


infra/src/dsl/map-to-values.ts
```diff
@@ -33,6 +33,21 @@ export const serializeService: SerializeMethod = (
   uberChart: UberChartType,
 ) => {
   let allErrors: string[] = []
+  const checkCollisions = (target: any, source: any) => {
```
sindrig, line 36: No way to create some type better than any here? Isn't the value always either a string or a secret ref?

sindrig (approved):

infra/src/dsl/map-to-values.ts
pshomov: I thought I would leave it open but this is way too wide, I agree. Will fix.

PR #5171, opened 2021-10-04
feat(judicial-system): Appeal can now show correct gender pronouns

oddsson:
# Appeal can now show correct gender pronouns

https://app.asana.com/0/1199153462262248/1201061586055548/f

## What

Add parameter `stakeholderGender` to `formatAppeal` function. If that parameter is set, we use the appropriate gender pronouns. Examples

`formatAppeal("APPEAL", "Kærða", "FEMALE")` returns "Kærða lýsir því yfir að **hún** kæri úrskurðinn til Landsréttar."

## Checklist:


libs/judicial-system/formatters/src/lib/formatters.ts
```diff
@@ -299,3 +300,38 @@ export function formatGender(gender?: CaseGender): string {
...
+    case CaseGender.OTHER:
+    default:
+      return 'hán'
+  }
+}
+
+export function formatAppeal(
+  appealDecision: CaseAppealDecision | undefined,
+  stakeholder: string,
+  stakeholderGender?: CaseGender,
```
gudjong, line 319: Consider simplifying this method, by defaulting to CaseGender.MALE. That way, there is no need to check below if stakeholderGender is set.

gudjong (changes requested):

libs/judicial-system/formatters/src/lib/formatters.ts
oddsson: Ah! Thank you.. Done!

gudjong (approved):

PR #5176, opened 2021-10-04
fix(judicial-system): Fix issues with dismissed cases

oddsson:
# Fix issues with dismissed cases

https://app.asana.com/0/1199153462262248/1201066149495299/f

## What

Fixes two issues

- Dismissed cases had validTo dates in the list of cases, which was not supposed to happen
- Dismissed cases showed up in the active cases list for prosecutors.

## Why

These are bugs.

## Checklist:


apps/judicial-system/web/src/routes/Shared/Requests/Requests.tsx
```diff
@@ -61,7 +61,9 @@ export const Requests: React.FC = () => {
       setActiveCases(
         casesWithoutDeleted.filter((c: Case) => {
           return isProsecutor
-            ? c.state !== CaseState.ACCEPTED && c.state !== CaseState.REJECTED
+            ? c.state !== CaseState.ACCEPTED &&
```
gudjong, line 64: Consider using completedCaseStates from case.ts in the types library.

gudjong (changes requested):

gudjong (approved):

PR #5184, opened 2021-10-04
feat(fa): New login

Rikkiabb:
# ...
https://app.asana.com/0/1201097099198335/1200897217304839

## What
Uses new login system for samband íslenskra Sveitarfélaga:
- Uses `IdsUserGuard` for all controllers. 
- Uses next-auth for new login
- Adds next api routes to handle new login
- Removes `TokenGuard` since `IdsUserGuard` can be used everywhere and it has better security.
- Creates new current user decorator for samband sveitarfélaga endpoints.
- Makes phone number optional since we don't get it from login - adds migration for it
- Makes base api a const since it is used all over the place
- Removes old login
- Removes financial-aid-auth library
- Removes unneeded `user.service` and uses backend instead

## Why
Newer login system, better security and needed to connect to services.

## Screenshots / Gifs

## Checklist:


apps/financial-aid/backend/src/app/decorators/current.user.decorator.ts
```diff
@@ -0,0 +1,18 @@
...
+} from '@nestjs/common'
+
+import { User } from '@island.is/financial-aid/shared/lib'
+import { getUserFromContext } from '../lib'
+
+export const CurrentUser = createParamDecorator(
+  (_: unknown, context: ExecutionContext): User => {
+    const user = getUserFromContext(context)
+    if (!user) {
+      throw new BadRequestException('Invalid user')
```
darrikonn, line 14: Unauthorized?

apps/financial-aid/api/src/app/modules/decorators/current.user.decorator.ts
```diff
@@ -0,0 +1,36 @@
...
+  (_: unknown, context: ExecutionContext): User => {
+    const req = GqlExecutionContext.create(context).getContext().req
+
+    const cookies = req.cookies
+    const sessionToken = cookies
+      ? cookies[IDENTITY_SERVER_SESSION_TOKEN_COOKIE_NAME]
+      : null
+
+    if (!sessionToken) {
+      throw new BadRequestException('Invalid user')
```
darrikonn, line 24: Ditto on unauthorized?

apps/financial-aid/backend/src/app/modules/municipality/municipality.controller.ts
```diff
@@ -39,19 +35,3 @@ export class MunicipalityController {
     return municipality
   }
 }
-
-// @UseGuards(JwtAuthGuard)
```
darrikonn, line 43: remove?

apps/financial-aid/api/src/app/modules/application/application.resolver.ts
```diff
@@ -18,19 +15,16 @@ import {
...
-
 import {
   Application,
   ApplicationFilters,
   UpdateApplicationTableResponseType,
   UpdateApplicationResponseType,
 } from '@island.is/financial-aid/shared/lib'
 
-@UseGuards(JwtGraphQlAuthGuard)
 @Resolver(() => ApplicationModel)
```
darrikonn, line 28: Should this resolver not be guarded?

apps/financial-aid/api/src/app/modules/decorators/current.user.decorator.ts
```diff
@@ -0,0 +1,36 @@
+import {
```
darrikonn, line 1: Btw. the filename `current.user.decorator` vs. `currentUser.decorator` :)

apps/financial-aid/api/src/app/modules/file/file.resolver.ts
```diff
@@ -13,7 +12,6 @@ import {
 } from './dto'
 import { SignedUrlModel, CreateFilesModel } from './models'
 
-@UseGuards(JwtGraphQlAuthGuard)
 @Resolver()
```
darrikonn, line 15: Should the graphql api be public?

apps/financial-aid/api/src/app/modules/user/user.resolver.ts
```diff
@@ -1,79 +1,58 @@
...
+      }
+      throw e
+    }
   }
 
   @Query(() => UserModel, { nullable: true })
-  async currentUser(
-    @CurrentGraphQlUser() user: User,
-  ): Promise<UserModel | undefined> {
+  async currentUser(@CurrentUser() user: User): Promise<UserModel | undefined> {
```
darrikonn, line 34: Does this work? (since it doesn't have a guard, I'm not sure the CurrentUser param decorator can be injected)

apps/financial-aid/backend/src/app/lib/userContextExtractor.ts
```diff
@@ -0,0 +1,34 @@
...
+  decodeToken,
+  IDENTITY_SERVER_SESSION_TOKEN_COOKIE_NAME,
+  User,
+} from '@island.is/financial-aid/shared/lib'
+import { ExecutionContext } from '@nestjs/common'
+import { GqlExecutionContext } from '@nestjs/graphql'
+
+export const getUserFromContext = (
+  context: ExecutionContext & { contextType?: string },
+): User | undefined => {
```
darrikonn, line 11: This is almost the same code as in `current.user.decorator`. Should this be in a lib?

apps/financial-aid/web-veita/pages/api/auth/[...nextauth].ts
```diff
@@ -0,0 +1,88 @@
+import NextAuth, { Callbacks } from 'next-auth'
```
Comment (darrikonn) [inline apps/financial-aid/web-veita/pages/api/auth/[...nextauth].ts:1]: Should this be in a lib (apps/financial-aid/web-osk/pages/api/auth/[...nextauth].ts)

apps/financial-aid/backend/src/app/modules/municipality/municipality.controller.ts
Rikkiabb: Yes removed this guard since we now use `IdsUserGuard` instead which is used for this whole controller 😄 

And this was just commented out code, decided to throw it in the bin 🗑️  and when we'll implement it then we will just add it back and use it 👼

apps/financial-aid/api/src/app/modules/application/application.resolver.ts
Rikkiabb: True, I'll add `IdsUserGuard` for this resolvers and others 😃

apps/financial-aid/api/src/app/modules/user/user.resolver.ts
Rikkiabb: Very true, added a guard 😄

apps/financial-aid/backend/src/app/lib/userContextExtractor.ts
Rikkiabb: Well you are not wrong 😅  I simplified this some, let me know if you still feel strongly about it. Currently we only have `financial-aid-shared`, not sure if this should be there.

apps/financial-aid/web-veita/pages/api/auth/[...nextauth].ts
Comment (Rikkiabb) [reply apps/financial-aid/web-veita/pages/api/auth/[...nextauth].ts:1]: Could you perhaps explain better? What do you want in a lib? The whole api route/s?

apps/financial-aid/api/src/app/modules/user/user.resolver.ts
```diff
@@ -1,79 +1,60 @@
...
   async currentApplication(
     @Parent() user: User,
-  ): Promise<CurrentApplicationModel | null> {
-    // Local development
-    this.logger.debug(
-      `Getting current application for nationalId: ${user.nationalId}`,
+    @Context('dataSources') { backendApi }: { backendApi: BackendAPI },
+  ): Promise<CurrentApplicationModel | undefined> {
+    this.logger.debug('Getting current application for nationalId')
+    return await this.handleNotFoundException(() =>
```
MargretFinnboga, line 47: Is this error not handled in the backend? application.controller? 🤓

MargretFinnboga (commented):
🏂

apps/financial-aid/api/src/app/modules/user/user.resolver.ts
Rikkiabb: Nope it is thrown there, originally I wanted to return undefined in the controller because `currentApplication` is optional. But when I threw `user.service.ts` into the trash and started to use `backend.ts` like every other route we use I ran into a problem that the controller returned undefined and then `apollo-datasource-rest` tries to run `.json()` on that value and then eats the error and returns an empty string, which I don't want. 

So then I decided to throw an error in the controller, if we have no current application and catch it in the resolver and return undefined for it. 

Up for discussion if we want to take another route 😄 

The reason I refactored this is I didn't like that all routes were going through the same flow except for routes used in this resolver.

apps/financial-aid/api/src/services/backend.ts
```diff
@@ -98,6 +100,14 @@ class BackendAPI extends RESTDataSource {
   ): Promise<CreateFilesResponse> {
     return this.post('file', createApplicationFiles)
   }
+
+  getCurrentApplication(nationalId: string): Promise<CurrentApplicationModel> {
+    return this.get(`currentApplication/?nationalId=${nationalId}`)
```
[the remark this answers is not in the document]
Rikkiabb: Agree 🙌  Fixed ✅

apps/financial-aid/web-veita/pages/api/auth/[...nextauth].ts
Comment (darrikonn) [reply apps/financial-aid/web-veita/pages/api/auth/[...nextauth].ts:1]: The content of this file handles all auth related (and is most likely valuable to other projects). Its content is the exact same as `apps/financial-aid/web-osk/pages/api/auth/[...nextauth].ts`.
Wondering if this file should just call something from a library, thus both files can share the same library and any additions/fixes will only be made in a single place.

apps/financial-aid/web-veita/graphql/authLink.ts
```diff
@@ -1,18 +1,15 @@
...
+    headers: {
+      ...headers,
+      authorization: session?.accessToken
+        ? `Bearer ${session.accessToken}`
+        : '',
+    },
   }
-
-  return headers
 })
```
MargretFinnboga, line 15: This is the same file as in web-osk can we maybe move this to libs? 🧎

MargretFinnboga (approved):
👏 🤩 🥇

apps/financial-aid/api/src/app/modules/decorators/currentUser.decorator.ts
```diff
@@ -0,0 +1,36 @@
...
+export const CurrentUser = createParamDecorator(
+  (_: unknown, context: ExecutionContext): User => {
+    const req = GqlExecutionContext.create(context).getContext().req
+
+    const sessionToken = req.cookies
+      ? req.cookies[IDENTITY_SERVER_SESSION_TOKEN_COOKIE_NAME]
+      : null
+
+    if (!sessionToken) {
+      throw new UnauthorizedException('Invalid user')
```
darrikonn, line 24: This is a decorator for GraphQL and your errorLink in the front-end is performing actions based on error extension code: `err.extensions?.code`. I don't think that `UnauthorizedException` creates those error codes for you.

I think the correct approach here is to throw the `AuthenticationError` from 'apollo-server-express' which is the server that @nestjs/graphql builds upon.
That one comes with error codes which your errorlink understands.

apps/financial-aid/backend/src/app/lib/userContextExtractor.ts
```diff
@@ -0,0 +1,33 @@
...
+} from '@island.is/financial-aid/shared/lib'
+import { ExecutionContext, UnauthorizedException } from '@nestjs/common'
+import { GqlExecutionContext } from '@nestjs/graphql'
+
+export const getUserFromContext = (
+  context: ExecutionContext & { contextType?: string },
+): User => {
+  const req =
+    context.contextType === 'graphql'
+      ? GqlExecutionContext.create(context).getContext().req
```
darrikonn, line 14: I don't think this will ever happen. `financial-aid/backend` is implementing a REST server and won't have any inkling of GraphQL context.

apps/financial-aid/backend/src/app/lib/userContextExtractor.ts
```diff
@@ -0,0 +1,33 @@
...
+      : context.switchToHttp().getRequest()
+
+  const sessionToken = req.cookies
+    ? req.cookies[IDENTITY_SERVER_SESSION_TOKEN_COOKIE_NAME]
+    : null
+
+  if (!sessionToken) {
+    throw new UnauthorizedException('Invalid user')
```
darrikonn, line 22: Here `UnauthorizedException` makes sense :+1:

darrikonn (changes requested):

apps/financial-aid/backend/src/app/lib/userContextExtractor.ts
Rikkiabb: True, I was originally going to use the same decorator, but then it ended up not making sense, will remove this.

apps/financial-aid/web-veita/graphql/authLink.ts
Rikkiabb: This is a good point, we could take most out of this folder and reuse it, but let's leave it for a different PR, if we decide to do it.

apps/financial-aid/web-veita/pages/api/auth/[...nextauth].ts
Comment (Rikkiabb) [reply apps/financial-aid/web-veita/pages/api/auth/[...nextauth].ts:1]: Gotcha, I extracted some of the functionality into `libs/financial-aid/shared/src/lib/identityProvider.ts`. But you want to go even further and create something like `libs/next-auth` ? 

I am all for it, just verifying before diving in.

Comment (darrikonn) [reply apps/financial-aid/web-veita/pages/api/auth/[...nextauth].ts:1]: Yeah, if we want to do that.. let's do it in a follow-up PR 👍

darrikonn (approved):
Nice job!

apps/financial-aid/web-osk/graphql/authLink.ts
```diff
@@ -1,18 +1,15 @@
 import { setContext } from '@apollo/client/link/context'
-import { CSRF_COOKIE_NAME } from '@island.is/financial-aid/shared/lib'
-import Cookie from 'js-cookie'
+import { getSession } from 'next-auth/client'
 
-export default setContext((_, { headers }) => {
-  const token = Cookie.get(CSRF_COOKIE_NAME)
+export default setContext(async (_, { headers }) => {
+  const session = await getSession()
```
gudjong, line 5: Was this our missing link?

apps/financial-aid/web-osk/pages/api/auth/[...nextauth].ts
```diff
@@ -0,0 +1,90 @@
...
+      idToken: user.idToken,
+      isRefreshTokenExpired: false,
+      folder: token.folder ?? uuid(),
+      service: RolesRule.OSK,
+    }
+  }
+
+  if (checkExpiry(token.accessToken, token.isRefreshTokenExpired)) {
+    try {
+      const [accessToken, refreshToken] = await refreshAccessToken(
```
Comment (gudjong) [inline apps/financial-aid/web-osk/pages/api/auth/[...nextauth].ts]: Is this working now?

apps/financial-aid/web-osk/pages/index.tsx
```diff
@@ -1,34 +1,20 @@
...
-        data-testid="logout-button"
-        preTextIconType="filled"
-        type="button"
-        variant="primary"
-      >
-        heim
-      </Button>
-    </ContentContainer>
-  )
+  return null
```
gudjong, line 17: What is this component used for?

gudjong (commented):
A few questions

apps/financial-aid/web-osk/pages/api/auth/[...nextauth].ts
Comment (Rikkiabb) [reply apps/financial-aid/web-osk/pages/api/auth/[...nextauth].ts]: Going to test it properly today, but it has been a while since I got the refresh error 😊

apps/financial-aid/web-osk/pages/index.tsx
Rikkiabb: In Osk, we want to have our application domain under `/umsokn` and there is going to be a landing page under `/` which has not been implemented so for now this page is only used to direct the user into the login flow, with a redirect url of `/umsokn`. And then later we will add the landing page :)

PR #5193, opened 2021-10-05
Attempt to squash error logs

sindrig:
Hopefully minimized error logs in island-is/infrastructure#837

We want to start to see if it fixes the graphs in https://app.datadoghq.eu/dashboard/ruv-d65-myv

### DIFF
--- a/libs/cache/src/lib/cache.ts
+++ b/libs/cache/src/lib/cache.ts
@@ -53,6 +53,7 @@ const getRedisClusterOptions = (
   return {
     keyPrefix: options.noPrefix ? undefined : `${options.name}:`,
     slotsRefreshTimeout: 2000,
+    slotsRefreshInterval: 10000,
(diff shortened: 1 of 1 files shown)


eirikurn (approved):
Do you mean the constant "New Connections" graph? Not sure if changing slotsRefreshInterval from 5000 to 10000 makes a big difference there or in error logs. 🤷 

But worth a try

sindrig:
> Do you mean the constant "New Connections" graph? Not sure if changing slotsRefreshInterval from 5000 to 10000 makes a big difference there or in error logs. shrug
> 
> But worth a try

Yes - the slots refresh creates a new connection every `slotsRefreshInterval` ms, so it should theoretically lower a lot. But I agree, I don't really believe it will change a lot wrt the logs, and really little we can do about it it seems :(

PR #5199, opened 2021-10-06
feat(search): Tweak search engine, fuzziness and default operator

kksteini:
## What
Tweaking the search engine to allow mistakes and be a bit more reliable.

Introducing fuzzy search for terms over a certain word-length via remapping.
See: https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-simple-query-string-query.html#simple-query-string-syntax

## Screenshots

On every image, the new search is on the left. 
Island.is as it is currently is on the right.

### Fuzziness for corrections
image

image

### Number of results

There are pros and cons. For example, here we have very many results in the new search engine however, arguably, the first results are still valid:

image

#### Garbled input
We can return correct results although a lot of noise can come with garbled input. A case can be made though that in this case "leigu" should return something corresponding which it does here. Room for more tweaking.

image

#### AND vs OR
Currently the search engine is `AND`-ing together terms. This reduces the result space but can on occasion hide the intended result.
Again, in contrast, this is a question between running the risk of cluttering the results with noise vs validity of proper searches.
image

## Checklist:


libs/content-search-toolkit/src/queries/search.ts
```diff
@@ -11,17 +11,31 @@ export const searchQuery = ({
...
+      .join(' ')
+  }
 
   should.push({
     simple_query_string: {
       query: queryString,
       fields: ['title.stemmed^15', 'title.compound', 'content.stemmed^5'],
       analyze_wildcard: true,
-      default_operator: 'and',
+      default_operator: 'or',
```
siggifv, line 38: When `OR`-ing together multiple terms, how is the ranking of result hits? Does it rank hits higher if it matches all terms? I.e. we want the topmost search result to be the most relevant based on all the terms.

kksteini: So basically "no, but it's not that simple"

kksteini: Actually, this is hasty and it's been rightfully pointed out that there are problems with `OR` which I didn't catch on. There are other ways and methods to be explored.
Closing this for now

PR #5203, opened 2021-10-06
feat(udicia-system): Mask Completed Investigation Cases

gudjong:
# Mask Completed Investigation Cases

https://app.asana.com/0/1199153462262248/1201085301643342

## What

☝️ 

## Why

Investigation cases need extra security.

## Screenshots / Gifs

image

## Checklist:


apps/judicial-system/api/src/app/modules/case/interceptors/case.transformer.spec.ts
```diff
@@ -1,6 +1,7 @@
-import { SessionArrangements } from '@island.is/judicial-system/types'
 import each from 'jest-each'
 
+import { SessionArrangements } from '@island.is/judicial-system/types'
```
oddsson, line 3: Is this an accidental import? Are you using it anywhere?

apps/judicial-system/web/src/utils/mutations.ts
```diff
@@ -16,7 +16,6 @@ export const CasesQuery = gql`
       id
       created
       type
-      description
```
oddsson, line 19: Do we not need this anymore? Isn't this "Efni kröfu" in investigation cases?

apps/judicial-system/api/src/app/modules/case/interceptors/case.transformer.spec.ts
gudjong: It is used in the tests below in an each-clause.

apps/judicial-system/web/src/utils/mutations.ts
oddsson: Nope! This looks good

PR #5222, opened 2021-10-07
feat(delegation): UX improvements

darrikonn:
# UX improvements

Implements:
- https://www.notion.so/Lets-revisit-the-date-of-scope-on-the-A-gangst-ring-panel-1bcf64c0e5b847aa90b9378cea21bf47
- https://www.notion.so/Improve-UX-when-many-scopes-3ba0dd40c9d544d3a53add498fd974c8

### Gif
datepickerfix

### Before
image

### After
image

## Checklist:


libs/service-portal/settings/access-control/src/screens/AccessControl/components/AccessCard/AccessCard.tsx
```diff
@@ -56,8 +56,8 @@ function AccessCard({
               ? format(new Date(validTo), 'dd.MM.yyyy')
               : formatMessage({
                   id:
-                    'service.portal.settings.accessControl:home-view-undetermined',
-                  defaultMessage: 'Óákveðið',
+                    'service.portal.settings.accessControl:home-view-undefined',
+                  defaultMessage: 'Óskilgreint',
```
saevarma, line 60: According to Eiki's comment in https://www.notion.so/Lets-revisit-the-date-of-scope-on-the-A-gangst-ring-panel-1bcf64c0e5b847aa90b9378cea21bf47 wasn't the plan to remove this or is the ticket missing some comment about another solution?

darrikonn: So this ticket has a longer history and was supposed to be completed (was being tracked in two places).

## Previous fix
The fix was to only display `validTo` if **all scopes agree on the expiry date**. Otherwise render `"Óákveðið"`

## This fix
This ticket bubbled up again regarding changing the name from "Óákveðið" to "Óskilgreint". At that point I saw Eiki's comment on not displaying anything and tried that out (656bdd9). This however implied that the delegation didn't have an expiry date at all, which was a worse UX, so I reverted the changes and made the change "Óákveðið" to "Óskilgreint" ... as can be seen in my commit message:
```
@darrikonn
Revert changes of valid to b89e56e
The previous fix implied that no date had been set for when the delegation expired, which wasn't true. Thus render undefined instead
```

Good spot though!

saevarma: Reading "Óskilgreint" feels kind of the same. I would understand that as the `validTo` to be `undefined` meaning no expiry date. Could we find some other word that could give the meaning: various. I.e. "Mismunandi" or "Ýmislegt"?

image

image

image

PR #5233, opened 2021-10-11
Feat driving license client v2

arnorhs:
# ...

## What

## Why

## Screenshots / Gifs

## Checklist:


jGeir (approved):
lgtm. +1 for api nx command

libs/clients/driving-license/v1/src/lib/drivingLicenseApiV1.module.ts
```diff
@@ -0,0 +1,45 @@
+import { DynamicModule } from '@nestjs/common'
+import { createEnhancedFetch } from '@island.is/clients/middlewares'
+import {
+  Configuration,
+  OkuskirteiniApi as OkuskirteiniApiV1,
```
eirikurn, line 5: Let's subclass the generated API classes and register those instead of using a injection token:

```typescript
import { OkuskirteiniApi } from '../../gen/fetch'
export class OkuskirteiniApiV1 extends OkuskirteiniApi {}
```

Just remember to re-export this class in the root index.ts (and not the original class from `gen/fetch`).

eirikurn (changes requested):
Nice to get a script to fix the downloaded OpenAPI file.

I've got a suggestion for the Nest naming conflict issue.

However, is it not safe/possible to upgrade all of our clients to the v2 API? I really don't like to collect so many versioned API clients in the monorepo. If it's risky, then maybe we can put it on the backlog.

If we **need** both (because they have different endpoints), then I wonder if we should have one client module with both, maybe with some additional Api class extending/massaging.

sindrig (approved):

PR #5240, opened 2021-10-11
fix(judicial-system): Fix 403 when revoking a case

oddsson:
# Fix 403 when sending a REVOKE notification

## What

Fix 403 when sending a REVOKE notification by sending first and then revoking

## Why

We get a 403 when we try to send a notification for a case that has been revoked

## Checklist:


apps/judicial-system/web/src/routes/Shared/Requests/Requests.tsx
```diff
@@ -89,14 +89,14 @@ export const Requests: React.FC = () => {
...
-      const caseDeleted = await transitionCase(
-        caseToDelete,
-        CaseTransition.DELETE,
+      const notificationSent = await sendNotification(
+        caseToDelete.id,
+        NotificationType.REVOKED,
       )
 
-      if (caseDeleted) {
+      if (notificationSent) {
```
gudjong, line 97: Consider not checking for notification success. The notification is not critical and we should probably delete the case even if notifications fail.

gudjong (changes requested):

PR #5252, opened 2021-10-12
feat(judicial-system): Get police case files

gudjong:
# Get police case files

https://app.asana.com/0/1199153462262248/1200403547319712

## What

Part of ☝️ 

## Why

In order to select police case files, we first need to get them from the police.

## Checklist:


oddsson (changes requested):
There is one type :)

apps/judicial-system/api/infra/judicial-system-api.ts
```diff
@@ -44,8 +44,8 @@ export const serviceSetup = (services: {
       },
       HIDDEN_FEATURES: {
         dev: '',
-        staging: '',
-        prod: '',
+        staging: 'POLICE_CASE_FILES',
```
pshomov, line 47: You can now use https://docs.devland.is/technical-overview/devops/service-setup#server-side-feature-flags maybe?

PR #5258, opened 2021-10-12
feat(fa): Adds phonenumber to form

Rikkiabb:
# ...
https://app.asana.com/0/1201097099198335/1201068271296084

## What
Adds phonenumber to osk flow

## Why
We don't get it from login anymore


apps/financial-aid/web-osk/src/routes/application/ContactInfo/contactInfo.tsx
```diff
@@ -14,24 +14,25 @@ import {
...
 } from '@island.is/financial-aid/shared/lib'
 
-const EmailForm = () => {
+const ContactInfo = () => {
   const router = useRouter()
 
   const { form, updateForm } = useContext(FormContext)
-  const [hasError, setHasError] = useState(false)
+  const [hasEmailError, setHasEmailError] = useState(false)
+  const [hasPhoneNumberError, setHasPhoneNumberError] = useState(false)
```
MargretFinnboga, line 22: Does this have to be two separate states, can we not just have hasError? ❌

apps/financial-aid/web-osk/src/routes/application/ContactInfo/contactInfo.tsx
```diff
@@ -14,24 +14,25 @@ import {
...
-    if (form?.emailAddress === undefined) {
-      setHasError(true)
+    if (form?.emailAddress === undefined || !isEmailValid(form?.emailAddress)) {
+      setHasEmailError(true)
       return
     }
 
-    if (!isEmailValid(form?.emailAddress)) {
-      setHasError(true)
+    if (form?.phoneNumber === undefined || form?.phoneNumber.length !== 7) {
```
MargretFinnboga, line 34: is it legit to check if number has 7 in length, what if I have a phone number from abroad? ✈️ 🛄

apps/financial-aid/web-osk/src/routes/application/ContactInfo/contactInfo.tsx
MargretFinnboga: Yess I agree, however you can add that to hasError value, so hasError checks if there is an error, and then the input takes in hasError && validation, 
so for example in the email input 
hasError={hasError && !isEmailValid(form?.emailAddress)}
And phone input 
hasError={hasError && form?.phoneNumber.length !== 7}. 

It is just a suggestion, I think it might be more of scalable solution if there will be added another input there, will we add another useState error for that too?  🤓 😇

PR #5259, opened 2021-10-12
fix(fa): Updates tax stepper

Rikkiabb:
# ...

https://app.asana.com/0/1201097099198335/1201038910099145


apps/financial-aid/web-osk/src/components/Files/Files.tsx
```diff
@@ -12,7 +12,7 @@ interface Props {
   header: string
   uploadFiles: UploadFile[]
   fileKey: UploadFileType
-  hasError: boolean
+  hasError?: boolean
```
MargretFinnboga, line 15: If this is default false, then why have it possible undefined? 👨‍🏫

PR #5277, opened 2021-10-13
fix(financial-aid): Files refactoring

MargretFinnboga:
# Files refactoring

https://app.asana.com/0/1201097099198335/1201176723556470

## What

When retrieving application files modal is included 
Fixing few errors here and there
Updated how we read through files array in application veita

## Checklist:

### DIFF
--- a/apps/services/endorsements/api/src/app/modules/endorsementList/endorsementList.controller.ts
+++ b/apps/services/endorsements/api/src/app/modules/endorsementList/endorsementList.controller.ts
@@ -88,6 +88,26 @@ export class EndorsementListController {
     )
   }
 
+  @ApiOperation({
+    summary:
+      'Finds all endorsement lists owned by the currently authenticated user',
+  })
+  @ApiOkResponse({ type: PaginatedEndorsementDto })
+  @Get('/endorsementLists')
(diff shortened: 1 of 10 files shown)


libs/financial-aid/shared/src/lib/const.ts
```diff
@@ -11,7 +11,7 @@ export const Routes = {
       : '/api/auth/login?nationalId=',
   apiLoginRouteForRealUsers: (id: string) =>
     id ? `/api/auth/login?applicationId=${id}` : '/api/auth/login',
-  filesPage: '/gogn',
+  filesPage: '/umsokn/skattframtal',
```
Rikkiabb, line 14: Isn't this page called skattagogn now?

libs/financial-aid/shared/src/lib/formatters.ts
```diff
@@ -96,6 +97,12 @@ export const getActiveTypeForStatus: KeyMapping<ApplicationState, string> = {
   Approved: 'Approved',
 }
 
+export const getFileTypeName: KeyMapping<FileType, string> = {
+  TaxReturn: 'Skattframtal',
```
Rikkiabb, line 101: Should this be Skattagögn?

PR #5290, opened 2021-10-14
feat(license-service): add validation w/cut-off point for driving license pkpass

osk:
Adds validation to driving licenses pkpass:

* License must have an image
* Image must be registered at least at 1997-08-15

When querying for a license a new `pkpassStatus` enum is returned with possible values:

* `Available`, a call to the `generatePkPassMutation` will return a URL with the pkpass
* `NotAvailable`, the license is not valid for pkpass generation
* `Unknown`, something happened so the status couldn't be validated

## Checklist:


libs/service-portal/licenses/src/screens/components/LicenseCards/LicenseCards.tsx
```diff
@@ -284,7 +285,9 @@ const LicenseCards = () => {
             <dt>Staða</dt>
             <dd>{license.license.status}</dd>
             <dt>pkpass?</dt>
-            <dd>{license.license.pkpass}</dd>
+            <dd>{license.license.pkpass ? 'Já' : 'Nei'}</dd>
```
thordurhhh, line 288: Is it ok to have this not translated here?

osk: yes, it's just a dummy ui for testing

PR #5293, opened 2021-10-15
feat(financial-aid): Sending email on state change

MargretFinnboga:
# Sending email on state change

https://app.asana.com/0/1201097099198335/1201017282842011

## What

When application is update, send email on applicant with right text

## Why

So the user always knows what the state of his application is

## Checklist:


apps/financial-aid/backend/src/app/modules/application/application.service.ts
```diff
@@ -207,6 +212,15 @@ export class ApplicationService {
     const files = await this.fileService.getAllApplicationFiles(id)
     updatedApplication?.setDataValue('files', files)
 
+    await this.sendEmail(
```
Rikkiabb, line 215: Aren't we here sending an email for every update to the application? 

For example when we are adding files and changing employee that is working on the application? In that case no email should be sent.

PR #5305, opened 2021-10-15
chore(judicial-system): Upgrades helm resources

gudjong:
# Upgrades helm resources

We need more resources temporarily


apps/judicial-system/backend/infra/judicial-system-backend.ts
```diff
@@ -81,3 +81,7 @@ export const serviceSetup = (): ServiceBuilder<'judicial-system-backend'> =>
     .liveness('/liveness')
     .readiness('/liveness')
     .postgres(postgresInfo)
+    .resources({
+      requests: { cpu: '750m', memory: '750Mi' },
+      limits: { cpu: '1000m', memory: '1Gi' },
```
AndriMar, line 86: Should be:
requests: { cpu: '100m', memory: '256Mi' },
 limits: { cpu: '400m', memory: '512Mii' },

PR #5326, opened 2021-10-19
feat(endorsement-system-api): Get endorsement list owner name from national registry

alexdiljar:
# ...

Owner was displayed as kennitala but should not be returning kennitala just owner name.

## What

Added National registry api to the provider for endorsement system and a resolver field that populates ownerName when queried.

## Why

So we dont expose owners ssn.

## Checklist:

### DIFF
--- a/libs/service-portal/endorsements/src/screens/ViewPetitionAdmin/ViewPetitionAdmin.tsx
+++ b/libs/service-portal/endorsements/src/screens/ViewPetitionAdmin/ViewPetitionAdmin.tsx
@@ -170,8 +170,8 @@ const ViewPetitionAdmin = () => {
           <Input
             backgroundColor="blue"
             disabled
-            name={petition?.ownerName}
-            value={petition?.ownerName}
+            name={petition?.ownerName ?? ''}
+            value={petition?.ownerName ?? ''}
             label={formatMessage(m.viewPetition.listOwner)}
           />
 


libs/api/domains/endorsement-system/src/lib/endorsementSystem.module.ts
```diff
@@ -7,11 +7,15 @@ import {
   Configuration,
   EndorsementListApi,
 } from '../../gen/fetch'
+import {
+  NationalRegistryApi,
+  NationalRegistryConfig,
+} from '@island.is/clients/national-registry-v1'
```
saevarma, line 13: Shouldn't we use the v2 service which is through XRoad?

saevarma (changes requested):

apps/services/endorsements/api/src/app/modules/endorsementList/endorsementList.service.ts
```diff
@@ -205,4 +207,12 @@ export class EndorsementListService {
     }
     return result
   }
+
+  async getOwnerInfo(endorsementList: EndorsementList) {
+    const fullName = (
+      await this.nationalRegistryApi.getUser(endorsementList.owner)
+    ).Fulltnafn
+    return (await this.nationalRegistryApi.getUser(endorsementList.owner))
+      .Fulltnafn //{fullName: fullName}
```
stefanhar, line 216: This seems a bit off, isn't this the same code repeated?

libs/api/domains/endorsement-system/src/lib/models/endorsementList.model.ts
```diff
@@ -30,6 +30,7 @@ export class EndorsementList {
   @Field(() => [ValidationRule])
   validationRules!: ValidationRule[]
 
+  // Remove this field when changes have been made to frontend so that we dont expose the owners ssn
```
stefanhar, line 33: It might be better to update the frontend in this PR as well.
Since that confirm that this flow works in frontend

apps/services/endorsements/api/src/app/modules/endorsementList/endorsementList.controller.ts
```diff
@@ -307,4 +308,23 @@ export class EndorsementListController {
...
+  async getOwnerInfo(
+    @Param(
+      'listId',
+      new ParseUUIDPipe({ version: '4' }),
+      EndorsementListByIdPipe,
+    )
+    endorsementList: EndorsementList,
+  ): Promise<String> {
+    return await this.endorsementListService.getOwnerInfo(endorsementList)
+  }
```
stefanhar, line 329: I wonder if it might be better to implement this as an [interceptor](https://docs.nestjs.com/interceptors).
This would allow us to centralise this functionality and might better ensure consistency

stefanhar (changes requested):

PR #5332, opened 2021-10-19
infra(services-auth-api): adding extra DD tracing info to services-auth-api

pshomov:
# ...

## What

## Why

## Screenshots / Gifs

## Checklist:


libs/auth-api-lib/src/lib/services/claims.service.ts
```diff
@@ -2,7 +2,9 @@ import { Inject, Injectable } from '@nestjs/common'
 import { InjectModel } from '@nestjs/sequelize'
 import { Claim } from '../entities/models/claim.model'
 import { Sequelize } from 'sequelize-typescript'
+import { trace } from '@theo.gravity/datadog-apm'
 
+@trace()
```
eirikurn, line 7: @pshomov The existing traces indicate that we won't find the real issue like this. 

There's a common span for a super simple service method which takes 4 seconds "just to call `this.logger.debug()`" (a no-op in production) before doing a simple SQL query which takes less than 0,1s.

This most likely revolves around "awaiting promises" which schedules a new tick in the event loop, and something else is DOSing the CPU during that time. We won't find the culprit with tracing spans because we can't see what it's actually doing in each span because of the async threaded nature of NodeJS.

We need to find the CPU bottleneck some other way and/or parallelize more (use more of the pods available CPU cores or more single-core pods).

eirikurn: It's a bit weird though that almost all SQL queries look super fast in APM ...

I imagine the SQL spans are traced in NodeJS just like everything else, and involves tracing network events which also run on the event loop. So they should be just as slow if the event loop is stuck while serializing JSON or something.

pshomov: Desperate times, desperate measures ;)

We are providing enough memory and CPU to the NodeJS service according to K8s. It autoscaled the service to about 40 pods and each one was barely above 50% both in CPU and memory utilisation. One concerning thing is the event loop delay is ~600ms, which sounds indicative of something strange going on on the main thread.

The tracing of Postgres is (I suspect) done by monkey-patching pg-postgres (+native) module.
We gave NodeJS a whole CPU core (we are kind of tight on those) and ~2GB RAM. Things did not improve much.

PR #5335, opened 2021-10-20
fix(service-portal): error handling for finance section

thordurhhh:
## What

Error handling for finance services in the service-portal

## Why

Was missing.

## Checklist:


libs/clients/finance/src/lib/finance.ts
```diff
@@ -38,20 +43,30 @@ export class FinanceService extends RESTDataSource {
     request.headers.set('X-Road-Client', this.options.xroadClientId)
   }
 
+  handleError(error: string): any {
```
osk, line 46: We must log the exception, something like
Also should return undefined? at least we know it'll always throw
```suggestion
  handleError(error: string, exception: Error): any {
```

libs/clients/finance/src/lib/finance.ts
```diff
@@ -38,20 +43,30 @@ export class FinanceService extends RESTDataSource {
...
+    this.logger.error(error)
```
osk, line 47:
```suggestion
    this.logger.error(error, {
      exception: e
    })
```

PR #5361, opened 2021-10-21
fix(data-providers): Make sure family data provider fails when there is an error

baering:
## What

Family data provider was not failing when there was an error in the query to fetch family members.

## Why

So that a user cannot start an application with incorrect state, without family members that may be required to correctly fill the application.

## Screenshots / Gifs

image

## Checklist:

### DIFF
--- a/libs/application/data-providers/src/providers/FamilyInformationProvider.ts
+++ b/libs/application/data-providers/src/providers/FamilyInformationProvider.ts
@@ -2,10 +2,14 @@ import {
   BasicDataProvider,
   SuccessfulDataProviderResult,
   FailedDataProviderResult,
+  StaticText,
+  coreErrorMessages,
 } from '@island.is/application/core'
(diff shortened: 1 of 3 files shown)


libs/application/data-providers/src/providers/FamilyInformationProvider.ts
```diff
@@ -35,16 +39,24 @@ export class FamilyInformationProvider extends BasicDataProvider {
...
   handleError(error: Error | unknown) {
+    if (isLocalDevelopment) {
+      return Promise.resolve([])
+    }
+
     console.error('Provider error - FamilyInformation:', error)
-    return Promise.resolve([])
+    return Promise.reject({
+      reason: coreErrorMessages.errorDataProviderDescription,
+    })
```
eirikurn, line 49: This error handling is weird 🙈

eirikurn (approved):

PR #5395, opened 2021-10-25
docs(application-system-payments): Adding docs for how to implement application payments

arnorhs:
The documentation also highlights how much this needs to be made simpler / easier to get right / harder to mess up

## Checklist:


eirikurn (approved):
Approved without automerge.

Would be nice if this guide was under the application-system folder, both in the repo and in gitbook. If you already tried this and the gitbook script doesn't support this, then feel free to add the automerge tag and merge it as is. 👍

arnorhs:
> Approved without automerge.
> 
> Would be nice if this guide was under the application-system folder, both in the repo and in gitbook. If you already tried this and the gitbook script doesn't support this, then feel free to add the automerge tag and merge it as is. 👍

yeah, i did actually try that.. i didn't dig deep though, i just tried adding in a place that i thought made sense.. i forget where it was thoguh

PR #5412, opened 2021-10-27
chore(infra): run tests on main to get code coverage

dabbeg:
## What

Run tests on main

## Why

So we get the test coverage to upload to CodeCov

## Checklist:


scripts/download-codecov-uploader.sh
```diff
@@ -1,7 +1,6 @@
 #!/usr/bin/env bash
```
sindrig, line 1: how long does this usually take? is it worth it to use the cache considering the time difference and the added boilerplate?

.github/workflows/push.yml
```diff
@@ -277,6 +305,58 @@ jobs:
           UNAFFECTED=$(./scripts/ci/list-unaffected.sh docker-next docker-express docker-static docker-cypress)
           echo "::set-output name=UNAFFECTED::$UNAFFECTED"
 
+  tests:
```
sindrig, line 308: Consider moving this to a [composite action](https://docs.github.com/en/actions/creating-actions/creating-a-composite-action)

pshomov: @sindrig I find it disappointing that composite actions do not support conditional steps. I guess you can use scripting when having a shell command but kind of hard when you are `use`-ing an action.

scripts/download-codecov-uploader.sh
pshomov: 61 MB binary

pshomov (approved):
LGTM, the cache for the codecov binary might be a bit too much, cache can be invalidated between jobs (but not after our 🦄  S3 cache)

scripts/download-codecov-uploader.sh
sindrig: caching it in azure vs downloading it from some cdn /shrug

dabbeg: It took 4s when I tested this in the pipeline. I was fooled by my very slow internet connection.
I am removing the caching

sindrig: image

sindrig: denmark

PR #5431, opened 2021-10-28
chore(ci): Use S3 for build cache

sindrig:


pshomov (approved):
LGTM

libs/cache/src/lib/cache.ts
```diff
@@ -27,9 +27,33 @@ class Cache {
...
+  getMap(key: string): Promise<Record<string, string>> {
+    return this.client.hgetall(key)
+  }
+
   expire(key: string, seconds: number): Promise<Redis.BooleanResponse> {
     return this.client.expire(key, seconds)
   }
+
+  async ping(): Promise<boolean> {
+    return (await this.client.ping()) == 'pong'
```
dabbeg, line 55: ===

apps/github-actions-cache/src/app/handlers.ts
```diff
@@ -0,0 +1,166 @@
...
+  }
+  try {
+    logger.debug(`Reserving cache id ${cacheId}`)
+    const reservation = await cache.setKeyIfNotExists(
+      cacheId,
+      'cacheId',
+      version,
+    )
+    logger.debug(`Reservation: ${reservation}`)
+    if (reservation != 1) {
```
dabbeg, line 34: Would be more explicit to use !== and convert to int if reservation is string

apps/github-actions-cache/src/app/handlers.ts
```diff
@@ -0,0 +1,166 @@
...
+  logger.debug(`Completing multipartupload with map`, MultipartUpload)
+  await s3
+    .completeMultipartUpload({
+      Bucket: bucket,
+      Key: cacheId,
+      MultipartUpload,
+      UploadId: uploadId,
+    })
+    .promise()
+  cache.expire(cacheId, 1)
```
dabbeg, line 130: Any reason for setting expiry to 1 second instead of just deleting?

apps/github-actions-cache/src/app/utils.ts
```diff
@@ -0,0 +1,63 @@
...
+  logger.warn(`Sending error "${error}" with status ${status}`)
+  return res.status(status).send({ error })
+}
+
+// In case we want to invalidate all caches in code
+const salt = '3'
+export const getCacheId = (key: string, version: string) =>
+  `${key}:${version}:${salt}`
+
+const uploadChunkSize = 32 * 1024 * 1024
```
dabbeg, line 17: Is this 32MB? Is this some recommended chunksize you came across?

libs/cache/src/lib/cache.ts
sindrig: ( ͡° ͜ʖ ͡°)

apps/github-actions-cache/src/app/utils.ts
sindrig: This is the default chunk size in the actions/cache package: https://github.com/actions/toolkit/blob/e2eeb0a784f4067a75f0c6cd2cc9703f3cbc7744/packages/cache/src/options.ts#L59

sindrig: We `could` try to figure it out ourselves, but I'm afraid it could get messy...

apps/github-actions-cache/src/app/handlers.ts
sindrig: I was not sure how deletions in redis clusters behaved and there was already some expiry method on the interface. I had some strange results locally when deleting keys where it did not propagate correctly.

dabbeg (approved):

apps/github-actions-cache/docker-compose.dev.yml
```diff
@@ -0,0 +1,13 @@
+version: '3.3'
+
+services:
+  redis-cluster:
+    container_name: ads_redis_cluster
```
darrikonn, line 5: ads is a short for the air-discount-scheme project.

darrikonn (commented):
What's the reason for these changes?

sindrig:
> What's the reason for these changes?

https://github.com/actions/cache/pull/679

I think this PR sums it up quite nicely. It'll also mean we won't hit the 5gb limit multiple time per day.

apps/github-actions-cache/docker-compose.dev.yml
sindrig: renamed

saevarma (approved):
LGTM

darrikonn (approved):

PR #5437, opened 2021-10-29
fix(application-system): Adding polyfill for submitter prop in Safari and validating invalid event for changeState

saevarma:
## What

The application-system depends on a property named `submitter`  on `SubmitEvent` for determine which button caused form submition. This property is not yet [supported in Safari](https://developer.mozilla.org/en-US/docs/Web/API/SubmitEvent/submitter). As a workaround we have added a polyfill to add this property to the event for Safari browsers.

We have also updated the `changeState` method to throw an error if the application-system tries to changeState for invalid event.

## Why

This was causing error for Safari users in Parental Leave, when other parent needed to approve requested days.
The UI displayed as the actions has succeeded but on the server nothing happened because of invalid event for the state.

## Checklist:


eirikurn (approved):
LGTM

PR #5444, opened 2021-11-01
fix(fa): Error state if aid amount is not calculated right

MargretFinnboga:
☝️ 

https://app.asana.com/0/1201301944561348/1201257203922140

## Why

Bug fixing

## Checklist:

### DIFF
--- a/apps/financial-aid/web-veita/src/components/ApplicationProfile/ApplicationProfile.tsx
+++ b/apps/financial-aid/web-veita/src/components/ApplicationProfile/ApplicationProfile.tsx
@@ -61,19 +61,19 @@ const ApplicationProfile = ({
 
   const currentYear = format(new Date(), 'yyyy')
 
-  const applicationInfo = aidAmount
-    ? [
-        {
-          title: 'Tímabil',
-          content:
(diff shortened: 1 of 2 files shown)


apps/financial-aid/web-veita/src/components/ApplicationProfile/ApplicationProfile.tsx
```diff
@@ -61,32 +61,46 @@ const ApplicationProfile = ({
...
-          application.usePersonalTaxCredit,
-          currentYear,
-        ).toLocaleString('de-DE')
-      } kr.`,
-      onclick: () => {
-        setAidModalVisible(!isAidModalVisible)
-      },
-    },
-  ]
+  const applicationInfo = aidAmount
```
Rikkiabb, line 64: Can we switch on áætluð aðstoð instead of the whole array?

MargretFinnboga: Good Point!! 🤓

Rikkiabb (approved):

PR #5450, opened 2021-11-01
An/add nationalid validation

sigruntg:
# Add custom national id input validation

https://app.clickup.com/t/1ht3ku4

## What

Added custom input validator, that uses the service call getIdentity to check if there is an actual person behind national Id.
There is no way for me to test this locally. So this has to be tested on prod. 

Tests
I have added to console.logs, one for if it succeeds and one if it fails. If it succeeds it should show no error message, and the user should be able to proceed to next screen. If it fails, the user should get error message saying that this national id does not belong to anyone. 

If this works, someone has to go into the code and uncomment where I wrote TODO and remove console.logs. 

This can be used on more places if this works, but you need to add validation to schema for each place this should be used.

## Checklist:


libs/application/templates/accident-notification/src/hooks/useLazyGetIdentity.ts
```diff
@@ -0,0 +1,22 @@
+import { gql } from '@apollo/client'
```
Thorrik58, line 1: No need for this, we need to use the apollo client usequery for this https://github.com/island-is/island.is/blob/main/libs/application/templates/accident-notification/src/fields/ApplicationStatus/index.tsx#L45

PR #5454, opened 2021-11-02
feat(financial-aid): Spouse status page

MargretFinnboga:
# Spouse status page

https://app.asana.com/0/1201301944561348/1201097099198303

## What

-When checking if user is spouse added also check if spouse has uploaded files
-Give spouse access to application 
-Added status page for spouse, where the spouse gets limited information about the application
-Retrieved municipality code on status page
-Removed session storage in my application

## Why

-Had to check whether the user(spouse) should go to form or status page
-(Retrieved municipality code): For the estimation calculation for applicant on status page to be correct 
-(Removed session storage): Because we always want the lates status on the application and if saved in storage we would need to check for update any ways 

## Checklist:


apps/financial-aid/api/src/app/modules/user/user.model.ts
```diff
@@ -24,8 +29,8 @@ export class UserModel implements User {
   @Field(() => String, { nullable: true })
   readonly currentApplication?: string
 
-  @Field(() => Boolean)
-  readonly isSpouse?: boolean
+  @Field(() => HasSpouseAppliedModel, { nullable: true })
```
Rikkiabb, line 32: Should this be named `SpouseModel` ?

apps/financial-aid/api/src/app/modules/user/user.model.ts
```diff
@@ -24,8 +29,8 @@ export class UserModel implements User {
...
+  readonly isSpouse?: HasSpouseAppliedModel
```
Rikkiabb, line 33: And should this field be named `Spouse` since it is an object now?

apps/financial-aid/api/src/app/modules/user/user.resolver.ts
```diff
@@ -37,13 +38,13 @@ export class UserResolver {
...
-  @ResolveField('isSpouse', () => Boolean)
+  @ResolveField('isSpouse', () => HasSpouseAppliedModel)
   async isSpouse(
     @Parent() user: User,
     @Context('dataSources') { backendApi }: { backendApi: BackendAPI },
-  ): Promise<Boolean> {
+  ): Promise<HasSpouseAppliedModel> {
     const isSpouse = await backendApi.isSpouse(user.nationalId)
-    return isSpouse.HasApplied
+    return isSpouse
```
Rikkiabb, line 47: You could join lines 46-47

apps/financial-aid/backend/src/app/guards/application.guard.ts
```diff
@@ -31,7 +31,10 @@ export class ApplicationGuard implements CanActivate {
         return false
       }
     } else if (user.service === RolesRule.OSK) {
-      if (application.nationalId !== user.nationalId) {
+      if (
+        application.nationalId !== user.nationalId &&
```
Rikkiabb, line 35: Shouldn't this be || ?

apps/financial-aid/backend/src/app/modules/application/application.controller.ts
```diff
@@ -95,11 +95,8 @@ export class ApplicationController {
   ): Promise<IsSpouseResponse> {
     this.logger.debug('Application controller: Checking if user is spouse')
 
-    return {
-      HasApplied: await this.applicationService.hasSpouseApplied(
-        spouseNationalId,
-      ),
-    }
+    const res = await this.applicationService.hasSpouseApplied(spouseNationalId)
+    return res
```
Rikkiabb, line 99: You could join these lines

apps/financial-aid/backend/src/app/modules/application/application.service.ts
```diff
@@ -29,6 +29,7 @@ import { EmailService } from '@island.is/email-service'
 
 import { ApplicationFileModel } from '../file/models'
 import { environment } from '../../../environments'
+import { IsSpouseResponse } from './models/isSpouse.response'
```
Rikkiabb, line 32: Should this be renamed spouseResponse?

apps/financial-aid/backend/src/app/modules/application/application.service.ts
```diff
@@ -55,15 +56,22 @@ export class ApplicationService {
     private readonly emailService: EmailService,
   ) {}
 
-  async hasSpouseApplied(spouseNationalId: string): Promise<boolean> {
+  async hasSpouseApplied(spouseNationalId: string): Promise<IsSpouseResponse> {
```
Rikkiabb, line 59: Should we rename this to `spouse` or `spouseInformation` or something like that?

apps/financial-aid/backend/src/app/modules/application/application.service.ts
```diff
@@ -55,15 +56,22 @@ export class ApplicationService {
...
     const application = await this.applicationModel.findOne({
       where: {
         spouseNationalId,
         created: { [Op.gte]: firstDateOfMonth() },
       },
     })
 
-    return Boolean(application)
+    const files = Boolean(application)
```
Rikkiabb, line 67: Do we need to wrap application with Boolean here?

apps/financial-aid/backend/src/app/modules/application/models/isSpouse.response.ts
```diff
@@ -3,4 +3,7 @@ import { ApiProperty } from '@nestjs/swagger'
 export class IsSpouseResponse {
   @ApiProperty()
   HasApplied: boolean
+
+  @ApiProperty()
+  HasSpouseFiles: boolean
```
Rikkiabb, line 8: If we rename the response to SpouseResponse, then this field could be hasFiles.

apps/financial-aid/backend/src/app/modules/file/file.service.ts
```diff
@@ -63,6 +65,18 @@ export class FileService {
     })
   }
 
+  async getSpouseApplicationFiles(
```
Rikkiabb, line 68: We could have this function take in fileType, rename it to getApplicationFilesByType. Then it could be used for more purposes then this single case, if we need it in the future, just a thought.

apps/financial-aid/web-osk/src/components/Status/Approved.tsx
```diff
@@ -25,10 +31,16 @@ const Approved = ({ state, amount, isStateVisible }: Props) => {
       <Text as="h2" variant="h3" color="mint600" marginBottom={[4, 4, 5]}>
         Umsókn {getState[state].toLowerCase()}
       </Text>
-
-      <Text as="h3" variant="h3" marginBottom={2}>
-        Veitt aðstoð :{amount?.toLocaleString('de-DE') + ' kr.'}
-      </Text>
+      {!isUserSpouse ? (
```
Rikkiabb, line 34: Wondering if we should revert this and have isApplicant? Like to have the default case rather then the edge case here.

apps/financial-aid/web-osk/src/components/Status/InProgress.tsx
```diff
@@ -16,9 +16,10 @@ import { useRouter } from 'next/router'
 
 interface Props {
   application: Application
+  isUserSpouse?: boolean
 }
 
-const InProgress = ({ application }: Props) => {
+const InProgress = ({ application, isUserSpouse = false }: Props) => {
```
Rikkiabb, line 22: and here.

apps/financial-aid/web-osk/src/components/Status/Rejected.tsx
```diff
@@ -7,9 +7,15 @@ interface Props {
...
   isStateVisible: boolean
+  isUserSpouse?: boolean
 }
 
-const Rejected = ({ state, rejectionComment, isStateVisible }: Props) => {
+const Rejected = ({
+  state,
+  rejectionComment,
+  isStateVisible,
+  isUserSpouse = false,
```
Rikkiabb, line 17: and here.

apps/financial-aid/web-osk/src/routes/application/Confirmation/confirmation.tsx
```diff
@@ -22,7 +22,9 @@ import { AppContext } from '@island.is/financial-aid-web/osk/src/components/AppP
 const Confirmation = () => {
   const router = useRouter()
   const { form } = useContext(FormContext)
-  const { municipality } = useContext(AppContext)
+  const { municipality, user } = useContext(AppContext)
+
+  const applicationId = form.applicationId || user?.currentApplication
```
Rikkiabb, line 27: currentApplication is now an id, right? Wondering if we should rename it to `currentApplicationId` ?

apps/financial-aid/web-osk/src/routes/status/MainPage/mainPage.tsx
```diff
@@ -26,29 +26,49 @@ import { AppContext } from '@island.is/financial-aid-web/osk/src/components/AppP
...
+  const {
+    myApplication,
+    loading,
+    error,
+    municipality,
+    setMunicipality,
+    user,
+  } = useContext(AppContext)
+
+  const isUserSpouse = user?.isSpouse?.HasApplied
```
Rikkiabb, line 38: Is this check correct?

Rikkiabb (changes requested):
📛

apps/financial-aid/api/src/app/modules/user/user.model.ts
MargretFinnboga: The user interface already has a spouse object, so maybe isSpouse should stay since it is object of two booleans? Or another name i am open to it 👐

apps/financial-aid/backend/src/app/guards/application.guard.ts
MargretFinnboga: No because it is !==, we are checking if application national id is not the same as the user and not the same as the spouse, then return forbidden

apps/financial-aid/web-osk/src/routes/status/MainPage/mainPage.tsx
MargretFinnboga: Yess, would you rather check if spouse has files, or maybe both, to be thorough?

apps/financial-aid/api/src/app/modules/user/user.model.ts
Rikkiabb: What about extending that object? So it will be:

`export interface Spouse {
  nationalId?: string
  name?: string
  maritalStatus?: string
  email?: string
  hasUploadedFiles?: boolean
}`

And then the user object has a boolean which is `isSpouse`

apps/financial-aid/web-osk/src/routes/status/MainPage/mainPage.tsx
Rikkiabb: No, just the naming confused me because I read this is has spouse applied and then it is extracted into a variable called isUserSpouse 🙈 
Which is bascially why I would rather go the way mentioned above 😄

apps/financial-aid/api/src/app/modules/user/user.resolver.ts
```diff
@@ -37,23 +38,22 @@ export class UserResolver {
     return user as UserModel
   }
 
-  @ResolveField('isSpouse', () => Boolean)
+  @ResolveField('isSpouse', () => SpouseModel)
```
Rikkiabb, line 41: Shouldn't this be 'spouse' ?

apps/financial-aid/api/src/app/modules/user/user.resolver.ts
```diff
@@ -37,23 +38,22 @@ export class UserResolver {
...
   async isSpouse(
```
Rikkiabb, line 42: And here, shouldn't this be spouse?

apps/financial-aid/api/src/app/modules/user/spouseModel.model.ts
```diff
@@ -0,0 +1,11 @@
+import { Spouse } from '@island.is/financial-aid/shared/lib'
+import { Field, ObjectType } from '@nestjs/graphql'
+
+@ObjectType()
+export class SpouseModel implements Spouse {
+  @Field()
+  readonly hasApplied!: boolean
```
Rikkiabb, line 7: Should this be hasPartnerApplied?

apps/financial-aid/api/src/services/backend.ts
```diff
@@ -88,11 +88,11 @@ class BackendAPI extends RESTDataSource {
     return this.post('file', createApplicationFiles)
   }
 
-  getCurrentApplication(nationalId: string): Promise<string | undefined> {
+  getCurrentApplicationId(nationalId: string): Promise<string | undefined> {
     return this.get(`application/nationalId/${nationalId}`)
   }
 
-  isSpouse(spouseNationalId: string): Promise<HasSpouseAppliedModel> {
+  isSpouse(spouseNationalId: string): Promise<SpouseModel> {
```
Rikkiabb, line 95: Should this also be spouse?

apps/financial-aid/api/src/app/modules/user/spouseModel.model.ts
MargretFinnboga: -_-

Rikkiabb (approved):
🎉

PR #5465, opened 2021-11-03
fix(api): Fixing dev url for auth-public-api

saevarma:
## What

Changing the `auth-public-api` to be operated on `/api` path, so configuring the `api` infra to update url.

## Why

So the service doesn't care about any prefixes and open for the `/swagger` url to expose the OpenAPI document of the service.

## Checklist:


apps/api/infra/api.ts
```diff
@@ -40,7 +40,7 @@ export const serviceSetup = (services: {
         prod: 'island-is-prod-upload-api',
       },
       AUTH_PUBLIC_API_URL: {
-        dev: 'https://identity-server.dev01.devland.is',
+        dev: 'https://identity-server.dev01.devland.is/public',
```
sindrig, line 43: why only for dev and didn't you forget to run `yarn charts`?

saevarma: Yes I forgot to run yarn charts.

I'm testing the ingress rewrite-target rule on dev by manually configuring in k9s.
Then the plan is to change this path to `/api` and do the corresponding changes in the Helm repo for the identity-server apis.

sindrig: ah right so this is work in progress. you know you can create draft pull requests, and then there will be no reviewers. kind of nice when you're playing a :ninja:

pshomov (approved):
LGTM

PR #5467, opened 2021-11-03
fix(dsl): Add support for files in dsl

sindrig:
This closes island-is/infrastructure#923

### DIFF
--- a/apps/api/infra/api.ts
+++ b/apps/api/infra/api.ts
@@ -161,6 +161,7 @@ export const serviceSetup = (services: {
       Education,
       NationalRegistry,
     )
+    .files('islyklar.p12')
     .ingress({
       primary: {
         host: {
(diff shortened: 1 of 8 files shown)


apps/api/infra/api.ts
```diff
@@ -161,6 +161,7 @@ export const serviceSetup = (services: {
       Education,
       NationalRegistry,
     )
+    .files('islyklar.p12')
```
pshomov, line 164: They usually need an env variable with the path to the file similar to [this](https://github.com/island-is/helm/blob/main/charts/identity-server/values.yaml#L30). Make this a hash where the key is the env variable name and the file is the value or the other way around 🙃

sindrig: fair enough

sindrig: done

pshomov (approved):
LGTM

PR #5471, opened 2021-11-03
fix(endorsement-system-api): Fixing small bugs in endorsement system api

alexandradilja:
- OwnerName was not querying and needed try catch
- Could not get single list if unauthorised, national id now optional
- endorsement scope stopped working due to function missing scope but checking access groups and one resolver function did not use auth  when service layer needed it.

## Checklist:


kksteini (approved):
LGTM

kksteini (approved):
Also LGTM

PR #5482, opened 2021-11-04
fix(fa): Orders staff list

Rikkiabb:
## What
Orders staff list by active and name
Applies different color to inactive users

## Checklist:


apps/financial-aid/backend/src/app/modules/staff/staff.service.ts
```diff
@@ -33,6 +34,9 @@ export class StaffService {
       where: {
         municipalityId,
       },
+      order: Sequelize.literal(
+        'CASE WHEN active = true THEN 0 ELSE 1 END, name ASC',
```
MargretFinnboga, line 38: You can also do   order: [['active', 'DESC']] I think 💭

MargretFinnboga (commented):
Suggestion 🎉

apps/financial-aid/backend/src/app/modules/staff/staff.service.ts
Rikkiabb: Didn't know that, but do you think it is more readable?

MargretFinnboga (approved):
Slides 🛷

PR #5487, opened 2021-11-04
fix(fa): Fixing spouse information in veita and uploading spouse name

MargretFinnboga:
# ...

https://app.asana.com/0/1201301944561348/1201213972251012

## What

Fixing in veita retreive spouse email and phone number 
When spouse logs in add to application spouse name 

## Why

For the staff to be able to know more information about spouse

## Checklist:


apps/financial-aid/web-osk/src/utils/hooks/useUpdateApplication.ts
```diff
@@ -38,6 +41,7 @@ const useUpdateApplication = () => {
             event: ApplicationEventType.FILEUPLOAD,
             spouseEmail: form.emailAddress,
             spousePhoneNumber: form.phoneNumber,
+            spouseName: user?.name,
```
Rikkiabb, line 44: When we create application shouldn't we use the `data` from national registry and null coalise it? Then we have this sometime before the spouse uploads data

Rikkiabb (approved):
📛

PR #5497, opened 2021-11-04
fix(general-petitions): adding loading dots while loading petition view + few minor tweaks for admin view

albinagu:
## Checklist:


alexandradilja (approved):

libs/service-portal/endorsements/src/screens/ViewPetition/ViewPetition.tsx
```diff
@@ -141,231 +147,254 @@ const ViewPetition = () => {
...
-          <Box>
-            <Text variant="h4">
-              {formatMessage(m.viewPetition.numberSigned)}
-            </Text>
-            <Text variant="default" marginBottom={3}>
-              {
-                (petitionEndorsements as PaginatedEndorsementResponse)
-                  .totalCount
-              }
+      {Object.entries(petition).length !== 0 ? (
```
arnorhs, line 150: This has become a very deeply nested component. Would it be a good idea to take some parts and create some non-exported smaller components in the same file?

albinagu: Yes, it would! I will followup on this

PR #5498, opened 2021-11-04
fix(auth): Remove user from session storage when signed out from IDS

eirikurn:
## What

Remove user from session storage when signed out from IDS

## Why

This reduces chances of a person seeing a flash of logged-in state if they revisit our systems after signing out.

There are two cases this addresses:

1. User is signed in. to our webapp in an open tab and the IDS session ends somehow (expired or signed out in another tab). The open tab is immediately pinged that the user has signed out and it redirects to the signin screen. But that tab might still have an active JWT. If the user manually navigates back, it'll look like they're still signed in for a split second before they're redirected back to the signin screen. This can be repeated until the JWT expires.

2. Same as case 1 but in this case the user is signed in with an active JWT but the tab is not open.

With this change, case 1 will not happen anymore, since we'll remove the JWT before going to the signin screen for the first time. With case 2, the user may see a signed in view exactly once, after which the JWT is removed and they're sent to the signin screen.

## Checklist:


saevarma (approved):
LGTM

PR #5503, opened 2021-11-05
fix(infra-nest-server): Fixing SwaggerUI trailing slash redirect

saevarma:
## What

Adding handler for SwaggerUI trailing slash redirect.

## Why

If API is using Ingress rewrite-target to add a path prefix the SwaggerUI breaks if the user
navigates to the swagger path without trailing slash. This is due to express.static making a redirect
with trailing slash to server static assets. So our infra-nest-server handles this if the service has
`PUBLIC_URL´ environment variable set, it redirects to the correct path with trailing slash.

## Screenshots / Gifs

## Checklist:


sindrig (approved):
This is starting to look like it'll create more problems than it was supposed to solve :sweat_smile: 
LGTM tho

PR #5514, opened 2021-11-05
feat(judicial-system): Get police case file for original ancestor

gudjong:
# Get police case file for original ancestor

https://app.asana.com/0/1199153462262248/1201185512430557

## What

☝️ 

## Why

Required

## Checklist:


apps/judicial-system/backend/src/app/modules/case/interceptors/test/caseOriginalAncestorInterceptor.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+
+  describe('next call', () => {
+    const theCase = {} as Case
+    const request = { case: theCase }
+    const result = {}
+    let then: Then
+
+    beforeEach(async () => {
+      mockRequest.mockImplementationOnce(() => request)
+      // mockFindOriginalAncestor.mockResolvedValueOnce(theCase)
```
oddsson, line 67: Consider removing this unused code

apps/judicial-system/backend/src/app/modules/case/guards/caseCompleted.guard.ts
```diff
@@ -12,7 +12,7 @@ import { Case } from '../models'
 
 @Injectable()
 export class CaseCompletedGuard implements CanActivate {
-  async canActivate(context: ExecutionContext): Promise<boolean> {
+  canActivate(context: ExecutionContext): boolean {
```
oddsson, line 15: I see that you are removing async in a couple of places but you are not removing any awaits. Was this function never being awaited on?

oddsson (changes requested):
Some minor things

apps/judicial-system/backend/src/app/modules/case/guards/caseCompleted.guard.ts
gudjong: This function is called as part of middleware execution. The framework uses awaits as neede.

oddsson (approved):

PR #5519, opened 2021-11-08
fix(fa): Css fix in veita and enum update in osk

MargretFinnboga:
# ...

https://app.asana.com/0/1201301944561348/1201337481363652
https://app.asana.com/0/1201301944561348/1201075565856649

## What

Removing overflow hidden in veita 
Updating family status enum removing and adding another value
Updating Summary page adding phone number beside the email, instead in user info

## Why

overflow was unnecessary and blocking styles 
More clarity

## Checklist:


apps/financial-aid/web-osk/src/components/Summary/FormInfo.tsx
```diff
@@ -21,6 +25,49 @@ const FormInfo = ({ info, error }: Props) => {
...
+                  <Text>{item.secInfo}</Text>
+                </Box>
+
+                <Box>
+                  <Button
+                    icon="pencil"
+                    iconType="filled"
+                    variant="utility"
+                    onClick={() => {
+                      router.push('samskipti')
```
Rikkiabb, line 61: Should this route be in our Route constant?

Rikkiabb (changes requested):
One comment and one question: Have you tried running Veita after these changes?

apps/financial-aid/web-osk/src/components/Summary/FormInfo.tsx
```diff
@@ -21,6 +25,49 @@ const FormInfo = ({ info, error }: Props) => {
     <>
       {info.map((item, index) => {
         const err = error && item.info === undefined
+        if (item.secLabel) {
```
Rikkiabb, line 28: Should this be a ternarny instead of a double return on line 28 and line 71

Rikkiabb: And what are the differences between these returns?

apps/financial-aid/web-osk/src/routes/application/SummaryForm/summaryForm.tsx
```diff
@@ -53,13 +54,13 @@ const SummaryForm = () => {
     {
       id: 'familyStatus',
       label: 'Hjúskaparstaða',
-      url: 'hjuskaparstada',
+      url: Routes.form.relationShip,
```
Rikkiabb, line 57: relationship is a single word and should hence not be camel cased.

apps/financial-aid/web-osk/src/components/Summary/FormInfo.tsx
Rikkiabb: Could I also get a screenshot of the UI when secLabel is defined 🙏

libs/financial-aid/shared/src/lib/const.ts
```diff
@@ -1,6 +1,18 @@
...
+    info: `${formRoutes}rettur`,
+    relationShip: `${formRoutes}hjuskaparstada`,
+    homeCircumstances: `${formRoutes}buseta`,
+    hasIncome: `${formRoutes}tekjur`,
+    employment: `${formRoutes}atvinna`,
+    usePersonalTaxCredit: `${formRoutes}personuafslattur`,
+    bankInfo: `${formRoutes}bankaupplysingar`,
+    contactInfo: `${formRoutes}samskipti`,
+    conformation: `${formRoutes}stadfesting`,
+  },
```
Rikkiabb, line 15: This is nice, we should maybe think about doing something similiar for other parts of our application

Rikkiabb (commented):
Some questions 😄

apps/financial-aid/web-osk/src/routes/application/SummaryForm/spouseSummary.tsx
```diff
@@ -42,29 +48,37 @@ const SpouseSummary = () => {
...
       id: 'hasIncome',
       label: 'Tekjur',
-      url: 'tekjur',
+      url: Routes.form.hasIncome,
       info:
         form?.hasIncome === undefined
           ? undefined
           : 'Ég hef ' +
-            (form.hasIncome ? '' : 'ekki') +
+            (form.hasIncome ? 'ekki ' : '') +
```
Rikkiabb, line 56: This is probably not correct......

libs/financial-aid/shared/src/lib/const.ts
```diff
@@ -1,6 +1,18 @@
+export const formRoutes = '/umsokn/'
```
Rikkiabb, line 1: This doesn't need to be exported since it is only a helper in the Routes obj

libs/financial-aid/shared/src/lib/const.ts
```diff
@@ -48,3 +60,10 @@ export const getNextPeriod = {
 }
 
 export const apiBasePath = 'api/financial-aid'
+
+export const scrollToId = (id: string) => {
```
Rikkiabb, line 64: I know we should split better into to helper files, but does this belong in `utils` instead of const?

Rikkiabb (changes requested):

Rikkiabb (approved):
☕  & 🚬

PR #5532, opened 2021-11-08
fix(dev): New approach on get secrets

sindrig:
This closes island-is/infrastructure#933


infra/src/cli/render-secrets.ts
```diff
@@ -0,0 +1,80 @@
...
+  services.forEach((svc) => {
+    Object.entries(svc).forEach(([serviceName, config]) => {
+      if (serviceName == service) {
+        Object.entries(config.secrets).forEach(([envName, ssmName]) => {
+          secretMap[ssmName] = envName
+        })
+      }
+    })
+  })
+  // There seem to be only 1 parameter that's labeled as dev?
```
sindrig, line 32: @darrikonn I seem to remember you had something to do about this script at some point. Any idea why there is only 1 parameter labeled as dev now? Can we/should we simply drop that requirement?

darrikonn: Yeah you're right. I wrote at least the first version of it.

There should be definitely more than 1 secret in dev labeled as dev.
Only 1 version had the label though. Might be that is what you're referring to?

sindrig: I'm mixing tags and labels :facepalm: 
Reworking this....

sindrig: Fixed now

infra/src/cli/render-secrets.ts
```diff
@@ -2,79 +2,59 @@ import { generateYamlForEnv } from '../dsl/serialize-to-yaml'
...
-      throw new Error('params.Parameters aws undefined?')
-    }
-    return
-  }
 }
 
-const getAllDevParamsNames = async (): Promise<string[]> => {
-  const client = new ResourceGroupsTaggingAPI(API_INITIALIZATION_OPTIONS)
-  const names = []
+const getAllDevParams = async (): Promise<{ [name: string]: string }> => {
```
darrikonn, line 38: Aren't there multiple environment variables that have the same name but for different projects (where the value might differ)?

sindrig: This will return a map of `ssm-parameter-name: ssm-parameter-value`. SSM parameter names are unique.

darrikonn: Yeah but then get-secrets will strip the environment variable name and load it using dotenv so they'll override each other if the final part of the aws param store variable matches.

sindrig: Final part? What do you mean?

Let's say service A wants to map `/k8s/serviceA/SOME_VAR` to the environment variable `SOME_VAR` and you run `get-secrets serviceA`.

- serviceB wants to map `/k8s/serviceB/SOME_VAR` to the environment variable `NOT_SOME_VAR`. That won't override `SOME_VAR` as `NOT_SOME_VAR != SOME_VAR`
- serviceC wants to map `/k8s/serviceC/NOT_SOME_VAR` to the environment variable `SOME_VAR`. That WILL override `SOME_VAR` in `.env.secrets`, but that's exactly how the script behaved before so I'm not addressing that.

darrikonn: Ahh sorry... I somehow assumed that you removed the service argument and you're just fetching every single environment variable, regardless of service.

darrikonn (approved):

PR #5539, opened 2021-11-09
feat(judicial-system): Display court case number in active case requests

telmag92:
# Display court case number in active case requests
https://app.asana.com/0/1199153462262248/1201289046861268/f

## What

Display court case number in active requests only if active case has assigned court case number

## Why

If a case has an assigned court case number that becomes the main reference number rather than the police number. 

## Screenshots / Gifs
Screenshot 2021-11-09 at 11 52 22

## Checklist:


apps/judicial-system/web/src/routes/Shared/Requests/ActiveRequests.tsx
```diff
@@ -165,12 +165,23 @@ const ActiveRequests: React.FC<Props> = (props) => {
               user?.role && onRowClick(c.id)
             }}
           >
-            <td className={styles.td}>
-              <Text as="span">{c.policeCaseNumber || '-'}</Text>
-            </td>
+            {c.courtCaseNumber ? (
```
oddsson, line 168: Both the if and the else use 

`<td className={styles.td}>`

so I think it would be better to place the if/else statement inside the <td> tag.

oddsson (changes requested):
I think there is a small issue with where the if/else statement is placed.

oddsson (approved):
Good job 🙌

PR #5548, opened 2021-11-09
fix(application-template-api-modules): pass type into makeGraphqlQuery

arnorhs:
Allowing types to be passed into the graphql query in the shared api service in the application system.

Allows us to skip a bunch of optimistic typecasts and duplicate code.

Unmodified code should work the same as before.

## Checklist:


alexandradilja (approved):
lgtm

eirikurn (approved):
LGTM

PR #5549, opened 2021-11-09
feat(judicial-system): Consistently masks names to max two characters

gudjong:
# Consistently masks names to max two characters

https://app.asana.com/0/1199153462262248/1157518359844597/f

## What

☝️ 

## Why

So that masked names can be distinguished from one another.

## Checklist:


telmag92 (approved):
Looks good and works as expected

PR #5558, opened 2021-11-10
fix(infra-nest-server): Fix SwaggerUI redirect loop

saevarma:
## What

Fixing path matching for SwaggerUI path.

## Why

Previous usage of `.baseUrl` strips away the trailing slash, so the match is always true, causing the too many redirects loop.

## Screenshots / Gifs

## Checklist:


eirikurn (approved):

sindrig (approved):
LGTM

PR #5565, opened 2021-11-10
chore(judicial-system): Remove pdf from court ready email to prosecutor

telmag92:
# Remove pdf from court ready email to prosecutor

https://app.asana.com/0/1199153462262248/1201312869487925/f

## What

* Remove pdf from court ready email to prosecutor
* Add more descriptive text to email body
* Add all strings to contentful

## Why

Security exposure

## Screenshots / Gifs
Before:
Screenshot 2021-11-10 at 11 36 41
After:
Screenshot 2021-11-10 at 11 44 13

## Checklist:


apps/judicial-system/backend/src/app/messages/requestPdf.ts
```diff
@@ -1,27 +1,6 @@
 import { defineMessage, defineMessages } from '@formatjs/intl'
 
 export const restrictionRequest = {
-  caseType: defineMessages({
```
gudjong, line 4: Isn't it dangerous to delete these strings in this PR? What happens when the changes are pushed to CF? Will that break prod until a new release is deployed?

I don't have the answer to this question. If deleting the strings poses a problem, then you may consider creating a new cleanup task that can be worked on after this PR is on prod.

gudjong (changes requested):
Nice, just some questions and considerations.

apps/judicial-system/backend/src/app/messages/requestPdf.ts
telmag92: Yes you are right and good point, I will do a separate task for removing these strings.

gudjong (approved):

PR #5583, opened 2021-11-11
feat(infra): Use test containers for running tests

dabbeg:
Issue: https://app.zenhub.com/workspaces/2021-q4-6144bf40d43cf9001036abef/issues/island-is/infrastructure/921

## What

Stop using docker-compose to run external services for tests, instead manage the containers with [testcontainers](https://www.npmjs.com/package/testcontainers) library.

## Why

To reduce boilerplate in services, simplify the dev experience and last but not least save some precious ci time.

## Checklist:


apps/reference-backend/docker-compose.yml
```diff
@@ -2,7 +2,7 @@ version: '3.3'
 
 services:
   db_reference_backend:
-    image: public.ecr.aws/bitnami/postgresql:11.12.0
+    image: postgres:11.12-alpine
```
sindrig, line 5: this might cause us problems with pull limits from dockerhub

scripts/ci/30_test.sh
```diff
@@ -25,11 +25,40 @@ if [ -f "$PROJECT_ROOT"/"$APP_HOME"/docker-compose.ci.yml ]; then
...
+  # Running postgres
+  docker run \
+    --rm \
+    --detach \
+    --publish 5432 \
+    --name "$CONTAINER_NAME" \
+    --env POSTGRES_NAME=test_db \
+    --env POSTGRES_USER=test_db \
+    --env POSTGRES_PASSWORD=test_db \
+    "postgres:11.12-alpine"
```
sindrig, line 55: ditto

apps/reference-backend/docker-compose.yml
dabbeg: Reverted back to bitnami

sindrig (approved):
LGTM

Rikkiabb (approved):
👌

darrikonn (changes requested):
Awesome work!
Nice improvements.

I believe there's one thing missing. If you take a look at the new test setup for Nest.js apps (`libs/testing/nest`), it exposes couple of hooks. One noticeable hook (which relates to your PR) is the `useDatabase` hook.
Can you add a support for testcontainers to that hook as well?
This would be in the `extend` section of the default export. It could call your `startPostgres` function and `stopPostres` in the teardown 🙏

dabbeg:
> Awesome work! Nice improvements.
> 
> I believe there's one thing missing. If you take a look at the new test setup for Nest.js apps (`libs/testing/nest`), it exposes couple of hooks. One noticeable hook (which relates to your PR) is the `useDatabase` hook. Can you add a support for testcontainers to that hook as well? This would be in the `extend` section of the default export. It could call your `startPostgres` function and `stopPostres` in the teardown pray

Thanks :)

Nice! I wasn't aware of this library, it would be nice to refactor the other services to use it so its existence is more visible (This PR is big enough, so I will leave that for some other lucky person).

But it seems the `useDatabase` hook is run for every spec file, that would mean postgres container for every spec file, so I think this will need to be separate. The `startPostgres` and `stopPostgres` are called from jests globalSetup and globalTeardown functions which run only once for all spec files.

darrikonn:
"that would mean postgres container for every spec file"
Yeah, this was previously raised and afaicr @pshomov was keen on that.

Petar, can you give your input on this as well?

libs/testing/containers/src/lib/testing-containers.ts
```diff
@@ -0,0 +1,39 @@
+import { GenericContainer, StartedTestContainer, Wait } from 'testcontainers'
+
+declare global {
+  // eslint-disable-next-line @typescript-eslint/no-namespace
+  namespace NodeJS {
+    interface Global {
+      __CONTAINER_POSTGRES__: StartedTestContainer
```
pshomov, line 7: Please enlighten me, what is this beauty for? 🧐

pshomov (approved):
Love it.

pshomov:
> "that would mean postgres container for every spec file" Yeah, this was previously raised and afaicr @pshomov was keen on that.
> 
> Petar, can you give your input on this as well?

I mean we are transitioning to using this tech which I believe has better DX than the docker-compose setup we had this far. I think we can take this further in some next step to start experimenting a bit more with not having this with the global* approach but I do agree that this PR goes far enough.

libs/testing/containers/src/lib/testing-containers.ts
dabbeg: To be able to access the container created in globalSetup in the globalTeardown function, I save a reference to the container in the global scope. 
[https://jestjs.io/docs/configuration#globalsetup-string](https://jestjs.io/docs/configuration#globalsetup-string)

dabbeg: As I write this I realize it is unnecessary to have it in the global scope. I can just place it at the top of this file as the variable is only used in this file.

arnorhs (approved):
LGTM

Snojo (approved):

apps/services/user-profile/docker-compose.yml
```diff
@@ -0,0 +1,17 @@
...
+    image: public.ecr.aws/bitnami/postgresql:11.12.0
+    container_name: db_user_profile
+    networks:
+      - local
+    environment:
+      - POSTGRES_DB=dev_db
+      - POSTGRES_USER=dev_db
+      - POSTGRES_PASSWORD=dev_db
+    ports:
+      - 5437:5432
```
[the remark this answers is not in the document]
dabbeg: Yes

valurefugl (approved):

darrikonn (approved):
LGTM

kksteini (approved):

PR #5588, opened 2021-11-11
feat(service-endorsement-api): Helper function checks if user has admin scope to access locked Lists 

alexandradilja:
Helper function in the service layer, checks if user has admin scope. With admin scope user can view adminLocked Lists.
Get lists by tags and get single lists endpoints use this helper function to determine if they should query after locked lists as well. Get single List is used by both endorsementListById and access guard, therefor those functions send the authenticated users into the function.

## Checklist:


rafnarnason (approved):
LGTM

kksteini (approved):
LGTM

thordurhhh (approved):
Service-portal LGTM.

PR #5594, opened 2021-11-11
feat(service-portal): Display finance pdf's using the download-service

thordurhhh:
## What

Use the download service for displaying finance documents in the service portal.

## Why

To display documents in the same manner as we do for the inbox documents. 
And the current way of displaying, could prompt the popup blocker to prevent the display of documents without the user opting in.

## Checklist:


disaerna (approved):
LGTM 💯

pshomov (approved):
LGTM

PR #5608, opened 2021-11-12
fix(clients-driving-license): using v1 path for v2 for now

arnorhs:
So apparently the v2 version of the service doesn't exist yet, and only two versions of v1.. but since it's actually still the same instance of the underlying service, continuing to call the v1 api with a v2 api header seems to work just as well.

When this gets resolved at RLS's end, I will update this config again

## Checklist:


apps/application-system/api/src/environments/environment.ts
```diff
@@ -186,7 +186,7 @@ const prodConfig = {
         xroadClientId: process.env.XROAD_CLIENT_ID,
         xroadBaseUrl: process.env.XROAD_BASE_PATH,
         xroadPathV1: process.env.XROAD_DRIVING_LICENSE_PATH,
-        xroadPathV2: process.env.XROAD_DRIVING_LICENSE_V2_PATH,
+        xroadPathV2: process.env.XROAD_DRIVING_LICENSE_PATH,
```
eirikurn, line 189: Should we fix this in helm (environment variables) rather than in code?

charts/islandis/values.dev.yaml
```diff
@@ -189,6 +189,7 @@ api:
     ISLYKILL_CERT: '/etc/config/islyklar.p12'
     NODE_OPTIONS: '--max-old-space-size=464'
     PARTY_LETTER_REGISTRY_BASE_API_URL: 'http://web-party-letter-registry-api.party-letter-registry.svc.cluster.local'
+    PAYMENT_SCHEDULE_XROAD_PROVIDER_ID: 'IS-DEV/GOV/10021/FJS-Public'
```
eirikurn, line 192: Where does all of these value changes come from? Can't see corresponding dsl changes. Also, should these start with XROAD_ to match sindri's recent naming convention?

apps/application-system/api/src/environments/environment.ts
arnorhs: probably a good idea

charts/islandis/values.dev.yaml
arnorhs: probably, i didn't even notice that those had changed. I thought we had a lint task for that

PR #5609, opened 2021-11-12
feat(judicial-system): Adds investigation type restraining order

gudjong:
# Adds investigation type restraining order

https://app.asana.com/0/1199153462262248/1201317701073314

## What

Adds restraining order as an investigation case type. This results in:

- A new case type in the database.
- A corresponding new value in the case type enum.
- The possibility to select restraining order when creating an investigation case in the client.

## Why

This is a useful feature because restraining orders are common.

## Screenshots / Gifs

image
image

## Checklist:


oddsson (approved):
Awesome!

PR #5610, opened 2021-11-12
feat(parental-leave): Update rights code with gender information

baering:
## What

For the other parent we've sent `FO` (`FO`reldri -> parent). This PR changes it so if the other parent is a male then we sent `F` instead of `FO`.

Also refactored Review screen to send users directly to screens which have side effects when answers are changed.

## Why

For Fæðingarolofssjóður statistical analysis.

### Reviewing full name before submitting application:
image

### Cannot edit full name:
image

### Reviewing full name after submitting application:
image

## Checklist:


Rikkiabb (approved):
Nice description and nice to see tests 👌👌👌

libs/application/templates/parental-leave/src/lib/parentalLeaveUtils.ts
```diff
@@ -625,6 +608,34 @@ export const getOtherParentId = (application: Application): string | null => {
...
+
+  return otherParentName
+}
+
+export const applicantIsMale = (application: Application): boolean => {
+  const { applicantGenderCode } = getApplicationExternalData(
+    application.externalData,
+  )
+
+  return applicantGenderCode === '1'
```
eirikurn, line 636: Should the client library have enums or constants for this?

eirikurn (approved):

PR #5612, opened 2021-11-12
feat(core): New library: @island.is/nest/config

eirikurn:
## What

Created helpers for Nest's ConfigModule which includes these features:

* Support required environment variables with fallbacks for local development.
* Support optional environment variables with default values.
* Support "not configured" modules with ServerSideFeatureFlags.
* Support "not configured" modules in local development.
* Simple number and JSON parsing.
* Validate Configuration using Zod.
* Friendly errors for better DX.

[RFC](https://www.notion.so/Config-Module-8c065e2983e6411389228baa149dc6f0)
[Docs](https://github.com/island-is/island.is/blob/3c5b1f106e659af7424ac0132dbb2a3d47877080/libs/nest/config/README.md)

## Why

The big goal is to move configuration to the module that needs the configuration, instead of configuring everything from the root app's `environment.ts`. This scales much better when using Nest's dependency injection system to instantiate large trees of modules, each of which needs configuration.

As an example, let's say we want our national registry client to use a theoretical IdentityServerModule to perform OAuth2 token exchange. The IdentityServerModule needs to know the address of the identity server, but to configure it, we need to pipe it down through all the modules from the app, eg:

> AppModule -> IdentityDomain -> NationalRegistryXRoadModule -> NationalRegistryClientModule -> IdentityServerModule

All apps that use this IdentityServerModule need to add code to read its configuration from the environment in development and production, which is error prone. If the module is imported multiple times, we end up with multiple instances of each service, which can cause subtle caching issues.

Another goal of this change is to properly validate that all environment variables are set correctly in production while still supporting fallback values and partial configuration for local development.

## Checklist:


libs/nest/config/src/lib/ConfigurationLoader.ts
```diff
@@ -0,0 +1,162 @@
...
+
+    // Only throw missing issue in production
+    if (hasMissingIssue && !this.isDev) {
+      throw new ConfigurationError(errorMessage)
+    }
+
+    // Re-throw other unexpected errors from loader.
+    if (error) {
+      throw error
+    }
```
arnorhs, line 62: there's a bunch of code around handling the issue and issue types both here and in `formatIssues()` etc.. i wonder if the issue thing would be more clear if it had its own class definition

eg.
```
class Issues {
  issues: Record<string, Issue> = {}
  hasMissingIssue() {}
  hasParseIssue() {}
  formatIssues () {}
}
```

might also read a lot clearer

libs/nest/config/src/lib/InvalidConfiguration.ts
```diff
@@ -0,0 +1,20 @@
...
+  constructor(errorMessage: string) {
+    return new Proxy(this, {
+      get(target, prop) {
+        if (
+          prop in target ||
+          InvalidConfiguration.allowedMembers.includes(String(prop))
+        ) {
+          return (target as any)[prop]
+        }
+        throw new ConfigurationError(errorMessage)
```
arnorhs, line 16: since this gets thrown when you access the property, is the configuration error clear enough, that the stack trace wouldn't end up seemingly pretty confusing?

libs/nest/config/README.md
```diff
@@ -0,0 +1,232 @@
...
+import { ConfigType } from '@island.is/nest/config'
+import { someModuleConfig } from './someModule.config'
+
+export class SomeModuleService {
+  constructor(
+    @Inject(someModuleConfig.KEY)
+    private config: ConfigType<typeof someModuleConfig>,
+  ) {
+    config.isConfigured // boolean
+    config.data // {url: string, secret: string, ttl: any} | undefined
```
arnorhs, line 67: is this from an older version? currently it wouldn't have data unless your config type had a data property, right?

libs/nest/config/src/lib/ConfigurationLoader.ts
```diff
@@ -0,0 +1,162 @@
...
+        `Failed loading configuration for ${this.definition.name}. Config was undefined.`,
+      )
+    }
+
+    if (this.isDev && hasMissingIssue) {
+      logger.warn({
+        message: `Could not load configuration for ${this.definition.name}. Missing ${issues.length} required environment variable(s).`,
+        category: 'ConfigModule',
+      })
+      return new InvalidConfiguration(errorMessage) as Configuration<T>
```
arnorhs, line 75: It seems like it would be simpler to throw, right away, but I have a feeling there is a particular use case you were addressing by throwing on property access, rather than at config time, I'd be curious to know what it is

arnorhs (commented):
I really like this API for configuring modules/services and the design is really good. I'm really glad that you guys are addressing the config issue in a generic way that makes it simpler and easier for everyone.

I have some minor comments/thoughts, but all in all this all LGTM

libs/nest/config/src/lib/ConfigurationLoader.ts
eirikurn: Right! It's to support partial configuration during local development, especially for the `api` and `application-system-api` projects.

When you're working on something, you should only need the dev-secrets (environment variables) and services for that thing on your computer. Otherwise, developers would need to have all of the secrets, for all of the modules, just to start these services (vmst, driver-license, health-insurance, etc).

I didn't want all of the required configuration attributes to be nullable to deal with this issue locally, as well as missing server-side features in production. Instead you can check `isConfigured`, or deal with the error at runtime. If you end up accessing an API that reads a missing configuration, you'll get a verbose error telling you precisely what 
the issue is (eg which environment variable which is missing or which server side feature is disabled). 

The only problem with this implementation is that in many cases, the configuration is read at server startup (eg when passed to OpenAPI classes or `createEnhancedFetch`). It could be tricky to delay this initialization until a request needs it.

libs/nest/config/README.md
eirikurn: Good catch. Will fix!

libs/nest/config/src/lib/InvalidConfiguration.ts
eirikurn: See comment below.

libs/nest/config/src/lib/ConfigurationLoader.ts
eirikurn: Good suggestion, will consider.

libs/nest/config/src/lib/ConfigurationLoader.ts
```diff
@@ -0,0 +1,162 @@
...
+import { ServerSideFeatureClient } from '@island.is/feature-flags'
+import { logger } from '@island.is/logging'
+import { InvalidConfiguration } from './InvalidConfiguration'
+import { ConfigurationError } from './ConfigurationError'
+import { ZodError } from 'zod'
+
+type Issue = 'missing' | 'could not be parsed as JSON'
+
+export class ConfigurationLoader<T> implements EnvLoader {
+  private isDev = process.env.NODE_ENV !== 'production'
```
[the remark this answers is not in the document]
eirikurn: Sounds good. 👍

eirikurn:
Thanks for the review @arnorhs!

libs/nest/config/src/lib/ConfigurationLoader.ts
```diff
@@ -0,0 +1,162 @@
...
+
+  required(envVariable: string, devFallback?: string): string {
+    const value = process.env[envVariable]
+    if (value !== undefined) {
+      return value
+    }
+    if (this.isDev && devFallback !== undefined) {
+      return devFallback
+    }
+    this.issues[envVariable] = 'missing'
```
saevarma, line 122: Should the known issues be a const object instead of "magic" strings?

```
const IssueReason = {
	MISSING: 'missing',
	PARSE_ERROR: 'could not be parsed as JSON',	
}
```

libs/nest/config/src/lib/ConfigurationLoader.ts
```diff
@@ -0,0 +1,162 @@
...
+    const value = process.env[envVariable]
+    if (value === undefined) {
+      return undefined
+    }
+
+    try {
+      return JSON.parse(value)
+    } catch (err) {
+      this.issues[envVariable] = 'could not be parsed as JSON'
+      return (undefined as unknown) as T
```
saevarma, line 159: This should just return `undefined` without typecasts

saevarma (approved):
Amazing work! Life will be so much simpler with this in it.
Just two minor comments.

libs/nest/config/src/lib/ConfigurationLoader.ts
eirikurn: Not really. This is to provide better TypeScript types for the configuration, and allow the loader to run as far as it can so we can list all issues.

PR #5614, opened 2021-11-14
chore(core): Upgrade postcss-svgo to remove is-svg dependency

eirikurn:
This fixes security vulnerability:

https://github.com/island-is/island.is/security/dependabot/yarn.lock/is-svg/open


saevarma (approved):
LGTM

PR #5615, opened 2021-11-14
chore(core): update slate and immer

eirikurn:
Upgrading immer to fix security vulnerability: https://github.com/island-is/island.is/security/dependabot/yarn.lock/immer/open

Needed resolution since react-dev-utils has not been published.
Was able to upgrade Slate which also depends on immer. It is currently only used in the markdown editor in our contentful extension. Verified that it works as expected after the upgrade.


saevarma (approved):
LGTM

darrikonn (approved):

PR #5616, opened 2021-11-14
chore(core): Update object-path

eirikurn:
Fixes security vulnerability: 
https://github.com/island-is/island.is/security/dependabot/yarn.lock/object-path/open


saevarma (approved):
LGTM

PR #5617, opened 2021-11-14
chore(core): Upgrade winston-cloudwatch and pac-resolver

eirikurn:
Fixes security vulnerability:

https://github.com/island-is/island.is/security/dependabot/yarn.lock/pac-resolver/open

The main change in winston-cloudwatch 3 is that it does not bundle its own version of aws-sdk anymore. We have our own peer version which should be compatible.


gudjong (approved):

saevarma (approved):
LGTM

darrikonn (approved):

PR #5634, opened 2021-11-16
fix(judicial-system): Fix null check for parentCase

telmag92:
# Fix null check for parentCase

## What

Handle null check for parent case in step four

## Why

## Screenshots / Gifs

## Checklist:


oddsson (changes requested):
The same mistake is being made in RulingStepTwo in restriction cases and in step four for prosecutor, also restriction cases. Could you please fix those as well?

oddsson (approved):

oddsson (approved):

PR #5645, opened 2021-11-16
chore(judicial-system): Formstepper fixes

oddsson:
# Formstepper fixes

https://app.asana.com/0/1199153462262248/1201353388542221/f

## What

Fixes two issues with the form stepper.

- When a case is extended, the form stepper would show "Niðurstaða" where the actual case result should be.

### Before

Screenshot 2021-11-16 at 15 08 01

### After

Screenshot 2021-11-16 at 15 07 07

- We would show case results based on the case decision instead of the case state. This is not ideal, since the following scenario could happen if we do this

👉 A prosecutor creates a case and sends it to the court
👉 The court selects ACCEPT the case but hasn't signed the verdict yet
👉 The prosecutor opens the case and sees "Gæsluvarðhald samþykkt" in the form stepper, even though the verdict has not been signed.

Now we only show case results base on case state.

## Why

Better UX

## Checklist:


apps/judicial-system/web/src/shared-components/PageLayout/utils/index.ts
```diff
@@ -0,0 +1,108 @@
...
+  translationStrings: TranslationStrings,
+  workingCase?: Case,
+) => {
+  if (!workingCase) {
+    return ''
+  }
+
+  const isRejected =
+    workingCase?.state === CaseState.REJECTED ||
+    workingCase?.parentCase?.state === CaseState.REJECTED
```
gudjong, line 31: I don't fully understand the logic here, but think it is strange to check if a parent case is rejected. Only accepted cases can be extended, so it follows that the parent case is always accepted.

gudjong (changes requested):
I do not have a specific change recommendation, but a comment to consider.

apps/judicial-system/web/src/shared-components/PageLayout/utils/index.ts
oddsson: You are absolutely right, thanks 🙏

gudjong (approved):

PR #5647, opened 2021-11-16
fix(public-debt-payment-plan): use schedule type keys instead of values

bensijons:
# Use schedule type keys instead of values

[ClickUp issue](https://app.clickup.com/t/1kr19xf)

## What

Since TBR uses the keys instead of values for schedule type we need to use the schedule types when submitting the applications instead of values. For example use "MR" instead of "Other fees".

## Why

Requested after testing

## Screenshots / Gifs

## Checklist:


libs/application/template-api-modules/src/lib/modules/templates/public-debt-payment-plan/public-debt-payment-plan.service.ts
```diff
@@ -59,35 +65,45 @@ export class PublicDebtPaymentPlanTemplateService {
...
+          organizationID: plan.organization,
+          chargeTypes: plan.chargetypes?.map((chargeType) => ({
+            chargeID: chargeType.id,
+          })),
+          payments: distribution.map((p) => ({
+            duedate: p.dueDate,
+            payment: p.payment,
+            accumulated: p.accumulated,
+          })),
+          type: ScheduleType[plan.id],
```
andriandresson, line 87: Is it Ok, if `ScheduleType[plan.id],` returns undefined ?

andriandresson (commented):
LGTM, just one concern.

libs/application/template-api-modules/src/lib/modules/templates/public-debt-payment-plan/public-debt-payment-plan.service.ts
bensijons: Yes since plan.id is actually of type PaymentScheduleType that maps to ScheduleType. If TBR (the external provider) adds a new schedule type we would have to update those types in our code. Plan.id comes from an external fetch to TBR's services, so plan.id should be of type PaymentScheduleType, otherwise something is wrong on their end.

andriandresson (approved):
LGTM 🔥

darrikonn (approved):

saevarma (approved):
LGTM

PR #5651, opened 2021-11-16
chore(judicial-system): Improves logging in court client

gudjong:
# Improves logging in court client

https://app.asana.com/0/1199153462262248/1201342867611338

## What

This is a major rewrite of error handling in the court service. Errors from the api are cryptic and distinguishing between different types of errors leads to brittle code. This PR is an attempt to make that distinction and at the same time tolerate changes to api error messages.

## Why

We were spamming the logs with excessive error messages.

## Checklist:


libs/judicial-system/court-client/src/lib/courtClient.service.ts
```diff
@@ -76,24 +80,48 @@ export class CourtClientService {
...
+        ) {
+          this.logger.warn(
+            'Error while calling the court service - attempting relogin',
+            { reason },
+          )
+
+          return this.wrappedRequest(clientId, request, true)
+        }
+
+        const { status, statusText } = reason as {
```
oddsson, line 104: I think this is a duplication of [this code](https://github.com/island-is/island.is/pull/5651/files#diff-30b1389d58ba77128bdb118fdd0fdc6213b60804863fc6b718dee745b6e39857R86). 

On line 86, consider writing 

```
const { statusCode, body, status, statusText } = reason as {
    statusCode: number
    body: string
    status: number
    statusText: string
}
``` 

instead. Also consider chaining the if statements together with `else if` on line 109 instead of if.

libs/judicial-system/court-client/src/lib/courtClient.service.ts
```diff
@@ -76,24 +80,48 @@ export class CourtClientService {
...
+      .catch((reason) => {
+        const { statusCode, body } = reason as {
+          statusCode: number
+          body: string
+        }
+
+        if (
+          !isRetry &&
+          statusCode === 400 &&
+          body?.startsWith('authenticationToken is expired')
```
oddsson, line 94: Making decisions based on whether a hard coded string is in the response body of a request seems fragile to me. Are we sure there is no better way to check for this scenario?

oddsson (changes requested):
A few things to consider ;)

libs/judicial-system/court-client/src/lib/courtClient.service.ts
```diff
@@ -51,53 +59,120 @@ export class CourtClientService {
     private readonly logger: Logger,
   ) {}
 
+  // The service has a 'logged in' state and at most one in progress
+  // login operation should be ongoing at any given time.
+  private loginPromise?: Promise<void>
+
+  // Detecting authentication token expioration is imperfect and brittle.
```
oddsson, line 66:
```suggestion
  // Detecting authentication token expiration is imperfect and brittle.
```

oddsson (approved):

PR #5653, opened 2021-11-17
chore(application-system-api): fix flaky tests in application-system-api

eirikurn:
Give jest a bit more time when running application-system tests. Jest lazily transforms dynamically imported files. In this case, we're using ts-jest (which is slow) and the application system uses dynamic imports for some large application templates. So the first time a test imports a large template, it stalls while jest starts transforming the template files. Note that jest has a cache for these transforms, so this is mainly an issue on a cold run.

Going forward, the plan is to replace ts-jest with babel-jest, esbuild-jest or swc-jest. We just need to get them working correctly with our typescript, decorators and type hints.

The previous fix from #5639 (passing --testTimeout during CI) doesn't work since the nx builder doesn't support this argument.

Additionally, this PR reduces verbose logging quite a bit throughout by disabling INTL warns and database debug logging while tests are running. Also updates browserlist.

## Checklist:


VilliGunn (approved):

sindrig (approved):
LGTM

saevarma (approved):
LGTM

PR #5655, opened 2021-11-17
fix(public-debt-payment-plan): check if wage deduction for alert

sigruntg:
# Check if wage deduction for alert message

[1kcjt49](https://app.clickup.com/t/1kcjt49)

## What

Added a check if wage deduction is not undefined, then we will show the alert message.

## Checklist:


bensijons (approved):
LGTM

PR #5660, opened 2021-11-17
fix(web): Make category accordion items expanded when hash exists in url

stjanilofts:
## What

Make category accordion items expanded when hash exists in url like it was before.

## Why

It was broken. The accordion item had the "plus" sign as if it was expanded but it was not.

## Checklist:


barabrian (approved):
LGTM

PR #5666, opened 2021-11-17
feat(cms): Indexable and searchable Links

kksteini:
Content independent websites (such as https://island.is/reglugerdir) are not indexable for the search engine due to their absense as a content type.

It is possible to create special Links for those and have them show up in the search engine.

To avoid indexing links which we do not want to be indexed there is a boolean field added to links, `searchable`, which dictates whether the elastic search importer will consider them or not.

## Screenshots / Gifs

image
contentful_searchable_link

searchable

## Checklist:


apps/web/screens/Search/Search.tsx
```diff
@@ -201,12 +206,18 @@ const Search: Screen<CategoryProps> = ({
       News &
       AdgerdirPage &
       SubArticle &
-      OrganizationSubpage
+      OrganizationSubpage &
```
kksteini, line 209: Imported with `Link as LinkItem` to avoid conflicts

apps/web/screens/Search/Search.tsx
```diff
@@ -201,12 +206,18 @@ const Search: Screen<CategoryProps> = ({
...
+      LinkItem
   >).map((item) => ({
     title: item.title,
     parentTitle: item.parent?.title,
     description: item.intro ?? item.description ?? item.parent?.intro,
-    link: linkResolver(item.__typename, item?.url ?? item.slug.split('/')),
+    link: linkResolver(
```
kksteini, line 215: Basically
```ruby
if typename == 'Link' then
    resolve as 'linkurl'
else
    resolve as either url, slug as applicable
```

apps/web/screens/queries/Search.tsx
```diff
@@ -245,6 +245,14 @@ export const GET_SEARCH_RESULTS_QUERY_DETAILED = gql`
           url
           intro
         }
+
+        ... on Link {
+          id
+          title: text
```
kksteini, line 251: TODO:
Change `url -> slug`, `description -> intro` for content model

libs/cms/src/lib/generated/contentfulTypes.d.ts
```diff
@@ -699,6 +699,12 @@ export interface IFrontpageFields {
   /** Title */
   title: string
 
+  /** Heading */
```
kksteini, line 702: Ran `yarn nx run api:contentful-types`.
This should be kept up to date if possible

libs/cms/src/lib/generated/contentfulTypes.d.ts
```diff
@@ -1123,6 +1135,15 @@ export interface ILinkFields {
 
   /** Link reference */
   linkReference?: IArticle | IArticleCategory | ILinkUrl | INews | undefined
+
+  /** Searchable */
+  searchable?: boolean | undefined
+
+  /** Search Description */
+  intro?: string | undefined
```
kksteini, line 1143: I'm not 100% whether I should name this intro or description.

If named description then in the graphql search query we need to do
```
...on Link {
  id
  intro: description
}
```

anyway.

siggifv (approved):
LGTM

apps/web/screens/Search/Search.tsx
```diff
@@ -192,6 +193,10 @@ const Search: Screen<CategoryProps> = ({
       }
     }
 
+    if (item?.labels && item.labels?.length) {
```
arnorhs, line 196: kannski?
```
if (item?.labels?.length) {
```

Annars er líka alltaf aðeins meira explicit að segja hreinlega
```
if (Array.isArray(item?.labels)) {
```

apps/web/screens/Search/Search.tsx
arnorhs: wouldn't it be better to leave graphql-type checking stuff outside of the component.. but looking at the rest of it, maybe we're a bit beyond that atm.

If I understand correctly, the item could be a link type or it could be a "normal link" .. and we want to handle both cases?

It would probably be nice then to make a helper for creating this link resolver.

libs/cms/src/lib/generated/contentfulTypes.d.ts
arnorhs: I think you are right naming it `intro` is the correct thing to do

arnorhs (changes requested):

apps/web/screens/Search/Search.tsx
kksteini: Yeah, actually. Maybe simply declaring the typename to be 'linkurl' in the importer and skipping this nonsense altogether.

Either way I should clean this up a bit

apps/web/screens/Search/Search.tsx
kksteini: Já vá. Fyndið að ég notaði optional chaining en ekki alla leið.
Takk fyrir að spotta þetta.

apps/web/screens/Search/Search.tsx
kksteini: Graphql doesn't like it when I lie to it :)
Moved special case to linkResolver similar to `linkurl`.

apps/web/screens/Search/Search.tsx
kksteini: Lagað

arnorhs (approved):

PR #5675, opened 2021-11-18
fix(github-actions-cache): Invalidate reservation

sindrig:
Invalidate reservation automatically after 5 minutes

This closes island-is/infrastructure#953

# ...

## What

## Why

## Screenshots / Gifs

## Checklist:

### DIFF
--- a/apps/github-actions-cache/src/app/handlers.ts
+++ b/apps/github-actions-cache/src/app/handlers.ts
@@ -34,6 +34,8 @@ export const reserveCache = async (
     if (reservation !== 1) {
       return error(res, 'cache id already reserved', 409)
     }
+    // Invalidate reservation after 300 seconds
+    await cache.expire(cacheId, 300)
     logger.debug(`Creating multipartupload for ${cacheId}`)
     const { UploadId } = await s3.createMultipartUpload(objectParams).promise()
     if (!UploadId) {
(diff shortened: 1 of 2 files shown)


apps/github-actions-cache/src/app/handlers.ts
```diff
@@ -34,6 +34,8 @@ export const reserveCache = async (
     if (reservation !== 1) {
       return error(res, 'cache id already reserved', 409)
     }
+    // Invalidate reservation after 300 seconds
+    await cache.expire(cacheId, 300)
```
dabbeg, line 38: Are you relying on hset to to clear the timeout?

dabbeg: If so see https://redis.io/commands/expire

```
The timeout will only be cleared by commands that delete or overwrite the contents of the key, including DEL, SET, GETSET and all the *STORE commands. This means that all the operations that conceptually alter the value stored at the key without replacing it with a new one will leave the timeout untouched. For instance, incrementing the value of a key with INCR, pushing a new value into a list with LPUSH, or altering the field value of a hash with HSET are all operations that will leave the timeout untouched.
```

dabbeg (approved):

apps/github-actions-cache/src/app/handlers.ts
sindrig: we agree

PR #5681, opened 2021-11-18
fix(service-portal): get citizenship from natreg + increase timeout

thordurhhh:
## What

get citizenship from natreg + increase timeout

## Why

The current setup sometimes returns the wrong citizenship, in rare cases.
The service can be a bit too slow sometimes, and time out.

## Checklist:


disaerna (approved):
LGTM 👍

PR #5683, opened 2021-11-18
feat(pdpp): slider number to thousand

sigruntg:
# Slider numbers display in thousands

[Attach a link to issue if relevant](https://app.clickup.com/t/1gwh55w)

## What

Make slider numbers display in thousands once sliding.
The min and max still stay the same.

## Screenshots / Gifs

image

## Checklist:


libs/clients/payment-schedule/src/lib/payment-schedule.ts
```diff
@@ -74,9 +74,7 @@ export class PaymentScheduleAPI extends RESTDataSource {
 
     return {
       ...response.distributionInitialPosition,
-      minPayment: roundToUpperThousand(
-        response.distributionInitialPosition.minPayment,
-      ),
+      minPayment: response.distributionInitialPosition.minPayment,
```
bensijons, line 77: No need to spread since the rounding is gone, this could just be

`return response.distributionIntialPostion`

libs/application/templates/public-debt-payment-plan/src/fields/components/Slider/Slider.tsx
```diff
@@ -118,11 +123,16 @@ const Slider = ({
...
       dragX.current = Math.max(0, Math.min(size.width, currentX))
       const index =
         dragX.current / sizePerCell + min >= max
           ? max
-          : roundByNum(dragX.current / sizePerCell, step) + min
+          : dragX.current / sizePerCell + min <= min
+          ? min
+          : roundByNum(dragX.current / sizePerCell, step) === 0
+          ? min
+          : roundByNum(dragX.current / sizePerCell, step) + roundedMin
```
bensijons, line 135: perhaps add a comment explaining what is going on here, I feel like this will be hard to understand when coming back to it later

libs/clients/payment-schedule/src/lib/payment-schedule.ts
sigruntg: Fixed

libs/application/templates/public-debt-payment-plan/src/fields/components/Slider/Slider.tsx
sigruntg: Done

libs/application/templates/public-debt-payment-plan/src/fields/components/Slider/Slider.tsx
```diff
@@ -118,11 +123,18 @@ const Slider = ({
   const dragBind = useDrag({
     onDragMove(deltaX) {
       const currentX = x + deltaX
+      const roundedMin = toFixedNumber(min, 1, 10)
       dragX.current = Math.max(0, Math.min(size.width, currentX))
+      // Get value to display in slider.
+      // Get max if more or equel to max, get min if less or equel to min and then show rest with only one desimal point.
```
bensijons, line 129:
```suggestion
      // Get max if more or equal to max, get min if less or equal to min and then show rest with only one decimal point.
```

bensijons (approved):

PR #5684, opened 2021-11-18
feat(services-notifications): Working queue setup using aws SQS

maggisk:
# ...

## What

Implement a queue using AWS SQS

## Screenshots / Gifs

## Checklist:


apps/services/notifications/src/app/modules/notifications/dto/createNotification.dto.ts
```diff
@@ -0,0 +1,36 @@
...
+  @IsEnum(MessageTypes)
+  @ApiProperty({ enum: MessageTypes })
+  type!: MessageTypes
+}
+
+export class NewPostholfMessage {
+  @ApiProperty({ enum: [MessageTypes.NewPostholfMessage] })
+  type!: MessageTypes.NewPostholfMessage
+
+  @IsString()
```
stefanhar, line 19: Should we maybe validate that this string is of a valid date format e.g. with `@isDateString`

apps/services/notifications/src/app/modules/notifications/dto/createNotification.dto.ts
```diff
@@ -0,0 +1,36 @@
...
+  from!: string
+
+  @IsNationalId()
+  @ApiProperty()
+  recipient!: string
+
+  @IsString()
+  @IsNotEmpty()
+  @ApiProperty()
+  postholfMessageId!: string
```
stefanhar, line 29: Maybe not use icelandic names for fields?

apps/services/notifications/src/app/modules/notifications/dto/createNotification.dto.ts
```diff
@@ -0,0 +1,36 @@
+import { IsEnum, IsNotEmpty, IsString } from 'class-validator'
+import { ApiProperty } from '@nestjs/swagger'
+import { IsNationalId } from '@island.is/nest/validators'
+
+export enum MessageTypes {
+  NewPostholfMessage = 'newPostholfMessage',
```
stefanhar, line 6: Should this maybe not include icelandic in name?

apps/services/notifications/src/app/modules/notifications/consumer.service.ts
```diff
@@ -0,0 +1,87 @@
...
+
+  async run(): Promise<void> {
+    this.logger.info('Starting consumer')
+
+    while (!this.shutdown) {
+      const { Messages: messages = [] } = await this.queue.client.send(
+        new ReceiveMessageCommand({
+          QueueUrl: this.queue.queueUrl,
+          MaxNumberOfMessages: 10,
+          WaitTimeSeconds: 20,
```
stefanhar, line 44: Maybe we should move these values to env variable for easy tuning in production?

apps/services/notifications/src/app/modules/notifications/consumer.service.ts
```diff
@@ -0,0 +1,87 @@
...
+
+    this.logger.info('Waiting for workers to finish')
+    await this.currentBatchProcessing
+
+    this.logger.info('Closing connection')
+    this.queue.client.destroy()
+  }
+
+  async run(): Promise<void> {
+    this.logger.info('Starting consumer')
```
stefanhar, line 37: Maybe it's better if these messages better reflect each other, one might not instantly equate `Closing connection` with `Shutting down consumer`

stefanhar (changes requested):

apps/services/notifications/src/app/modules/notifications/connection.provider.ts
```diff
@@ -0,0 +1,72 @@
...
+  SQSClient,
+  SQSClientConfig,
+  CreateQueueCommand,
+  GetQueueAttributesCommand,
+} from '@aws-sdk/client-sqs'
+import { SqsChannel } from '../../../types'
+
+const minute = 60
+const hour = minute * 60
+const day = hour * 24
```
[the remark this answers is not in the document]
maggisk: well, if you change the constant of how time is represented then yes, you will break the code :/
AWS SQS accepts all intervals as seconds. These are not exported and IMO this is a lot more readable and less error prone than `360` and `8640`

apps/services/notifications/src/app/modules/notifications/connection.provider.ts
```diff
@@ -0,0 +1,72 @@
...
+}): Promise<SqsChannel> => {
+  const client = new SQSClient(sqsConfig)
+
+  // Create a dead letter queue
+  // https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-dead-letter-queues.html
+  const deadLetterQueue = await client.send(
+    new CreateQueueCommand({
+      QueueName: deadLetterQueueName,
+      Attributes: {
+        MessageRetentionPeriod: (day * 14).toString(),
```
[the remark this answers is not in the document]
maggisk: I agree that it might be a good idea to make these configurable.
If you however change the `Attributes` of an existing queue AWS will throw an error.
So to change the attributes of an existing queue you have to run another command.
I'd suggest we make it a separate task/ticket since this is not a simple as replacing these with environment constants.

apps/services/notifications/src/app/modules/notifications/connection.provider.ts
```diff
@@ -0,0 +1,72 @@
+import assert from 'assert'
```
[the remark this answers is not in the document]
maggisk: agreed. will change

apps/services/notifications/src/constants.ts
```diff
@@ -0,0 +1,2 @@
+export const CONFIG_PROVIDER = 'CONFIG'
+export const CONNECTION_PROVIDER = 'CONNECTION'
```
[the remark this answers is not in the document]
maggisk: the time constants are not used outside of 1 single file and could cause confusion/errors since aws sqs could represent 1 second differently than other libraries (second  being 1 vs 1000)

As we don't need the constants anywhere else I think is safer to not expose them or make them re-usable

apps/services/notifications/src/app/modules/notifications/notifications.controller.ts
```diff
@@ -0,0 +1,62 @@
...
+  BadRequestException,
+  Req,
+  HttpCode,
+} from '@nestjs/common'
+import { ApiBody, ApiExtraModels, getSchemaPath } from '@nestjs/swagger'
+import { validate, ValidationError } from 'class-validator'
+import { Request } from 'express'
+import { NotificationProducerService } from './producer.service'
+import {
+  NewPostholfMessage,
```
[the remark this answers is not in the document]
maggisk: I'm following the RFC which you were invited to comment on :)

https://www.notion.so/Notifications-RFC-v2-e62a3e1b9c0e4469af838919b1c996ef

apps/services/notifications/src/environments/environment.prod.ts
```diff
@@ -1,3 +1,21 @@
...
+import assert from 'assert'
+
+const environment = {
   production: true,
+
+  mainQueueName: process.env.MAIN_QUEUE_NAME,
+  deadLetterQueueName: process.env.DEAD_LETTER_QUEUE_NAME,
+
+  sqsConfig: {
+    region: process.env.SQS_REGION ?? 'eu-west-1',
```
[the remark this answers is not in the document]
maggisk: I don't really have an opinion on this.
I just see `eu-west-1` hardcoded in a few places in the repo.
Maybe this could be hardcoded as well.
I thought this would be a sinsible default.

apps/services/notifications/src/app/modules/notifications/consumer.service.ts
```diff
@@ -0,0 +1,87 @@
+import { Inject, Injectable, OnModuleDestroy } from '@nestjs/common'
```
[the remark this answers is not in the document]
maggisk: IMO that is just more code for an unnecessary abstraction.

apps/services/notifications/src/app/modules/notifications/consumer.service.ts
maggisk: I think this describes pretty well what is actually being done, but I'm open to suggestions :)

apps/services/notifications/src/app/modules/notifications/consumer.service.ts
maggisk: Maybe. Not sure.

These values are the maxmimum allowed.
If we make them configurable it would cause runtime errors to set them higher than SQS allows.

maggisk: Maybe that's not such a big deal since this is only the worker and it would not cause any real downtime/interruption

apps/services/notifications/src/app/modules/notifications/dto/createNotification.dto.ts
maggisk: Again, RFC. I don't care one way or the other

apps/services/notifications/src/app/modules/notifications/dto/createNotification.dto.ts
maggisk: RFC

apps/services/notifications/src/app/modules/notifications/dto/createNotification.dto.ts
maggisk: This is actually the name of the institution sending the message.

Maybe it would be clearer to name this field `sender`

apps/services/notifications/src/app/modules/notifications/e2e/notifications.spec.ts
```diff
@@ -0,0 +1,104 @@
...
+
+  async waitFor(cond: (messages: Message[]) => boolean, max = 1000) {
+    const start = Date.now()
+    while (Date.now() - start < max) {
+      if (!cond(this.received)) {
+        await sleep(10)
+      }
+    }
+  }
+}
```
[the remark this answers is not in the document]
maggisk: Would that make sense in e2e tests?

Don't you want to run external services/depdencies in docker if possible?

apps/services/notifications/src/app/modules/notifications/notifications.controller.ts
```diff
@@ -0,0 +1,62 @@
...
+  Object.assign(type, body)
+  throwIfError(await validate(type, { forbidUnknownValues: false }))
+
+  const ValidatorClass = ValidatorTypeMap[type.type]
+  const message = new ValidatorClass()
+  Object.assign(message, body)
+  throwIfError(await validate(message, { forbidUnknownValues: true }))
+
+  return message
+}
```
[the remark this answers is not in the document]
maggisk: I spent WAY too much time on this and this was the result.

You want to first validate the type, and then validate the rest of the object differently bases on the type.
It is AFAIK totally impossible to do in a single custom class-validator or decorator.

I strongly dislike class-validator. What comes closest is `validateNested` from `class-transformer`, but that requires the `type` to be a part of a nested object in stead of the top level input object.

E.g. class-validator can validate something like `{ input: { type: 'a' | 'b', ... } }` but not `{ type: 'a' | 'b', ... }` where rest of the validation is based on the value of `type`

apps/services/notifications/src/app/modules/notifications/notifications.controller.ts
```diff
@@ -0,0 +1,62 @@
...
+  ) {}
+
+  @Post()
+  @ApiBody({
+    schema: {
+      type: 'object',
+      oneOf: [{ $ref: getSchemaPath(NewPostholfMessage) }],
+    },
+  })
+  @HttpCode(202)
```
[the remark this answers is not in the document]
maggisk: ```A 201 status code indicates that a request was successful and as a result, a resource has been created (for example a new page). The status code 202 indicates that server has received and understood the request, and that it has been accepted for processing, although it may not be processed immediately```

https://benramsey.com/blog/2008/04/http-status-201-created-vs-202-accepted/

I think 202 is a better fit, but I don't have a strong opinion on it

apps/services/notifications/src/app/modules/notifications/connection.provider.ts
```diff
@@ -0,0 +1,72 @@
...
+        // If a message hasn't been processed for this amount of time it will
+        // be discarded (or moved to dead letter queue?)
+        MessageRetentionPeriod: (day * 7).toString(),
+        // If a message isn't acknowledged for this amount of time after
+        // delivery it will be re-inserted into the queue
+        VisibilityTimeout: (minute * 10).toString(),
+        // If a message delivery fails `maxReceiveCount` times it will be
+        // moved to the dead letter queue for inspection/debugging
+        RedrivePolicy: JSON.stringify({
+          maxReceiveCount: 3,
```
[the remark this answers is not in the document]
maggisk: same issue as above with the dead letter queue

apps/services/notifications/src/app/modules/notifications/connection.provider.ts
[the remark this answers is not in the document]
stefanhar: Yes works either way, I don't mind this staying as is

apps/services/notifications/src/constants.ts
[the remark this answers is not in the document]
stefanhar: See your point, marking this as resolved

apps/services/notifications/src/app/modules/notifications/notifications.controller.ts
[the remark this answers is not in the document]
stefanhar: Got me there :) I should have spotted this there.
I'll counter with the [codeStandards](https://docs.devland.is/technical-overview/code-standards#language), which hold true unless we think that `Postholf` can be classified as  a government domain word?

apps/services/notifications/src/app/modules/notifications/notifications.controller.ts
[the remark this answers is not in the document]
stefanhar: This seems very valid, I meant comment to the code though 😄, so that when someone comes across this they know why it's there

apps/services/notifications/src/app/modules/notifications/notifications.controller.ts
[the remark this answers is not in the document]
stefanhar: I agree that 202 is a better fit, but I think it's better to keep to the docs wherever possible to maintain consistency.
Another way to think of this is that the entry was successfully created in the queue 🤷

apps/services/notifications/src/app/modules/notifications/e2e/notifications.spec.ts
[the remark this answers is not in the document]
stefanhar: I think it makes sense here MSW intercepts requests on the network layer,
and runs as part of your application so you don't need to run external services

apps/services/notifications/src/app/modules/notifications/dto/createNotification.dto.ts
stefanhar: Haha, yes that might be a bit clearer

apps/services/notifications/src/app/modules/notifications/consumer.service.ts
stefanhar: I think the messages describe what is being done just fine, I'm thinking when someone who does not know this app is debugging the logs they see `Starting consumer` but the shutting down message is `Closing connection`.
My concern is that it might not be obvious that the application is done running, therefor it might be more clear to have a message that better reflects the start message e.g. `Shutting down consumer`

stefanhar (approved):

workspace.json
```diff
@@ -7990,6 +7990,12 @@
             "buildTarget": "services-notifications:build"
           }
         },
+        "worker": {
+          "builder": "@nrwl/workspace:run-commands",
+          "options": {
+            "command": "yarn ts-node -P apps/services/notifications/tsconfig.app.json apps/services/notifications/src/worker.ts"
+          }
+        },
```
eirikurn, line 7998: I don't think this will work well with deployments. Not sure if you've checked with Andes, but I think this will either need to be a standalone "app" (using shared models/code like the auth services), or extra process arguments to the "main.ts" file.

We could maybe expand infra-nest-server#bootstrap to support workers or arbitrary commands when the process is started with extra arguments, eg `node main.js --worker`.

arnorhs: yes, that's right. We need to iterate on this part a bit

arnorhs: @eirikurn this is only for local development while we are iterating on this. we have a separate pull request for the infra setup, which is very much a draft, and waiting for this one, where we will take the next steps forward with andes

apps/services/notifications/src/environments/environment.prod.ts
```diff
@@ -1,3 +1,23 @@
-export const environment = {
+import assert from 'assert'
+
+const environment = {
```
eirikurn, line 3: File replacement does not work anymore (with esbuildnx) to keep prod and dev environments separate.

Please consider using `@island.is/nest/config`, or join these files together.

apps/services/notifications/src/app/modules/notifications/queueConnection.provider.ts
```diff
@@ -0,0 +1,73 @@
+import assert from 'assert'
+import { Inject, Injectable, OnApplicationBootstrap } from '@nestjs/common'
+import {
+  SQSClient,
+  CreateQueueCommand,
+  GetQueueAttributesCommand,
+} from '@aws-sdk/client-sqs'
```
eirikurn, line 7: Can you use/extend `libs/message-queue` in this service?

The reason it exists is that we wanted to abstract proprietary SQS code in one place so it is easier to replace if we have to.

workspace.json
eirikurn: Sure. As discussed on slack, I recommend starting the worker from `main.ts`, and use a `@nrwl/node:execute` builder with `"args": "--something"`.

apps/services/notifications/src/app/modules/notifications/queueConnection.provider.ts
maggisk: We probably could. I didn't use it because it was undocumented, uses AWS SNS which I'm not sure why we need, and doesn't support graceful shutdown of workers.

The documentation and graceful shutdown is something I could fix.
Can you explain to me why it uses SNS in stead of pushing directly to the queue?

apps/services/notifications/src/environments/environment.prod.ts
maggisk: I see. I used the code generator to bootstrap the project, but it seems to be quite outdated :/

apps/services/notifications/src/app/modules/notifications/queueConnection.provider.ts
dabbeg: I can give some context into that. We used SNS to fanout messages to multiple SQS queues.
So using SNS in front of SQS gives you the option to send a single message to SNS and SNS will send that message to all the connected SQS queues.

apps/services/notifications/infra/notifications.ts
```diff
@@ -0,0 +1,21 @@
...
+    .serviceAccount('notifications')
+    .command('node')
+    .args('--tls-min-v1.0', 'main.js')
+    .env({
+      MAIN_QUEUE_NAME: 'notifications',
+      DEAD_LETTER_QUEUE_NAME: 'notifications-failure',
+      AWS_REGION: 'eu-west-1',
+    })
+    .secrets({
+      SQS_ACCESS_KEY: '/k8s/notifications/SQS_ACCESS_KEY',
```
dabbeg, line 16: SQS_ACCESS_KEY and SQS_SECRET_ACCESS_KEY is unnecessary.
We need to create a irsa to give the service SQS permissions in the infra repo.

apps/services/notifications/src/app/modules/notifications/queueConnection.provider.ts
dabbeg: It's a little bit more boilerplate creating the exhange(SNS) and then binding it to the queue but it also gives us the possibility to do more advanced routing of our messages.

charts/islandis/values.staging.yaml
```diff
@@ -799,11 +799,56 @@ namespaces:
...
+  namespace: 'notifications'
+  podSecurityContext:
+    fsGroup: 65534
+  replicaCount:
+    default: 2
+    max: 3
+    min: 2
+  secrets:
+    CONFIGCAT_SDK_KEY: '/k8s/configcat/CONFIGCAT_SDK_KEY'
+    XXXX: '/k8s/notifications/XXXX'
```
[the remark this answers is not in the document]
arnorhs: sorry.. hadn't regenerated the values..

apps/services/notifications/infra/notifications.ts
arnorhs: ok, so skip it .. makes sense

workspace.json
arnorhs: yup, it's been updated and docs have been added to indicate.

eirikurn (approved):
Approving code owner files. Hope to see the AWS code abstracted soon.

apps/services/notifications/infra/notifications.ts
```diff
@@ -0,0 +1,17 @@
...
+  service('notifications')
+    .image('services-notifications')
+    .namespace('notifications')
+    .serviceAccount('notifications')
+    .command('node')
+    .args('--tls-min-v1.0', 'main.js')
+    .env({
+      MAIN_QUEUE_NAME: 'notifications',
+      DEAD_LETTER_QUEUE_NAME: 'notifications-failure',
+      AWS_REGION: 'eu-west-1',
```
arnorhs, line 13: tek út

apps/services/notifications/infra/notifications.ts
```diff
@@ -0,0 +1,17 @@
+import { service, ServiceBuilder } from '../../../../infra/src/dsl/dsl'
+
+export const serviceSetup = (): ServiceBuilder<'notifications'> =>
+  service('notifications')
+    .image('services-notifications')
+    .namespace('notifications')
+    .serviceAccount('notifications')
+    .command('node')
+    .args('--tls-min-v1.0', 'main.js')
```
arnorhs, line 9: taka út

apps/services/notifications/infra/notifications.ts
```diff
@@ -0,0 +1,17 @@
...
+    .env({
+      MAIN_QUEUE_NAME: 'notifications',
+      DEAD_LETTER_QUEUE_NAME: 'notifications-failure',
+      AWS_REGION: 'eu-west-1',
+    })
+    .grantNamespaces('islandis', 'notfications')
```
arnorhs, line 15: taka út

apps/services/notifications/infra/notifications.ts
```diff
@@ -0,0 +1,17 @@
+import { service, ServiceBuilder } from '../../../../infra/src/dsl/dsl'
+
+export const serviceSetup = (): ServiceBuilder<'notifications'> =>
+  service('notifications')
+    .image('services-notifications')
+    .namespace('notifications')
+    .serviceAccount('notifications')
```
arnorhs, line 7: sitthvort service account

apps/services/user-notification/Dockerfile.test
```diff
@@ -0,0 +1,11 @@
...
+
+RUN apt-get update && apt-get install netcat -y
+
+RUN addgroup runners && useradd -m runner -G runners
+USER runner
+
+WORKDIR "/code"
+VOLUME ["/code"]
+
+ENTRYPOINT /code/scripts/ci/_wait-for.sh localstack:4566 -t 60 -- yarn test services-notifications --runInBand --runner=docker-test --codeCoverage
```
dabbeg, line 11: services-user-notification, right?

apps/services/user-notification/src/main.ts
```diff
@@ -0,0 +1,25 @@
...
+}
+
+const { argv } = yargs(process.argv.slice(2))
+
+if (argv.job === 'worker') {
+  worker()
+} else {
+  bootstrap({
+    appModule: AppModule,
+    name: 'services-notifications',
```
dabbeg, line 21: services-user-notification

pshomov:
@dabbeg if you have the time, can you please take a look at this failing tests issue for `message-queue` and `notification-service`. I thought it was the health check timing out due to the size of the docker image. Increased the retries to 120 with a second interval between retries. The log timestamps do not match the two minute timeout though. Maybe you have some idea what this could be.

apps/services/user-notification/Dockerfile.test
```diff
@@ -0,0 +1,11 @@
+FROM public.ecr.aws/bitnami/node:14.17.0
```
dabbeg: Is this still being used?

apps/services/user-notification/src/main.ts
dabbeg: This should be `services-user-notification`, right?

apps/services/user-notification/docker-compose.base.yml
```diff
@@ -0,0 +1,14 @@
+version: '3.3'
```
dabbeg: Take a look at how other services are configured regarding docker-compose.
There is only a single docker-compose.yml and no `base`, `ci`, `dev`

dabbeg (approved):
Looks good!

PR #5686, opened 2021-11-18
feat(fa): When creating municipality creates admin

MargretFinnboga:
# ...

https://app.asana.com/0/1201365172053613/1201380382032049

## What

Creates admin when municipality is created

## Why

It is mandatory when super admin creates municipality it has to have one admin

## Checklist:


apps/financial-aid/api/src/app/modules/municipality/dto/createMunicipality.input.ts
```diff
@@ -1,15 +1,19 @@
 import { Allow } from 'class-validator'
 
 import { Field, InputType } from '@nestjs/graphql'
-import type { CreateMunicipality } from '@island.is/financial-aid/shared/lib'
+import { CreateStaffInput } from '../../staff'
 
 @InputType()
-export class CreateMunicipalityInput implements CreateMunicipality {
+export class CreateMunicipalityInput {
```
Rikkiabb, line 7: Since this is not implementing `CreateMunicipality` are we still using that variable anywhere? And if not should we throw it in the trash 🗑️

apps/financial-aid/backend/src/app/modules/staff/staff.service.ts
```diff
@@ -72,6 +74,26 @@ export class StaffService {
...
+    municipalityInput: CreateMunicipalityDto,
+    t: Transaction,
+  ): Promise<StaffModel> {
+    return await this.staffModel.create(
+      {
+        nationalId: input.nationalId,
+        name: input.name,
+        municipalityId: municipalityInput.municipalityId,
+        email: input.email,
+        roles: input.roles,
```
Rikkiabb, line 88: Should we maybe use `CreateStaff` function instead?

Rikkiabb (changes requested):
Nice, minor things to check out

apps/financial-aid/api/src/app/modules/municipality/dto/createMunicipality.input.ts
MargretFinnboga: Yess I am using it in backend here for example: 

`
  createMunicipality(
    createMunicipality: CreateMunicipality,
    createAdmin: CreateStaff,
  ): Promise<Municipality> {
    return this.post('municipality', {
      municipalityInput: createMunicipality,
      adminInput: createAdmin,
    })
  }
`

apps/financial-aid/backend/src/app/modules/staff/staff.service.ts
MargretFinnboga: We already have a `CreateStaff` function, so keep the name or change ? 🙃

Rikkiabb: No use that instead of creating a new one 👼

apps/financial-aid/backend/src/app/modules/staff/staff.service.ts
```diff
@@ -61,34 +61,25 @@ export class StaffService {
...
-    })
-  }
-
-  async createAdmin(
+  async createStaff(
     input: CreateStaffDto,
-    municipalityInput: CreateMunicipalityDto,
-    t: Transaction,
+    municipality: {
+      id: string
```
Rikkiabb, line 67: Should this be typed?

Rikkiabb (approved):
🍔

PR #5698, opened 2021-11-19
fix(pdpp): if minpayment is less than minwagepayment show alert message, else not

sigruntg:
# Add condition for alert message visibility

[1m3cypq](https://app.clickup.com/t/1m3cypq)

## What

If min payment is less than min wage payment the user should see alert message.

## Checklist:


bensijons (approved):
LGTM

PR #5702, opened 2021-11-19
feat(application-system-form): adding plausible for stats

arnorhs:
wip

I had some varied attempts at doing this conditionally in the html depending on the environment. That part is easy in the next.js projects, since the head is in JS land - I might need to configure a webpack plugin to make this env dependent in the html file, or generating the html and/or header with react-helmet.

I think the more reasonable thing would be to actually have JS version of the include script.. which did not rely on the presence of the script tag on the page..

Anyways.. open to ideas.

## Checklist:


apps/application-system/form/src/index.html
```diff
@@ -38,6 +38,13 @@
     <meta name="viewport" content="width=device-width, initial-scale=1" />
     <link rel="icon" type="image/x-icon" href="favicon.ico" />
 
+    <script
+      async
+      defer
+      data-domain="island.is/umsoknir"
```
eirikurn, line 44: Interesting. Is it an issue if this is hard coded and running on all environments?

It looks like this is what [service portal](https://github.com/island-is/island.is/blob/99c6b7c33cad398ce6f70bd7deee4e66d89c923d/apps/service-portal/src/index.html#L44) is doing. They're also using the outbound-links version of the script if we want to be consistent.

eirikurn (approved):

arnorhs (commented):
Im going to put these comments in here for future reference.. and i suppose it's better to have something tracked, rather than nothing at all. But definitely something that needs to be considered across the whole project.

PR #5708, opened 2021-11-22
fix(ci): api service schema generation

pshomov:
# ...

Server-side features cannot be evaluated at build time.

## What

## Why

## Screenshots / Gifs

## Checklist:


arnorhs (approved):

arnorhs (approved):
This has got something to do with the env variables having different values when they are imported?

pshomov:
> This has got something to do with the env variables having different values when they are imported?

No, just lazy evaluating the configuration info since it is affected by the server-side features.

apps/api/src/app/environments/environment.ts
```diff
@@ -1,333 +1,341 @@
...
+          username: process.env.PAYMENT_SCHEDULE_USER,
+          password: process.env.PAYMENT_SCHEDULE_PASSWORD,
+        },
+        islykill: {
+          cert: process.env.ISLYKILL_CERT,
+          passphrase: process.env.ISLYKILL_SERVICE_PASSPHRASE,
+          basePath: process.env.ISLYKILL_SERVICE_BASEPATH,
+        },
+      }
+    : {
```
darrikonn, line 163: Having this huge ternary makes it hard to find where the dev config vs. prod config begins/ends.

My proposal would be to rather keep the same format as before with two variables devConfig and prodConfig and return the previous ternary at the bottom ; makes it easier to read.

darrikonn (changes requested):

apps/api/src/app/environments/environment.ts
pshomov: 👍🏻

darrikonn (approved):

PR #5709, opened 2021-11-22
fix(fa): Exception thrown from guards

Rikkiabb:
# ...

We are throwing forbidden exception from some guards, fixing it and throwing unauth.


MargretFinnboga (approved):
🏈 🏉 🏀 ⚽

PR #5714, opened 2021-11-22
fix(tracing): Don't mask nationalids on http>4xx

sindrig:
Avoid masking antionalids in request urls when we don't get HTTP OK
Also add masking to express handlers as they were exposing national
ids accidentally.

This closes island-is/infrastructure#958


libs/infra-tracing/src/lib/datadog-tracer.ts
```diff
@@ -1,20 +1,41 @@
...
+import { Span } from 'opentracing'
 
 if (process.env.NODE_ENV !== 'development') {
+  const rewriteUrl = (url: string): string =>
+    url
+      .split('/')
+      .map((segm) => (isPerson(segm) ? '--MASKED--' : segm))
+      .join('/')
+  const shouldMask = (statusCode?: number): boolean =>
+    !statusCode || statusCode < 400
```
pshomov, line 12: We need an additional input here which is the magic env variable that will allow this to be possible at all. By default shouldMask should return `true` for all status codes.

libs/infra-tracing/src/lib/datadog-tracer.ts
```diff
@@ -1,20 +1,41 @@
...
         request: (span, req, res) => {
           const url =
-            res?.url || `http(s)://${req?.getHeader('host')}${req?.path}`
-          const urlRewritten = url
-            .split('/')
-            .map((segm) => (isPerson(segm) ? '--MASKED--' : segm))
-            .join('/')
-          span?.setTag('http.url', urlRewritten)
+            res?.url ||
+            `${req?.protocol}//${req?.getHeader('host')}${req?.path}`
```
pshomov, line 37: I was under the impression req.protocol was not present very much, hence I forced it to be `http(s):`. Maybe have that as a fallback?

pshomov (commented):
Needs a bit more functionality

libs/infra-tracing/src/lib/datadog-tracer.ts
sindrig: It's a required attribute on `ClientRequest` and I have only seen it being present in my tests.

libs/infra-tracing/src/lib/datadog-tracer.ts
sindrig: oh so this shouldn't be always on? talk in person

pshomov (approved):
LGTM

PR #5723, opened 2021-11-23
fix(general-petitions): base url update

albinagu:
baseUrl fix 

## Checklist:


alexandradilja (approved):

libs/application/templates/general-petition/src/fields/ListSubmitted/index.tsx
```diff
@@ -16,12 +16,12 @@ const ListSubmited: FC<FieldBaseProps> = ({ application }) => {
   const isStaging = window.location.origin.includes('beta.staging01.devland.is')
 
   const baseUrlForm = isLocalhost
-    ? 'http://localhost:4200/undirskriftalistar/'
```
arnorhs, line 19: wouldn't it be better to just do
```
baseUrlFrom = `${document.location.origin}/medmaelendalistar/`
```
and skip all the isStaging / isDev business?

arnorhs (changes requested):

libs/application/templates/general-petition/src/fields/ListSubmitted/index.tsx
albinagu: ok vá! takk!

arnorhs (approved):

PR #5736, opened 2021-11-23
fix(application-system): Do not filter out answer from external data provider screen

baering:
## What

Currently when you check the 'I agree' box on the external data screen it is not saved until you transition from the state. This PR changes it so that you save the answer when submitting the external data screen.

## Why

So that users are not taken back to the external data screen should they refresh while in the same state.

## Checklist:


eirikurn (approved):
LGTM

PR #5741, opened 2021-11-24
feat(application-system-ui): prevent persisted errors from being displayed

arnorhs:
image

## Checklist:


libs/application/ui-shell/src/components/FormExternalDataProvider.tsx
```diff
@@ -145,6 +153,7 @@ const FormExternalDataProvider: FC<{
   const activateBeforeSubmitCallback = (checked: boolean) => {
     if (checked) {
       setBeforeSubmitCallback(async () => {
+        setSuppressProviderErrors(false)
```
baering, line 156: Maybe place it after the request so that if the user had errors previously they won't be briefly shown during the fetching of data/

baering (changes requested):

baering (approved):

PR #5746, opened 2021-11-24
feat(service-portal): add verification back to email and mobile for settings

thordurhhh:
## What

Add back verification options to email + phone.
We do not have any way to store the `emailVerified` boolean in the islykill service, so we will take advantage of the userprofile service for that. 

When looking at this you will see when a user creates an account

* The user will create the account with both islykill and userprofile
* Phone in islykill and locale, phone and email in userprofile.
* When the email from userprofile is confirmed with the verification service, it will be updated within the islykill service.

Feature deployment for testing: https://service-portalconfirm-email-mobile-beta.dev01.devland.is/minarsidur/stillingar/personuupplysingar

For email edit you need to be on the verified SES list on dev.

## Why

The locale option is not available in the islykill service
The confirmation for email can not be used for the islykill email update

## Checklist:


disaerna (approved):
LGTM

libs/api/domains/user-profile/src/lib/userProfile.service.ts
```diff
@@ -62,14 +62,17 @@ export class UserProfileService {
...
         mobilePhoneNumber:
           islyklarData?.mobilePhoneNumber ?? profile.mobilePhoneNumber,
-        email: islyklarData?.email ?? profile.email,
-        mobilePhoneNumberVerified: islyklarData?.mobilePhoneNumber
-          ? true
-          : profile.mobilePhoneNumberVerified,
-        emailVerified: islyklarData?.email ? true : profile.emailVerified,
+        email:
+          profile.emailVerified === false
+            ? profile.email
```
eirikurn, line 74: What is the default value of `profile.emailVerified` for newly signed up users? If false, then this API will return an undefined email, right?

Is the plan now to have all island.is users verify their existing email when they sign in?

libs/service-portal/settings/personal-information/src/screens/EditPhoneNumber/EditPhoneNumber.tsx
```diff
@@ -57,14 +66,27 @@ export const EditPhoneNumber: ServicePortalModuleComponent = () => {
...
+          mobilePhoneNumber: formData.tel,
+        })
+      } else {
+        await createIslykillSettings({
+          mobile: `+${parsePhoneNumber?.countryCallingCode}-${parsePhoneNumber?.nationalNumber}`,
+        })
+        await createUserProfile({
+          mobilePhoneNumber: formData.tel,
+        })
+      }
```
eirikurn, line 89: I think it would be better to remove the Islykill domain and instead wrap it behind the user profile domain/service.

We also need to verify the phone number, right?

eirikurn (commented):
Couple of comments and a discussion on Slack.

libs/api/domains/user-profile/src/lib/userProfile.service.ts
thordurhhh: The default value is false I believe.
Yes the plan is to have everyone who enters their email to verify it.

So the way this is set up:

User enters email in input
email is then saved with the userprofile service + a verification email is sent.
A get request for the user email would return the userprofile email at this point + `emailVerified: false`.

Once the user verifies, only then the email would be saved in the islyklar db. 
A get request at this point would return the verified email (saved in the islyklar db) + `emailVerified: true`

thordurhhh:
> Couple of comments and a discussion on Slack.

Yes, lets continue with discussion on slack. I can see I need good documentation along with this change.

libs/service-portal/settings/personal-information/src/screens/EditPhoneNumber/EditPhoneNumber.tsx
thordurhhh: We do want to verify the phone number yes. Phone verification is included here. The only way for the user to call this is to have already verified beforehand.

PR #5762, opened 2021-11-25
fix(secrets): Fetch all secrets required by svc

sindrig:
Stop depending on dev label. If they should not be fetched, specify
them in code rather than aws.


infra/src/cli/render-secrets.ts
```diff
@@ -3,14 +3,15 @@ import { UberChart } from '../dsl/uber-chart'
...
 import { SSM } from 'aws-sdk'
-import { GetParametersByPathResult } from 'aws-sdk/clients/ssm'
 
 const API_INITIALIZATION_OPTIONS = {
   region: 'eu-west-1',
   maxRetries: 10,
   retryDelayOptions: { base: 200 },
 }
 
+const EXCLUDED_ENVIRONMENT_NAMES = ['DB_PASSWORD']
```
darrikonn, line 13: Would rather go for opt-in approach since we're dealing with secrets.

infra/src/cli/render-secrets.ts
```diff
@@ -20,41 +25,44 @@ export const renderSecrets = async (service: string) => {
     (chart) => generateYamlForEnv(uberChart, ...chart.dev).services,
```
darrikonn, line 25: I'm not sure at what point this got changed ... but previously you could request environment variables for the whole project without specifying which part, e.g.
```
yarn get-secrets air-discount-scheme
```
used to work ... but now you've to request
```
yarn get-secrets air-discount-scheme-web
yarn get-secrets air-discount-scheme-backend
yarn get-secrets air-discount-scheme-api
```

sindrig: previously it only used SSM prefixes, e.g. if you did `yarn get-secrets air-discount-scheme` it would only fetch ssm parameters with prefix `/k8s/air-discount-scheme*`. one of the problems with this was that a lot of projects share secrets (examples include `api`s usage of `/k8s/shared/api/RSK_API_USERNAME` and `/k8s/documents/POSTHOLF_CLIENTID`

sindrig: Now it uses the actual services defined in the dsl, so you get exactly what you need, nothing more, nothing less

pshomov: I mean we have the dependencies of the services so we can pull all the secrets for all of them but would you use them in the same shell?

pshomov (approved):
LGTM

infra/src/cli/render-secrets.ts
pshomov: Nevermind, it is not in the shell but a file. But my point still stands about having the option for the related services.

PR #5768, opened 2021-11-26
fix(template-api-modules-driving-license): re-adding the return statement

arnorhs:
re-adding the return statement I idiotically removed when trying to fix the consistency of the action

## Checklist:


albinagu (approved):
nice!

PR #5771, opened 2021-11-26
feat(testing): Infra/support for external contract tests

sindrig:
This closes island-is/infrastructure#916
Related: https://github.com/island-is/helm/pull/585


pshomov (approved):
LGTM

apps/external-contracts-tests/test-suites/national-registry.spec.ts
```diff
@@ -0,0 +1,25 @@
...
+
+describe('is.island.external.national', () => {
+  let client: NationalRegistryApi
+  beforeAll(async () => {
+    client = await NationalRegistryApi.instantiateClass({
+      baseSoapUrl: process.env.SOFFIA_SOAP_URL!,
+      user: process.env.SOFFIA_USER!,
+      password: process.env.SOFFIA_PASS!,
+      host: process.env.SOFFIA_HOST_URL!,
+    })
```
eirikurn, line 14: Would be great to update this with the new NationalRegistry client and use the NestJS DI to construct and get the service instance. This way it is configured exactly the same way as in other apps.

We can either use the same `Test.createTestingModule()` setup as we use in other jest tests, or have a shared AppModule that contains everything which all the tests need.

eirikurn (changes requested):

apps/external-contracts-tests/test-suites/national-registry.spec.ts
pshomov: Done. Although I am getting 401 on both Staging and Dev.

apps/external-contracts-tests/jest-to-dd.ts
```diff
@@ -0,0 +1,26 @@
...
+//   .map((tc) => tc.name)
+// const failedTests = testCasesInfo
+//   .filter((tc) => tc.status === 'failure')
+//   .map((tc) => tc.name)
+
+// console.log(`Failed test: ${failedTests}`)
+// console.log(`Successful tests: ${successfulTests}`)
+
+// successfulTests.forEach((tc) => client.check(tc, client.CHECKS.OK))
+// failedTests.forEach((tc) => client.check(tc, client.CHECKS.CRITICAL))
```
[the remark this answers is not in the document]
pshomov: Restored. This is a script developed while I was researching unknown areas of this task. Not used immediately but in the next phase of this Story.

apps/external-contracts-tests/test-suites/national-registry.spec.ts
```diff
@@ -1,24 +1,55 @@
...
+      imports: [
+        NationalRegistryModule.register({
+          xRoadPath: createXRoadAPIPath(
+            process.env.XROAD_BASE_PATH_WITH_ENV!,
+            XRoadMemberClass.GovernmentInstitution,
+            process.env.XROAD_TJODSKRA_MEMBER_CODE!,
+            process.env.XROAD_TJODSKRA_API_PATH!,
+          ),
+          xRoadClient: XROAD_CLIENT,
+        }),
```
eirikurn, line 34: You need to merge or rebase from main, since all of this has changed to use the ConfigModule.

Same with the environment variable comment above. Shouldn't need any environment variables locally as is (currently, the config has good default values for development).

apps/external-contracts-tests/test-suites/national-registry.spec.ts
```diff
@@ -1,24 +1,55 @@
...
+    client = await module.get(EinstaklingarApi)
   })
   it('should get user correctly', async () => {
-    const user = await client.getUser('0101302989')
-    expect(user.Fulltnafn).toEqual('Gervimaður Ameríka')
+    const user = await client.einstaklingarGetEinstaklingur({
+      id: '0101302989',
+      xRoadClient: XROAD_CLIENT,
+    })
+    expect(user.fulltNafn).toEqual('Gervimaður Ameríka')
```
eirikurn, line 44: Ahh ... this API needs an IDS access token with specific scopes. Need to set up a machine client and configure EnhancedFetch#autoAuth. Hopefully we can call their primary endpoint without a nationalId claim.

But there are some endpoints (in this service and others) which require a nationalId claim, so we should also discuss how we'll set up external-contract-tests against those. Eg with Gervimenn and some custom OAuth2 flow on dev/staging?

eirikurn (changes requested):

apps/external-contracts-tests/test-suites/national-registry.spec.ts
pshomov: Since this API usage seems like a lot of work that I am not familiar with I will comment this out for future reference and replace with fake success responses. Since this PR is to put the infrastructure in place, I will focus on that.
Alternatively I can revert to using the SOFIA API since that was working and is a bit better than fake tests.

@eirikurn Any preferences on the above?

eirikurn (approved):
Decided to start with the old SOFIA test to have some value in this first v1. Devs should contribute further integrations.

PR #5780, opened 2021-11-27
feat(core): Cached enhancedFetch + national registry integration

eirikurn:
## What

Extends createEnhancedFetch to support caching using [cache-manager](https://www.npmjs.com/package/cache-manager). Builds on top of standard cache-control semantics so server-side cache headers "just work" while still allowing the client to override the cache behaviour.

Includes these advanced cache-control features:

* `stale-if-error`, provides fallback content from cache if remote server is down.
* `stale-while-revalidate`, returns responses from cache while updating it behind the scenes for future requests.
* `private` cache based on the nationalId of the authenticated user.
* Conditional requests using "If-None-Match" and "If-Modified-Since".
* Shared vs private max-age.

[Readme](https://github.com/island-is/island.is/blob/cache-fetch/libs/clients/middlewares/README.md#caching)

This PR also includes a `defineConfig` refactor and cached fetch integration for the National Registry XRoad Client.

## Why

Because we want to provide good user experience even though some of our dependencies aren't reliable or fast.

## Todo

- [x] Create new ENVs in helm for Auth API, Auth Public API and GraphQL API.
- [x] Test national registry xroad domain, see if everything works after the refactor.
- [x] Configure cache-manager against redis, including any helm ENVs.
- [x] Add back national registry timeout values for auth-public-api and graphql api.

## Checklist:


infra/src/dsl/xroad.ts
```diff
@@ -178,6 +183,19 @@ export const Education = new XroadConf({
...
   env: {
+    XROAD_NATIONAL_REGISTRY_SERVICE_PATH: {
+      dev: 'IS-DEV/GOV/10001/SKRA-Protected/Einstaklingar-v1',
+      staging: 'IS-TEST/GOV/6503760649/SKRA-Protected/Einstaklingar-v1',
+      prod: 'IS/GOV/6503760649/SKRA-Protected/Einstaklingar-v1',
+    },
+    // Only cache on dev for now.
+    XROAD_NATIONAL_REGISTRY_REDIS_NODES: {
+      dev:
+        '["clustercfg.general-redis-cluster-group.5fzau3.euw1.cache.amazonaws.com:6379"]',
```
sindrig, line 194: Could this be a comma separated list instead?

eirikurn: No ... 😆 

The new [config library](https://github.com/island-is/island.is/tree/main/libs/nest/config) has native support for parsing and validating JSON environment variables. IMO it is more precise, flexible and has [better DX](https://github.com/island-is/island.is/blob/7dbef7d5134b832458a9653fed45fd68c53f82e7/libs/clients/national-registry/v2/src/lib/nationalRegistryClient.config.ts) for some types of configuration.

It would be cool to upgrade the dsl to support this better:

```
XROAD_NATIONAL_REGISTRY_REDIS_NODES: {
      dev: json([
        'clustercfg.general-redis-cluster-group.5fzau3.euw1.cache.amazonaws.com:6379',
      ]),
},
```

Happy to jump on a call to argue about this. God knows there's plenty of stuff in these [two](https://github.com/island-is/island.is/pull/5612) PRs worth chatting about. 😃

sindrig: It's also much more error prone. This'll get translated to regular ENV variables in our deployments. In the case of an emergency (example: we need to provision a new redis cluster and point all deployments to that) the first step would be to manually update the deployments, then to fix it in the dsl and wait until it builds etc.

I'd be much more comfortable editing a comma separated list than a json encoded array :grimacing:

apps/services/auth-public-api/src/app/modules/delegations/national-registry.config.ts
```diff
@@ -1,17 +0,0 @@
...
-export const NationalRegistryConfig: NationalRegistryModuleConfig = {
-  xRoadPath: createXRoadAPIPath(
-    environment.nationalRegistry.xroad.basePath,
-    environment.nationalRegistry.xroad.memberClass,
-    environment.nationalRegistry.xroad.memberCode,
-    environment.nationalRegistry.xroad.apiPath,
-  ),
-  xRoadClient: environment.nationalRegistry.xroad.clientId,
-  fetch: {
-    timeout: 5000,
```
Rikkiabb, line 15: Where is this configured now?

infra/src/dsl/xroad.ts
eirikurn: I have absolute faith in the JSON writing skills of the devops team. 🤓 Worst case the config validation stops the deployment and tells you what you did wrong.

I've also seen comma separated environment variables messed up many times, with whitespace or buggy empty/undefined handling. I also see devs often adding more elaborate syntax and stuff to these variables to support extra metadata.

apps/services/auth-public-api/src/app/modules/delegations/national-registry.config.ts
eirikurn: In config and helm. Have to add it:

image

Rikkiabb (approved):
Awesome refactoring and very nice to se caching in enhancedFetch 👏

saevarma (approved):
Amazing 🙌

sindrig (approved):
LGTM

valurefugl (approved):

PR #5791, opened 2021-11-29
feat(p-mark-application): Template and start of project for P-mark

alexandradilja:
This is the base template for P-mark application. This is a copy of the driving licence application project with some unused functionality in the api removed. Template for the application is still heavily based on the driving license template so that the other developers in this application can use it for guidance. 
The api part of the application has not been fully cleaned off unused driving license features since we are still figuring out what we can reuse. 
This application can be run locally and accessed at http://localhost:4242/umsoknir/p-merki/

## Checklist:


eirikurn:
We don't want any unfinished code on main since it gets deployed to production every now and then.

For features that are fairly isolated, they should be developed on a long-running feature branch with feature deployments. 

If the feature changes a lot of existing code, or two teams/projects are collaborating, then it can be split into small "complete" PRs, merged to main and hidden from deployment behind feature flags.

It looks like this project will be fairly isolated. In either case it can make code reviews easier to first create project skeletons (with minimal copied code) and make your team codeowner for it, then do work in smaller PRs (either against main or your "big" feature branch).

PR #5804, opened 2021-11-29
feat(core/web): External/Internal links

kksteini:
There were a bunch of implementations for `isLinkExternal` or `isLinkInternal`.
Moved to a single implementation and tested.

Previously did not account for full URL to island.is.

## What?

The `isLinkExternal` check or similar checks are all used to determine whether links should open in a new tab or not.
All links that do not navigate away from island.is should open in the same window whereas external links should not.

## Screenshots / Gifs
image

## Checklist:


eirikurn (approved):
Nice!

darrikonn (approved):

barabrian (approved):
Nice! LGTM

arnorhs (changes requested):
Looks good with tests etc

A few issues I have though:

is `isLinkExternal` a good name? since it depends on the context what is considered external or internal. if the purpose is to decide whether a link should open in a tab or not, you could just name it `shouldLinkOpenInCurrentTab`

it might be better to construct this function from options where you supply the hostnames you want to match or even paths. there's definitely cases where you want links that are "internal" to open in a tab.

even though this is a step towards a better future, as is it seems too broadly scoped for a particular use case

libs/shared/utils/src/lib/linkExternal.ts
```diff
@@ -0,0 +1,18 @@
...
+const islandisRe = new RegExp(template)
+
+export const isLinkExternal = (href: string): boolean => {
+  const externalCandidate =
+    typeof href === 'string' && href.indexOf('://') !== -1
+
+  return externalCandidate && !href.match(islandisRe)
+}
+
+export const isLinkInternal = (href: string): boolean => {
```
arnorhs, line 16: isn't this unneccessary?

arnorhs (changes requested):

libs/shared/utils/src/lib/linkExternal.ts
kksteini: Perhaps. I decided not to mess with semantics in other files.
`isLinkExternal` and `isLinkInternal` were being used alternatingly and so I considered this.

kksteini:
> Looks good with tests etc
> 
> A few issues I have though:
> 
> is `isLinkExternal` a good name? since it depends on the context what is considered external or internal. if the purpose is to decide whether a link should open in a tab or not, you could just name it `shouldLinkOpenInCurrentTab`
> 
> it might be better to construct this function from options where you supply the hostnames you want to match or even paths. there's definitely cases where you want links that are "internal" to open in a tab.
> 
> even though this is a step towards a better future, as is it seems too broadly scoped for a particular use case

Good points.

External to the domain vs internal to the domain is how I understood the names. Maybe it isn't as clear cut as that?
 `shouldLinkOpenInCurrentTab` might be a better name for the current situation but the usage of the function is tied to `target="_blank"` in the React implementation of those who choose to use the function. 
Further down the line, programmers might want to choose to use this function for other purposes, f.x, sorting links by whether they point out of domain or not.

---
 
It felt a little hacky when I introduced the hosts to the regex's template so I agree that a better construction is worth considering. I'm not too up to date on the environment/host debate but I assume that this can be inferred from the window object or the process object, depending on context.

oddsson (approved):

libs/shared/utils/src/lib/linkExternal.ts
kksteini: Actually, the more I look at it the less I like it. I'll just change the semantics and remove the extra function

arnorhs (approved):

PR #5815, opened 2021-11-30
fix(ci): Use temporary published version of island-is/cache

sindrig:
To maybe get rid of API rate limiting... This should be temporary until either https://github.com/actions/toolkit/pull/947 or https://github.com/actions/cache/pull/679 are accepted and released


pshomov (approved):
LGTM

darrikonn (approved):

PR #5816, opened 2021-11-30
fix(web): set fixed height for search section on desktop

stjanilofts:
- Minor fix to avoid the "jump" when the image/video has yet to load.
- Make shortTitle nullable

## Checklist:


kksteini (approved):
LGTM

PR #5822, opened 2021-11-30
fix(accident-notification): Fix issues with application

Thorrik58:
# Fix issues with application

## What

Report error to user when application fails
Send correct id when applying as proxy or juridicial person
Handle non successful submits
Text changes
When reporting as a juridicial person fill the company info from the applicant and juridicial info.

## Why

Needed for application

## Screenshots / Gifs

## Checklist:


libs/application/template-api-modules/src/lib/modules/templates/accident-notification/accident-notification.service.ts
```diff
@@ -49,10 +49,14 @@ export class AccidentNotificationService {
       const answers = application.answers as AccidentNotificationAnswers
       const xml = applictionAnswersToXml(answers, attachments)
 
-      const { ihiDocumentID } = await this.documentApi.documentPost({
+      const { success, ihiDocumentID } = await this.documentApi.documentPost({
         document: { doc: xml, documentType: 801 },
       })
 
+      if (success === 0) {
```
obmagnusson, line 56: does this coincide with the 412 errors your catching here below

libs/application/template-api-modules/src/lib/modules/templates/accident-notification/accident-notification.service.ts
```diff
@@ -78,6 +82,14 @@ export class AccidentNotificationService {
       }
     } catch (e) {
       this.logger.error('Error submitting application to SÍ', e)
+      // In the case we get a precondition error we present it to the user
+      if (e.body && e.body.errorList && e.body.errorList.length > 0) {
+        throw new Error(
+          `Villa kom upp við vistun á umsókn. ${e.body.errorList
+            .map((e: any) => e.errorDesc)
```
obmagnusson, line 89: Is this safe or even informative in all cases to propogate to the client? 
I would think an error of "Innri villa í kerfum Sí" or "Villa við staðfestingu á heilleika skjals. Villuboð: ORA-30937... " would cause major confusion or atleast some kind of panic.

Thorrik58: Good point, ill get a list of possible errors we can get from the precondition failed and whitelist the ones we want to present to the users. Putting the task on hold until then.

libs/application/template-api-modules/src/lib/modules/templates/accident-notification/accident-notification.service.ts
Thorrik58: Nope we can get an error less response with success 0 :/

obmagnusson: Dont we want to log something here ?

PR #5823, opened 2021-12-01
feat(core): Feature flag library for Nest APIs

eirikurn:
## What

A feature flag library for our APIs. Wraps the low level API to provide support for our user objects, consistent with the `@island.is/react/feature-flags` library.

Also includes a NestJS guard to wrap entire endpoints and resolvers behind feature flags.

[Readme](https://github.com/island-is/island.is/blob/nest-feature-flag/libs/nest/feature-flags/README.md)

## Why

To make it easier to use feature flags in our APIs.

## Checklist:

### DIFF
--- a/libs/nest/feature-flags/src/lib/feature-flag.config.ts
+++ b/libs/nest/feature-flags/src/lib/feature-flag.config.ts
@@ -1,0 +1,11 @@
+import { defineConfig } from '@island.is/nest/config'
+
+export const FeatureFlagConfig = defineConfig({
+  name: 'FeatureFlag',
+  load: (env) => ({
+    sdkKey: env.required(
+      'CONFIGCAT_SDK_KEY',
+      'YcfYCOwBTUeI04mWOWpPdA/KgCHhUk0_k2BdiKMaNh3qA',
+    ),
+  }),
+})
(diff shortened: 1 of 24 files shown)


libs/nest/feature-flags/src/lib/feature-flag.config.ts
```diff
@@ -0,0 +1,11 @@
+import { defineConfig } from '@island.is/nest/config'
+
+export const FeatureFlagConfig = defineConfig({
+  name: 'FeatureFlag',
+  load: (env) => ({
+    sdkKey: env.required(
+      'CONFIGCAT_SDK_KEY',
+      'YcfYCOwBTUeI04mWOWpPdA/KgCHhUk0_k2BdiKMaNh3qA',
```
saevarma, line 8: Should the SDK key be in the source?

eirikurn: It's already in the codebase and included in all of our production browser bundles. So I'll leave this for now until we decide to remove it globally at some point.

pshomov (approved):
LGTM

saevarma (approved):
Great addition!

PR #5828, opened 2021-12-01
fix(icelandic-names-registry): a few tweaks

stjanilofts:
- GET name by id under `@Get('/id/:id')` instead of just `@Get(':id')`
- Add pipe for making sure id is a number
- Add pipe for making sure string contains characters from Icelandic alphabet
- Remove some webpack stuff which I believe is no longer used

This should fix [this error](https://app.datadoghq.eu/logs?query=service%3Aicelandic-names-registry-backend&cols=host%2Cservice&event=AQAAAX1w8vBLZplCEAAAAABBWDF3OHZpVUFBQkplQXVIZUs3dGxBQUE&index=&messageDisplay=inline&stream_sort=desc&from_ts=1635686183023&to_ts=1638278183023&live=true
).

## Checklist:


albinagu (approved):

apps/icelandic-names-registry/backend/src/app/modules/icelandic-name/icelandic-name.controller.ts
```diff
@@ -53,15 +54,15 @@ export class IcelandicNameController {
     return await this.icelandicNameService.getAll()
   }
 
-  @Get(':id')
+  @Get('id/:id')
```
darrikonn, line 57: Not sure this falls inline with the API design guide (notably the naming conventions).
This doesn't specify what resource you're fetching and isn't really a RESTful design.
I personally haven't used this API before, so I'm assuming this is fetching Icelandic names, thus the following would make more sense for URI structure:
```
api/icelandic-names-registry/names/:id
api/icelandic-names-registry/names?search=xxx
```

Also feels like initial-letter can be combined with search ... it's just a matter of skipping `%` in the database query.

darrikonn (changes requested):

darrikonn (approved):

PR #5834, opened 2021-12-01
feat(sakavottord2): new application type

johannaagma:
# ...

## What

A new application in the application system where people can apply for their criminal record. 

## Why

## Screenshots / Gifs

## Checklist:


darrikonn (commented):
Can you remove the `kubectl` binary?

package.json
```diff
@@ -138,6 +139,7 @@
     "date-fns": "2.14.0",
     "dd-trace": "0.33.1",
     "deepmerge": "4.2.2",
+    "document-register-element": "1.13.1",
```
darrikonn, line 142: as far as I can see, this is not used

darrikonn: maybe a peer dep of react-pdf?

apps/service-portal/src/assets/images/bus.svg
```diff
@@ -1,433 +1 @@
-<svg width="432" height="445" viewBox="0 0 432 445" fill="none" xmlns="http://www.w3.org/2000/svg">
```
thordurhhh, line 1: This file seems to be still in use in the service-portal. 😱

johannaagma: Seems like the file changed from 433 lines to 1 line (but we didnt change anything), but I have reverted the changes in this file

package.json
johannaagma: We have removed this dependency

darrikonn (commented):
Can you add your team to the CODEOWNERS file as well.

johannaagma:
@darrikonn We have already been added to codeowners here:
https://github.com/island-is/island.is/pull/5873/files

libs/api/domains/criminal-record/src/lib/graphql/main.resolver.ts
```diff
@@ -0,0 +1,22 @@
...
+
+  //Note: not being used
+  @Query(() => CriminalRecord)
+  async getCriminalRecord(@Args('ssn') ssn: string) {
+    return await this.criminalRecordService.getCriminalRecord(ssn)
+  }
+
+  @Query(() => Boolean)
+  async checkCriminalRecord(@Args('ssn') ssn: string) {
+    return await this.criminalRecordService.checkCriminalRecord(ssn)
```
darrikonn, line 20: Both of these queries look dangerous.
By implementing it like this you're exposing every single criminal record to users that have a valid login. It's most always incorrect to pass in the nationalId as an argument.
In addition, this is not following the naming conventions for GraphQL, ref: https://docs.devland.is/technical-overview/api-design-guide/graphql-naming-conventions

Arguments should be input objects and names `input`. The name of the queries should be the domain itself:
`CriminalRecord`. And this should most likely use the CurrentUser as the nationalId in order to prevent exposing the pre-mentioned danger ^.

darrikonn (changes requested):

libs/island-ui/core/src/lib/PdfViewer/PdfViewer.tsx
```diff
@@ -0,0 +1,58 @@
+import React, { FC, useState } from 'react'
+import { Box } from '../Box/Box'
+import { pdfjs, Document, Page } from 'react-pdf'
```
stjanilofts, line 3: I don't know for sure but according to [this](https://github.com/wojtekmaj/react-pdf#webpack) I think you can skip `pdfjs` for the worker stuff by importing this way:

```javascript
import { Document, Page } from 'react-pdf/dist/esm/entry.webpack';
```

stjanilofts (approved):
Approved (for island-ui codeowners) but with one question.

libs/island-ui/core/src/lib/PdfViewer/PdfViewer.tsx
unnursolingimars: Yes we actually imported it from 'react-pdf/dist/esm/entry.webpack' first but got some errors on the automatic checks so we changed it in consultation with Eiríkur :)

Ingimarsson (approved):
Changes in syslumenn client look good.

libs/application/templates/criminal-record/src/forms/CriminalRecordForm.ts
```diff
@@ -0,0 +1,117 @@
...
+            buildDataProviderItem({
+              id: 'userProfile',
+              type: 'UserProfileProvider',
+              title: m.userProfileInformationTitle,
+              subTitle: m.userProfileInformationSubTitle,
+            }),
+            buildDataProviderItem({
+              id: 'payment',
+              type: 'FeeInfoProvider',
+              title: '',
```
baering, line 47: Intentionally no title here?

libs/application/templates/criminal-record/src/forms/CriminalRecordForm.ts
```diff
@@ -0,0 +1,117 @@
...
+            buildCustomField({
+              id: 'payment.over',
+              title: '',
+              component: 'OverviewPaymentCharge',
+            }),
+          ],
+        }),
+        buildDescriptionField({
+          id: 'final',
+          title: 'Takk',
```
baering, line 90: Missing translations here.

libs/application/templates/criminal-record/src/lib/CriminalRecordTemplate.ts
```diff
@@ -0,0 +1,141 @@
...
+              read: 'all',
+            },
+          ],
+        },
+        type: 'final' as const,
+      },
+    },
+  },
+  mapUserToRole(): ApplicationRole | undefined {
+    return Roles.APPLICANT
```
baering, line 137: You only want to return this role if the `nationalId` (passed in as the first argument) is equal to `application.applicant` (application is passed in as the second argument).

libs/application/templates/criminal-record/src/forms/Payment.tsx
```diff
@@ -0,0 +1,75 @@
...
+          title: m.payment,
+          condition: () => {
+            return !window.document.location.href.match(/\?done$/)
+          },
+          description: (application) => {
+            const { paymentUrl } = application.externalData.createCharge
+              .data as { paymentUrl: string }
+
+            if (!paymentUrl) {
+              throw new Error()
```
baering, line 39: I think this will break the application and likely require manual database state change so that the user can interact with it again.

baering: You might want to conditionally show a screen in case the url is missing to allow the user to move the application to draft and try again. Similarly only showing this screen when the url does exist.

baering (commented):
Good job on your first application. I did see some things that need to be addressed and then left some minor comments as well.

libs/api/domains/criminal-record/src/lib/graphql/models/criminalRecord.model.ts
```diff
@@ -0,0 +1,7 @@
+import { Field, ObjectType, ID } from '@nestjs/graphql'
+
+@ObjectType()
+export class CriminalRecord {
+  @Field()
+  contentBase64!: string
```
darrikonn, line 6: Wondering about this approach.
This GraphQL field is not used so it doesn't really matter (might be worth removing it). But do we plan on being able to download a Criminal Record file, like a PDF?

Another possible solution instead of returning base64 encoded file (which has a huge payload), is to use the download service. Might be worth checking out if you plan on being able to download the file via browser?

libs/clients/criminal-record/src/lib/criminalRecordApi.service.ts
```diff
@@ -0,0 +1,29 @@
...
+@Injectable()
+export class CriminalRecordApi {
+  constructor(private readonly api: CrimeCertificateApi) {}
+
+  public async getCriminalRecord(ssn: string): Promise<CriminalRecord> {
+    const contentBlob = await this.api.apiPdfCreatePersonIdGet({
+      personId: ssn,
+    })
+    const contentArrayBuffer = await contentBlob.arrayBuffer()
+    const contentBase64 = Buffer.from(contentArrayBuffer).toString('base64')
```
darrikonn, line 14: Another solution would be to upload it to the download S3 bucket... but might be better to look at the download service?

darrikonn (commented):
Looking good!
Core files mostly approved, just waiting for answers regarding my PR review questions. Otherwise approved 👍

libs/application/templates/criminal-record/src/forms/CriminalRecordForm.ts
unnursolingimars: Yes it's supposed to be hidden from the user because it contains no personal data.

libs/application/templates/criminal-record/src/forms/CriminalRecordForm.ts
unnursolingimars: Thanks! Fixing it.

libs/api/domains/criminal-record/src/lib/graphql/models/criminalRecord.model.ts
johannaagma: Ah forgot to remove this model class, after I removed an unecessary function (getCriminalRecord) in the graphQL layer.

But we are going to allow users to download the criminal record as PDF (through libs/client, not libs/api - if that matters, so it is not going through the graphQL layer).
I hadn't realized we had a download service, but we also need to display the PDF in the browser, so it will look something like this:
image

Is it possible to use this download service to access the base64 string for the PDF to display in the browser (as well as being able to download the file)?

libs/application/templates/criminal-record/src/forms/Payment.tsx
unnursolingimars: Okay thanks, fixing it :)

libs/application/templates/criminal-record/src/dataProviders/FeeInfoProvider.ts
```diff
@@ -0,0 +1,35 @@
+import { PaymentCatalogItem } from '@island.is/api/schema'
+import {
+  SuccessfulDataProviderResult,
+  FailedDataProviderResult,
+} from '@island.is/application/core'
+import { PaymentCatalogProvider } from '@island.is/application/data-providers'
+import { m } from '../lib/messages'
+
+const CHARGE_ITEM_CODE = 'AY101'
```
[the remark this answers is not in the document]
unnursolingimars: @baering Do you have a suggestion of where I should store these values?

libs/api/domains/criminal-record/src/lib/graphql/models/criminalRecord.model.ts
darrikonn: Alright, nice!

Yeah I believe you can use it.. it's just a REST API. Implementation can be seen in `apps/download-service/*`, although you probably won't need it since if your interaction is through a REST API you can just stream it yourself :).

darrikonn (approved):
LGTM

johannaagma:
@Ingimarsson We made some more small changes to Syslumenn client, you can see them here: https://github.com/island-is/island.is/pull/5834/commits/012b0288a12edfe53ba1f5c083025cbe99cd9bf6
Could you review that as well? :)

libs/application/templates/criminal-record/src/dataProviders/FeeInfoProvider.ts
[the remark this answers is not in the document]
baering: Maybe add this as an enum to a new constant file here `libs/shared/constants/src/lib`

baering (approved):
LGTM

PR #5837, opened 2021-12-01
refactor(auth-public-api): Refactor endpoints and add api tests

saevarma:
# Endpoint refactor and adding API tests

This PR closes the following Notion tickets: 
* [Fix Auth Public API endpoints to match API design guide](https://www.notion.so/Fix-Auth-Public-API-endpoints-to-match-API-design-guide-084ffb5adca34bc986f8101a1314e0c3)
* [Create API tests for Auth Public API endpoints](https://www.notion.so/Create-API-tests-for-Auth-Public-API-endpoints-bcdb6e4dc3094cbc8a145784207cab45)
* [NationalId in URLs](https://www.notion.so/NationalId-in-URLs-e5b33f535f2c4cdbafc3d40793aba262)

### Refactor endpoints

Simplifying and making the `auth-public-api` match the API Design Guide.
This creates two endpoints:
* `/v1/actor/delegations`: To get receiving delegations for the actor if it exists on the authentication. Fallbacks to the user if there is not actor.
* `/v1/me/delegations`: CRUD on delegations for the current user.

### Adding API tests

We are adding API tests using @island.is/testing with the SQlite in-memory database.
Our aim is to make this as a reference for other API projects in the future.

### NationalId in URL

NationalId in URLs have been removed and we are now using the delegation id instead.

### Other

* Few fixes in the `@island.is/testing` related to mock authentication and creating fixtures.
* Adds HttpProblemResponse to `@island.is/nest/problem`

## Why

We want to make this API public for other organisations to consume so they can use delegations with island.is authentication.

## Screenshots / Gifs

## Checklist:


valurefugl (approved):

libs/api/domains/auth/src/lib/meDelegations.service.ts
```diff
@@ -0,0 +1,94 @@
...
+  constructor(private delegationsApi: MeDelegationsApi) {}
+
+  private delegationsApiWithAuth(auth: Auth) {
+    return this.delegationsApi.withMiddleware(new AuthMiddleware(auth))
+  }
+
+  getDelegations(user: User): Promise<DelegationDTO[]> {
+    return this.delegationsApiWithAuth(user).meDelegationsControllerFindAll({
+      direction: MeDelegationsControllerFindAllDirectionEnum.Outgoing,
+      isValid: true,
```
eirikurn, line 30: Fyi. You should also see delegations that are valid in the future. Not sure if this encompasses that or if it should be a `valid` string enum filter instead.

libs/api/domains/auth/src/lib/meDelegations.service.ts
```diff
@@ -0,0 +1,94 @@
...
+  ): Promise<DelegationDTO> {
+    return this.delegationsApiWithAuth(user).meDelegationsControllerFindOne({
+      delegationId,
+    })
+  }
+
+  async getDelegationByOtherUser(
+    user: User,
+    { toNationalId }: DelegationByOtherUserInput,
+  ): Promise<DelegationDTO> {
```
eirikurn, line 46: This might be undefined, right? I'd suggest this and `getDelegationById` should be consistent, either throw or return null.

libs/api/domains/auth/src/lib/meDelegations.service.ts
```diff
@@ -0,0 +1,94 @@
...
+    const payload = {
+      delegationId,
+      updateDelegationDTO: {
+        scopes: scopes?.map((scope) => ({
+          ...scope,
+          validTo: scope.validTo ? scope.validTo : undefined,
+        })) as UpdateDelegationScopeDTO[],
+      },
+    }
+    console.log(payload)
```
eirikurn, line 89: Remove

libs/api/domains/auth/src/lib/meDelegations.service.ts
```diff
@@ -0,0 +1,94 @@
...
+
+    return delegations[0]
+  }
+
+  createDelegation(
+    user: User,
+    { toNationalId, scopes }: CreateDelegationInput,
+  ): Promise<DelegationDTO> {
+    return this.delegationsApiWithAuth(user).meDelegationsControllerCreate({
+      createDelegationDTO: { toNationalId, scopes },
```
eirikurn, line 62: Do the POST/PUT APIs accept data nested in a createDelegationDTO property? Does that follow the Rest Design Guide?

libs/api/domains/auth/src/lib/meDelegations.service.ts
```diff
@@ -0,0 +1,94 @@
...
+    return true
+  }
+
+  updateDelegation(
+    user: User,
+    { delegationId, scopes }: UpdateDelegationInput,
+  ): Promise<DelegationDTO> {
+    const payload = {
+      delegationId,
+      updateDelegationDTO: {
```
eirikurn, line 82: Same here

libs/api/domains/auth/src/lib/meDelegations.service.ts
```diff
@@ -0,0 +1,94 @@
...
+        scopes: scopes?.map((scope) => ({
+          ...scope,
+          validTo: scope.validTo ? scope.validTo : undefined,
+        })) as UpdateDelegationScopeDTO[],
```
eirikurn, line 86: What is this `map` for? Mapping null to undefined?

In the national registry I embraced DTOs allowing both undefined and null, which better reflected the OpenAPI schema. Might make sense to do the same here.

libs/api/domains/auth/src/lib/resolvers/delegationScope.resolver.ts
```diff
@@ -19,19 +19,11 @@ export class DelegationScopeResolver {
 
   @ResolveField('type')
   resolveType(@Parent() delegationScope: DelegationScopeDTO): string {
-    return delegationScope.scopeName
-      ? ScopeType.ApiScope
-      : delegationScope.identityResourceName
-      ? ScopeType.IdentityResource
-      : ''
+    return delegationScope.scopeName ? ScopeType.ApiScope : ''
```
eirikurn, line 22: When is this condition false?

libs/api/domains/auth/src/lib/resolvers/delegationScope.resolver.ts
```diff
@@ -19,19 +19,11 @@ export class DelegationScopeResolver {
...
   }
 
   @ResolveField('displayName')
   resolveDisplayName(@Parent() delegationScope: DelegationScopeDTO): string {
-    return (
-      delegationScope.apiScope?.displayName ??
-      delegationScope.identityResource?.displayName ??
-      ''
-    )
+    return delegationScope.displayName
```
eirikurn, line 27: I think this is a noop. Can probably remove this function.

libs/clients/auth-public-api/src/lib/auth-public-api-client.module.ts
```diff
@@ -1,37 +1,58 @@
...
-          provide: DelegationsApi,
+          provide: MeDelegationsApi,
+          useFactory: () =>
+            new MeDelegationsApi(
+              new Configuration({
+                fetchApi: fetch,
+                basePath: config.baseApiUrl,
+              }),
+            ),
+        },
```
eirikurn, line 33: Fyi. In the national registry client we do this as a `...apis.map()` expression, which reduces repitition.

libs/nest/problem/src/lib/httpProblemResponse.ts
```diff
@@ -0,0 +1,36 @@
+import { ApiProperty, ApiPropertyOptional } from '@nestjs/swagger'
+
+import { HttpProblem, ProblemType } from '@island.is/shared/problem'
+
+export class HttpProblemResponse implements HttpProblem {
```
eirikurn, line 5: Nice!

libs/service-portal/settings/access-control/src/screens/AccessControl/components/Accesses/Accesses.tsx
```diff
@@ -77,7 +77,7 @@ function Accesses(): JSX.Element {
                   validTo={delegation.validTo}
                   description={kennitala.format(delegation.to.nationalId)}
                   tags={delegation.scopes.map((scope) => scope.displayName)}
-                  href={`${pathname}/${delegation.to.nationalId}`}
+                  href={`${pathname}/${delegation.id}`}
```
eirikurn, line 80: Great improvement!

libs/testing/nest/src/lib/hooks/useAuth.ts
```diff
@@ -1,31 +1,21 @@
...
-      .useValue({
-        canActivate: (context: ExecutionContext) => {
-          const request = getRequest(context)
-          request.user = currentUser
-          return true
-        },
-      })
-      .overrideGuard(ScopesGuard)
-      .useValue({ canActivate: () => true }),
+      .useValue(new MockAuthGuard(auth)),
```
eirikurn, line 20: Nice!

eirikurn (commented):
Beautiful!

libs/api/domains/auth/src/lib/meDelegations.service.ts
saevarma: Oops!

libs/api/domains/auth/src/lib/meDelegations.service.ts
saevarma: Fixed

libs/clients/auth-public-api/src/lib/auth-public-api-client.module.ts
saevarma: Nice, did that and moved the config to use the new ConfigModule

libs/api/domains/auth/src/lib/resolvers/delegationScope.resolver.ts
saevarma: Never, after I removed IdentityResources so I updated this to return just `ScopeType.ApiScope`

libs/api/domains/auth/src/lib/meDelegations.service.ts
saevarma: Strictly the scopes should be a sub-resource of the delegations. But we might want to rethink the delegation model a bit to simplify this resource definitions.

Leaving as is for now.

libs/api/domains/auth/src/lib/meDelegations.service.ts
saevarma: I just moved this around with out looking to much at it. This makes no sense and has been removed.
Also embracing null values more on the API side for optional values.

apps/services/auth-public-api/src/app/modules/delegations/actorDelegations.controller.ts
```diff
@@ -0,0 +1,86 @@
...
+export class ActorDelegationsController {
+  constructor(private readonly delegationsService: DelegationsService) {}
+
+  @ActorScopes(AuthScope.actorDelegations)
+  @Get()
+  @ApiOkResponse({ type: [DelegationDTO] })
+  @ApiBadRequestResponse({ type: HttpProblemResponse })
+  @ApiForbiddenResponse({ type: HttpProblemResponse })
+  @ApiUnauthorizedResponse({ type: HttpProblemResponse })
+  @ApiInternalServerErrorResponse()
```
darrikonn, line 53: Do you believe this is mandatory to specify the default behaviours of a REST api?
Wondering since these swagger decorators are getting really noisy, and ApiBadRequestResponse, ApiForbiddenResponse, ApiForbiddenResponse, and ApiInternalServerErrorResponse don't provide any additional documentation to the route, since they're the behaviours of a REST api.
On the other-side, ApiOkResponse provides a good documentation for this endpoint.

apps/services/auth-public-api/src/app/modules/delegations/meDelegations.controller.ts
```diff
@@ -0,0 +1,313 @@
...
+    }
+
+    // Check if the requested scopes are valid
+    const scopes = requestedScopes.map((scope) => scope.name)
+    const allowedIdentityResources = await this.resourcesService.findAllowedDelegationIdentityResourceListForUser(
+      scopes,
+    )
+    const allowedApiScopes = await this.resourcesService.findAllowedDelegationApiScopeListForUser(
+      scopes,
+    )
```
darrikonn, line 295: Wondering if the count should be done on the database layer to avoid loading bunch of objects into memory.


But this probably won't be a lot of data, so it might pass.

apps/services/auth-public-api/test/fixtures/delegation.fixture.ts
```diff
@@ -0,0 +1,129 @@
...
+
+/**
+ * Creates a Delegation fixture to be used for testing.
+ * @param fromNationalId NationalId of the user granting the delegation
+ * @param toNationalId NationalId of the user receiving the delegation
+ * @param scopes Scopes the delegation grants access to
+ * @param today A date the fixture uses as the current day to set valid or expired scopes.
+ *              The test suite should use useFakeTimers and provide this value for consistent testing.
+ *              Defaults to current datetime, but note that it could lead to undesired test results.
+ * @param expired Flag to indicate if the delgation should be expired. Defaults to false.
```
darrikonn, line 109: nit-picky, but comments should be treated like second-class citizens. They tend to get outdated once developers update the code.
Code should be able to explain itself with good names, thus reduces the need for comments which then only provide extra noise.

e.g. if a developers decides to rename or remove a parameter, there's nothing that forces him/her to remove/rename the comment as well, thus the comments become outdated.

libs/auth-api-lib/src/lib/entities/models/api-resource-user-claim.model.ts
```diff
@@ -7,6 +7,7 @@ import {
   UpdatedAt,
   ForeignKey,
   PrimaryKey,
+  BelongsTo,
```
darrikonn, line 10: unused

apps/services/auth-public-api/src/app/modules/delegations/actorDelegations.controller.ts
saevarma: This provides documentation on the error type the API returns. As we are returning Problem JSON. But I agree that this is really noisy and was thinking about making a custom decorator `@ApiErrorResponses()` fyrir Nest, sem myndi apply-a öllum þessum ásamt Problem týpunni.

darrikonn: sounds much better! Like that approach ... or just really extend swagger as a whole ... have one decorator for swagger:
```typescript
@Documentation({
  response: {
    ok: [DelegationDTO],
    errors: [] // this is provided by default
  },
  description: `Finds all incoming delegations for the signed in user or actor.
			Including the custom delegations as well as natural delegations from NationalRegistry and CompanyRegistry.`,
  query: {
    name: 'direction',
    required: true,
    schema: {
      enum: ['incoming'],
      default: 'incoming',
    },
  },
})
async findAll() {}
```

apps/services/auth-public-api/test/fixtures/delegation.fixture.ts
saevarma: Valid point. Simplified the comment. I still find it valuable to get further details on autocomplete pop up.

apps/services/auth-public-api/src/app/modules/delegations/meDelegations.controller.ts
saevarma: Good comment. I created a count function to return just the count from db. I also just removed the `identityResource` part as we are not using it.

apps/services/auth-public-api/test/fixtures/delegation.fixture.ts
darrikonn: Yeah I see your point ... `vim` user here :D.

darrikonn (approved):
Awesome work!

PR #5842, opened 2021-12-02
fix(service-portal): updates + bugfixes in user profile

thordurhhh:
## What

- Save language to user profile when updating in user menu
- Hide settings page when acting as other user
- Bug-fixing for some issues in onboarding modal & updating profile info with islykill service

## Checklist:


thordurhhh:
Just a little heads up. I was doing the same thing as what is being done in this PR: [#5866](https://github.com/island-is/island.is/pull/5866/files#diff-7953eeb3cc4ff0cbd3ca9a5025948a34279bc8603656a42b25e4f89cc94421c4R2) but a little bit differently.
So I changed mine, because this was a better approach ☝️ 

The reason why I keep it in is because there was a request to get it out with today's (tomorrows) deployment and I was not sure if the other PR would be in by then. (Should be an easy merge conflict to solve, whichever comes in first)

libs/service-portal/settings/personal-information/src/components/Forms/PhoneForm/Steps/FormStep.tsx
```diff
@@ -78,6 +78,7 @@ export const FormStep: FC<Props> = ({
               hasError={errors.tel}
               errorMessage={errors.tel?.message}
               onChange={onChange}
+              maxLength={7}
```
osk, line 81: We don't allow spaces or dashes? E.g. `123-4567` or `123 4567` and strip out dash/space?

libs/service-portal/settings/personal-information/src/screens/UserProfile/UserProfile.tsx
```diff
@@ -48,7 +48,7 @@ const UserProfile: ServicePortalModuleComponent = ({ userInfo }) => {
                 labelColumnSpan={['8/12', '3/12']}
                 editColumnSpan={['1/1', '2/12']}
                 valueColumnSpan={['1/1', '7/12']}
-                content={settings?.mobile ?? ''}
+                content={settings?.mobile?.split('-')[1] ?? ''}
```
osk, line 51: Move into a function, this would fail if the split returns the empty array or array of length 1.
Could fix with `?.[1]` but this is getting a bit hairy for inline wrangling

libs/shared/components/src/auth/UserMenu/UserProfileInfo.tsx
```diff
@@ -40,13 +43,13 @@ export const UserProfileInfo = () => {
             <Box display="flex" alignItems="center" marginRight={2}>
               <Icon type="outline" icon="call" color="blue300" />
             </Box>
-            <Text>{settings?.mobilePhoneNumber}</Text>
+            <Text>{settings?.mobilePhoneNumber?.split('-')[1]}</Text>
```
osk, line 46: Same as with the other `split`  and +1 for refactoring into a function

libs/service-portal/settings/personal-information/src/components/Forms/PhoneForm/Steps/FormStep.tsx
thordurhhh: We do not, but come to think of this now, this should replicate the behaviour of the log-in 2factor screen where there is a space after the first three characters.

libs/shared/components/src/auth/UserMenu/userProfile.graphql.ts
```diff
@@ -8,3 +8,11 @@ export const USER_PROFILE = gql`
     }
   }
 `
+
+export const UPDATE_USER_PROFILE = gql`
+  mutation updateProfile($input: UpdateUserProfileInput!) {
```
darrikonn, line 13: Shouldn't this be called `updateUserProfile`?

libs/shared/components/src/auth/UserMenu/userProfile.graphql.ts
```diff
@@ -8,3 +8,11 @@ export const USER_PROFILE = gql`
...
+    updateProfile(input: $input) {
+      locale
```
darrikonn, line 15: You should always query the ID field of the object you're mutating ... that way, apollo knows how to update related cached objects.

In this case, nationalId should be queried as well.

darrikonn (changes requested):

libs/shared/components/src/auth/UserMenu/userProfile.graphql.ts
thordurhhh: Ah yes, thanks.

Valur (approved):

libs/shared/components/src/auth/UserMenu/userProfile.graphql.ts
thordurhhh: Makes sense, updated!

thordurhhh:
@darrikonn - thanks.
Your comments have been addressed/fixed.

darrikonn (approved):
Core looks good

PR #5851, opened 2021-12-02
fix(skilavottord): auth and add guard

qdd1:
# ...

## What

## Why

## Screenshots / Gifs

## Checklist:

### DIFF
--- a/apps/skilavottord/ws/src/app/modules/recyclingRequest/recyclingRequest.resolver.ts
+++ b/apps/skilavottord/ws/src/app/modules/recyclingRequest/recyclingRequest.resolver.ts
@@ -21,9 +21,6 @@ export class RecyclingRequestResolver {
     private logger: Logger,
   ) {}
 
-  /*
-    List for all recycled vehicles
-  */
   @Authorize({ roles: ['developer', 'recyclingFund'] })
   @Query(() => [RecyclingRequestModel])
   async skilavottordAllRecyclingRequests(): Promise<RecyclingRequestModel[]> {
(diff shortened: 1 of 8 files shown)


apps/skilavottord/ws/src/app/modules/recyclingRequest/recyclingRequest.resolver.ts
```diff
@@ -20,7 +21,10 @@ export class RecyclingRequestResolver {
     private logger: Logger,
   ) {}
 
-  @Authorize({ throwOnUnAuthorized: false })
+  /*
+    List for all recycled vehicles
+  */
```
darrikonn, line 26: Food for thought.
Comments should be treated as secondary class citizens. The code should explain itself with good names and clean code.
Unnecessary comments tend to get outdated and bring extra noise.

darrikonn (commented):
Would like to get verification that this was smoke tested by changing your roles.

apps/skilavottord/ws/src/app/modules/recyclingRequest/recyclingRequest.resolver.ts
darrikonn: So IMO these comments should be deleted since they're not needed and might get outdated.

darrikonn (approved):
Nice work!

iceadapter (approved):
LGTM

tommigitgaur (approved):
LFTM

miamiultras (approved):

PR #5863, opened 2021-12-03
feat(auth-admin-web): Misc minor changes

valurefugl:
# ...

https://www.notion.so/Small-changes-to-admin-UI-a8dc3cc314c640e198707916b2ead65f
https://www.notion.so/Hide-archived-items-in-lists-in-Admin-UI-b23e927aa4074f1e977f7ac2ba4d0641
https://www.notion.so/order-secrets-by-date-a471d4952b224a05ae88f85ff5b6c84d

## What

## Why

## Screenshots / Gifs

## Checklist:


saevarma (changes requested):
I think we should rather let the API filter out the archived items, instead of returning them to the UI just to be filtered out.

valurefugl:
Yeah, you're probably right. Don't you think we should have the option of requesting archived items also though? If we want to enable viewing them/undelete from the ui?

saevarma:
> Yeah, you're probably right. Don't you think we should have the option of requesting archived items also though? If we want to enable viewing them/undelete from the ui?

Yes that would be good, provided via some query parameter.

saevarma (approved):
LGTM

unnarb (approved):

PR #5866, opened 2021-12-03
feat(service-portal): Configure access control on all service portal modules

eirikurn:
## What

Add scope checks to all service portal modules and associated API resolvers/controllers.

## Why

To allow us to release the delegation system. When users are able to get an access token in another's name, we want to make sure that access token can only be used for authorised services. This means that all of our services need to check on some scope in addition to the token audience.

## Checklist:

- [ ] Verify that our clients still work.
- [ ] Add scope migrations or create them in staging/production.
- [x] Review remaining API domains, check if some scope checks are missing.


valurefugl (approved):

JuliusGG (approved):

thordurhhh (approved):
Nice!

LGTM

arnorhs (approved):
LGTM

darrikonn (approved):

miamiultras (approved):

Rikkiabb (approved):
🚢

eirikurn:
Fyi, the CI failure is some strange false negative. Always fails with a node16 issue the first build after push then succeeds when I do "re-run all jobs". Will monitor main.

PR #5886, opened 2021-12-07
chore(auth): Add new scopes to our clients

eirikurn:
## What

Add new scopes to our clients.

## Why

To minimize disruption during the next release. This way, signed in users will already have access tokens with the scopes they need when we deploy new UIs and APIs which check for those scopes.

## Checklist:


thordurhhh (approved):
LGTM

PR #5887, opened 2021-12-07
chore(judicial-system): Move from uncontrolled inputs to controlled inputs

oddsson:
# Move from uncontrolled inputs to controlled inputs

https://app.asana.com/0/1199153462262248/1201423192590376/f

## What

Stop using uncontrolled inputs, i.e. inputs that use defaultValues, and start using controlled inputs, i.e. inputs that use value. 

## Why

This is in response to a regression in #5818. That implementation creates the following flow in our application.

- Pages are rendered with a default case, most of which is undefined.
- The actual case is fetched.
- The page re-renders with the new case.

This introduces an issue with defaultValue, because defaultValue does not change with updated state and this fact broke our autofill functionality.

## Checklist:


gudjong (commented):
I noticed the following:
- Selects are still using defaultValue.
- RulingStepTwo / RulingStepTwoForm use defaultValue in the courtEndTime field.

Not sure if the above is wrong.

oddsson:
I've fixed the default value in Select but not the courtEndTime field. I think that works as intended because we are not autofilling these fields. I've created a task to move over to value instead of defaultValue in these fields (https://app.asana.com/0/1199153462262248/1201473256874575/f).

gudjong (approved):

PR #5888, opened 2021-12-07
fix(fa): Updates env variables

Rikkiabb:
Updates env varibles for clients base url

Environments are already updated.


MargretFinnboga (approved):
✅ 🍾

dabbeg (approved):
:balloon:

PR #5889, opened 2021-12-07
fix(public-debt-payment-plan): Info screen for users that have no debts

bensijons:
# Info screen for users that have no debts

https://app.clickup.com/t/1mugzfj

## What

Info screen for users that have no debts, so that users can't continue with the application.

## Why

Requested after testing

## Screenshots / Gifs

image

## Checklist:


libs/application/templates/public-debt-payment-plan/src/lib/paymentPlanUtils.ts
```diff
@@ -14,6 +15,11 @@ export const prerequisitesFailed = (data: ExternalData) => {
     | PaymentScheduleConditions
     | undefined
 
+    const debts = (data as PaymentPlanExternalData)
+    .paymentPlanPrerequisites?.data?.debts as PaymentScheduleDebts[]
+
+    console.log('debts', debts)
```
sigruntg, line 21: console.log

sigruntg (changes requested):
Looks great, just remove the console.log

libs/application/templates/public-debt-payment-plan/src/lib/paymentPlanUtils.ts
Thorrik58, line 21: Console.log
```suggestion
```

Thorrik58 (changes requested):
LGTM apart from the console log. There is a small thing that bugs me and that is that the application will stay listed when the prerequisites fail (that you have a debt) and we might want to keep the application unlisted until that prerequisite passes. 
We are doing this with the health insurance application but it is a bigger feature if we want to go that route. Just a thought.

sigruntg (approved):

Thorrik58 (approved):

PR #5891, opened 2021-12-07
fix(api): Adding nullable true for resolve fields in auth domain.

saevarma:
## What

Fields can be null if no result is found in NationalRegistry, or the service is offline.

## Why

Missing nullable config for GQL schema.

## Screenshots / Gifs

## Checklist:


eirikurn (approved):
May need to address some clientside typescript issues.

PR #5894, opened 2021-12-07
fix(service-portal): Fixing toast error handling in GrantAccess and typescript issues for delegations

saevarma:
## What

When no identity is found the GQL returns null object so adding handling for `onComplete` if `identity` was not found to display the toast error message.

Fixing typescript compile errors when `to` and `from` fields on `delegation` are missing due to errors in `identity` service.
## Why

Fixing the regression when returning null instead of throwing error when identity was not found in the identity service.

## Screenshots / Gifs

## Checklist:

### DIFF
--- a/libs/service-portal/settings/access-control/src/screens/AccessControl/components/Accesses/Accesses.tsx
+++ b/libs/service-portal/settings/access-control/src/screens/AccessControl/components/Accesses/Accesses.tsx
@@ -70,17 +70,19 @@ function Accesses(): JSX.Element {
                 <EmptyImage width="100%" />
               </Box>
             ) : (
-              authDelegations.map((delegation) => (
-                <AccessCard
-                  key={delegation.id}
-                  title={delegation.to.name}
-                  validTo={delegation.validTo}
-                  description={kennitala.format(delegation.to.nationalId)}
-                  tags={delegation.scopes.map((scope) => scope.displayName)}
-                  href={`${pathname}/${delegation.id}`}
-                  group="Ísland.is"
-                />
-              ))
+              authDelegations.map((delegation) =>
+                delegation.to ? (
+                  <AccessCard
+                    key={delegation.id}
+                    title={delegation.to.name}
+                    validTo={delegation.validTo}
+                    description={kennitala.format(delegation.to.nationalId)}
+                    tags={delegation.scopes.map((scope) => scope.displayName)}
+                    href={`${pathname}/${delegation.id}`}
+                    group="Ísland.is"
+                  />
+                ) : undefined,
+              )
             )}
           </Stack>
         </GridColumn>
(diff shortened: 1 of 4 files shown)


libs/service-portal/settings/access-control/src/screens/GrantAccess/GrantAccess.tsx
```diff
@@ -64,13 +73,11 @@ function GrantAccess() {
...
     IdentityQuery,
     {
-      onError: (error) => {
-        toast.error(
-          formatMessage({
-            id: 'service.portal.settings.accessControl:grant-identity-error',
-            defaultMessage: 'Enginn notandi fannst með þessa kennitölu.',
-          }),
-        )
+      onError: noUserFoundToast,
```
eirikurn, line 76: Ideally "errors" are handled separately in some globally consistent way. But that's outside the scope of this PR.

eirikurn (approved):

libs/service-portal/settings/access-control/src/screens/AccessControl/components/Accesses/Accesses.tsx
```diff
@@ -70,17 +70,19 @@ function Accesses(): JSX.Element {
...
-                  title={delegation.to.name}
-                  validTo={delegation.validTo}
-                  description={kennitala.format(delegation.to.nationalId)}
-                  tags={delegation.scopes.map((scope) => scope.displayName)}
-                  href={`${pathname}/${delegation.id}`}
-                  group="Ísland.is"
-                />
-              ))
+              authDelegations.map((delegation) =>
+                delegation.to ? (
```
saevarma, line 74: We should think about how we should handle if the `to` object is not found in NationalRegistry.

libs/shared/components/src/auth/UserMenu/UserDelegations.tsx
```diff
@@ -62,11 +62,13 @@ export const UserDelegations = ({
     delegations.push(
       ...data.authActorDelegations
         .map((delegation) => ({
-          nationalId: delegation.from.nationalId,
-          name: delegation.from.name,
+          nationalId: delegation.from?.nationalId ?? '',
+          name: delegation.from?.name ?? '',
```
saevarma, line 66: We should think about how we should handle the delegations list in the user menu when the `from` object is not found in NationalRegistry.

eirikurn (approved):

libs/service-portal/settings/access-control/src/screens/GrantAccess/GrantAccess.tsx
darrikonn: Yeah true, but it's related to my project's link: https://github.com/orgs/island-is/projects/9#card-60800557

Currently there's no error handling for the service portal 🙈 
All errors are "swallowed" and just console logged.

libs/service-portal/settings/access-control/src/screens/AccessControl/components/Accesses/Accesses.tsx
darrikonn: I assume this is only for data that's on dev (before we hooked up the national registry if I remember correctly)... don't see this ever happening on prod.

darrikonn (approved):

PR #5897, opened 2021-12-07
feat(island-ui): add description field for select option

stjanilofts:
Asana ticket: https://app.asana.com/0/1199123945323117/1201373464015677

## What

Select options only show the label but this adds an optional description field as well (see screenshot). Description is truncated (after 2 lines) by default.

## Why

We needed this for to show more details about an option.

## Screenshots / Gifs

image

## Checklist:


libs/island-ui/core/src/lib/Select/Components/index.tsx
```diff
@@ -23,11 +23,31 @@ export const Menu = (props: MenuProps<ReactSelectOption>) => (
...
       {...props}
-    />
+    >
+      <>
+        {props.children}
+        {!!description && (
+          <div
+            className={cn(
+              styles.optionDescription,
+              styles.optionDescriptionSizes[size!],
```
barabrian, line 42: do you need the non-null assertion operator here?
```suggestion
              styles.optionDescriptionSizes[size],
```

barabrian (approved):
LGTM, minor comment

barabrian (approved):
LGTM

PR #5902, opened 2021-12-07
fix(fa): Use our own scope for national registry

Rikkiabb:
# ...
https://app.asana.com/0/1201427499183230/1201472387205525

## What
Forking national registry to use our own scope.

## Why
We don't have the default scope that national registry uses

## Checklist:


apps/financial-aid/api/src/app/modules/nationalRegistryXRoadModule/nationalRegistryXRoad.resolver.ts
```diff
@@ -0,0 +1,51 @@
...
+  Scopes,
+  ScopesGuard,
+} from '@island.is/auth-nest-tools'
+import { Audit } from '@island.is/nest/audit'
+import { NationalRegistryXRoadService } from '@island.is/api/domains/national-registry-x-road'
+import { Person } from './models/person.model'
+import { UserSpouse } from './models/userSpouse.model'
+
+@UseGuards(IdsAuthGuard, IdsUserGuard, ScopesGuard)
+@Scopes('@skra.is/individuals')
```
saevarma, line 17: You should use your GQL api scope `@samband.is/internal` here as the skra.is scope is intended for skra.is resources. And here you are protecting your gql query.

saevarma (changes requested):

saevarma (approved):
LGTM

MargretFinnboga (approved):
💯

PR #5904, opened 2021-12-07
feat(island-ui): try out adding a new breakpoint

stjanilofts:
This is just a little test to see how it would work to add one more breakpoint (XXL)


andes-it:

andes-it:

PR #5926, opened 2021-12-08
fix(core): Fix empty response body handling in enhanced fetch cache

eirikurn:
## What

See title.

## Why

Because JS is sometimes evil and conflates empty strings with null/undefined.

## Checklist:


saevarma (approved):
LGTM

PR #5928, opened 2021-12-08
fix(fa): Showing accepted aid calculations and refactoring  

MargretFinnboga:
# ...

https://app.asana.com/0/1201427499183230/1200897217304857
https://app.asana.com/0/1201427499183230/1201473665291472

## What

- Created Breakdown component that osk and veita share (Breakdown of calculations) 
- Added current year to taxcalculation instead of sending it as prop 
- Added two function to taxcalculation since estimated aid and provided aid are similary displayed in osk and veita 

## Why

- Usability purposes, for user and staff to see clearly why accepted aid is the amount that it is

## Checklist:


apps/financial-aid/web-osk/src/components/Estimation/Estimation.tsx
```diff
@@ -94,30 +55,13 @@ const Estimation = ({
...
-            )
-          })}
-
-          <Divider />
-        </>
+      {municipality && aidAmount && (
+        <Breakdown
+          calculations={estimatedBreakDown(
+            aidAmount,
+            usePersonalTaxCredit ?? false,
```
Rikkiabb, line 62: Should `usePersonalTaxCredit` rather default to false?

apps/financial-aid/web-osk/src/components/Estimation/Estimation.tsx
```diff
@@ -94,30 +55,13 @@ const Estimation = ({
...
-                  <Text variant="small">{item.title}</Text>
-                  <Text>{item.calculation}</Text>
-                </Box>
-              </span>
-            )
-          })}
-
-          <Divider />
-        </>
+      {municipality && aidAmount && (
```
Rikkiabb, line 58: Should breakdown rather take these variables and retur null if they are false like we do in other places?

apps/financial-aid/web-veita/src/components/Profile/ApplicationProfile.tsx
```diff
@@ -192,14 +195,28 @@ const ApplicationProfile = ({
...
+      {application.amount && (
+        <AidAmountModal
+          headline="Veitt aðstoð"
+          calculations={acceptedAmountBreakDown(application.amount)}
+          isVisible={isAmountModalVisible}
+          onVisibilityChange={(isVisibleBoolean) => {
+            setAmountModalVisible(isVisibleBoolean)
+          }}
+        />
+      )}
```
Rikkiabb, line 219: Should we join these two?

libs/financial-aid/shared/src/components/Breakdown/Breakdown.tsx
```diff
@@ -0,0 +1,36 @@
...
+          <Box
+            display="flex"
+            justifyContent="spaceBetween"
+            paddingY={2}
+            paddingX={3}
+            borderTopWidth="standard"
+            borderColor="blue200"
+            background={index === calculations.length - 1 ? 'blue100' : 'white'}
+            borderBottomWidth={
+              index === calculations.length - 1 ? 'standard' : undefined
```
Rikkiabb, line 24: Should index === calculations.length -1 be drawn out to a const?

libs/financial-aid/shared/src/lib/taxCalculator.ts
```diff
@@ -13,11 +13,14 @@ export interface TaxInfo {
   personalTaxAllowance: number
   taxPercentage: number
 }
+import format from 'date-fns/format'
+import { Amount, Calculations } from './interfaces'
+
+export const currentYear = format(new Date(), 'yyyy')
```
Rikkiabb, line 19: Is this imported in other places?

Rikkiabb (changes requested):
Some minor comments

libs/financial-aid/shared/src/lib/taxCalculator.ts
MargretFinnboga: We weere always sending it as props, but nor really using it again, so I thought this would be better?

Rikkiabb: Just wondering if it should be exported? And if it should be exported then it should be in another file :)

Rikkiabb (approved):
🏑

PR #5929, opened 2021-12-08
fix(fa): Fixing to create admin 

MargretFinnboga:
# Fixing to create admin and allows use_pseudo_name to be null

https://app.asana.com/0/1201427499183230/1201380382032049

## What

- use_pseudo_name prop didn't allow null 
- created migration to allow null
- fixes error when creating admin for municipality

## Why

- use_pseudo_name doesn't apply to all staff and therefore it should be allowed to be null

## Checklist:


Rikkiabb (changes requested):
I actually fixed this here in a nicer way IMO: https://github.com/island-is/island.is/pull/5933/files#diff-64fc454323bdd8bb8113545cb770b235544f40fc8c1b1ba8690841be0b06ee4bR74 
Feel free to disagree

PR #5978, opened 2021-12-13
fix(island-ui): a11y inputs

thordurhhh:
## What

We were doing a little accessibility review in the service-portal, and noticed some minor issues.

Increased accessibility on inputs select and buttons.
Adding aria-label to buttons and select on datepicker.
Adding aria-labelledby to react select.

## Why

When using voiceover to test these, they are not read out for a user relying on assistive technologies.
When running an automated tool for accessibility on a page with these elements, an error marked critical will show up.

In the case of the Select, there are two input fields rendered. One is associated with the label, but the other one, the one that the user focuses on by keyboard, is blank. So in that case, use the same label.

## Checklist:


thordurhhh:
@oddsson - Any chance you could take a look at this?

oddsson (approved):

oddsson:
Awesome! Sorry for the late reply :)

PR #5989, opened 2021-12-14
fix(feature-flags): Allow feature flagging an entire controller/resolver

eirikurn:
## What

Upgrade the FeatureFlagGuard to support the `@Feature()` decorator on an whole class instead of just individual methods.

## Why

For better DX and avoid mistakes.

## Checklist:


darrikonn (approved):
Nice 🙏

PR #5993, opened 2021-12-14
fix(core): Don't mix in-memory memoization and singletons

eirikurn:
Apollo RESTDataSource instances are designed to be used by the Apollo
Data Source system. Apollo creates new instances of each Data Source
class per request. Each instance memoizes GET requests to better handle
common GraphQL data loading situations.

This can have cause serious caching issues when mixed with NX's
dependency injection system, which by default has singleton services.
We don't want to change these to per-request instances either to avoid
excessive per-request allocations in our growing GraphQL API.

The current workaround is to call `this.memoizedResults.clear()` in the
`willSendRequest` method.

Going forward, we should steer clear from the RESTDataSource class in
favour of [Enhanced Fetch](https://docs.devland.is/libs/clients/middlewares)
as well as OpenAPI generated clients whenever possible.


osk (approved):
👍

Rikkiabb (approved):
👌

Thorrik58 (approved):
LGTM

arnorhs (approved):
TY, 👍 codeowned files.

gudjong (approved):

PR #6006, opened 2021-12-15
fix(endorsement-system): Removing unused code for party letter and party application

alexandradilja:
Removing unused code for party letter and party application.
All unused code regarding these two applications has been removed. The goal of this is to reduce build errors andto make the endorsement system a little less complicated.

## Checklist:


arnorhs (approved):
LGTM! 

image

darrikonn (approved):
Core looks good... love seeing code removals :D!

thordurhhh (approved):
Service-portal LGTM.

eirikurn (approved):
Nice!

apps/services/endorsements/api/migrations/20210325113632-create-initial-tables.js
```diff
@@ -25,10 +25,6 @@ module.exports = {
       tags: {
         type: Sequelize.ARRAY(Sequelize.STRING),
       },
-      validation_rules: {
-        type: Sequelize.JSONB,
-        defaultValue: '[]',
-      },
       owner: {
```
pshomov, line 28: In general we cannot be modifying existing migrations. Unless this is not a real migration but some kind of hack I kind of recall we were using to create data.

apps/services/endorsements/api/migrations/20210811145737-endorsement-list-metadata-sync.js
```diff
@@ -18,61 +18,7 @@ module.exports = {
...
-                'partyApplicationSudurkjordaemi2021',
-                'partyApplicationSudvesturkjordaemi2021',
-              ],
-              'varchar[]',
-            ),
-          },
-        },
-        { transaction: t },
-      )
     })
```
pshomov, line 22: Same comment as the other migration.

pshomov:
The schema generation is failing due to some missing file, probably needs fixing.

apps/services/endorsements/api/migrations/20210325113632-create-initial-tables.js
alexandradilja: Ok, that makes sense! I'll add it back in.

PR #6009, opened 2021-12-15
fix(web): fix process entry links that are using NextLink

stjanilofts:
See https://island.is/samstarf-vid-stafraent-island and click on the blue process entry link button. It links to https://island.is/umsoknir/samstarf/ which `NextLink` tries to prefetch and open once clicked on as if it were an article.

This quick fix adds the `openAsExternalLink` prop which the process entry button now uses.

We should remove NextLink from the ui library Link button and instead make an optional link wrapper for it.

Edit: No need to change `Link` actually, only made change to process entry link. Also added a rewrite to `next.config.js` in case there is any `next/link` pointing to /umsoknir/:slug. It should redirect to https://island.is/umsoknir/:slug as if it were an [external url](https://nextjs.org/docs/api-reference/next.config.js/rewrites#rewriting-to-an-external-url).

## Checklist:


oddsson:
I'm confused about two things

1. Why do we want to make this button "non-tabbable" via the `skipTab` prop?
2. Why do we need the `openAsExternalLink` props? I gather from the description that this bug is happening because the link is currently being prefetched so shouldn't passing `prefetch={false}` be enough to fix this?

stjanilofts:
1. `skipTab` is used to make the anchor itself non-tababble which is useful for when the child component (such as a button) is tabbable already (see here: https://island.is/rikisborgararettur-fyrir-barn-1964-1982). The way it is now is you can "double-tab" the same link. There is probably a better way to handle this (this is unrelated to the process link bug though).
2. I'm not sure if setting `prefetch={false}` will fix it. I think it will handle the link the same way just without trying to prefetch on hover. For these links (https://island.is/umsoknir/*, which next/link tries to find internally and gets 404 article not found) we need to avoid using next/link. Adding `openAsExternalLink` is for forcing the link to be opened in a new tab.

stjanilofts:
No need to change Link actually, only made change to process entry link. Also added a rewrite to next.config.js in case there is any next/link pointing to /umsoknir/:slug. It should redirect to https://island.is/umsoknir/:slug as if it were an external url.

kksteini (approved):
LGTM

oddsson (approved):

PR #6012, opened 2021-12-15
chore(judicial-system): Remove all code related to session arrangment REMOTE_SESSION

oddsson:
# Remove all code related to session arrangement REMOTE_SESSION 

https://app.asana.com/0/1199153462262248/1201310419657908/f

## What

Remove all code related to session arrangement REMOTE_SESSION 

## Why

It's not being used and won't be added later.

## Checklist:


apps/judicial-system/backend/migrations/20211220090749-update-case.js
```diff
@@ -0,0 +1,36 @@
...
+    // replaceEnum does not support transactions
+    return replaceEnum({
+      queryInterface,
+      tableName: 'case',
+      columnName: 'session_arrangements',
+      newValues: [
+        'ALL_PRESENT',
+        'ALL_PRESENT_SPOKESPERSON',
+        'PROSECUTOR_PRESENT',
+      ],
```
gudjong, line 16: This migration will fail because some cases have "REMOTE_SESSION" assigned to "session_arrangements".

I suggest starting the migration by replacing "REMOTE_SESSION" entries with "PROSECUTOR_PRESENT". That way, the enum modification will work and the rest of the code changes will make sense, including pdf document generation.

gudjong (changes requested):

apps/judicial-system/backend/migrations/20211220090749-update-case.js
oddsson: Awesome! Done :)

apps/judicial-system/backend/migrations/20211220090749-update-case.js
```diff
@@ -0,0 +1,44 @@
+'use strict'
+
+const replaceEnum = require('sequelize-replace-enum-postgres').default
+
+module.exports = {
+  up: (queryInterface) => {
+    // replaceEnum does not support transactions
+    return queryInterface.sequelize.transaction(() =>
```
gudjong, line 8: This should do the job. The transaction is redundant here, however, as it is not used.

Consider removing the transaction.

gudjong (changes requested):

apps/judicial-system/backend/migrations/20211220090749-update-case.js
oddsson: Awesome, fixed :)

gudjong (approved):

PR #6021, opened 2021-12-16
feat(application): Adding quality photo and jurisdiction providers to shared providers

alexandradilja:
# ...

Moving qualityPhoto Provider and Jurisdiction providers to shared providers. P-merki application will be using them as well as the driving license application. Changing them to shared providers will reduce code duplicates. Has been tested well locally.

## Checklist:


albinagu (approved):

libs/application/data-providers/src/providers/QualityPhotoProvider.ts
```diff
@@ -3,18 +3,19 @@ import {
...
   type = 'QualityPhotoProvider'
 
   async provide(application: Application) {
     const fakeData = application.answers.fakeData as
-      | DrivingLicenseFakeData
+      | DataProviderFakeData
       | undefined
 
+    // To use fake data for the quality photo provider take a look at the implementation in libs/application/templates/driving-license/src/forms/application.ts
     if (fakeData?.useFakeData === YES) {
```
baering, line 19: Would be good to make sure that `useFakeData === YES` does not work on production.

baering (approved):

libs/application/data-providers/src/providers/QualityPhotoProvider.ts
alexandradilja: This should be addressed now

PR #6025, opened 2021-12-16
fix(auth-public-api): Filter out delegations with scopes that or no longer allowed for delegations

saevarma:
https://www.notion.so/Filter-out-saved-delegation-scopes-which-aren-t-supported-anymore-da60974f61cd4fbab1447c0612c89247

## What

Filtering out delegation where scope is no longer allowed for delegations.

## Why

To prevent delegations from showing up in the menu.

## Screenshots / Gifs

## Checklist:


libs/auth-api-lib/src/lib/services/delegations.service.ts
```diff
@@ -419,15 +434,21 @@ export class DelegationsService {
           model: DelegationScope,
           required: true,
           where: this.getScopeValidWhereClause(DelegationValidity.NOW),
+          include: [
+            {
+              model: ApiScope,
+              as: 'apiScope',
+              required: true,
```
eirikurn, line 441: Is this `required` redundant?

eirikurn (approved):
Great!

libs/auth-api-lib/src/lib/services/delegations.service.ts
saevarma: Yes actually, this and the one above in L#435.

PR #6026, opened 2021-12-16
fix(service-portal): islykill + userprofile services.

thordurhhh:
## What

- Get userprofile returns data from combined userprofile and islyklar data.
- Create/edit islyklar data is a separate service from userprofile as we currently need data from both services and will not be combining the create/put to not over complicate the create+put of each individual service. These will be kept separate.
- Cleanup after combining services, deleting previous islykill resolver + islykill screen.
- Feature flag for Create and Update user profile
- CRU for bank info

This is still behind a feature flag.

## Why

- Keeping separate services for easier migration later on.
- Cleanup of previous addition.

## Checklist:


thordurhhh:
Adding @eirikurn as a reviewer as well on this, since this is something we had a brief meeting about right before the Christmas holidays.

libs/api/domains/user-profile/src/lib/userProfile.service.ts
```diff
@@ -86,15 +81,38 @@ export class UserProfileService {
...
-    return await this.userProfileApiWithAuth(user)
+
+    const feature = await this.featureFlagService.getValue(
+      Features.personalInformation,
+      false,
+      user,
+    )
+
+    if (!feature) {
+      handleError({ status: 'User profile create is feature flagged for user' })
```
eirikurn, line 100: Should not throw an error? If the feature is disabled we still want to update the locale.

libs/api/domains/user-profile/src/lib/userProfile.service.ts
```diff
@@ -104,13 +122,40 @@ export class UserProfileService {
...
     }
+
+    const feature = await this.featureFlagService.getValue(
+      Features.personalInformation,
+      false,
+      user,
+    )
+
+    if (!feature) {
+      handleError({ status: 'User profile update is feature flagged for user' })
```
eirikurn, line 142: Same here?

libs/api/domains/user-profile/src/lib/userProfile.service.ts
```diff
@@ -104,13 +122,40 @@ export class UserProfileService {
...
+      user.nationalId,
+    )
+
+    // if email exists and *only* mobile is updated, email will become undefined. And vice-versa.
+    await this.islyklarService
+      .updateIslykillSettings(user.nationalId, {
+        email: input.email || islyklarData.email,
+        mobile: input.mobilePhoneNumber || islyklarData.mobile,
+        canNudge: input.canNudge,
+        bankInfo: input.bankInfo,
```
eirikurn, line 155: Should `bankInfo` only be in update, not create?

libs/service-portal/settings/personal-information/src/screens/EditNudge/EditNudge.tsx
```diff
@@ -19,39 +22,30 @@ import {
...
-          input: {
-            email: settings?.email,
-            mobile: settings?.mobile,
-            canNudge: formData.nudge,
-          },
-        },
-      })
+      if (settings) {
+        await updateUserProfile({
+          canNudge: formData.nudge,
```
eirikurn, line 42: Does the UserProfile and Islykill APIs support partial updates? We wouldn’t want to accidentally clear saved locale/email/phone/bankinfo.

In standard REST APIs, PUT requests don’t support partial updates, and require all fields to be included (compared with PATCH).

libs/api/domains/islykill/src/lib/islykill.module.ts
```diff
@@ -6,15 +6,14 @@ import {
...
-import { MainResolver } from './graphql/main.resolver'
 import { IslykillService } from './islykill.service'
 
 @Module({})
 export class IslykillModule {
   static register(config: IslykillApiModuleConfig): DynamicModule {
     return {
       module: IslykillModule,
-      providers: [MainResolver, IslykillService],
+      providers: [IslykillService],
```
eirikurn, line 16: If this domain doesn’t have a resolver anymore, we should get rid of it and either move the service to the client (if it is useful for “everyone” integrating the client) or just to the UserProfile domain (if it’s mainly useful there).

libs/api/domains/user-profile/src/lib/userProfile.service.ts
```diff
@@ -59,18 +46,26 @@ export class UserProfileService {
         nationalId: user.nationalId,
       })
 
-      const islyklarData = await this.getIslyklarData(user.nationalId)
+      const islyklarData = await this.islyklarService.getIslykillSettings(
```
eirikurn, line 49: During the last meeting we discussed that it would be useful to add this logic to the UserProfile service so other “User Profile” consumers don’t need to integrate directly with the temporary Islykill service.

I guess it’s not an issue for the Application System, which currently requests the user profile through the GraphQL API (and gets this logic). But we also need this in the IDP so it can put these claims into ID tokens and userinfo responses. 

Probably not the end of the world to integrate it directly in that one place. Not a blocker for this PR at least.

eirikurn:
Almost there. Great job :)

libs/api/domains/user-profile/src/lib/userProfile.service.ts
thordurhhh: Yeah, whoops, this should definitely not throw an error.

libs/api/domains/user-profile/src/lib/userProfile.service.ts
thordurhhh: Yeah, so the bankInfo is not in create with the iskykill service. When a user is created the service will look for an existing bankInfo with Þjóðskrá. If it exists, the bankInfo will be updated "automatically".

If it does not exist, the user can edit on the "persónuupplýsingar" page.

While onboarding the user, the UI will only ask for email, phone + language.

libs/api/domains/islykill/src/lib/islykill.module.ts
thordurhhh: Sounds good. 

I went with placing it with the userprofile. I think it's mainly (only?) going to be used there.

disaerna (approved):

thordurhhh:
all review comments should have been addressed now.

thordurhhh:
@eirikurn - any chance of a re-review on this PR 🙏

libs/api/domains/user-profile/src/lib/userProfile.service.ts
```diff
@@ -84,26 +81,84 @@ export class UserProfileService {
...
+    if (feature) {
+      await this.islyklarService
+        .createIslykillSettings(user.nationalId, {
+          email: input.email,
+          mobile: input.mobilePhoneNumber,
+          canNudge: input.canNudge,
+        }) // Current version does not return the newly created user in the response.
+        .catch(handleError)
+    } else {
+      logger.error('User profile create is feature flagged for user')
```
eirikurn, line 108: Should be info/warning to not raise alarms in production if someone changes their locale. Same in the other mutation.

eirikurn (approved):
Great! One minor comment. Approving without automerge.

libs/api/domains/user-profile/src/lib/userProfile.service.ts
thordurhhh: Yes! That makes sense. 🙈

PR #6027, opened 2021-12-16
fix(auth): Filter expired delegations from edit screen

eirikurn:

https://www.notion.so/In-delegation-edit-screen-handle-errors-from-update-mutation-somehow-72893f0edf9a4b16af2d0a84019db1e6
https://www.notion.so/In-delegation-edit-screen-don-t-mark-expired-scopes-as-checked-98ff8f6cfd8643a6a044a8bc2a07c6a8

## What

Filter expired delegations from edit screen. Also add error handling

## Why

To reduce confusing states and errors in the access control UI.

## Checklist:


apps/services/auth-public-api/src/app/modules/delegations/meDelegations.controller.ts
```diff
@@ -132,6 +132,14 @@ export class MeDelegationsController {
     type: 'string',
     description: 'Delegation ID.',
   })
+  @ApiQuery({
+    name: 'valid',
```
darrikonn, line 136: renaming this to `validity` makes more sense IMO ... `valid` almost sounds like a boolean.

darrikonn: ^ or anything that would describe it better

eirikurn: This query argument is already used in other AuthPublicAPI endpoints, which should be consistent. It would be a bigger change to rename it everywhere.

Not sure if this is a big issue, since the argument is documented in swagger as an enum rather than a boolean, and validated properly.

darrikonn (approved):

valurefugl (approved):

PR #6028, opened 2021-12-16
fix(sp): When switching users in service portal, redirect to the front page

eirikurn:

https://www.notion.so/Villa-egar-skipt-er-um-pers-nu-1d2bc2f7b48d4471a2936ad9f51becad

## What

See title

## Why

Since many times, the user may not have access to the page they were on before switching. This is the simplest fix for now, but not ideal UX-wise.

Later, it would be nice to check after switching users if they have access to the current page, and only redirect if not. But that's tricky, especially with routes that depend on data access.

## Checklist:


thordurhhh (approved):
LGTM

saevarma (approved):
LGTM

PR #6032, opened 2021-12-17
feat(accident-notification): Use hash do determine if attachment has been sent before

obmagnusson:
# ...

## What

In attachment review the form was sending to the external api other documents on each submit. Use hashes by filecontent to determine if the file has been sent and does not send them.

## Why

Reviewers at SÍ where complaining that they have to go over the same document over and over again each time a user added a document

## Screenshots / Gifs

## Checklist:


libs/application/template-api-modules/src/lib/modules/templates/accident-notification/attachments/attachment.utils.ts
```diff
@@ -0,0 +1,97 @@
...
+export const attachmentStatusToAttachmentRequests = (
+  receivedAttachments?: AccidentNotificationAttachmentStatus,
+): AccidentNotificationAttachmentGatherRequest[] => {
+  if (!receivedAttachments) return allAttachmentRequestConfig.requests
+
+  const attachmentRequests: AccidentNotificationAttachmentGatherRequest[] = []
+
+  if (
+    !receivedAttachments.InjuryCertificate &&
+    receivedAttachments.InjuryCertificate != null
```
bensijons, line 29: Couldn't you just do this? I feel like it is more readable
```suggestion
     receivedAttachments?.InjuryCertificate === false
```

libs/application/template-api-modules/src/lib/modules/templates/accident-notification/attachments/attachment.utils.ts
```diff
@@ -0,0 +1,97 @@
...
+  ) {
+    attachmentRequests.push(injuryCertificateRequest)
+  }
+  if (
+    !receivedAttachments.ProxyDocument &&
+    receivedAttachments.ProxyDocument != null
```
bensijons, line 35: same here

libs/application/template-api-modules/src/lib/modules/templates/accident-notification/attachments/attachment.utils.ts
```diff
@@ -0,0 +1,97 @@
...
+  ) {
+    attachmentRequests.push(powerOfAttorneyRequest)
+  }
+  if (
+    !receivedAttachments.PoliceReport &&
+    receivedAttachments.PoliceReport != null
+  ) {
+    attachmentRequests.push(policeReportRequest)
+  }
+  if (!receivedAttachments.Unknown && receivedAttachments.Unknown != null) {
```
bensijons, line 45: aand here

libs/application/template-api-modules/src/lib/modules/templates/accident-notification/attachments/attachment.utils.ts
obmagnusson: Well the receivedAttachement object is not null it is the InjuryCertificate that the null check is for.

obmagnusson: Tests fail if i change the implementation to this suggestion

bensijons (approved):

PR #6057, opened 2021-12-21
feat(core): Add "AutoAuth" functionality in Enhanced Fetch

eirikurn:
## What

You can now configure Enhanced Fetch to automatically perform OAuth2 Client Credential Grant and Token Exchange against our IDP.

[Docs](https://github.com/island-is/island.is/blob/ca77419a07a3a937d61b94a8022f79376c0764fb/libs/clients/middlewares/README.md#authorization)

## Why

To reduce copied boilerplate, cache tokens and configure scopes and clients more securely.

## Checklist:


libs/clients/middlewares/src/lib/withAutoAuth.ts
```diff
@@ -0,0 +1,244 @@
...
+    if (result.token_type !== 'Bearer') {
+      throw new Error(
+        `Fetch (${name}): Token exchange failed, invalid token type (${result.token_type})`,
+      )
+    }
+
+    if (
+      isTokenExchange &&
+      result.issued_token_type !=
+        'urn:ietf:params:oauth:token-type:access_token'
```
saevarma, line 220: Shouldn't the `issued_token_type` match the `requested_token_type`? So if `requestActorToken` is `true` then it would be `islandis:oauth:token-type:actor-access-token`?

Also we should maybe have a enums for `TokenType` and `GrantType` as these uris are not so nice copy-paste around. i.e.

```
export enum TokenType { 
  AccessToken = 'urn:ietf:params:oauth:token-type:access_token',
  ActorAccessToken = 'islandis:oauth:token-type:actor-access-token'
}

export enum GrantType {
  TokenExchane = 'urn:ietf:params:oauth:grant-type:token-exchange',
  ClientCredentials = 'client_credentials',
  AuthorizationCode = 'code'
}
```

libs/clients/middlewares/test/setup.ts
```diff
@@ -0,0 +1,91 @@
...
+export const fakeResponse = (...args: ConstructorParameters<typeof Response>) =>
+  new Response(...args)
+
+export const fakeAuthResponse = (override?: Partial<TokenResponse>) => {
+  const response: TokenResponse = {
+    expires_in: 60,
+    scope: 'testScope',
+    token_type: fakeAuthentication.split(' ')[0],
+    access_token: fakeAuthentication.split(' ')[1],
+    issued_token_type: 'urn:ietf:params:oauth:token-type:access_token',
```
saevarma, line 36: Missing test cases for `actor_access_token`?

saevarma (commented):
Amazing! Just two minor comments about the token types.

Rikkiabb (approved):
👌

PR #6062, opened 2021-12-21
chore(judicial-system): Improved error responses

gudjong:
# Improved error responses

https://app.asana.com/0/1199153462262248/1201515654267416/f

## What

- Redesigns error responses in the judicial system.
- Modifies log levels.
- Uses shared exception filters in backend, api and xrd-api.
- TODO later: replace HttpExceptions with custom ProblemErrors.

## Why

Clearer logs and easier incident analysis.

## Checklist:


libs/nest/problem/src/lib/problem.module.ts
```diff
@@ -25,6 +26,22 @@ import { ApolloErrorFilter } from './apolloError.filter'
...
+    return {
+      module: ProblemModule,
+      providers: [
+        {
+          provide: PROBLEM_OPTIONS,
+          useValue: options,
+        },
+      ],
+    }
+  }
```
eirikurn, line 46: 😘

libs/nest/problem/src/lib/base-problem.filter.ts
```diff
@@ -37,6 +47,12 @@ export abstract class BaseProblemFilter implements ExceptionFilter {
     const response = ctx.getResponse<Response>()
     const problem = this.getProblem(error)
 
+    if (problem.status && problem.status < 500) {
+      this.options.logAllErrors && this.logger.info(error)
+    } else {
+      this.logger.error(error)
+    }
```
eirikurn, line 54:
```suggestion
    if (problem.status && problem.status >= 500) {
      this.logger.error(error)
    } else if (this.options.logAllErrors) {
       this.logger.info(error)
    }
```

Not a fan of `condition && doStuff` for clarity's sake. Also think it makes sense to reverse the condition to not log all problems with `!problem.status` as errors.

Same above.

libs/nest/problem/src/lib/problem.module.ts
eirikurn: Oh, and can you please add a test for this logic? 🙏

eirikurn (commented):
Really nice and clean! Thank you. Just a couple of minor comments.

libs/nest/problem/src/lib/base-problem.filter.ts
gudjong: Good points - will change.

apps/judicial-system/backend/src/app/modules/case/test/requestRulingSignature.spec.ts
```diff
@@ -69,12 +58,9 @@ describe('CaseController - Request ruling signature', () => {
       then = await givenWhenThen(caseId, user, theCase)
     })
 
-    it('should initiate signing', () => {
-      expect(then.result.statusCode).toBe(201)
-    })
-
     it('should return a control code and a document token', () => {
-      expect(then.result.message).toEqual({
+      console.log(then)
```
oddsson, line 62:
```suggestion
```

oddsson (changes requested):

oddsson (approved):

libs/nest/problem/src/lib/base-problem.filter.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+      // Act
+      await request()
+
+      // Assert
+      expect(infoLog).not.toHaveBeenCalled()
+      expect(errorLog).not.toHaveBeenCalled()
+    })
+  })
+
+  describe('default loggging', () => {
```
eirikurn, line 61:
```suggestion
  describe('with logAllErrors', () => {
```

eirikurn (commented):
Looks awesome! Just one minor comment.

pshomov (approved):
LGTM

libs/nest/problem/src/lib/base-problem.filter.spec.ts
gudjong: Thank you for spotting this

eirikurn (approved):

PR #6065, opened 2021-12-21
feat(user-notification): ingress for x-road

arnorhs:
## Checklist:

### DIFF
--- a/apps/services/user-notification/infra/user-notification.ts
+++ b/apps/services/user-notification/infra/user-notification.ts
@@ -16,6 +16,17 @@ export const userNotificationServiceSetup = (): ServiceBuilder<'user-notificatio
     })
     .liveness('/liveness')
     .readiness('/liveness')
+    .ingress({
+      primary: {
+        host: {
+          dev: 'user-notification-xrd',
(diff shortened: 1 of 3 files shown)


apps/services/user-notification/infra/user-notification.ts
```diff
@@ -16,6 +16,17 @@ export const userNotificationServiceSetup = (): ServiceBuilder<'user-notificatio
     })
     .liveness('/liveness')
     .readiness('/liveness')
+    .ingress({
+      primary: {
+        host: {
+          dev: 'user-notification-xrd',
```
dabbeg, line 22: I am wondering about the naming convention. I assume xrd means X-Road.

Will this service be exposed through xroad?

arnorhs: xrd stands for x-road, not sure where the naming comes from but we have another internal ingress targetted for x-road with a similar name

dabbeg: For reference this was resolved on slack.

dabbeg (approved):

PR #6076, opened 2021-12-22
feat(web): Project Page / Link Groups in navigation

siggifv:
# Project Page / Link Groups in navigation

## What

Add support for nested navigation links (2nd level navigation) via Link Groups in Project Page navigation menu, similar to what we have in Organization Pages.

Additionally, we've made some improvements for the `activeItemTitle` value logic in the Navigation on both the Project Page and the Organization Page.

## Why

In order to support 2 levels of navigation items, as the current functionality (using only Links) only supports one level.

## Screenshots / Gifs

island-project-page-link-groups-support

## Checklist:


apps/web/screens/Project/Project.tsx
```diff
@@ -100,6 +107,23 @@ const ProjectPage: Screen<PageProps> = ({ projectPage, news, namespace }) => {
...
+      active: router.asPath === primaryLink.url,
+      items: childrenLinks.map(({ text, url }) => ({
+        title: text,
+        href: url,
+      })),
+    }),
+  )
+
+  const navigationData: NavigationData = {
+    title: n('navigationTitle', 'Efnisyfirlit'),
```
arnorhs, line 123: is `activeItemTitle` never set to something?

arnorhs (changes requested):

apps/web/screens/Project/Project.tsx
siggifv: Good catch. It's never set, but we should be setting it, in order to improve the UX on mobile. We'll fix this.

ingolfurs (approved):

apps/web/components/Organization/Wrapper/OrganizationWrapper.tsx
```diff
@@ -285,15 +284,27 @@ export const OrganizationWrapper: React.FC<WrapperProps> = ({
...
   const secondaryNavList: NavigationItem[] =
     organizationPage.secondaryMenu?.childrenLinks.map(({ text, url }) => ({
       title: text,
       href: url,
-      active: Router.asPath === url,
+      active: router.asPath === url,
     })) ?? []
 
+  // Keep track of what navigation link is active and update it when the browser url changes
+  const activeNavigationItem = useMemo(() => {
```
arnorhs, line 297: seems like with this approach you keep iterating over the whole list and children even if you've found the corresponding item.

maybe a normal `for (const x of z) {` loop would be better suited for the early exit?

also, should `activeNavigationItem` be named `activeNavigationItemTitle`

apps/web/screens/Project/Project.tsx
```diff
@@ -108,20 +109,42 @@ const ProjectPage: Screen<PageProps> = ({ projectPage, news, namespace }) => {
...
-      href: primaryLink.url,
-      active: router.asPath === primaryLink.url,
-      items: childrenLinks.map(({ text, url }) => ({
-        title: text,
-        href: url,
-      })),
-    }),
+    ({ primaryLink, childrenLinks }) => {
+      let isAnyChildActive = false
+      const items = childrenLinks.map(({ text, url }) => {
```
arnorhs, line 114: a `.map()` function with side effects is generally considered an anti-pattern. i wonder if you'd actually just want to create either some helpers to move these "is active" logics to some shared place (even if in the same file)

and it might also be helpful to actually create a more useful type to begin with for all these items of things with a title and an active state.
```
interface NavItemWithActiveState extends NavigationItem {
  isActive: boolean
}
```
and then you could just map all the items to this type in a single pass.

this means that eg. the logic here would seem much simpler to read at least

arnorhs (changes requested):

apps/web/components/Organization/Wrapper/OrganizationWrapper.tsx
siggifv: Good point. This has been addressed accordingly.

apps/web/screens/Project/Project.tsx
siggifv: These map calls were in fact both mapping the items and calculating the active flag at the same time, but not necessarily easy to read through. This logic has now been extracted to helpers to improve readability of the code.  The `NavigationItem` model already contains an `active` field, so there's no need to create such interface as you suggested.

arnorhs (approved):

PR #6083, opened 2021-12-22
fix(skilavottord): added check on vehicle before recycling

qdd1:
# ...

## What

## Why

## Screenshots / Gifs

## Checklist:


apps/skilavottord/ws/src/app/modules/recyclingRequest/recyclingRequest.resolver.ts
```diff
@@ -4,8 +4,9 @@ import { Query, Resolver, Args, Mutation } from '@nestjs/graphql'
 import type { Logger } from '@island.is/logging'
 import { LOGGER_PROVIDER } from '@island.is/logging'
 
-import { Authorize, Role } from '../auth'
+import { Authorize, Role, CurrentUser } from '../auth'
 import { VehicleModel } from '../vehicle'
+import type { User } from '../auth'
```
darrikonn, line 9: it's a class, not a type.

apps/skilavottord/ws/src/app/modules/recyclingRequest/recyclingRequest.resolver.ts
```diff
@@ -33,7 +34,6 @@ export class RecyclingRequestResolver {
     return res
   }
 
-  @Authorize({ roles: [Role.developer, Role.recyclingFund] })
```
darrikonn, line 36: This would expose every single recycling requests regardless of if a user should be able to view it or not.

Would rather create a new query that exposes a single `RecyclingRequestModel` and verifies that the user owns the car.

darrikonn (changes requested):
MIssing same checks for `createSkilaVottorVehicle(Owner?)`

apps/skilavottord/ws/src/app/modules/recyclingRequest/recyclingRequest.resolver.ts
qdd1: Finished to implement a new query (in resolver and service, not frontend ). But I'm wondering if we should put a check on old query: whether user owns the vehicle, return only last recyclingRequest, else return all. If we do this way, we could skip new query and we dont have to change codes in frontend.  @darrikonn what do you think?

darrikonn: Makes more sense to have another query IMO

apps/skilavottord/ws/src/app/modules/recyclingRequest/recyclingRequest.resolver.ts
```diff
@@ -64,13 +82,15 @@ export class RecyclingRequestResolver {
...
   async createSkilavottordRecyclingRequest(
+    @CurrentUser() user: User,
     @Args({ name: 'requestType', type: () => RecyclingRequestTypes })
     requestType: RecyclingRequestTypes,
     @Args('permno') permno: string,
     @Args('nameOfRequestor', { nullable: true }) name: string,
     @Args('partnerId', { nullable: true }) partnerId: string,
   ): Promise<typeof RecyclingRequestResponse> {
     return this.recyclingRequestService.createRecyclingRequest(
+      user,
```
darrikonn, line 93: This solution isn't very intuitive... same checks are being done everywhere.

```typescript
const vehicle = await this.recyclingRequestService.getUsersVehicle(permno, user.nationalId)

if (!vehicle) {
  throw new NotFoundError(`Vehicle does not exist`) // or BadRequest ... or Forbidden
}

return this.recyclingRequestService.createRecyclingRequest(
  vehicle,
  requestType,
  name,
  user.partnerId
)
```


^ btw. here's another breaking issue ... this mutation shouldn't allow `partnerId` as an argument (needs to be removed) ... this would allow anyone to set any partner id.

Same with the name ... this is the user's name. Remove that one as well.



In addition ... doesn't this need check on the `requestType` ... should a citizen be able to set any requestType?

apps/skilavottord/ws/src/app/modules/recyclingRequest/recyclingRequest.resolver.ts
```diff
@@ -33,7 +33,25 @@ export class RecyclingRequestResolver {
     return res
   }
 
-  @Authorize({ roles: [Role.developer, Role.recyclingFund] })
+  @Query(() => [RecyclingRequestModel])
+  async skilavottordUserRecyclingRequest(
+    @CurrentUser() user: User,
+    @Args('permno') perm: string,
+  ): Promise<RecyclingRequestModel[]> {
+    const res = await this.recyclingRequestService.findRecyclingRequestWithPermno(
```
darrikonn, line 41: Why the logging?


The name `findRecyclingRequestWithPermno` doesn't imply that this is for a single user.

apps/skilavottord/ws/src/app/modules/recyclingRequest/recyclingRequest.service.ts
```diff
@@ -110,6 +120,31 @@ export class RecyclingRequestService {
...
   }
 
+  async findRecyclingRequestWithPermno(
+    permno: string,
+    user: User,
+  ): Promise<RecyclingRequestModel[]> {
+    this.logger.info(
+      `---- Starting findRecyclingRequestWithPermno for ${permno} ----`,
+    )
+    const carList = await this.samgongustofaService.getVehicleInformation(
```
darrikonn, line 130: `getVehicleInformation` implies that it needs permno to read it ... but this is actually `getUsersVehicles`

apps/skilavottord/ws/src/app/modules/recyclingRequest/recyclingRequest.service.ts
```diff
@@ -110,6 +120,31 @@ export class RecyclingRequestService {
...
+      user.nationalId,
+    )
+    if (
+      carList.find((car) => car.isRecyclable && car.permno === permno) ||
+      [Role.recyclingCompany, Role.developer].includes(user.role)
```
darrikonn, line 135: Would like to see this split for readability:
```typescript
const car = carList.find((car) => car.isRecyclable && car.permno === permno)
const hasPermission = [Role.recyclingCompany, Role.developer].includes(user.role)

if (...)
```

But scrap that... the permno should be handled by the `samgongustofaService` .. and the role checks should be handled by the resolver.

apps/skilavottord/ws/src/app/modules/recyclingRequest/recyclingRequest.service.ts
```diff
@@ -268,16 +305,20 @@ export class RecyclingRequestService {
...
       // 3. Then deregistered the vehicle from Samgongustofa
       // 4. Set requestType to 'deregistered'
       // 5. Then call fjarsysla for payment
       // 6. Set requestType to 'paymentInitiated'
       // If we encounter error then update requestType to 'paymentFailed'
       // If we encounter error with 'partnerId' then there is no request saved
-      if (requestType == 'deregistered') {
-        // 1. Check 'pendingRecycle' requestType
+      if (
+        requestType == 'deregistered' &&
```
darrikonn, line 318: `===`

apps/skilavottord/ws/src/app/modules/recyclingRequest/recyclingRequest.service.ts
```diff
@@ -429,6 +470,19 @@ export class RecyclingRequestService {
       }
       // requestType: 'pendingRecycle' or 'cancelled'
       else {
+        // Check if user has the vehicle
+        const carList = await this.samgongustofaService.getVehicleInformation(
+          user.nationalId,
+        )
+        if (!carList.find((car) => car.isRecyclable && car.permno === permno)) {
```
darrikonn, line 477: Abstract this.

apps/skilavottord/ws/src/app/modules/vehicle/vehicle.service.ts
```diff
@@ -63,11 +72,24 @@ export class VehicleService {
...
         `starting creating vehicle with vehicle id - ${vehicle.vehicleId}...`,
       )
+
+      const carList = await this.samgongustofaService.getVehicleInformation(
+        user.nationalId,
+      )
+      if (
+        !(
+          carList.find(
+            (car) => car.isRecyclable && car.permno === vehicle.vehicleId,
```
darrikonn, line 87: Same check again.

Needs better abstraction.

darrikonn (changes requested):

apps/skilavottord/ws/src/app/modules/recyclingRequest/recyclingRequest.service.ts
```diff
@@ -110,6 +120,31 @@ export class RecyclingRequestService {
...
+      carList.find((car) => car.isRecyclable && car.permno === permno) ||
+      [Role.recyclingCompany, Role.developer].includes(user.role)
+    ) {
+      const res = this.findAllWithPermno(permno)
+      return [res[0]]
+    }
+    this.logger.error(
+      `User doesn't have right to fetch the vehicle's information`,
+    )
+    throw new UnauthorizedException(
```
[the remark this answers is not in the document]
qdd1: used BadRequestException

apps/skilavottord/ws/src/app/modules/accessControl/accessControl.resolver.ts
```diff
@@ -26,6 +26,14 @@ export class AccessControlResolver {
     }
   }
 
+  private verifyRecyclingCompanyInput(
+    input: CreateAccessControlInput | UpdateAccessControlInput,
+  ) {
+    if (input.role === Role.recyclingCompany && !input.partnerId) {
+      throw new BadRequestException()
```
darrikonn, line 33: Missing message

apps/skilavottord/ws/src/app/modules/auth/user.model.ts
```diff
@@ -19,4 +19,7 @@ export class User {
 
   @Field(() => Role)
   role!: Role
+
+  @Field({ nullable: true })
```
darrikonn, line 23: Let's remove the field ... but keep the property. Don't need to expose this to GraphQL

apps/skilavottord/ws/src/app/modules/recyclingRequest/recyclingRequest.resolver.ts
```diff
@@ -1,17 +1,18 @@
...
-import { Authorize, Role } from '../auth'
+import { Authorize, Role, CurrentUser, User } from '../auth'
 import { VehicleModel } from '../vehicle'
 import {
   RecyclingRequestModel,
   RecyclingRequestTypes,
   RecyclingRequestResponse,
 } from './recyclingRequest.model'
 import { RecyclingRequestService } from './recyclingRequest.service'
+import { SamgongustofaService } from '../samgongustofa/samgongustofa.service'
```
darrikonn, line 15: Don't dig into the modules for imports ... index.ts is the `api` of the module and things should only be imported from there.
Keep the imports clean and move it upwards below `../vehicle`

darrikonn: `*.module` are the only exceptions.

apps/skilavottord/ws/src/app/modules/recyclingRequest/recyclingRequest.resolver.ts
```diff
@@ -20,29 +21,65 @@ export class RecyclingRequestResolver {
...
   ): Promise<RecyclingRequestModel[]> {
-    const res = await this.recyclingRequestService.findAllWithPermno(perm)
+    const vehicle = await this.samgongustofaService.getUserVehicle(
+      user.nationalId,
+      perm,
+    )
+    const hasPermission = [Role.recyclingCompany, Role.developer].includes(
+      user.role,
+    )
+    if (!vehicle || hasPermission) {
```
darrikonn, line 51: wrong check ... `if (!vehicle || !hasPermission) {}`

apps/skilavottord/ws/src/app/modules/recyclingRequest/recyclingRequest.resolver.ts
```diff
@@ -20,29 +21,65 @@ export class RecyclingRequestResolver {
...
+      this.logger.error(
+        `User does not have permission to call skilavottordUserRecyclingRequest action`,
+      )
+      throw new NotFoundException()
```
darrikonn, line 55: missing message

apps/skilavottord/ws/src/app/modules/recyclingRequest/recyclingRequest.resolver.ts
```diff
@@ -20,29 +21,65 @@ export class RecyclingRequestResolver {
...
+    )
+    this.logger.info(
+      'skilavottordUserRecyclingRequest responce:' +
+        JSON.stringify(userLastRecyclingRequest, null, 2),
+    )
+    return userLastRecyclingRequest
+  }
+
+  @Authorize({
+    roles: [Role.developer, Role.recyclingCompany, Role.recyclingCompany],
```
darrikonn, line 69: recyclingCompany 2x ... is this supposed to be recyclingFund?

apps/skilavottord/ws/src/app/modules/recyclingRequest/recyclingRequest.resolver.ts
```diff
@@ -64,17 +101,15 @@ export class RecyclingRequestResolver {
 
   @Mutation(() => RecyclingRequestResponse)
   async createSkilavottordRecyclingRequest(
+    @CurrentUser() user: User,
     @Args({ name: 'requestType', type: () => RecyclingRequestTypes })
     requestType: RecyclingRequestTypes,
```
darrikonn, line 106: Missing check for this ... should citizen be able to set any kind of recyclingRequestType?

apps/skilavottord/ws/src/app/modules/recyclingRequest/recyclingRequest.service.ts
```diff
@@ -1,13 +1,23 @@
...
 import format from 'date-fns/format'
 
 import type { Logger } from '@island.is/logging'
 import { LOGGER_PROVIDER } from '@island.is/logging'
 
+import { Role, User } from '../auth'
 import { environment } from '../../../environments'
 import { FjarsyslaService } from '../fjarsysla'
 import { RecyclingPartnerService } from '../recyclingPartner'
+import { SamgongustofaService } from '../samgongustofa/samgongustofa.service'
```
darrikonn, line 19: ditto .... samgongustofa is already imported below ... keep it clean :)

apps/skilavottord/ws/src/app/modules/recyclingRequest/recyclingRequest.service.ts
darrikonn: not resolved.

apps/skilavottord/ws/src/app/modules/recyclingRequest/recyclingRequest.service.ts
```diff
@@ -429,6 +462,30 @@ export class RecyclingRequestService {
...
       // requestType: 'pendingRecycle' or 'cancelled'
       else {
+        if (
+          !(requestType === 'pendingRecycle' || requestType === 'cancelled')
+        ) {
+          this.logger.error(
+            `User could not use this requestType: ${requestType}`,
+          )
+          throw new BadRequestException(
+            `User doesn't have right call this action`,
```
darrikonn, line 472: This should be handled in the resolver.

apps/skilavottord/ws/src/app/modules/recyclingRequest/recyclingRequest.service.ts
```diff
@@ -429,6 +462,30 @@ export class RecyclingRequestService {
...
+          )
+        }
+        // Check if user has the vehicle
+        const vehicle = await this.samgongustofaService.getUserVehicle(
+          user.nationalId,
+          permno,
+        )
+        if (!vehicle) {
```
darrikonn, line 480: vehicle should be passed in as arg

darrikonn: already fetched in resolver.

apps/skilavottord/ws/src/app/modules/vehicle/vehicle.resolver.ts
```diff
@@ -54,6 +61,17 @@ export class VehicleResolver {
     @Args('newRegDate') newReg: Date,
     @Args('vinNumber') vin: string,
   ) {
+    const vehicle = await this.samgongustofaService.getUserVehicle(
+      user.nationalId,
+      permno,
+    )
+    if (!(vehicle || Role.developer === user.role)) {
```
darrikonn, line 68: Why `Role.developer === user.role`?

apps/skilavottord/ws/src/app/modules/vehicle/vehicle.resolver.ts
```diff
@@ -54,6 +61,17 @@ export class VehicleResolver {
...
+      this.logger.error(
+        `User does not have right to call createSkilavottordVehicle action`,
+      )
+      throw new NotFoundException()
```
darrikonn, line 72: Missing message

darrikonn (changes requested):

apps/skilavottord/ws/src/app/modules/vehicle/vehicle.resolver.ts
```diff
@@ -16,51 +18,68 @@ export class VehicleResolver {
...
+      throw new NotFoundException(`User does not have this vehicle`)
+    }
+
     const newVehicle = new VehicleModel()
-    newVehicle.vinNumber = vin
-    newVehicle.newregDate = newReg
-    newVehicle.vehicleColor = color
-    newVehicle.vehicleType = type
+    newVehicle.vinNumber = vehicle.vinNumber
+    newVehicle.newregDate = parse(
```
darrikonn, line 74: hmmm ... I think we rather want to save this as a Date in the database ... if we want to parse it for readability, then let's do so in the front-end.

darrikonn (changes requested):
Nice work ... minor comment.

apps/skilavottord/ws/src/app/modules/vehicle/vehicle.resolver.ts
qdd1: We are saving it as a Date but vehicle.firstRegDate is string. That's why we have to parse it here

apps/skilavottord/web/auth/utils.ts
```diff
@@ -16,9 +16,20 @@ export const hasPermission = (page: Page, role: Role) => {
...
+    ],
     citizen: ['myCars', 'recycleVehicle'],
-    recyclingFund: ['recycledVehicles', 'recyclingCompanies', 'accessControl'],
+    recyclingFund: [
+      'recycledVehicles',
+      'recyclingCompanies',
+      'accessControl',
+      'myCars',
+      'recycleVehicle',
+    ],
```
darrikonn, line 32: not used?

apps/skilavottord/ws/src/app/modules/vehicle/vehicle.resolver.ts
darrikonn: aaah I see 👍

darrikonn (approved):
Nice work!

You added some permissions there in `apps/skilavottord/web/auth/utils.ts` that aren't used.
That should be reverted ... but otherwise approving.

essgje (approved):
lftm

PR #6096, opened 2021-12-27
feat(application-templates-driving-license): allow only one active application

arnorhs:
## What

There is now only a single active (draft) application / payment application per user - this is loosely based on the approach taken by the parental leave application.

## Why

We are having issues helping people and debug issues, when people have multiple applications. Currently people are creating new applications when something isn't working, which is fine for the first, prerequisites step, but fails for eg. when you have something payable or ok for submission.

This addresses the issue by allowing you as many times to create a new application in the prerequisite step, as long as you don't have an active application ongoing.

## Todo:
- [ ] add some tests to make sure the state transition goes ok
- [ ] fix issues where people with a saved state don't accidentally get taken too far in the application

## Checklist:


kksteini (approved):
LGTM

PR #6097, opened 2021-12-28
fix(auth-nest-tools): Use x-orginal-forwarded-for for ip in audit logs

dabbeg:
island-is/infrastructure#1003

## What

Use correct source ip of client in audit logs.
X-Real-Ip has either 127.0.0.1 or cloudfront ip

## Checklist:

### DIFF
--- a/libs/auth-nest-tools/src/lib/jwt.strategy.ts
+++ b/libs/auth-nest-tools/src/lib/jwt.strategy.ts
@@ -54,7 +54,11 @@ export class JwtStrategy extends PassportStrategy(Strategy) {
         nationalId: actor.nationalId,
         scope: this.parseScopes(actor.scope),
       },
-      ip: String(request.headers['x-real-ip']) ?? request.ip,
+      ip: String(
+        request.headers['x-original-forwarded-for'] ??
+          request.headers['x-real-ip'] ??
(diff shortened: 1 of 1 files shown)


Snojo (approved):

libs/auth-nest-tools/src/lib/jwt.strategy.ts
```diff
@@ -54,7 +54,11 @@ export class JwtStrategy extends PassportStrategy(Strategy) {
         nationalId: actor.nationalId,
         scope: this.parseScopes(actor.scope),
       },
-      ip: String(request.headers['x-real-ip']) ?? request.ip,
+      ip: String(
+        request.headers['x-original-forwarded-for'] ??
+          request.headers['x-real-ip'] ??
+          request.ip,
+      ),
```
eirikurn, line 61: Can you please post links proving that our infrastructure always sets/overrides these headers? I.e. that our users aren't able to spoof their IP address by just changing a header.

dabbeg: I see that the X-Original-Forwarded-For is overwritten for all ingress routes in the nginx config but it is undocumented so I will not go further with this approach. Too risky.

PR #6098, opened 2021-12-28
feat(user-profile): Client scope migration

rafnarnason:
# ...

## What

creating client and scope as described here https://docs.devland.is/libs/auth-api-lib/seeders#creating-clients

## Why

for internal communications between user profile and user notifications apis

## Screenshots / Gifs

## Checklist:


libs/auth-api-lib/seeders/data/client-scope-user-profile-notifications.ts
```diff
@@ -0,0 +1,17 @@
+import { compose, createClient, createScope } from './helpers'
+
+export const up = compose(
+  createClient({
+    clientId: '@island.is/system/notifications',
+    clientType: 'machine',
+    displayName: 'Notifications machine client',
+    description: '',
+    allowedScopes: ['@island.is/user-profile:admin'],
```
eirikurn, line 9: This will break because the scope doesn't exist yet. addToClient below is enough.

eirikurn (approved):

libs/auth-api-lib/seeders/data/client-scope-user-profile-notifications.ts
rafnarnason: so empty array at that point => "allowedScopes: []," ?

eirikurn: Yeah. I guess you can just the field.

PR #6099, opened 2021-12-29
ci: Retry if cache fails

dabbeg:
island-is/infrastructure#1017

## What

Bake in retry mechanism for the cache actions in our CI

It's not ideal that we need a second step after the cache action to check if it failed or not but composite actions do not yet support continue-on-error.

## Why

Because our cache action was failing with 500 every now and then.

## Checklist:


pshomov (approved):
Excellent!

PR #6105, opened 2022-01-03
fix(user-notification): Change to default swagger path

maggisk:
The openapi json schema path is generated by adding '-json' to the
swagger path. By having the swagger path on the root the openapi schema
is hosted on /-json which doesn't make any sense


rafnarnason (approved):
lgtm

PR #6110, opened 2022-01-03
fix(skilavottord): fix ingress path for ws

darrikonn:
# Fix ingress path for ws

## What

Fix ingress path for ws

## Why

Otherwise next-auth won't be able to handle auth, since it relies on the auth handling being at route `/api/auth/...`

## Checklist:


qdd1 (approved):

dabbeg (approved):
LGTM

PR #6117, opened 2022-01-04
feat(application-templates-driving-license): Changes to the payment field

arnorhs:
## What

Single custom field in the application system for both redirecting, and waiting and submitting the application after payment. Should be easy to extract into a shared field now.

## Why

There's a lot of fragile glue code required to have the redirect flow / UI and the behaviour of things to be nice as it stands. This should make it so it's easier for different teams who end up copying this to re-use the UI etc.

## Checklist:


kksteini (approved):
LGTM

PR #6125, opened 2022-01-04
feat(island-ui): image/logo option for CategoryCard

stjanilofts:
## What

- Add option for adding logo.
- Add option to automatically stack the logo below content once the card reaches a certain width.
- Auto stack option was added in case the logo needs to be visible when the card is very narrow. This way an event listener will not be started for the card needlessly if there is not even an image being used in the first place.
- Add hyphenate option for the heading

## Why

- A new design that adds an image to `CategoryCard`.

## Checklist:


libs/island-ui/core/src/lib/CategoryCard/CategoryCard.tsx
```diff
@@ -41,62 +60,155 @@ const colorSchemes = {
...
-      color={colorScheme}
-    >
-      <Text
-        as={headingAs}
-        variant={headingVariant}
-        color={textColor}
-        truncate={truncateHeading}
-        title={heading}
+    const hasTags = Array.isArray(tags) && tags.length > 0
+    const hasImage = (!!src && !!alt) || customImage
```
kksteini, line 105: If either `src` or `alt` are invalid then `hasImage` will contain whatever value is in `customImage`.
Is this on purpose?

```js
// node terminal session
> customImage = 'https://example.com/catpic.png'
> src = undefined
> alt = 'image of cat'
> const hasImage = (!!src && !!alt) || customImage
> hasImage
'https://example.com/catpic.png'
```

kksteini (approved):
LGTM

oddsson (approved):

barabrian (approved):
LGTM

PR #6131, opened 2022-01-04
chore(assets): Configure token exchange for Properties API

eirikurn:
## What

Refactor assets client to use new config system, auto auth and token exchange.

## Why

This is to limit the spread of third party scopes and audience in our public access tokens. Our tokens should mainly give access to our own resources, especially now that we're issuing delegated tokens.

Let's say island.is has two scopes (read: permissions) which can be granted to another person; `@island.is/assets` and `@island.is/documents`. However, to fulfil requests for "assets", we need a third party scope (eg `@skra.is/properties`). There are a few ways to do this:

1. Let all tokens have `@skra.is/properties` but verify that the token has `@island.is/assets` before calling skra.is. This is not ideal when dealing with your own data, but becomes really bad when you get a delegated token for another user. If you can reach the skra.is API directly then you can access data you're not supposed to have access to.
2. Let tokens which have the `@island.is/assets` scope also have the `@skra.is/properties` scope. This is better, but still can cause a potential privilege escalation since we don't know the full reach of the `@skra.is/properties` scope. We only know the parts of their APIs which we expose. We should not have the power to give users full access to third party scopes on someone's behalf.
3. Only list `@island.is` scopes in our public tokens. Validate them in our APIs and perform token exchange to get a new access token with third party scopes. The user never gets access to these tokens.

With these PRs we're moving from setup 1 to 3.

## Screenshots / Gifs

## Checklist:


libs/clients/assets/src/lib/properties-api.provider.ts
```diff
@@ -0,0 +1,48 @@
+import { Provider } from '@nestjs/common/interfaces/modules/provider.interface'
```
darrikonn, line 1: Name of file not following naming conventions.

ref: https://docs.devland.is/technical-overview/adr/0009-naming-files-and-directories#decision-outcome

libs/nest/config/src/lib/lazy-during-dev.scope.ts
```diff
@@ -0,0 +1,11 @@
+import { Scope } from '@nestjs/common'
```
darrikonn, line 1: Ditto on filename

libs/nest/config/src/index.ts
```diff
@@ -1,4 +1,5 @@
 export { ConfigModule } from '@nestjs/config'
 export { XRoadConfig } from './lib/configurations/XRoadConfig'
 export * from './lib/defineConfig'
+export * from './lib/lazy-during-dev.scope'
```
darrikonn, line 4: Safer to explicitly specify what you're exporting outside of this domain:
```typescript
export { LazyDuringDevScope } from './lib/lazyDuringDev.scope'
```

darrikonn (changes requested):
Looks good. Only comments regarding naming conventions.

libs/clients/assets/src/lib/properties-api.provider.ts
```diff
@@ -0,0 +1,48 @@
...
+      new Configuration({
+        fetchApi: createEnhancedFetch({
+          name: 'clients-assets',
+          ...config.fetch,
+          autoAuth: {
+            mode: 'tokenExchange',
+            ...config.auth,
+          },
+          fetch: (url, init) => {
+            // The Properties API expects two different authorization headers for some reason.
```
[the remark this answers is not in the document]
eirikurn: @hrefnalind Þjóðskrá is planning to ask us to call their webservice with two tokens:

Authorization: Bearer <innskra.island.is access token>
Authorization-Identity: Bearer <Þjóðskrá IDS access token>

It would be good to chat with them and dig into their concerns and see if we can convince them against this path. As I understand it, this PR already addresses the issue they describe since our public tokens won't have access to their scope/service anymore.

libs/clients/assets/src/lib/properties-api.provider.ts
eirikurn: This is a bit annoying since it goes against the NX templates and NestJS conventions.

Our repo is now full of `something.module|resolver|controller|model|dto.ts` files... :/

darrikonn: Yeah. I saw that you changed this to Pascal case, which is incorrect. Pascal case should only be for React components.

The filename should follow camelcase.
From docs: "If that default export is a React Component, or a class, then the file name should be in PascalCase. Otherwise, the filename should be in camelCase."

Your file is not a React component, nor does it contain a default export... so the filename should be in camelCase.

darrikonn: Which would agree with our repo's files:
```
Our repo is already full of something.module|resolver|controller|service|model|dto.ts files
```

thordurhhh (approved):
👍

libs/clients/assets/src/lib/properties-api.provider.ts
eirikurn: Hm, I think the decision may need a review.

First, I feel like the ecosystem is generally moving away from default exports and always use named exports, which are arguably more consistent, are nicer to use in tunnel files and IntelliSense. So we should maybe document that and define naming conventions for modules which contain a "primary export" (as opposed to a collection of similar exports like in a `utils.ts` module).

My intention when originally reviewing that decision was that the filename should be written exactly the same as the "primary export". I feel like "React Components and classes" is not an exhausted list of pascal cased filenames. IMHO it's an unnecessary mental load to think about this arbitrary classification of the primary export when naming the file. I.e. why is the Button file pascal-cased and not files exporting SomeEnum, SomeConstant, SomeType or SomeInterface?

Also, most of the repo files I listed have a "primary export" of a class. So according to the naming convention they should be named AuthService.ts rather than auth.service.ts or Auth.service.ts, right? Meaning we should get rid of all dots from filenames, except some postfixes like `.spec.ts`.

darrikonn (approved):

PR #6144, opened 2022-01-05
chore(api): configcat-node to esbuild

thordurhhh:
## What

See details here:
https://github.com/island-is/island.is/pull/6059


saevarma (approved):
LGTM

PR #6165, opened 2022-01-06
feat(audit): Updating audit to support auditing from auth object without nationalId.

saevarma:
# ☝️

## What

When machine clients calls an API they by default don't have `nationalId` claim. This allows for audit logging also for machine clients without nationalId.

## Why

To support audit logging for machine clients.

## Screenshots / Gifs

## Checklist:

### DIFF
--- a/libs/nest/audit/src/lib/audit.service.spec.ts
+++ b/libs/nest/audit/src/lib/audit.service.spec.ts
@@ -1,13 +1,12 @@
+import { Test, TestingModule } from '@nestjs/testing'
 import { mock } from 'jest-mock-extended'
 
-import { Test, TestingModule } from '@nestjs/testing'
-
-import type { Logger } from '@island.is/logging'
 import { LOGGER_PROVIDER } from '@island.is/logging'
+import { AUDIT_OPTIONS } from '@island.is/nest/audit'
+import type { Logger } from '@island.is/logging'
+import type { Auth, User } from '@island.is/auth-nest-tools'
 
 import { AuditService } from './audit.service'
-import { AUDIT_OPTIONS } from '@island.is/nest/audit'
-import type { Auth, User } from '@island.is/auth-nest-tools'
 import SpyInstance = jest.SpyInstance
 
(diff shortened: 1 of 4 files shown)


libs/nest/audit/src/lib/audit.service.ts
```diff
@@ -1,24 +1,24 @@
...
 import type { Logger } from '@island.is/logging'
 import { LOGGER_PROVIDER } from '@island.is/logging'
-import type { User } from '@island.is/auth-nest-tools'
+import type { Auth } from '@island.is/auth-nest-tools'
 import type { AuditOptions } from './audit.options'
 import { AUDIT_OPTIONS } from './audit.options'
 
 export interface AuditMessage {
-  user: User
+  user: Auth
```
eirikurn, line 13: Ideally all of these are changed to be `auth: Auth` rather than `user: Auth`. But I guess that's a pretty big change across the monorepo. 

Maybe we can support both for backwards compatibility and deprecate `user`? Or just do a big PR :)

saevarma: I actually started doing that, but realised it would be a pretty big PR so a backed away from it. But I'm up for a big PR so just let's do it!

libs/nest/audit/src/lib/audit.service.spec.ts
```diff
@@ -72,6 +72,7 @@ describe('AuditService against Cloudwatch', () => {
   // Cleanup
   afterEach(() => {
     jest.restoreAllMocks()
+    //spy.mockReset()
```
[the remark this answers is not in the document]
saevarma: 🤦‍♂️ yup!

eirikurn (approved):
Removed automerge and approving this pending the comment removal.

Makes sense to get this in to unblock devs, and then rename the field in a separate PR.

saevarma:
> Removed automerge and approving this pending the comment removal.
> 
> Makes sense to get this in to unblock devs, and then rename the field in a separate PR.

On it!

PR #6166, opened 2022-01-06
fix(auth-nest-tools): Use x-forwarded-for in audit logs

dabbeg:
island-is/infrastructure#1003

## Problem

Currently `X-Real-IP` holds the cloudfront ip on production and on dev and staging it holds the value `127.0.0.1` . So all our audit logs have those ips instead of the real client ip. The difference between environments is because linkerd is running on dev and staging but not yet in production. See the example below to see why. 

## What

Use full computed X-Forwarded-For in audit logs.

We will configure our ingress controller to use [https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/configmap/#compute-full-forwarded-for](https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/configmap/#compute-full-forwarded-for) which will append ip's instead of replacing the header.

So it will look like this:
```
     client defined, real client ip, cloudfront ip, linkerd ip
ip: "123.123.123.123, 88.149.58.121, 130.176.222.7, 127.0.0.1"
```

So when we go through our audit logs we will need to parse the correct ip.

## Why

Because there is not a fully supported way of getting the exact client ip with our infrastructure setup. Parsing the right client ip automatically before we set it in the audit log would be error prone.

## Checklist:


libs/auth-nest-tools/src/lib/jwt.strategy.ts
```diff
@@ -54,7 +54,7 @@ export class JwtStrategy extends PassportStrategy(Strategy) {
         nationalId: actor.nationalId,
         scope: this.parseScopes(actor.scope),
       },
-      ip: String(request.headers['x-real-ip']) ?? request.ip,
+      ip: String(request.headers['x-forwarded-for'] ?? request.ip),
```
eirikurn, line 57: The plan was originally to use Express's built in support for parsing X-Forwarded-For:

http://expressjs.com/en/guide/behind-proxies.html

There's already an [environment variable](https://github.com/island-is/island.is/blob/54cb834e0157b87da61932109debdcf1f00ca6db/libs/infra-nest-server/src/lib/bootstrap.ts#L82-L87) set up which we can use to configure this everywhere.

The tricky thing is that all of this needs to also support [forwarding](https://github.com/island-is/island.is/blob/2bf24efe74404baaebcfd53846b52854688cd867/libs/auth-nest-tools/src/lib/auth.middleware.ts#L67) within our infrastructure, which currently uses `x-real-ip`.

We at least need to switch that to set the x-forwarded-for header, but I would really like for us to configure the "trust proxy" value correctly and just use express's calculated `request.ip` here.

Our internal APIs could have `EXPRESS_TRUST_PROXY=1`. But our "public" apps might need some environment-specific value, either the correct number of proxies, or an array with whitelisted IP ranges.

dabbeg: I like the whitelist approach but because the cloudfront ips can change it might be troublesome to hardcode them.

The number approach feels a bit fragile, if either the number of ip's would increase or decrease it would break.
If it would decrease the user could enter any ip in the X-Forwarded-For header and we would use it.

I would choose to log them all for the time being, but I know this is not a great solution for the long term.

eirikurn (approved):
Yeah, I get your point. This is your call. :)

PR #6167, opened 2022-01-06
feat(core): simplify swagger documentation

darrikonn:
# Simplify swagger documentation

## What

Simplify swagger documentation

## Why

> It gets tedious and noisy to read code that has 10 decorators that share similar purpose. Combining those will improve DX

## Screenshots / Gifs

swagger

## Checklist:


libs/nest/swagger/src/lib/documentation.decorator.ts
```diff
@@ -0,0 +1,93 @@
...
+  params,
+  description,
+}: Options): MethodDecorator =>
+  applyDecorators(
+    HttpCode(response.status),
+    ApiNotFoundResponse,
+    ApiInternalServerErrorResponse(),
+    ApiBadRequestResponse({ type: HttpProblemResponse }),
+    ApiForbiddenResponse({ type: HttpProblemResponse }),
+    ApiConflictResponse({ type: HttpProblemResponse }),
```
eirikurn, line 87: Should all operations list all of these responses even though they are never returned?

eirikurn: Note that NotFoundResponse and InternalServerErrorResponse should also include a HttpProblemResponse body.

apps/services/auth-public-api/src/app/modules/delegations/meDelegations.controller.ts
```diff
@@ -123,28 +109,28 @@ export class MeDelegationsController {
...
+    response: { status: 200, type: DelegationDTO },
     description: `Finds a single delegation by ID where the authenticated user is either giving or receiving.
        Does not include delegations from NationalRegistry or CompanyRegistry.`,
+    params: [
+      {
+        name: 'delegationId',
+        type: 'string',
+        description: 'Delegation ID.',
+      },
+    ],
```
eirikurn, line 122: What do you think about saving more space for param/query by doing this:

```
params: {
  delegationId: {
    type: 'string',
    description: 'Delegation ID.',
  }
}
```

I know you're re-using the @nestjs/swagger type, but they could probably still be used like this:

```
params: Record<string, Omit<ApiParamOptions, 'name'>>,
queries: Record<string, Omit<ApiQueryOptions, 'name'>>,
```

apps/services/auth-public-api/src/app/modules/delegations/meDelegations.controller.ts
```diff
@@ -123,28 +109,28 @@ export class MeDelegationsController {
...
+    queries: [
```
eirikurn, line 123: The name `queries` confused me, since I associate the plural form with search/database queries rather than the query string.

Would `query` be ok? Note that in Express, there's `request.params` and `request.query`.

eirikurn (commented):
Doesn't really save much on scrolling, but I think I can grow to prefer this.

apps/services/auth-public-api/src/app/modules/delegations/meDelegations.controller.ts
darrikonn: Definitely..  agree with that.

libs/nest/swagger/src/lib/documentation.decorator.ts
darrikonn: > Should all operations list all of these responses even though they are never returned?

Nah, http conflict does not always apply. But I don't see anything wrong with always applying not found, bad request, forbidden, 500.

eirikurn (approved):
Like it. Do you think we can unit test this decorator logic somehow?

Also curious what @saevarma  thinks :)

valurefugl (approved):

apps/services/auth-public-api/src/app/modules/delegations/meDelegations.controller.ts
```diff
@@ -254,11 +237,9 @@ export class MeDelegationsController {
   @Scopes(AuthScope.writeDelegations)
   @FeatureFlag(Features.customDelegations)
   @Delete(':delegationId')
-  @HttpCode(204)
```
saevarma, line 257: FYI: This is not Swagger related decorator, this actually changes the response status of the operation. I find this a nice feature to let the new `@Documentation` decorator handle this in relation to the documented response status. But we should maybe document this behaviour and usage in the `Nest/Swagger` library's `README`, as all our operation should either way use this new decorator, and therefore use the `response` property of that to control this.

libs/nest/swagger/README.md
```diff
@@ -0,0 +1,7 @@
+# Nest Swagger
```
saevarma, line 1: Maybe add a bit of a description in the `README` what this is and does, and how it should be used with few code samples.

libs/nest/swagger/src/lib/documentation.decorator.ts
```diff
@@ -0,0 +1,107 @@
...
+  request,
+  ...extra
+}: Options): MethodDecorator =>
+  applyDecorators(
+    /* BEGIN DEFAULT DECORATORS */
+    HttpCode(response.status),
+    ApiInternalServerErrorResponse(),
+    ApiNotFoundResponse({ type: HttpProblemResponse }),
+    ApiBadRequestResponse({ type: HttpProblemResponse }),
+    ApiForbiddenResponse({ type: HttpProblemResponse }),
```
saevarma, line 101: I think that the `ApiForbiddenResponse`  should be part of the `getExtraDecorators` and applied when `ApiUnauthorizedResponse` is applied. As the response indicate a forbidden action due to lack of authorization, that is, an authentication token was provided but is missing the required scopes to execute the action. So if the action is unauthenticated it will never return `ApiForbiddenResponse`.

libs/nest/swagger/src/lib/documentation.decorator.ts
```diff
@@ -0,0 +1,107 @@
...
+export const Documentation = ({
+  response,
+  request,
+  ...extra
+}: Options): MethodDecorator =>
+  applyDecorators(
+    /* BEGIN DEFAULT DECORATORS */
+    HttpCode(response.status),
+    ApiInternalServerErrorResponse(),
+    ApiNotFoundResponse({ type: HttpProblemResponse }),
```
saevarma, line 99: This one is a bit tricky. Because a `GET` operation returning list of resources, should never return `404` as it would instead just return empty list (like if no resource is found matching some query params). So the client shouldn't be concerned about `404` response for those endpoints and we shouldn't document it in our OpenAPI doc. But for `GET` with resourceId, `PUT` and `DELETE`  the `404` should be applied.

saevarma (changes requested):
This is a really nice addition to our repo and I believe it will give our developers a better DX documenting their api endpoints.

I just had two concerns about the default decorators.

libs/nest/swagger/src/lib/documentation.decorator.ts
saevarma: See my review comment about the not found and forbidden responses.

libs/nest/swagger/src/lib/documentation.decorator.ts
darrikonn: Great feedback. We can base it on the params decorator. See: 898fc2b9791803ed0aefd3aa01d1df4cbabe0e06

eirikurn: That's nice!

eirikurn (approved):

libs/nest/swagger/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+1. @HttpCode(200)
+2. @ApiInternalServerErrorResponse({ type: HttpProblemResponse })
+3. @ApiBadRequestResponse({ type: HttpProblemResponse })
+4. @ApiOkResponse({ status: 200, type: AnimalDTO })
+5. @ApiQuery({ name: 'query', required: true, schema: { enum: [SearchEnum], default: SearchEnum.query, } })
+6. @ApiParam({ name: 'animalId', type: 'string', description: 'ID of the animal' })
+7. @ApiNotFound({ type: HttpProblemResponse })
+8. @ApiUnauthorizedResponse({ type: HttpProblemResponse }),
+9. @ApiForbiddenResponse({ type: HttpProblemResponse }),
```
saevarma, line 52: It would also add the `@ApiOperation(...)` with the description.

saevarma (approved):
Approved but removed the `automerge` label so you can update the missing decorator in the README and then re-add the `automerge` to merge it.

PR #6170, opened 2022-01-06
chore(web): a11y tweaks

stjanilofts:
- Main content for the sidebar layout used on many pages (especially articles) now has the #main-content id set around the whole section instead of just the content so when "jumping to content" the focus will be from left to right (sidebar first).
- Content links in search result suggestions now work with keyboard when pressing enter and show an underline when highlighted.

Jan-06-2022 16-26-37

## Checklist:


apps/web/components/SearchInput/SearchInput.tsx
```diff
@@ -426,30 +452,35 @@ const Results = ({
                 News[] &
                 SubArticle[])
                 .slice(0, 5)
-                .map((item) => {
+                .map((item, i) => {
                   const { onClick, ...itemProps } = getItemProps({
-                    item: '',
+                    item:
```
arnorhs, line 457: Would it make sense to create a an object for this, rather than doing string interpolation?

eg. 

```typescript
interface SidebarNavAction<T extends ('directNav' | 'submit')> {
  type: T
  href: string
}

---
item: {
  type: 'directNav',
  href: .. etc
}

// and then you could just have something like
const handleResultAction = (item: SiebarNavAction) => {
  if (item.type === 'directNav') {
    // etc
  } else {
    // something submit
  }
}
```

At least then the logic could be a little bit more centralized and you wouldn't have to be comparing and composing strings

arnorhs (changes requested):

apps/web/components/SearchInput/SearchInput.tsx
stjanilofts: I actually tried that but downshift turns whatever `item` is into a string so it ends up being [object Object]... So I ended up going this sort of hacky way.

stjanilofts: Now I see why it ends up being a string 🤦 
I'm gonna unhackify this then 😁

apps/web/components/SearchInput/SearchInput.tsx
```diff
@@ -392,7 +426,10 @@ const Results = ({
                 ? suggestion.replace(search.term, '')
                 : ''
               const { onClick, ...itemProps } = getItemProps({
-                item: suggestion,
+                item: {
+                  type: 'query',
+                  string: suggestion,
+                } as SubmitType,
```
arnorhs, line 432: as doesnt help much here unless getItemProps knows about it

apps/web/components/SearchInput/SearchInput.tsx
```diff
@@ -426,30 +463,36 @@ const Results = ({
...
+                .map((item, i) => {
                   const { onClick, ...itemProps } = getItemProps({
-                    item: '',
+                    item: {
+                      type: 'link',
+                      string: linkResolver(
+                        item.__typename as LinkType,
+                        item.slug.split('/'),
+                      ).href,
+                    } as SubmitType,
```
arnorhs, line 474: same as above

apps/web/components/SearchInput/SearchInput.tsx
```diff
@@ -244,17 +253,34 @@ export const SearchInput = forwardRef<HTMLInputElement, SearchInputProps>(
         id={id}
         initialInputValue={initialInputValue}
         onChange={(q) => {
-          return onSubmit(`${search.prefix} ${q}`.trim() || '')
+          if (q && typeof q === 'object') {
```
arnorhs, line 256: can q ever not be an object?

apps/web/components/SearchInput/SearchInput.tsx
```diff
@@ -244,17 +253,34 @@ export const SearchInput = forwardRef<HTMLInputElement, SearchInputProps>(
...
+            }
+          }
         }}
         onInputValueChange={(q) => setSearchTerm(q)}
         itemToString={(v) => {
-          const str = `${search.prefix ? search.prefix + ' ' : ''}${v}`.trim()
-
-          if (str === 'null') {
-            return ''
+          if (v && typeof v === 'object') {
```
arnorhs, line 273: isn't it always an object?

arnorhs (changes requested):

arnorhs (approved):

Issue #5226, opened 2021-10-08
@island.is/island-ui: Datepicker needs to check if label is in order to place popper


darrikonn:
Setting the `top` css prop for the Datepicker popper should check if a label is set or not. Otherwise the fixed pixels will be off, resulting in this:

![image](https://user-images.githubusercontent.com/5694851/136539612-922f0a2c-c471-4d66-ad02-169b85983e15.png)

PR #6175, opened 2022-01-07
fix(island-ui): fix datepicker without label

miamiultras:
# ...

[@island.is/island-ui: Datepicker needs to check if label is in order to place popper](https://github.com/island-is/island.is/issues/5226)

## What

Refactor datepicker popper to use className instead of globalStyles and fix datepicker without label, it shouldn't happen, but it is possible to add empty string as a label.

## Why

Because it breaks datepicker when there is no label for it

### default datepicker with label
Screenshot 2022-01-07 at 11 13 12

### default datepicker without label
Screenshot 2022-01-07 at 11 12 49

### small datepicker with label
Screenshot 2022-01-07 at 11 13 28

### small datepicker without label
Screenshot 2022-01-07 at 11 12 12

## Checklist:


darrikonn (approved):

stjanilofts (approved):
🆒

PR #6177, opened 2022-01-07
chore(user-notification): removing worker health check

arnorhs:
## What

Removing the health checks for the notification worker

## Why

Because the worker service doesn't require it and is getting shut down over and over

## Checklist:


pshomov (approved):
LGTM

PR #6185, opened 2022-01-10
fix(web): Article organization slug not changing on sync

RunarVestmann:
# Article organization slug not changing on sync

## What

* If an organization slug changes then articles that are referencing that organization still point to the old slug

## Why

* This fixes the problem regarding no articles showing up on the organization services page due to a change of organization slug (since articles will then keep pointing to the old slug)

## Checklist:


JuliusGG (approved):
lgtm

PR #6212, opened 2022-01-11
feat(quality-photo): Provides if photo exists, components queries on demand

alexandradilja:
Quality photo data provider returns if user has a quality photo. The Quality photo then gets the photo when loaded, instead of storing it in the external data.

## Checklist:


libs/application/data-providers/src/providers/QualityPhotoProvider.ts
```diff
@@ -5,36 +5,15 @@ import {
...
 } from '@island.is/application/core'
-import { isRunningOnEnvironment } from '@island.is/shared/utils'
-import { DataProviderFakeData, YES } from '../libs/constants'
 
 export class QualityPhotoProvider extends BasicDataProvider {
   type = 'QualityPhotoProvider'
 
   async provide(application: Application) {
-    // If running locally or on dev allow for fake data
-    if (isRunningOnEnvironment('local') || isRunningOnEnvironment('dev')) {
```
arnorhs, line 16: I don't remember adding this.. checking for application.answers.useFakeData should be enough.. but we don't really want to remove this feature do we?

libs/application/data-providers/src/providers/QualityPhotoProvider.ts
```diff
@@ -48,29 +27,16 @@ export class QualityPhotoProvider extends BasicDataProvider {
     }
 
     const response = await res.json()
-
+    console.log(response)
```
arnorhs, line 30: consoooole

libs/application/templates/driving-license/src/fields/QualityPhoto/hooks/getQualityPhoto.ts
```diff
@@ -0,0 +1,60 @@
+import { Application } from '@island.is/application/core'
+import { isRunningOnEnvironment } from '@island.is/shared/utils'
```
arnorhs, line 2: we should try to avoid using this as much as possible

libs/application/templates/driving-license/src/fields/QualityPhoto/hooks/getQualityPhoto.ts
```diff
@@ -0,0 +1,60 @@
...
+
+type HasQualityPhotoData = {
+  data: {
+    hasQualityPhoto: boolean
+  }
+}
+
+export const getQualityPhoto = (application: Application): QualityPhotoType => {
+  // If running locally or on dev allow for fake data
+  if (isRunningOnEnvironment('local') || isRunningOnEnvironment('dev')) {
```
arnorhs, line 28: ideally this would not be aware of fake data :O

libs/application/templates/driving-license/src/fields/QualityPhoto/hooks/getQualityPhoto.ts
```diff
@@ -0,0 +1,60 @@
...
+    const fakeData = application.answers.fakeData as
```
arnorhs, line 29: using as is an anti pattern.. to enforce type safety for answers, esp. when it comes to undefined etc, we should try to always use
```typescript
const value = getValueViaPath<'yes'|'no'>(application.answers, 'fakeData')
```
the resulting type is 'yes'|'no'|undefined, which ensures that the rest of the code is correct

libs/application/templates/driving-license/src/fields/QualityPhoto/hooks/getQualityPhoto.ts
```diff
@@ -0,0 +1,60 @@
...
+        success: fakeData.qualityPhoto === YES,
+        qualityPhoto:
+          fakeData.qualityPhoto === YES
+            ? `data:image/jpeg;base64,/9j/4AAQSkZJRgABAQAAAQABAAD/2wCEAAQEBAQEBAQEBAQGBgUGBggHBwcHCAwJCQkJCQwTDA4MDA4MExEUEA8QFBEeFxUVFx4iHRsdIiolJSo0MjRERFwBBAQEBAQEBAQEBAYGBQYGCAcHBwcIDAkJCQkJDBMMDgwMDgwTERQQDxAUER4XFRUXHiIdGx0iKiUlKjQyNEREXP/CABEIAAIAAgMBIgACEQEDEQH/xAAUAAEAAAAAAAAAAAAAAAAAAAAH/9oACAEBAAAAAHP/xAAUAQEAAAAAAAAAAAAAAAAAAAAH/9oACAECEAAAADv/xAAUAQEAAAAAAAAAAAAAAAAAAAAG/9oACAEDEAAAAHn/xAAZEAABBQAAAAAAAAAAAAAAAAAAAgMTU5H/2gAIAQEAAT8AjbrTh//EABQRAQAAAAAAAAAAAAAAAAAAAAD/2gAIAQIBAT8Af//EABQRAQAAAAAAAAAAAAAAAAAAAAD/2gAIAQMBAT8Af//Z`
+            : null,
+      }
+    }
+  }
+  if (
+    (application.externalData.qualityPhoto as HasQualityPhotoData)?.data
```
arnorhs, line 45: same with external data and `getValueViaPath` -- try to never use `as`

libs/application/templates/p-sign/src/dataProviders/DoctorsNoteProvider.ts
```diff
@@ -19,25 +19,29 @@ export class DoctorsNoteProvider extends BasicDataProvider {
...
-      return Promise.reject({})
-    }
-
-    return Promise.resolve(response.data.getSyslumennCertificateInfo)
+    return Promise.resolve({
+      nationalId: '0101303019',
+      expirationDate: '2023-05-29T13:52:33.22',
+      releaseDate: '2021-11-22T13:54:07.5',
+    })
+    // const res = await this.useGraphqlGateway(query)
```
arnorhs, line 27: should this be removed? rather than commented out?

arnorhs (changes requested):

libs/api/domains/driving-license/src/lib/graphql/main.resolver.ts
```diff
@@ -27,13 +27,21 @@ const namespace = '@island.is/api/driving-license'
 
 @UseGuards(IdsUserGuard, ScopesGuard)
 @Scopes(ApiScope.internal)
-@Resolver()
+@Resolver(QualityPhoto)
```
eirikurn, line 30: Might want to create a new resolver for QualityPhoto fields.

libs/api/domains/driving-license/src/lib/graphql/models/qualityPhoto.model.ts
```diff
@@ -1,10 +1,16 @@
-import { Field, ObjectType } from '@nestjs/graphql'
+import { Field, ObjectType, ID } from '@nestjs/graphql'
 
 @ObjectType()
 export class QualityPhoto {
+  @Field(() => ID)
```
eirikurn, line 5: I wonder if we can remove this decorator but keep the field for your field resolver. 

Since it feels like internal logic, should always match the authenticated user, and not relevant to the client.

libs/api/domains/driving-license/src/lib/drivingLicense.service.ts
```diff
@@ -219,21 +219,45 @@ export class DrivingLicenseService {
...
+  async getQualityPhoto(
+    nationalId: User['nationalId'],
+  ): Promise<QualityPhotoResult> {
+    const hasQualityPhoto = await this.drivingLicenseApi.getHasQualityPhoto({
+      nationalId,
+    })
+    if (!hasQualityPhoto) {
+      return {
+        nationalId: nationalId,
+        success: hasQualityPhoto,
```
arnorhs, line 252: success should probably just be `true`

arnorhs: suggestion:
```
let hasQualityPhoto
try {
  hasQualityPhoto = await .. etc
} catch (e) {
  return { success: false, errorMessage: e, nationalId }
}

return {
  success: true,
  hasQualityPhoto,
  nationalId,
}
```

libs/api/domains/driving-license/src/lib/graphql/models/qualityPhoto.model.ts
```diff
@@ -1,10 +1,16 @@
-import { Field, ObjectType } from '@nestjs/graphql'
+import { Field, ObjectType, ID } from '@nestjs/graphql'
 
 @ObjectType()
 export class QualityPhoto {
```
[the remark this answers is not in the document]
arnorhs: @eirikurn "quality photo" was the translation picked for "gæðamerkt ljósmynd" .. possibly a better name would be "registrationPhoto" or something like that.. apparently there are also photos that are not "gæðamerktar" and the intent is to distinguish between the two in code as well as for the user

libs/application/templates/p-sign/src/forms/application.ts
```diff
@@ -190,8 +191,8 @@ export const getApplication = (): Form => {
             title: m.qualityPhotoTitle,
             condition: (_, externalData) => {
               return (
-                (externalData.qualityPhoto as QualityPhotoData)?.data
```
arnorhs, line 193: i suggest we use getValueViaPath here as well

arnorhs (changes requested):
tiny tiny comments

libs/api/domains/driving-license/src/lib/graphql/models/qualityPhoto.model.ts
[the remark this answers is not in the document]
alexandradilja: - I removed the `success` field since it was hardly being used. 
- Removed the id field since its possible to use `CurrentUser` decorator in a resolver field
- renamed to `drivingLicenseQualityPhoto`
- Added error messages if QualityPhoto cant be found and there is a warning filed in the ui
- The query takes a second to fetch the qualityPhoto and @albinagu will come in with a follow up pr with a skeleton loader

libs/api/domains/driving-license/src/lib/drivingLicense.service.ts
```diff
@@ -243,20 +243,26 @@ export class DrivingLicenseService {
...
-      nationalId,
-    })
-    if (!hasQualityPhoto) {
+    let hasQualityPhoto
+
+    try {
+      hasQualityPhoto = await this.drivingLicenseApi.getHasQualityPhoto({
+        nationalId,
+      })
+    } catch (e) {
```
eirikurn, line 252: Best practice in error handling is to only catch and handle “expected” issues which we know how to handle, or the User should know about in a special way.

It looks like in this case the only “expected” issue is if there is no quality photo stored (404). In GraphQL land, that could just be represented as `hasQualityPhoto: false`.

If there’s more than one expected issue that can occur, it’s best to have them in an Enum, so the UI as more flexibility in how to display them, eg with translated messages. Only if there is a big variability in “expected” issues which we can’t know about in GraphQL land should we allow some kind of string errorMessage from the external API. But a string “errorMessage” is generally bad since we can’t easily localize it.

ALL other (“unexpected”) errors should **not** be handled in services/resolvers. This error could be anything from a JS syntax error to network issues, authentication issues and service issues. We should never swallow it, hide it or try to recover from it. Instead we need to log these issue and handle them consistently (globally, not in each and every UI/Resolver/Service).

If you catch an error which you don’t expect, you should re-throw it. Then a global API interceptor can log it in a consistent way and send it to the client where it can also be handled in a consistent way (toast or error in external data provider). The ops team will notice the errors in the logs so they can be addressed.

Tldr, I suggest getting rid of errorMessage and just return `hasQualityPhoto: true/false`. Below the 404 case you should just `throw e` again.

libs/api/domains/driving-license/src/lib/graphql/models/qualityPhoto.model.ts
```diff
@@ -2,15 +2,12 @@ import { Field, ObjectType, ID } from '@nestjs/graphql'
 
 @ObjectType()
```
eirikurn, line 3: Should also namespace the “ObjectType” in the schema. Think somehow like this:
```suggestion
@ObjectType(‘DrivingLicenseQualityPhoto’)
```

libs/api/domains/driving-license/src/lib/graphql/models/qualityPhoto.model.ts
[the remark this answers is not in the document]
eirikurn: Nice!

eirikurn:
Almost there. 🤗

libs/api/domains/driving-license/src/lib/graphql/main.resolver.ts
darrikonn: I  thought the syntax would be:
```typescript
@Resolver(() => QualityPhoto)
```

But yeah, agree with Eiki. Should have a new file specific for that resolver if it's resolving the GraphQL type's fields.

libs/api/domains/driving-license/src/lib/graphql/models/qualityPhoto.model.ts
```diff
@@ -1,10 +1,13 @@
...
 export class QualityPhoto {
   @Field()
-  success!: boolean
+  hasQualityPhoto!: boolean
 
   @Field({ nullable: true })
-  qualityPhoto!: string | null
+  qualityPhotoDataUri?: string
```
darrikonn, line 9: Just `dataUri`?
Since the GraphQL type is named QualityPhoto, the fields don't have to be prefixed with the same name.

libs/api/domains/driving-license/src/lib/graphql/qualityPhoto.resolver.ts
```diff
@@ -0,0 +1,32 @@
...
+export * from '@island.is/nest/audit'
+import { QualityPhoto } from './models'
+
+@UseGuards(IdsUserGuard, ScopesGuard)
+@Scopes(ApiScope.internal)
+@Resolver(QualityPhoto)
+export class QualityPhotoResolver {
+  constructor(private readonly drivingLicenseService: DrivingLicenseService) {}
+
+  @ResolveField('qualityPhotoDataUri', () => String, { nullable: true })
```
darrikonn, line 21: The usage of the field resolver can be improved.
So both the parent resolver (`drivingLicenseQualityPhoto`) and the field resolver are performing a service call that **BOTH** make the same API call:
```typescript
const image = await this.drivingLicenseApi.getHasQualityPhoto({
  nationalId,
})
 ```

So under the hood, a single GraphQL query is making two identical API calls.

It would be better if the field resolver would just check if the parent is `null` and skip that check.

**Review by darrikonn — CHANGES_REQUESTED** [review]

File: libs/api/domains/driving-license/src/lib/graphql/models/qualityPhoto.model.ts [hunk 17]
```diff
@@ -1,10 +1,13 @@
-import { Field, ObjectType } from '@nestjs/graphql'
+import { Field, ObjectType, ID } from '@nestjs/graphql'
 
 @ObjectType()
 export class QualityPhoto {
   @Field()
-  success!: boolean
+  hasQualityPhoto!: boolean
```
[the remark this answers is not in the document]
alexandradilja: Yes! The `hasQualityPhoto` check is stored in the `qualityPhotoDataProvider` and the `dataUri` is queried later on in the driving licence and p-sign applications. This is so we don't have to store the image as base64 in the external data.
But an early return from the resolver field if `hasQualityPhoto` is `false` is a very good idea!

libs/api/domains/driving-license/src/lib/drivingLicense.service.ts
```diff
@@ -250,21 +246,11 @@ export class DrivingLicenseService {
...
-        return {
-          hasQualityPhoto: false,
-          errorMessage: 'QualityPhoto not available',
-        }
-      }
-      return {
-        hasQualityPhoto: false,
-        errorMessage: 'Something went wrong',
-      }
+      throw e
```
darrikonn, line 266: Remove try catch ... since it's just throwing the error anyway.

darrikonn: And then you can basically return the promise :)

libs/api/domains/driving-license/src/lib/graphql/qualityPhoto.resolver.ts
```diff
@@ -0,0 +1,35 @@
...
+import { QualityPhoto } from './models'
+
+@UseGuards(IdsUserGuard, ScopesGuard)
+@Scopes(ApiScope.internal)
+@Resolver(QualityPhoto)
+export class QualityPhotoResolver {
+  constructor(private readonly drivingLicenseService: DrivingLicenseService) {}
+
+  @ResolveField('dataUri', () => String, { nullable: true })
+  resolvedataUri(
```
darrikonn, line 22: resolveDataUri

eirikurn (approved):
Sweet! Thanks for cleaning this up :)

darrikonn (approved):
Nice
 LGTM

arnorhs (approved):

PR #6216, opened 2022-01-11
feat(fa): Prereq

Rikkiabb:
# ...
https://app.asana.com/0/1201600280110882/1201636033387155

## What
Sets up prereq step for financial aid in application system so we can know which form the user should route to.
Fixes validation bug in in relationship screen

## Why
We need to change the flow to be able to route the user to the correct page 😭 

## Checklist:


MargretFinnboga (approved):
We gonna be alright 👍

PR #6225, opened 2022-01-12
feat(application-system-api): Application lifecycle

obmagnusson:
# Application lifecycle

## What

Implement deletion of personal data and attachments from applications according their set prune date.

## Why

For applications to able to be compliant to current GDPR regulations the pruneAt needs to actually remove personal data at the set prune date.

## Checklist:


apps/application-system/api/src/app/modules/application/lifecycle/application-lifecycle.service.ts
```diff
@@ -0,0 +1,165 @@
...
+  private async getJustOneById(id: string) {
+    const application = await this.applicationService.findOneById(id)
+    if (application) {
+      this.processingApplications.push({
+        pruned: false,
+        application,
+        failedAttachments: {},
+      })
+    }
+  }
```
eirikurn, line 52: Not used?

apps/application-system/api/src/app/modules/application/application.service.ts
```diff
@@ -84,6 +83,31 @@ export class ApplicationService {
...
+          },
+          answers: {
+            [Op.not]: '{}',
+          },
+          attachments: {
+            [Op.not]: '{}',
+          },
+          externalData: {
+            [Op.not]: '{}',
+          },
```
eirikurn, line 104: This looks scary to me. Will it always compare these as JSON strings?

apps/application-system/api/src/app/modules/application/application.service.ts
```diff
@@ -84,6 +83,31 @@ export class ApplicationService {
     })
   }
 
+  async findAllDueToBePruned(): Promise<Application[]> {
+    return this.applicationModel.findAll({
```
eirikurn, line 87: This query might be quite heavy (with more applications) and large (in the first run).

Heavy:
* We should tweak the columns/indices for this so we're not performing a full table scan over all applications ever submitted.
* Maybe have an index on pruneAt, but then it still always processes all pruned applications from the start.
* We could either null pruneAt after pruning successfully (not ideal?) or add a new column indicating that the application is pruned and have a multicolumn index on `(pruned, pruneAt)`, even a partial index `WHERE (pruned = FALSE AND pruneAt IS NOT NULL)`.
* If we add a `pruned` column, then we should not filter on the other fields.

Large:
* Only select the id and attachments fields?
* Should remove the order statement, unless needed for some reason?

apps/application-system/api/src/app/modules/application/lifecycle/application-lifecycle.service.ts
```diff
@@ -0,0 +1,165 @@
...
+      const applicationAttachments = prune.application.attachments as {
+        key: string
+        name: string
+      }
+
+      const attachments = this.attachmentsToKeyArray(applicationAttachments)
+      try {
+        if (attachments) {
+          await this.awsService.deleteObjects(this.getBucketName(), attachments)
+        }
```
eirikurn, line 77: Maybe simplify this logic to use `deleteObject`, handle any error with `failedAttachments` and skip the extra `fileExists` check.

Then we can get the bucket name from each attachment's url using AmazonS3URI since some attachments may be stuck in the upload bucket which might fail the other command. Though I'm not sure if this service account has delete permissions for the upload bucket ...

Note that the upload bucket has its own lifecycle rules so the object might be already gone. Would be great to verify what deleteObject does for objects that are already deleted or don't exist for some other reason.

workspace.json
```diff
@@ -1696,6 +1696,13 @@
           },
           "outputs": ["coverage/apps/application-system/api"]
         },
+        "worker": {
+          "builder": "@nrwl/node:execute",
+          "options": {
+            "buildTarget": "application-system-api:build",
+            "args": ["--job", "worker"]
+          }
+        },
```
eirikurn, line 1705: Nice

eirikurn: Guess we need to configure this cron job in the infra dsl somehow?

apps/application-system/api/src/app/modules/application/lifecycle/application-lifecycle.service.ts
```diff
@@ -0,0 +1,165 @@
...
+
+@Injectable()
+export class ApplicationLifeCycleService {
+  private processingApplications: ApplicationPruning[] = []
+
+  constructor(
+    @Inject(APPLICATION_CONFIG)
+    private readonly config: ApplicationConfig,
+    @Inject(LOGGER_PROVIDER)
+    private logger: Logger,
```
eirikurn, line 25: We might want to create a child logger with its own context to make these easier to query. Something like:

```
    this.logger = logger.child({ context: 'ApplicationLifeCycleService' })
```

eirikurn:
ps, loving this PR and the test coverage! great job!

apps/application-system/api/src/app/modules/application/application.service.ts
obmagnusson: Yes the pruned field was my thought also. When pruneAt = null it was my understanding that this meant that the applictation should not be pruned at all.

apps/application-system/api/src/app/modules/application/lifecycle/application-lifecycle.service.ts
obmagnusson: Well the  the `deleteObjects` and `deleteObject` just returns success if the file does not exist. So it wont tell us if we actually removed it or it never was there to begin with.
https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/S3.html#deleteObject-property

obmagnusson: Switched over to use `deleteObject` and  `AmazonS3URI` to determine the bucket name. But I kept the `fileExists` check because `deleteObject` does not appear to be giving us a good enough feedback on wether the file was delted or not. I still however worry that this will amount to alot of calls to the aws sdk for a single run.

apps/application-system/api/migrations/20220112162352-pruneAt-and-prune-index.js
```diff
@@ -0,0 +1,19 @@
+'use strict'
+
+module.exports = {
+  up: (queryInterface) => {
+    return queryInterface.sequelize.query(`
+      BEGIN;
+
+        CREATE INDEX pruned_and_pruneAt_idx ON application (prune_at) WHERE (pruned = FALSE AND prune_at IS NOT NULL);
```
eirikurn, line 8: I expected that we'd need to index both properties, `(pruned, prune_at)` for the query planner to pick it up, since we filter on both of these properties. 

But maybe it's smart enough to know that the index is already filtered in a similar way. Please verify.

apps/application-system/api/src/app/modules/application/lifecycle/application-lifecycle.service.ts
```diff
@@ -40,19 +43,12 @@ export class ApplicationLifeCycleService {
...
-      })
-    }
-  }
-
   private async fetchApplicationsToBePruned() {
-    const applications = await this.applicationService.findAllDueToBePruned()
+    const applications = (await this.applicationService.findAllDueToBePruned()) as Pick<
+      Application,
+      'id' | 'attachments' | 'answers' | 'externalData'
+    >[]
```
eirikurn, line 50: Why the type cast? You should only have `id` and `attachments` here, same as the function returns.

apps/application-system/api/src/app/modules/application/lifecycle/application-lifecycle.service.ts
```diff
@@ -142,24 +137,22 @@ export class ApplicationLifeCycleService {
   private attachmentsToKeyArray(
     attachments: object,
   ): {
-    Key: string
+    s3key: string
+    key: string
+    bucket: string
+    value: string
```
eirikurn, line 143: Maybe define an interface for this inside the file?

apps/application-system/api/src/main.ts
```diff
@@ -15,6 +15,8 @@ const worker = async () => {
   const app = await NestFactory.createApplicationContext(AppModule)
   app.enableShutdownHooks()
   await app.get(ApplicationLifeCycleService).run()
+  console.log('sdfsdf')
```
eirikurn, line 18: Remove?

apps/application-system/api/src/app/modules/application/files/aws.service.ts
```diff
@@ -58,4 +58,24 @@ export class AwsService {
         () => false,
       )
   }
+
+  public async deleteObjects(bucket: string, fileObjects: { Key: string }[]) {
```
eirikurn, line 62: Remove this?

apps/application-system/api/src/app/modules/application/lifecycle/application-lifecycle.service.ts
```diff
@@ -0,0 +1,158 @@
...
+
+      const attachments = this.attachmentsToKeyArray(applicationAttachments)
+
+      if (attachments) {
+        for (const attachment of attachments) {
+          const { key, s3key, bucket, value } = attachment
+          try {
+            await this.awsService.deleteObject(bucket, s3key)
+
+            const exists = await this.awsService.fileExists(bucket, value)
```
eirikurn, line 77: Hm. This feels redundant and expensive. Would be interested in knowing more about any false positive deletes you have experienced.

apps/application-system/api/src/main.ts
obmagnusson: whoops yes

apps/application-system/api/src/app/modules/application/lifecycle/application-lifecycle.service.ts
obmagnusson: Well later on update it will be populated with the resulting updated values . Checking these values in the tests.

pshomov (approved):
LGTM

apps/application-system/api/migrations/20220112162352-pruneAt-and-prune-index.js
```diff
@@ -5,7 +5,7 @@ module.exports = {
     return queryInterface.sequelize.query(`
       BEGIN;
 
-        CREATE INDEX pruned_and_pruneAt_idx ON application (prune_at) WHERE (pruned = FALSE AND prune_at IS NOT NULL);
+        CREATE INDEX pruned_and_pruneAt_idx ON application (prune_at, pruned) WHERE (pruned = FALSE AND prune_at IS NOT NULL);
```
eirikurn, line 8: Sorry, forgot to mention that `pruned` needs to come before `prune_at`. Basically any property you filter for a specific value (`pruned = FALSE`) should come before properties that you filter by a range (`prune_at < now()`)

apps/application-system/api/src/app/modules/application/lifecycle/application-lifecycle.service.ts
eirikurn: I see. Not great to lose type safety just for the tests. Wonder if there's a better way, but whatever :)

eirikurn (approved):
LGTM. Just fix this one minor thing, then we're good :)

(removed the automerge tag and approved the changes, you can add automerge back after fixing this one thing)

sigruntg (approved):

PR #6235, opened 2022-01-12
feat(island-ui): adding to buildFileUploadField, ui for image upload with preview

albinagu:
Image Upload with preview UI proposal vol. 1

EDIT:

https://user-images.githubusercontent.com/47886428/152549929-4a177157-06bc-4851-b252-4a8cbff89875.mov

## Checklist:


libs/application/templates/p-sign/src/fields/UploadedPhoto/index.tsx
```diff
@@ -0,0 +1,52 @@
...
+  }
+`
+
+const UploadedPhoto = ({ application }: UploadedPhotoProps) => {
+  const { formatMessage } = useLocale()
+
+  const attachmentKey = (application.answers.attachments as Array<{
+    key: string
+    name: string
+  }>)[0].key
```
kksteini, line 26: Is this guaranteed for ordering?

kksteini (approved):
LGTM

darrikonn:
Design comment: would remove the loading bar once completed ... or change the color to green in order to indicate that this has finished ... when first looking at this, it felt like it hadn't finished.

apps/application-system/api/src/app/modules/application/application.controller.ts
```diff
@@ -1007,4 +1008,45 @@ export class ApplicationController {
...
+    allowEmptyValue: false,
+  })
+  @ApiParam({
+    name: 's3Key',
+    type: String,
+    required: true,
+    description: 's3Key',
+    allowEmptyValue: false,
+  })
+  @ApiOkResponse({ type: PresignedUrlResponseDto })
```
darrikonn, line 1028: Checkout the @Documentation decorator from @island.is/nest/swagger.

libs/api/domains/application/src/lib/dto/getAttachmentPresignedUrl.input.ts
```diff
@@ -0,0 +1,12 @@
+import { Field, InputType } from '@nestjs/graphql'
+import { IsString, IsEnum } from 'class-validator'
+@InputType()
+export class GetAttachmentPresignedUrlInput {
+  @Field(() => String)
+  @IsString()
+  id!: string
+
+  @Field(() => String)
+  @IsString()
```
darrikonn, line 10: You can remove IsString ... doesn't do anything for the schema.
class-validator should be removed as well.

libs/api/domains/application/src/lib/application.resolver.ts
```diff
@@ -177,4 +178,12 @@ export class ApplicationResolver {
   ): Promise<PresignedUrlResponse> {
     return this.applicationService.presignedUrl(input, user)
   }
+
+  @Query(() => PresignedUrlResponse, { nullable: true })
+  getAttachmentPresignedURL(
```
darrikonn, line 183: Not following naming conventions for GraphQL ... remove `Get`.

libs/api/domains/application/src/lib/dto/getAttachmentPresignedUrl.input.ts
```diff
@@ -0,0 +1,12 @@
+import { Field, InputType } from '@nestjs/graphql'
+import { IsString, IsEnum } from 'class-validator'
+@InputType()
+export class GetAttachmentPresignedUrlInput {
```
darrikonn, line 4: Ditto on naming conventions.

libs/island-ui/core/src/lib/InputImageUpload/InputImageUpload.tsx
```diff
@@ -0,0 +1,340 @@
+import React, { useMemo } from 'react'
```
darrikonn, line 1: This component should probably not live inside core. It's making a GraphQL query which would break every application that is not coupled by the shared GraphQL API.
Should most likely be moved to `shared/components`.

darrikonn (changes requested):
Looks good... checkout my design comment as well (above)

apps/application-system/api/src/app/modules/application/application.controller.ts
rafnarnason: hi looks good from what i see here, https://github.com/island-is/island.is/blob/main/libs/nest/swagger/README.md, not superclear what i need to import to get it running ...

rafnarnason: amazingly figured it out
`import { Documentation } from '@island.is/nest/swagger'
`

apps/application-system/api/src/app/modules/application/application.controller.ts
```diff
@@ -1007,4 +1009,49 @@ export class ApplicationController {
...
+          type: 'string',
+          description: 's3 key for attachment',
+          required: true,
+        },
+      },
+    },
+  })
+  async getAttachmentPresignedURL(
+    @Param('id', new ParseUUIDPipe()) id: string,
+    @Param('s3Key') s3Key: string,
```
eirikurn, line 1036: Let's call this `attachmentKey` instead. S3 is an implementation detail.

apps/application-system/api/src/app/modules/application/application.controller.ts
```diff
@@ -1007,4 +1009,49 @@ export class ApplicationController {
...
+    )
+
+    if (!existingApplication.attachments) {
+      throw new NotFoundException('Attachments not found')
+    }
+
+    try {
+      const str = s3Key as keyof typeof existingApplication.attachments
+      const fileName = existingApplication.attachments[str]
+      const { bucket, key } = AmazonS3URI(fileName)
```
eirikurn, line 1051: Nitpick, but this line and maybe even the one above could be inside the `getAttachmentPresignedURL` helper.

libs/api/domains/application/src/lib/application.service.ts
```diff
@@ -196,4 +197,15 @@ export class ApplicationService {
       pdfType: type,
     })
   }
+
+  async attachmentPresignedURL(input: AttachmentPresignedUrlInput, auth: Auth) {
+    const { id, s3Key } = input
```
eirikurn, line 202: Also here.

libs/api/domains/application/src/lib/dto/AttachmentPresignedUrl.input.ts
```diff
@@ -0,0 +1,10 @@
+import { Field, InputType } from '@nestjs/graphql'
+
+@InputType()
+export class AttachmentPresignedUrlInput {
+  @Field(() => String)
+  id!: string
+
+  @Field(() => String)
+  s3Key!: string
```
eirikurn, line 9: Also here.

libs/island-ui/core/src/lib/InputImageUpload/InputImageUpload.tsx
eirikurn: I'd also argue that since it has application-system specific logic it should be in libs/application/ui-components or something.

Would be nice though (now or later) to split up the component logic so the same UI can be used with other "backends" later.

eirikurn (changes requested):
Nice stuff. Adding a couple minor issues on top of Darri's core review 👍

libs/island-ui/core/src/lib/InputImageUpload/InputImageUpload.tsx
darrikonn: Yeah, I agree on libs/application/ui-components ... just not core 😄 .

But I agree, the InputImageUpload could/should be more generic so it can be moved to `core`, and then have another wrapper component in libs/application/ui-components that uses InputImageUpload component from core alongside queries and application specific logic.

apps/application-system/api/src/app/modules/application/application.controller.ts
rafnarnason: yep - had been brewing in my mind - good call

apps/application-system/api/src/app/modules/application/application.controller.ts
rafnarnason: this one is a bit tricky, maybe this attachments object is not so great to work with ... anyways ... updates are served

libs/island-ui/core/src/lib/InputImageUpload/InputImageUpload.tsx
albinagu: since this is a part of the buildFileUploadField and is sort of an addon to it (used if the template field has 'forImageUpload' prop), im not sure i understand why it needs to be moved away from InputFileUpload

but i will do so if u say so! :D

albinagu: aaaah okok, yes, due to the query within this file?

darrikonn: Yeah. Core components shouldn't have any side effects.

eirikurn (approved):

darrikonn (approved):

stjanilofts (approved):

Rikkiabb (approved):

PR #6237, opened 2022-01-12
feat(nest/audit): Support token exchange client chain in audit logs

eirikurn:
## What

Include token exchange client chain in audit logs.

## Why

So we can see in audit logs which OAuth2 clients were involved with each change.

Let's say a user is authenticated into client A, which calls an API X which performs token exchange with client B which calls another API Y. In the audit log of API Y, we would only see that some change was performed by client A, which is only half of the truth.

With this change, we'll see that client A is the original authenticating client, but the actual change was performed by client B.

## Checklist:


saevarma (approved):
LGTM 👍

PR #6238, opened 2022-01-12
fix(application-templates-driving-license): fixing applicationFor check when filled

arnorhs:
## What

Fixing photo step for application full, when you can pick a license (currently behind feature flag)

## Why

To not break the application for people when we have the temporary license selection

## Checklist:


libs/application/templates/driving-license/src/lib/utils/formUtils.spec.ts
```diff
@@ -0,0 +1,116 @@
...
+  allowFakeCondition,
+  hasCompletedPrerequisitesStep,
+} from './formUtils'
+
+describe('isVisible', () => {
+  it('returns true when all functions return true', () => {
+    expect(
+      isVisible(
+        () => true,
+        () => true,
```
JuliusGG, line 16: is () => true, supposed to come twice?

JuliusGG (approved):
lgtm, 1 question so I removed automerge.

libs/application/templates/driving-license/src/lib/utils/formUtils.spec.ts
arnorhs: yeah, this isVisible function should return true when all functions return true

PR #6240, opened 2022-01-13
feat(service-portal): National registry children

thordurhhh:
## What

* Add service for children of logged in user
* Add detailed children info on minarsidur

.. There will be more info added with updated design in the next release.

## Why

Not enough data is present in the family service, so the children service was needed. Using the user service for the spouse info.

## Checklist:


libs/service-portal/family/src/screens/Spouse/Spouse.tsx
```diff
@@ -0,0 +1,106 @@
...
+        />
+        <Divider />
+        <UserInfoLine
+          label={defineMessage(m.natreg)}
+          content={formatNationalId(nationalId)}
+          loading={loading}
+        />
+        <Divider />
+        <UserInfoLine
+          label="Tengsl"
```
disaerna, line 92: Þýðing?

libs/api/domains/national-registry/src/lib/nationalRegistry.service.ts
```diff
@@ -28,6 +29,9 @@ export class NationalRegistryService {
       fullName: user.Fulltnafn,
       gender: this.formatGender(user.Kyn),
       maritalStatus: this.formatMaritalStatus(user.hju),
+      spouseName: user.nafnmaka,
+      spouseNationalId: user.MakiKt,
+      spouseCohab: user.Sambudarmaki,
```
osk, line 34: Namespace in string, should be an object, see comment on type

libs/api/domains/national-registry/src/lib/types/user.type.ts
```diff
@@ -31,4 +31,7 @@ export interface User {
     city: string
     date: string
   }
+  spouseName?: string
+  spouseNationalId?: string
+  spouseCohab?: string
```
osk, line 36: Lets have a spouse object that's optional, then we wont have to check each property if no spouse
```
spouse?: {
  name?: string,
  nationalId?: string,
  cohab?: boolen, // boolean not string, right?
}
```

libs/clients/national-registry/v1/src/lib/dto/getViewISLBorninMin.dto.ts
```diff
@@ -0,0 +1,119 @@
...
+
+interface Diffgram {
+  DocumentElement: DocumentElement
+}
+
+interface DocumentElement {
+  ISLBorninMin: ISLBorninMin[]
+}
+
+export interface ISLBorninMin {
```
osk, line 20: aren't most of these nullable?

libs/clients/national-registry/v1/src/lib/nationalRegistryApi.ts
```diff
@@ -87,6 +89,25 @@ export class NationalRegistryApi {
...
+    )
+
+    if (!borninMinResponse) {
+      throw new NotFoundException(
+        `children for nationalId ${nationalId} not found`,
+      )
+    }
+
+    const documentData =
+      borninMinResponse.table.diffgram.DocumentElement.ISLBorninMin
```
osk, line 107: I'd be more defensive here since it's a 3rd party service that might just randomly one day change this/break this and then we might get a null reference. i.e. add as potentially null/undefined in type and check each is available

libs/service-portal/family/src/components/FamilyMemberCard/FamilyMemberCard.tsx
```diff
@@ -8,22 +8,45 @@ import {
...
 import React, { FC } from 'react'
 import { Link } from 'react-router-dom'
+import { defineMessage } from 'react-intl'
 import * as styles from './FamilyMemberCard.css'
 
 interface Props {
   title: string
   nationalId?: string
-  familyRelation?: string
+  familyRelation?: 'child' | 'spouse'
```
osk, line 17: 👍

libs/service-portal/family/src/screens/FamilyMember/FamilyMember.tsx
```diff
@@ -89,11 +97,48 @@ const FamilyMember: ServicePortalModuleComponent = () => {
...
-                  natRegGenderMessageDescriptorRecord[person.gender],
-                )
-              : ''
+              : person?.genderDisplay || ''
+          }
+          loading={loading}
+        />
+        <Divider />
+        <UserInfoLine
+          label="Fæðingarstaður"
```
osk, line 106: translate

libs/service-portal/family/src/screens/FamilyMember/FamilyMember.tsx
```diff
@@ -89,11 +97,48 @@ const FamilyMember: ServicePortalModuleComponent = () => {
...
+          content={
+            error
+              ? formatMessage(dataNotFoundMessage)
+              : person?.birthplace || ''
+          }
+          loading={loading}
+        />
+        <Divider />
+        <UserInfoLine
+          label="Foreldrar"
```
osk, line 116: translate

osk (changes requested):

libs/api/domains/national-registry/src/lib/nationalRegistry.service.ts
thordurhhh: Yes, makes sense!

libs/api/domains/national-registry/src/lib/types/user.type.ts
thordurhhh: Cohab is actually a string, it can return all sorts of different results 😃

libs/clients/national-registry/v1/src/lib/dto/getViewISLBorninMin.dto.ts
thordurhhh: whoops, only changed that in one place. good catch.

osk (approved):

PR #6244, opened 2022-01-13
feat(web): Syslumenn auctions, show respondents and petitioners for ships and real estate

RunarVestmann:
# Syslumenn auctions, show respondents and petitioners for both ships and real estate

## What

* Respondents and petitioners were only showing up for real estate auctions
* Show where the auction takes place if that data is available

## Why

* We want to show the respondents and petitioners for both ships and real estates
* Recently, a new field `uppbodStadur` was added to Syslumenn API, to contain information about where the auction takes place. This field is currently always empty, but soon it will be fully implemented in the Syslumenn API. Therefore, in this PR, we are rendering this field in the view if it has a value.

### Before

ship-auction-before

### After

ship-auction-after

## Checklist:


arnorhs (approved):
Only reviewed code-owned files. LGTM

siggifv (approved):

PR #6260, opened 2022-01-14
chore(gitbook): Cleanup navigation

eirikurn:
## What

* Allow app/lib README to have a separate document title from the navigation title.
* Rename nested navigation titles to not contain name of parent folder.
* Ignore api-domain, application-template and service-portal submodule libs.

## Why

To make out gitbook navigation nicer. Instead of a navigation like this:

```
Nest
- Nest Config Module
- Nest Feature Flag Module
```

It's something like this:

```
Nest
- Config Module
- Feature Flag Module
```

The README title should still have the full name to provide good context to users that deep link to the documentation.

Also, our gitbook should focus on providing high-level documentation for our apps and shared libraries. Most "App" submodules don't have a README, and there are so many of them that they clutter up the Gitbook navigation. So they are not available anymore in gitbook, but devs can still explore them in the codebase, on demand.

Currently these modules are manually "ignored". In a future PR it would be nice to automatically ignore specific subtrees, but  we might still want to include some of these, like service portal core? We could do this with `<!-- gitbook-include -->`, or just move all the service-portal libs we don't need into a subfolder which we can ignore, like "libs/service-portal/modules".

## Checklist:


eirikurn:
Reviewer hint: Focus on `SUMMARY.md` and `gitbook.js`.

darrikonn (approved):
LGTM

Rikkiabb (approved):
🥇

disaerna (approved):
👍

PR #6264, opened 2022-01-14
infra(services-auth-api): adding extra DD tracing info and use read replicas in services-auth-api

dabbeg:
# ...

This is something I picked up from @pshomov and here is the old PR: https://github.com/island-is/island.is/pull/5332

Closes: island-is/infrastructure#960

## What

* Start using read replicas
* Setup tracing capabilities for datadog

## Why

* For better db performance
* For better monitoring insights

## Checklist:


apps/services/auth-api/esbuild.json
```diff
@@ -26,7 +26,21 @@
...
+    "@protobufjs/aspromise",
+    "@protobufjs/base64",
+    "@protobufjs/codegen",
+    "@protobufjs/eventemitter",
+    "@protobufjs/fetch",
+    "@protobufjs/float",
+    "@protobufjs/inquire",
+    "@protobufjs/path",
+    "@protobufjs/pool",
+    "@protobufjs/utf8"
```
eirikurn, line 43: This is annoying :(

libs/nest/sequelize/README.md
```diff
@@ -0,0 +1,7 @@
+# Sequelize
```
eirikurn, line 1: Add some docs, tests maybe?

eirikurn (commented):
Is the read replica configuration only used in auth-api here? Do you plan to do a load test on a feature branch before integrating it elsewhere?

I’m concerned about configuring read replicas (in auth-api or globally) without reviewing all endpoints to see if any of them have read-after-write logic which might encounter race-condition issues with read replicas.

Regarding the tracing setup. I’m not sure about the value we get from tracing “everything”, versus potential overhead or loss of data. We should focus on tracing everything that happens asynchronously (dbs, requests). Datadog does this pretty well by default — except Redis for some reason, which would be reaaally good to fix.

Generally in NodeJS, with or without tracing, it’s difficult to see if the stuff we’re tracing is actually slow, or if CPU/IO is bottlenecked handling other stuff. Haven’t found good solutions for this unfortunately.

dabbeg:
> Is the read replica configuration only used in auth-api here? Do you plan to do a load test on a feature branch before integrating it elsewhere?
> 
Yes only configured for the auth-api and I put the commonly used sequelize options in a library so we can easily add this to more services later. No I had not planned on loadtesting this on a feature branch.

> I’m concerned about configuring read replicas (in auth-api or globally) without reviewing all endpoints to see if any of them have read-after-write logic which might encounter race-condition issues with read replicas.
> 
That's a good point. If there are any read operations that need read-after-write consistency we would need to explicitly set the option to use the main rds instance instead of the read replica. Something like:
```
const user = await User.findOne(
  { where: { nationalId } },
  { useMaster: true }
);
```
Can you take a look at this @valurefugl?

> Regarding the tracing setup. I’m not sure about the value we get from tracing “everything”, versus potential overhead or loss of data. We should focus on tracing everything that happens asynchronously (dbs, requests). Datadog does this pretty well by default — except Redis for some reason, which would be reaaally good to fix.
> 
> Generally in NodeJS, with or without tracing, it’s difficult to see if the stuff we’re tracing is actually slow, or if CPU/IO is bottlenecked handling other stuff on the event loop. Haven’t found good solutions for this unfortunately.

@pshomov can you comment on the tracing setup.

libs/nest/sequelize/README.md
dabbeg: Yeah good idea

apps/services/auth-api/esbuild.json
dabbeg: Yeah, this was necessary for the tracing. Am I right @pshomov?

valurefugl:
Think we would like to ensure consistency for read operations in `GrantsController `and `UserIdentitiesController`.

dabbeg:
@valurefugl @eirikurn Can I get a review again?

I removed the tracing and I added `useMaster` to the services @valurefugl pointed out.

This is only configured on dev so we can try it out for a bit there.

infra/src/environments.ts
```diff
@@ -3,6 +3,8 @@ import { EnvironmentConfigs } from './dsl/types/charts'
 export const Envs: EnvironmentConfigs = {
   dev: {
     auroraHost: 'postgres-applications.internal',
+    auroraReplica:
+      'dev-vidspyrna-aurora.cluster-ro-c6cxecmrvlpq.eu-west-1.rds.amazonaws.com',
```
eirikurn, line 7: Is this safe to configure in all the monorepo charts? Should judicial-system use the same aurora clusters as islandis?

eirikurn (approved):

infra/src/environments.ts
dabbeg: Yeah it should be safe, we have one aurora cluster for all of our applications (judicial-system included) and the plan is to keep on doing that for additional applications.

charts/islandis/values.dev.yaml
```diff
@@ -58,6 +58,7 @@ air-discount-scheme-backend:
   env:
     DB_HOST: 'postgres-applications.internal'
     DB_NAME: 'air_discount_scheme_backend'
+    DB_REPLICAS_HOST: 'dev-vidspyrna-aurora.cluster-ro-c6cxecmrvlpq.eu-west-1.rds.amazonaws.com'
```
pshomov, line 61: @dabbeg please create follow up tasks to get a pretty DNS name for the replica

pshomov (approved):
LGTM

PR #6269, opened 2022-01-14
ci: Cache action success

dabbeg:
## What

The cache retry is not working, because the cache-hit output is not behaving like I expected.
So I added a success output to the caching action here: https://github.com/island-is/cache/pull/1

This PR starts to use that success output mentioned above.

## Checklist:


pshomov (approved):
LGTM

PR #6271, opened 2022-01-14
feat(as): Connect to Veita

Rikkiabb:
# ...
https://app.asana.com/0/1201600280110882/1201646079999939

## What
Connects application system to Veita
Adds data provider that talks to Veita and fetches if user has application
Adds domain that exposes api points in Veita to application system
We don't throw an error if we don't find application because we don't need it if we cut out rest data source, which we do for this look up
Adds auto generation to Veita api

TODO: Add env params to helm
TODO: Read other data from Veita

## Why
Needed data in our flow

## Screenshots / Gifs

## Checklist:


libs/api/domains/financial-aid/src/lib/financialAid.service.ts
```diff
@@ -0,0 +1,22 @@
...
+@Injectable()
+export class FinancialAidService {
+  constructor(private applicationApi: ApplicationApi) {}
+
+  applicationApiWithAuth(auth: Auth) {
+    return this.applicationApi.withMiddleware(new AuthMiddleware(auth))
+  }
+
+  async hasUserApplicationForCurrentPeriod(auth: Auth, nationalId: string) {
+    console.log('auth', auth)
```
eirikurn, line 15: Remove?

libs/api/domains/financial-aid/src/lib/financialAid.module.ts
```diff
@@ -0,0 +1,36 @@
...
+        FinancialAidResolver,
+        FinancialAidService,
+        {
+          provide: ApplicationApi,
+          useFactory: async () =>
+            new ApplicationApi(
+              new Configuration({
+                fetchApi: createEnhancedFetch({
+                  name: 'FinancialAidModule.applicationApi',
+                  timeout: 60000,
```
eirikurn, line 26: Why the long timeout? Is the API slow? :)

libs/api/domains/financial-aid/src/lib/dto/currentApplication.ts
```diff
@@ -0,0 +1,7 @@
+import { Field, ObjectType } from '@nestjs/graphql'
+
+@ObjectType()
+export class CurrentApplicationResponse {
```
eirikurn, line 4: This ObjectType should be namespaced. 

Eg `@ObjectType(‘LocalGovFinancialSupportApplicationStatus‘)`

libs/api/domains/financial-aid/src/lib/financialAid.module.ts
```diff
@@ -0,0 +1,36 @@
...
+import { FinancialAidService } from './financialAid.service'
+import { ApplicationApi, Configuration } from '../../gen/fetch'
+import { createEnhancedFetch } from '@island.is/clients/middlewares'
+
+export interface Config {
+  baseApiUrl: string
+}
+
+@Module({})
+export class FinancialAidModule {
```
eirikurn, line 12: I know this is consistent with your “app” name, but I’m concerned this might conflict with some “government financial aid” programs.

Should we create a more focused domain like Veita, Samband or “LocalGov” which includes FinancialAidApplication-related resolvers and payloads?E

libs/api/domains/financial-aid/src/lib/financialAid.module.ts
```diff
@@ -0,0 +1,36 @@
...
+  static register(config: Config): DynamicModule {
```
eirikurn, line 13: Should use the new “@island.is/nest/config” system instead of the root environments.ts and `register` pattern.

libs/api/domains/financial-aid/src/lib/financialAid.module.ts
```diff
@@ -0,0 +1,36 @@
...
+    return {
+      module: FinancialAidModule,
+      providers: [
+        FinancialAidResolver,
+        FinancialAidService,
+        {
+          provide: ApplicationApi,
```
eirikurn, line 20: Should move this to a stand-alone client library so you can share it with your `submitApplication` template API.

libs/api/domains/financial-aid/src/lib/financialAid.service.ts
```diff
@@ -0,0 +1,22 @@
...
+    return await this.applicationApiWithAuth(
+      auth,
+    ).applicationControllerGetCurrentApplication({
```
eirikurn, line 18: What does this API return if there is no valid application? null? If it returns 404, you might want to catch it and return null instead.

libs/api/domains/financial-aid/src/lib/financialAid.resolver.ts
```diff
@@ -0,0 +1,28 @@
...
+import { FinancialAidService } from './financialAid.service'
+import { CurrentApplicationResponse } from './dto/currentApplication'
+
+@UseGuards(IdsUserGuard, ScopesGuard)
+@Resolver()
+export class FinancialAidResolver {
+  constructor(private financialAidService: FinancialAidService) {}
+
+  @Query(() => CurrentApplicationResponse, { nullable: true })
+  async hasUserFinancialAidApplicationForCurrentPeriod(
```
eirikurn, line 19: Should be something like `{namespace}FinancialAidApplicationStatus` to better fit our naming conventions?

Or `{namespace}FinancialAidApplications` with optional filters like “period=CURRENT” or whatever.

workspace.json
```diff
@@ -5014,8 +5041,7 @@
             "main": "apps/icelandic-names-registry/backend/src/main.ts",
             "tsConfig": "apps/icelandic-names-registry/backend/tsconfig.app.json",
             "webpackConfig": "apps/icelandic-names-registry/backend/webpack.config.js",
-            "assets": [
-              {
+            "assets": [{
```
eirikurn, line 5044: Where do all these changes come from. Is this formatted by prettier?

libs/api/domains/financial-aid/src/lib/financialAid.module.ts
Rikkiabb: No not at all, will remove, this is just dev c/p and forgot to remove 🙈

MargretFinnboga (approved):

libs/api/domains/financial-aid/src/lib/financialAid.module.ts
Rikkiabb: Very good point, we have been using the 'Municipalities'. So going to change to `MunicipalitiesFinancialAid`, you'll let me know if that is also too vague.

libs/api/domains/financial-aid/src/lib/financialAid.module.ts
Rikkiabb: Great point, then I can also use that client in our api and remove rest-datasource as the way to talk to our backend from our api, will do 🙌

workspace.json
Rikkiabb: My local prettier seems to be configured incorrectly for .json files

apps/financial-aid/backend/src/environments/environment.ts
```diff
@@ -53,7 +53,7 @@ const devConfig = {
   },
   identityServerAuth: {
     issuer: 'https://identity-server.dev01.devland.is',
-    audience: '@samband.is',
+    audience: '@island.is', // TODO: Figure out how we support both island.is and samband audience
```
Rikkiabb, line 56: @eirikurn Is there someone using IDS currently which allows 2 audiences?

In our case we would like to allow '@island.is' and '@samband.is' and then we would do extra ACL in our backend.

eirikurn: Why do you want to have a `@island.is` audience? We should try and keep your backend audience/scopes completely separate from @island.is.

libs/api/domains/financial-aid/src/lib/financialAid.service.ts
eirikurn: Ah. Good question. The API Design Guide says it should return 404 if there’s no data to be returned.

That also fits better when documenting the API with OpenAPI, to always have a consistent DTO or a different response status.

libs/api/domains/financial-aid/src/lib/financialAid.module.ts
eirikurn: Ooh, that would be nice!

libs/api/domains/financial-aid/src/lib/financialAid.module.ts
eirikurn: Sounds great! If anything, the domain could be “Municipalities” and contain multiple resolvers for different things. But this is fine for now.

libs/api/domains/municipalities-financial-aid/src/lib/applicationApiProvider.ts
```diff
@@ -0,0 +1,22 @@
...
+
+import { ConfigType } from '@island.is/nest/config'
+
+import {
+  ApplicationApi,
+  Configuration,
+} from '@island.is/clients/municipalities-financial-aid'
+import { MunicipalitiesFinancialAidConfig } from './municipalitiesFInancialAid.config'
+
+export const ApplicationApiProvider: Provider<ApplicationApi> = {
```
eirikurn, line 11: This should be exported from the client library. The config should be there as well.

The domain (and other modules) can then import the client library module, and get ApplicationAPI injected wherever.

If you want, the client library could also have its own service class if you want to share some “integration” logic between your projects. E.g. 404 handling. Just be mindful that it doesn’t get too coupled to one of the apps.

libs/api/domains/financial-aid/src/lib/financialAid.service.ts
Rikkiabb: Nice, will throw 404 😄

apps/financial-aid/backend/src/environments/environment.ts
Rikkiabb: Because when this goes into the application system, we will use the same login as the application system does for the applicant flow, which means the audience in the user jwt will be @island.is. 

Is there another way of doing this that I am missing?

eirikurn: Yeah, I think you might have this backwards. APIs are "Resource Servers" in OAuth2 lingo. We define these resources in IDS, for example named "@island.is" and "@samband.is" and associate scopes with them.

When the Application System client gets access to some scopes (eg "@island.is/applications:read" and "@samband.is/financial-aid:read"), the IDS issues a JWT with audience = all the associated resource servers.

It means the audience for this token are resource servers which handle the list of scopes. Then when I send it to different APIs, they verify that the token was "meant" for that API by checking if the audience matches.

tldr: tokens can have multiple audiences. If a token has a samband scope, it will have a samband audience (resource) in the token as well. Your API should only check that.

Rikkiabb: Makes sense, thanks for the write up :)

PR #6273, opened 2022-01-16
feat(parental-leave): Be able to fully use rights

baering:
## What

For certain periods greater than remaining rights it is not always possible to fully use all days since there can be a remainder of days from spreading rights over some amount of months with a certain percentage.

By specifying that the user wishes to fully use days the VMST API will make sure to use all the remaining rights.

This is done by prefixing the ratio with `D` so the API will process it as days to use rather than a ratio to use for each month within the period.

## Why

So that there won't be some days left when the user means to use all of them for a single period.

## Screenshots / Gifs

image

## Checklist:

### DIFF
--- a/libs/api/domains/directorate-of-labour/src/dto/getParentalLeavesEstimatedPaymentPlan.input.ts
+++ b/libs/api/domains/directorate-of-labour/src/dto/getParentalLeavesEstimatedPaymentPlan.input.ts
@@ -1,4 +1,4 @@
-import { Field, Float, InputType } from '@nestjs/graphql'
+import { Field, InputType } from '@nestjs/graphql'
 import { IsString, IsNumber, IsBoolean, IsArray } from 'class-validator'
 
 @InputType()
@@ -11,9 +11,9 @@ class Period {
   @IsString()
   to!: string
 
-  @Field(() => Float)
+  @Field(() => String)
   @IsNumber()
-  ratio!: number // Ratio of usage in period.
+  ratio!: string // Ratio of usage in period.
 
   @Field(() => Boolean)
   @IsBoolean()
--- a/libs/api/domains/directorate-of-labour/src/lib/directorate-of-labour.repository.ts
+++ b/libs/api/domains/directorate-of-labour/src/lib/directorate-of-labour.repository.ts
@@ -206,7 +206,7 @@ export class DirectorateOfLabourRepository {
         period: {
           from: '01-01-2020',
           to: '01-01-2020',
-          ratio: 0.8,
+          ratio: '0.8',
           approved: true,
           paid: true,
         },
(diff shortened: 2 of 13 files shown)


libs/application/templates/parental-leave/src/fields/PeriodPercentage/PeriodPercentage.tsx
```diff
@@ -94,13 +98,45 @@ export const PeriodPercentage: FC<PeriodPercentageField> = ({
...
+      setMaxPercentageValue(max)
+
+      options.splice(0, 0, {
+        value: max,
+        label: formatMessage(
+          parentalLeaveFormMessages.duration.fullyUsedRatio,
+          { maxPercentage: max },
+        ),
+      })
+    }
```
eirikurn, line 122: Minor: it's a bit weird to use `useEffect` to initialize component state from available data (this whole function?). Since it performs a redundant render, possibly with a flash of wrong data.

In the future, you might want to have this kind of logic in the component or a utility function. If you want to make sure to only run this once (instead of during each render), `useMemo` can help.

libs/api/domains/directorate-of-labour/src/lib/directorate-of-labour.repository.ts
```diff
@@ -206,7 +206,7 @@ export class DirectorateOfLabourRepository {
         period: {
           from: '01-01-2020',
           to: '01-01-2020',
-          ratio: 0.8,
+          ratio: '0.8',
```
eirikurn, line 209: Should this be 80?

libs/api/domains/directorate-of-labour/src/dto/getParentalLeavesEstimatedPaymentPlan.input.ts
```diff
@@ -11,9 +11,9 @@ class Period {
   @IsString()
   to!: string
 
-  @Field(() => Float)
+  @Field(() => String)
   @IsNumber()
-  ratio!: number // Ratio of usage in period.
+  ratio!: string // Ratio of usage in period.
```
eirikurn, line 16: Would be good to document what kind of values this field supports. The name "ratio" is not great. Not sure if we want to "clean" that up in our schema.

libs/application/templates/parental-leave/src/fields/PeriodPercentage/PeriodPercentage.tsx
```diff
@@ -133,6 +171,15 @@ export const PeriodPercentage: FC<PeriodPercentageField> = ({
...
       )}
+
+      {isUsingAllRemainingDays && (
+        <input
+          type="hidden"
+          ref={register}
+          name={`periods[${currentIndex}].daysToUse`}
+          value={remainingRights}
+        />
+      )}
```
eirikurn, line 182: Looks like it doesn't clear an existing value. Would be nice if this step could support edits/back even though it doesn't happen now.

eirikurn (approved):
LGTM. Some minor comments but your call, happy to merge as is.

PR #6274, opened 2022-01-16
feat(auth-react): Checking session info from IDS

saevarma:
## What

Adding usage of the new `connect/sessioninfo` endpoint in IDS in new component `CheckIdpSession` in the `auth-react` library. 

## Why
So service-portal and application-system can monitor when the user session expires to redirect to sign in page.

## Checklist:


libs/auth/react/src/lib/Authenticator/CheckIdpSession.tsx
```diff
@@ -0,0 +1,117 @@
...
+      return userManager.signinRedirect({
+        state: getReturnUrl(history, authSettings),
+      })
+      // Nothing more happens here since browser will redirect to IDS.
+    },
+    [dispatch, userManager, authSettings, history],
+  )
+
+  const onLoadHandler = (event: SyntheticEvent<HTMLIFrameElement>) => {
+    console.log('***** iFrame loaded', event)
```
darrikonn, line 56: debug.

libs/auth/react/src/lib/Authenticator/CheckIdpSession.tsx
```diff
@@ -0,0 +1,117 @@
...
+          new Date().getTime() +
+          1000
+
+        if (timeout <= 0) {
+          // The session is expired but for some reason the `isExpired` was not correctly set
+          return signIn()
+        }
+
+        if (!sessionTimeout) {
+          console.log(`**** Set timeout to check again in ${timeout} ms`)
```
darrikonn, line 96: ditto

libs/auth/react/src/lib/Authenticator/CheckIdpSession.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+          sessionTimeout = setTimeout(() => {
+            console.log('Checking again if session is still valid')
```
darrikonn, line 99: ditto

darrikonn (changes requested):
Do we want these console logs?


Shouldn't the GraphQL error link be handling these? Which makes the API take care of returning Unauthorized, which the errorLink can pick up and make the appropriate action, e.g. refresh the access token, and if successful retry the batched calls without the user noticing anything.

libs/auth/react/src/lib/Authenticator/CheckIdpSession.tsx
saevarma: Thought I had cleaned up all my console logs ...  apparently not 😅

saevarma:
> 
> mmm I see, smart! The main thing I'm worried about is this clashing with the GraphQL error link, since it should handle similar logic. It'll be hard to debug what's happening if that happens.
> 
> Scenario:
> 
> 1. The session expires
> 2. The errorLink tries to refresh the access token
> 3. In the meantime this session handler redirects the user to the signin page
> 4. The errorLink can't do what it was supposed to do.

Currently the application system and service portal are not requesting the offline_access scope, so they are not using refresh token. They depend on the silent login by iframe using the IDS authentication cookie and if the cookie is expired the user has to login again, same as this handler does.

But this is a valid point for clients using the offline_access scope. So I added check if the offline_access scope is in the scope array in the AuthSettings this session handler isn't loaded.

https://github.com/island-is/island.is/blob/0c58ebf2945e56b7089e2ea74e56f06ae941629c/libs/auth/react/src/lib/Authenticator/CheckAuth.tsx#L27

darrikonn (approved):

PR #6295, opened 2022-01-18
fix(web): Stricter lang toggling and news added to lang toggle

kksteini:
A user on an icelandic news article presses the `EN` button to switch to english.
User is thrown to -> `/en/news` even though the english version of the news article exists.

The language toggler was missing the contentful id for path lookup on switching.

Added a stricter check for toggles.
Some content models auto generate slugs from the title and that auto generation is triggered for both locales even if only the icelandic title is given.

21 out of 23 content models who have slugs also have a title so adding this check adds some robustness. Work is underway to standardize the other two content models to have a 'title' field instead of their respective custom ones, 'pageTitle' and 'question'.

The two content models in question, `Open Data Page` and `SupportQNA` are not relying on this functionality at the moment so introducing this check at this stage is OK.

## Checklist:


JuliusGG (approved):
lgtm

PR #6301, opened 2022-01-18
feat(air-discount-scheme): xroad user relations.

JuliusGG:
# ...

## What

## Why

## Screenshots / Gifs

## Checklist:

### DIFF
--- a/apps/air-discount-scheme/backend/src/app/modules/user/user.controller.ts
+++ b/apps/air-discount-scheme/backend/src/app/modules/user/user.controller.ts
@@ -19,6 +19,9 @@ import { AirlineUser, User } from './user.model'
 import { DiscountService } from '../discount'
 import { FlightService } from '../flight'
 import { AuthGuard } from '../common'
+import { CurrentUser, IdsUserGuard, Scopes, ScopesGuard } from '@island.is/auth-nest-tools'
+import { NationalRegistryXRoadService } from '@island.is/api/domains/national-registry-x-road'
+import { User as AuthUser } from '@island.is/auth-nest-tools'
 
 @ApiTags('Users')
 @Controller('api/public')
@@ -52,18 +55,32 @@ export class PublicUserController {
   }
 }
 
+@UseGuards(IdsUserGuard, ScopesGuard)
+@Scopes('@vegagerdin.is/air-discount-scheme-scope')
 @Controller('api/private')
 export class PrivateUserController {
   constructor(
     private readonly flightService: FlightService,
     private readonly userService: UserService,
+    private thjodskraXroad: NationalRegistryXRoadService,
   ) {}
 
   @Get('users/:nationalId/relations')
   @ApiExcludeEndpoint()
   async getUserRelations(
     @Param() params: GetUserRelationsParams,
+    @CurrentUser() user: AuthUser,
   ): Promise<User[]> {
+    let person
+    try {
+      person = await this.thjodskraXroad.getNationalRegistryPerson(user, user.nationalId)
+    } catch (e) {
+      console.log(e)
+    }
+    console.log(person)
+
+
+
     const relations = await this.userService.getRelations(params.nationalId)
     const users = await Promise.all([
       this.userService.getUserInfoByNationalId(params.nationalId),
--- a/apps/air-discount-scheme/backend/src/app/modules/user/user.controller.ts
+++ b/apps/air-discount-scheme/backend/src/app/modules/user/user.controller.ts
@@ -71,17 +71,29 @@ export class PrivateUserController {
     @Param() params: GetUserRelationsParams,
     @CurrentUser() user: AuthUser,
   ): Promise<User[]> {
-    let person
+    // let person
+    // try {
+    //   person = await this.thjodskraXroad.getNationalRegistryPerson(user, user.nationalId)
+    // } catch (e) {
+    //   console.log(e)
+    // }
+    // console.log(person)
+    let children
     try {
-      person = await this.thjodskraXroad.getNationalRegistryPerson(user, user.nationalId)
+      children = await this.thjodskraXroad.getChildrenForsja(user, user.nationalId)
     } catch (e) {
       console.log(e)
     }
-    console.log(person)
+    console.log(children)
 
+    const relations = await this.thjodskraXroad.getChildrenForsja(user, params.nationalId)
 
-
-    const relations = await this.userService.getRelations(params.nationalId)
+    if(relations === undefined) {
+      return [] as User[]
+    }
+    console.log(relations)
+    console.log(relations[0])
+    //const relations = await this.userService.getRelations(params.nationalId)
     const users = await Promise.all([
       this.userService.getUserInfoByNationalId(params.nationalId),
       ...relations.map((nationalId) =>
--- a/apps/air-discount-scheme/backend/src/app/modules/user/user.controller.ts
+++ b/apps/air-discount-scheme/backend/src/app/modules/user/user.controller.ts
@@ -19,7 +19,12 @@ import { AirlineUser, User } from './user.model'
 import { DiscountService } from '../discount'
 import { FlightService } from '../flight'
 import { AuthGuard } from '../common'
-import { CurrentUser, IdsUserGuard, Scopes, ScopesGuard } from '@island.is/auth-nest-tools'
+import {
+  CurrentUser,
+  IdsUserGuard,
+  Scopes,
+  ScopesGuard,
+} from '@island.is/auth-nest-tools'
 import { NationalRegistryXRoadService } from '@island.is/api/domains/national-registry-x-road'
 import { User as AuthUser } from '@island.is/auth-nest-tools'
 
@@ -71,8 +76,11 @@ export class PrivateUserController {
     @Param() params: GetUserRelationsParams,
     @CurrentUser() user: AuthUser,
   ): Promise<User[]> {
-    const relations = await this.thjodskraXroad.getChildrenForsja(user, params.nationalId)
-    if(relations === undefined) {
+    const relations = await this.thjodskraXroad.getChildrenForsja(
+      user,
+      params.nationalId,
+    )
+    if (relations === undefined) {
       return [] as User[]
     }
     const users = await Promise.all([
--- a/apps/air-discount-scheme/backend/src/app/modules/user/user.controller.ts
+++ b/apps/air-discount-scheme/backend/src/app/modules/user/user.controller.ts
@@ -71,29 +71,10 @@ export class PrivateUserController {
     @Param() params: GetUserRelationsParams,
     @CurrentUser() user: AuthUser,
   ): Promise<User[]> {
-    // let person
-    // try {
-    //   person = await this.thjodskraXroad.getNationalRegistryPerson(user, user.nationalId)
-    // } catch (e) {
-    //   console.log(e)
-    // }
-    // console.log(person)
-    let children
-    try {
-      children = await this.thjodskraXroad.getChildrenForsja(user, user.nationalId)
-    } catch (e) {
-      console.log(e)
-    }
-    console.log(children)
-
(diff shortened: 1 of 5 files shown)


libs/api/domains/national-registry-x-road/src/lib/nationalRegistryXRoad.service.ts
```diff
@@ -75,6 +75,20 @@ export class NationalRegistryXRoadService {
...
+  async getChildrenForsja(
+    user: User,
+    parentNationalId: string,
+  ): Promise<string[] | undefined> {
+    const nationalRegistryApi = this.nationalRegistryApiWithAuth(user)
+    const childrenNationalIds = await nationalRegistryApi
+      .einstaklingarGetForsja({
+        id: parentNationalId,
+      })
+      .catch(this.handle404)
```
Rikkiabb, line 88: I see we are doing the same on lines 96-101, we should extract that to a function and reuse that instead of having this duplicated. 

Also in that function (on line 103) we return an empty array if there are no `id's` found. I see you are doing that in your controller, but if you want then you can extract that also to the function and then you can remove it from your controller. 😄

apps/air-discount-scheme/backend/src/app/modules/user/user.controller.ts
```diff
@@ -52,19 +60,29 @@ export class PublicUserController {
...
   @Get('users/:nationalId/relations')
   @ApiExcludeEndpoint()
   async getUserRelations(
     @Param() params: GetUserRelationsParams,
+    @CurrentUser() user: AuthUser,
   ): Promise<User[]> {
-    const relations = await this.userService.getRelations(params.nationalId)
+    const relations = await this.thjodskraXroad.getChildrenForsja(
+      user,
+      params.nationalId,
```
Rikkiabb, line 81: Why don't you use the national id from the user instead of using params here?

Don't know the flow but it feels like more sanitized.

JuliusGG: I originally did this because I had access to params.nationalId, meaning I could verify that the nationalId from currentUser-authNestTools would have the same nationalid as was sent into the function. This line of thought was only possible because nationalRegistry would throw error if user.authorization !== nationalId

saevarma: Is the UI only sending the nationalId of the currently logged in user? If so I agree with Rikki about removing the params and read the nationalId from the current user.

apps/air-discount-scheme/backend/src/environments/environment.ts
```diff
@@ -12,6 +12,18 @@ const devConfig = {
     url: process.env.NATIONAL_REGISTRY_URL,
     username: process.env.NATIONAL_REGISTRY_USERNAME,
     password: process.env.NATIONAL_REGISTRY_PASSWORD,
+    authMiddlewareOptions: {
+      forwardUserInfo: true,
```
saevarma, line 16: You should have forwardUserInfo set to false as it only adds the User-Agent X-Forwarded-For headers to the request, which Þjóðskrá doesn't need.

apps/air-discount-scheme/backend/src/environments/environment.ts
```diff
@@ -12,6 +12,18 @@ const devConfig = {
...
+      tokenExchangeOptions: {
+        issuer: process.env.IDENTITY_SERVER_DOMAIN
+          ? `https://${process.env.IDENTITY_SERVER_DOMAIN}`
```
saevarma, line 19: Why doesn't the `IDENTITY_SERVER_DOMAIN` include the `https://`  as well?

apps/air-discount-scheme/backend/src/app/modules/user/user.controller.ts
JuliusGG: the params.nationalId is taken from ADS-API currentUser.decorator from cookies.

apps/air-discount-scheme/backend/src/environments/environment.ts
JuliusGG: Im using the same IDSDomain secret everywhere. So I excluded the https:// because Nextauth wants the domain without https in one place.
Adding https to the secret makes Nextauth contact IDS with https://https://identy......

But I might refactor these secrets completely.

apps/air-discount-scheme/web/infra/web.ts
```diff
@@ -7,12 +7,20 @@ export const serviceSetup = (services: {
...
+      NEXT_PUBLIC_NEXTAUTH_URL: {
+        dev: 'https://loftbru.dev01.devland.is',
+        staging: 'https://loftbru.staging01.devland.is',
+        prod: 'https://loftbru.island.is',
+      },
+      NEXT_PUBLIC_IDENTITY_SERVER_DOMAIN: {
+        dev: 'identity-server.dev01.devland.is',
+        staging: 'identity-server.staging01.devland.is',
+        prod: 'innskra.island.is',
+      },
```
JuliusGG, line 19: @pshomov @Snojo This is not working, on build it throws an error. What is the best practice when having infra values available for web production builds ?

apps/air-discount-scheme/api/src/environments/environment.ts
```diff
@@ -37,8 +39,8 @@ const prodConfig = {
     admins: process.env.ADMINS,
   },
   identityServerAuth: {
-    issuer: process.env.IDENTITY_SERVER_DOMAIN
-      ? `https://${process.env.IDENTITY_SERVER_DOMAIN}`
+    issuer: isProd
+      ? getStaticEnv('SI_PUBLIC_IDENTITY_SERVER_ISSUER_DOMAIN')
       : '',
     audience: '@vegagerdin.is',
```
pshomov, line 45: Why is the prod configuration with this condition? Seems redundant to me.

apps/air-discount-scheme/backend/src/environments/environment.ts
```diff
@@ -54,6 +63,18 @@ const prodConfig = {
     url: process.env.NATIONAL_REGISTRY_URL,
     username: process.env.NATIONAL_REGISTRY_USERNAME,
     password: process.env.NATIONAL_REGISTRY_PASSWORD,
+    authMiddlewareOptions: {
+      forwardUserInfo: false,
+      tokenExchangeOptions: {
+        issuer: isProd
+          ? getStaticEnv('SI_PUBLIC_IDENTITY_SERVER_ISSUER_DOMAIN')
+          : '',
```
pshomov, line 71: Same issue as above

apps/air-discount-scheme/backend/src/environments/environment.ts
```diff
@@ -65,8 +86,8 @@ const prodConfig = {
   },
   baseUrl: process.env.BASE_URL,
   identityServerAuth: {
-    issuer: process.env.IDENTITY_SERVER_DOMAIN
-      ? `https://${process.env.IDENTITY_SERVER_DOMAIN}`
+    issuer: isProd
+      ? getStaticEnv('SI_PUBLIC_IDENTITY_SERVER_ISSUER_DOMAIN')
       : '',
```
pshomov, line 91: Same comment as above

pshomov (commented):
A couple of minor issues

apps/air-discount-scheme/web/lib/environment.ts
```diff
@@ -1,21 +1,30 @@
...
 }
 
+if (isProd) {
+  if (process.env.NEXT_PUBLIC_NEXTAUTH_URL === undefined) {
+    throw new Error('Missing Nextauth_Url in web environment.')
+  }
+  if (process.env.NEXT_PUBLIC_IDENTITY_SERVER_DOMAIN === undefined) {
+    throw new Error('Missing IDS domain in web environment.')
+  }
+}
```
pshomov, line 21: I think it is best to remove this code.

apps/air-discount-scheme/api/src/environments/environment.ts
JuliusGG: Without it there will be errors when running it locally.

apps/air-discount-scheme/backend/src/environments/environment.ts
JuliusGG: Without it there will be errors when running it locally.

apps/air-discount-scheme/backend/src/environments/environment.ts
JuliusGG: Without it there will be errors when running it locally.

apps/air-discount-scheme/web/lib/environment.ts
JuliusGG: First I need to get the infra env values to work without breaking build.

apps/air-discount-scheme/api/src/environments/environment.ts
JuliusGG: `issuer: getStaticEnv('SI_PUBLIC_IDENTITY_SERVER_ISSUER_DOMAIN'),`
gives
`[App] Error	1/23/2022, 5:20:18 PM uncaughtException: document is not defined
ReferenceError: document is not defined
    at getStaticEnv (/home/jgg/Code/island.is/dist/apps/air-discount-scheme/api/webpack:/libs/shared/utils/src/lib/getStaticEnv.ts:5:5)`

apps/air-discount-scheme/api/infra/api.ts
```diff
@@ -12,6 +12,11 @@ export const serviceSetup = (services: {
         staging: 'loftbru.staging01.devland.is',
         prod: 'loftbru.island.is',
       },
+      SI_PUBLIC_IDENTITY_SERVER_ISSUER_URL: {
```
dabbeg, line 15: The `SI_PUBLIC_` prefix and `getStaticEnv` are used for react forntends that need injected environment variables like service-portal and application-system-form. It wold be better to just name it `IDENTITY_SERVER_ISSUER_URL`

dabbeg (changes requested):

apps/air-discount-scheme/api/infra/api.ts
JuliusGG: @dabbeg Can it stay in .env part of infra?

dabbeg: yes

apps/air-discount-scheme/web/infra/web.ts
dabbeg: I think it depends on whether the next backend or client code is referencing this.

If the client code is referencing it you will probably need to use the nextjs publicruntimeconfig

JuliusGG: I suppose it wouldnt work for local environment. This is how it would be running locally:
`{"loc":"http://localhost:4200","dev":"https://loftbru.dev01.devland.is","staging":"https://loftbru.staging01.devland.is","prod":"https://loftbru.island.is"}`
`[next-auth][error][signin_oauth_error] 
https://next-auth.js.org/errors#signin_oauth_error Invalid URL: https://[object Object]/connect/authorize?response_type=code`

dabbeg: Okay if I understand the Nextauth code correctly it is not being run in the browser so you don't need the publicruntimeconfig.

For anything local you would set the variable in the devConfig object in the environment.ts file.

For dev, staging and prod you would set the process.env.BLA in the prodConfig object in the environment.ts file and set the `BLA: { dev: 'stuff', staging: 'stuff', prod: 'stuff }` inside the infra/web.ts file

dabbeg: But if you are unsure that this works after it's deployed, you can set the `deploy feature` label on the PR and you can try this out without merging the PR.

dabbeg (approved):
LGTM

PR #6302, opened 2022-01-18
fix(judicial-system): Sidebar navigation fixes pt.2

oddsson:
# Sidebar navigation fixes pt.2

https://app.asana.com/0/1199153462262248/1201678975488492/f

## What

We forgot to add routes for extended investigation cases in the sidebar. Therefore, we always routed to a non-investigation case when the user used the sidebar to navigate.

## Why

It's a bug.

## Screenshots / Gifs

https://user-images.githubusercontent.com/3789875/150019914-f312bac7-b8bf-4961-9881-6a1ac3d84b55.mov

## Checklist:


apps/judicial-system/web/src/components/PageLayout/utils/index.ts
```diff
@@ -109,6 +110,12 @@ export const getSections = (
           workingCase || ({} as Case),
           activeSubSection,
         ),
-    getCourtSections(workingCase || ({} as Case), user, activeSubSection),
+    isRestrictionCase(workingCase?.type)
+      ? getCourtSections(workingCase || ({} as Case), user, activeSubSection)
```
gudjong, line 114: Do we not need a special getExtensionCourtSection for restriction cases?

oddsson: No we don't, because the routing is identical whether it is an extension case or not. You do however point out a flaw in this pr which is that we don't need a special "extension" section for investigation cases either. I've removed it and use "getInvestigationCaseCourtSections" instead.

gudjong (approved):

PR #6303, opened 2022-01-19
fix(service-portal): unable to scroll access list on mobile

disaerna:
# Accesses  - table mobile fix 

## What

Removing overflow 'visible' to 'auto'. It was preventing mobile users to scroll horizontal on table

## Why

🐞 

## Screenshots / Gifs

-
## Checklist:


eirikurn (approved):
Can't test this feature deployment :(

The overflow styles on this page are real messy because the island.is date picker component doesn't really work inside scrolling tables ...

Think we may need to either go for a completely different design for this, or upgrade DatePicker to support some kind of Portal system for the dropdown so it can be rendered outside the overflow container with correct positioning.

disaerna:
> Can't test this feature deployment :(
> 
> The overflow styles on this page are real messy because the island.is date picker component doesn't really work inside scrolling tables ...
> 
> Think we may need to either go for a completely different design for this, or upgrade DatePicker to support some kind of Portal system for the dropdown so it can be rendered outside the overflow container with correct positioning.

I agree - Keli (designer) is going to redesign this part in the next weeks 👍

PR #6306, opened 2022-01-19
feat(api-domains-driving-license): adding the reason for rejection from RLS

arnorhs:
## What

This adds a reason for why RLS will not let people move forward in the application

## Why

People who are applying and get the generic notice that people need to contact Syslumenn, will now get a user-friendly message indicating what the error may be.

## Checklist:


kksteini (approved):
LGTM

PR #6307, opened 2022-01-19
feat(web): Stepper - state machine

RunarVestmann:
# Stepper - state machine

Documentation: https://www.notion.so/StepperFSM-island-is-c73a41e32c7d40a2b157b75b3f156c2c

## What

* A new and improved Stepper, that's still under development, (called StepperFSM since we don't want to replace the current stepper yet) which will allow for radio and dropdown answers as well as a helper unit that allows you to see errors stated in the stepper configuration (intended to be used by those that set up the Stepper in Contentful). And also, the question answers are stored via query parameters so when you refresh or send the link to someone your previous answers will be remembered. 
* Now when you click on the EN button on a baby article then you'll actually visit the baby article's english page instead of jumping to the article's english page.

## Why

* As of right now the current Stepper only allows for yes and no answers and doesn't allow you to see what your previous answers were

## Information and future plans

This is a Stepper that is intended to be used on baby articles. When setting up the Stepper in Contentful you'll have to provide a JSON configuration for the Stepper state machine. This can be hard to do and that's why we made a helper unit that's only visible when the web is running locally or on the dev environment (there's even a fail-safe since you need to click four times on a question title in order to make the helper unit appear just in case the helper could be viewed in production). This helper unit shows you what errors reside in your JSON configuration and help you in defining the state machine correctly. In the future we hope to improve the helper unit in a way that the user could even draw their own state diagram in the helper unit which will generate the appropriate configuration that the user can then copy and paste into Contentful.

If you want to run this branch locally here are a few Stepper examples to check out:

(Very simple example that's shown in the documentation)
http://localhost:4200/stepper-demo/example-stepper

(Notice that in the helper here there is an error but everything still works, that's because there's an extra unnecessary state in the configuration that isn't used)
http://localhost:4200/stepper-demo/stepper-aritun

(Simple example that demonstrates the optionsFromSource field that a step can have, meaning that we read options from a UI configuration and in this example we are utilizing a predetermined list of countries)
http://localhost:4200/stepper-demo/stepper-demo-page

## Screenshots / Gifs

image

image

## Checklist:


apps/web/components/StepperFSM/StepperFSM.tsx
```diff
@@ -0,0 +1,461 @@
...
+      props.stepper?.steps
+        ?.map((s) => s.slug)
+        ?.map((s) => getStepBySlug(props.stepper, s)) ?? []
+
+    const stepConfigErrors = steps.map((step) => ({
+      step,
+      errors: validateStepConfig(step),
+    }))
+
+    const showStepperConfigHelper = STEPPER_HELPER_ENABLED
```
kksteini, line 128: Curious about this. Is there a reason the `showStepperConfigHelper` is declared instead of using `STEPPER_HELPER_ENABLED`?

apps/web/components/StepperFSM/StepperFSM.tsx
```diff
@@ -0,0 +1,461 @@
...
+              underline="small"
+              underlineVisibility="always"
+              color="blue400"
+              shallow={true}
+              href={{
+                pathname: urlWithoutQueryParams,
+                query: query,
+              }}
+            >
+              {activeLocale === 'en' ? 'Change' : 'Breyta'}
```
kksteini, line 280: Can this not be achieved with translations and passed as props, perhaps?

apps/web/components/StepperFSM/StepperFSM.tsx
```diff
@@ -0,0 +1,461 @@
...
+                },
+              })
+              .then(() => window.scrollTo(0, 0))
+
+            if (!transitionWorked) {
+              setTransitionErrorMessage(
+                activeLocale === 'en'
+                  ? 'Sadly, the next step could not be loaded'
+                  : 'Því miður gekk ekki að hlaða niður næsta skrefi',
+              )
```
kksteini, line 330: Can this not be achieved with translations?

apps/web/components/StepperFSM/StepperFSM.tsx
```diff
@@ -0,0 +1,461 @@
...
+          />
+        )
+      })
+
+    if (currentStepType === STEP_TYPES.QUESTION_DROPDOWN)
+      return (
+        <Select
+          name="step-option-select"
+          noOptionsMessage={
+            activeLocale === 'en' ? 'No options' : 'Enginn valmöguleiki'
```
kksteini, line 387: Same here.
In the future, Polish might be added f.x.

apps/web/components/StepperFSM/StepperFSMUtils.tsx
```diff
@@ -0,0 +1,386 @@
...
+
+const stepContainsQuestion = (step: Step) => {
+  return (
+    step.subtitle &&
+    step.subtitle.length > 0 &&
+    step.subtitle[0].__typename === 'Html' &&
+    step.subtitle[0].document.content.length > 0 &&
+    step.subtitle[0].document.content[0].content &&
+    step.subtitle[0].document.content[0].content.length > 0
+  )
```
kksteini, line 356: Assuming step is defined, isn't it enough to do:
`step?.subtitle[0]?.document?.content[0]?.content?.length > 0`?
image

kksteini: Along with the `html` typename check, I assume

kksteini (approved):

apps/web/components/StepperFSM/StepperFSM.tsx
RunarVestmann: I agree with your comment. We will address this in the next PR related to the Stepper.

PR #6314, opened 2022-01-19
feat(application-system): Removing attachment field from resolver response

alexandradilja:
# ...

Remove attachment field from graphql queries so that the templates can not access attachment urls in aws

## What

Removed field from application model in `libs/api/domains/application/src/lib/application.model.ts`. Add and delete attachment mutations query after application id for return. 

No templates are using the attachment filed so no changes needed there. 

## Why

Security reasons.

## Checklist:


libs/application/core/src/types/Application.ts
```diff
@@ -28,7 +28,7 @@ export interface ActionCardMetaData {
   }
 }
 
-export interface Application {
+export interface ApplicationModel {
```
saevarma, line 31: To have this more explicit and clear in the naming how about the calling model with attachments `ApplicationWithAttachments`?

saevarma (changes requested):
Looks good just question about the name of the models

libs/application/core/src/types/Application.ts
alexandradilja: Yes! good idea!

libs/application/core/src/types/Application.ts
```diff
@@ -45,3 +45,5 @@ export interface Application {
   progress?: number
   status: ApplicationStatus
 }
+
+export type Application = Omit<ApplicationWithAttachments, 'attachments'>
```
miamiultras, line 49: maybe the other way around and we can stick with interfaces

```
interface Application {
...
}
```

```
interface ApplicationWithAttachments extends Application {
attachments...
}
```

alexandradilja: Yes sure!

miamiultras (approved):

Thorrik58 (approved):
LGTM

apps/application-system/api/src/app/modules/payment/payment.service.ts
```diff
@@ -12,7 +12,7 @@ import type { User } from '@island.is/auth-nest-tools'
 import { CreateChargeResult } from './payment.type'
 import { logger } from '@island.is/logging'
 import { ApolloError } from 'apollo-server-express'
-import { Application as ApplicationModel } from '../application/application.model'
+import { Application as ApplicationWithAttachments } from '../application/application.model'
```
saevarma, line 15: Why import the `Application` model without attachments to rename it to `ApplicationWithAttachments` ?

apps/financial-aid/api/src/app/modules/application/application.resolver.ts
```diff
@@ -6,7 +6,7 @@ import { BackendAPI } from '../../../services'
 
 import {
   ApplicationFiltersModel,
-  ApplicationModel,
+  ApplicationWithAttachments,
```
saevarma, line 9: I think you accidentally renamed the `financial-aid` related stuff?

saevarma (changes requested):

apps/application-system/api/src/app/modules/payment/payment.service.ts
alexandradilja: yes, looks like it.. I'll fix that

saevarma (approved):
LGTM

libs/application/template-api-modules/src/lib/modules/templates/children-residence-change/children-residence-change.service.ts
```diff
@@ -20,7 +20,7 @@ import {
   generateSyslumennNotificationEmail,
   transferRequestedEmail,
 } from './emailGenerators'
-import { Application } from '@island.is/application/core'
+import { ApplicationWithAttachments as Application } from '@island.is/application/core'
```
Rikkiabb, line 23: We are not using Attachments so we can use `Application` instead 😄

darrikonn (approved):

Margretol (approved):

johannaagma (approved):

PR #6334, opened 2022-01-20
fix(service-portal): missing variable in query

disaerna:
# Service Portal - Documents - Missing variable

## What

Added missing "opened" variable in query so both styles for unopened and red bubble works!

## Why

It's not working!

## Checklist:


thordurhhh (approved):
👍

PR #6338, opened 2022-01-20
fix(judicial-system): Autofills isCustodyIsolation

gudjong:
# Autofills isCustodyIsolation

https://app.asana.com/0/1199153462262248/1201634243867932/f

## What

☝️ 

## Why

it should.

## Checklist:

### DIFF
--- a/apps/judicial-system/web/src/utils/hooks/use-case/useCase.tsx
+++ b/apps/judicial-system/web/src/utils/hooks/use-case/useCase.tsx
@@ -12,6 +12,7 @@ import type {
   CaseTransition,
   RequestSignatureResponse,
   UpdateCase,
+  SessionArrangements,
 } from '@island.is/judicial-system/types'
 
 import { CreateCaseMutation } from './createCaseGql'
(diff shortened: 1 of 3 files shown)


apps/judicial-system/web/src/utils/hooks/use-case/useCase.tsx
```diff
@@ -313,6 +316,38 @@ const useCase = () => {
     [updateCase],
   )
 
+  const autofillSessionArrangements = useMemo(
```
oddsson, line 319: The bodies of these functions are almost identical to the `autofill` function. Would it be possible to change the value parameter in the autofill function to have type `string | SessionArrangements | boolean` and use that instead? Inside autofill we could cast the value parameter to string if needed?

oddsson (changes requested):
One question

apps/judicial-system/web/src/utils/hooks/use-case/useCase.tsx
gudjong: It is possible, but would be somewhat messy, as we would need to base our decisions on field names rather than field types. Also, the logic for the boolean autofill function is different from the other functions. Instead of checking for falsie current values we need to check for undefined/null.

The reason, I created a separate function for the SessionArrangements type was to get rid of a typescript error.
The reason, I created a separate function for the boolean type was the different logic.

If you don't mind, I would like to leave it like this for now and come pack to it later for possible refactoring.

oddsson (approved):

PR #6344, opened 2022-01-21
fix(fa): Uses staffguard for staffcontroller

Rikkiabb:
# ...
https://app.asana.com/0/1201664508282204/board

## What
Adds staff guard to whole staff controller

## Why
Security fix - The staff controller is not properly guarded. There's a couple of StaffGuards on individual methods, but not the entire controller. - Get staff member endpoint is not properly guarded. It allows anyone to fetch a staff member regardless of permission

## Checklist:


MargretFinnboga (approved):
🔒 🔒

darrikonn (approved):

PR #6348, opened 2022-01-21
fix(fa): Security fixes

Rikkiabb:
# ...
https://app.asana.com/0/1201664508282204/board

## What
Fixing security issues

TODO: Adds scopes to staging and prod on IDS

## Why

- The municipality controller is not properly guarded. It allows anyone to fetch a municipality regardless of permission
- The file controller is trying to guard for web veita but there's only a role check that is passed from the client and thus can be bypassed by any client.
- The application controller does not have global controller guard. The update application endpoint is totally unguarded as anyone with an authentication can update any application.

## Checklist:


apps/financial-aid/backend/src/app/modules/file/test/fileControllerGuards.spec.ts
```diff
@@ -38,7 +50,7 @@ describe('FileController - Creates a new signed url guards', () => {
     )
   })
 
-  it('should have one guard', () => {
+  it('should have two guards', () => {
     expect(guards).toHaveLength(1)
```
MargretFinnboga, line 54: it('should have two guards', () => {
    expect(guards).toHaveLength(1) 

there is a mismatch here? 👀

MargretFinnboga (approved):
One minor comment on test text, otherwise looks good 💥 💣

apps/financial-aid/backend/src/app/modules/application/application.controller.ts
```diff
@@ -272,6 +271,8 @@ export class ApplicationController {
     return this.applicationService.create(application, user)
   }
 
+  @UseGuards(ScopesGuard)
+  @Scopes(MunicipalitiesFinancialAidScope.write)
```
darrikonn, line 275: This endpoint can be bypassed.

Don't you need the application guard?

apps/financial-aid/backend/src/app/modules/file/file.controller.ts
```diff
@@ -34,6 +36,7 @@ export class FileController {
   @Get('url/:id')
   @UseGuards(RolesGuard)
   @RolesRules(RolesRule.VEITA)
+  @Scopes(MunicipalitiesFinancialAidScope.read)
```
darrikonn, line 39: This is still open for attacks.
The scope is set by the client, so it doesn't really guard this. You can get a valid access token (and its scopes) from Veita, even though you don't have access and use that to call this endpoint.
This needs a StaffGuard.

apps/financial-aid/backend/src/app/modules/municipality/municipality.controller.ts
```diff
@@ -26,13 +26,16 @@ import {
...
+import { MunicipalitiesFinancialAidScope } from '@island.is/auth/scopes'
 
 @UseGuards(IdsUserGuard)
 @Controller(`${apiBasePath}/municipality`)
 @ApiTags('municipality')
 export class MunicipalityController {
   constructor(private readonly municipalityService: MunicipalityService) {}
 
+  @UseGuards(ScopesGuard)
+  @Scopes(MunicipalitiesFinancialAidScope.read)
```
darrikonn, line 38: ditto on last comment.

apps/financial-aid/backend/src/app/modules/staff/staff.controller.ts
```diff
@@ -40,10 +41,8 @@ export class StaffController {
     type: StaffModel,
     description: 'Gets staff by nationalId',
   })
-  async getStaffByNationalId(
-    @Param('nationalId') nationalId: string,
-  ): Promise<StaffModel> {
-    const staff = await this.staffService.findByNationalId(nationalId)
+  async getStaffByNationalId(@CurrentUser() user: User): Promise<StaffModel> {
```
darrikonn, line 44: The endpoint below (need to comment here from Github) is still unguarded (Get(id/:id))

darrikonn (changes requested):

apps/financial-aid/backend/src/app/modules/municipality/municipality.controller.ts
Rikkiabb: The municipality data is not sensitive. At least for the time being. It has the following columns: 

`id`, `municipality_id`, `name`, `homepage`, `active`, `individual_aid_id`, `cohabitation_aid_id`, `email` and `rules_homepage`

This data is then used to see if user can apply for financial aid in the applicants municipality and to show data relevant to that municipality. 

This endpoint is for example called after the applicant approves data gathering to see if their municipality supports financial aid.

With that in mind, do you feel it should be better guarded?

apps/financial-aid/backend/src/app/modules/staff/staff.controller.ts
darrikonn: That's just the RolesGuard ... which is just checking for a service that's passed by the client.
If I get an authentication token from Veita (which I can do regardless of permission) I'll get the correct service and thus bypass this protection.

apps/financial-aid/backend/src/app/modules/municipality/municipality.controller.ts
darrikonn: I did my initial report based on who was supposed to call the endpoints. If I remember correctly, only web-veita was supposed to call this. That's why I flagged it in the report.

So it's up to you :).

Rikkiabb: This is called from `useMunicipality` hook: https://github.com/island-is/island.is/blob/3666ae66b8c6e0b02d3f91f1da39f151dc552c74/libs/financial-aid/shared/src/components/hooks/useMunicipality.ts 

Which is used in both clients, will resolve conversation but feel to unresolve if you feel the need :)

apps/financial-aid/backend/src/app/modules/staff/staff.controller.ts
darrikonn: see it now .. you're right!

darrikonn (approved):
Nice work!

 LGTM

PR #6358, opened 2022-01-24
fix(fa): Adds protobuf to esbuild

Rikkiabb:
# ...
https://app.asana.com/0/1201664508282204/1201698304881585

## What
Fixing broken environments

## Why
we need dem envs


MargretFinnboga (approved):
🥇

PR #6410, opened 2022-01-27
fix(user-notification): Infrastructure fixes

arnorhs:
## What

- setting correct user profile env variables
- running the worker the same way as the web server, but also starts the
  worker, allowing us to re-use the health checks and liveness checks

## Why

So we can use the already built and tested health checks for the notification worker and so that the config works for the user notification user profile stuffs

## Checklist:


maggisk (approved):

pshomov (approved):
LGTM

PR #6411, opened 2022-01-28
fix(api-domains-driving-license): allowing expired licenses for student lookup

arnorhs:
## What

Looks up in api v1 for all licenses of a student in case they have an expired one that has their name.

## Why

So their teacher is able to submit their driving assessment

## Checklist:


barabrian (approved):
LGTM

PR #6423, opened 2022-01-28
feat(auth): Integrate with RSK's new Procuring service

eirikurn:
## What

Create client library for the new procuring API from RSK. Removed old clients/rsk/v2 which has been deprecated.

## Why

To implement procuring holder delegations in the new delegation system with a proper API from RSK.

## Checklist:


libs/auth-api-lib/src/lib/services/delegations.service.ts
```diff
@@ -369,7 +368,7 @@ export class DelegationsService {
 
   /**
    * Finds all companies that the user is procuration holder from CompanyRegistry
-   * @param toNationalId
+   * @param user.nationalId
```
darrikonn, line 371: just user

darrikonn (approved):
Nice

PR #6426, opened 2022-01-31
feat(web): Organization page alert banner support

RunarVestmann:
# Organization page alert banner support

## What

* Added an alert banner field to Organization Pages
* Now an organization can define an alert banner in Contenful, link it to an organization page and it will show up on the web

## Why

* We want to prevent having to hotfix in a change when an organization wants to show an alert, instead the organization can add an alert banner themselves via Contentful.

## Screenshots / Gifs

In Contentful there's a new field on the OrganizationPage model:
image

An organization can choose to hide an Alert Banner in Contentful:
image
image

An organization can also choose to show an Alert Banner in Contentful:
image
image

## Checklist:


apps/web/components/Organization/OrganizationAlert/OrganizationAlert.tsx
```diff
@@ -0,0 +1,87 @@
...
+        href: linkResolver(alertBanner.link.type as LinkType, [
+          alertBanner.link.slug,
+        ]).href,
+        title: alertBanner.linkTitle,
+      }),
+  }
+
+  const handleDismiss = () => {
+    if (alertBanner.dismissedForDays === 0) return
+    Cookies.set(alertBannerId, 'hide', {
```
arnorhs, line 65: would it make sense to have something higher up in the state tree deciding whether or not to show this banner?

eg. how would you reuse the logic for another type of alert?

Would local storage be more appropriate for storing this kind of setting? Rather than a cookie.

arnorhs (changes requested):

apps/web/components/Organization/OrganizationAlert/OrganizationAlert.tsx
RunarVestmann: I used a cookie since that's what is used for the homepage of island.is, line 296 in apps/web/layouts/main.tsx
image

There's a specific functionality for how long the dismiss is active so I did the same thing that was already in the code.

RunarVestmann: And I decided not to have this state higher up since you might've noticed that I could have skipped using the OrganizationAlert component completely and only used the AlertBanner component which would have bloated up the OrganizationWrapper component, so that's my reasoning.

arnorhs: makes sense

arnorhs (approved):

siggifv (approved):

PR #6431, opened 2022-01-31
Exclude user-notification from feature deployment

dabbeg:
# ...

## What

## Why

## Screenshots / Gifs

## Checklist:

### DIFF
--- a/infra/src/feature-env.ts
+++ b/infra/src/feature-env.ts
@@ -1,5 +1,7 @@
 import yargs from 'yargs'
+import { hideBin } from 'yargs/helpers'
 import AWS from 'aws-sdk'
+
 import {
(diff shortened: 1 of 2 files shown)


infra/src/feature-env.ts
```diff
@@ -1,5 +1,7 @@
 import yargs from 'yargs'
+import { hideBin } from 'yargs/helpers'
```
pshomov, line 2: What's that all about?

pshomov (approved):
Yes, I think that's it.

infra/src/feature-env.ts
dabbeg: just changed it from using require into import and moved it next to the yargs import

PR #6433, opened 2022-01-31
feat(judicial-system): Enable court users to modify case dates

oddsson:
# Enable court users to modify case dates 

https://app.asana.com/0/1199153462262248/1201437594297191/f

## What

Enable court users to modify validToDate and/or isolationToDate active custody cases.

## Why

Sometimes, errors are made in the hearing that need to be corrected after a verdict has been made.

## Screenshots / Gifs

Screenshot 2022-02-01 at 12 01 39

image

## Checklist:


apps/judicial-system/web/src/routes/Shared/SignedVerdictOverview/SignedVerdictOverviewForm.tsx
```diff
@@ -131,60 +134,9 @@ const SignedVerdictOverviewForm: React.FC<Props> = (props) => {
...
-        <Box component="span">
-          {`${
-            isTravelBan ? 'Farbann' : 'Gæsla' // ACCEPTING
-          } til ${formatDate(theCase.validToDate, 'PPP')} kl. ${formatDate(
-            theCase.validToDate,
-            TIME_FORMAT,
-          )}`}
-        </Box>
-      </>
+      user?.role === UserRole.JUDGE && workingCase.type === CaseType.CUSTODY
```
gudjong, line 139: Consider who should be allowed to change.

- Here, only judges are allowed to change dates.
- Server side, both judges and registrar are allowed to change dates.
- In the examples from Guðný in Asana prosecutors are changing dates.

Who should actually be allowed to change dates?

gudjong (changes requested):

apps/judicial-system/web/src/routes/Shared/SignedVerdictOverview/SignedVerdictOverviewForm.tsx
gudjong: Note also that in Guðný's example, the prosecutor's office is also shown.

gudjong (approved):

PR #6440, opened 2022-02-01
fix(fa): Fixing accepted aid final amount

MargretFinnboga:
# Fixing accepted aid modal and more

https://app.asana.com/0/1201664508282204/1201722624788136

## What

- Changing accepted aid modal from object to array 
- Added  to application event comment when aid is accepted  "Samþykkt upphæð: kr. {accepted aid }.-"
- Fixing time stamps
- Changing staffComment component to timelineComment component since accepted aid will also be showing
- Using familyStatus instead of spouse.nationalId to calculate aidAmount
- Removing CreatedAmount 
- Removed isSpouseDataNeeded 

## Why

- Otherwise it takes random accepted aid, we want the newest accepted aid 
- To show in timeline 
- Date format in MM was not right, noticed by accident screenshot below shows the two timestamps 
- Because the comment is also shown when aid is accepted more appropriate
- I change this because we use familyStatus in osk and I wanted consistency
-  Neither is used

## Screenshots / Gifs

Screenshot 2022-02-01 at 08 27 29

## Checklist:


Rikkiabb (approved):
nice

PR #6441, opened 2022-02-01
fix(fa): fixing comment section in Veitu

MargretFinnboga:
# Fixing comment section in Veitu

https://app.asana.com/0/1201722624788165/1201749252246245

## What

- Added in application guard to check for body.applicationId too 

## Why

- When only event is created it doesn't have params.id in the request the id is applicationId in the body

## Screenshots / Gifs

## Checklist:


Rikkiabb (approved):
👌

PR #6444, opened 2022-02-01
chore(core): Update node-fetch

eirikurn:
To fix dependabot alert


darrikonn (approved):
Does the major upgrade not need any changes in our codebase?

PR #6457, opened 2022-02-01
fix(air-discount-scheme): Error handling for Thjodskra

JuliusGG:
# ...

## What

Fixing OOM problem with Loftbru by fixing National registry edge case handling.

## Why

## Screenshots / Gifs

Working even if national registry cannot find individual children of user.

Screenshot from 2022-02-01 21-31-08

## Checklist:


apps/air-discount-scheme/backend/src/app/modules/user/user.service.ts
```diff
@@ -82,4 +92,15 @@ export class UserService {
   async getUserInfoByNationalId(nationalId: string): Promise<User | null> {
     return this.getUserByNationalId<User>(nationalId, User)
   }
+
+  async getMultipleUsersByNationalIdArray(ids: string[]): Promise<User[]> {
```
eirikurn, line 96:
```suggestion
  async getMultipleUsersByNationalIdArray(ids: string[]): Promise<Array<User | null>> {
```

apps/air-discount-scheme/backend/src/app/modules/user/user.service.ts
```diff
@@ -82,4 +92,15 @@ export class UserService {
...
+    const allUsers = ids.map(
+      async (nationalId) =>
+        (this.getUserInfoByNationalId(nationalId) as unknown) as User,
+    )
+
+    const result = (await Promise.all(allUsers)) as User[]
```
eirikurn, line 102: Are all of these type casts necessary/correct? You need to handle the nulls somewhere.

apps/air-discount-scheme/backend/src/app/modules/user/user.controller.ts
```diff
@@ -80,12 +80,19 @@ export class PrivateUserController {
...
-      ),
-    ])) as User[]
-    return users.filter((user) => user) as User[]
+
+    // Adding User in beginning of array so order is correct.
+    relations.unshift(authUser.nationalId)
+
+    const userAndRelatives = (await this.userService.getMultipleUsersByNationalIdArray(
+      relations,
+    )) as User[]
```
eirikurn, line 89: Necessary typecast?

apps/air-discount-scheme/backend/src/app/modules/nationalRegistry/nationalRegistry.service.ts
```diff
@@ -204,19 +204,24 @@ export class NationalRegistryService {
...
     } = await this.httpService
       .get(`${this.baseUrl}/general-lookup?ssn=${nationalId}`)
       .toPromise()
 
+    if (response && response.data[0].error) {
+      this.logger.error(
+        `National Registry general lookup failed for User<${nationalId}> due to: ${response.data[0].error}`,
+      )
+      return null
+    }
```
eirikurn, line 223: More swallowing but this is more understandable, since it is used in an N+1 lookup in the backend. Do you want the chance for partial results and hide any errors from the user, or do we want to show all or nothing with a clear error in the UI.

Generally I would return "missing" national ids as null or 404, but throw other errors.

JuliusGG: My line of thought is to solve the error boundaries of dev/staging while test data is not sufficient, to me it is better to have partial results, silent error and usable site. 

This shouldn't be a problem on production since all individual SSNs exist - however every error would be logged and if no users are returned then the controller would throw error.

apps/air-discount-scheme/backend/src/app/modules/user/user.service.ts
```diff
@@ -16,27 +16,37 @@ import {
...
+          id: authUser.nationalId,
+        })
+    } catch (e) {
+      this.logger.error(
+        'NatReg xRoad forsja: ' + e.name + ' - ' + e.message + ' - ' + e.url,
       )
-      .einstaklingarGetForsja(<EinstaklingarGetForsjaRequest>{
-        id: authUser.nationalId,
-      })
+    }
```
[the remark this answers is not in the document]
JuliusGG: Changed so if this service fails it'll return a null and controller throws an error.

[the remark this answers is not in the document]
JuliusGG: Hopefully we can switch to Enhanced fetch in future.

apps/air-discount-scheme/backend/src/app/modules/user/user.service.ts
```diff
@@ -82,4 +93,17 @@ export class UserService {
   async getUserInfoByNationalId(nationalId: string): Promise<User | null> {
     return this.getUserByNationalId<User>(nationalId, User)
   }
+
+  async getMultipleUsersByNationalIdArray(
+    ids: string[],
+  ): Promise<Array<User> | null> {
```
eirikurn, line 99:
```suggestion
  ): Promise<Array<User | null>> {
```
Never returns null, but can return null inside the array.

apps/air-discount-scheme/backend/src/app/modules/nationalRegistry/nationalRegistry.service.ts
eirikurn: Yeah, we had a similar issue in the delegation system and followed the same line of thought.

This is something that GraphQL can actually help with. Instead of having one resolver for everything, it's possible to split it up, so if one resolution fails, you might still get partial data while still having the partial error sent to the client/UI. Anyways...

apps/air-discount-scheme/backend/src/app/modules/user/user.service.ts
```diff
@@ -16,27 +16,38 @@ import {
...
+            environment.nationalRegistry
+              .authMiddlewareOptions as AuthMiddlewareOptions,
+          ),
+        )
+        .einstaklingarGetForsja(<EinstaklingarGetForsjaRequest>{
+          id: authUser.nationalId,
+        })
+    } catch (e) {
+      this.logger.error(
+        'NatReg xRoad forsja: ' + e.name + ' - ' + e.message + ' - ' + e.url,
```
eirikurn, line 47: Why catch, log, return null, then throw a different error?

It's better to let original errors bottle up so they can be logged and handled consistently. It also saves a lot of boilerplate error handling code all over the place, which might be spotty.

E.g. this log line might miss some important properties from the error object, and it doesn't contribute much which is not apparent from the original stack trace or other logs.

apps/air-discount-scheme/backend/src/app/modules/user/user.service.ts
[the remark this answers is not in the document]
eirikurn: This national registry client is using Enhanced Fetch, that's why the backend has pretty good logs for these errors.

It's only the api project which is using RESTDataSource to call the backend, but that too throws adequate error objects. Still makes no sense why the server crashes with OOM right after...

apps/air-discount-scheme/backend/src/app/modules/user/user.service.ts
eirikurn: So this whole catch and and the null logic on the other side should be removed IMO.

apps/air-discount-scheme/backend/src/app/modules/user/user.controller.ts
```diff
@@ -79,13 +79,20 @@ export class PrivateUserController {
...
+    const userAndRelatives = await this.userService.getMultipleUsersByNationalIdArray(
+      relations,
+    )
+
+    if (Array.isArray(userAndRelatives) && userAndRelatives[0] === null) {
+      throw new Error(
+        'Could not find NationalRegistry records of both User and relatives.',
+      )
+    }
+    return userAndRelatives as Array<User>
```
eirikurn, line 96: "this is a lie" since there might be nulls in there.

Should either filter them out here, or update the schema and return type to match what is actually returned.

apps/air-discount-scheme/backend/src/app/modules/user/user.service.ts
```diff
@@ -16,27 +16,35 @@ import {
...
+            authUser,
+            environment.nationalRegistry
+              .authMiddlewareOptions as AuthMiddlewareOptions,
+          ),
+        )
+        .einstaklingarGetForsja(<EinstaklingarGetForsjaRequest>{
+          id: authUser.nationalId,
+        })
+    } catch (e) {
+      this.logger.error(e)
```
eirikurn, line 46: In my previous reviews I was missing the fact that this endpoint should still return the authenticated user, even though the forsja call fails. Now I understand why you are handling this error at all.

The general point is that errors should always be handled as high up as possible, either by business logic, or by the user. Arguably it should not be in this function but somewhere higher, but it doesn't really matter in this case since the lines between client logic and domain/business logic is a bit blurry here.

eirikurn (approved):
Sweet!

apps/air-discount-scheme/backend/src/app/modules/user/user.controller.ts
JuliusGG: fixed :+1:

apps/air-discount-scheme/backend/src/app/modules/user/user.service.ts
JuliusGG: I find this very reasonable.
I guess initially I went overboard with logging errors after trying to debug staging on release - The fact that there were so few traces astounded me.

kksteini (approved):

PR #6458, opened 2022-02-02
chore(api): Strict types in api project and all its libs

eirikurn:
## What

Fix all strict type issues in the api project and all its libs. Configure strict: true.

Note that I generally kept existing null/undefined (mis)handling using "!", especially when it comes to the `environment`. This at least highlights some of the issues that `@island.is/nest/config` aims to solve. Would be great to clean this up in future PRs.

This PR should generally only affect (and improve) type checking, not actual code/logic. Please let me know if you spot potential issues.

## Why

To catch more errors going forward. Also to get a smarter TypeScript since some type-checking tricks just don't work except in strict typescript.

Now we just need to fix the `web` project and we'll be rolling in strict code! :)

## Checklist:


JuliusGG (approved):

libs/regulations/src/lib/utils.ts
```diff
@@ -59,11 +59,7 @@ export function toISODate(
       date = undefined
     }
   }
-  // TODO: Remove this silly `as Date` type coersion as soon as
-  // the "api" project turns on `strict: true`
```
thordurhhh, line 63: 🤝

libs/api/domains/icelandic-names-registry/src/lib/models/icelandicName.model.ts
```diff
@@ -25,19 +25,19 @@ export class IcelandicName implements TIcelandicName {
   @IsEnum(StatusType)
   status!: StatusType | null
 
-  @Field({ nullable: true })
+  @Field(() => Boolean, { nullable: true })
   @IsBoolean()
   visible!: boolean | null
```
darrikonn, line 30: Shouldn't these be:
```typescript
visible?: boolean
```

(same goes for other fields)

darrikonn (commented):
Nice work ... had one question; once addressed, I'll approve.

thordurhhh (approved):
Nice 👍 
Looks good.

Rikkiabb (approved):
👏

libs/api/domains/icelandic-names-registry/src/lib/models/icelandicName.model.ts
eirikurn: In strict TypeScript, the Field decorator can not reflect union types:

```
// BAD
@Field({ nullable: true })
description: string | undefined

@Field({ nullable: true })
description!: string | null

@Field({ nullable: true })
description?: string | null

// GOOD
@Field({ nullable: true })
description?: string

description: string | undefined

description!: string | null

description?: string | null
```

I preferred the first "GOOD" option and tried to make it consistent like that, but that caused issues when casting these GraphQL types back and forth with other types.

These other types are usually from client libraries and often based on OpenAPI schemas which sometimes marks properties as nullable (null) and sometimes as not required (undefined). Depending on the flow of data, this causes typescript errors if they don't match correctly. 

So rather than chasing down too many rabbit holes (some of these types come from external OpenAPI schemas), I just fixed the remaining issues by specifying the type in the `@Field` decorator.

darrikonn (approved):

bensijons (approved):

PR #6460, opened 2022-02-02
fix(web): Organization subpage language toggle

RunarVestmann:
# Fix: Organization subpage language toggle

## What

* Previously if you'd toggle the language on an Organization subpage then the Organization page would come up instead of the subpage you were currently viewing

## Why

* We don't want the language toggle on an Organization subpage to move you to a different page

### Before

If I would press the EN button in the top right corner, then I would leave this subpage
image

Here's the page I go to when I press the button
image

### After
After this change then we actually stay on the subpage
image

Here's the same subpage but in English
image

## Checklist:


kksteini (approved):
LGTM

siggifv (approved):

PR #6473, opened 2022-02-03
feat(judicial-system): Search for defendants in national registry

oddsson:
# Search for defendants in national registry

https://app.asana.com/0/1199153462262248/1200221568911085/f

## What

When a prosecutor enters a defendant's national id, we perform a lookup in the national registry and fill in the name, address and gender.

## Why

This saves the prosecutors from having to manually type in the name, address and gender. This also ensures that the info provided is correct.

## Screenshots / Gifs

https://user-images.githubusercontent.com/3789875/152319242-6247f208-9a34-4075-bd55-e44732c549cc.mov

## Checklist:


gudjong (approved):

darrikonn:
So, we actually had to solve this for the delegation system (access control part of service portal).
The UX of your implementation could be misunderstood.. are the fields read-only? can I modify them after they've been prefilled? etc.

We came up with this solution, which we interpreted as a better UX:

https://user-images.githubusercontent.com/5694851/152359726-0e1c1bae-423f-49b1-9734-49b3cbc0dc1f.mov

darrikonn (changes requested):
Please checkout me previous comment :)

apps/judicial-system/web/pages/api/nationalRegistry/getByNationalId/index.ts
```diff
@@ -0,0 +1,27 @@
+import { NextApiRequest, NextApiResponse } from 'next'
+
+import { NationalRegistryResponse } from '@island.is/judicial-system-web/src/types'
+
+async function getByNationalId(
+  nationalId: string,
+): Promise<NationalRegistryResponse> {
+  const response = await fetch(
+    `https://api.ja.is/skra/v1/people?kennitala=${nationalId}`,
```
eirikurn, line 9: Why is this using a new national registry integration through Já instead of the new API from skra.is?

hrefnalind: I am also interested to hear why we are using ja.is

oddsson:
Thanks for the feedback @darrikonn. I agree that this is a bit confusing as is, because the gender is above the national id. The flow is awkward because the user selects the gender, then the national id and on blur, the other fields are populated. However, we are implementing a new layout for these input fields that will look like this 

Screenshot 2022-02-03 at 15 30 30

By moving the national id to the top we are hoping the flow becomes clearer. The fields that are filled are not read only, users should be able to change the values as they please. We want to start by implementing this and see what our users think and then make this clearer if needed.

apps/judicial-system/web/pages/api/nationalRegistry/getByNationalId/index.ts
oddsson: Right, I feel a bit strange answering this since I'm not directly responsible for the decision to use Já's API over Skrá's API. What I can say is that this decision was made by @baldur-kolibri and our PO (Guðlaug Þórhallsdóttir at Dómsmálaráðuneytið). They included Vigfús from SÍ who gave the go ahead.

I realise that this is a bit of a non answer. For more details, please contact any or all of the people mentioned ☝️, they are all on Slack.

eirikurn: Hehe. Thanks, "Mr Messenger"! 😄

eirikurn (approved):

apps/judicial-system/web/pages/api/nationalRegistry/getByNationalId/index.ts
baldur-kolibri: Back from a week long vacation 😎

The decision to use a standard out-of-the-box API was taken by Dómsmálaráðuneytið for both economic and practical reasons. 

We (DMR, Kolibri and Fúsi) explored all options - Einstaklingur, Soffía and the standard APIs used by Icelandic organizations (powered by Já, Creditinfo, Advania, Ferli, ..) for this use case and the third option came out as the best fit for Réttarvörslugátt, i.e. lowest cost and risk for DMR both to get started and operate. People from the Island.is project office were involved and/or informed before any code was written.

Note that this is not the typical Island.is use case where an authenticated individual needs government services and agrees to have their data pulled from ÞÍ. This is a typical lookup of a third party where the consumer is an internal system. We fully support the journey of Einstaklingur to maturity for use by Island.is (application/auth/Mínar síður) and Kolibri has experience using it in two projects (for Island.is and Sambandið).

If you think there is benefit in discussing the reasons behind this choice further, I suggest we have a live conversation and include Gulla from DMR there 🙂

darrikonn (approved):

apps/judicial-system/web/pages/api/nationalRegistry/getByNationalId/index.ts
eirikurn: Afaik Einstaklingur also supports lookup of a third party, as we're using it for that purpose already in a couple of places.

I understand there may be many practical reasons for the decision (requires new auth, doesn't include companies, maturity/reliability). Just sad since it feels like a step back in many ways. IMO the service needs more consumers to become more mature.

Then again, I'm not fully aware of the complete terms of use for the Einstaklingar service, given Þjóðskrá's general need to monetise access to their datasets. Maybe it's naive for me to assume this new non-billed API is available to government organisations (other than island.is) to look up third parties.

baldur-kolibri: Got it! I agree it is wise to stick to the policy to use Einstaklingur for all upcoming Island.is use cases and harden it in the coming months. 

Governance wise using Einstaklingur for RVG has the potential to generate blockers to progress even integrating the dev version, given the experience of other teams. ÞÍ has for some reason been wary of all non-standard use cases for Einstaklingur (different service owner, different uses) and used it as a reason to require specific GDPR processing agreements between parties and even started duplicating Einstaklingur for these uses to block the new consumer from using the standard API over X-road. 

Thus I think no one (not DMR and not Island.is) would benefit from the judicial system being used to test these boundaries. But I'm sure DMR will consider switching to Einstaklingur in order to save operating costs in due time, e.g. when larger sets of cases start coming in and the governance issues have (mostly) been settled.

PR #6476, opened 2022-02-03
feat(fa): personal tax return

Rikkiabb:
# ...
https://app.asana.com/0/1201722624788165/1201672092806096

## What

- Creates a new client which connect to personal tax return. 
- The client currently exposes one endpoint which returns personal tax return in pdf format.
- Adds types for xml2js.
- Adds data gathering section in financial aid application which tells user that we are fetching personal tax return.
- Sets up query and call to personal tax return client.
- Sets up service which calls data client and uploads file to s3 if it is successful.
- Fixes a bug which lead to that user could not download files on overview screen.
- Makes taxReturnForm screen dynamic depending on if call to personal tax return client was successful.

## Why

- Needed data for processing of application
- Client is set up for others to use as well
- Was not in the monorepo
- Better ux
- Needed
- We need to store the file
- Bug fixing is niceeeee
- We have a backup case telling the user to upload their tax return if the call to the service fails.

## Screenshots / Gifs

## Checklist:


apps/financial-aid/web-osk/src/routes/application/Info/info.tsx
```diff
@@ -117,6 +145,20 @@ const ApplicationInfo = () => {
...
+          Afrit af skattframtali og upplýsingar um staðgreiðslu í
+          staðgreiðsluskrá.
+        </Text>
+
+        <Text marginBottom={3}>
+          Við þurfum að afla þessara gagna til að staðfesta hjá hvaða
+          sveitarfélagi þú eigir að sækja um fjárhagsaðstoð, einfalda umsóknar-
+          og vinnsluferlið og staðfesta réttleika upplýsinga.
+        </Text>
+
```
JuliusGG, line 161: Should all these texts be hardcoded/translatable ?

apps/financial-aid/web-osk/src/routes/application/TaxReturnForm/taxReturnForm.tsx
```diff
@@ -13,6 +13,35 @@ import useFormNavigation from '@island.is/financial-aid-web/osk/src/utils/hooks/
...
+        staðfesting á öllum þeim tekjum, eignum og skuldum sem þú áttir á
+        skattárinu sem leið og er nauðsynlegt fylgigagn fyrir úrvinnslu á
+        fjárhagsaðstoð.
+      </Text>
+    ) : (
+      <Text marginBottom={[4, 4, 5]}>
+        Staðgreiðsluskrá er staðfesting/yfirlit frá Skattinum um skattskyldar
+        tekjur umsækjanda á árinu. Það er nauðsynlegt fylgigagn fyrir úrvinnslu
+        umsóknar um fjárhagsaðstoð.
+      </Text>
```
JuliusGG, line 40: Should these texts be translatable ?

JuliusGG (approved):
Nice work! - few places with open questions regarding hardcoded / translatable strings.

apps/financial-aid/web-osk/src/routes/application/TaxReturnForm/taxReturnForm.tsx
```diff
@@ -28,59 +57,70 @@ const TaxReturnForm = () => {
...
+                  </a>
+                ),
+              }}
+            >
+              <Text marginBottom={[3, 3, 5]}>
+                Á vef Skattsins finnur þú{' '}
+                <a href="https://www.skatturinn.is/einstaklingar/framtal-og-alagning/stadfest-afrit-framtals/">
+                  leiðbeiningar
+                </a>{' '}
+                um hvernig sækja má staðfest afrit skattframtals.
```
JuliusGG, line 119: same here :)

apps/financial-aid/web-osk/src/routes/application/TaxReturnForm/taxReturnForm.tsx
```diff
@@ -28,59 +57,70 @@ const TaxReturnForm = () => {
...
-        <Text marginBottom={2}>
-          Við þurfum að fá afrit af nýjasta <strong>skattframtali</strong> þínu
-          og staðfestingarskjal úr <strong>staðreiðsluskrá</strong> Skattsins.
-        </Text>
+        {taxReturnFetchFailed && (
+          <Box marginBottom={4}>
+            <AlertMessage
+              type="warning"
+              title="Ekki tókst að sækja skattframtal"
+              message="Þú getur hlaðið upp skattframtalinu þínu hérna ásamt staðgreiðsluskrá"
```
JuliusGG, line 77: and here :)

darrikonn (approved):
Ditto on Júlli's review about the Icelandic. Otherwise:
 LGTM

apps/financial-aid/web-osk/src/routes/application/Info/info.tsx
Rikkiabb: Yes so the client is not localized and won't be. It was suppose to be future work (I would have love to have it straight away). But now there is a project in progress about moving this application into the application system where we will set up the localization. 😄

apps/financial-aid/web-osk/src/components/Summary/AllFiles.tsx
```diff
@@ -31,7 +36,13 @@ const AllFiles = () => {
             if (file) {
               return (
                 <a
-                  href={file.name}
+                  onClick={() =>
+                    createSignedUrlMutation({
+                      variables: { input: { fileName: file.name } },
+                    }).then((response) => {
+                      window.open(response.data?.getSignedUrl.url, '_blank')
```
MargretFinnboga, line 43: window.open(response.data?.getSignedUrl.url, '_blank'), '_blank' is that not unneccary since this a tag has target="_blank"?

apps/financial-aid/web-osk/src/routes/application/TaxReturnForm/taxReturnForm.tsx
```diff
@@ -28,59 +57,70 @@ const TaxReturnForm = () => {
...
+                    target="_blank"
+                  >
+                    {children}
+                  </a>
+                ),
+              }}
+            >
+              <Text marginBottom={[3, 3, 5]}>
+                Á vef Skattsins finnur þú{' '}
+                <a href="https://www.skatturinn.is/einstaklingar/framtal-og-alagning/stadfest-afrit-framtals/">
```
MargretFinnboga, line 116: if I press this will it lead me there or open in another tab?

apps/financial-aid/web-osk/src/utils/hooks/useApplication.ts
```diff
@@ -41,6 +41,7 @@ const useApplication = () => {
         const files = formatFiles(form.taxReturnFiles, FileType.TAXRETURN)
           .concat(formatFiles(form.incomeFiles, FileType.INCOME))
           .concat(formatFiles(form.otherFiles, FileType.OTHER))
+          .concat(formatFiles(form.taxReturnFromRskFile, FileType.TAXRETURN))
```
MargretFinnboga, line 44: Do we want to differ from the tax that we get from user vs rsk?

libs/clients/rsk/personal-tax-return/src/lib/personalTaxReturn.config.ts
```diff
@@ -0,0 +1,26 @@
+import { defineConfig } from '@island.is/nest/config'
+import * as z from 'zod'
+
+const schema = z.object({
+  agentNationalId: z.string(),
```
MargretFinnboga, line 5: agent national id, is that the applicant?

MargretFinnboga (commented):
Few questions here and there 👀

apps/financial-aid/backend/src/app/modules/personalTaxReturn/personalTaxReturn.service.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    await fetch(presignedUrl.url, {
+      method: 'PUT',
+      body: Buffer.from(base64, 'binary'),
+      headers: {
+        'x-amz-acl': 'bucket-owner-full-control',
+        'Content-Type': 'application/pdf',
+        'Content-Length': size.toString(),
+      },
+    })
```
eirikurn, line 50: Not handling a non-ok response here. Should use AWS SDK's PutObject instead?

Also, I assume this is a pretty sensitive document we're storing in our S3. What's the data lifecycle like?

libs/clients/rsk/personal-tax-return/src/lib/requests/pdf.ts
```diff
@@ -0,0 +1,21 @@
...
+  return `<soap:Envelope xmlns:soap="http://www.w3.org/2003/05/soap-envelope" xmlns:tem="http://tempuri.org/" xmlns:ns="http://skattur.is/UmbodsmadurSkuldaraThjonusta/2011/01/03">
+  <soap:Header xmlns:wsa="http://www.w3.org/2005/08/addressing"><wsa:Action>http://tempuri.org/IUSStadgreidslaFramtalGogn/SaekjaPDFAfritFramtalsEinstaklings</wsa:Action><wsa:To>https://vefurp.rsk.is/ws/securep/UMS/WS/USStadgreidslaFramtalGogn.svc</wsa:To></soap:Header>
+  <soap:Body>
+     <tem:SaekjaPDFAfritFramtalsEinstaklings>
+        <!--Optional:-->
+        <tem:inntak>
+           <ns:KennitalaUmbodsmanns>${agentNationalId}</ns:KennitalaUmbodsmanns>
+           <ns:AudkenniUmbodsmanns>${agentId}</ns:AudkenniUmbodsmanns>
+           <ns:KennitalaSkuldara>${requesterNationalId}</ns:KennitalaSkuldara>
+           <ns:Tekjuar>${year}</ns:Tekjuar>
```
eirikurn, line 16: Should defend against XML injection. I much prefer setting up some kind of WSDL-to-TypeScript workflow to make soap integrations nicer.

No-one has done this before in the monorepo unfortunately, but we did set up something like that in the vidspyrna repo, so it's definitely doable.

apps/financial-aid/backend/src/app/modules/personalTaxReturn/personalTaxReturn.service.ts
```diff
@@ -0,0 +1,54 @@
...
+  private async callPersonalTaxReturn(nationalId: string, year: number) {
+    return await this.personalTaxReturnApi
+      .personalTaxReturnInPdf(nationalId, year.toString())
+      .catch(() => {
+        return { success: false, errorText: '', content: '' }
+      })
+  }
+
+  async personalTaxReturn(nationalId: string, folder: string) {
+    let year = new Date().getFullYear() - 1
```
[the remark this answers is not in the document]
Rikkiabb: This varibles mutate, so it could be last year og two years ago.  

The reason for this is the tax api doesn't fetch latest tax return but only by year. So if the call fails then we try the year before that if that fails then we allow the user to upload the tax return for themself.

apps/financial-aid/backend/src/app/modules/personalTaxReturn/personalTaxReturn.service.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    let taxReturn = await this.callPersonalTaxReturn(nationalId, year)
+
+    if (taxReturn.success === false) {
+      year -= 1
```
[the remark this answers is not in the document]
Rikkiabb: See response above ☝️  😄

apps/financial-aid/web-osk/src/utils/hooks/useApplication.ts
Rikkiabb: We have too, if we don't then the user will get to the tax form screen and that will look like he has already uploaded a file. And that screen has changes depending if this file was successfully fetched or not.

libs/clients/rsk/personal-tax-return/src/lib/personalTaxReturn.config.ts
Rikkiabb: Nopes it is bascially user and pass for the api

apps/financial-aid/web-osk/src/routes/application/TaxReturnForm/taxReturnForm.tsx
Rikkiabb: yes it does

libs/clients/rsk/personal-tax-return/src/lib/requests/pdf.ts
Rikkiabb: Implemented in a naive manner, would love feedback on it @eirikurn and WSDL-to-TS sounds nice, never seen, do you have an example which I could look at for curiosity?

apps/financial-aid/web-osk/src/routes/application/TaxReturnForm/taxReturnForm.tsx
```diff
@@ -13,6 +13,35 @@ import useFormNavigation from '@island.is/financial-aid-web/osk/src/utils/hooks/
 
 import { NavigationProps } from '@island.is/financial-aid/shared/lib'
 
+const getContent = (fetchFailed: boolean) => {
```
MargretFinnboga, line 16: why is this not in utils or formatters? 🤔

MargretFinnboga: And why have the text component along since there is no difference between them? why not just below? 
```
   <Text marginBottom={2}>
        {content.data}
</Text> 

   <Text marginBottom={[4, 4, 5]}>
        {content.reason}
</Text> 
```

apps/financial-aid/web-osk/src/routes/application/TaxReturnForm/taxReturnForm.tsx
```diff
@@ -28,59 +57,70 @@ const TaxReturnForm = () => {
...
 
-        <Text marginBottom={[4, 4, 5]}>
-          Við þurfum að fá afrit af nýjasta skattframtali þínu. Skattframtal er
-          staðfesting á öllum þeim tekjum, eignum og skuldum sem þú áttir á
-          skattárinu sem leið og er nauðsynlegt fylgigagn fyrir úrvinnslu á
-          fjárhagsaðstoð.
-        </Text>
+        {content.data}
+
+        {content.reason}
```
MargretFinnboga, line 84: Here, why not wrapped in component?

apps/financial-aid/web-osk/src/utils/hooks/useApplication.ts
MargretFinnboga: talk IRL

MargretFinnboga (commented):
Few more nitpicks 🤌

apps/financial-aid/web-osk/src/routes/application/TaxReturnForm/taxReturnForm.tsx
Rikkiabb: Very legit questions.

1. Because we are using strong tag in the return values and then we would have to wrap them in <></>, instead of having needless fragments I decided to use what we actually use in the layout.


2. I have been thinking about this and feel we have been little on the line of making things exported when they are just used in a single place, like this will be used nowhere else so I feel it is better to use just here and not export it.

Rikkiabb: Feel free do disagree 😄

apps/financial-aid/web-osk/src/routes/application/TaxReturnForm/taxReturnForm.tsx
Rikkiabb: Answered above ☝️

MargretFinnboga (approved):
🚢  it!

libs/clients/rsk/personal-tax-return/src/lib/requests/pdf.ts
eirikurn: My concern was mostly to escape the input, eg using `_.escape()`. Sanitization gets the job done though. Thankfully there is also no user input in this case, but always good to escape. Data validation can occur in other places (ideally as late as possible).

As I said, there is an example in [Vidspyrna here](https://github.com/island-is/vidspyrna/blob/de193ce0dac722a818596da595d20dabd9ab247d/apps/api/src/services/rsk/soap/types.ts). Haven't looked more into it though.

eirikurn (approved):

PR #6481, opened 2022-02-03
feat(driving-instructor-registrations): Driving Instructor's Registrations

albinagu:
New Application Template for Driving Instructor's Registrations (Skráningar ökukennara) VOL1

## Checklist:


libs/application/templates/driving-instructor-registrations/src/forms/instructorRegistrations.ts
```diff
@@ -0,0 +1,65 @@
...
+        id: 'externalData',
+        title: m.dataCollectionTitle,
+        children: [
+          buildExternalDataProvider({
+            id: 'approveExternalData',
+            title: m.dataCollectionTitle,
+            subTitle: m.dataCollectionSubtitle,
+            checkboxLabel: m.dataCollectionCheckboxLabel,
+            dataProviders: [
+              /*buildDataProviderItem({
```
arnorhs, line 30: this should probably not be commented out, right?

libs/application/templates/driving-instructor-registrations/src/lib/InstructorRegistrationsTemplate.ts
```diff
@@ -0,0 +1,75 @@
...
+
+const InstructorRegistrationsTemplate: ApplicationTemplate<
+  ApplicationContext,
+  ApplicationStateSchema<Events>,
+  Events
+> = {
+  type: ApplicationTypes.DRIVING_INSTRUCTOR_REGISTRATIONS,
+  name: 'Skráningar Ökukennara',
+  dataSchema: dataSchema,
+  readyForProduction: true,
```
arnorhs, line 23: this should probably be false?

libs/application/templates/driving-instructor-registrations/src/lib/dataSchema.ts
```diff
@@ -0,0 +1,27 @@
...
+const isValidEmail = (value: string) => emailRegex.test(value)
+
+const isValidPhoneNumber = (phoneNumber: string) => {
+  const phone = parsePhoneNumberFromString(phoneNumber, 'IS')
+  return phone && phone.isValid()
+}
+
+export const dataSchema = z.object({
+  approveExternalData: z.boolean().refine((v) => v),
+  address: z.string(),
```
arnorhs, line 15: are all these fields used?

libs/application/templates/driving-instructor-registrations/src/fields/mock.ts
```diff
@@ -0,0 +1,105 @@
+export const STUDENTS = [
+  {
```
arnorhs, line 2: same re: including actual users in mock data - and ideally not having any mock data in running code paths (beyond tests)

arnorhs (changes requested):
Looks like this is coming along well, a few comments re: application and mock data etc.

Separately, I wonder if this could be split into two PRs.. one for the client/service/api part and another for the actual template

libs/application/templates/driving-instructor-registrations/src/forms/instructorRegistrations.ts
albinagu: ó yea

libs/application/templates/driving-instructor-registrations/src/fields/FindStudentModal/index.tsx
```diff
@@ -67,8 +98,10 @@ const FindStudentModal = ({
                 name="search_student"
                 backgroundColor="blue"
                 size="sm"
+                hasError={studentNotFoundError}
+                errorMessage="Enginn nemandi skráður á eftirfarandi kennitölu"
```
Thorrik58, line 102: Shouldn't this be defined as a translation string?

libs/application/templates/driving-instructor-registrations/src/fields/StudentsOverview/index.tsx
```diff
@@ -0,0 +1,182 @@
...
+                </T.HeadData>
+                <T.HeadData box={{ textAlign: 'center' }}>
+                  {formatMessage(m.studentsOverviewTableHeaderCol3)}
+                </T.HeadData>
+                <T.HeadData></T.HeadData>
+              </T.Row>
+            </T.Head>
+            <T.Body>
+              {pageStudents && pageStudents.length ? (
+                pageStudents.map((student, key) => {
```
Thorrik58, line 118: We shouldn't use index as a key

libs/application/templates/driving-instructor-registrations/src/fields/StudentsOverview/index.tsx
```diff
@@ -0,0 +1,182 @@
...
+                    {formatMessage(m.studentsOverviewNoStudentFound)}
+                  </T.Data>
+                  <T.Data></T.Data>
+                  <T.Data></T.Data>
+                  <T.Data></T.Data>
+                </T.Row>
+              )}
+            </T.Body>
+          </T.Table>
+          {pageStudents && !!pageStudents.length && (
```
Thorrik58, line 153: Is there a reason we are using `!!pageStudents.length` instead of `pagesStudents.length > 0?`

libs/application/templates/driving-instructor-registrations/src/fields/StudentsOverview/pagination.ts
```diff
@@ -0,0 +1,15 @@
+export const PAGE_SIZE = 10
+
+export function paginate(students: any, pageSize: number, pageNumber: number) {
+  if (students === undefined) return
```
Thorrik58, line 4: What happens if student is null?

libs/application/templates/driving-instructor-registrations/src/fields/StudentsOverview/pagination.ts
```diff
@@ -0,0 +1,15 @@
+export const PAGE_SIZE = 10
+
+export function paginate(students: any, pageSize: number, pageNumber: number) {
```
Thorrik58, line 3: Would be better if we could type students. You can use union types to define object or undefined.

libs/application/templates/driving-instructor-registrations/src/fields/StudentsOverview/pagination.ts
```diff
@@ -0,0 +1,15 @@
...
+  if (students === undefined) return
+  else {
+    return students.slice((pageNumber - 1) * pageSize, pageNumber * pageSize)
+  }
+}
+
+export function pages(studentsLength: number) {
+  if (studentsLength === undefined) return 0
```
Thorrik58, line 11: What happens if studentsLength is null?

libs/api/domains/driving-license-book/src/dto/activeBookId.input.ts
```diff
@@ -0,0 +1,13 @@
+import { IsString } from 'class-validator'
+import { Field, InputType } from '@nestjs/graphql'
+
+@InputType()
+export class ActiveBookIdInput {
+  @Field({ nullable: true })
+  @IsString()
+  ssn!: string
```
darrikonn, line 8: we use nationalId in island.is ... btw. the class-validators don't do anything here... can be removed.

libs/api/domains/driving-license-book/src/dto/createPracticalDrivingLesson.input.ts
```diff
@@ -0,0 +1,32 @@
+import { IsString, IsNumber, IsOptional } from 'class-validator'
```
darrikonn, line 1: Ditton on class-validator ... not needed for GraphQL unless it's doing some complex validation, like isEmail.

libs/api/domains/driving-license-book/src/dto/createPracticalDrivingLesson.input.ts
```diff
@@ -0,0 +1,32 @@
...
+
+@InputType()
+export class CreatePracticalDrivingLesson {
+  @Field()
+  @IsString()
+  bookId!: string
+
+  @Field()
+  @IsString()
+  teacherSsn!: string
```
darrikonn, line 12: ditto on nationalId

libs/application/templates/driving-instructor-registrations/src/forms/instructorRegistrations.ts
```diff
@@ -0,0 +1,65 @@
...
+          }),
+        ],
+      }),
+      buildSection({
+        id: 'students',
+        title: 'Skráningar',
+        children: [
+          buildMultiField({
+            id: 'students',
+            title: 'Mínir ökunemar',
```
Thorrik58, line 52: Should this be defined as a translation string?

libs/application/templates/driving-instructor-registrations/src/forms/instructorRegistrations.ts
```diff
@@ -0,0 +1,65 @@
...
+                title: '',
+                subTitle: '',
+              }),
+            ],
+          }),
+        ],
+      }),
+      buildSection({
+        id: 'students',
+        title: 'Skráningar',
```
Thorrik58, line 48: Should this be defined as a translation string?

libs/application/templates/driving-instructor-registrations/src/lib/InstructorRegistrationsTemplate.ts
```diff
@@ -0,0 +1,75 @@
...
+                  Promise.resolve(val.getInstructorRegistrations()),
+                ),
+              actions: [
+                {
+                  event: DefaultEvents.SUBMIT,
+                  name: 'Staðfesta',
+                  type: 'primary',
+                },
+              ],
+              write: 'all',
```
Thorrik58, line 58: Something to consider before the application is released out to the public, we need to make sure the 3rd party can only read and write the data that that role needs to work with but not everything. (all)

libs/application/templates/driving-instructor-registrations/src/shared/constants.ts
```diff
@@ -0,0 +1,24 @@
+export enum ApiActions {
+  createApplication = 'createApplication',
+  doStuffThatFails = 'doStuffThatFails',
+  completeApplication = 'completeApplication',
+}
```
Thorrik58, line 5: Is this used anywhere?

libs/api/domains/driving-license-book/src/dto/student.input.ts
```diff
@@ -0,0 +1,9 @@
+import { IsString } from 'class-validator'
+import { Field, InputType } from '@nestjs/graphql'
+
+@InputType()
+export class StudentInput {
+  @Field({ nullable: true })
+  @IsString()
+  ssn!: string
```
darrikonn, line 8: ditto

libs/api/domains/driving-license-book/src/models/activeBookId.response.ts
```diff
@@ -0,0 +1,12 @@
+import { Field, ObjectType } from '@nestjs/graphql'
+@ObjectType()
+export class BookId {
+  @Field({ nullable: true })
+  bookId?: string
```
darrikonn, line 5: This should probably be marked as a GraphQL ID

libs/api/domains/driving-license-book/src/models/createPracticalDrivingLicense.response.ts
```diff
@@ -0,0 +1,13 @@
...
+@ObjectType()
+export class PracticalDrivingLessonId {
+  @Field({ nullable: true })
+  id?: string
+}
+
+@ObjectType()
+export class CreatePracticalDrivingLessonResponse {
+  @Field(() => PracticalDrivingLessonId, { nullable: true })
+  data?: PracticalDrivingLessonId
```
darrikonn, line 12: The data object structure is not following GraphQL conventions ... this way you probably have to map data twice in the front-end... rather return PracticalDrivingLesson

libs/api/domains/driving-license-book/src/models/activeBookId.response.ts
```diff
@@ -0,0 +1,12 @@
+import { Field, ObjectType } from '@nestjs/graphql'
+@ObjectType()
+export class BookId {
```
darrikonn, line 3: This goes for other GraphQL fields as well ... the name of an object type (field) should rather be Book, instead of BookId ... then you fetch the id of the book with book.id

libs/api/domains/driving-license-book/src/models/drivingLicenseTestResult.resopnse.ts
```diff
@@ -0,0 +1,46 @@
+import { Field, ObjectType } from '@nestjs/graphql'
+
+@ObjectType()
+export class DrivingLicenceTestResult {
```
darrikonn, line 4: Skip suffix

darrikonn (changes requested):
Gonna stop here, since most of my comments are the same.
The GraphQL schema needs a cleanup to follow conventions.

Thorrik58 (changes requested):
Some comments

libs/api/domains/driving-license-book/src/drivingLicenseBook.service.ts
```diff
@@ -0,0 +1,124 @@
...
+@Injectable()
+export class DrivingLicenseBookService {
+  constructor(
+    @Inject(LOGGER_PROVIDER) private readonly logger: Logger,
+    @Inject(DrivingLicenseBookClientService)
+    private drivingLicenseBookClientService: DrivingLicenseBookClientService,
+  ) {}
+
+  async apiWithAuth() {
+    return await this.drivingLicenseBookClientService.api()
```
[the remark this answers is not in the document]
alexandradilja: The function` this.drivingLicenseBookClientService.api()`  is calling `createsEnhancedFetch` for the `drivingLicenseBookApi` and calls the async `authenticate` in the `drivingLicenseBookApi` and sets the `auth` provided into the header with `new AuthMiddleware.` Do you know of any examples that I could look at where this login functionality exists in a constructor?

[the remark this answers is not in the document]
darrikonn: mm I see. No this is fine, didn't know the client needed the auth token for each request. Can be ignored.

libs/api/domains/driving-license-book/src/drivingLicenseBook.resolver.ts
```diff
@@ -0,0 +1,71 @@
...
+import { Student } from './models/studentsTeacherNationalId.response'
+
+@UseGuards(IdsUserGuard, ScopesGuard)
+@Resolver()
+export class DrivinLicenseBookResolver {
+  constructor(
+    private readonly drivingLicenseBookService: DrivingLicenseBookService,
+  ) {}
+
+  @Query(() => [DrivingBookStudent], { nullable: true })
```
darrikonn, line 29: I don't think that lists should be nullable ... rather return an empty list.

libs/api/domains/driving-license-book/src/drivingLicenseBook.resolver.ts
```diff
@@ -0,0 +1,71 @@
...
+  drivingBookStudentList(@Args('input') input: StudentListInput) {
```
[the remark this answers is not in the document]
darrikonn: Should this query be open for all?
This is open so anyone with a valid auth can fetch every driving license student, regardless of permission.

libs/api/domains/driving-license-book/src/drivingLicenseBook.resolver.ts
```diff
@@ -0,0 +1,71 @@
...
+    return this.drivingLicenseBookService.getStudentList(input)
+  }
+
+  @Query(() => [Student])
+  drivingBookStudentListByTeacherNationalId(@CurrentUser() user: User) {
```
darrikonn, line 35: This comment depends on if the query above should be open or not.
`ByTeacherNationalId` this should most likely be a part of the input ... or if the previous query gets removed.. than this name should just be:
```typescript
drivingLicenseBookStudents
```

darrikonn: Or actually .. should be named `drivingLicenseBookStudents` ... but the input should contain a boolean specifying something like `fetchMyStudents: true`

darrikonn (changes requested):
Missing guards

libs/api/domains/driving-license-book/src/models/drivingLicenseTestResult.resopnse.ts
alexandradilja: This result is referring to test results not result from the query 😄

libs/api/domains/driving-license-book/src/drivingLicenseBook.resolver.ts
```diff
@@ -0,0 +1,71 @@
...
+  ) {
+    return this.drivingLicenseBookService.updatePracticalDrivingLesson(input)
+  }
+
+  @Mutation(() => SuccessResponse)
+  drivingBookDeletePracticalDrivingLesson(
+    @Args('input') input: DeletePracticalDrivingLessonInput,
+  ) {
+    return this.drivingLicenseBookService.deletePracticalDrivingLesson(input)
+  }
```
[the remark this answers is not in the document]
alexandradilja: adding checks on update and delete to see if user is marked as teacher on that lesson. And looking into the guards.

apps/api/infra/api.ts
```diff
@@ -155,6 +156,10 @@ export const serviceSetup = (services: {
       ISLYKILL_SERVICE_PASSPHRASE: '/k8s/api/ISLYKILL_SERVICE_PASSPHRASE',
       ISLYKILL_SERVICE_BASEPATH: '/k8s/api/ISLYKILL_SERVICE_BASEPATH',
       IDENTITY_SERVER_CLIENT_SECRET: '/k8s/api/IDENTITY_SERVER_CLIENT_SECRET',
+      DRIVING_LICNESE_BOOK_XROAD_PATH:
+        '/k8s/api/DRIVING_LICNESE_BOOK_XROAD_PATH',
+      DRIVING_LICNESE_BOOK_USERNAME: '/k8s/api/DRIVING_LICNESE_BOOK_USERNAME',
+      DRIVING_LICNESE_BOOK_PASSWORD: '/k8s/api/DRIVING_LICNESE_BOOK_PASSWORD',
```
pshomov, line 162: We should go for an X-Road "module" like the ones below in the `.xroad` section.

pshomov (changes requested):
See comment please

libs/api/domains/driving-license-book/src/models/drivingLicenseTestResult.resopnse.ts
darrikonn: aaah I see 👍

charts/islandis/values.dev.yaml
```diff
@@ -345,6 +346,9 @@ api:
     SYSLUMENN_HOST: '/k8s/api/SYSLUMENN_HOST'
     SYSLUMENN_PASSWORD: '/k8s/api/SYSLUMENN_PASSWORD'
     SYSLUMENN_USERNAME: '/k8s/api/SYSLUMENN_USERNAME'
+    DRIVING_LICNESE_BOOK_XROAD_PATH: '/k8s/api/DRIVING_LICNESE_BOOK_XROAD_PATH',
+    DRIVING_LICNESE_BOOK_USERNAME: '/k8s/api/DRIVING_LICNESE_BOOK_USERNAME',
+    DRIVING_LICNESE_BOOK_PASSWORD: '/k8s/api/DRIVING_LICNESE_BOOK_PASSWORD',
```
darrikonn, line 351: there's a typo in the name btw. :) ... licnese vs. license

libs/api/domains/driving-license-book/src/models/studentsTeacherNationalId.response.ts
```diff
@@ -0,0 +1,16 @@
+import { Field, ID, ObjectType } from '@nestjs/graphql'
+
+@ObjectType()
+export class DrivingLicenseBookStudentForTeacher {
+  @Field(() => ID, { nullable: true })
```
darrikonn, line 5: I might be wrong here ... but this comment goes for the other GraphQL types as well:
Doesn't the schema throw an error if you try to set an ID as nullable?
I assumed this wasn't possible 🤔

libs/application/templates/driving-instructor-registrations/src/fields/StudentsOverview/index.tsx
```diff
@@ -0,0 +1,187 @@
...
+      )
+
+      handlePagination(1, filteredList)
+    } else {
+      handlePagination(1, data?.drivingLicenseBookStudentsForTeacher)
+    }
+  }
+
+  useEffect(() => {
+    filter(searchTerm)
```
darrikonn, line 72: Shouldn't there be a delay on this?
Otherwise there will be x queries fired for a string of length x

libs/clients/driving-license-book/src/lib/drivingLicenseBookClient.module.ts
```diff
@@ -0,0 +1,10 @@
+import { Module } from '@nestjs/common'
+// import { ApiConfiguration } from './apiConfiguration'
+// import { exportedApis } from './apis'
```
saevarma, line 3: Remove redundant comments

libs/clients/driving-license-book/src/lib/drivingLicenseBookClient.service.ts
```diff
@@ -0,0 +1,47 @@
...
+
+import { createEnhancedFetch } from '@island.is/clients/middlewares'
+import { XRoadConfig } from '@island.is/nest/config'
+import type { ConfigType } from '@island.is/nest/config'
+import { DrivingLicenseBookClientConfig } from './drivingLicenseBookClient.config'
+import { Injectable, Inject } from '@nestjs/common'
+import { AuthHeaderMiddleware } from '@island.is/auth-nest-tools'
+
+@Injectable()
+export class DrivingLicenseBookClientService {
```
saevarma, line 11: I would consider to rename this class to `DrivingLicenseBookClientFactory` as it is more of a factory creating a new instance of the api with auth setup. The `api()` could then be renamed to `create()`

saevarma (changes requested):

apps/api/infra/api.ts
kristofer-lp8f: Done. Can you confirm?

charts/islandis/values.dev.yaml
kristofer-lp8f: Fixed.

libs/api/domains/driving-license-book/src/models/studentsTeacherNationalId.response.ts
kristofer-lp8f: I don't think there is anything forbidding nullable IDs. I believe any GraphQL field can be null. The schema doesn't seem to enforce non-null on ID, and in fact is omitted for most classes.

libs/application/templates/driving-instructor-registrations/src/fields/StudentsOverview/index.tsx
kristofer-lp8f: If we add delay, would the user experience be worse?

kristofer-lp8f: Nvm, changed to debounce with 500 ms delay. How do you like https://github.com/island-is/island.is/pull/6481/commits/38464708d822da50c21bb53070780a8005d9f6f3?

apps/api/src/app/app.module.ts
```diff
@@ -93,6 +95,17 @@ const autoSchemaFile = environment.production
...
+    // DrivingLicenseBookModule needs to register DrivingLicenseModule and uses the same config to do so
+    DrivingLicenseBookModule.register({
+      clientConfig: {
+        xroadBaseUrl: environment.xroad.baseUrl!,
+        xroadClientId: environment.xroad.clientId!,
+        secret: environment.drivingLicense.secret!,
+        xroadPathV1: environment.drivingLicense.v1.xroadPath!,
+        xroadPathV2: environment.drivingLicense.v2.xroadPath!,
+      },
+    }),
```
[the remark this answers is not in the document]
kristofer-lp8f: We are under some pressure to get this through and get going with `DrivingTeacherBooking`. Do you mind me adding the refactoring to our internal sprint backlog and deal with it for the next release?

libs/clients/driving-license-book/src/lib/drivingLicenseBookClient.module.ts
kristofer-lp8f: Done.

libs/clients/driving-license-book/src/lib/drivingLicenseBookClient.service.ts
kristofer-lp8f: Done.

libs/api/domains/driving-license-book/src/models/studentsTeacherNationalId.response.ts
kristofer-lp8f: How would I go about testing your hypothesis?

darrikonn: So basically my assumption was that the graphql eslint would notify you that ID's should not be null. But you generally never want to set IDs as nullable, otherwise the in memory cache normalization won't work on the front-end. So you'd rather want the error to propagate.
There are really unique cases when you want the ID to be nullable.

If you want to test it, just try and strip out the ID from the returned result. The cache won't be updated correctly since it won't know how to normalize the returned data -> this is the same reason for why we should always query the GraphQL field's ID

libs/api/domains/driving-license-book/src/dto/createPracticalDrivingLesson.input.ts
```diff
@@ -0,0 +1,19 @@
+import { Field, InputType } from '@nestjs/graphql'
+
+@InputType()
+export class CreatePracticalDrivingLessonInput {
+  @Field()
+  bookId!: string
+
+  @Field()
+  teacherNationalId!: string
```
darrikonn, line 9: Should driving license teachers be able to create driving licenses for other teachers?

If not, then this input field should not be here, and it should be using the currently logged in user's national id (which will be a driving license teacher)

libs/api/domains/driving-license-book/src/models/success.response.ts
```diff
@@ -0,0 +1,7 @@
+import { Field, ObjectType } from '@nestjs/graphql'
+
+@ObjectType()
+export class Success {
```
darrikonn, line 4: The name of this field will clash with other fields in the schema.

libs/application/templates/driving-instructor-registrations/src/fields/StudentsOverview/index.tsx
```diff
@@ -0,0 +1,193 @@
...
+
+  const handlePagination = (page: number, students: Array<object>) => {
+    setPage(page)
+    setTotalPages(pages(students?.length))
+    setFilteredStudents(students)
+    setPageStudents(paginate(students, PAGE_SIZE, page))
+  }
+
+  const filter = (searchTerm: string) => {
+  console.log("########## Filter search ###############")
```
darrikonn, line 59: console.

libs/api/domains/driving-license-book/src/models/success.response.ts
kristofer-lp8f: Renamed it to `DrivingLicenceBookSuccess`. Thanks for the catch.

libs/application/templates/driving-instructor-registrations/src/fields/StudentsOverview/index.tsx
kristofer-lp8f: Removed.

libs/api/domains/driving-license-book/src/dto/createPracticalDrivingLesson.input.ts
kristofer-lp8f: Fixed.

libs/application/templates/driving-instructor-registrations/src/fields/StudentsOverview/pagination.ts
kristofer-lp8f: Added to backlog :)

pshomov (approved):
LGTM

apps/api/src/app/app.module.ts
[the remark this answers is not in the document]
alexandradilja: A simplification of the DrivingLicenseModule is a good idea, we will implement it in a follow up PR, since it effects the Driving License application more and would be nice to do it in a clean PR.

darrikonn (approved):

Thorrik58 (approved):
LGTM for all files related to the application system team.

apps/api/src/app/app.module.ts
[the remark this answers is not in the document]
saevarma: Agreed.

saevarma (approved):

PR #6482, opened 2022-02-03
feat(enhancedfetch): Adding a way for us to send custom metrics to DD

dabbeg:
Closes: https://github.com/island-is/infrastructure/issues/1000

[Link to metrics in DD](https://app.datadoghq.eu/metric/explorer?from_ts=1643961964561&to_ts=1643965564561&live=false&tile_size=m&exp_metric=islandis.FJS_financeIsland.requests%2Cislandis.clients_assets.requests%2Cislandis.ApplicationModule.applicationsApi.requests%2Cislandis.clients_auth_public_api.requests%2Cislandis.clients_national_registry_v2.requests&exp_agg=avg&exp_row_type=metric)

## What

Add custom metrics to islandis. As an example I added a simple counter in enhanced fetch.

## Why

So we can stop using APM metrics as they are sample metrics but not reliable for counters.

## Checklist:


libs/clients/middlewares/src/lib/withMetrics.ts
```diff
@@ -0,0 +1,21 @@
...
+export const withMetrics = ({
+  fetch,
+  name,
+}: MetricsMiddlewareOptions): FetchAPI => {
+  const client = dogStatsD({
+    prefix: `${name}.`,
+  })
+  return (input, init) => {
+    client.increment('requests')
+    return fetch(input, init)
```
eirikurn, line 19: Would be nice to add a couple more metrics while we're at it:

* ${name}.requests.ok
* ${name}.requests.fail (either separate metrics or tag: `reason=4xx|5xx|network`)

Maybe also a responseTime metric?

The only tricky thing here is that the same API might have quite different endpoints which in some cases we'd want to count separately. If that happens we could expand the Enhanced Fetch API with a callback that customizes the metric keys/tags using request details.

eirikurn (commented):
Love it! Would be nice to expand this a bit we're at it, but your call.

libs/clients/middlewares/src/lib/withMetrics.ts
pshomov: I would suggest distinguishing between 4xx and 5xx errors, so have
${name}.requests.ok - {100-399}
${name}.requests.error - {400-499}
${name}.requests.fail - {500 - 599}

pshomov:
Great integration with our extended fetch api. Does it make sense to provide more examples in different parts of the code -NextJS page rendering, application submitted, etc.? Not sure it makes sense to be part of this PR (with this title anyways) but maybe a continuation on this work.

pshomov (approved):

libs/clients/middlewares/src/lib/withMetrics.ts
dabbeg: I added more descriptive error metrics, they are little bit different than suggested. I followed the pattern I was seeing on existing metrics in datadog.

The responseTime metric does not seem to be trivial and the other thing can be an improvement later on.

Just to be clear this PR is a simple example on how to add metric collecting so we have an example to follow.

dabbeg:
> Great integration with our extended fetch api. Does it make sense to provide more examples in different parts of the code -NextJS page rendering, application submitted, etc.? Not sure it makes sense to be part of this PR (with this title anyways) but maybe a continuation on this work.

I would prefer getting another issue with specific metrics we want to collect.

libs/clients/middlewares/src/lib/withMetrics.ts
```diff
@@ -0,0 +1,40 @@
...
+      } else if (response.status >= 400 && response.status < 500) {
+        client.increment('requests.4xx')
+        client.increment('requests.error')
+      } else if (response.status >= 500 && response.status < 600) {
+        client.increment('requests.5xx')
+        client.increment('requests.error')
+      }
+      return response
+    } catch (error) {
+      client.increment('requests.client_error')
```
eirikurn, line 36: Let's also increment `requests.error` here. These errors almost always occur because of some network issue, so maybe:

```suggestion
      client.increment('requests.network_error')
      client.increment('requests.error')
```

libs/clients/middlewares/src/lib/withMetrics.ts
```diff
@@ -0,0 +1,40 @@
...
+
+export interface MetricsMiddlewareOptions {
+  fetch: FetchAPI
+  name: string
+}
+
+export const withMetrics = ({
+  fetch,
+  name,
+}: MetricsMiddlewareOptions): FetchAPI => {
```
eirikurn, line 13: Sorry, one more thing for your "simple example PR". :D 

It would be reaaaally good to add tests for this. At least that 500 responses trigger the correct increments. Depending on the ordering of the middlewares, 500 responses are returned normally or thrown as an error.

I think it's not an issue as currently ordered, but this could silently break if someone reorders middlewares in the future, unless we have a test for it.

eirikurn (changes requested):
Awesome! Agree with finishing this integration and leaving more metrics for future issues/PRs.

libs/clients/middlewares/src/lib/withMetrics.ts
dabbeg: I added tests, can you review again?

eirikurn (approved):
Beautiful! :shipit:

PR #6484, opened 2022-02-03
fix(user-profile): Only create islykill settings if email OR tel is in input

thordurhhh:
## What

Only create islykill settings if `email` OR `mobile` is in input

## Why

IF islykill fails when creating a user with only `locale` or `documentNotifications` an error will be thrown. And that error will not have any relevance for a user with only `locale` or `documentNotifications` in the input

## Checklist:


disaerna (approved):
Nice catch 👍

PR #6490, opened 2022-02-03
feat(air-discount-scheme): loftbru-roll-forward

JuliusGG:
# ...

## What

## Why

## Screenshots / Gifs

## Checklist:

### DIFF
--- a/apps/air-discount-scheme/backend/src/app/modules/user/user.controller.ts
+++ b/apps/air-discount-scheme/backend/src/app/modules/user/user.controller.ts
@@ -79,10 +79,11 @@ export class PrivateUserController {
         '[/relations] Request parameters do not correspond with user authentication.',
       )
     }
-    const relations = await this.userService.getRelations(authUser)
+    const relations = [
+      authUser.nationalId,
+      ...(await this.userService.getRelations(authUser)),
+    ]
 
-    // Adding User in beginning of array so order is correct.
-    relations.unshift(authUser.nationalId)
     const userAndRelatives = await this.userService.getMultipleUsersByNationalIdArray(
       relations,
     )
--- a/apps/air-discount-scheme/backend/src/app/modules/user/user.controller.ts
+++ b/apps/air-discount-scheme/backend/src/app/modules/user/user.controller.ts
@@ -79,13 +79,22 @@ export class PrivateUserController {
         '[/relations] Request parameters do not correspond with user authentication.',
       )
     }
-    const relations = await this.userService.getRelations(params.nationalId)
-    const users = (await Promise.all([
-      this.userService.getUserInfoByNationalId(params.nationalId),
-      ...relations.map((nationalId) =>
-        this.userService.getUserInfoByNationalId(nationalId),
-      ),
-    ])) as User[]
-    return users.filter((user) => user) as User[]
+    const relations = await this.userService.getRelations(authUser)
+
+    // Adding User in beginning of array so order is correct.
+    relations.unshift(authUser.nationalId)
+    const userAndRelatives = await this.userService.getMultipleUsersByNationalIdArray(
+      relations,
+    )
+
+    const returnUsers = userAndRelatives.filter(
+      (user) => user !== null,
+    ) as Array<User>
+    if (returnUsers.length === 0) {
+      throw new Error(
+        'Could not find NationalRegistry records of neither User or relatives.',
+      )
+    }
+    return returnUsers
   }
 }
--- a/apps/air-discount-scheme/backend/src/app/modules/user/user.controller.ts
+++ b/apps/air-discount-scheme/backend/src/app/modules/user/user.controller.ts
@@ -88,14 +88,11 @@ export class PrivateUserController {
       relations,
     )
 
-    const returnUsers = userAndRelatives.filter(
-      (user) => user !== null,
-    ) as Array<User>
-    if (returnUsers.length === 0) {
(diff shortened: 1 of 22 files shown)


Snojo (approved):

apps/air-discount-scheme/backend/src/app/modules/user/user.service.ts
```diff
@@ -60,4 +90,17 @@ export class UserService {
   async getUserInfoByNationalId(nationalId: string): Promise<User | null> {
     return this.getUserByNationalId<User>(nationalId, User)
   }
+
+  async getMultipleUsersByNationalIdArray(
+    ids: string[],
+  ): Promise<Array<User | null>> {
+    const allUsers = ids.map(
+      async (nationalId) =>
+        (this.getUserInfoByNationalId(nationalId) as unknown) as User,
```
arnorhs, line 99: it would be nice to get rid of the the `as` -- or simply filtering out nulls, since it looks like they can return null

so if you change the thing below to
```
return (await Promise.all(allUsers)).filter(Boolean)
```

you should get all the correct types without type cooercion (and it will handle these edge cases somewhat gracefully)

arnorhs (changes requested):

apps/air-discount-scheme/backend/src/app/modules/user/user.service.ts
arnorhs: also, you could make the return type `Array<User>` instead of having them possibly be null

apps/air-discount-scheme/backend/src/app/modules/user/user.controller.ts
```diff
@@ -79,13 +79,22 @@ export class PrivateUserController {
...
     }
-    const relations = await this.userService.getRelations(params.nationalId)
-    const users = (await Promise.all([
-      this.userService.getUserInfoByNationalId(params.nationalId),
-      ...relations.map((nationalId) =>
-        this.userService.getUserInfoByNationalId(nationalId),
-      ),
-    ])) as User[]
-    return users.filter((user) => user) as User[]
+    const relations = await this.userService.getRelations(authUser)
```
arnorhs, line 82: personally i think it would be a cleaner approach to do something like:
```
const relations = [
  ...(await this.userService.getRelations(authUser)),
  authUser.nationalId
]
```
then you wouldn't have to modify the array, and i think it communicates the intent a bit better

apps/air-discount-scheme/backend/src/app/modules/user/user.controller.ts
```diff
@@ -79,13 +79,22 @@ export class PrivateUserController {
...
+
+    // Adding User in beginning of array so order is correct.
+    relations.unshift(authUser.nationalId)
+    const userAndRelatives = await this.userService.getMultipleUsersByNationalIdArray(
+      relations,
+    )
+
+    const returnUsers = userAndRelatives.filter(
```
arnorhs, line 90: hopefully with the other suggested changes, you don't have to perform this filter statement, and type coercion would be necessary

arnorhs (changes requested):
It's going!

One suggestion, not only for this pull request, but in general is to really try to avoid `as` as much as possible, unless you are dealing with a truly un-inferrable type.

typescript is pretty precise, and when you do an `as` you are really opting out of a lot of the safety that the types get you.

apps/air-discount-scheme/backend/src/app/modules/user/user.controller.ts
JuliusGG: I like it

apps/air-discount-scheme/backend/src/app/modules/user/user.controller.ts
JuliusGG: Since getUserInfoByNationalId can return either User or null - depending on the data of the server(not a problem in production 99.9999999999% of the time) then I believe the filter statement will be required for dev/staging and the most extreme edgecase.
This filter statement could be moved into getMultipleUserByNationalArray but I feel it is more appropriate that the controller handles what is legit and what isn't.
Feel free to voice your opinion on that

apps/air-discount-scheme/backend/src/app/modules/user/user.service.ts
JuliusGG: Just saw this now - Needed to refresh the page.

Snojo (approved):

apps/air-discount-scheme/backend/src/app/modules/user/user.controller.ts
arnorhs: not really an opinion, but like i said.. moving the filter statement to the other method would fix that

JuliusGG: I fixed it - earlier I missed two other comments.

Snojo (dismissed):

Snojo (approved):

arnorhs (approved):

eirikurn (approved):
Good discussions all around 👍

eirikurn (approved):

eirikurn:
@pshomov don't forget to apply this to the main branch.

arnorhs (approved):

pshomov:
> @pshomov don't forget to apply this to the main branch.

Yes, already did that #6503

Snojo (approved):

PR #6492, opened 2022-02-04
Gfl/payment

sigruntg:
# Payment

[1pkmetm](https://app.clickup.com/t/1pkmetm)

## What

Added payment into the project, we still do not have items added from FJS, so for now I just added items coming from Sýslumaður to test.

## Checklist:


libs/application/templates/general-fishing-license/src/fields/Declined/WarningSection.css.ts
```diff
@@ -0,0 +1,52 @@
...
+})
+
+export const sectionNumberRequiresAction = style({
+  background: theme.color.red600,
+  color: theme.color.red600,
+})
+
+export const sectionNumberComplete = style({
+  background: theme.color.mint600,
+})
```
darrikonn, line 36: Most of these can be created using the Box component, and thus this isn't needed.

libs/application/templates/general-fishing-license/src/graphql/queries.ts
```diff
@@ -37,3 +37,11 @@ export const queryFishingLicense = gql`
     }
   }
 `
+
+export const PAYMENT_STATUS = gql`
+  query status($applicationId: String!) {
```
darrikonn, line 42: The name of the query is too generic and might override existing ones.

Additionally, always query the GraphQL ID of the type as well in order to update the store.

darrikonn (changes requested):
Decided to stop here since I wasn't sure if the PR was ready for review or not.

If it's ready for a review, I'll continue :)

libs/application/templates/general-fishing-license/src/fields/Declined/WarningSection.css.ts
sigruntg: Perfect thank you for this, I'll take a look at this :D. Just copy pasted this from drivers license to get the payment going, I can also see that I'm only using two of these styles!

libs/application/templates/general-fishing-license/src/graphql/queries.ts
sigruntg: Do you mean query the __typename?

sigruntg: @darrikonn

darrikonn: No (sorry for late answers, missed this somehow).
So now there's a query name called `status` in the GraphQL schema that's accessible for every single application. The name needs to relate to e.g. payment. Check https://docs.devland.is/technical-overview/api-design-guide/graphql-naming-conventions#integrating-naming-conventions-into-shared-api for reference.

Regarding the GraphQL ID. I'm assuming that you're querying the payment status of the application. You however don't query the application id as well, so Apollo doesn't know how to construct its cache, making the payment status not accessible in other components that might need it. A better approach would be to do something like:
```typescript
query application($input: ApplicationInput!) {
  application(input: $input) {
    id
    paymentStatus
  }
}
```
where variables are:
```typescript
variables: {
  input: {id: applicationId}
}
```

This way, the Apollo cache can construct the cache by setting the paymentStatus to the appropriate application because the GraphQL ID is fetched as well (id).

sigruntg: I ended up removing this part since Arnór made a reusable PaymentPending ui component.

darrikonn (approved):
Core looks good.

libs/application/templates/general-fishing-license/src/fields/components/ShipInformation/index.tsx
```diff
@@ -56,23 +56,23 @@ export const ShipInformation: FC<ShipInformationProps> = ({
...
+      {!!grossTons && (
         <ValueLine
           label={formatMessage(shipSelection.labels.grossTonn)}
           value={grossTons.toString()}
           disabled={isExpired}
           color={isExpired ? 'grey' : 'black'}
         />
       )}
-      {length && (
+      {!!length && (
```
Thorrik58, line 75: Why did you go with throwing everything to truthy values?

libs/application/templates/general-fishing-license/src/forms/GeneralFishingLicensePaymentForm.ts
```diff
@@ -0,0 +1,66 @@
...
+export const GeneralFishingLicensePaymentForm: Form = buildForm({
+  id: 'GeneralFishingLicensePaymentForm',
+  title: '',
+  mode: FormModes.APPLYING,
+  renderLastScreenButton: false,
+  children: [
+    buildSection({
+      id: 'ExternalDataSection',
+      title: externalData.dataProvider.sectionTitle,
+      children: [],
```
Thorrik58, line 27: Why are there so many empty sections? Are you doing this to fake the progress of the status bar?

libs/application/templates/general-fishing-license/src/fields/Declined/index.tsx
```diff
@@ -0,0 +1,30 @@
...
+    <>
+      <Box paddingTop={2}>
+        <Text>{formatMessage(declined.labels.helpText)}</Text>
+      </Box>
+      <Box marginTop={5} marginBottom={8}>
+        <WarningSection
+          application={application}
+          step={{
+            title: formatMessage(declined.labels.otherCountryTitle),
+            key: 'key',
```
Thorrik58, line 22: This key isnt used anywhere is it?

libs/application/templates/general-fishing-license/src/fields/Declined/WarningSection.tsx
```diff
@@ -0,0 +1,79 @@
...
+  return (
+    <Box
+      position="relative"
+      border="standard"
+      borderRadius="large"
+      padding={4}
+      marginBottom={2}
+      borderColor={'red600'}
+    >
+      {/* Warning Icon */}
```
Thorrik58, line 38: wouldnt really say that these comments are needed.

Thorrik58 (changes requested):
Minor comments

libs/application/templates/general-fishing-license/src/fields/components/ShipInformation/index.tsx
sigruntg: Because sometimes some values returned 0 and then I would only see 0.

libs/application/templates/general-fishing-license/src/forms/GeneralFishingLicensePaymentForm.ts
sigruntg: Yes, this is to fake the status bar, it is done in all the applications that use payment

libs/application/templates/general-fishing-license/src/fields/Declined/index.tsx
sigruntg: And neither is the state, truthfully just copy pasted this from another application with payment hehe

libs/application/templates/general-fishing-license/src/fields/Declined/WarningSection.tsx
sigruntg: Removed them

PR #6504, opened 2022-02-04
feat(fa): Spouse tax return

Rikkiabb:
# ...
https://app.asana.com/0/1201722624788165/1201672092806127

## What

- Fetches spouse tax return file
- Adds data gathering screen to spouse flow
- Adds privacy policy to spouse flow
- Adds bunch of reusable components
- Adds file type to taxReturnFromRskFile

## Why
- Needed for processing application
- Users needs to accept we are fetching data about them
- Should have been here but wasn't 🤷 
- Better to reusseeeee
- Needed to know who uploads file

## Checklist:


MargretFinnboga (commented):
Few paths that need fixing 📍

apps/financial-aid/web-osk/src/routes/application/DataGathering/applicantDataGathering.tsx
```diff
@@ -149,58 +142,17 @@ const ApplicationInfo = () => {
...
-          og vinnsluferlið og staðfesta réttleika upplýsinga.
-        </Text>
-
-        <Text marginBottom={[4, 4, 5]}>
-          Við þurfum að fá þig til að renna yfir nokkur atriði varðandi þína
-          persónuhagi og fjármál til að reikna út fjárhagsaðstoð til útgreiðslu
-          í byrjun {getNextPeriod.month}. Í lok umsóknar getur þú sent hana inn
-          eða eytt henni og öllum tengdum gögnum.
-        </Text>
+        <AboutDataGathering nextMonth={getNextPeriod.month} />
```
MargretFinnboga, line 147: why are you sending this as props when you can get straight access to this in @island.is/financial-aid/shared/lib _question_

apps/financial-aid/web-osk/src/routes/application/DataGathering/spouseDataGathering.tsx
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  return (
+    <>
+      <ContentContainer>
+        <DataGatheringHeader />
+
+        <TaxDataGathering />
+
+        <AboutDataGathering nextMonth={getNextPeriod.month} />
```
MargretFinnboga, line 88: same question

MargretFinnboga (commented):
more nitpicks 🙃

apps/financial-aid/web-osk/src/routes/application/DataGathering/applicantDataGathering.tsx
Rikkiabb: Nice will change, I thought for a second it had something to do with the application state but it doesn't so will move

apps/financial-aid/web-osk/src/utils/hooks/useApplication.ts
```diff
@@ -41,7 +41,12 @@ const useApplication = () => {
         const files = formatFiles(form.taxReturnFiles, FileType.TAXRETURN)
           .concat(formatFiles(form.incomeFiles, FileType.INCOME))
           .concat(formatFiles(form.otherFiles, FileType.OTHER))
-          .concat(formatFiles(form.taxReturnFromRskFile, FileType.TAXRETURN))
+          .concat(
+            formatFiles(
+              form.taxReturnFromRskFile,
+              form.taxReturnFromRskFile[0].type,
+            ),
```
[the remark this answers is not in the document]
Rikkiabb: Very good question but I don't think so, because this calls a function `formatFiles` which maps over the files and returns them in a certain format. This array is initialized to [], so the array itself won't be undefined and if the list is empty then we are mapping over empty array and return nothing. 

However I think a better implementation of our files would be that we would have one filelist in the form and that list has the filetype like this does, instead of concatting here, but that is a latter task.

MargretFinnboga (approved):
Well done 👏🏼 ⭐

PR #6509, opened 2022-02-07
fix(fa): making muncipality homepage nullable

MargretFinnboga:
# ... ☝🏼 

https://app.asana.com/0/1201722624788165/1201672092806062

## What
 - Making municipality homepage nullable in the backend

## Why

- So when admin and municipality is created at the same time the new admin can log in without

## Checklist:


davidpxn (approved):
Amazing 😍

Rikkiabb (changes requested):
Are we null checking for this where we use it?

Rikkiabb:
You will have to change interfaces and dtos for municipality as well

Rikkiabb (approved):

PR #6511, opened 2022-02-07
feat(ci): Adding Lifecycle policy tags to all images

Snojo:
# ...

## What

## Why

## Screenshots / Gifs

## Checklist:


.github/workflows/push.yml
```diff
@@ -171,10 +172,16 @@ jobs:
           fi
           echo "Using SHA='$SHA' as docker tag sha"
           export DOCKER_TAG=${GITHUB_RUN_NUMBER}_${DOCKER_BRANCH_TAG}_${SHA:0:7}
+          export POLICY_TAG=${DOCKER_BRANCH_TAG}_${GITHUB_RUN_NUMBER}_${SHA:0:7}
```
dabbeg, line 175: Any idea how hard it would be two make this the new DOCKER_TAG instead of adding this additional one? Are there many places referencing the current format of the dockertag? @Snojo @pshomov

.github/workflows/push.yml
```diff
@@ -545,6 +562,9 @@ jobs:
       - name: Retag unaffected Docker images
         if: "!startsWith(env.GIT_BRANCH_DEPLOY, 'feature/')"
         run: ./scripts/ci/retag-unaffected.sh "$UNAFFECTED"
+      - name: Retag unaffected Docker images with lifecycle
+        if: "!startsWith(env.GIT_BRANCH_DEPLOY, 'feature/')"
+        run: ./scripts/ci/retag-lifecycle-policy.sh "$UNAFFECTED" #Retags feature images for lifecycle.
```
dabbeg, line 567: Can you do something like `DOCKER_TAG=feature-$POLICY_TAG ./scripts/ci/retag-unaffected.sh "$UNAFFECTED"` instead of creating the additional retag script?

dabbeg (commented):
Good job, the functionality looks solid! If possible I would like changing the current docker tag to match the format needed for the ecr lifecycle policy to reduce the complexity.

.github/workflows/push.yml
Snojo: That's actually something I meant to do after reading over my code. 
I'll make the change

.github/workflows/push.yml
Snojo: I do agree. I did mention that to Petar but he was hesitant to take that step. I think it would require quite a big refactoring session to do.

.github/workflows/push.yml
```diff
@@ -171,10 +172,16 @@ jobs:
           fi
           echo "Using SHA='$SHA' as docker tag sha"
           export DOCKER_TAG=${GITHUB_RUN_NUMBER}_${DOCKER_BRANCH_TAG}_${SHA:0:7}
+          export POLICY_TAG=${DOCKER_BRANCH_TAG}_${GITHUB_RUN_NUMBER}_${SHA:0:7}
+          
           echo "Docker tag will be ${DOCKER_TAG}"
           echo "::set-output name=DOCKER_TAG::${DOCKER_TAG}"
           echo "DOCKER_TAG=$DOCKER_TAG" >> $GITHUB_ENV
 
+          echo "Docker tag will be ${POLICY_TAG}"
```
dabbeg, line 181: `echo "Policy tag will be ${POLICY_TAG}"`

.github/workflows/push.yml
```diff
@@ -120,6 +120,7 @@ jobs:
       UNAFFECTED: ${{ steps.unaffected.outputs.UNAFFECTED }}
       BUILD_CHUNKS: ${{ steps.build_map.outputs.BUILD_CHUNKS }}
       IMAGES: ${{ steps.deploy_map.outputs.IMAGES }}
+      POLICY_TAG: ${{steps.docker_tags.outputs.POLICY_TAG }}
```
dabbeg, line 123: Also one thing, consider changing `POLICY_TAG` to something like `LIFECYCLE_TAG`.

.github/workflows/push.yml
dabbeg: Okay I understand. As I think about it, this would be a necessary step if we would want to do that refactor.

We would need to do something like:
1. Tag the images with both versions of the tag. (This PR)
2. Refactor all references to use the new version of the tag.
3. Remove the old version of the tag.

andes-it:

andes-it:

PR #6513, opened 2022-02-07
feat(auth-public-api): Add special rules for access control delegation

eirikurn:
## What

For now no-one should be able to give other users the access control scope
except procuring holders on behalf of companies, and in those cases, the
receiving user should not be able to give it forward.

Fixes https://www.notion.so/Only-allow-delegations-with-access-control-scopes-for-companies-865dc75af3d3413bb071713a687ce787

## Why

To provide default delegation security and prevent recursive access control delegation.

## Checklist:


darrikonn (approved):
LGTM

libs/auth-api-lib/src/lib/config/DelegationConfig.ts
```diff
@@ -0,0 +1,40 @@
...
+      //
+      // * Self: Normal "person" users can grant this scope to other users on their behalf.
+      // * ProcurationHolder: "Company owners" can grant this scope to other users on behalf of their company.
+      // * LegalGuardian: "Parents" can grant this scope to other users on behalf of their child.
+      // * Custom: Custom delegatee can grant this scope to other users on behalf of the original delegator.
+      //
+      // In all cases, the active user still needs to have this scope and the
+      // `delegation:write` scope in their access token.
+      onlyForDelegationType: z.array(
+        z.enum(['Self', 'ProcurationHolder', 'LegalGuardian', 'Custom']),
```
valurefugl, line 23: How about personal representatives?

libs/auth-api-lib/src/lib/config/DelegationConfig.ts
```diff
@@ -0,0 +1,40 @@
...
+  ),
+})
+
+export const DelegationConfig = defineConfig({
+  name: 'DelegationConfig',
+  schema,
+  load: (env) => ({
+    customScopeRules: env.optionalJSON('DELEGATION_CUSTOM_SCOPE_RULES') ?? [
+      {
+        scopeName: '@island.is/auth/delegations:write',
```
valurefugl, line 35: Should we not use `AuthScope.writeDelegations` here?

valurefugl (approved):

PR #6521, opened 2022-02-07
fix(judicial-system): Fixes to modify case dates functionality

oddsson:
# Fixes to modify case dates functionality

https://app.asana.com/0/1199153462262248/1201437594297191/f

## What

- Add a loading indicator to the primary button in the modify case dates modal and disable it while sending notifications.
- Now, when a user sets the validToDate to a date before isolation to date, the isolation to date is changed accordingly. This means that if valid to date is 20.02.2022 and isolation to date is 15.02.2022 and the valid to date is changed to 10.02.2022, the isolation is automatically changed to 05.02.2022.

## Why

This is overall, better UX

## Checklist:


apps/judicial-system/web/src/routes/Shared/SignedVerdictOverview/SignedVerdictOverview.tsx
```diff
@@ -609,6 +649,7 @@ export const SignedVerdictOverview: React.FC = () => {
                 )}
                 isPrimaryButtonDisabled={isCaseModificationInvalid()}
                 handlePrimaryButtonClick={() => handleDateModification()}
+                isPrimaryButtonLoading={isSendingNotification}
```
gudjong, line 652: Consider modifying the condition to include isUpdatingCase. Otherwise, the button will be enabled for a brief moment while the client waits for the update, before the notification is sent.

gudjong (changes requested):

PR #6538, opened 2022-02-08
fix(an): file upload check for multiples

sigruntg:
# Check if files are still being uploaded

[1x1ppfj](https://app.clickup.com/t/1x1ppfj)

## What

If you are adding multiple files you can get through as long as one of the files have been loaded. I made it so that if there are still any files with the status "uploading", no files will be uploaded, hence the data schema will stop you from going to the next page before all files have been uploaded.

## Checklist:


libs/shared/form-fields/src/lib/FileUploadController/index.tsx
```diff
@@ -100,15 +100,23 @@ export const FileUploadController: FC<FileUploadControllerProps> = ({
...
+      const onlyUploadedFiles = state.filter(
+        (f: UploadFile) => f.key && f.status === 'done',
+      )
 
-    setValue(id, uploadAnswer)
+      const uploadAnswer: UploadFileAnswer[] = onlyUploadedFiles.map(
+        transformToAnswer,
+      )
+
+      setValue(id, uploadAnswer)
```
eirikurn, line 118: What happens if I finish uploading 1 file completely, then start uploading a new file. Can it also block the continue button somehow in that case?

sigruntg: This ended up going to draft state, there where more issues with the file uploader that we are looking into with Júní.

PR #6543, opened 2022-02-09
feat(service-portal): Add better loading and error state for dynamic finance pages

thordurhhh:
## What
Dynamic finance pages return a 404 before the service has responded.

* Add loader to dynamic finance pages & Hide in nav while loading..
* If the user does not have access it will render the accessdenied page
* If the user does have access it will render the page

## Checklist:


Valur (approved):
LGTM

PR #6557, opened 2022-02-09
fix(service-portal): get access token on clicking document

disaerna:
# FIX - access token invalid when downloading file from Pósthólf

## What

Fixing a bug in Pósthólf when user tries to download a file after waiting a bit with the screen open.
User gets "Access denied" error. 
Fetch token on click

## Why

🐞 

## Checklist:

### DIFF
--- a/apps/services/auth-public-api/test/setup.ts
+++ b/apps/services/auth-public-api/test/setup.ts
@@ -63,7 +63,7 @@ export const setupWithAuth = async ({
   scopes = Scopes,
 }: SetupOptions): Promise<TestApp> => {
   // Setup app with authentication and database
-  const app = await testServer<AppModule>({
+  const app = await testServer({
     appModule: AppModule,
     override: (builder: TestingModuleBuilder) =>
       builder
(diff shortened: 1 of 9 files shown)


libs/service-portal/documents/src/components/DocumentLine/DocumentLine.tsx
```diff
@@ -24,6 +25,7 @@ interface Props {
 const DocumentLine: FC<Props> = ({ documentLine, userInfo, img }) => {
   const { formatMessage } = useLocale()
   const onClickHandler = () => {
+    getAccessToken()
```
eirikurn, line 28: Unfortunately you need to await and use this access token. When the token is not expired it takes 0ms, but may take up to a second if it's expired.

We also need to test this in the main browsers and see if this async delay triggers a popup blocker. :(

disaerna: Ah I see, too good to be true then 😆

libs/service-portal/documents/src/components/DocumentLine/DocumentLine.tsx
```diff
@@ -23,12 +17,16 @@ interface Props {
...
-  const onClickHandler = () => {
+
+  const onClickHandler = async () => {
     // Create form elements
     const form = document.createElement('form')
     const documentIdInput = document.createElement('input')
     const tokenInput = document.createElement('input')
 
+    const token = await getAccessToken()
+    if (!token) return
```
Valur, line 28: Maybe there should be a toast.error when this fails?

Valur (approved):

PR #6569, opened 2022-02-10
feat(data-protection-complaint): Application to pdf added as attachment.

obmagnusson:
# ...

Added pdf version of application as attachment to the application. 

- Added add attachemnt to application in shared template api service

## What

Making the pdf available to the applicant as a download when the application is submitted.

## Why

Request from Persónuvernd

## Screenshots / Gifs

## Checklist:

### DIFF
--- a/apps/services/user-notification/src/app/modules/notifications/messageProcessor.service.ts
+++ b/apps/services/user-notification/src/app/modules/notifications/messageProcessor.service.ts
@@ -46,7 +46,7 @@ export class MessageProcessorService {
           title: t.formatMessage(title, formatArgs),
           body: t.formatMessage(body, formatArgs),
           category: 'NEW_DOCUMENT',
-          appURI: `${this.appProtocol}://document/${message.documentId}`,
+          appURI: `${this.appProtocol}://inbox/${message.documentId}`,
         }
       }
     }
--- a/libs/shared/components/src/auth/UserMenu/UserMenu.tsx
+++ b/libs/shared/components/src/auth/UserMenu/UserMenu.tsx
@@ -42,7 +42,7 @@ export const UserMenu = ({
 
   return (
     <Box display="flex" position="relative" height="full">
-      <Hidden below="lg">
+      <Hidden below="md">
         <UserLanguageSwitcher user={user} />
       </Hidden>
       <UserButton user={user} onClick={handleClick} small={small} />


eirikurn (changes requested):
Nice job. Just a few tweaks needed.

libs/application/template-api-modules/src/lib/modules/shared/shared.service.ts
```diff
@@ -205,4 +211,61 @@ export class SharedTemplateApiService {
...
+      authorization,
+      ADD_ATTACHMENT_MUTATION,
+      {
+        input: {
+          id: applicationId,
+          key,
+          url,
+        },
+      },
+    )
```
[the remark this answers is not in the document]
eirikurn: Ah, and that of course :D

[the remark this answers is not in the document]
obmagnusson: So keep the gql query for now?

libs/application/template-api-modules/src/lib/modules/templates/data-protection-complaint/data-protection-complaint.service.ts
```diff
@@ -18,16 +19,36 @@ export class DataProtectionComplaintService {
...
+      if (!complaintPdf?.content) throw new Error('No pdf content')
+
+      const key = await this.sharedService.addAttachment(
+        auth.authorization,
+        application.id,
+        'kvörtun.pdf',
+        complaintPdf.fileBuffer,
+        {
+          ContentEncoding: 'base64',
+          ContentDisposition: 'inline',
```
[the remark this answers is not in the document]
obmagnusson: Not sure

libs/application/template-api-modules/src/lib/modules/templates/data-protection-complaint/data-protection-complaint.service.ts
```diff
@@ -18,16 +19,36 @@ export class DataProtectionComplaintService {
...
+
+      if (!complaintPdf?.content) throw new Error('No pdf content')
+
+      const key = await this.sharedService.addAttachment(
+        auth.authorization,
+        application.id,
+        'kvörtun.pdf',
+        complaintPdf.fileBuffer,
+        {
+          ContentEncoding: 'base64',
```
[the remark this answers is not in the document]
obmagnusson: no not necessary

libs/application/template-api-modules/src/lib/modules/shared/shared.service.ts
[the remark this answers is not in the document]
eirikurn: Hm, no.

What stands in the way of this saving the attachment to the application model in the same way as the controller?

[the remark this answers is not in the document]
obmagnusson: Well nothing really. So then just inject the Application Service and calling update? Just have to have a place for the "am I allowed to edit this application" check.

libs/application/template-api-modules/src/lib/modules/shared/shared.service.ts
```diff
@@ -205,4 +213,39 @@ export class SharedTemplateApiService {
...
+    },
+  ): Promise<string> {
+    const uploadBucket = getConfigValue(
+      this.configService,
+      'attachmentBucket',
+    ) as string
+
+    const uploadParams = {
+      Bucket: uploadBucket,
+      Key: fileName,
```
eirikurn, line 234: This might override the same "Kvörtun.pdf" file for all users in the root.

The S3 key structure on the application system is something like this (please double check):

`{applicationId}/{uuid}-{fileName}`?

libs/application/template-api-modules/src/lib/modules/templates/accident-notification/accident-notification.service.spec.ts
```diff
@@ -8,7 +8,10 @@ import {
 import { logger, LOGGER_PROVIDER } from '@island.is/logging'
 import { EmailService } from '@island.is/email-service'
 import { SharedTemplateApiService } from '../../shared'
-import { TemplateApiModuleActionProps } from '../../../types'
+import {
+  BaseTemplateApiApplicationService,
+  TemplateApiModuleActionProps,
```
eirikurn, line 13: Unused?

apps/application-system/api/src/app/modules/application/application.module.ts
```diff
@@ -46,14 +47,18 @@ if (process.env.INIT_SCHEMA === 'true') {
   })
 }
 
+@Global()
```
eirikurn, line 50: Would be best to refactor some of this "app" into "libs" which are easier to share without circular dependencies.

libs/application/template-api-modules/src/lib/modules/shared/shared.service.ts
```diff
@@ -205,4 +213,39 @@ export class SharedTemplateApiService {
         return data?.applicationPaymentStatus
       })
   }
+
+  async addAttachment(
+    applicationId: string,
```
eirikurn, line 218: Should accept `application` like other methods.

apps/application-system/api/src/app/modules/application/template-api.service.ts
```diff
@@ -0,0 +1,34 @@
...
+  ): Promise<void> {
+    const existingApplication = await this.applicationService.findOneById(
+      applicationId,
+    )
+
+    if (!existingApplication) {
+      throw new NotFoundException(
+        `An application with the id ${applicationId} does not exist`,
+      )
+    }
```
eirikurn, line 25: Should accept an application object. No need to check if it exists.

The `key` and `url` arguments feel too low level since we don't want anyone to mess with either. Maybe we should move the upload and key logic here? Could also reuse logic from S3/FileService.

eirikurn (changes requested):
Almost there. Just one issue and a couple of nitpicks.

libs/application/template-api-modules/src/lib/modules/shared/shared.service.ts
obmagnusson: Ahh of course missed that when going between buckets.

apps/application-system/api/src/app/modules/application/application.module.ts
obmagnusson: Yes agree that would be preferable. Could we let this one slip by for now? I would like to create another PR with this refactor.

eirikurn: Totally!

apps/application-system/api/src/app/modules/application/template-api.service.ts
```diff
@@ -0,0 +1,54 @@
...
+      ContentType?: string
+      ContentDisposition?: string
+      ContentEncoding?: string
+    },
+  ): Promise<string> {
+    const uploadBucket = this.config.attachmentBucket
+    if (!uploadBucket) throw new Error('No attachment bucket configured')
+
+    const fileId = uuid()
+    const key = `${application.id}/${fileId}-${fileName}`
```
eirikurn, line 37: Super minor but currently the attachment key is `${fileId}-${fileName}` while the S3 key is `${application.id}/${fileId}-${fileName}`.

Would be good to keep this consistent.

Rikkiabb (approved):

PR #6571, opened 2022-02-10
feat(fa): Autofill spouse email

davidpxn:
# ... 👆

https://app.asana.com/0/1201722624788165/1201785888802173

## What

When an applicant fills out the form and selects that he has a spouse, the spouse email is required. Then when the spouse fills out the form we want the email field to be autofilled with the email that the applicant already entered.

## Why

Better UX.

## Checklist:

### DIFF
--- a/apps/financial-aid/web-osk/src/components/FormProvider/FormProvider.tsx
+++ b/apps/financial-aid/web-osk/src/components/FormProvider/FormProvider.tsx
@@ -29,6 +29,7 @@ export interface Form {
   ledger?: string
   accountNumber?: string
   emailAddress?: string
+  emailEdited: boolean
   interview?: boolean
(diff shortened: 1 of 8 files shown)


apps/financial-aid/web-osk/src/components/FormProvider/FormProvider.tsx
```diff
@@ -29,6 +29,7 @@ export interface Form {
   ledger?: string
   accountNumber?: string
   emailAddress?: string
+  emailEdited: boolean
```
Rikkiabb, line 32: Is this needed? Can't we just send the value even though it hasn't been updated?

Rikkiabb (commented):
Check it, nice work btw 🕴️

PR #6574, opened 2022-02-10
feat(fa): Autofill spouse email

davidpxn:
# ... 👆

https://app.asana.com/0/1201722624788165/1201785888802173

## What

When an applicant fills out the form and selects that he has a spouse, the spouse email is required. Then when the spouse fills out the form we want the email field to be autofilled with the email that the applicant already entered.

## Why

Better UX.

## Checklist:


apps/financial-aid/backend/src/app/modules/application/models/spouse.response.ts
```diff
@@ -8,5 +8,8 @@ export class SpouseResponse {
   hasFiles: boolean
 
   @ApiProperty()
-  spouseName?: string
+  applicantName?: string
+
+  @ApiProperty()
+  applicantSpouseEmail?: string
```
Rikkiabb, line 14: Shouldn't this be `applicantName` and `applicantSpouseEmail`, that is no `?` since we return an empty string but not undefined from the service?

Rikkiabb (commented):
One question

Rikkiabb (approved):
👏

PR #6577, opened 2022-02-11
refactor(testing): Remove generic typing from testServer

programm-ingovals:
# Remove generic typing from testServer

## What

The generic typing in testServer is unneccesary.
The function itself does not use the type for anything than to pass it along to InfraModule
Changing the typing in the options object allows TypeScript to infer the type.

Also changes to the useDatabase hook since Sequelize 5 and lower use
Bluebird's library version of promises, while hooks are expected to return ES6 promises.

## Why

The generic version was causing typescript errors on usage even though it built fine.

## Checklist:


darrikonn (approved):

FridrikMProgramm (approved):
LGTM

eirikurn (approved):

PR #6584, opened 2022-02-11
fix(fa): Format date of attached files

davidpxn:
# ... 👆

https://app.asana.com/0/1201722624788165/1201722624788121

## What

Make the format of date on files more readable.

## Why

Better UX

## Screenshots / Gifs

image

## Checklist:


libs/financial-aid/shared/src/components/FileList/FileList.tsx
```diff
@@ -71,7 +72,11 @@ const FileList = ({ className, files }: Props) => {
                   item,
                 )} KB`}</Text>
                 {'created' in item && (
-                  <Text variant="small"> {`${item.created}`}</Text>
+                  <div className={styles.date}>
+                    <Text variant="small">
+                      {format(new Date(item.created), 'dd.MM.y  · kk:mm')}
```
Rikkiabb, line 77: Not to be addressed in this pr but maybe we should have a date format funtion in utils. It would help if we do ever change the formats then we just have to do it in a single place.

Rikkiabb (approved):

PR #6596, opened 2022-02-14
feat(auth-api): Update auth api with personal-representative features

programm-ingovals:
# Personal Representative - Update to auth-api to enable Identity Server to use the delegation system for Personal Representatives

## What

Changes to two endpoints on the _Delgations controller_ of the _services-auth-api_ project.
First off is the **findAllTo** that lists all the entities that a user is delegating for. This PR adds a check on the PersonalRepresentative relationship that have been added via the [PersonalRepresentative service](https://github.com/island-is/island.is/blob/main/apps/services/personal-representative/src/app/modules/personalRepresentatives/personalRepresentatives.controller.ts). National registry is used to fetch the names of the delegated users.
The new type claim for the delegation is `PersonalRepresentative`

Second is **findAllScopesTo** that fetches the scopes a user has access to when delegating. Now it is possible to use the new type claim `PersonalRepresentative` to fetch scopes a user has delegated to him according to the rules of _PersonalRepresentative right types_.

## Why

This is an addition to the delegation system made specifically for represented persons. This allows projects that use **island.is** identityServer authentication and authorization to enable PersonalRepresentatives access to their clients resources according to the rules set by the regulatory authority. 

## Changes

- Added functionality to delegations controller methods for personalRepresentatives
- Defines a new type of implicit delegations that identity server can fetch from auth-api
- Enables identity server to use PersonalRepresentative features that are configured via the PersonalRepresentative API
- Moved ScopePermissons model, DTO and Service into PersonalRepresentative subfolder in library
- Added new feature-flag to easily turn on/off feature
- Added reference from api-scope to PersonalRepresentativeScopePermission
- Added local seed data
- Changed testing setup for auth-api ( was unused ) to a use TestServer similar to auth-api-public
- Fixed some circular reference issues

## Checklist:


libs/auth-api-lib/src/lib/services/delegationScope.service.ts
```diff
@@ -144,6 +150,63 @@ export class DelegationScopeService {
...
+                    {
+                      model: PersonalRepresentative,
+                      required: true,
+                      where: {
+                        [Op.and]: [
+                          {
+                            nationalIdPersonalRepresentative: toNationalId,
+                          },
+                          {
+                            nationalIdRepresentedPerson: fromNationalId,
```
eirikurn, line 193: 4 joins, yay!

Should there be a couple of indices on these two fields to help with common queries?

libs/auth-api-lib/src/lib/services/delegationScope.service.ts
```diff
@@ -144,6 +150,63 @@ export class DelegationScopeService {
     ]
   }
 
+  async findPersonalRepresentativeScopes(
+    toNationalId: string,
+    fromNationalId: string,
+  ): Promise<string[]> {
+    const apiScopes = await this.apiScopeModel.findAll({
```
eirikurn, line 157: Might want to select only the name of the top model.

apps/services/auth-api/test/setup.ts
```diff
@@ -1,58 +1,78 @@
...
+  einstaklingarGetForsja = () => []
+}
+
+interface SetupOptions {
+  user: User
 }
 
-export const setup = async (options?: Partial<TestServerOptions>) => {
-  app = await testServerActivateAuthGuards({
+export const setupWithAuth = async ({
```
eirikurn, line 26: Nice upgrades!

apps/services/auth-api/src/app/modules/delegations/delegations.controller.spec.ts
```diff
@@ -0,0 +1,393 @@
...
+      describe.each([
+        [1, 0, 0],
+        [2, 0, 0],
+        [0, 0, 0],
+        [0, 1, 0],
+        [0, 0, 1],
+        [0, 1, 1],
+        [1, 1, 1],
+      ])(
+        'and given user has %d active clients with valid rights, %d active clients with outdate rights and %d active clients with unactivated',
```
eirikurn, line 138: Are we using "clients" in many places like this? Would like to choose another word that isn't confused with OAauth2 client. (was super confused 😆 )

apps/services/auth-api/src/app/modules/delegations/delegations.controller.spec.ts
```diff
@@ -0,0 +1,393 @@
...
+import {
+  getPersonalRepresentativeRelationship,
+  getPersonalRepresentativeRights,
+  getPersonalRepresentativeRightType,
+  getPRenabledApiScope,
+  getScopePermission,
+  personalRepresentativeType,
+} from '../../../../test/stubs/personalRepresentativeStubs'
+
+describe('DelegationsController', () => {
```
eirikurn, line 36: Great job with these tests!

FridrikMProgramm (approved):
Changes look good to me.
Nice testing setup :)

eirikurn:
Also, will you add auth-public-api changes to another PR?

programm-ingovals:
@eirikurn 

I have no changes to the auth-public-api. The findAll endpoint there uses the service from the private api so should work regardless. I was under the impression that other endpoints there were only relevant to explicit delegations.

Something I'm missing?

programm-ingovals:
Added indices to **PersonalRepresentative**. Two separate indices ( since fields might be searched separately ) using hash ( since these are strings, not comparisons ).
I absolutely willing to defer to someone better at DBs than me.

eirikurn:
> @eirikurn
> 
> I have no changes to the auth-public-api. The findAll endpoint there uses the service from the private api so should work regardless. I was under the impression that other endpoints there were only relevant to explicit delegations.
> 
> Something I'm missing?

No, you're probably good. Right @saevarma? The only thing I can think of is API tests. There are thorough tests for this on the auth-api side, but none on the auth-public-api side. Maybe understandable since it's the same logic being tested.

eirikurn:
> Added indices to **PersonalRepresentative**. Two separate indices ( since fields might be searched separately ) using hash ( since these are strings, not comparisons ). I absolutely willing to defer to someone better at DBs than me.

I'm not a DBA either, but we should probably use the default B-tree index for this reason:

image

We are using cloud replicas, don't know if those are affected. But best to follow their recommendation and be safe with backups and replicas.

saevarma:
> > @eirikurn
> > I have no changes to the auth-public-api. The findAll endpoint there uses the service from the private api so should work regardless. I was under the impression that other endpoints there were only relevant to explicit delegations.
> > Something I'm missing?
> 
> No, you're probably good. Right @saevarma? The only thing I can think of is API tests. There are thorough tests for this on the auth-api side, but none on the auth-public-api side. Maybe understandable since it's the same logic being tested.

Yes, you are right. It's a bit tricky situation with the test setup for these apis. But maybe we should add happy path tests for the `ActorDelegationsController` to make sure it returns the personal representative rights and only those you should see?

programm-ingovals:
I've added test for **auth-public-api** and changed to using B-tree index.

eirikurn (approved):
Nice!

PR #6439, opened 2022-02-01
feat: Scripts to recreate DB schema for feature-deployed services

pshomov:
# ...

Creating this to support self-service

## What

Avoid manual work in the future

## Why

Less devops friction

## Screenshots / Gifs

Attach Screenshots / Gifs to help reviewers understand the scope of the pull request

## Checklist:

- [ ] I have performed a self-review of my own code
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] Formatting passes locally with my changes
- [ ] I have rebased against main before asking for a review

PR #6600, opened 2022-02-14
fix(core): regression in proxy scripts

eirikurn:
## What

Fix a regression in the proxy scripts. They broke when `run-proxy.js` was refactored in #6439.

## Why

So DX works.


eirikurn:
cc @pshomov 😊

pshomov (approved):
LGTM

PR #6618, opened 2022-02-15
chore(web): html validation tweaks

stjanilofts:
Using [Nu Html Checker](https://validator.github.io/validator/site/nu-about.html) to scan the Sýslumenn pages and fix the most glaring markup errors/warnings. Some of those changes are made in the island-ui library so they will also fix the markup elsewhere where those components are used.

## Checklist:


apps/web/screens/Home/Home.tsx
```diff
@@ -114,6 +114,7 @@ const Home: Screen<HomeProps> = ({ categories, news, page }) => {
           <NewLinks
             heading={n('newsTickerHeading')}
             seeMoreText={n('newsTickerSeeMore')}
+            showMoreButton
```
RunarVestmann, line 117: This adds a button on the homepage:

image

But shouldn't something happen if I click the button?

apps/web/components/Organization/Wrapper/OrganizationWrapper.tsx
```diff
@@ -281,8 +282,17 @@ export const OrganizationWrapper: React.FC<WrapperProps> = ({
...
+  const { width } = useWindowSize()
+  const [isMobile, setIsMobile] = useState(false)
   const router = useRouter()
 
+  useEffect(() => {
+    if (width < theme.breakpoints.md) {
+      return setIsMobile(true)
+    }
+    setIsMobile(false)
+  }, [width])
```
RunarVestmann, line 294: This could be written in fewer lines but how it is right now can be argued to be more clarifying (your call on what you want to do here)

`useEffect(() => {
  setIsMobile(width < theme.breakpoints.md)
}, [width])`

RunarVestmann (changes requested):
Some minor comments from me.

apps/web/screens/Home/Home.tsx
stjanilofts: Forgot that in there :)

libs/island-ui/core/src/lib/Navigation/Navigation.tsx
```diff
@@ -350,17 +350,18 @@ const MobileButton = ({ title, colorScheme }: MobileButtonProps) => {
         variant="h5"
         color={colorSchemeColors[colorScheme]['activeColor']}
       >
-        {title}
+        {title} asd
```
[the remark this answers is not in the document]
stjanilofts: oops

apps/web/components/Organization/Wrapper/OrganizationWrapper.tsx
stjanilofts: After looking at it again I think we don't need the useEffect/useState for this. The `width` value from `useWindowSize` is causing a render on window width change so we can just have a simple `const isMobile = width < theme.breakpoints.md`.

RunarVestmann (approved):

albinagu (approved):
👍

oddsson (approved):

PR #6638, opened 2022-02-16
chore(judicial-system): Stops using barrel files in the api project

gudjong:
# Stops using barrel files in the api project

https://app.asana.com/0/1199153462262248/1201721129628614/f

## What

☝️ 

## Why

Barrel files are not recommended in NestJS.

## Checklist:


Rikkiabb (changes requested):
Looks good if you want to stop using barrel files. 

It kind off goes against being domain driven which the monorepo wants to be, but I guess very few are using judicial-system specific domains.

Also wanted to put here, another solution which fixes the barrel issue problem, which is having all modules exported from a single barrel file which exports nothing else and then other domain specific barrel files export other items. Like we do here: https://github.com/island-is/island.is/blob/600464010f111ab3acacd07fb6add8fe8d538765/apps/financial-aid/backend/src/app/modules/index.ts

PR #6645, opened 2022-02-17
feat(application-system): delegations

alexdiljar:
**Delegations in application system**

Screenshot 2022-03-11 at 09 50 30

Screenshot 2022-03-16 at 13 03 55

Further info here ->  https://www.notion.so/The-exclusive-on-applications-delegations-in-progress-06de4e1eda2248e19b3e6867eec80b43

Can be tested here https://application-systemdelegations-beta.dev01.devland.is/umsoknir/p-merki
with login 010-7789

## Checklist:


handbook/reference/problems/actor-validation-failed.md
```diff
@@ -0,0 +1,29 @@
...
+
+The server will not process the request due to validation issues around actor delegations.
+
+The client can try again after delegating to correct delegation.
+
+## Extra metadata
+
+### `fields`
+
+A JSON object listing validation issues where delegatedUser is the nationalId of the user assigned to the resource that the actor is a and actor is the nationalId of the user trying to access the resource.
```
kksteini, line 11: > resource that the **actor is a and** actor is the nationalId of the user trying to access the resource.
Don't understand this. Is something missing?

albinagu: @alexandradilja perhaps? :D

handbook/reference/problems/actor-validation-failed.md
```diff
@@ -0,0 +1,41 @@
+# 400 Actor Validation Failed
+
+The server will not process the request due to validation issues around actor delegations.
+This error should be thrown when user (actor) tries to access a resource that they do not have access to but someone they have delegation rights for (subject) has access to.
```
kksteini, line 4:
```suggestion
This error should be thrown when a user (actor) tries to access a resource that they do not have access to but someone they have delegation rights for (subject) has access to.
```

handbook/reference/problems/actor-validation-failed.md
```diff
@@ -0,0 +1,41 @@
...
+The error should contain the national id of the subject that is assigned to the resource and the types of delegations the actor has for the subject.
+Handeling this error could then involve allowing the user to delegate to the subject assigned to the resource so that they can access it as an actor.
```
kksteini, line 6:
```suggestion
Handling this error could then involve allowing the user to delegate to the subject assigned to the resource so that they can access it as an actor.
```

handbook/reference/problems/actor-validation-failed.md
```diff
@@ -0,0 +1,41 @@
...
+
+### `fields`
+
+A JSON object where delegatedUser is the national id of the subject that is assigned to the resource that the actor is trying to access.
+The delegations array contains delegations that the actor has for the subject assigned to the resource.
+The delegation object contain the name of the subject, the subjects national id as well as the type of delegation the actor has for the subject.
```
kksteini, line 12:
```suggestion
The delegation object contains the name of the subject, the subject's national id as well as the type of delegation the actor has for the subject.
```

libs/application/templates/reference-template/README.md
```diff
@@ -74,3 +74,37 @@ Run `ng test application-templates-reference-template` to execute the unit tests
 ## Custom API functionality
 
 Should your template require custom API actions, like calling an external API or sending an email you should head over to `libs/application/template-api-modules/README.md`
+
+## Applications that support delegations
+
+Applications do not support user delegations by default, however it is simple to add this feature to your application type.
+
+When an application supports user delegations of a specific type then a user with actor delegations of the corresponding delegation type will be prompted to choose what user they are applying for before creating a new application. They can choose to switch to a subject from their actor delegations and the subject will be the applicant for a new application and the actor's national id will be stored in the actors field for on the application.
```
kksteini, line 82:
```suggestion
When an application supports user delegations of a specific type then a user with actor delegations of the corresponding delegation type will be prompted to choose what user they are applying for before creating a new application. They can choose to switch to a subject from their actor delegations and the subject will be the applicant for a new application and the actor's national id will be stored in the actors field on the application.
```

kksteini (approved):

siggifv (approved):

apps/application-system/api/src/app/modules/application/tools/applicationAccess.service.ts
```diff
@@ -22,4 +46,77 @@ export class ApplicationAccessService {
 
     return existingApplication as Application
   }
+
+  async findOneByIdAndDelegations(id: string, auth: Auth) {
+    const actorDelegations = await this.delegationsApiWithAuth(
+      auth,
+    ).actorDelegationsControllerFindAll({
+      direction: ActorDelegationsControllerFindAllDirectionEnum.Incoming,
+    })
```
eirikurn, line 55: We need to add guard logic to most application controller endpoints to make sure delegations can't access or update application which don't allow the current delegation (403 Forbidden).

    E.g. so I can't sign into my child and apply for parental leave.

I would like to get rid of the new `/delegated/:id` endpoint and move this logic (or similar) into `ApplicationAccessService#findOneByIdAndNationalId` so the applicationApplication query returns the correct problem.

Like mentioned elsewhere, we don't need to depend on Auth Public API here, and we shouldn't use the deprecated `delegation.fromName` and `delegation.toName`.

This logic can be much simpler. Just check if `user.nationalId` or `actor?.nationalId` is in the `application.applicantActors` array, then throw our fancy problem with the `application.applicant` nationalId.

eirikurn (changes requested):
Really good first draft. Since we're doing a fairly complex change to a fairly complex project it's difficult to figure out all the implementation details in advanced. Thus this review is pretty loaded. 🤗❤️

Didn't have a place to put this comment, but it's very important to test all application API behaviour thoroughly. We need tests that verify:

1. That the correct problem is thrown in all the right circumstances.
2. That delegated users are unable to access/modify applications which don't support that type of delegations.

apps/application-system/api/src/app/modules/application/tools/applicationAccess.service.ts
alexandradilja: Thank you for the feedback! I think these are really good ideas and a good direction, making this as simple as it can be is the goal 😄  
One thing I question is 

_Just check if user.nationalId or actor?.nationalId is in the application.applicantActors array, then throw our fancy problem with the application.applicant nationalId._

Since more than one person can be actors for the same subject. If one actor or the subject itself creates the application then the second actor would not be assigned in the `applicantActors` array and the error would not be thrown. 
How can we check if the user has the `application.applicant `in their delegation, and therefore should get the fancy problem if we don't check if the applicant exists in their delegations in the `applicationAccess.service`

For all other use cases your suggestions are spot on!

eirikurn: Great point. IMHO this is an edge case (second actor joining an existing application) we should either not worry about, or solve in a different way.

This only happens if someone opens a direct link to an application they have a delegation for but have never opened before. Typically, the user would sign into My Pages, switch to the delegation, find the application there to open it, in which case they'd drop right in.

I think this is rare enough to warrant simplifying the code and skipping the extra Auth Public API integration. 

A different way would be to have a "generic" delegation switcher on some application system "Access Denied" page. Google Drive does something similar: "You do not have access. You are signed in as Eirikur Nilsson. Switch accounts." We can show a delegation dropdown right there, or a simple link which sends the user to the IDP delegation prompt:

image

alexandradilja: This is a good point! Let's go with that and try to handle it else where. Project simplification STARTS NOW 🦾

libs/api/domains/syslumenn/src/lib/syslumenn.resolver.ts
```diff
@@ -56,7 +56,6 @@ export class SyslumennResolver {
   }
 
   @Query(() => CertificateInfoResponse)
-  @Scopes(ApiScope.internal)
```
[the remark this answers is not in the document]
alexdiljar: The internal scope does not travel with delegations. For testing purposes I removed it and once we figure out the scope functionality with delegations I'll add it back in!

libs/application/templates/p-sign/src/lib/pSignTemplate.ts
```diff
@@ -22,6 +23,7 @@ const PSignTemplate: ApplicationTemplate<
   name: 'Stæðiskort',
   dataSchema: dataSchema,
   readyForProduction: true,
+  allowedDelegations: [AuthDelegationType.Custom],
```
alexdiljar, line 26: For the purpose of deploy feature testing I have this as custom to be able to test with gervimaður Afríka as subject and gervimaður Útlönd as actor. Gervimaður Afríka is the only gervimaður with læknisvottorð for the pSign application checks.

libs/application/core/src/types/ApplicationTemplate.ts
```diff
@@ -26,6 +26,7 @@ export interface ApplicationTemplate<
   readonly name: StaticText
   readonly institution?: StaticText
   readonly translationNamespaces?: string[]
+  readonly allowedDelegations?: string[]
```
eirikurn, line 29: It might be a good idea to have an enum here, and for V1 we might want it to not support Custom delegations...

handbook/reference/problems/bad-subject.md
```diff
@@ -0,0 +1,23 @@
+# 403 Bad Subject
+
+The server will not process the request since the current subject claims don't have access.
+This problem is different from `403 Forbidden` in that it indicates that the user might be able to switch to another subject to get access, e.g. using the delegation system. It can both mean that the user needs to switch to a specific delegation, or that they must switch from a delegation to the original authenticated user.
+
+## Example
```
eirikurn, line 6: Would be nice to document the optional "alternativeSubjects" field.

eirikurn (commented):
Reviewed everything except the React stuff.

Awesome job! Really clean, well commented and documented and nice details all around. 💯 

Now we just need API Tests for the application system api. 😄 Should verify the actorApplicant logic and test the guards in at least the primary endpoints. Can look into using `it.each` to do the same test for more than one endpoint (Sævar used it in auth-public-api tests).

libs/application/graphql/src/lib/fragments/application.ts
```diff
@@ -7,6 +7,7 @@ export const ApplicationFragment = gql`
     modified
     applicant
     assignees
+    applicantActors
```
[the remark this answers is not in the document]
alexdiljar: Me and Gunnar discussed being able to test the P-merki application with actors all the way, including submitting the actors to sýslumenn. Maybe it should be a follow up PR and this one will focus on only adding the functionality and testing with tests and not real applications?

libs/application/template-api-modules/src/lib/modules/templates/parental-leave/parental-leave.utils.spec.ts
```diff
@@ -27,6 +27,7 @@ const createApplicationBase = (): Application => ({
   answers: {},
   applicant: '',
   assignees: [],
+  applicantActors: [],
```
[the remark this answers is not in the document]
alexdiljar: Wow, didn't know that was an option for testing. I'll take it up in the next umsóknir faghópur meeting!

johannaagma (approved):

workspace.json
```diff
@@ -1952,7 +1952,7 @@
           },
           "configurations": {
             "production": {
-              "optimization": true,
+              "optimization": false,
```
darrikonn, line 1955: Why wouldn't you want to have optimization set to true?

qdd1 (approved):

libs/application/templates/reference-template/README.md
```diff
@@ -74,3 +74,37 @@ Run `ng test application-templates-reference-template` to execute the unit tests
...
+When an application supports user delegations of a specific type then a user with actor delegations of the corresponding delegation type will be prompted to choose what user they are applying for before creating a new application. They can choose to switch to a subject from their actor delegations and the subject will be the applicant for a new application and the actor's national id will be stored in the applicantActors field on the application.
+
+If a user should open up a drafted application where the applicant is a user they have correct delegation for the user will be prompted to switch to the correct subject.
+
+If another user with the correct delegation rights for the applicant should update the application they will be added to the applicantActors list.
+
+Should your template require user delegation support you will need to configure the types of delegations that the application should support to the template:
+
+```ts
+import { DelegationType } from '@island.is/auth-api-lib'
```
saevarma, line 91: This PR made me realise that we have duplicate defintion for `DelegationType` one in `@island.is/auth-api-lib`  and another one in `@island.is/auth-nest-tools` (the `ApplicationTemplate.ts` uses the on from auth-nest-tools). 

I think I will move this type to `@island.is/shared/types` in a separate PR.

libs/application/ui-shell/src/components/DelegationsScreen.tsx
```diff
@@ -0,0 +1,228 @@
...
+  coreMessages,
+  getTypeFromSlug,
+} from '@island.is/application/core'
+import { getApplicationTemplateByTypeId } from '@island.is/application/template-loader'
+import { LoadingShell } from './LoadingShell'
+import { format as formatKennitala } from 'kennitala'
+import { useLocale } from '@island.is/localization'
+import { useHistory } from 'react-router-dom'
+import { ScreenType, DelegationsScreenDataType, Delegation } from '../types'
+import { ErrorShell } from './ErrorShell'
```
saevarma, line 24: Unused

libs/application/ui-shell/src/components/DelegationsScreen.tsx
```diff
@@ -0,0 +1,228 @@
...
+    screenType: ScreenType.LOADING,
+  })
+  const { formatMessage } = useLocale()
+  const type = getTypeFromSlug(slug)
+  const { switchUser, userInfo: user } = useAuth()
+  const history = useHistory()
+  const featureFlagClient: FeatureFlagClient = useFeatureFlagClient()
+
+  // Check for user delegations if application supports delegations
+  const { data: delegations, error } = useQuery(ACTOR_DELEGATIONS, {
```
saevarma, line 48: This query now supports array of delegationTypes to filter by.

libs/application/ui-shell/src/lib/ApplicationForm.tsx
```diff
@@ -43,13 +58,30 @@ const ApplicationLoader: FC<{
...
   })
+
   const application = data?.applicationApplication
 
   if (loading) {
     return <LoadingShell />
   }
 
   if (!applicationId || error) {
+    const foundError = findProblemInApolloError(error as any, [
```
saevarma, line 69: I think you shouldn't need to do the any type casting.

apps/application-system/api/src/app/modules/application/application.controller.ts
```diff
@@ -116,6 +118,7 @@ export class ApplicationController {
   ) {}
 
   @Scopes(ApplicationScope.read)
+  @UseGuards(DelegationGuard)
```
saevarma, line 121: The `DelegationGuard` is applied to all methods expect for the `delete`, shouldn't it be there as well? Could we just add it in the `UseGuards`  on the controller level?

saevarma (changes requested):

apps/application-system/api/src/app/modules/application/application.controller.ts
```diff
@@ -1107,7 +1130,7 @@ export class ApplicationController {
     const { nationalId } = user
     const existingApplication = (await this.applicationAccessService.findOneByIdAndNationalId(
       id,
-      nationalId,
+      user,
```
obmagnusson, line 1133: Is there a reason you left out a DelegationGuard for Delete?

eirikurn (approved):
Just reviewed my previous comments. Let me know if there's any other change you feel I should review especially.

Otherwise, great job with this!

apps/application-system/api/src/app/modules/application/e2e/utils.ts
```diff
@@ -0,0 +1,23 @@
+import request, { CallbackHandler } from 'supertest'
+
+export const getRequestMethod = (
```
saevarma, line 3: I saw that you added this to `@island.is/testing/nest` so you should remove this file.

libs/application/ui-shell/src/components/DelegationsScreen.tsx
saevarma: You didn't address this in the changes.

libs/application/ui-shell/src/components/DelegationsScreen.tsx
saevarma: You didn't address this in the changes

saevarma (approved):

obmagnusson (approved):
Awesome!

PR #6648, opened 2022-02-17
feat(email-service): generate email template to html file

stjanilofts:
### What

Added a command that generates an email template to an html file just for testing purposes.

```bash
yarn nx run email-service:generate-html
```

### Why

The other day I needed to make an example email for a designer so he could look at the template component variations. There was no easy way to do that other than setting up `email-service` and test them out using NodemailerApp, which is fine, but I did not have any project using `email-service` to quickly do that so I really wished there was a way to just generate an example html file using the CLI.

## Checklist:


eirikurn (approved):
Nice!

gudjong (approved):
Very nice!

PR #6652, opened 2022-02-18
feat(web): Organization alert banner on top of the page

RunarVestmann:
# Organization alert banner on top of the page

## What

* Placed the Organization alert banner above the header
* Deleted an unused OrganizationAlert component

## Why

* Previously the alert banner was below the header which goes against the guidelines for the AlertBanner component

### Before:
image

### After:
image

## Checklist:


kksteini (approved):
LGTM

siggifv (approved):

apps/web/layouts/main.tsx
```diff
@@ -199,9 +201,28 @@ const Layout: NextComponentType<
...
+          bannerId: `organization-alert-${stringHash(
+            JSON.stringify(organizationAlertBannerContent ?? {}),
+          )}`,
+          ...organizationAlertBannerContent,
+        },
+      ].filter(
+        (banner) => !Cookies.get(banner.bannerId) && banner?.showAlertBanner,
+      ),
+    )
+  }, [alertBannerContent, organizationAlertBannerContent])
```
barabrian, line 225: Just wondering, is there any particular reason the banners are stored in state? looks like the content is never changing, and the bannerId is stored in cookie

```suggestion
  const alertBanners = [
    {
      bannerId: `alert-${stringHash(
        JSON.stringify(alertBannerContent ?? {}),
      )}`,
      ...alertBannerContent,
    },
    {
      bannerId: `organization-alert-${stringHash(
        JSON.stringify(organizationAlertBannerContent ?? {}),
      )}`,
      ...organizationAlertBannerContent,
    },
  ].filter(
    (banner) => !Cookies.get(banner.bannerId) && banner?.showAlertBanner,
  )
```

barabrian (commented):
LGTM, just one comment

apps/web/layouts/main.tsx
RunarVestmann: I placed the alert banners in state for two reasons. Number one, since now we don't have to access the cookie on every re-render. Number two, because after I dismissed the alert banners and hit refresh I still saw the background color of them even though they weren't supposed to be rendered and somehow adding them to the state fixed that issue:
image

barabrian: That's weird, I don't see why the content would not be populated in the banner if this is a cookie issue. There might be something off with the AlertBanner component, but if this works it should be fine for now 🙃

barabrian (approved):

PR #6661, opened 2022-02-18
feat(fa):  Making application info collapsible 

MargretFinnboga:
# ...

https://app.asana.com/0/1201813977904020/1201752166226723

## What

- Making profile unit collapsible
- use animated height for comment section 

## Why

- More nice for staff
- cleaner and better

## Checklist:


apps/financial-aid/web-veita/src/components/ProfileUnit/ProfileUnit.tsx
```diff
@@ -14,16 +16,75 @@ interface Props {
...
+
+  if (collapsible) {
+    return (
+      <>
+        <button
+          className={cn({
+            [`${styles.toggleButton}`]: true,
+            [`${className}`]: className,
+          })}
+          onClick={() => setToggle(!toggle)}
```
Rikkiabb, line 40: Should send in the variable here as well :)

apps/financial-aid/web-veita/src/components/ProfileUnit/Unit.tsx
```diff
@@ -0,0 +1,66 @@
...
+            )}
+
+            {item.onclick && (
+              <button onClick={item.onclick} className={styles.button}>
+                {item.content}
+              </button>
+            )}
+
+            {!item.link && !item.onclick && <Text>{item.content}</Text>}
+
```
Rikkiabb, line 46: Can we take this out into a function, so the function returns the content instead of doing it inline?

Rikkiabb (commented):
Small stuff 🐛

apps/financial-aid/web-veita/src/components/ProfileUnit/Unit.tsx
```diff
@@ -0,0 +1,98 @@
...
+interface Props {
+  info: {
+    title: string
+    content?: string
+    link?: string
+    onclick?: () => void
+    other?: string
+  }[]
+  className?: string
+}
```
Rikkiabb, line 48: I like this a lot, I've been thinking of the structure, should we have this in the component self or have it live next to it and we have a naming convention... Something like `getUnitContent.ts`, what do you think?

PR #6663, opened 2022-02-18
fix(service-portal): Migration of current users + minor fixes.

thordurhhh:
## What

When a user is migrating from the old web, the data will be preserved.
But we need to set the status of the data in the userprofile db.
Updating the migration modal.
Updating the userprofile.service Readme.

## Why

Migration of islyklar data over to the /minarsidur pages.

## Checklist:


disaerna (approved):
Nice job 👍  love the mermaid 🧜‍♀️

PR #6664, opened 2022-02-18
feat(judicial-system): Removes heightened security when sharing requests

gudjong:
# Removes heightened security when sharing requests

Removes heightened security when sharing requests

## What

- Sets isHeightenedSecurityLevel to false when sharing a request.

## Why

Sharing is useless for requests with a heightened security level.

## Checklist:


apps/judicial-system/web/src/routes/Shared/SignedVerdictOverview/SignedVerdictOverview.tsx
```diff
@@ -429,6 +429,9 @@ export const SignedVerdictOverview: React.FC = () => {
             created: new Date().toString(),
             modified: new Date().toString(),
           },
+          isHeightenedSecurityLevel: workingCase.isHeightenedSecurityLevel
```
oddsson, line 432: What is the reasoning here? Why can't this be `isHeightenedSecurityLevel: false`? Aren't we always turning the heightened security level off when we share requests?

oddsson (changes requested):
One question

apps/judicial-system/web/src/routes/Shared/SignedVerdictOverview/SignedVerdictOverview.tsx
gudjong: Most of the time this field is undefined and I don't want to set it to false if it is undefined.

oddsson (approved):

PR #6667, opened 2022-02-18
fix(fa): Add redirect from /umsokn paths

davidpxn:
# ...  👆

https://app.asana.com/0/1201813977904020/1201752115305192

## What

Check if user has already submitted an application and redirect from /umsokn if so.

## Why

So the user cannot submit more than one application.

## Checklist:


apps/financial-aid/web-osk/src/components/AppLayout/AppLayout.tsx
```diff
@@ -11,18 +12,32 @@ import {
...
+  const redirectApplicant =
+    !user?.spouse?.hasPartnerApplied && user?.currentApplicationId
+
+  if (
+    router.pathname.startsWith(Routes.application) &&
+    (redirectApplicant || redirectSpouse)
+  ) {
+    router.push(`${Routes.statusPage(user.currentApplicationId as string)}`)
+    return null
+  }
```
MargretFinnboga, line 39: We already have a similar function returnUrl in pages/index.tsx? Maybe use that and putt it in shared?

MargretFinnboga (commented):
One suggestion, for reusability

Rikkiabb (changes requested):
Question in general about this PR, what does make `/umsokn`special? 

We route from '/' since we dont know where the user is heading there, but on `/umsokn` the user knows they are applying. Yes you can route to it and then we block the create application request in the end of the form. 

The user can route hack and go to `/umsoknir` but can't apply. After this change comes in then the user can`t route hack to `/umsoknir` but they can do it to every other page. So we are we making a special case here?

Rikkiabb:
> Question in general about this PR, what does make `/umsokn`special?
> 
> We route from '/' since we dont know where the user is heading there, but on `/umsokn` the user knows they are applying. Yes you can route to it and then we block the create application request in the end of the form.
> 
> The user can route hack and go to `/umsoknir` but can't apply. After this change comes in then the user can`t route hack to `/umsoknir` but they can do it to every other page. So we are we making a special case here?

My bad, didn't see you put it in applayout :)

Rikkiabb (commented):
My bad didn't see you put in AppLayout, thought it was in umsoknir because of the description :D

apps/financial-aid/web-osk/src/components/AppLayout/AppLayout.tsx
davidpxn: I decided to move all the redirection in index.tsx into AppLayout.tsx.

apps/financial-aid/web-osk/src/components/AppLayout/AppLayout.tsx
```diff
@@ -11,18 +12,48 @@ import {
...
+
+    return `${Routes.application}`
+  }
+
+  if (isAuthenticated && user) {
+    const redirectPath = getRedirectPath()
+
+    if (
+      router.pathname === '/' ||
+      (router.pathname.startsWith(Routes.application) &&
```
Rikkiabb, line 49: I would probably start with having this in the if check on line 44 and would also draw that out as a variable for readability. Since we are checking for 4 things then it can be nice to draw it out.

apps/financial-aid/web-osk/src/components/AppLayout/AppLayout.tsx
```diff
@@ -11,18 +12,48 @@ import {
...
+        !redirectPath.startsWith(Routes.application))
```
Rikkiabb, line 50: Instead of checking for this here, would it be better to check for it on line 41 and there return Routes.application if the pathname is '/'?

apps/financial-aid/web-osk/src/components/AppLayout/AppLayout.tsx
```diff
@@ -11,18 +12,48 @@ import {
...
+    ) {
+      router.push(redirectPath)
+      return null
```
Rikkiabb, line 53: Is this return needed?

Rikkiabb (changes requested):
Check it

apps/financial-aid/web-osk/src/components/AppLayout/AppLayout.tsx
davidpxn: It is not necessary but makes the UX a bit better. If we don't have the return then the page we are redirecting from gets displayed for a second, as it takes a bit of time for the redirect to happen.

Rikkiabb (approved):
🕶️

PR #6677, opened 2022-02-21
feat(fa): File upload

MargretFinnboga:
# File upload for income and tax return files

https://app.asana.com/0/1201849781730850/1201645065535607

## What

- get signed url from AS to our backend 
- Income files site 
- tax return file site 
- File upload functions
- Added files to summary 
- encoded and normalize file in utils

## Why

- Required

## Checklist:


libs/application/templates/financial-aid/src/forms/Application.ts
```diff
@@ -96,6 +96,29 @@ export const Application: Form = buildForm({
             }),
           ],
         }),
+        buildSubSection({
+          condition: (answers) => answers.income === ApproveOptions.Yes,
+          id: 'incomeFiles',
+          title: m.incomeFilesForm.general.sectionTitle,
+          children: [
+            buildCustomField({
+              id: routes.INCOMEFILES,
```
Thorrik58, line 105: Just wondering why do most of the custom fields have their id assigned from the routes constant but some don't?

libs/application/templates/financial-aid/src/lib/types.ts
```diff
@@ -51,6 +51,7 @@ export interface Applicant {
   fullName: string
   address: Address
   spouse?: Spouse
+  // folderId: UUID
```
Thorrik58, line 54: should this be removed?

Thorrik58 (commented):
Just some minor comments. 
Also you probably want to update the codeowners file so that kolibri owns all things related to municipalities-financial-aid.

libs/application/templates/financial-aid/src/fields/FileUploadContainer/FileUploadContainer.tsx
```diff
@@ -0,0 +1,34 @@
...
+  return (
+    <GridRow marginBottom={3}>
+      <GridColumn
+        span={['12/12', '12/12', '12/12', '9/12']}
+        offset={['0', '0', '0', '1/12']}
+      >
+        <Box marginBottom={[1, 1, 2]}>{children}</Box>
+        <div
+          className={cn({
+            [`${styles.errorMessage}`]: true,
```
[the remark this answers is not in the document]
MargretFinnboga: Yess it should, it has the same core styles, and then I am only transforming the opacity to show or now to show the message 😇

libs/application/templates/financial-aid/src/forms/Application.ts
MargretFinnboga: We are routing from the summary file also, so to avoid having to change the routes in two places, we put it in one and use it in the file and in summaryform file 😸

apps/financial-aid/backend/src/app/modules/personalTaxReturn/personalTaxReturn.controller.ts
```diff
@@ -22,18 +28,17 @@ export class PersonalTaxReturnController {
...
-  @UseGuards(RolesGuard)
-  @RolesRules(RolesRule.OSK)
   @ApiOkResponse({
     type: PersonalTaxReturnResponse,
     description: 'Fetches personal tax return and uploads it to s3',
   })
   async municipalitiesPersonalTaxReturn(
     @CurrentUser() user: User,
   ): Promise<PersonalTaxReturnResponse> {
+    // TODO: Get application id as query parameter input
```
MargretFinnboga, line 38: This was merge

VilliGunn (commented):
Just one minor comment and then I'm thinking if we could do a little better with the typings of the incomeFiles and taxReturnFiles like we discussed in a huddle 🙌 

PS. we could also maybe change the import order in some of the files 🙈

libs/application/templates/financial-aid/src/fields/Files/Files.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+  const {
+    files,
+    uploadErrorMessage,
+    onChange,
+    onRemove,
+    onRetry,
+  } = useFileUpload(uploadFiles, folderId)
+
+  const stringifyFile = (file: UploadFile) => {
```
davidpxn, line 30: Is there a better name for this function? Feel like stringifyFile is saying that it will turn the file into a string 🤔

davidpxn (commented):
One minor comment 👶

libs/application/templates/financial-aid/src/fields/Files/Files.tsx
MargretFinnboga: Do you have any suggestions? We named it this because this is doing the same as json.stringify, however It is not possible to directly convert a File object into JSON using JSON.stringify. So this function manually puts in the information of the file we need

davidpxn: But is this function not returning an object, unlike json.stringify that returns a string?

MargretFinnboga: Well this is basically what the function does, what name do you think will suit best? 😇 
Screenshot 2022-03-01 at 16 18 09

davidpxn (approved):
🏸

PR #6686, opened 2022-02-22
fix(fa): Add error message when national id is already in use

davidpxn:
# ... 👆

https://app.asana.com/0/1201849781730850/1201814168458673

## What

Check if the national id is already in use and display error message about that.

## Why

So the user know what the error is about.

## Screenshots / Gifs

image

## Checklist:

### DIFF
--- a/apps/financial-aid/web-veita/src/components/NewUserModal/NewUserModal.tsx
+++ b/apps/financial-aid/web-veita/src/components/NewUserModal/NewUserModal.tsx
@@ -21,8 +21,8 @@ interface newUsersModalState {
   staffName: string
   staffEmail: string
   hasError: boolean
-  hasNationalIdError: boolean
   hasSubmitError: boolean
+  errorMessage: string
   roles: StaffRole[]
 }
 
@@ -41,7 +41,7 @@ const NewUserModal = ({
     staffEmail: '',
     hasError: false,
     hasSubmitError: false,
-    hasNationalIdError: false,
+    errorMessage: '',
     roles: predefinedRoles,
   })
   const [createStaff] = useMutation(StaffMutation)
@@ -88,9 +88,11 @@ const NewUserModal = ({
       setState({
         ...state,
         hasSubmitError: true,
-        hasNationalIdError:
+        errorMessage:
           (e as ApolloError).graphQLErrors[0]?.extensions?.response.status ===
-          400,
+          400
+            ? 'Mögulega er notandi með þessa kennitölu til nú þegar'
+            : 'Eitthvað fór úrskeiðis, vinsamlega reynið aftur síðar',
       })
     }
   }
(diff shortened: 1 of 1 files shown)


apps/financial-aid/web-veita/src/components/NewUserModal/NewUserModal.tsx
```diff
@@ -93,7 +101,11 @@ const NewUserModal = ({
       setIsVisible={setIsVisible}
       header={'Nýr notandi'}
       hasError={state.hasSubmitError}
-      errorMessage={'Eitthvað fór úrskeiðis, vinsamlega reynið aftur síðar'}
+      errorMessage={
+        state.hasNationalIdError
+          ? 'Mögulega er notandi með þessa kennitölu til nú þegar'
+          : 'Eitthvað fór úrskeiðis, vinsamlega reynið aftur síðar'
+      }
```
Rikkiabb, line 108: Should we rather add the error message to the state and display that?

apps/financial-aid/web-veita/src/components/NewUserModal/NewUserModal.tsx
```diff
@@ -83,7 +85,13 @@ const NewUserModal = ({
         onStaffCreated()
       })
     } catch (e) {
-      setState({ ...state, hasSubmitError: true })
+      setState({
+        ...state,
+        hasSubmitError: true,
+        hasNationalIdError:
+          (e as ApolloError).graphQLErrors[0]?.extensions?.response.status ===
+          400,
```
Rikkiabb, line 93: I know we do it like this in other places, but is there another way of getting the error?

Rikkiabb (changes requested):
One small comment

apps/financial-aid/web-veita/src/components/NewUserModal/NewUserModal.tsx
davidpxn: This is is only place in the error object that has the status as a number

apps/financial-aid/web-veita/src/components/NewUserModal/NewUserModal.tsx
```diff
@@ -83,7 +85,15 @@ const NewUserModal = ({
         onStaffCreated()
       })
     } catch (e) {
-      setState({ ...state, hasSubmitError: true })
+      setState({
+        ...state,
+        hasSubmitError: true,
```
Rikkiabb, line 90: Do we need `hasSubmitError` ? Can't we just check if errorMessage if undefined? 

That is if we have it by default as undefined

davidpxn: goood idea😍

Rikkiabb (approved):
🍸

PR #6688, opened 2022-02-22
feat(services-user-profile): Remove email/sms verification in POST /userProfile

kristofer-lp8f:
## What

Remove code which is realistically never be used. The `userProfile.controller.ts:create()` function can only run if the calling user is not registered in the database. However, when a user is registering their email/phone they will always already have an account registered. Thus, in practice, a user will never have a registered email when signing up. At least, the email/phone will never be confirmed at registration.

## Why

Remove complexity and chance for exploitability. No tests cover e.g. sms verification portion of the `POST /userProfile` endpoint. Instead of writing tests for it (which can easily be mocked) I suggest we remove code which can never happen in the real world.

## Note
I'm not 💯  sure the relevant code is unused in all cases. Please correct me 🙂

## Checklist:

- [x] ~I have made corresponding changes to the documentation~
- [x] ~I have added tests that prove my fix is effective or that my feature works~

### DIFF
--- a/apps/services/user-profile/src/app/user-profile/e2e/userProfile.spec.ts
+++ b/apps/services/user-profile/src/app/user-profile/e2e/userProfile.spec.ts
@@ -78,41 +78,6 @@ describe('User profile API', () => {
       expect(response.body.id).toBeTruthy()
     })
 
-    it('POST /userProfile should register userProfile and create verification', async () => {
-      // Act
-      await request(app.getHttpServer())
-        .post('/emailVerification/')
-        .send({
-          nationalId: mockProfile.nationalId,
-          email: mockProfile.email,
-        })
-        .expect(204)
-
-      const verification = await EmailVerification.findOne({
-        where: { nationalId: mockProfile.nationalId },
-      })
-
-      const spy = jest
-        .spyOn(emailService, 'sendEmail')
-        .mockImplementation(() => Promise.resolve('user'))
-      const response = await request(app.getHttpServer())
-        .post('/userProfile')
-        .send({ ...mockProfile, emailCode: verification.hash })
-        .expect(201)
-      expect(spy).toHaveBeenCalled()
-      expect(response.body.id).toBeTruthy()
-
-      // Assert
-      expect(response.body.emailStatus).toEqual(DataStatus.VERIFIED)
-      expect(verification.nationalId).toEqual(mockProfile.nationalId)
-      expect(response.body).toEqual(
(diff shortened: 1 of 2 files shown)


kristofer-lp8f:
@island-is/hugsmidjan Any thoughts on this?

thordurhhh:
We definitely need this in the userprofile.create 

* The user should be able to sign up by posting their email and/or mobile number.
* If they are doing so, they should already have verified their data with the verification service.

kristofer-lp8f:
How can a user sign up with an already verified email/phone? Verification email/sms is only sent after signup, right?

thordurhhh:
> How can a user sign up with an already verified email/phone? Verification email/sms is only sent after signup, right?

It goes something like this:

User enters a registration page.
Enter their email. Get verification code sent to their email. Enter said code into input. Press "confirm". We call the userprofile.create with the code and end up [here](https://github.com/island-is/island.is/pull/6688/files#diff-9276af058e9e8d06b8aeee90d2123347077f0cae494d3f5da25b8aaace0404b5L117) check for a match. If it matches then we create a profile with said email and a status of VERIFIED.

Note: this is currently a feature that's behind a feature flag, and is being worked on. I have a PR I'm working on currently that will come in once we have tried and tested the features to work as they should.

This is a "little" more confusing as things stand now because of a little thing called the islyklar service.
We are in a migration phase right now, so we keep the email+tel in a different service, so this is definitely not as straight forward as this could otherwise be.

thordurhhh:
I am happy to jump on a Slack call to expand further on this.

kristofer-lp8f:
Then this PR is invalid.

Additional questions:
1. Is an email/phone required to sign up? If so, some constraint should be present in the `create()` function. Currently a user *could* sign up with no email/phone, or with unverified email and phone.
2. Does there exist an explanation of this flow (graph or prose)? The `services-user-profile/README.md` isn't helpful in this regard.

thordurhhh:
> Then this PR is invalid.
> 
> Additional questions:
> 
> 1. Is an email/phone required to sign up? If so, some constraint should be present in the `create()` function. Currently a user _could_ sign up with no email/phone, or with unverified email and phone.
> 2. Does there exist an explanation of this flow (graph or prose)? The `services-user-profile/README.md` isn't helpful in this regard.

1. No it's not required. A user can create a profile without setting their email/phone. And that's perfectly valid.
2. The readme is to be updated with the last additions which should be coming in later this week.

kristofer-lp8f:
@thordurhhh Thank you for your help 😄

PR #6708, opened 2022-02-24
feat(fa): State changes in AS on submission

Rikkiabb:
##
https://app.asana.com/0/1201849781730850/1201645065535623
https://app.asana.com/0/1201849781730850/1201645065535613

## What
- Renames routes to Routes
- Changes state of application on submission. Changes it to spouse if there is a spouse otherwise change it to submitted
- Adds blank spouse form
- Adds blank waiting for spouse form
- Adds blank submitted form
- Maps user to role for applicant and spouse
- Assigns application to spouse when application enters spouse state

## Why
- For naming convention
- We need it to handle different states in application
- Needed for spouse state
- Needed for applicant when application is in spouse state
- Needed for status page
- Access control for the application
- So spouse has access to application

### DIFF
--- a/libs/application/templates/financial-aid/src/lib/FinancialAidTemplate.ts
+++ b/libs/application/templates/financial-aid/src/lib/FinancialAidTemplate.ts
@@ -5,6 +5,7 @@ import {
   ApplicationStateSchema,
   DefaultEvents,
   ApplicationConfigurations,
+  Application,
 } from '@island.is/application/core'
 
 import { Roles, ApplicationStates, ONE_DAY, ONE_MONTH } from './constants'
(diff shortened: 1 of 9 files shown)


libs/application/templates/financial-aid/src/lib/utils.ts
```diff
@@ -8,3 +10,15 @@ export const isValidPhone = (value: string) => {
   return Boolean(phone && phone.isValid())
 }
 export const isValidNationalId = (value: string) => kennitala.isValid(value)
+
+export function hasSpouse(context: ApplicationContext) {
+  const {
+    externalData,
+    answers,
+  } = (context.application as unknown) as FAApplication
```
MargretFinnboga, line 18: as unknown, do we not have type for application?

libs/application/templates/financial-aid/src/lib/FinancialAidTemplate.ts
```diff
@@ -75,12 +82,87 @@ const FinancialAidTemplate: ApplicationTemplate<
...
+          }
+        }
+        return { ...context }
+      }),
     },
   },
 
-  mapUserToRole() {
-    return Roles.APPLICANT
+  mapUserToRole(id: string, application: Application) {
```
MargretFinnboga, line 158: nice 🌻

MargretFinnboga (commented):
minor check otherwise looks really good 🎖️

libs/application/templates/financial-aid/src/lib/utils.ts
Rikkiabb: Yes this is bacially doing that, we know that our application is of type `FAAppliction` but application context is {} so we have to cast it first to unknown to be able to cast it to our application :( 

Welcome to the application system ❤️

MargretFinnboga (approved):
🌟 ⭐ 🌟

PR #6714, opened 2022-02-24
chore(user-notification): Return empty message list from message-queue

maggisk:
## What

Make it clearer in logs if the worker is alive or dead when there are no incoming messages

## Why

There is currently no way of seeing from logs if the worker is dead or
idling because the queue is empty.
By returning an empty array, message-queue will log the empty response
every 20 seconds, making it clear in logs if the worker is dead or
just chilling because of no incoming messages

## Checklist:


rafnarnason (approved):
lgtm 🌮

Issue #6424, opened 2022-01-29
Add deployed code revision information to our audit logs.


pshomov:
This is follow up on a security meeting we had. The audit log statements should contain the revision of the code that emitted them.

PR #6724, opened 2022-02-25
chore: The audit log should output app version as well

dabbeg:
https://github.com/island-is/island.is/issues/6424

## What

Audit log should include app version

## Why

So we can determine what version of the app was running in the audit log

## Checklist:


pshomov:
Just wanted to put this out here for consideration. The `APP_VERSION` value is coming from an environment variable that is in tern rendered by Helm based on the docker image tag. So as far as immutability goes it is pretty flexible. Additionally due to re-tagging of docker images (and any bugs in that process) the revision number that we end up rendering deployment might be different from the build that originally created the docker image. We will still be able to find the original revision of the code by finding the "youngest" tag for a particular docker image but the whole thing does not give you feeling for being bullet-proof.
I would suggest a bit more involved process that should be more reliable and give us  the build revision that is more immutable and we should probably set both (DEPLOYMENT_VERSION=APP_VERSION and BUILD_VERSION) in the audit logs:

So I would suggest something like this:
- add to .gitignore a path to a .ts file(A) that will be created at build time 
- at build time, create the the content of file A something like `const version='${REV}'` (for local dev should be some distinctive value suggesting the build might not be clean)
- import and use file A in the audit logging code and use `version` value.

This will bake the build number in the source code so that should make it more reliable.
Thoughts?

darrikonn (approved):

PR #6734, opened 2022-02-28
fix(fa): Fix state change when assigning staff

davidpxn:
# ... 👆

https://app.asana.com/0/1201849781730850/1201552270013239

## What

- Change the state to INPROGRESS when assigning staff to a NEW application.

## Why

So the "Sjá um" button behaves correctly. 

## Checklist:


apps/financial-aid/web-veita/src/components/ApplicationHeader/ApplicationHeader.tsx
```diff
@@ -42,7 +43,7 @@ const ApplicationHeader = ({
 
     await changeApplicationState(
       application.id,
-      application.state,
+      ApplicationState.INPROGRESS,
```
MargretFinnboga, line 46: Are we sure we are always changing the state to inprogress??

MargretFinnboga (commented):
Just one check 🤓

MargretFinnboga (approved):
😎

PR #6737, opened 2022-03-01
chore(user-notification): disable fetch client circuit breaker

maggisk:
Requests from user-notification worker to user-profile service are super
slow and are triggering the fetch-client circuit breaker, so we're
disabling it (for now) while debugging

## Checklist:


rafnarnason (approved):
lgtm ✅

PR #6744, opened 2022-03-01
fix(fa): Fix state change error

davidpxn:
# ... 👆

https://app.asana.com/0/1201849781730850/1201892339168944

## What

Make directTaxPayments optional when updating application.

## Why

So that we don´t have to send a directTaxPayments object when updating other fields of the application.

## Checklist:


MargretFinnboga (approved):
🍾

PR #6745, opened 2022-03-01
fix(API): Fix missing graphql type for delegations

programm-ingovals:
Adds type for Personal Representative delegations.

Adds exhaustive type checking to ensure user get build errors when
adding new types to delegationType enum without adding new type.

## What

New PersonalRepresentative delegation did not work with GraphQL since it was missing type.
This adds a new type.

In addition a exhaustive check was added do ensure this issue does not repeat.

## Checklist:

### DIFF
--- a/libs/api/domains/auth/src/lib/models/delegation.model.ts
+++ b/libs/api/domains/auth/src/lib/models/delegation.model.ts
@@ -17,6 +17,10 @@ import { DelegationScope } from './delegationScope.model'
 registerEnumType(DelegationProvider, { name: 'AuthDelegationProvider' })
 registerEnumType(DelegationType, { name: 'AuthDelegationType' })
 
+const exhaustiveCheck = (param: never) => {
+  throw new Error(`Missing interfaceType ${param}`)
+}
+
 @InterfaceType('AuthDelegation', {
   resolveType(delegation: Delegation) {
     switch (delegation.type) {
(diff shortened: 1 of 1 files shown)


libs/api/domains/auth/src/lib/models/delegation.model.ts
```diff
@@ -24,9 +24,15 @@ registerEnumType(DelegationType, { name: 'AuthDelegationType' })
         return LegalGuardianDelegation
       case DelegationType.ProcurationHolder:
         return ProcuringHolderDelegation
+      case DelegationType.PersonalRepresentative:
+        return PersonalRepresentativeDelegation
       case DelegationType.Custom:
         return CustomDelegation
     }
+    const exhaustiveCheck = (param: never) => {
+      throw new Error(`Missing interfaceType ${param}`)
```
darrikonn, line 33: More readable to have this as the default statement of switch.

This error will show up at build time, right?

darrikonn (changes requested):

libs/api/domains/auth/src/lib/models/delegation.model.ts
programm-ingovals: Yes, this will cause a compile error if not all cases are covered. The error thrown should never happen unless `Delegation.Type` would be changed to something like `any`, in fact I believe it is removed from the resulting javascript.

Regarding setting it as default case, the linter will complain `Unexpected lexical declaration in case block.` since I'm declaring a const there. I could disable it. What way to do we prefer?

darrikonn (approved):

PR #6749, opened 2022-03-02
fix(ci): Force Codecov to return true

Snojo:
# ...

To make sure we can still build, even though codecov is having issues, we can just return the upload job as true and bipass it. 
It's not a good solution, but we can't use continue-on-failure in runs, and we're using free version of codecov, so we can't expect any help when hitting things like rate-limit. 

## What

## Why

## Screenshots / Gifs

## Checklist:


pshomov (approved):
LGTM

PR #6752, opened 2022-03-02
fix(ads): Children outside ADS location, flight endpoint fix

kksteini:
## What

Children, who were granted a flightcode based on their custodians location, but lived outside the ADS required area themselves would get denied on the `api/public/discounts/{discountCode}/flights` endpoint with 403.

Leveraging the cache from userService to fix that.

## Checklist:


apps/air-discount-scheme/backend/src/app/modules/flight/flight.controller.ts
```diff
@@ -213,9 +215,36 @@ export class PublicFlightController {
...
     }
 
-    const meetsADSRequirements = this.flightService.isADSPostalCode(
+    let meetsADSRequirements = this.flightService.isADSPostalCode(
       user.postalcode,
     )
+
+    // TODO: this is a quickly made temporary hotfix and should be rewritten
+    // along when the nationalregistry module is rewritten for the client V2 completely
+    if (!meetsADSRequirements) {
```
JuliusGG, line 224: You might consider adding a age check as well incase the edge case where user is actually adult and shouldn't be able to proceed.

kristofer-lp8f (approved):
Much nesting, but LGTM 👍

apps/air-discount-scheme/backend/src/app/modules/flight/flight.controller.ts
kksteini: The age check already takes place for when the cache is set here:
https://github.com/island-is/island.is/blob/d50c699ab9237f809a8c8aa643f110c47e1d080e/apps/air-discount-scheme/backend/src/app/modules/user/user.service.ts#L95-L111

JuliusGG: Up to you.
I reckon its cheaper to check age vs search cache.
Every user that has !adsRequirements will search cache.

kksteini: Ah, that however is a good point

kksteini: Fixed

PR #6757, opened 2022-03-03
fix(fa): Hot fix when application created without municipality code

MargretFinnboga:
# ... ☝🏼 

https://app.asana.com/0/1201849781730850/1201873101852414

## What

- If national registry isn't available but municipality information is stored, use that when creating application 

## Why
- Application are created without municipality code and that is necessary for application

## Checklist:


davidpxn (approved):
🦶💅

PR #6760, opened 2022-03-03
feat(fa): Add application_id to applications table

davidpxn:
# ... 👆

## What

- Add field application_id to applications table
- Update so the field is available in the backend and api
- Update so that it is possible to create an applications with application_id
- Update all interfaces and dto´s

## Why

To be able to create an application from AS.

## Screenshots / Gifs

image

## Checklist:


apps/financial-aid/backend/migrations/20220303095125-applicaitons-add-column-applicationid.js
```diff
@@ -0,0 +1,30 @@
...
+  up: (queryInterface, Sequelize) => {
+    return queryInterface.sequelize.transaction((t) =>
+      Promise.all([
+        queryInterface.addColumn(
+          'applications',
+          'application_id',
+          {
+            type: Sequelize.UUID,
+            allowNull: true,
+            unique: true,
```
MargretFinnboga, line 13: unique: true this should be false, because there can come a case where the applicant is able to create application in our DB but it didn't go through in the Application System DB

apps/financial-aid/backend/src/app/modules/application/dto/createApplication.dto.ts
```diff
@@ -156,4 +157,9 @@ export class CreateApplicationDto {
   @IsArray()
   @ApiProperty()
   readonly directTaxPayments: DirectTaxPayment[]
+
+  @IsOptional()
+  @IsUUID()
```
MargretFinnboga, line 162: nice 🥇

apps/financial-aid/backend/src/app/modules/application/models/application.model.ts
```diff
@@ -283,4 +283,12 @@ export class ApplicationModel extends Model<Application> {
   @HasMany(() => DirectTaxPaymentModel, 'applicationId')
   @ApiProperty({ type: DirectTaxPaymentModel, isArray: true })
   directTaxPayments: DirectTaxPaymentModel[]
+
+  @Column({
+    type: DataType.UUID,
+    allowNull: true,
+    unique: true,
```
MargretFinnboga, line 290: unique: true, should be false here also 🙃

libs/financial-aid/shared/src/lib/interfaces.ts
```diff
@@ -214,6 +214,7 @@ export interface CreateApplication {
   postalCode?: string
   city?: string
   municipalityCode?: string
+  applicationId?: string
```
MargretFinnboga, line 217: I search this file and sound we are using applicationId in quite a few places, so I think we need a more specific name applicationSystemId to not get confused

MargretFinnboga (commented):
🥸 🤝

PR #6771, opened 2022-03-03
fix(fa): Make municipalityCode mandatory in applications

davidpxn:
# ... 👆

https://app.asana.com/0/1201849781730850/1201905938180582

## What

- Update column municipality_code in applications table
- Update all relevant interfaces and dto´s
- Create migration script that replaces all null municipality_code values with "1400" (Hafnafjörður), and then updates the column to ban null values.

## Why

To prevent the creation of applications with no municipality_code. 

## Checklist:


gudjong (approved):
Nice, I think this should do the trick.

PR #6773, opened 2022-03-04
feat(judicial-system): Introduction field on ruling screen

oddsson:
# Introduction field on ruling screen

https://app.asana.com/0/1199153462262248/1201900945687128/f

## What

Introduction field on ruling screen

## Why

This is something that is being displayed in the ruling pdf but needs to be editable.

## Screenshots / Gifs

screencapture-localhost-4200-domur-rannsoknarheimild-urskurdur-76aeb066-f7c0-483b-ae6d-60780edcefdc-2022-03-04-10_21_19

## Checklist:


apps/judicial-system/web/src/routes/Court/InvestigationRequest/Ruling/RulingForm.tsx
```diff
@@ -80,6 +81,47 @@ const RulingForm: React.FC<Props> = (props) => {
...
+                updateCase,
+                setIntroductionEM,
+              )
+            }
+            errorMessage={introductionEM}
+            hasError={introductionEM !== ''}
+            textarea
+            rows={7}
+            autoExpand={{ on: true, maxHeight: 300 }}
+            required
```
Arnarkari93, line 122: Var ekki talað um að þetta myndi vera optional svæði? er ég að fara með rangð mál?

gudjong: Good question, but I don't think so. Until now, we have added a fixed text to the PDF. This will replace it, but there should always be some intro to the ruling.

apps/judicial-system/backend/src/app/formatters/rulingPdf.ts
```diff
@@ -65,12 +65,7 @@ function constructRulingPdf(
     }),
   )
   setLineGap(doc, 1)
-  addNormalJustifiedText(
-    doc,
-    formatMessage(ruling.intro, {
-      courtStartDate: formatDate(theCase.courtStartDate, 'PPP'),
-    }),
-  )
+  addNormalJustifiedText(doc, theCase.introduction ?? '')
```
gudjong, line 68: As this is a required field, consider defaulting to some CF text such as "Aðfararorð hafa ekki verið skráð".

gudjong: Also consider deleting the CF string `ruling.intro`. It has not been modified in CF and should be safe to delete.

gudjong (changes requested):

apps/judicial-system/web/src/routes/Court/InvestigationRequest/Ruling/RulingForm.tsx
oddsson: It's marked as required in the design 🤷

apps/judicial-system/backend/src/app/formatters/rulingPdf.ts
oddsson: Ruling.intro removed.

I'm confused about defaulting to a CF text. Why would we default to something when the string is required? That means that there will always be a value and no need for a default string, right?

PR #6778, opened 2022-03-04
fix(judicial-system): Fixes translator updates in investigation cases

gudjong:
# Fixes translator updates in investigation cases

https://app.asana.com/0/1199153462262248/1201918336079042/f

## What

☝️ 

## Why

Bug

## Checklist:


oddsson (changes requested):
Consider changing the type of the `field` parameter in `validateAndSendToServer` from `string` to `keyof Case` to prevent this from happening again.

Arnarkari93 (approved):
LGTM

Issue #6787, opened 2022-03-07
Spec up `service-portal-api`


pshomov:
Seems user-notification-worker service is hitting `service-portal-api` with 10 requests at a time. The worker services might be getting scaled up by HPA so that load might be higher. Thus we should do the following for this release:

1. Set min and max replica count for the worker to be 2
2. Double the memory and cpu for service-portal-api and allow it to have max replicas 30

PR #6788, opened 2022-03-07
fix(service-portal-api): Add resources to service portal api

Snojo:
# ...

Fixes https://github.com/island-is/island.is/issues/6787

## What

## Why

## Screenshots / Gifs

## Checklist:

### DIFF
--- a/apps/service-portal/infra/service-portal.ts
+++ b/apps/service-portal/infra/service-portal.ts
@@ -5,6 +5,15 @@ export const serviceSetup = (services: {}): ServiceBuilder<'service-portal'> =>
     .namespace('service-portal')
     .liveness('/liveness')
     .readiness('/readiness')
+    .replicaCount({
+      default: 2,
+      max: 20,
+      min: 2
+    })
+    .resources({
+      limits: { cpu: '400m', memory: '512Mi' },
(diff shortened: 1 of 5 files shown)


rafnarnason (approved):
lgtm 🌮

apps/service-portal/infra/service-portal.ts
```diff
@@ -5,6 +5,15 @@ export const serviceSetup = (services: {}): ServiceBuilder<'service-portal'> =>
...
     .readiness('/readiness')
+    .replicaCount({
+      default: 2,
+      max: 20,
+      min: 2,
+    })
+    .resources({
+      limits: { cpu: '400m', memory: '512Mi' },
+      requests: { cpu: '200m', memory: '256Mi' },
+    })
```
pshomov, line 16: Not sure why this is needed but maybe it's not a bad idea.

pshomov (approved):
LGTM

apps/service-portal/infra/service-portal.ts
Snojo: Neither was present, so I figured I'd add it, since I wanted to double resources there as well, away from the default.

PR #6790, opened 2022-03-07
fix(service-portal): print and dl button not visible when status 0

disaerna:
# Fix buttons not visible when status = 0

## What

Display buttons even though status is equal to 0. 

## Why

Users may want to print or download files even though their status is 0.

## Checklist:


thordurhhh (approved):
💯

PR #6796, opened 2022-03-07
feat(web): Project page stepper and new header

RunarVestmann:
# Project page stepper and new header

## What

* Added the new stepper to the Project pages
* Added a new header

## Why

* We want to be able to use the new stepper on Project pages
* We want to use the new header on a Ukranian support page

## Screenshots / Gifs
image

## Checklist:


kristofer-lp8f:
Shouldn't the banner image be stored in Contentful? Is there a reason to bundle the Ukraine page with the stepper? Where is dark mode for `'ukraine'`?

PR #6804, opened 2022-03-08
fix(auth): Handle edge cases better in delegation list

eirikurn:
## What

See commit messages.

## Why

Because our delegation list broke completely when failing to resolve user names.

## Checklist:


saevarma (approved):

libs/shared/utils/src/lib/formatNationalId.ts
```diff
@@ -0,0 +1,7 @@
+export const formatNationalId = (nationalId: string): string => {
```
darrikonn, line 1: The kennitala package already does this, so this isn't needed:

```typescript
import * as kennitala from 'kennitala'

const formattedNationalId = kennitala.format(someNationalId)
```

darrikonn (changes requested):
Nice work ... just the unneeded util that can be removed .. otherwise approved.

libs/shared/utils/src/lib/formatNationalId.ts
eirikurn: Ah of course. Copied this utility from another place in the monorepo. Should probably get rid of those too, just didn't want to pull in more code reviewers in this PR.

PR #6805, opened 2022-03-08
fix(judicial-system): fix prefill for ruling input

Arnarkari93:
# ...

[Asana task](https://app.asana.com/0/1199153462262248/1201918336079019/f)

## What

- Prefill input when the judge is drafting the ruling 

## Why

- It is prefilled later in the flow with the name of the judge

## Screenshots / Gifs

## Checklist:

### DIFF
--- a/apps/judicial-system/web/src/components/RulingInput/RulingInput.tsx
+++ b/apps/judicial-system/web/src/components/RulingInput/RulingInput.tsx
@@ -14,41 +14,32 @@ import {
 interface Props {
   workingCase: Case
   setWorkingCase: React.Dispatch<React.SetStateAction<Case>>
-  isCaseUpToDate: boolean
   isRequired: boolean
   rows?: number
 }
 
 const RulingInput: React.FC<Props> = (props) => {
-  const {
-    workingCase,
-    setWorkingCase,
-    isCaseUpToDate,
-    isRequired,
-    rows,
-  } = props
+  const { workingCase, setWorkingCase, isRequired, rows } = props
   const { updateCase, autofill } = useCase()
   const { formatMessage } = useIntl()
   const [rulingErrorMessage, setRulingErrorMessage] = useState('')
 
   useEffect(() => {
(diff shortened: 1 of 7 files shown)


apps/judicial-system/web/src/components/RulingInput/RulingInput.tsx
```diff
@@ -14,41 +14,32 @@ import {
...
-    isRequired,
-    rows,
-  } = props
+  const { workingCase, setWorkingCase, isRequired, rows } = props
   const { updateCase, autofill } = useCase()
   const { formatMessage } = useIntl()
   const [rulingErrorMessage, setRulingErrorMessage] = useState('')
 
   useEffect(() => {
-    if (isCaseUpToDate) {
```
Arnarkari93, line 35: isCaseUpToDate is set to false after one sec. Therefore, it is never set to `true` in modals.

oddsson (changes requested):
I agree with you, `isCaseUpToDate` will never be true in modals. However the `rulingInput` component is not only used in this modal, it's used on the ruling pages as well, and on those pages we need the `isCaseUpToDate` functionality. 

Maybe we should move the autofill out of the rulingInput component and onto the overview page? That way, the `isCaseUpToDate` functionality continues to work. What do you think?

Arnarkari93:
> I agree with you, `isCaseUpToDate` will never be true in modals. However the `rulingInput` component is not only used in this modal, it's used on the ruling pages as well, and on those pages we need the `isCaseUpToDate` functionality.
>   
> Maybe we should move the autofill out of the rulingInput component and onto the overview page? That way, the `isCaseUpToDate` functionality continues to work. What do you think?

I was thinking if we should rethink the autofill logic. Feels a bit hacky to control with a state that is reset with a `setTimeout`. I was thinking we could write a custom hook that we use for each field that needs to be autofill and make sure there it is only updated when we want it to.

Arnarkari93:
Close in favour of #6817

PR #6817, opened 2022-03-09
fix(judicial-system): autofill draft ruling

Arnarkari93:
# ...

[Asana taks](https://app.asana.com/0/1199153462262248/1201918336079019/f)

closes #6805 

## What

- clean up timeout in `useEffect`
- `DraftConclusionModal`
  - move from `Overview` to `OverviewForm`
  - move autofill hook outside of `RulingInput` to the `OverviewForm`
  - add e2e tests
  - mark `RulingInput` as required in `DarftConculsionModal`
- Move autofill hook call to top level

## Why

- We are only auto-filling inputs right after we have fetched the `workingCase`. We are controlling this with a boolean (`isCaseUpToDate`) in `FormContext` and we reset the boolean after one second with a timeout. This way the component needs to be rendered if their corresponding autofill hook is to trigger correctly. Thus, making it not work in modals. 
- To fix autofill for the `DraftConclusionModal` we move the autofill hook logic to the `OverviewForm`, which is rendering the modal.
- `RulingInput` in `DraftConclusionModal` was not marked required with a start but was treated as such.

## Screenshots / Gifs

Screenshot 2022-03-08 at 13 14 40

## Checklist:


apps/judicial-system/web/src/components/RulingInput/RulingInput.tsx
```diff
@@ -14,23 +14,16 @@ import {
...
-const RulingInput: React.FC<Props> = (props) => {
-  const {
-    workingCase,
-    setWorkingCase,
-    isCaseUpToDate,
-    isRequired,
-    rows,
-  } = props
-  const { updateCase, autofill } = useCase()
+export const useRulingAutofill = (
```
Arnarkari93, line 21: Moving the autofill outside of the component. This way the logic stays in the same place as the input.

apps/judicial-system/web/src/components/FormProvider/FormProvider.tsx
```diff
@@ -114,10 +114,14 @@ const FormProvider = ({ children }: Props) => {
...
   useEffect(() => {
+    let timeout: undefined | NodeJS.Timeout
     if (state === 'up-to-date') {
       // The case may change on the server so we only stay up to date for a short time
       // The time needs to be long enough to let hooks take appropriate actions, for instance auto fill
-      setTimeout(() => setState('ready'), 1000)
+      timeout = setTimeout(() => setState('ready'), 1000)
+    }
+    return () => {
+      timeout && clearInterval(timeout)
```
oddsson, line 124: Nice! 💪

oddsson (changes requested):
Neat! We generally handle all autofill logic in the data component. I'm wondering if we should autofill in apps/judicial-system/web/src/routes/Court/InvestigationRequest/Ruling/Ruling.tsx instead of using the useRulingAutofill hook in apps/judicial-system/web/src/routes/Court/InvestigationRequest/Ruling/RulingForm.tsx.

Arnarkari93:
> Neat! We generally handle all autofill logic in the data component. I'm wondering if we should autofill in apps/judicial-system/web/src/routes/Court/InvestigationRequest/Ruling/Ruling.tsx instead of using the useRulingAutofill hook in apps/judicial-system/web/src/routes/Court/InvestigationRequest/Ruling/RulingForm.tsx.

Yes, good point. Will update!

apps/judicial-system/web/src/routes/Court/InvestigationRequest/Ruling/Ruling.tsx
```diff
@@ -26,6 +27,8 @@ const Ruling = () => {
     document.title = 'Yfirlit kröfu - Réttarvörslugátt'
   }, [])
 
+  useRulingAutofill(isCaseUpToDate, workingCase)
```
oddsson, line 30: I think, instead of having a useRulingAutofill hook we should add the code inside it in the useEffect in this file, where we are already doing some autofilling.

oddsson (changes requested):
I see that you moved the `useRulingAutofill` hook into the data component. I'm thinking we should not have a specific hook for this autofill and do it in the useEffect in the data component, where we are already doing some autofilling.

oddsson (approved):

PR #6864, opened 2022-03-14
chore(user-notification): Fix error detection from firebase-admin

maggisk:
Fix error detection from firebase-admin

Firebase error documentation is rather poor.
When firebase returns an error on sending a push notification we must decide based on the response if the error is caused by an issue with the push token or not.
If we decide incorrectly that the issue is not with the token when it actually is the tokens fault, the message will end up in the sqs dead-letter-queue, which we'd rather avoid.

This change is based on errors we're seeing in datadog.

## Checklist:


rafnarnason (approved):
lgtm 🌮

PR #6878, opened 2022-03-15
feat(parental-leave): add explicit union confirmation

miamiultras:
# ...

If a union isn't specified, then let the user know that they might lose their rights

Currently its easy for user to accidentally skip that field without being notified, now it has to explicit choose to either use union or not

## Payment
image
## Review
image
## Review edit mode
image

## Checklist:


darrikonn (changes requested):
The select box should be hidden when no answer (or the answer is no) has been selected.

libs/application/templates/parental-leave/src/fields/Review/Review.tsx
```diff
@@ -320,30 +321,57 @@ export const Review: FC<ReviewScreenProps> = ({
...
+                    label: formatMessage(
+                      parentalLeaveFormMessages.shared.noOptionLabel,
+                    ),
+                    value: NO,
+                  },
+                ]}
+                onSelect={(s: string) => {
+                  setStateful((prev) => ({
+                    ...prev,
+                    useUnion: s as YesOrNo,
```
darrikonn, line 368: This probably applies for the other conditional rendering as well (the one blow)... but when no is selected, you've to manually reset the states of relevant inputs. ... so basically, you need to set this as null.

How to replicate?
1. radio = yes
2. select something in select box
3. radio = no
4. continue
5. now, union is set even though the user specified no.

darrikonn (changes requested):

eirikurn (approved):

libs/application/templates/parental-leave/src/fields/Review/Review.tsx
miamiultras: true, done

libs/application/ui-shell/src/components/Screen.tsx
```diff
@@ -255,6 +322,19 @@ const Screen: FC<ScreenProps> = ({
...
+    resetUnionValues()
+  }, [resetUnionValues, watchUseUnion, watchUnion])
+
+  useEffect(() => {
+    resetPensionFundValues()
+  }, [
+    resetPensionFundValues,
+    watchUsePrivatePensionFund,
+    watchPrivatePensionFund,
+    watchPrivatePensionFundPercentage,
```
darrikonn, line 335: The night's watch has begun

darrikonn (approved):

Thorrik58 (changes requested):
I need more info on what we are trying to fix here to give a decent review.
I really dont like the idea of changing the screen component to change something in a specific application.

libs/application/ui-shell/src/components/Screen.tsx
```diff
@@ -40,6 +40,7 @@ import {
   ProblemType,
 } from '@island.is/shared/problem'
 import { handleServerError } from '@island.is/application/ui-components'
+import { NO, YES } from '@island.is/application/templates/parental-leave'
```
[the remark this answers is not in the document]
miamiultras: good point

libs/application/ui-shell/src/components/Screen.tsx
```diff
@@ -239,6 +254,77 @@ const Screen: FC<ScreenProps> = ({
   const { width } = useWindowSize()
   const headerHeight = 85
 
+  const resetUnionValues = useCallback(async () => {
```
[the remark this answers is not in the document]
miamiultras: Original ticket was to make union more explicit, so now it follows the approach from privatePensionFund, where u have radio button and u have to explicitly choose yes or no.

But Darri found a bug:
image

TLDR: when u click yes, fill in the fields and then click no, u are still submitting fields that shouldn't be there.

I know that this solution is bad workaround, but I tried few others and I can't make it work, because of the app "frameworky" structure and the fact that part of the data is coming from api, part of the data is stored in local state and all of that has to be synced together. If u have better idea I am all ears or let me know if u need more info.

darrikonn (approved):

Thorrik58 (approved):
LGTM much better 👍

qdd1 (approved):

PR #6892, opened 2022-03-16
feat(web): English support for Sýslumenn specific pages

RunarVestmann:
# English support for Sýslumenn specific pages

## What

* Toggling english on and off on Sýslumenn specific pages now works
* Rendered the description field on the operating licenses page

## Checklist:


apps/web/screens/Organization/Syslumenn/Homestay.tsx
```diff
@@ -198,7 +201,13 @@ const Homestay: Screen<HomestayProps> = ({
   )
 }
 
-Homestay.getInitialProps = async ({ apolloClient, locale, query }) => {
+Homestay.getInitialProps = async ({ apolloClient, locale, pathname }) => {
+  const slug = pathname.split('/')?.[locale !== 'is' ? 3 : 2] ?? 'syslumenn'
+  const subSlug =
+    pathname.split('/')?.[locale !== 'is' ? 4 : 3] ?? 'heimagisting'
```
kristofer-lp8f, line 210: Can you make this more robust? Maybe searching for parameters or regex? Indexing slug seems fragile.

apps/web/screens/Organization/Syslumenn/OperatingLicenses.tsx
```diff
@@ -536,7 +541,15 @@ const OperatingLicenses: Screen<OperatingLicensesProps> = ({
...
 
-OperatingLicenses.getInitialProps = async ({ apolloClient, locale }) => {
+OperatingLicenses.getInitialProps = async ({
+  apolloClient,
+  locale,
+  pathname,
+}) => {
+  const slug = pathname.split('/')?.[locale !== 'is' ? 3 : 2] ?? 'syslumenn'
+  const subSlug =
+    pathname.split('/')?.[locale !== 'is' ? 4 : 3] ?? 'rekstrarleyfi'
```
kristofer-lp8f, line 551: Same fragile indexing

kristofer-lp8f (commented):
Aside from possibly-fragile indexing, this looks good. Thank you for nice l10n 😄 

Are you 💯  sure the pathname indexing will *never* break?

apps/web/screens/Organization/Syslumenn/Homestay.tsx
RunarVestmann: If we so happen to index out of the array then javascript will return undefined and the subSlug will become 'heimagisting'. So I believe that this is stable even though it looks sketchy. Perhaps moving this into a utility function would allow us to revisit this at a later point if the url structure changes in the future

RunarVestmann: I moved this into a util function, I don't see a better way at this moment how to make this more robust, this was previously hard coded in which didn't work since the english and icelandic slugs are different.

kristofer-lp8f: I think your solution is good. At least this way we only need to change the url lookup once. Is this overengineering? I like this better than the alternative 🤷

kristofer-lp8f (approved):
LGTM 👍

siggifv (approved):

PR #6897, opened 2022-03-16
feat(nest-config): Add support for optional configs

saevarma:
# Add support for optional configs

## What

This adds support for optional configs. So module can register config type to be optional, and handle it using the `isConfigured` property.

## Why

This was inspired when I was going to add token exchange for the `NationalRegistryClientModule` but other apps like Air Discount Scheme so changing the `NationalRegistryClientModule`  to require client credentials in config would have broken ADS. 

## Screenshots / Gifs

## Checklist:


eirikurn (approved):
Nice!

PR #6901, opened 2022-03-17
feat(api): assets service error handling

thordurhhh:
## What

Add proper error handling to api assets

## Why

Now when the service should return for example a 403 or 403, it will always return a 500. This fixes that.

## Checklist:


libs/api/domains/assets/src/lib/api-domains-assets.service.ts
```diff
@@ -36,127 +36,133 @@ export class AssetsXRoadService {
...
+              municipality: item.sjalfgefidStadfang?.sveitarfelagBirting,
+              postNumber: item.sjalfgefidStadfang?.postnumer,
+              locationNumber: item.sjalfgefidStadfang?.stadfanganumer,
+            },
+          })),
+        }
       }
+    } catch (e) {
+      this.handleError(e)
     }
```
darrikonn, line 65: return null by default as well ... since there isn't an explicit return statement in this function.

libs/api/domains/assets/src/lib/api-domains-assets.service.ts
```diff
@@ -36,127 +36,133 @@ export class AssetsXRoadService {
...
+              display: item.sjalfgefidStadfang?.birting,
+              propertyNumber: item.sjalfgefidStadfang?.landeignarnumer,
+              municipality: item.sjalfgefidStadfang?.sveitarfelagBirting,
+              postNumber: item.sjalfgefidStadfang?.postnumer,
+              locationNumber: item.sjalfgefidStadfang?.stadfanganumer,
+            },
+          })),
+        }
       }
+    } catch (e) {
```
darrikonn, line 63: Is this catching 4xx error as well?

They might rather return `null` or `[]` since we're working inside a graphql context.

darrikonn (commented):
My explicit return comment goes for the other methods as well

libs/api/domains/assets/src/lib/api-domains-assets.service.ts
```diff
@@ -26,6 +27,13 @@ export class AssetsXRoadService {
     )
   }
 
+  private handle403(error: FetchError) {
+    if (error.status === 403) {
```
darrikonn, line 31: what about 404?

darrikonn (approved):

libs/api/domains/assets/src/lib/api-domains-assets.service.ts
thordurhhh: Yes, should be both. 🙈

disaerna (approved):
👍

PR #6920, opened 2022-03-18
chore(judicial-system): Delete case timing and animations

oddsson:
# Delete case timing and animations

https://app.asana.com/0/1199153462262248/1201955349331115/f

## What

Orchestrate the delete case animations so that we first delete a case, then run animations and then remove the case from the list.  

## Why

If there is a delay in deleting a case, it's not clear what's happening.

### Before 

https://user-images.githubusercontent.com/3789875/159015794-68578c9a-3894-415e-bc36-3d59fb3c5605.mov

### After

https://user-images.githubusercontent.com/3789875/159015586-afa19b4a-916f-45fe-9bd2-1b4608cce520.mov

## Checklist:


apps/judicial-system/web/src/routes/Shared/Requests/ActiveRequests.tsx
```diff
@@ -194,171 +216,191 @@ const ActiveRequests: React.FC<Props> = (props) => {
...
+                      c.state === CaseState.DRAFT ||
+                      c.state === CaseState.SUBMITTED ||
+                      c.state === CaseState.RECEIVED) && (
+                      <Box
+                        data-testid="deleteCase"
+                        component="button"
+                        aria-label="Viltu afturkalla kröfu?"
+                        className={styles.deleteButton}
+                        onClick={async (evt) => {
+                          evt.stopPropagation()
```
Arnarkari93, line 353: It would be nice to move onClick functions to use `React.useCallback`.

apps/judicial-system/web/src/routes/Shared/Requests/ActiveRequests.tsx
```diff
@@ -194,171 +216,191 @@ const ActiveRequests: React.FC<Props> = (props) => {
...
+                      </Box>
+                    )}
+                </td>
+                <td className={cn(styles.deleteButtonContainer, styles.td)}>
+                  <Button
+                    colorScheme="destructive"
+                    size="small"
+                    loading={isDeletingCase}
+                    onClick={async (evt) => {
+                      if (onDeleteCase && setActiveCases) {
```
Arnarkari93, line 376: Also this one, move to `React.useCallback`

Arnarkari93 (approved):
LGMT, I don't really understand where the animation is configured. Are we just using the default values from AnimatePresence?

I also have a minor nitpick regarding `useCallback`

PR #6924, opened 2022-03-18
chore(judicial-system): Use NextJS' Head component for page headers

oddsson:
# Use NextJS' Head component for page headers

https://app.asana.com/0/1199153462262248/1201934486933573/f

## What

Use NextJS' Head component for page headers

## Why

We were using useEffects for this, which is not best practice given that NextJS has a standard way to do this. Also, we moved all titles into CF.

## Checklist:


gudjong (approved):
LGTM

PR #6927, opened 2022-03-21
refactor(core): Add token exchange support in NationalRegistryClientModule

saevarma:
https://www.notion.so/Perform-token-exchange-when-calling-person-api-from-skra-is-4ba9f2f0d1ca4d4bb221e92b113e3242

## What

Update `NationalRegistryClientModule` to use the new optional config feature to allow for token exchange configuration.

## Why

When communicating with 3rd party we want to use as small access token as possible. We also don't want to give our clients access to scope to other organisation, that should be done via token exchange.

This also allows for using client credential were allowed to save the overhead of the token exchange.
Like when doing person name lookup in Þjóðskrá, it can be done via client credentials rather than token exchange.

## Checklist:

### DIFF
--- a/apps/financial-aid/web-osk/src/components/Status/Rejected.tsx
+++ b/apps/financial-aid/web-osk/src/components/Status/Rejected.tsx
@@ -26,29 +26,47 @@ const Rejected = ({
   return (
     <>
       <Text as="h2" variant="h3" color="red400" marginBottom={[4, 4, 5]}>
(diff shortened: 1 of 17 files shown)


eirikurn (commented):
LGTM after offline fixes.

eirikurn (approved):

valurefugl (approved):

PR #6929, opened 2022-03-21
chore(download-service): Audit PDF requests

eirikurn:
## What

Add audit records for download-service requests.

## Why

So we know which files were actually requested.

## Checklist:


disaerna (approved):
Nice work 👍

PR #6932, opened 2022-03-21
refactor(auth-public-api): Change validTo to be required for create delegation

saevarma:
https://www.notion.so/Update-auth-public-api-to-require-validTo-property-for-delegations-1bd831c467b147fd80b606e72f840cb6

## What

- Make `validTo` property required to stop supporting infinite delegations.
- Fix Typescript error in `getClientDelegationInfo()`

## Why

Before launching the delegations it was decided that we should require validTo to be set.
The UI will add a default value, and the backend will return a 400 Bad Request response if validTo is not set.

## Checklist:


libs/auth-api-lib/src/lib/entities/dto/delegation.dto.ts
```diff
@@ -6,6 +6,7 @@ import {
   IsArray,
   IsDateString,
   ValidateNested,
+  IsDefined,
```
eirikurn, line 9: Used?

eirikurn (commented):
LGTM. What about the API domain?

eirikurn (approved):

PR #6946, opened 2022-03-22
feat(judicial-system): S3 Iteration

gudjong:
# S3 Iteration

https://app.asana.com/0/1199153462262248/1201777996978574/f

## What

- No longer removes case files from S3 when they have been uploaded to court.
- A case file can be viewed in the client until the case has been completed and the case file uploaded to court.

## Why

Usability improvements.

## Checklist:


apps/judicial-system/backend/src/app/modules/file/file.service.ts
```diff
@@ -220,22 +215,22 @@ export class FileService {
       courtCaseNumber,
     )
 
-    const documentId = await this.throttle
+    await this.throttle
 
-    const s3Key = file.key
-
-    const [nrOfRowsUpdated] = await this.fileModel.update(
-      { state: CaseFileState.STORED_IN_COURT, key: documentId },
```
Arnarkari93, line 228: So are we saying that we keep the same `case_file.key` after uploading to court? No need to get the `documentId` anymore?

apps/judicial-system/web/src/components/CaseFileList/CaseFileList.tsx
```diff
@@ -44,17 +46,19 @@ const CaseFileList: React.FC<Props> = (props) => {
               file={file as TCaseFile}
               showFileSize={true}
               defaultBackgroundColor={
-                file.state === CaseFileState.BOKEN_LINK ? 'dark100' : 'blue100'
+                file.status === 'broken' ? 'dark100' : 'blue100'
```
Arnarkari93, line 49: I was a bit confused reading this. First I read `file.status` as `file.state` and I did not know where this 'broken' was defined. Then I figured out these were two different properties of the file.

I want to suggest that we rename the `file.state` to `file.storage` with the options `IN_COURT | IN_RVG | DELETED`
we can keep the `file.status` as it is.

Just a suggestion 🤓

apps/judicial-system/backend/src/app/modules/file/file.service.ts
gudjong: Yes, we have never used the documentId we get from Auður, so I decided to just drop it.

Arnarkari93 (approved):
LGTM

apps/judicial-system/web/src/components/CaseFileList/CaseFileList.tsx
gudjong: I agree, this is somewhat confusing. However, renaming the `state` field is a rather big change, it comes all the way from the database. On the other hand `status` is defined in the client and could be easily changed to something different, for instance `uploadStatus`. I will create a task in the Asana inbox for refactoring this.

PR #6963, opened 2022-03-24
feat(fa): Submit spouse application

davidpxn:
# ... 👆

https://app.asana.com/0/1201983411973883/1201995298577462

## What

- Create new application in Veita backend when spouse submits. 
- Make changes to the "create" endpoint so it accepts data about the spouse. Also needed to make the endpoint accept nationalId of the applicant, because before it was using the nationalId of the current user that would be the spouse nationalId when the spouse submits - we want the applicant nationalId to be associated with the new application created.
- Make sure form comments are submitted

## Why

So the application is accessible in Veita.

## Checklist:


libs/application/templates/financial-aid/src/fields/Summary/SpouseSummaryForm.tsx
```diff
@@ -52,11 +84,21 @@ const SummaryForm = ({ application, goToScreen }: FAFieldBaseProps) => {
...
+
+      <Box
+        className={cn(styles.errorMessage, {
+          [`${styles.showErrorMessage}`]: formError,
+        })}
+      >
+        <Text color="red600" fontWeight="semiBold" variant="small">
+          {formatMessage(m.summaryForm.general.errorMessage)}
+        </Text>
+      </Box>
```
MargretFinnboga, line 99: this is the same as in summary, should this maybe be a component?

libs/application/templates/financial-aid/src/lib/FinancialAidTemplate.ts
```diff
@@ -137,14 +137,26 @@ const FinancialAidTemplate: ApplicationTemplate<
...
+              id: Roles.APPLICANT,
+              formLoader: () =>
+                import('../forms/Submitted').then((module) =>
+                  Promise.resolve(module.Submitted),
+                ),
+              // TODO: Limit this
+              read: 'all',
+            },
+            {
+              id: Roles.SPOUSE,
```
MargretFinnboga, line 159: was this necessary, did it not work like Roles.APPLICANT || Roles.SPOUSE? Because we are redirecting to the same page?

MargretFinnboga (commented):
Two thoughts 🤔

libs/application/templates/financial-aid/src/lib/FinancialAidTemplate.ts
davidpxn: Yeah this was necessary, kinda annoying. It didn´t know where to redirect the spouse otherwise. Also saw other projects doing it like this (children-residence-change):
image

libs/application/templates/financial-aid/src/fields/Summary/SummaryForm.tsx
```diff
@@ -21,19 +21,27 @@ import { Routes } from '../../lib/constants'
...
 import { hasSpouse } from '../../lib/utils'
-import { FormInfo, SummaryComment, UserInfo, ContactInfo, Files } from './index'
+import {
+  FormInfo,
+  SummaryComment,
+  UserInfo,
+  ContactInfo,
+  Files,
+  SummaryError,
+} from './index'
```
MargretFinnboga, line 33: nice 😎

MargretFinnboga (approved):
beautiful

PR #6985, opened 2022-03-25
fix(fa): Adding service to decouple the dependency on api-domains for National Registry

saevarma:
## What

Adding service class that uses the `NationalRegistryClientModule` in financial-aid instead of using the `api/domains/nationalRegistryXroadModule`. 

## Why

After the `api/domains/nationalRegistryXroadModule` was refactored to use token exchange or client credentials to communicate with Þjóðskrá, it broke the Financial Aid GraphQL communications with Þjóðskrá.

## Screenshots / Gifs

## Checklist:


MargretFinnboga (approved):
Tested and works 👏🏼 🤩

PR #6992, opened 2022-03-28
fix(fa): Fixing prereq & margin & mismatch in db

MargretFinnboga:
# ...

https://app.asana.com/0/1202024607727436/1202025305621757

## What

 - Municipality check updated 
 - Margin not right in summary 
 - Temp fix in db

## Why

- Needed  & requested

## Checklist:

### DIFF
--- a/libs/application/templates/financial-aid/src/lib/FinancialAidTemplate.ts
+++ b/libs/application/templates/financial-aid/src/lib/FinancialAidTemplate.ts
@@ -68,8 +68,10 @@ const FinancialAidTemplate: ApplicationTemplate<
         on: {
           SUBMIT: [
             //TODO check if works when national registry works
-            { target: ApplicationStates.DRAFT, cond: isMuncipalityRegistered },
-            { target: ApplicationStates.MUNCIPALITYNOTREGISTERED },
+            {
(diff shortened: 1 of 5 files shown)


libs/application/templates/financial-aid/src/lib/FinancialAidTemplate.ts
```diff
@@ -68,8 +68,10 @@ const FinancialAidTemplate: ApplicationTemplate<
         on: {
           SUBMIT: [
             //TODO check if works when national registry works
-            { target: ApplicationStates.DRAFT, cond: isMuncipalityRegistered },
-            { target: ApplicationStates.MUNCIPALITYNOTREGISTERED },
+            {
+              target: ApplicationStates.MUNCIPALITYNOTREGISTERED,
+              cond: isMuncipalityRegistered,
```
davidpxn, line 73: Shouldn´t the condition function be named isMuncipalityNotRegistered or something like that?
isMunicipalityUnregistered?

davidpxn (changes requested):
Ein pæling 👀

davidpxn (approved):

PR #7008, opened 2022-03-29
fix(service-portal): audit + minor fixes for Access control

disaerna:
# Access control - fixes

## What

- Add more detailed audit logs
- Revert button size
- Fix error check

## Why

Needed to remove feature flag ! 

## Checklist:


thordurhhh (approved):
finance resolver LGTM 👍

libs/api/domains/auth/src/lib/resolvers/delegation.resolver.ts
```diff
@@ -74,6 +76,13 @@ export class DelegationResolver {
       })
     }
 
+    this.auditService.audit({
+      auth: user,
+      namespace: '@island.is/api/auth',
+      action: 'createDelegation',
+      resources: [input.toNationalId],
+    })
+
```
saevarma, line 85: This is audit logged in the service so we don't need it in the resolver.

libs/api/domains/finance/src/lib/api-domains-finance.resolver.ts
```diff
@@ -108,22 +108,37 @@ export class FinanceResolver {
...
       user,
     )
+
+    this.auditService.audit({
+      auth: user,
+      namespace: '@island.is/api/finance',
+      action: 'getDocumentList',
+      resources: [input.listPath, input.dayFrom, input.dayTo],
+    })
+
```
saevarma, line 118: What is the reason to move from the decorator?
The decorator can be updated to log the resources, it logs the authentication be default and uses the namespace from the decorator on the controller.
I personally prefer the decorator over the service when I can.

eirikurn: Should add these as meta properties instead. The resources field is designed for specific resource ids which you want to query all audit entries for.

saevarma (approved):
LGTM

PR #7029, opened 2022-03-30
fix(application-system): Store a nonce token for assignments

obmagnusson:
# Store a nonce token for assignment

## What

Make assign tokens single use . 

## Why
So the same token cannot be used multiple times.

## Screenshots / Gifs

## Checklist:


libs/application/api/core/src/lib/application/application.service.ts
```diff
@@ -199,6 +202,21 @@ export class ApplicationService {
     return { numberOfAffectedRows, updatedApplication }
   }
 
+  async removeNonce(application: Application, assignNonce: string) {
```
eirikurn, line 205: Could also have a mirror function `addNonce` instead of extending `update`.

apps/application-system/api/src/app/modules/application/application.controller.ts
```diff
@@ -326,14 +326,19 @@ export class ApplicationController {
       return existingApplication
     }
 
+    if (!existingApplication.assignNonces.includes(decodedToken.nonce)) {
```
eirikurn, line 329: Might want to provide backwards compatibility for tokens which don't have a nonce. Otherwise some applications may become stuck.

Ideally the old unsecure tokens would expire pretty soon but unfortunately the parental leave and accident notification assignments are valid for 6 months ... For parental leaves I think the thinking was that the application may not be submitted for some time if the baby is still far away.

Not sure if we should allow the unsecure tokens to be used for the full 6 months or just for some weeks or something. Note that `decodedToken` has an `iat` claim which we can use to override the expiry (`new Date(decodedToken.iat * 1000)`).

apps/application-system/api/src/app/modules/application/application.controller.ts
```diff
@@ -326,14 +326,19 @@ export class ApplicationController {
...
+      throw new NotFoundException('Token no longer usable.')
```
eirikurn, line 330: Should create new API tests around this functionality. 😊

eirikurn (changes requested):
Looks good, like we discussed. 👍🎉

apps/application-system/api/src/app/modules/application/application.controller.ts
obmagnusson: Ahhh good point

darrikonn: Not sure if we should allow them at all.
Whatever we do, the end user is bound to experience bad UX, wether it's disabling the tokens for some weeks or immediately. Might be good to raise a custom error when the token is valid, but it's from the previous logic.
Explain that "The token has expired, and the applicant needs to resend the application" <-- we probably have to do something like this at some point, no matter what approach we take.
We however might reduce the number of support tickets related to this by enabling the older tokens for some weeks.

But this kind of comes down to `Security` vs. `UX`.

obmagnusson: I added support for the legacy tokens and reduced the validity to 6 weeks as a suggestion. 
At least with this we are in a better place security wise than before.

Thorrik58 (approved):

eirikurn (approved):
Great.

Have some additional comments which would be nice to fix but nothing that blocks the release.

apps/application-system/api/src/app/modules/application/application.controller.ts
```diff
@@ -674,7 +684,7 @@ export class ApplicationController {
     const onExitStateAction = helper.getOnExitStateAPIAction(application.state)
     const status = helper.getApplicationStatus()
     let updatedApplication: BaseApplication = application
-
+    await this.applicationService.clearNonces(updatedApplication.id)
```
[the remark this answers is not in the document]
obmagnusson: Good point.  Ill do that

PR #7030, opened 2022-03-30
feat(fa): Allow all roles to have multiple municipalities 

MargretFinnboga:
# Make veita work when logged in user has multiple municipalities

https://app.asana.com/0/1202024607727436/1201234405372285

## What

- Create end point that retrieves all info for all the municipality the current staff has 
- Keep all municipalities in context as array 
- Created application municipality state when looking at application profile for right amounts 

## Why

- Feature that was asked for 

## Checklist:


apps/financial-aid/backend/src/app/modules/municipality/municipality.service.ts
```diff
@@ -54,6 +54,40 @@ export class MunicipalityService {
     })
   }
 
+  async findByMunicipalityIds(
+    staffNationalId: string,
+  ): Promise<MunicipalityModel[]> {
+    const currentMuncipalities = await this.staffService.findByNationalId(
```
davidpxn, line 60: Should this variable not be named staff or something like that?

davidpxn (changes requested):
One small comment

davidpxn (approved):

PR #7042, opened 2022-03-31
feat(judicial-system): Add CaseType ADMISSION_TO_FACILITY

Arnarkari93:
[Asana]()
## What

- Add case type for 'admission to appropriate facility'

## Why

- We are adding a new restriction case type

## Checklist:


apps/judicial-system/backend/src/app/modules/court/court.service.ts
```diff
@@ -8,8 +8,10 @@ import type { CaseType, User } from '@island.is/judicial-system/types'
 import { now } from '../../factories'
 import { EventService } from '../event'
 
+type SubTypes = { [c in CaseType]: string | [string, string] }
```
gudjong, line 11: Nice

apps/judicial-system/backend/src/app/modules/court/court.service.ts
```diff
@@ -19,6 +21,10 @@ export const subTypes = {
   // 'Framsalsmál',
   // 'Frestur',
   CUSTODY: ['Gæsluvarðhald', 'Framlenging gæsluvarðhalds'],
+  ADMISSION_TO_FACILITY: [
```
gudjong, line 24: Looks like this subtype does not exist in Auður.

The existing subtypes are all listed in this file, some in comments as they are not yet part of RVG.

If I am correct, the value must be "Annað" and an extension is not possible.

apps/judicial-system/backend/src/app/modules/court/test/createCourtCase.spec.ts
```diff
@@ -153,6 +154,7 @@ describe('CourtService - Create court case', () => {
     type
     ${CaseType.CUSTODY}
     ${CaseType.TRAVEL_BAN}
+    ${CaseType.ADMISSION_TO_FACILITY}
```
gudjong, line 157: Not sure if this is possible.

libs/judicial-system/types/src/lib/case.ts
```diff
@@ -14,6 +14,7 @@ export enum CaseOrigin {
 export enum CaseType {
   CUSTODY = 'CUSTODY',
   TRAVEL_BAN = 'TRAVEL_BAN',
+  ADMISSION_TO_FACILITY = 'ADMISSION_TO_FACILITY',
```
gudjong, line 17: The const caseTypes in formatters.ts in the formatters lib needs to be updated as well.

gudjong (changes requested):
Two issues:
- The new type probably does not exist in Auður, which means that we need to map it to the type Annað when creating a court case.
- The caseTypes constant in the formatter library must be updated to include the new subtype.

apps/judicial-system/backend/src/app/modules/court/test/createCourtCase.spec.ts
Arnarkari93: It should be possible according to Guðný. Are there some guards clauses that check the `CaseType`?

apps/judicial-system/backend/src/app/modules/court/court.service.ts
Arnarkari93: Should I change this to `ADMISSION_TO_FACILITY: ['Annað', 'Annað']` since this an extendable case?

gudjong: Not yet. First we should make sure it does not exist the Auður xRoad API.

gudjong: Ef við þurfum að nota Annað þá mundum við nota `ADMISSION_TO_FACILITY: 'Annað'`.

gudjong (approved):

PR #7044, opened 2022-03-31
feat(judicial-system): Archive Old Cases

gudjong:
# Archive Old Cases

https://app.asana.com/0/1199153462262248/1201714251473992/f

## What

- Archives old cases 90 days after ruling or end of custody.
- Archives deleted cases.
- Encrypts selected database properties in a json object when archiving a case.
- Stores the encrypted json object in a new table `case_archive`.
- Clears the encrypted database properties.
- Addresses some linting and naming issues.

## Why

Security reasons - sensitive information should not be accessible longer than necessary.

## Checklist:


pshomov (approved):
LGTM

apps/judicial-system/backend/src/app/modules/case/filters/case.filters.ts
```diff
@@ -151,6 +153,70 @@ function isHightenedSecurityCaseHiddenFromUser(
...
+            decision: [
+              CaseDecision.ACCEPTING,
+              CaseDecision.ACCEPTING_PARTIALLY,
+            ],
+          },
+        ],
+      }
+}
+
+export const oldFilter = {
```
Arnarkari93, line 182: Nice

apps/judicial-system/backend/src/app/modules/case/models/case.model.ts
```diff
@@ -863,4 +863,14 @@ export class Case extends Model<Case> {
   })
   @ApiProperty()
   caseResentExplanation?: string
+
+  /**********
+   * Indicates whether the case has been archived - optional
+   **********/
+  @Column({
+    type: DataType.BOOLEAN,
+    allowNull: true,
```
Arnarkari93, line 872: Should we not allow a null value here?
```suggestion
    allowNull: false,
```

apps/judicial-system/backend/src/app/modules/case/test/archive.spec.ts
```diff
@@ -0,0 +1,403 @@
...
+          [{ model: Defendant, as: 'defendants' }, 'created', 'ASC'],
+          [{ model: CaseFile, as: 'caseFiles' }, 'created', 'ASC'],
+        ],
+        where: {
+          isArchived: false,
+          [Op.or]: [
+            { state: CaseState.DELETED },
+            {
+              [Op.or]: [
+                {
```
Arnarkari93, line 123: I don't really see the value in checking if `caseModel.fineOne` has been called with the right parameters. Can we test the return value? 
If we are not able to test the return value, can we reuse the `oldFilter` here?

apps/judicial-system/backend/src/app/modules/file/file.service.ts
```diff
@@ -233,4 +235,33 @@ export class FileService {
...
+    if (numberOfAffectedRows > 1) {
+      // Tolerate failure, but log error
+      this.logger.error(
+        `Unexpected number of rows (${numberOfAffectedRows}) affected when updating file ${fileId} of case ${caseId}`,
+      )
+    } else if (numberOfAffectedRows < 1) {
+      throw new InternalServerErrorException(
+        `Could not update file ${fileId} of case ${caseId}`,
+      )
+    }
```
Arnarkari93, line 265: Do we want to return `{ success }` here like the other methods in the service?

apps/judicial-system/scheduler/src/app/app.config.ts
```diff
@@ -0,0 +1,13 @@
...
+
+export const appModuleConfig = defineConfig({
+  name: 'AppModule',
+  load: (env) => ({
+    backendAccessToken: env.required(
+      'BACKEND_ACCESS_TOKEN',
+      'secret-backend-api-token',
+    ),
+    backendUrl: env.required('BACKEND_URL', 'http://localhost:3344'),
+    timeToLiveMinutes: env.requiredJSON('TIME_TO_Live_MINUTES', 1),
```
Arnarkari93, line 11: Should this be all caps?
```suggestion
    timeToLiveMinutes: env.requiredJSON('TIME_TO_LIVE_MINUTES', 1),
```

Arnarkari93 (commented):
Great job! I just have a couple of questions.

apps/judicial-system/backend/src/app/modules/case/models/case.model.ts
gudjong: Agreed, that is in line with the database migration.

apps/judicial-system/backend/src/app/modules/case/test/archive.spec.ts
gudjong: I am mocking the return value and that is the reason I test the call parameters. We have no way of testing the return value unless we call a running database and that involves a pretty hefty test setup cost.

Basically, I am saying: If our code makes the correct call, then I trust the database to return the expected value.

Using oldFilter in the test is a good idea.

apps/judicial-system/backend/src/app/modules/file/file.service.ts
gudjong: Thank you for pointing out the discrepancy. Actually, there are other methods that do not return `{ success }`, but all the other methods return something. The reason I did not do that here is that this method's return value will not be passed to the client and is not used by the caller (in this case the `archive` method in the `CaseService`. However, to stick to the patterned used elsewhere on the server, this method should return the updated case file. I have made that change.

apps/judicial-system/backend/src/app/modules/case/case.controller.ts
```diff
@@ -526,4 +522,16 @@ export class CaseController {
...
   }
+
+  @UseGuards(TokenGuard)
+  @Post('internal/cases/archive')
+  @ApiOkResponse({
+    type: ArchiveResponse,
+    description: 'Archives a single case if any case is ready to be archived',
+  })
+  archive(): Promise<ArchiveResponse> {
+    this.logger.debug('Archiving a case')
```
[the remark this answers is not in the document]
Arnarkari93: Sounds good!

apps/judicial-system/backend/src/app/modules/case/test/archive.spec.ts
Arnarkari93: I see, then this makes sense.

Arnarkari93 (approved):
🕵️‍♀️ LGTM

PR #7045, opened 2022-03-31
refactor(auth-react): CheckIdpSession refactor

saevarma:
https://www.notion.so/Fix-session-info-CSP-d23e2dd03672407b99cabf2ef9147cd9

## What

- Update the `postMessage` payload to include the type expected by the IDS.
- Clean up on the event registration to handle message events.
- Use `signOut` instead of `signIn` when expiration is triggered. Because the session is expiring on the IDS we need to do a `signOut` in the ServicePortal and ApplicationSystemForm to destroy the authentication state. Otherwise we see a flash of a logged in page before getting a redirect to the IDS login page. This will loose the location of the user where it was when the session expired. But does it even make sense to try to send him back to some previous location if the session expired due to inactivity? I don't think so.

## Why

For improvements on the session expiration monitoring.

## Screenshots / Gifs

## Checklist:


libs/auth/react/src/lib/Authenticator/CheckIdpSession.tsx
```diff
@@ -73,42 +72,55 @@ export const CheckIdpSession = () => {
...
+  )
 
-  window.addEventListener(messageEventName, messageHandler)
+  useEffect(() => {
+    window.addEventListener(messageEventName, messageHandler)
+
+    return () => {
+      window.removeEventListener(messageEventName, messageHandler)
+    }
+  }, [messageHandler, iframeChecksum])
```
eirikurn, line 123: This effect does not depend on `iframeChecksum`.

libs/auth/react/src/lib/Authenticator/CheckIdpSession.tsx
```diff
@@ -18,45 +20,42 @@ export interface SessionInfo {
...
-    )
+    if (authSettings.authority) {
+      event?.currentTarget?.contentWindow?.postMessage(
+        { type: 'SessionInfo' } as SessionInfoMessage,
+        authSettings.authority,
+      )
+    }
   }
 
   const checkIdpSessionIframe = (
```
eirikurn, line 61: Inline this in the return statement to make the component easier to read?

libs/auth/react/src/lib/Authenticator/CheckIdpSession.tsx
```diff
@@ -18,45 +20,42 @@ export interface SessionInfo {
...
+  const signOut = useCallback(
+    async function signOut() {
       dispatch({
-        type: ActionType.SIGNIN_START,
+        type: ActionType.LOGGING_OUT,
       })
-      return userManager.signinRedirect({
-        state: getReturnUrl(history, authSettings),
+      await userManager.signoutRedirect({
+        post_logout_redirect_uri: authSettings.baseUrl,
```
eirikurn, line 46: Can we somehow clear the user session and refresh the page? I.e. instead of triggering signoutRedirect? I prefer coming back to where I was...

eirikurn: Note that the baseUrl may not currently be a valid post redirect uri. The normal signout takes the user to the island.is front page.

libs/auth/react/src/lib/Authenticator/CheckIdpSession.tsx
```diff
@@ -73,42 +72,55 @@ export const CheckIdpSession = () => {
...
-
-        if (timeout <= 0) {
-          // The session is expired but for some reason the `isExpired` was not correctly set
-          return signIn()
-        }
+  const messageHandler = useCallback(
+    async ({ data, origin }: MessageEvent): Promise<void> => {
+      if (origin !== authSettings.authority) {
+        return
+      }
```
eirikurn, line 79: Is there some more checks we can make here to avoid conflicts with other messages?

libs/auth/react/src/lib/Authenticator/CheckIdpSession.tsx
```diff
@@ -18,45 +20,42 @@ export interface SessionInfo {
...
-  const history = useHistory()
   const reducerInstance = useReducer(reducer, initialState)
   const [state, dispatch] = reducerInstance
   const userManager = getUserManager()
   const authSettings = getAuthSettings()
   const iframeSrc = `${authSettings.authority}${authSettings.checkSessionPath}`
   const [iframeChecksum, setIframeChecksum] = useState(0)
+  const [sessionTimeout, setSessionTimeout] = useState<ReturnType<
+    typeof setTimeout
+  > | null>(null)
```
eirikurn, line 38: Should use `useRef` for stuff like this, when we don't need/want to re-render the component.

eirikurn (commented):
Nice cleanup. Some minor comments.

libs/auth/react/src/lib/Authenticator/CheckIdpSession.tsx
saevarma: Fixed :)

libs/auth/react/src/lib/Authenticator/CheckIdpSession.tsx
saevarma: Added type property in the message with specific `SessionInfoResponse` value.

libs/auth/react/src/lib/Authenticator/CheckIdpSession.tsx
```diff
@@ -1,13 +1,22 @@
...
+} from 'react'
 
 import { getAuthSettings, getUserManager } from '../userManager'
-import { ActionType, initialState, reducer } from './Authenticator.state'
 
-import type { History } from 'history'
-import { AuthSettings } from '../AuthSettings'
+enum SessionInfoMessageType {
+  SessionInfoRequest = 'SessionInfoRequest',
+  SessionInfoReponse = 'SessionInfoResponse',
```
eirikurn, line 13: Typo :)

eirikurn (approved):

PR #7046, opened 2022-04-01
fix(fa): Status page aid amount calculation

MargretFinnboga:
# ... ☝🏼 

https://app.asana.com/0/1202024607727436/1201995298577458

## What

- When applicant is married the aid amount calculated differently from veita to osk 
- We were using national registry, which we dont retrieve when logged into status page osk 
- Added family status to my application to use to decide between individual or cohabitation 

## Why

- Necessary that osk and veita provide the same information 

## Checklist:


davidpxn (approved):
⚡️

PR #7047, opened 2022-04-01
fix(judicial-system): use correct font

Arnarkari93:
## What

Use correct font across the judicial-system project

## Checklist:


baldur-kolibri (changes requested):
Consider adding an Asana task elaborating on the reasoning for this change 😜

PR #7051, opened 2022-04-01
fix(core): Fix renewal time check in next-ids-auth

saevarma:
## What

The `getSeconds()` and `setSeconds()` only work with the seconds segments of the date, so values from 0-59.
So the previous calculations can create some unexpected behaviour, specially with short lived 5min access tokens.

## Why

This causes the token to renewed more often than it needs to be.

## Screenshots / Gifs

## Checklist:


eirikurn (approved):

MargretFinnboga (approved):
🏆

PR #7054, opened 2022-04-02
feat(ids): Add more claim integrations to auth-api, integrate company-registry in identity domain

eirikurn:
## What

* Implement more national registry claims in Auth API.
* Implement user profile claims in Auth API.
* Implement company registry claims in Auth API (missing address for now).
* Implement company registry in Identity domain (missing address for now).
* Extend fixture helpers to support creating different types of valid national ids.
* Remove old RSK API integration.

## Why

* To support more claims in IDS.
* To support up-to-date company names in delegation system.
* To get rid of old code/integrations.
* To improve the Identity implementation to support a consistent query for people and companies by nationalId.

## Checklist:


eirikurn:
Sorry about the big PR but these changes are connected and must be merged in order, which is annoying to deal with in multiple PRs ...

Might help to review the commits in isolation!

libs/auth-api-lib/src/lib/entities/dto/address.dto.ts
```diff
@@ -0,0 +1,21 @@
+import { ApiPropertyOptional } from '@nestjs/swagger'
+
+export class AddressDTO {
+  @ApiPropertyOptional({ type: String })
```
saevarma, line 4: Feel like the `type` property is redundant.
But we need to set the `nullable` property to `true`. As the `@ApiPropertyOptional()`  only sets `required: false`  and default nullable is not allowed.

libs/auth-api-lib/src/lib/services/user-profile.service.ts
```diff
@@ -1,34 +1,164 @@
...
-      return { name: result.nafn }
-    } catch (error) {
-      this.logger.error('Error in findIndividual', error)
-      throw error
+  userProfileApiWithAuth(auth: Auth) {
+    return this.userProfileApi.withMiddleware(new AuthMiddleware(auth))
+  }
+
+  async getUserProfileClaims(auth: User): Promise<UserProfileDTO> {
+    const isCompany = !!auth.nationalId.match(/^[4-7]\d{9}$/)
```
saevarma, line 34: In another places we have been using the `isCompany` check from the `kennitala` package. Wondering if we should be consistent with this check.

apps/services/auth-api/src/app/modules/user-profile/user-profile.controller.ts
```diff
@@ -16,10 +16,19 @@ import type { User } from '@island.is/auth-nest-tools'
...
+  async userProfile(@CurrentUser() user: User): Promise<UserProfileDTO> {
+    return this.userProfileService.getUserProfileClaims(user)
+  }
+
   @Scopes('@identityserver.api/authentication')
   @Get('individual')
   @ApiOkResponse()
-  async findIndividual(@CurrentUser() user: User): Promise<IndividuaInfoDTO> {
-    return this.userProfileService.findIndividual(user)
+  async individualUserProfile(
```
saevarma, line 29: We should add the `@ApiOperation({ deprecated: true })`  decorator to indicate that this endpoint is deprecated and will be removed.

saevarma (changes requested):

pshomov (approved):
LGTM

kristofer-lp8f (approved):
Maybe remove Júní from code-owners, since this PR removes files we own?
Otherwise, LGTM 👍

miamiultras (approved):

libs/auth-api-lib/src/lib/services/user-profile.service.ts
eirikurn: Good point.

I wan't this logic to support temporary ids which `kennitala` does not support until [they release 2.0](https://github.com/HermannBjorgvin/Kennitala/pull/23).

libs/auth-api-lib/src/lib/entities/dto/address.dto.ts
eirikurn: I'm pretty sure `type` is neccessary when typescript is `strict`, since then it can't reflect the "string" part of the `string | null` union.

In any case, I decided to get rid of all nulls from this API since we don't want null claims anywhere.

saevarma (approved):
LGTM

valurefugl (approved):

obmagnusson (approved):
LGTM

PR #7063, opened 2022-04-04
fix(fa): Salary breakdown in direct payments and minor fixes

MargretFinnboga:
# ... ☝🏼 

https://app.asana.com/0/1202037685216853/1202037685216822

## What

- Salary breakdown can come as a object or array, dealing with when it comes as object
- Fixing the look on content data gathering page 
- Removing hardcoded ágúst and set current month 

## Why

- Read when the salary is an object 
- requested and is a bug

## Checklist:


davidpxn (approved):
🏅

PR #7065, opened 2022-04-04
fix(ads): Patch null-values from national registry propagating to DB

kksteini:
## What
Admin overview having problems with null values in `user_info -> postalCode`.
image-20220404111314845

Problems patched with devops via
```sql
UPDATE flight                                                                            
SET user_info = jsonb_set(user_info, '{postalCode}', '0', false)
WHERE user_info -> 'postalCode'::text = 'null';
```

Now patching future null insertions with this patch.

## Checklist:


barabrian (approved):
LGTM

PR #7067, opened 2022-04-04
feat(fa): Multiple muni & update staff add on muni

MargretFinnboga:
# ... ☝🏼 

https://app.asana.com/0/1202037685216853/1201234405372285

## What

- Endpoint that retrieves get all muni 
- Removes shared hook
- Add use muni hook to veita(gets all muni) and osk(gets one muni) 
- When staff is updated send municipaliy Ids with to update that 

## Why

- Veita and osk are using muni differently now so shared muni hook is outdated 
- We want to be able to add muni to staff 

## Checklist:


apps/financial-aid/web-veita/src/components/MultiSelection/MultiSelection.css.ts
```diff
@@ -0,0 +1,12 @@
+import { style } from '@vanilla-extract/css'
+import { theme } from '@island.is/island-ui/theme'
+
+export const selectContainer = style({
```
davidpxn, line 4: Is this necessary?  If we are using these styles for a Box component then we can just send it in as a prop:
display="block"

apps/financial-aid/web-veita/src/utils/useMunicipalities.ts
```diff
@@ -0,0 +1,95 @@
...
+        livesWithParents
+        unknown
+        type
+      }
+    }
+  }
+`
+
+export const useMunicipalities = () => {
+  const storageKey = 'currentMunicipalities'
```
davidpxn, line 42: SHould this be in constants?

MargretFinnboga: Its not really used anywhere else, so I am not sure if that is necessary? 🙃

davidpxn (approved):
Very næs 💁‍♂️

PR #7080, opened 2022-04-05
fix(ids): HOTFIX Revert user profile integration

eirikurn:
## What

Revert user profile integration in auth-api.

## Why

Because it doesn't work on staging/prod and sometimes slows down auth by around 3 seconds.

## Checklist:


saevarma (approved):
😥

PR #7081, opened 2022-04-05
fix(auth-api): Handle integration errors better

eirikurn:
## What

Handle error responses better in the new IDS / Auth API integrations.

* Handle 404 responses from Company Registry, National Registry and User Profile Service.
* Handle empty National Registry 204 responses with new shared code and client service.

## Why

So we can add more user claims in IDS.

## Checklist:


saevarma (approved):
LGTM

PR #7087, opened 2022-04-05
feat(clients): Add cache support in UserProfile and CompanyRegistry clients

eirikurn:
## What

Add configurable cache for UserProfile and CompanyRegistry clients.

* CompanyRegistry cache is shared and enabled by default for company lookups, with a fairly long cache.
* UserProfile cache is private short term cache and disabled by default. Used for IDS claims.

## Why

Better performance with recurring lookups

## Checklist:


libs/clients/rsk/company-registry/src/lib/apiConfiguration.ts
```diff
@@ -0,0 +1,54 @@
...
+              redisInstance: createRedisCluster({
+                name: 'clients-company-registry',
+                nodes: config.redis.nodes,
+                ssl: config.redis.ssl,
+                noPrefix: true,
+              }),
+            }),
+            overrideCacheControl: (request: Request) => {
+              // Only cache company lookups. Not search.
+              if (request.url.match(registryEndpoint)) {
```
saevarma, line 34: FYI this could match the search path if the search term is the national Id of the company ;)

saevarma (approved):
LGTM 🙌

libs/clients/rsk/company-registry/src/lib/apiConfiguration.ts
eirikurn: Good point. I think that's fine. It's ok to cache search results as well. The goal here is to avoid filling redis up with search phrases (one letter at a time).

obmagnusson (approved):
LGTM

darrikonn (approved):
Nice

PR #7101, opened 2022-04-06
fix(fa): Fix attachments placement when printing application

davidpxn:
# ... 👆

https://app.asana.com/0/1202037685216853/1202070037203638

## What

Fix styling of the print page.

## Why

So the attachments don't overlap with the timeline component.

## Screenshots / Gifs

Before: 
image

After:
image

## Checklist:

### DIFF
--- a/apps/financial-aid/web-veita/src/routes/PrintableApplication/printableApplication.tsx
+++ b/apps/financial-aid/web-veita/src/routes/PrintableApplication/printableApplication.tsx
@@ -14,6 +14,7 @@ import {
   ApplicationProfile,
   PrintableFiles,
 } from '@island.is/financial-aid-web/veita/src/components'
+import { Box } from '@island.is/island-ui/core'
 
 interface ApplicantData {
   application: Application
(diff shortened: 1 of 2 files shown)


apps/financial-aid/web-veita/src/styles.css
```diff
@@ -212,7 +212,6 @@ input[type='number'] {
   .printableSection {
     float: left;
     width: 100vw;
-    height: 100vh;
```
MargretFinnboga, line 215: why are we removing this? 🤔

apps/financial-aid/web-veita/src/routes/PrintableApplication/printableApplication.tsx
```diff
@@ -47,9 +48,9 @@ const PrintableApplication = () => {
             isPrint={true}
           />
           {application?.files && (
-            <>
+            <Box marginTop={20}>
```
MargretFinnboga, line 51: We could add the margin when print under css print, should we maybe do that instead?

MargretFinnboga (commented):
questions 💭

apps/financial-aid/web-veita/src/routes/PrintableApplication/printableApplication.tsx
davidpxn: I decided to have it here to create a bit of a margin between the timeline and the attachments. 
So before it looked like this:
image

Now its like this: 
image

PR #7111, opened 2022-04-07
feat(fa): Many municipalites settings

MargretFinnboga:
# Allow admin to update settings for all municipalities that it has access to

https://app.asana.com/0/1202037685216853/1201234405372285

## What

- When updating muni added that returns array of muni that current staff(admin) has access to 
- order muni alphabetically 
- Added setMunicipality to admin context 
- Refactoring settings
- Added useEffect to number input

## Why

- To make updating easier for the context 
- more efficient 
- for consistency and so everything would be update 
- Much needed 
- So would we always get the newest value 

## Checklist:


davidpxn (approved):
Veryy nicee

PR #7113, opened 2022-04-07
feat(ci): Github Actions Runners on EC2

Snojo:
## What
GHA runners in kubernetes are unreliable. So we've set up a new runner system that uses lambda and EC2 to run, scale and power

## Why

## Screenshots / Gifs

## Checklist:


darrikonn (commented):
There are still bunch of jobs running on the kubernetes runner.
If that's the plan, then when do we want the kubernetes runners and when the ec2 runners?

pshomov (changes requested):
Loogs good, need a bit of clean up. What's up with all the white space changes? 🤔

pshomov:
> There are still bunch of jobs running on the kubernetes runner.
> If that's the plan, then when do we want the kubernetes runners and when the ec2 runners?

@darrikonn We are trying a new stack to run GHA runners. These are with a different labels since we want to control the experiment. In the end it will all be running on one of these stacks. The labels can be moved between the stacks so we can move both labels to the same setup if we need/want to.

pshomov (approved):
LGTM

PR #7131, opened 2022-04-08
feat(web): Organization published material page - default ordering

RunarVestmann:
# Organization published material page - default ordering

## What

* Changed the default ordering on the Organization published material page
* Only show a menu link as active if it's a direct match
* Added comments indicating that the page number is 1-based

## Why

* The default ordering used to be the ascending release date which is not what users expect
* Menu links are active if there is a prefix match which is not ideal
* A colleague recommended that I'd add comments to indicate that the page number is 1-based since that's not at all obvious

## Checklist:


alexandradilja (approved):
LGTM

siggifv (approved):

PR #7140, opened 2022-04-11
feat(web): Project page default header visual changes

RunarVestmann:
# Project page default header visual changes

## What

* Minor visual changes for the default header of Project pages

## Why

* The header's height used to be too large

### Before
image

### After
image

## Checklist:

### DIFF
--- a/apps/web/components/Project/Header/DefaultProjectHeader.tsx
+++ b/apps/web/components/Project/Header/DefaultProjectHeader.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useRef } from 'react'
+import React, { useRef } from 'react'
 import {
   Box,
   Text,
@@ -37,11 +37,11 @@ export const DefaultProjectHeader = ({
 
   const { width } = useWindowSize()
 
-  const isBelowLarge = !(width < theme.breakpoints.lg)
+  const isBelowLarge = width < theme.breakpoints.lg
 
   console.log(textRef.current?.getBoundingClientRect())
 
-  const maxImageHeight = isBelowLarge
+  const maxImageHeight = !isBelowLarge
     ? textRef.current?.getBoundingClientRect()?.height ?? undefined
     : undefined
 
--- a/apps/web/components/Project/Header/DefaultProjectHeader.tsx
+++ b/apps/web/components/Project/Header/DefaultProjectHeader.tsx
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { useEffect, useRef } from 'react'
 import {
   Box,
   Text,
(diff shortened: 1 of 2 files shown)


apps/web/components/Project/Header/DefaultProjectHeader.tsx
```diff
@@ -31,45 +33,63 @@ export const DefaultProjectHeader = ({
...
+  const textRef = useRef<HTMLDivElement | null>(null)
+
+  const { width } = useWindowSize()
+
+  const isBelowLarge = width < theme.breakpoints.lg
+
+  const maxImageHeight = !isBelowLarge
+    ? textRef.current?.getBoundingClientRect()?.height ?? undefined
+    : undefined
+
```
kksteini, line 45: Consider not using javascript for style changes.
If I'm not mistaken, computed javascript style changes can create layout thrashing.

Is it possible to do this with CSS (maybe breakpoints)?
Perhaps @barabrian is better suited to comment on this than I am.

apps/web/components/Project/Header/DefaultProjectHeader.tsx
```diff
@@ -31,45 +33,63 @@ export const DefaultProjectHeader = ({
...
   return (
     <Box className={defaultImageIsProvided ? styles.headerWrapper : undefined}>
-      <Box
+      <div
```
kksteini, line 48: Is there a reason for changing Box -> Div?

apps/web/components/Project/Header/DefaultProjectHeader.tsx
RunarVestmann: This was the only way that I could make the height of the container follow the text on the left.
There are two columns here and I want to allow the left one to control the height of both. I was having a lot of issues trying do to that in css so I went for a Javascript solution.

apps/web/components/Project/Header/DefaultProjectHeader.tsx
RunarVestmann: Perhaps not, I've changed it back to being just a Box

apps/web/components/Project/Header/DefaultProjectHeader.tsx
barabrian: If the right column only contains an image, it should be possible to position the image absolute inside the right column so it wont take any space (if thats the issue, I haven't really gone through the code only based on the screenshots). But regarding this code I would strongly suggest never to use getBoundingClientRect inside `render`, it is quite expensive performance wise. I would at least suggest wrapping it in a useCallback
ex.
```js
const [height, setHeight] = useState(0);

const measuredRef = useCallback(node => {
  if (node !== null) {
    setHeight(node.getBoundingClientRect().height);
  }
}, [])
```
Then you would use `measuredRef` instead of `textRef`

But the best solution would still be css only if possible 😄

RunarVestmann: Yep, I'll have to have another go at this. Thanks for the advice guys :)

kksteini (approved):

PR #7149, opened 2022-04-11
feat(application-system): Delete applications

obmagnusson:
# Delete Applications

## What

Allow users to delete their applications in selected states

## Why

Users should be able to delete their drafts

## Screenshots / Gifs
image
image

## Checklist:


thordurhhh (approved):
service portal lgtm 👍

apps/application-system/api/src/app/modules/application/application.controller.ts
```diff
@@ -1058,4 +1059,35 @@ export class ApplicationController {
...
+    required: true,
+    description: 'The id of the application to delete.',
+    allowEmptyValue: false,
+  })
+  async delete(
+    @Param('id', new ParseUUIDPipe()) id: string,
+    @CurrentUser() user: User,
+  ) {
+    const { nationalId } = user
+    const existingApplication = (await this.applicationAccessService.findOneByIdAndNationalId(
```
Thorrik58, line 1077: Should we try to return early here if we know the application doesnt exist for the specific user?

apps/application-system/api/src/app/modules/application/tools/applicationAccess.service.ts
```diff
@@ -20,6 +26,19 @@ export class ApplicationAccessService {
...
+  }
+
+  async canDeleteApplication(
+    application: Application,
+    nationalId: string,
+  ): Promise<boolean> {
+    const templateId = application.typeId as ApplicationTypes
+    const template = await getApplicationTemplateByTypeId(templateId)
+    const helper = new ApplicationTemplateHelper(application, template)
+    const userRole = template.mapUserToRole(nationalId, application) ?? ''
```
Thorrik58, line 39: Should we return early if the userRole cant be found e.g. returns ''?

Thorrik58 (approved):
LGTM, some minor comments that are just a question of readability.

libs/island-ui/core/src/lib/ActionCard/ActionCard.tsx
```diff
@@ -172,7 +200,10 @@ export const ActionCard: React.FC<ActionCardProps> = ({
           <Text variant="small">{date}</Text>
         </Box>
 
-        {!eyebrow && renderTag()}
+        <Inline alignY="center" space={1}>
+          {!eyebrow && renderTag()}
+          {!eyebrow && renderDelete()}
```
atligudl, line 205: This is confusing, render delete when eyebrow is not defined? But I guess that comes more down to the structure of this component

Sidenote: Rendering React jsx in this way avoides React's optimization services and should generally be avoided.

atligudl (approved):

karenbjorg (commented):
lgtm

PR #7160, opened 2022-04-13
feat(fa): Filter applications

davidpxn:
# ... 👆

https://app.asana.com/0/1202037685216853/1202088096213656

## What

- Create new filter endpoint
- Create seperate page for listing applications with filtering
- Create new component FilterPopover to control the filters

## Why

So the staff can easily find applications

## Screenshots / Gifs

image
image
image

## Checklist:


apps/financial-aid/web-veita/src/routes/ApplicationsOverview/applicationsOverviewProcessed.tsx
```diff
@@ -0,0 +1,153 @@
...
+    fetchPolicy: 'no-cache',
+    errorPolicy: 'all',
+  })
+
+  const currentNavigationItem =
+    navigationItems.find((i) => i.link === router.pathname) ||
+    navigationItems[0]
+
+  useEffect(() => {
+    getApplications({
```
Arnarkari93, line 81: I feel like we don't have one single source of truth for the state of the filers. We are using both states and getting it from the query. If the user manually changes the query in the URL it will call `getApplications` with old filter data? I'm I mistaken?

Can we get by not using state and use the router as the source of truth?

Arnarkari93 (commented):
Looks good, just one thought 🤓

apps/financial-aid/web-veita/src/routes/ApplicationsOverview/applicationsOverviewProcessed.tsx
davidpxn: If the user manually changes the query and presses enter then the page will reload and the states are set from the query. If the user changes the query but does not press enter then nothing will happen. So getApplications can't be triggered by the user in that way. 

But I agree about the source of truth. I feel like using the query to manage the state is a bit unconventional so I am going to change so that the React states are the source of truth and the query string is just there to be displayed and get the initial states. So we will no longer listen to the changes to the Router on useEffect

apps/financial-aid/web-veita/src/routes/ApplicationsOverview/applicationsOverviewProcessed.tsx
```diff
@@ -0,0 +1,171 @@
...
+      checked
+        ? filters.selectedStates.push(item)
+        : filters.selectedStates.splice(filters.selectedStates.indexOf(item), 1)
+    }
+
+    setFilters(filtersCopy)
+  }
+
+  const onFilterClear = () => {
+    setFilters({ selectedMonths: [], selectedStates: [] })
```
Arnarkari93, line 56: Consider refactoring the state to use a reducer for a single state change.

apps/financial-aid/web-veita/src/routes/ApplicationsOverview/applicationsOverviewProcessed.tsx
```diff
@@ -0,0 +1,171 @@
...
+    setCurrentPage(1)
+    filter(1, { selectedMonths: [], selectedStates: [] })
```
Arnarkari93, line 58: Please use the `useEffect` hook with dependencies to the state to trigger the filtering.

Arnarkari93 (approved):
Discussed offline, filter should only be fired manually not automatically on state change.

linavidars (approved):

PR #7161, opened 2022-04-13
feat(web): Organization published material, request-response matching

RunarVestmann:
# Organization published material, request-response matching

## What

* Matched what was searched with what got returned by comparing the hashed input with what the response gave us

## Why

* If the communication between the front- and backend suddenly lags then the UI on the page might be out of sync with what got returned so this change is to circumvent that 

## Checklist:

### DIFF
--- a/apps/web/screens/Organization/PublishedMaterial/PublishedMaterial.tsx
+++ b/apps/web/screens/Organization/PublishedMaterial/PublishedMaterial.tsx
@@ -50,6 +50,7 @@ import {
   getGenericTagGroupHierarchy,
   getInitialParameters,
   Ordering,
+  responseInputMatchesRequestInput,
 } from './utils'
 import { OrderByItem } from './components/OrderByItem'
 import { useSelect } from 'downshift'
@@ -114,6 +115,11 @@ const PublishedMaterial: Screen<PublishedMaterialProps> = ({
         sort: ordering,
       },
     },
+  })
+  const [publishedMaterialData, setPublishedMaterialData] = useState({
+    total: 0,
+    items: [],
+    input: {},
   })
 
   const { data, loading, fetchMore } = useQuery<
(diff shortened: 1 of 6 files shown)


kristofer-lp8f (commented):
A few questions:
1. Is this a problem?
2. Is it worth the extra complexity?
3. Have you taken a look at `debounce` for delaying searching?
4. Have you taken a look at invalidating previous HTTP requests?
5. Have you considered relying on timestamps instead?

This seems like over-engineering for something I see not big a deal.

RunarVestmann:
1. We are trying to prevent this from being a possible problem
2. It's not too much extra complexity since it's just one more string value to keep track of
3. I am debouncing
4. I don't see a simple way to do that since I'm using the useQuery hook
5. I could use timestamps instead of hashing the input since all I really need is a simple comparison between requests and responses

stefanhar:
> 1. We are trying to prevent this from being a possible problem (for example this was an issue for the Sýslumenn auctions where the data was not reflecting your search string since the backend wasn't answering at a consistant enough rate)
> 2. It's not too much extra complexity since it's just one more string value to keep track of
> 3. I am debouncing
> 4. I don't see a simple way to do that since I'm using the useQuery hook
> 5. I could use timestamps instead of hashing the input since all I really need is a simple comparison between requests and responses

I believe there is already handling for some race conditions within apollo client.
Might there be some other reason we are seeing a weird result here?
Does the issue persist if `fetchPolicy` is set to `network-only`?

kristofer-lp8f:
Some resources I've stumbled upon:
- https://github.com/apollographql/react-apollo/issues/3717
- https://github.com/apollographql/apollo-client/blob/main/src/core/QueryManager.ts#L1039

RunarVestmann:
The implementation works fine without this change. This is just something a colleague pointed out and we added in case there is a halt in network communication. We could either just bail on this change and I'll close the PR or I could maybe do something a bit simpler like sending a timestamp instead of hashing the request. What do you guys think?

kristofer-lp8f:
If this is a real problem or will be in the foreseeable future, I'd prefer the timestamp version. If not, I vote to close this PR.

PR #7178, opened 2022-04-19
perf(search-indexer): Cluster reduction for local development performance

kksteini:
## What

Initially, cluster behaviour set up locally in order to test cluster specific problems.
These have proven to be unnecessary and are bogging down the performance on local setups due to high memory and some processing demands.

If cluster related problems arise, this is easily reversed in testing since this cluster removal operation is purely subtractive.

## Why

Improve performance when testing elasticsearch locally.

## Checklist:


rafnarnason (approved):
lgtm

PR #7181, opened 2022-04-19
chore(core): Fix IDS client credential DX

eirikurn:
## What

Update get-secrets to return a special identity server secret for local development.

After this change, when you run `get-secrets` for a project which includes a `IDENTITY_SERVER_CLIENT_SECRET`, you'll instead get the secret for the `@island.is/clients/dev` client, which is configured as a `devFallback` in [IdsClientConfig](https://github.com/island-is/island.is/blob/main/libs/nest/config/src/lib/configurations/IdsClientConfig.ts).

For everything to work you should remove the `IDENTITY_SERVER_CLIENT_ID` variable from your `.env.secret` file or update it to match.

## Why

To improve DX and make "get-secrets" just work.

## Checklist:


pshomov (approved):
LGTM

PR #7188, opened 2022-04-20
feat(web): Feature flag support for web

RunarVestmann:
# Feature flag support for web

## What

* Added a utility function that gets a feature flag from config cat

## Why

* We want to be able to toggle whether a feature is enabled on not

## Checklist:


kristofer-lp8f (changes requested):
This looks good. Could you add a positive and negative test case? And if no too much trouble add some documentation on the usage of feature flag in Web?

RunarVestmann:
Wow, that comment went past me. I was wondering why there was no reaction on Slack :/. Anyways, where should I document the usage of feature flags on the web. Should I add comments, create a simple Notion page or both?

PR #7202, opened 2022-04-22
refactor(core): Refactor type dependencies for localisation and cms-translation

saevarma:
# Refactoring type dependencies for types from localisation and cms-translation

## What

Turned out that there was some spaghetti of dependencies for the `FormatMessage`, `MessageDescriptor`, ´StaticText` and other localisation/translation related types. They were also wrongly imported in react vs nestjs applications.

## Why

Simplify dependencies across apps. For example Judicial system was depending on `StaticText` from the `application/core library`.

## Checklist:


eirikurn (approved):

Thorrik58 (approved):

tommigitgaur (approved):
LFTM

Arnarkari93 (approved):
LGMT

thordurhhh (approved):

saevarma:
We made a lot of changes to the previous approach so I requested a re-review from @Thorrik58, @Arnarkari93 and @tommigitgaur

libs/cms-translations/src/lib/intl.service.ts
```diff
@@ -1,29 +1,29 @@
...
-): string => {
-  if (!descriptor) {
-    logger.warn('No descriptor passed to formatMessage function.')
-    return ''
-  }
-
-  if (typeof descriptor === 'string') {
-    return descriptor
-  }
+export const formatMessage = (intl: IntlShape<string>) =>
```
Arnarkari93, line 16: Do we really need to wrap the default `formatMessage` with this extra check? Are we not just making it worse to enforce code that is depending on the cms-translations to use a custom formatMessage?

Arnarkari93 (changes requested):

Arnarkari93:
I have some small changes to cms-translations in an open PR https://github.com/island-is/island.is/pull/7256. Just so you know if you get conflicts.

libs/cms-translations/src/lib/intl.service.ts
eirikurn: Good question. This check is to make the NestJS implementation consistent with the React implementation (in `@island.is/localization`). 

I think the string support was originally added to improve DX when dealing with strings/translations of different origins, eg as props in shared components. Some UIs don't need translations for some reason and instead use hard coded strings.

The `undefined` sugar was added in #6507 to simplify code like this: `message ? formatMessage(message) : undefined`.

Arnarkari93: I believe this check should instead be done in the individual project them selfs rather than trying to support all different use-cases in this library. I am mostly concerned here about changing the return value to be optionally undefined.

I just wanted to raise my concern, let me know how you want to proceed?

eirikurn: I don't have a strong opinion on this. Some devs liked the sugar enough to implement it. But I'll leave the cms-translation consistency change out of this PR at least. Worth discussing on slack someday.

Arnarkari93 (approved):

Thorrik58 (approved):
LGTM

PR #7209, opened 2022-04-22
feat(service-portal): Translation support for access control scopes

eirikurn:
## What

Implement translation support in api scope resolver/controller.

## Why

To provide better support for translations in the access control module.

https://www.notion.so/Add-translation-support-to-auth-public-api-41eb54c032f2450d87f46c43654b7781

## Checklist:


libs/api/domains/auth/src/lib/dto/apiScopes.input.ts
```diff
@@ -0,0 +1,9 @@
+import { Field, InputType } from '@nestjs/graphql'
+import { IsString } from 'class-validator'
+
+@InputType('AuthApiScopesInput')
+export class ApiScopesInput {
+  @Field(() => String)
+  @IsString()
+  lang = 'is'
```
darrikonn, line 8: should this be an enum?

FYI: `IsString` is not needed, doesn't do anything for the GraphQL schema validation

darrikonn (approved):

libs/api/domains/auth/src/lib/dto/apiScopes.input.ts
eirikurn: Ah. Good to know. Copied the input pattern from the cms resolvers. They have an enum there but not sure if this should be an enum. 

To add a new language we "just" need to update the locale picker and add content in the backend. Enums just increase the number of code changes required :)

saevarma (approved):
Nice!

PR #7230, opened 2022-04-25
feat(auth-api): Add feature flag for IDS user-profile integration

eirikurn:
## What

Add feature flag so we can toggle and test the user-profile integration in auth-api.

## Why

Because it may not currently work in production, failures may delay logins and this integration is not fully implemented yet in IDS anyways. Still would be good to test it safely in production.

## Checklist:


valurefugl (approved):

saevarma (approved):
LGTM

PR #7238, opened 2022-04-26
chore(core): Update oidc-client library

saevarma:
# Update oidc-client library to use the new oidc-client-ts

## What

The oidc-client-js library has been abandoned and there is a new successor project oidc-client-ts 

## Why

The old oidc-client-js library does a time sensitive check on the iat (issued at) claim, which depends on the clock on the user system. This can lead to users being unable to authenticate if they have more then a 5 min time skew from our IDS server.

## Checklist:


eirikurn (approved):
LGTM, one minor comment.

darrikonn (approved):

thordurhhh (approved):

PR #7239, opened 2022-04-26
fix(fa): Tax break down adding data fetched

MargretFinnboga:
# ...

https://app.asana.com/0/1202167941440767/1202037685216798

## What

- Added data fetched string to tax breakdown 
- Added fillInDictionary so if direct payments is last three months dont have to check 
- Sorted also by year

## Why

- Because we need to know when the data was fetched to go back to three months, not just current month if looking at an older application 
- More efficiency 
- Requested

## Screenshots / Gifs

Dont want the months to pile up 
Screenshot 2022-04-26 at 16 03 24

## Checklist:


davidpxn (approved):
Looks good 👩‍🎓

PR #7242, opened 2022-04-27
feat(service-portal): Block access if legal guardian & child >= 16 years

disaerna:
# Service portal - Block "Pósthólf" access 

## What

If user is acting as legal guardian and child is 16 years or older, the "Pósthólf" should be blocked.

* Add dateOfBirth to User type
* Add delegationType User type
* Add getBirthday func because kennitala package was throwing errors & tests
* Add birthday to User object in Authenticator state

## Why

Laws and rules 👮 


thordurhhh (approved):
👍

libs/auth/react/src/lib/Authenticator/Authenticator.state.ts
```diff
@@ -39,6 +40,20 @@ export const initialState: AuthReducerState = {
...
+// Add dateOfBirth Date object to user profile
+const formatUser = (payload: User): User | null => {
+  const dateOfBirth = getBirthday(payload?.profile?.nationalId)
+
+  return {
+    ...payload,
+    scopes: payload.scopes || [],
+    profile: {
+      ...payload.profile,
+      dateOfBirth: dateOfBirth,
```
saevarma, line 52:
```suggestion
      dateOfBirth,
```

saevarma (approved):
Looks good. Will be nice to clean this up in the future when the IDS will return the birthDate.

PR #7245, opened 2022-04-27
(application-system-api): Initial state onEntry

kksteini:
Allow triggering of Api Template Actions for the `meta.onEntry` directive on states marked as `initial`.

## Why

Prefetching of relevant user data which dictates an application and other use cases of a similar vein.

## Checklist:

### DIFF
--- a/apps/application-system/api/src/app/modules/application/application.controller.ts
+++ b/apps/application-system/api/src/app/modules/application/application.controller.ts
@@ -288,6 +288,41 @@ export class ApplicationController {
       resources: updatedApplication.id,
       meta: { type: application.typeId },
     })
+
+    const actionDto: BaseApplication = {
+      ...updatedApplication,
+      id: createdApplication.id,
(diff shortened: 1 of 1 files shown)


apps/application-system/api/src/app/modules/application/application.controller.ts
```diff
@@ -288,6 +288,41 @@ export class ApplicationController {
...
+        updatedApplication: withUpdatedExternalData,
+      } = await this.performActionOnApplication(
+        actionDto,
+        template,
+        user,
+        onEnterStateAction,
+      )
+
+      // TODO: decision on how to handle errors
+      // Proposal:  return the external Data anyway.
```
obmagnusson, line 319: Yes agree with this. If programmers choose this state they should be responsible in handling the errors.

obmagnusson (approved):
LGTM! Nice work!

PR #7256, opened 2022-04-27
test(j-s): refactor createTestIntl setup

Arnarkari93:
## What

- change createTestIntl setup
- refactor formatters test intl setup

## Why

- It dose not make sense to have to pass in the messages when creating `intl`. Instead we can mock the `onError` function to dissmiss error message about missing translations strings, defaulting to `defaultMessage`

## Checklist:


gudjong (approved):
LGTM

PR #7266, opened 2022-04-28
feat(j-s): support admission cases in custody restrictions

Arnarkari93:
[Asana](https://app.asana.com/0/0/1202188672896413/f)
## What

- Support admission to facility cases in custody notice pdf and court record autofill of `endOfSessionBookings`
- Move strings to contentful and spliting the format so we have one for backend and another for web.

## Why

- Support new case type
- Seperation of consern formatting the strings for different usecases but also keeping the shared things shared

## Checklist:


gudjong (approved):
Cleaner code - good change.

PR #7275, opened 2022-04-29
feat(j-s): validate court case number

Arnarkari93:
[Asana](https://app.asana.com/0/1199153462262248/1202107610712222/f)

## What

- move the validation test to a separate file
- validate the input of court case number

## Why

- Input mistakes

## Screenshots / Gifs
Screenshot 2022-04-29 at 11 43 00

## Checklist:


apps/judicial-system/web/src/utils/validate.spec.ts
```diff
@@ -0,0 +1,209 @@
...
+    ${'R-1-2019'}
+    ${'R-1/201'}
+    ${'R-1/201'}
+    ${'R-12345/2014'}
+  `(
+    'should fail if case number as wrong format $courtCaseNumber',
+    ({ courtCaseNumber }) => {
+      const result = validate(courtCaseNumber, 'court-case-number')
+      expect(result.isValid).toEqual(false)
+      expect(result.errorMessage).toEqual('Dæmi: R-1234/2022')
```
gudjong, line 206: Won't this test fail next year, because the year will change? Consider injecting the year or use some kind of date factory which can be mocked in the test.

gudjong (changes requested):

apps/judicial-system/web/src/utils/validate.spec.ts
Arnarkari93: Ofc! 🤦‍♂️

PR #7278, opened 2022-04-29
chore(swagger): specify what to import in docs

darrikonn:
## Why
Devs have been trying to import from `@nestjs/common`


kristofer-lp8f:
I would love for some more examples, and a better explanation of how elements of the `request/param/body` map to `@Api*`. The current (only) example is unclear about where `search` comes from.

saevarma (approved):
Always nice to improve docs

PR #7283, opened 2022-04-29
feat(service-portal): Add company modules

thordurhhh:
## What

Add company option to service-portal.

When logged in as procuration holder of a company, you can switch to the company a delegate. Then you will get a different view of the service-portal.

A different view meaning:

* Not the same pages in the sidebar as an individual
* Company details page available.

## Why

Allow company delegation view in the service-portal

## Screenshots / Gifs

Screenshot 2022-05-16 at 12 57 36

## Checklist:


Valur (approved):

libs/service-portal/information/src/lib/queries/getCompanyRegistryCompany.ts
```diff
@@ -0,0 +1,33 @@
+import { gql } from '@apollo/client'
+
+export const COMPANY_REGISTRY_INFORMATION = gql`
+  query companyRegistryCompanyQuery($input: RskCompanyInfoInput!) {
```
darrikonn, line 4: I know this is done almost everywhere, but we really shouldn't store queries in a central location and query everything. This takes the power away from GraphQL, and treats it like a Rest API.
Components should query the needed data. GraphQL will take care of merging those requests into a batch operation and remove redundant fields.
For example, this query is making two API requests to external data providers... every single time, regardless of if it's needed or not. This makes field resolvers obsolete, since their purpose are bypassed.
If the component/screen only needs the name and nationalId, then we skip a whole API request to another external API.
In some cases we can be saving 5 external requests using field resolvers... and if that query is a relay like connection that fetches 100 items, then it can save 500 external requests (you get the point 😄 ).
We should take this up in the next dev meeting to solve this globally for service-portal, but IMO this PR should extract the GraphQL AST to the component.

Rule of thumb, if we want to share some GraphQL ASTs, then it should be shared via fragments.

darrikonn (changes requested):
Nice work. Minor regarding the location of the GraphQL AST

libs/service-portal/information/src/lib/queries/getCompanyRegistryCompany.ts
thordurhhh: Yes, totally agree with you there.
I'll add a ticket on our board to refactor this a bit in the service-portal, and try to make sure this is done properly in the future 😃

libs/service-portal/graphql/src/lib/queries/getCompanyRegistryCompany.ts
```diff
@@ -0,0 +1,16 @@
+import { gql } from '@apollo/client'
+import { CompanyInfoFragment } from '../fragments/companyInfo'
+
+export const COMPANY_REGISTRY_INFORMATION = gql`
```
darrikonn, line 4: Better. Almost there!
Would be an improvement if this query would be moved to the `libs/service-portal/information/src/screens/Company/CompanyInfo.tsx` component. Since we don't want other components/devs to reuse this query. Other components should specifically request the `companyInfo fragment` if it's needed, since otherwise we are still introducing the same problem of making field resolvers obsolete and making unnecessary API requests when it isn't needed.

darrikonn (changes requested):

darrikonn (approved):

libs/service-portal/graphql/src/lib/queries/getCompanyRegistryCompany.ts
darrikonn: IMO keeping the GraphQL AST inside the component makes a great documentation, similar to PropTypes.
You can see what the purpose of the component is, and what it needs in order to render. Additionally it's easy to clean up the queries and saving requests since you're just comparing the component to the AST in a single file :).

thordurhhh: Yeah, putting it that way, that's a solid point. 💯

PR #7291, opened 2022-05-02
fix(company-registry): Fix client to match server response

eirikurn:
## What

* Update company registry client to return correct data from server.
* Implement client service and map server response to cleaner english-language typescript types.
* Update graphql domain to match.
* Update auth-api integration to match.

## Why

So we can implement more company functionality.

## Checklist:


Thorrik58 (approved):

libs/auth-api-lib/src/lib/services/user-profile.service.ts
```diff
@@ -119,47 +144,58 @@ export class UserProfileService {
...
     }
 
+    const country = this.formatCountry(address.country, hasCountryCode)
     return {
-      formatted: this.formattedAddressString(address),
+      formatted: this.formattedAddressString(address, country),
       streetAddress: address.streetAddress ?? undefined,
       locality: address.locality ?? undefined,
       postalCode: address.postalCode ?? undefined,
+      country,
```
saevarma, line 172: Everything can be `undefined` except country which will be `Ísland` if undefined. I'm wondering if we could get some weird edge cases by providing default value for country, and if we should instead allow country to be undefined?

saevarma: That makes sense.

saevarma (approved):

PR #7299, opened 2022-05-02
feat(fa): Tax data in Application System for Applicant and spouse

MargretFinnboga:
# ...

https://app.asana.com/0/1202167941440767/1202037685216798

What 

- Adding personalTaxreturn and directTaxPayments to api 
- Added Tax data fetch provider to retrieve 
- Added component for summary (directtaxpaymentmodal) 
- Added personal tax return to files in summary 
- Updating get family status 
- Updating tax return files form depending on what comes from tax data gathering
- Added to external data taxDataFetchSpouse and taxDataFetch
- Show only tax site if tax data fails 
- Added prerequisitesSpouse
- when creating application added tax data and checks 

Why
- To retrive the data 
- to show the tax data in summary
- family status more detail on whether the applicant is individual or not 
- for spouse flow

## Checklist:


libs/application/templates/financial-aid/src/fields/Summary/Files.tsx
```diff
@@ -21,20 +22,26 @@ interface Props {
...
+      ? incomeFiles.concat(taxFiles)
+      : incomeFiles.concat(taxFiles).concat([personalTaxReturn])
 
   return (
     <SummaryBlock editAction={() => goToScreen?.(route)}>
       <Text fontWeight="semiBold" marginBottom={1}>
         {formatMessage(summaryForm.formInfo.filesTitle)}
       </Text>
       {allFiles &&
+        allFiles.length > 0 &&
```
davidpxn, line 43: Do we need to check the length of allFiles? Could we remove this check and then if the length is 0 then the allFiles.map will just return nothing?

libs/application/templates/financial-aid/src/fields/Summary/Files.tsx
```diff
@@ -13,6 +13,7 @@ import { summaryForm } from '../../lib/messages'
 interface Props {
   goToScreen?: (id: string) => void
   route: Routes
+  personalTaxReturn: UploadFile | null
```
davidpxn: Is it possible to do this instead? 
personalTaxReturn?: UploadFile

libs/application/templates/financial-aid/src/fields/Summary/DirectTaxPaymentCell.tsx
```diff
@@ -0,0 +1,54 @@
...
+}
+
+const DirectTaxPaymentCell = ({
+  directTaxPayments,
+  hasFetchedPayments,
+  setIsModalOpen,
+}: Props) => {
+  const fetchFailed = directTaxPayments.length === 0 && !hasFetchedPayments
+
+  if (fetchFailed) {
```
davidpxn: Maybe en unnecessary variable fetchFailed. Could maybe just do
if(directTaxPayments.length === 0 && !hasFetchedPayments)

libs/application/templates/financial-aid/src/fields/Summary/DirectTaxPaymentCell.tsx
```diff
@@ -0,0 +1,54 @@
...
+      <Divider />
+      <Box
+        display="flex"
+        justifyContent="spaceBetween"
+        alignItems="flexStart"
+        paddingY={[4, 4, 5]}
+      >
+        <Box marginRight={3}>
+          <Text fontWeight="semiBold" color={'dark400'}>
+            Staðgreiðsluskrá
```
davidpxn, line 30: Move texts to messages?

libs/application/templates/financial-aid/src/fields/TaxBreakdown/TaxBreakdown.css.ts
```diff
@@ -0,0 +1,60 @@
...
+      marginBottom: theme.spacing[4],
+    },
+  },
+})
+
+globalStyle(`${tableContainer} th, ${tableContainer} td`, {
+  paddingRight: theme.spacing[3],
+  paddingLeft: theme.spacing[3],
+  paddingTop: theme.spacing[2],
+  paddingBottom: theme.spacing[2],
```
davidpxn, line 19: Same question as before ⬆️

libs/application/templates/financial-aid/src/fields/TaxBreakdown/TaxBreakdown.tsx
```diff
@@ -0,0 +1,110 @@
...
+  formatNationalId,
+  getMonth,
+} from '@island.is/financial-aid/shared/lib'
+import TaxBreakdownHeadline from './TaxBreakdownHeadline'
+import { Dictionary, groupBy } from 'lodash'
+
+import * as styles from './TaxBreakdown.css'
+
+export const taxBreakDownHeaders = [
+  'Fyrirtæki',
```
davidpxn, line 18: Move to messages?

libs/application/templates/financial-aid/src/fields/TaxBreakdown/TaxBreakdown.tsx
```diff
@@ -0,0 +1,110 @@
...
+                        formatNationalId(item.payerNationalId),
+                        `${item.totalSalary.toLocaleString('de-DE')} kr.`,
+                        `${item.personalAllowance.toLocaleString('de-DE')} kr.`,
+                        `${item.withheldAtSource.toLocaleString('de-DE')} kr.`,
+                      ]}
+                    />
+                  ) : (
+                    <TaxBreakdownItem
+                      key={`${index}-${item.month}-taxBreakdown`}
+                      items={['Engin staðgreiðsla']}
```
davidpxn: Move text to messages?

libs/application/templates/financial-aid/src/fields/DirectTaxPaymentsModal/DirectTaxPaymentModal.tsx
```diff
@@ -19,44 +19,46 @@ const DirectTaxPaymentModal = ({
...
+          <Box
+            position="relative"
+            background="white"
+            borderRadius="large"
+            paddingY={4}
+            paddingX={4}
+            className={styles.modal}
+          >
+            <Text variant="h3" marginBottom={4}>
+              Staðgreiðsluskrá
```
davidpxn, line 47: Move to messages?

libs/application/templates/financial-aid/src/fields/DirectTaxPaymentsModal/DirectTaxPaymentModal.tsx
```diff
@@ -19,44 +19,46 @@ const DirectTaxPaymentModal = ({
...
 
-    //         <Box paddingTop={4} display="flex" justifyContent="flexEnd">
-    //           <Button onClick={closeModal}>Loka</Button>
-    //         </Box>
-    //       </Box>
-    //     </Box>
-    //   )}
-    // </ModalBase>
+            <Box paddingTop={4} display="flex" justifyContent="flexEnd">
+              <Button onClick={closeModal}>Loka</Button>
```
davidpxn, line 56: Move to messages?

libs/application/templates/financial-aid/src/fields/TaxBreakdown/TaxBreakdown.tsx
```diff
@@ -93,10 +100,7 @@ const TaxBreakdown = ({ items, dateDataWasFetched }: Props) => {
                       ]}
                     />
                   ) : (
-                    <TaxBreakdownItem
-                      key={`${index}-${item.month}-taxBreakdown`}
-                      items={['Engin staðgreiðsla']}
-                    />
+                    <TaxBreakdownItem items={['Engin staðgreiðsla']} />
```
davidpxn, line 103: Move text to messages?

libs/application/templates/financial-aid/src/fields/TaxReturnFilesForm/TaxReturnFilesForm.tsx
```diff
@@ -10,12 +10,22 @@ import { getTaxFormContent } from './taxFormContent'
...
 const TaxReturnFilesForm = ({ field, application }: FAFieldBaseProps) => {
   const { formatMessage } = useIntl()
-  const { id, answers, externalData } = application
+  const { id, answers, externalData, assignees } = application
 
-  const {
-    municipalitiesDirectTaxPayments,
-    municipalitiesPersonalTaxReturn,
-  } = externalData.taxDataFetch.data
+  const isSpouse = (assignees: string[], nationalId?: string) => {
```
davidpxn, line 15: We have a similar check in mapUserToRole in FinancialAidTemplate.ts. Just wondering if we could create a function in utils.ts and reuse it for both places?

libs/application/templates/financial-aid/src/fields/Summary/SummaryForm.tsx
```diff
@@ -14,28 +13,56 @@ import * as m from '../../lib/messages'
...
+import { DirectTaxPaymentsModal } from '..'
 
 const SummaryForm = ({ application, goToScreen }: FAFieldBaseProps) => {
   const { formatMessage } = useIntl()
   const { id, answers, externalData } = application
   const summaryCommentType = SummaryCommentType.FORMCOMMENT
 
+  const [isModalOpen, setIsModalOpen] = useState(false)
+
+  const getFamilyStatus = (
```
davidpxn, line 40: There is a similar function findFamilyStatus in utils. Can we reuse that?

davidpxn (commented):
Couple of questions and suggestion 👩‍🎓. 
Also saw some commits and changes for the driving-license and parental-leave projects that I am not sure how they got into this PR 🤷‍♂️

libs/application/templates/financial-aid/src/fields/DirectTaxPaymentsModal/DirectTaxPaymentModal.css.ts
```diff
@@ -0,0 +1,49 @@
...
+export const tableItem = style({
+  borderBottom: `1px solid ${theme.color.blue200}`,
+})
+
+globalStyle(`${tableContainer} th, td`, {
+  paddingRight: theme.spacing[7],
+  paddingLeft: theme.spacing[7],
+  paddingTop: theme.spacing[3],
+  paddingBottom: theme.spacing[3],
+})
```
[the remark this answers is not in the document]
MargretFinnboga: We can, but i never remember the order of the padding 🙈  so i think this is more clearer for forgetful people like me 🤷🏼

libs/application/templates/financial-aid/src/fields/TaxBreakdown/TaxBreakdown.css.ts
MargretFinnboga: We can, but i never remember the order of the padding 🙈  so i think this is more clearer for forgetful people like me 🤷🏼

libs/application/templates/financial-aid/src/fields/TaxReturnFilesForm/TaxReturnFilesForm.tsx
MargretFinnboga: changed it made it similar, what do you think?

davidpxn: I was thinking of creating a function in utils (getUserRole or something like that) using the code from mapUserToRule in FinancialAidTemplate:

```
    if (id === application.applicant) {
      return Roles.APPLICANT
    }
    if (application.assignees.includes(id)) {
      return Roles.SPOUSE
    }
    return undefined
```

Then call this utils function inside of mapUserToRule and TaxReturnFilesForm.ts
How do you feel about that?

MargretFinnboga: I like how you think, however in the application system the mapUserToRole has to be in the FinancialAidTemplate, because there is some magic happening there and all other template do it the same way. And the function would return Roles.SPOUSE and then i would also have to check that as well, instead of doing the check if the nationalId is in assignees, I can add that it to utils, but i thought it was unnecessary since it is only used here, however i do agree if we need the check again we should move the function to utils.

davidpxn (approved):
Love it 😍

PR #7305, opened 2022-05-03
fix(dpo): Security patch for document-provider api

fjandakornid:
# Security patch for document-provider api

## What

Check if caller is last modifier of organisation so that only he/she can perform methods on organisation.

## Why

Methods were callable by all logged-in users.

## Checklist:


libs/api/domains/document-provider/src/lib/document-provider.service.ts
```diff
@@ -241,14 +235,23 @@ export class DocumentProviderService {
...
+      nationalId,
+      authorization,
+    )
+
+    if (!isLastModifier) {
+      throw new ApolloError(
+        'Forbidden. User is not last modifier of organisation.',
+        '403',
+      )
+    }
```
saevarma, line 254: As we are re-using this around this file we should move this into a helper function.

saevarma (changes requested):

saevarma (approved):
LGTM, just add the `automerge`  label

PR #7309, opened 2022-05-03
feat(fa): Populate state page with data from backend

davidpxn:
# ... 👆

https://app.asana.com/0/1202167941440767/1202189059355217

## What

- Create new endpoint in backend to get application by AS id
- Create api endpoints for AS to retrieve data about application
- Populate the UI with data from the backend

## Checklist:


apps/financial-aid/backend/src/app/modules/application/application.service.ts
```diff
@@ -218,6 +218,45 @@ export class ApplicationService {
...
+        },
+        {
+          model: DirectTaxPaymentModel,
+          as: 'directTaxPayments',
+        },
+      ],
+    })
+
+    if (application?.amount) {
+      application.setDataValue('amount', application.amount['0'])
```
MargretFinnboga, line 254: hmm this is interesting, do we need this? 🤔

apps/financial-aid/backend/src/app/modules/application/application.service.ts
```diff
@@ -218,6 +218,45 @@ export class ApplicationService {
...
+          include: [{ model: DeductionFactorsModel, as: 'deductionFactors' }],
+          separate: true,
+          order: [['created', 'DESC']],
+        },
+        {
+          model: DirectTaxPaymentModel,
+          as: 'directTaxPayments',
+        },
+      ],
+    })
```
MargretFinnboga, line 251: thinking findbyId is using the same includes, can we maybe make an one array that they both could use, so not to forget to update either 😃 
Or use the same endpoint, and have a boolean where isApplicationId?, because the endpoints are the same

libs/application/templates/financial-aid/src/dataProviders/VeitaProvider.ts
```diff
@@ -24,7 +23,9 @@ export class VeitaProvider extends BasicDataProvider {
         const returnObject =
           response.data.municipalitiesFinancialAidCurrentApplication
 
-        return Promise.resolve({ currentApplicationId: returnObject })
+        return Promise.resolve({
+          currentApplicationId: returnObject,
```
MargretFinnboga, line 27: unnecessary , here right?

libs/application/templates/financial-aid/src/lib/formatters.ts
```diff
@@ -159,28 +159,35 @@ export const getStateMessageAndColor: KeyMapping<
...
+  created?: string,
+  modified?: string,
   showSpouseStep?: boolean,
 ) => {
+  const createdFormatted = created
+    ? format(new Date(created), 'dd/MM/yyyy HH:mm')
+    : null
+  const modifiedFormatted = modified
+    ? format(new Date(modified), 'dd/MM/yyyy HH:mm')
+    : null
```
MargretFinnboga, line 171: this could be a function, if the format changes, only have to change one place 🤓

MargretFinnboga (commented):
few comments 🐑

apps/financial-aid/backend/src/app/modules/application/application.service.ts
davidpxn: Yes this is necessary. When we use "include" in the query then the included results are always an array and here we need to turn that array into an object because Amount in Application is not an array.

libs/application/templates/financial-aid/src/dataProviders/VeitaProvider.ts
davidpxn: What do you mean is unnecessary?

apps/financial-aid/backend/src/app/modules/application/application.service.ts
```diff
@@ -176,39 +212,22 @@ export class ApplicationService {
...
+
+    return application
+  }
+
+  async findByApplicationSystemId(
+    id: string,
+  ): Promise<ApplicationModel | null> {
+    const application = await this.applicationModel.findOne({
+      where: { applicationSystemId: id },
+      include: applicationInlcudeClause(false),
```
MargretFinnboga, line 230: why are you not just using the  findById endpoint, rather than making a new one?

libs/application/templates/financial-aid/src/dataProviders/VeitaProvider.ts
MargretFinnboga: I have no idea what i meant 🙈

libs/application/templates/financial-aid/src/fields/Status/SpouseStatus.tsx
```diff
@@ -11,26 +11,28 @@ import {
...
       {state === ApplicationState.DATANEEDED && <MissingFilesCard />}
 
-      {/* TODO: we might need to use the dates from Veita*/}
       <Timeline
         state={state}
-        created={application.created}
-        modified={application.modified}
+        created={currentApplication?.created ?? application.created.toString()}
+        modified={
+          currentApplication?.modified ?? application.modified.toString()
```
MargretFinnboga, line 34: instead of changing application.created and application.modified from date to string, then string to date, why not move formatDate on currentApplication?.created and currentApplication?.modified, and just change those from string to date, and then props for timeline modified and created to Date again? I think that is better right?

MargretFinnboga (commented):
Lets go 🚀

apps/financial-aid/backend/src/app/modules/application/application.service.ts
davidpxn: I did it because I thought we didn't have access to the application id. But I realised we can get the id from the response of the createApplication request.
So I have removed the unnecessary endpoint and reusing the old one.

libs/application/templates/financial-aid/src/fields/Status/SpouseStatus.tsx
MargretFinnboga: yess!

MargretFinnboga (commented):
Just fixing the timeline and then we good to go

libs/application/templates/financial-aid/src/fields/Status/ApplicantStatus.tsx
```diff
@@ -39,16 +45,26 @@ const ApplicantStatus = ({ application }: FAFieldBaseProps) => {
...
+        created={
+          currentApplication?.created
+            ? new Date(currentApplication.created)
+            : application.created
+        }
+        modified={
+          currentApplication?.modified
+            ? new Date(currentApplication.modified)
+            : application.modified
+        }
```
MargretFinnboga, line 67: 🤩

libs/application/templates/financial-aid/src/fields/Status/Timeline/Timeline.tsx
```diff
@@ -9,16 +9,20 @@ import { ApplicationState } from '@island.is/financial-aid/shared/lib'
 
 import * as styles from './Timeline.css'
 import { timeline } from '../../../lib/messages'
-import { timelineSections } from '../../../lib/formatters'
+import { formatDate, timelineSections } from '../../../lib/formatters'
```
MargretFinnboga, line 12: this is unused import

MargretFinnboga (approved):
take one thing out, otherwise very nice 👍🏼

PR #7375, opened 2022-05-10
chore(testing): Improve code coverage by ignoring more

kristofer-lp8f:
## What
Improve code coverage for all apps, by ignoring untestable and some auto-generated code.

## Why
Code coverage is important for confidence in code quality. Currently code coverage is at an abysmal 28%. Ideally code coverage should be 80+%.

## Checklist:


jest.preset.js
```diff
@@ -18,6 +18,7 @@ module.exports = {
     '!**/*.config.*',
     '!**/infra/*',
     '!**/seeders/*',
-    '!**/migration/*',
+    '!**/migrations/*',
+    '!**/{buildOpenApi,openApi,main,index}.ts',
```
darrikonn, line 22: Some files include the actual logic in `index.ts` and those should be tested + collected for coverage.
Most of them are re-exports to define the API of the domain. So we can't safely ignore all of them.

darrikonn (changes requested):
The other files (except index.ts) are valid

jest.preset.js
kristofer-lp8f:
```suggestion
    '!**/{buildOpenApi,openApi,main}.ts',
```
What about `main.ts`?

darrikonn: That's fine, since it's not targeting `main.tsx` files. The other ones are just bootstraps, which is OK to ignore, but not mandatory :D.

darrikonn: Nah, thinking again, we should skip it since it could leave us in a bad debug trying to figure out why a coverage isn't taking affect on valid files.

kristofer-lp8f: Can you explain further? Can bootstrapping even be unit/integration tested? Should bootstrapping be tested in e2e (system testing)?

jest.preset.js
```diff
@@ -18,6 +18,7 @@ module.exports = {
     '!**/*.config.*',
     '!**/infra/*',
     '!**/seeders/*',
-    '!**/migration/*',
+    '!**/migrations/*',
+    '!**/{buildOpenApi,openApi,main}.ts',
```
kristofer-lp8f, line 22:
```suggestion
```
Is this better?

kristofer-lp8f (commented):
Should we not test these files? Coverage has no effect on whether or not they are tested. They may or may not be tested, but will not show in coverage.

jest.preset.js
darrikonn: We can't be 100% certain that logic in all `main.ts` files will all be the island.is nestjs bootstrap.

jest.preset.js
darrikonn: See my comment from above

jest.preset.js
kristofer-lp8f: What about `buildOpenApi.ts` and `openApi.ts`? Are there more files which don't contain logic or are only for bootstrapping/setup?

darrikonn (approved):

PR #7388, opened 2022-05-11
chore(web): Updated Sýslumenn's email list

RunarVestmann:
# Updated Sýslumenn's email list

## What

* Updated the email list for when someone contacts Sýslumenn from the service web
* Made the code for the email list more readable by replacing id's with variables

## Why

* This was a requested feature

## Checklist:


libs/api/domains/communications/src/lib/emailTemplates/serviceWebForms.ts
```diff
@@ -4,55 +4,55 @@ import { environment } from '../environments/environment'
...
+  | typeof ANDLAT_OG_DANARBU
+  | typeof THINGLYSINGAR
+  | typeof GJOLD_OG_INNHEIMTA
+  | typeof LOGGILDINGAR
+  | typeof VOTTORD
+  | typeof LOGRADAMAL
+  | typeof ONNUR_THJONUSTA_SYSLUMANNA
+  | typeof LEYFI
+  | typeof FULLNUSTUGERDIR
+  | typeof DEFAULT
```
kristofer-lp8f, line 55:
```suggestion
  | string
```
This doesn't really make sense to me. The types are all `string`.

libs/api/domains/communications/src/lib/emailTemplates/serviceWebForms.ts
```diff
@@ -4,55 +4,55 @@ import { environment } from '../environments/environment'
...
+
+// Leyfi
+const LEYFI = '7HbSNTUHJReJ2GPeT1ni1C'
+
+// Fullnustugerðir
+const FULLNUSTUGERDIR = '7LkzuYSzqwM7k8fJyeRbm6'
+
+// Default/fallback email address
+const DEFAULT = 'default'
+
```
kristofer-lp8f, line 42: Have you considered using enum?

kristofer-lp8f (approved):
LGTM 👍 

I'd recommend refactoring to use enums instead of copy-pasting/retyping IDs. The change you are doing is for the better, nonetheless :)

libs/api/domains/communications/src/lib/emailTemplates/serviceWebForms.ts
RunarVestmann: Aren't the types the actual possible value, like typeof FJOLSKYLDUMAL is the id, right?

libs/api/domains/communications/src/lib/emailTemplates/serviceWebForms.ts
RunarVestmann: I've never used enums in TypeScript before but it'd probably be a smarter move: https://www.typescriptlang.org/docs/handbook/enums.html

RunarVestmann: image
This is way better :)

kristofer-lp8f: @RunarVestmann

RunarVestmann: I addressed this in another PR: https://github.com/island-is/island.is/pull/7410

PR #7392, opened 2022-05-11
fix(core): Fixing timeout bug in enhancedFetch.

saevarma:
# Fixing timeout bug in enhancedFetch

## What

The introduction of the `agentkeepalive`  package introduced a regression in the enhancedFetch. This regression causes the enhanceFetch withAgent to ignore the timeout property and use the default provided in the `agentkeepalive` package. Meaning if an api is configured to use a enhancedFetch with 60sec timeout, it is resulting in a 20 sec timeout at runtime.

## Why

This is causing issue in the Child Residence Change application, where the api is configured to use 60 sec timeouts due to long requests waiting for the signed document. 

## Screenshots / Gifs

## Checklist:


libs/clients/middlewares/src/lib/createEnhancedFetch.ts
```diff
@@ -129,7 +129,10 @@ export const createEnhancedFetch = (
 
   builder.wrap(withAgent, {
     clientCertificate,
-    agentOptions,
+    agentOptions: {
+      timeout: typeof timeout === 'number' ? timeout : undefined,
```
eirikurn, line 133: If `timeout` is false, there should be no timeout. Can we disable agentkeepalive's timeout?

Also, now we have two implementations for the `timeout` argument. Should maybe get rid of `withTimeout`?

 Or should we use

eirikurn (changes requested):

eirikurn (approved):
LGTM

PR #7402, opened 2022-05-12
fix(j-s): Fix Client Build

gudjong:
# Fix Client Build

https://app.asana.com/0/1199153462262248/1202212024251105/f

## What

- Rewrites `tsconfig` files so the build catches `typescript` errors.

## Why

Typescript errors were not caught in the ci/cd pipeline.

## Checklist:

### DIFF
--- a/apps/judicial-system/web/tsconfig.app.json
+++ b/apps/judicial-system/web/tsconfig.app.json
@@ -3,7 +3,5 @@
   "compilerOptions": {
     "outDir": "../../../dist/out-tsc",
     "types": ["node", "jest"]
-  },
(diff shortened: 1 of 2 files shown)


eirikurn:
@saevarma curious if this is consistent with later nx, or if there's potential regressions.

saevarma:
> @saevarma curious if this is consistent with later nx, or if there's potential regressions.

It is consistent with later Nx versions.
New NextJS app now only contains tsconfig.json and tsconfig.spec.json.

Arnarkari93 (approved):

PR #7404, opened 2022-05-12
feat(Application-system): custom data provider error title and summary

Thorrik58:
# Custom data provider error title and summary

## What

Adding custom data provider error title and summary.

I went with a solution that lets the old dataprovider reason work to not have to change the older code and make decisions on what would be the right title as the summary is only supplied.

Throwing a typeError on the missing title when summary is defined and vice versa seems a bit hacky but it lets the dev know right away with an error why the error message isnt getting displayed.

## Why

Needed for better ux, user would always be presented by the fetch error title when that wasnt the case. For example if a user didnt have any debts in the payment distribution application.

## Screenshots / Gifs

The original default error message that appears if the dataproviders fails to fetch data or fails in any way not handled in the data provider:
image

The original way of representing custom error summary would look like this and is still supported to not having to change multiple applications with this pr:
image

```
return Promise.reject({
      reason:
        'The old way to represent error example summary. Something here didnt meet the prerequisites for example.',
    })

  onProvideError(error: { reason: StaticText }): FailedDataProviderResult {
    return {
      date: new Date(),
      data: {},
      reason: error.reason,
      status: 'failure',
    }
  }
```
___

Custom error summary and title on an error:
image

```
return Promise.reject({
      reason: {
        title: 'very bad error',
        summary: 'something really bad',
      },
      statusCode: 500,
    })

  onProvideError(error: {
    reason: ProviderErrorReason
    statusCode?: number
  }): FailedDataProviderResult {
    return {
      date: new Date(),
      data: {},
      reason: error.reason,
      status: 'failure',
      statusCode: error.statusCode,
    }
  }
```
___

Custom error summary and title supplied on a handled warning when user does not meet prerequisites:
image

```
    return Promise.reject({
      reason: {
        title: errorModal.noDebts.title,
        summary: errorModal.noDebts.summary,
      },
      statusCode: 404,
    })

  onProvideError(error: {
    reason: ProviderErrorReason
    statusCode?: number
  }): FailedDataProviderResult {
    return {
      date: new Date(),
      data: {},
      reason: error.reason,
      status: 'failure',
      statusCode: error.statusCode,
    }
  }
```
___

## Checklist:


libs/application/core/src/lib/dataProviderUtils.ts
```diff
@@ -29,9 +29,24 @@ function callProvider(
       return Promise.resolve(provider.onProvideSuccess(result))
     },
     (error) => {
+      if (
+        error.reason &&
+        ((!('title' in error.reason) && 'summary' in error.reason) ||
+          ('title' in error.reason && !('summary' in error.reason)))
+      ) {
+        throw new TypeError(
```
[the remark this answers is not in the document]
Thorrik58: I believe i fixed this in the commit just before you commented, sorry. Think its addressed with the object check here https://github.com/island-is/island.is/pull/7404/files#diff-8fa9d447c058a57e82e219d12bd6ef2539d3a8348d29b38c1ce325d8fc2d7d22R34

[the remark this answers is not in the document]
darrikonn: true :)

darrikonn (approved):

libs/application/core/src/lib/dataProviderUtils.ts
```diff
@@ -1,10 +1,67 @@
...
+  }
+  if (
+    typeof reason === 'object' &&
+    'title' in reason &&
+    'summary' in reason &&
+    reason.title !== '' &&
+    reason.summary !== ''
+  ) {
+    return true
+  }
```
darrikonn, line 35: Can be replaced with:
```typescript
if (typeof reason === 'object' && reason.title && reason.summary) {
  return true
}
```

libs/application/ui-shell/src/components/FormExternalDataProvider.tsx
```diff
@@ -51,12 +52,12 @@ const ItemHeader: React.FC<{ title: StaticText; subTitle?: StaticText }> = ({
   )
 }
 
-const isTranslationObject = (reason?: StaticText) => {
-  if (typeof reason !== 'object') {
+const isTranslationObject = (text?: StaticText) => {
```
darrikonn, line 55: Rather use the one from `libs/application/core/src/lib/dataProviderUtils.ts`

libs/application/core/src/lib/dataProviderUtils.ts
```diff
@@ -1,10 +1,67 @@
...
+    if (isProviderErrorReason(reason)) {
+      return {
+        title: reason.title,
+        summary: reason.summary,
+      }
+    }
+    if (
+      isTranslationObject(reason) ||
+      typeof reason === 'string' ||
+      reason instanceof String
```
darrikonn, line 52: Better readability to use `isString` from `lodash`:
```typescript
import isString from 'lodash/isString'
```

libs/application/core/src/lib/dataProviderUtils.ts
darrikonn: In addition, it's better for readability to use:
```typescript
import isObject from 'lodash/isObject`
```

libs/application/core/src/lib/dataProviderUtils.ts
```diff
@@ -1,10 +1,67 @@
...
 import {
   BasicDataProvider,
   CustomTemplateFindQuery,
 } from '../types/BasicDataProvider'
-import { FormatMessage } from '../types/Form'
+import { FormatMessage, StaticText } from '../types/Form'
+import { coreErrorMessages } from './messages'
+
+const isTranslationObject = (text?: StaticText) => {
+  if (typeof text !== 'object') {
```
darrikonn, line 14: ditto

libs/application/core/src/lib/dataProviderUtils.ts
```diff
@@ -1,10 +1,67 @@
...
+    return false
+  }
+
+  return text.id !== undefined
+}
+
+export const isProviderErrorReason = (
+  reason: ProviderErrorReason | StaticText,
+): reason is ProviderErrorReason => {
+  if (typeof reason === 'string' || reason instanceof String) {
```
darrikonn, line 24: ditto isString

darrikonn (changes requested):
Nice, better.

libs/application/core/src/lib/dataProviderUtils.ts
```diff
@@ -9,10 +9,10 @@ import {
 } from '../types/BasicDataProvider'
 import { FormatMessage, StaticText } from '../types/Form'
 import { coreErrorMessages } from './messages'
-import { isString } from 'lodash'
+import { isString, isObject } from 'lodash'
```
darrikonn, line 12: There should be an eslint rule that forbids this. But we don't want to import destructed objects from lodash since it'll import the whole lodash library and increase the bundle size.

This is preferred due to a much lower bundle size:
```typescript
import isString from 'lodash/isString'
```

darrikonn (changes requested):

darrikonn (approved):

PR #7429, opened 2022-05-16
chore(j-s): Posts Error Events on Signing Failure

gudjong:
# Dokobit Signing

## What

- Rewrites module configuration using new ConfigModule pattern.
- Posts an error event when failing to request a signature.
- Posts an error event when failing to get a signature confirmation.

## Why

This is the recommended pattern.

## Checklist:


Thorrik58 (approved):

libs/dokobit-signing/src/lib/signing.service.ts
```diff
@@ -61,14 +57,48 @@ interface DokobitStatusResponse {
...
+        502, // Bad gateway
+        99999,
+        'Checksum verification failed',
+      )
+    }
+
+    return documentContent
+  }
+
+  private delay(ms: number) {
```
oddsson, line 98: Is this the right place for a delay function? Seems more like a helper then a property of the signing functionality.

oddsson (changes requested):
One question

libs/dokobit-signing/src/lib/signing.service.ts
gudjong: Good question. I think it belongs here as this service is handling the nuance of talking to Dokobit.

The flow goes something like this:
- Ask Dokobit to request a document signature from an end user.
- Return feedback (including control code) to the end user.
- Wait until the user has completed the signature or time out.
- Return the signed document (if not time out).

We want to give the end user feedback on signature completion/cancellation/error as soon as Dokobit is ready, so repeated calls in rapid succession are necessary. Also, the timeout is rather short, so we don't have the option of not calling the api in this way. Pushing this logic further up the call chain seems unnecessary to me and it would make it necessary to duplicate that logic everywhere the signing service is used. Currently, it is used in at least three places in two applications.

The intended function of the `getSignedDocument` method is to wait for the signed document to become available and then return it. Perhaps, it would be useful to rename the method to `waitForSignedDocument` or something similar.

Also note that this is not new code. It was just moved within the source file.

oddsson (approved):

PR #7431, opened 2022-05-16
feat(j-s): show message when filetype is not supported

Arnarkari93:
[Asana](https://app.asana.com/0/1199153462262248/1201976501087613/f)
## What

`island-ui`
  - Support `label` prop being component in `AccordionItem`

`judicial-system`
- Show an error message when the file type is not supported
- Fix padding for the error message and upload to court button.

## Screenshots / Gifs
Screenshot 2022-05-16 at 10 39 34

## Checklist:


gudjong (approved):
LGTM

barabrian (approved):
reviewed island-ui part. LGTM

libs/island-ui/core/src/lib/Accordion/AccordionItem/AccordionItem.tsx
```diff
@@ -145,12 +145,16 @@ export const AccordionItem = forwardRef<HTMLButtonElement, AccordionItemProps>(
                     component="span"
                     height="full"
                     width="full"
-                    display="flex"
+                    display="block"
                     alignItems="center"
```
barabrian, line 149: one late comment, what is the reason for the display to change to `block`? `alignItems="center"` is probably useless here if we change the display property to block

barabrian: and this could cause some side-effects
e.g. here I'm toggling between block and flex

barabrian (changes requested):
Sorry for the review change 😄
I want to resolve the display change comment before this gets merged 🙂

libs/island-ui/core/src/lib/Accordion/AccordionItem/AccordionItem.tsx
Arnarkari93: I added it because I needed the parent element (the label in this case) to take full with. So I could space these two elements with justify content.
Screenshot 2022-05-17 at 09 55 15

I fixed the issue in the judicial-system by adding `flex-grow: 1` to the custom label.

PR #7440, opened 2022-05-17
fix(parental-leave): self employed and attachment

mahanmashoof:
## What

Erasing employer/self-employment data from the UI if the user changes employment options while filling the form or at the form review stage.

## Why

https://dit-iceland.atlassian.net/jira/software/projects/UUFV/boards/71?selectedIssue=UUFV-67

## Checklist:


libs/application/templates/parental-leave/src/forms/ParentalLeaveForm.ts
```diff
@@ -388,21 +388,12 @@ export const ParentalLeaveForm: Form = buildForm({
           id: 'employer',
           title: parentalLeaveFormMessages.employer.subSection,
           children: [
-            buildRadioField({
+            buildCustomField({
+              component: 'SelfEmployed',
               id: 'employer.isSelfEmployed',
               title: parentalLeaveFormMessages.selfEmployed.title,
               description: parentalLeaveFormMessages.selfEmployed.description,
               width: 'half',
```
Thornvik, line 396: Dont think you need to specify `width: 'half'` here as well as you have it in the custom component.

Thornvik (approved):
Looks good

PR #7442, opened 2022-05-17
feat(login-service): company search

sigruntg:
# Adding company search to login service form

[1pdu0u3](https://app.clickup.com/t/1pdu0u3)

## What

* Adding company search to company search form. 
* Making name, national id and type of operation field disabled - get data from company registry.
* Adding alert message and hinder user from going forward if ÍSAT number does not start with 84 (Opinber stofnun).

## Screenshots / Gifs

image

## Checklist:


libs/application/templates/login-service/src/fields/InformationAboutApplication/InformationAboutApplication.tsx
```diff
@@ -0,0 +1,201 @@
...
+}
+
+export const InformationAboutApplication = ({
+  application,
+  errors,
+}: FieldBaseProps) => {
+  const { setValue } = useFormContext()
+  const { formatMessage } = useLocale()
+  const [nationalId, setNationalId] = useState<string>('')
+  const [isatNumber, setIsatNumber] = useState<string>('')
```
darrikonn, line 56: You can use `watch` from `react-hook-form` to track the actual value instead of storing it in two places :)

sigruntg: Cool, I'll look into that! 😄

PR #7444, opened 2022-05-17
chore(j-s): Cleanup on the Cases page

oddsson:
# Cleanup on the Cases page

[Asana](https://app.asana.com/0/1199153462262248/1202282745438085/f)

## What

- Remove console warning "p cannot be a defendant of p"
- Extract the rendering of `fd. [nationalid]` and `kt. [nationalid]` into a function
- Fix regression where the institution name was being rendered in a single line instead of broken up into two lines
- Remove unnecessary p tags from the skeleton loader component 

## Why

Cleaner product.

## Screenshots / Gifs

Nothing has changed visually.

## Checklist:


apps/judicial-system/web/src/components/Logo/Logo.tsx
```diff
@@ -11,14 +11,21 @@ interface Props {
...
+  )
 
   return (
     <div className={styles.logoContainer}>
       <Box marginRight={[0, 0, 0, 2]} marginBottom={[0, 0, 1, 0]}>
         <LandWightsLogo />
       </Box>
       <p className={styles.logoText}>
-        {user?.institution?.name ?? defaultInstitution}
+        <span>{institutionNameFirstHalf.toString().replace(',', ' ')}</span>
```
Arnarkari93, line 27: Can we not use `word-wrap: break-word` in CSS instead of doing this in JS?

Arnarkari93 (changes requested):
One small nitpick or question.

apps/judicial-system/web/src/components/Logo/Logo.tsx
oddsson: The problem with that is that sometimes the institution name is three words and sometimes two, f.x. Lögreglustjórinn á höfuðborgarsvæðinu and Héraðsdómur Reykjaness. We obviously don't want 

Lögreglustjórinn
á
höfuðborgarsvæðinu

Arnarkari93 (approved):

apps/judicial-system/web/src/components/Logo/Logo.tsx
```diff
@@ -11,14 +11,21 @@ interface Props {
...
+    0,
+    institutionNameArr.length - 1,
+  )
 
   return (
     <div className={styles.logoContainer}>
       <Box marginRight={[0, 0, 0, 2]} marginBottom={[0, 0, 1, 0]}>
         <LandWightsLogo />
       </Box>
       <p className={styles.logoText}>
```
oddsson, line 26: Should this be extracted into a separate component and that component tested, instead of mounting the entire Cases page to test this, like we are doing in this pr? What do you think?

Arnarkari93: This component is so small. This is basically the only logic in it. I think either testing this component or you can extract this logic to a function and test that.

apps/judicial-system/web/src/routes/Shared/Cases/PastCases.tsx
```diff
@@ -77,26 +77,19 @@ const PastCases: React.FC<Props> = (props) => {
...
-                    </Text>
-                  </Text>
-                )
-              ) : (
-                <Text as="span" variant="small" color="dark400">
-                  {`+ ${theCase.defendants.length - 1}`}
-                </Text>
-              )}
+              <Text as="span" variant="small">
+                {theCase.defendants.length === 1
```
oddsson, line 81: Should this be extracted into a separate component and that component tested, instead of mounting the entire Cases page to test this, like we are doing in this pr? What do you think?

PR #7465, opened 2022-05-19
chore(core): Tackle unused vars to reduce lint warnings

darrikonn:
# Tackle unused vars to reduce lint warnings

## What

Silence Github Action's lint suggestions on PRs

## Why

There's a lot of noise going on when viewing and reviewing PRs since we have a lot of lint warnings that Github Actions is posting to PRs.

## Checklist:


johannaagma (approved):

.eslintrc.json
```diff
@@ -67,7 +67,19 @@
...
+            }
+          },
+          {
+            "selector": "variable",
+            "format": ["camelCase", "PascalCase", "UPPER_CASE"],
+            "filter": {
+              "regex": "^_([0-9])?$",
+              "match": false
+            }
           },
```
kristofer-lp8f, line 83: I don't understand what this change does. Can you rationalize it for me? What are you trying to do?

darrikonn: We have an eslint rule called `@typescript-eslint/no-unused-vars` which throws an error of unused variables. We however sometimes need dummy variables in code:
```typescript
const [_, setFoo] = useState('')
```
So we make an exception by naming the variable as `_`. In some cases we need multiple dummy variables in the same scope, so we can suffix them with a number:
```typescript
const [_1, _2, height] = [2, 1, 5]
```

We recently introduced the `@typescript-eslint/naming-convention` eslint rule, which clashes with the `@typescript-eslint/no-unused-vars`, since we yet again (what this PR implements) have to make exceptions for `^_([0-9])?$`

kristofer-lp8f: I would suggest removing this. Dummy variables can still have descriptive names even though they are not used. Using `_1`, `_2` is opaque and a naive solution to having to declare a variable. Much better, using your code, would be:
```js
const [_width, _depth, height] = [2, 1, 5]
```
This conveys meaning to future readers and makes no difference in code, since they aren't used.

.eslintrc.json
```diff
@@ -67,7 +67,19 @@
...
+            "custom": {
+              "regex": "^_[a-zA-Z0-9]*$",
+              "match": true
+            }
+          },
+          {
+            "selector": "variable",
+            "format": ["camelCase", "PascalCase", "UPPER_CASE"],
+            "filter": {
+              "regex": "^_[a-zA-Z0-9]*$",
```
kristofer-lp8f, line 80:
```suggestion
              "regex": "^\w+$",
```

.eslintrc.json
```diff
@@ -55,8 +55,8 @@
           "warn",
           {
             "ignoreRestSiblings": true,
-            "argsIgnorePattern": "^_([0-9])?$",
-            "varsIgnorePattern": "^_([0-9])?$"
+            "argsIgnorePattern": "^_[a-zA-Z0-9]*$",
+            "varsIgnorePattern": "^_[a-zA-Z0-9]*$"
```
kristofer-lp8f, line 59:
```suggestion
            "argsIgnorePattern": "^\w+$",
            "varsIgnorePattern": "^\w+$"
```

kristofer-lp8f (approved):
LGTM 🍪 

[MDN regex docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions/Character_Classes)

.eslintrc.json
kristofer-lp8f:
```suggestion
            "argsIgnorePattern": "^_[a-zA-Z]\w*$",
            "varsIgnorePattern": "^_[a-zA-Z]\w*$"
```
This might be better if we want to prohibit e.g. `_1`, `_2`, etc.

.eslintrc.json
kristofer-lp8f:
```suggestion
              "regex": "^_[a-zA-Z]\w*$",
```

gudjong (approved):
LGTM

.eslintrc.json
darrikonn: No, we don't want to prohibit it. The dummy variable isn't always descriptive, and can improve code readability by "ignoring" its unnecessary meaning behind `_`

.eslintrc.json
darrikonn: `\w` is not the same as `[a-zA-Z0-9]` since it allows `_`, which is not in line with our naming standards:
`_foo_bar` should be flagged as incorrect.

Thorrik58 (approved):
LGTM for all files related to the application system.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "island.is",
-  "version": "0.0.1",
+  "version": "0.0.2",
```
saevarma, line 3: Maybe we should be updating this field according to the version number of the releases?

saevarma (approved):
Aranja and Core files looks good, great initiative 🙌

essgje (approved):
LF

package.json
darrikonn: As per the change in `nx.json`, this can be treated as when we want to trigger builds for all applications.
If that's the purpose, then it would clash with the actual release version number.
So IMO I don't see the benefit of keeping it in sync, but I see the benefit of keeping it separate. But it would be interesting to see the counter-argument?

valurefugl (approved):

package.json
peturgq: I'm not sure about this change here, why are we just first changing this version now? 
I'm wondering if we have to worry about any side effects from this. 
I think this is something @pshomov should take a look at.

RunarVestmann (approved):
The files owned by Stefna look good to me :)

package.json
darrikonn: Since we're not using the package.json version for anything, we haven't had the need to change it before.
Can't think of any possible side effects, other than what this PR introduces, i.e. making projects affected by nx due to the change to nx.json.

saevarma: > As per the change in `nx.json`, this can be treated as when we want to trigger builds for all applications. If that's the purpose, then it would clash with the actual release version number. So IMO I don't see the benefit of keeping it in sync, but I see the benefit of keeping it separate. But it would be interesting to see the counter-argument?

Yes I don't have any counter argument other than just to be in sync, which is probably just the effort with no extra value.
So you changed this value to cause `package.json` to be affected and trigger build for everything? I can see the value in having that possibility, so maybe we just need to document it somewhere.

darrikonn: True.
Yeah, needs to be documented 👍

Snojo (approved):
LGTM on DevOps owned files

PR #7474, opened 2022-05-19
feat(regulations): Regulations/UI addons

Valur:
## What

Add ui tweaks used in regulations admin and update some related code

## Checklist:


thordurhhh (approved):
👍

libs/island-ui/core/src/lib/Tag/Tag.tsx
```diff
@@ -79,10 +87,12 @@ export const Tag = forwardRef<HTMLButtonElement & HTMLAnchorElement, TagProps>(
     const hyphenated = hyphenate && typeof children === 'string' && (
       <Hyphen>{children}</Hyphen>
     )
+    const renderCloseIcon = removable && !!onClick && !href
 
     const content = (
       <Text variant="eyebrow" as="span" truncate={truncate}>
         {!truncate && hyphenate ? hyphenated : children}
+        {renderCloseIcon && <Icon icon="close" className={styles.closeIcon} />}
```
Arnarkari93, line 95: I think this should not be implemented in `island-ui` until the designs in Figma have been updated. If this should be added we need to account for accessibility.

Arnarkari93 (changes requested):

libs/island-ui/core/src/lib/Tag/Tag.tsx
Arnarkari93: I think we should instead add a close icon by allowing the user to inject a custom icon instead—that way we allow for more flexibility for different use-cases.

saevarma (commented):
Why does this PR include the `package.json` changes?

PR #7495, opened 2022-05-23
feat(auth-public-api): Delegation types filter for authActorDelegations

raRaRa:
## What
This PR adds new optional delegationTypes filter to authActorDelegations that can filter actorDelegations based on one or more delegation types. 

## Why

Currently we are always fetching all actor delegations, but in some cases we only need to fetch certain type of delegations, which will make the query faster and only return the necessary data. One example is the Island.is Application System.

## Screenshots / Gifs

image

## Checklist:


apps/services/auth-public-api/src/app/modules/delegations/actorDelegations.controller.spec.ts
```diff
@@ -356,6 +356,28 @@ describe('ActorDelegationsController', () => {
...
+
+        it('should return a legal guardian delegation since the type is included in the delegationTypes filter', async () => {
+          // Act
+          const res = await server.get(
+            `${path}${query}&delegationTypes=${DelegationType.Custom}&delegationTypes=${DelegationType.LegalGuardian}`,
+          )
+
+          // Assert
+          expect(res.status).toEqual(200)
+          expect(res.body).toHaveLength(1)
```
saevarma, line 368: I think we should add assert for the type of the delegation to make sure we get the LegalGuardian delegation.
This test might pass later on in the future with wrong type of delegation.

apps/services/auth-public-api/src/app/modules/delegations/actorDelegations.controller.spec.ts
```diff
@@ -408,6 +430,28 @@ describe('ActorDelegationsController', () => {
...
+
+        it('should return a procuring holder delegation since the type is included in the delegationTypes filter', async () => {
+          // Act
+          const res = await server.get(
+            `${path}${query}&delegationTypes=${DelegationType.Custom}&delegationTypes=${DelegationType.ProcurationHolder}&delegationTypes=${DelegationType.PersonalRepresentative}`,
+          )
+
+          // Assert
+          expect(res.status).toEqual(200)
+          expect(res.body).toHaveLength(1)
```
saevarma, line 442: Same about asserting the type.

apps/services/auth-public-api/src/app/modules/delegations/actorDelegations.controller.spec.ts
```diff
@@ -516,6 +560,28 @@ describe('ActorDelegationsController', () => {
...
 
+        it('should return a personal representative delegation since the type is included in the delegationTypes filter', async () => {
+          // Act
+          const res = await server.get(
+            `${path}${query}&delegationTypes=${DelegationType.Custom}&delegationTypes=${DelegationType.ProcurationHolder}&delegationTypes=${DelegationType.PersonalRepresentative}`,
+          )
+
+          // Assert
+          expect(res.status).toEqual(200)
+          expect(res.body).toHaveLength(1)
```
saevarma, line 571: Same about asserting the type

libs/auth-api-lib/src/lib/services/delegations.service.ts
```diff
@@ -323,21 +323,43 @@ export class DelegationsService {
...
     const delegationPromises = []
 
-    if (!client || client.supportsLegalGuardians) {
+    const hasDelegationTypeFilter =
+      delegationTypes && delegationTypes.length > 0
+
+    if (
+      (!client || client.supportsLegalGuardians) &&
+      (!hasDelegationTypeFilter ||
+        delegationTypes?.includes(DelegationType.LegalGuardian))
```
saevarma, line 340: Nit pick, but the optional chaining is unnecessary as you have checked it before when setting `hasDelegationTypeFilter`

saevarma (changes requested):
Great job!
Just minor improvements in the testing, and we should maybe add a test case for filtering the custom delegations?

libs/auth-api-lib/src/lib/services/delegations.service.ts
raRaRa: Fixed. TS was happy with this change, which I thought wasn't the case

saevarma (approved):

PR #7498, opened 2022-05-23
feat(j-s): Email Error Event

gudjong:
# Email Error Event

[Villa á #rettarvorslugatt-errors þegar ekki tekst að senda tölvupóst](https://app.asana.com/0/1199153462262248/1201962718337603/f)

## What

- Post an error event when sending email fails.

## Why

Better monitoring for the development team.

## Checklist:


apps/judicial-system/backend/src/app/modules/case/case.service.ts
```diff
@@ -342,15 +346,40 @@ export class CaseService {
...
       this.logger.error('Failed to send email', { error })
+
+      this.eventService.postErrorEvent(
+        'Failed to send email',
+        {
+          courtCaseNumber,
+          subject,
+          to: Array.isArray(to)
+            ? to
+                .reduce(
```
oddsson, line 364: I was a bit confused about what this code did until I wrote it in a terminal and saw what it did with my own eyes. I was mainly confused about the `.slice(2)` part. I think it would be clearer to include a index parameter to the reduce function and use that to do what slice does in this version, i.e.

`[].reduce((acc, recipient, index) => `${index > 0 ? `${acc}, ` : ''}${recipient.name} (${recipient.address})`, '')`

What do you think?

oddsson (changes requested):
One question

oddsson (approved):

PR #7517, opened 2022-05-25
feat(j-s): rename restrictedCase to limitedAccessCase

Arnarkari93:
[Asana - Consider renaming restricted case ](https://app.asana.com/0/1199153462262248/1202211680145826/f)

## What

- Rename `restrictedCase` to `limitedAccessCase`

## Why

- To avoid confusion between `RestrictionCase`

## Checklist:


apps/judicial-system/api/src/app/modules/file/file.module.ts
```diff
@@ -1,7 +1,7 @@
 import { Module } from '@nestjs/common'
 
 import { FileController } from './file.controller'
-import { RestrictedFileController } from './restrictedFile.controller'
+import { RestrictedFileController } from './limitedAccessFile.controller'
```
oddsson, line 4: Please rename this to `LimitedAccessFileController`

oddsson (approved):

PR #7523, opened 2022-05-25
fix: license service driving validation

osk:
When trying to get a pkpass url we are validating that the license has an image of the person and it's not older than a set cutoff point. This happens, but if it's not valid be just log and send the user a url...

This fixes that 🙈 

## What

## Why

## Screenshots / Gifs

## Checklist:


thordurhhh (approved):
👍

PR #7536, opened 2022-05-27
feat(j-s): report when create internal case fails

Arnarkari93:
[Asana - Villa á #rettarvorslugatt-error þegar LÖKE nær ekki að stofna RVG mál](https://app.asana.com/0/1199153462262248/1202010789863360/f)

## What

- report if create interal case fails

## Why

- better transparency when this process fails 

## Checklist:

### DIFF
--- a/apps/judicial-system/backend/src/app/modules/case/case.service.ts
+++ b/apps/judicial-system/backend/src/app/modules/case/case.service.ts
@@ -606,48 +606,61 @@ export class CaseService {
     let prosecutorId: string | undefined
     let courtId: string | undefined
 
-    if (caseToCreate.prosecutorNationalId) {
-      const prosecutor = await this.userService.findByNationalId(
-        caseToCreate.prosecutorNationalId,
-      )
-
-      if (!prosecutor || prosecutor.role !== UserRole.PROSECUTOR) {
-        throw new BadRequestException(
-          `User ${prosecutor.id} is not registered as a prosecutor`,
+    try {
+      if (caseToCreate.prosecutorNationalId) {
+        const prosecutor = await this.userService.findByNationalId(
+          caseToCreate.prosecutorNationalId,
         )
+
+        if (!prosecutor || prosecutor.role !== UserRole.PROSECUTOR) {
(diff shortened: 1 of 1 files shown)


apps/judicial-system/backend/src/app/modules/case/case.service.ts
```diff
@@ -606,48 +606,61 @@ export class CaseService {
...
 
-    if (caseToCreate.prosecutorNationalId) {
-      const prosecutor = await this.userService.findByNationalId(
-        caseToCreate.prosecutorNationalId,
-      )
-
-      if (!prosecutor || prosecutor.role !== UserRole.PROSECUTOR) {
-        throw new BadRequestException(
-          `User ${prosecutor.id} is not registered as a prosecutor`,
+    try {
```
Arnarkari93, line 609: This is a bad diff. I added `try_catch` around everything

oddsson (approved):

apps/judicial-system/backend/src/app/modules/case/case.service.ts
```diff
@@ -606,48 +606,61 @@ export class CaseService {
...
+              gender: caseToCreate.accusedGender,
+              address: caseToCreate.accusedAddress,
+            },
+            transaction,
+          )
+
+          return caseId
+        })
+        .then((caseId) => this.findById(caseId))
+    } catch (error) {
```
gudjong, line 652: Catching an error here is one way to post error events for LÖKE created cases.

However, this will not catch all possible errors. In particular, failure to call the backend from the xrd-api will not be caught.

A more robust approach would be to catch errors in the xrd-api itself and post events from there. This approach will require either a duplication of the event service in the xrd-api application or moving the event service to a shared library. I recommend using a shared library.

So, consider:
- modifying this PR to handle errors in the xrd-api, or
- creating a refactoring task to move the error handling to the xrd-api later.

gudjong (changes requested):

Arnarkari93:
Closed in favor of: https://github.com/island-is/island.is/pull/7619

PR #7541, opened 2022-05-27
feat(web): Hotfix - GridContainer side padding added back in mobile

RunarVestmann:
Cherry pick commit from (#7530 )


stjanilofts (approved):
Approved, but maybe @oddsson would like to have a look in case this will affect something we don't know about?

PR #7543, opened 2022-05-28
feat(core): Contentful entry tagging service

RunarVestmann:
# Contentful entry tagging service

## Why

* We want to manage user permissions in Contentful and the approach chosen was to utilize roles and tags.
* When a user creates an entry then this service will automatically apply the correct tag to that entry so the user permissions are correctly set (since the tag will be connected to the role that the user is linked to)

## Checklist:


workspace.json
```diff
@@ -9569,6 +9569,59 @@
...
+          "configurations": {
+            "production": {
+              "optimization": true,
+              "extractLicenses": true,
+              "inspect": false,
+              "fileReplacements": [
+                {
+                  "replace": "apps/services/contentful-entry-tagger/src/environments/environment.ts",
+                  "with": "apps/services/contentful-entry-tagger/src/environments/environment.prod.ts"
+                }
```
pshomov, line 9597: We stopped using this method of configuration.

pshomov (changes requested):
Will add the docker image to our registry so it is ready to go.
Needs to have `infra` folder with a .ts file describing the deployment. For example, see other similar services. Also needs to be wired up in the [islandis chart/deployment](https://github.com/island-is/island.is/blob/main/infra/src/uber-charts/islandis.ts)

workspace.json
```diff
@@ -9569,6 +9569,59 @@
...
+        },
+        "test": {
+          "builder": "@nrwl/jest:jest",
+          "outputs": ["coverage/apps/services/contentful-entry-tagger"],
+          "options": {
+            "jestConfig": "apps/services/contentful-entry-tagger/jest.config.js",
+            "passWithNoTests": true
+          }
+        }
+      }
```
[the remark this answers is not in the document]
pshomov: This is so a docker image is created

workspace.json
RunarVestmann: Should I remove the replace and with then? This got added automatically when I bootstrapped the service

pshomov: Just remove this to begin with. There is a library that is used to retrieve config if I recall correctly.

workspace.json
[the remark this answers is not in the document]
pshomov: Sorry, yes - that sounds good.

apps/services/contentful-entry-tagger/infra/contentful-entry-tagger-service.ts
```diff
@@ -0,0 +1,21 @@
+import { service, ServiceBuilder } from '../../../../infra/src/dsl/dsl'
+
+export const serviceSetup = (): ServiceBuilder<'contentful-entry-tagger-service'> =>
+  service('contentful-entry-tagger-service')
+    .image('contentful-entry-tagger')
```
pshomov, line 5: Please use the name as defined in `workspace.json`

workspace.json
RunarVestmann: Done :)

infra/src/uber-charts/islandis.ts
```diff
@@ -168,4 +171,5 @@ export const FeatureDeploymentServices = []
 export const ExcludedFeatureDeploymentServices = [
   userNotificationService,
   userNotificationWorkerService,
+  contentfulEntryTagger,
```
pshomov, line 174: Just FYI - this means the service cannot be feature-deployed.

infra/src/uber-charts/islandis.ts
```diff
@@ -108,6 +110,7 @@ export const Services: EnvironmentServices = {
     appSystemApiWorker,
     userNotificationService,
     userNotificationWorkerService,
+    contentfulEntryTagger,
```
pshomov, line 113: You should probably start by having the service added only to the dev stack, while it is being developed

apps/services/contentful-entry-tagger/infra/contentful-entry-tagger-service.ts
```diff
@@ -0,0 +1,21 @@
...
+    .resources({
+      requests: {
+        cpu: '100m',
+        memory: '256Mi',
+      },
+      limits: {
+        cpu: '400m',
+        memory: '1024Mi',
+      },
+    })
```
pshomov, line 21: No change needed - You can skip this part if the service has no particular needs.

apps/services/contentful-entry-tagger/infra/contentful-entry-tagger-service.ts
```diff
@@ -0,0 +1,21 @@
+import { service, ServiceBuilder } from '../../../../infra/src/dsl/dsl'
+
+export const serviceSetup = (): ServiceBuilder<'contentful-entry-tagger-service'> =>
+  service('contentful-entry-tagger-service')
```
pshomov, line 4: This name has to match the one in the [helm repo](https://github.com/island-is/helm/pull/645).

pshomov (changes requested):

apps/services/contentful-entry-tagger/infra/contentful-entry-tagger-service.ts
RunarVestmann: I changed the name in the helm repo to be contentful-entry-tagger-service

infra/src/uber-charts/islandis.ts
RunarVestmann: Ok, yep I agree

infra/src/uber-charts/islandis.ts
RunarVestmann: That's fine, I don't see a reason for it to be feature deployed

apps/services/contentful-entry-tagger/infra/contentful-entry-tagger-service.ts
RunarVestmann: I think I've got matching names now

Snojo (approved):
LGTM on the DevOps Side

RunarVestmann:
As of right now the ingress is open so it can be called from anywhere, I could look into request verification to make sure the request is coming from Contentful.

PR #7545, opened 2022-05-30
feat(j-s): feature flag modify ruling button

Arnarkari93:
[Asana - feature flag Leiðrétta úrskurð](https://app.asana.com/0/1199153462262248/1202360581950457/f)
## What

- But Leiðrétta button under feature flag

## Checklist:

### DIFF
--- a/apps/judicial-system/web/src/routes/Shared/SignedVerdictOverview/SignedVerdictOverviewForm.tsx
+++ b/apps/judicial-system/web/src/routes/Shared/SignedVerdictOverview/SignedVerdictOverviewForm.tsx
@@ -36,6 +36,7 @@ import {
   UserRole,
   isAcceptingCaseDecision,
   Case,
+  Feature,
 } from '@island.is/judicial-system/types'
 import {
   capitalize,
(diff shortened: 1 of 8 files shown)


apps/judicial-system/web/src/routes/Shared/SignedVerdictOverview/SignedVerdictOverviewForm.tsx
```diff
@@ -390,6 +394,7 @@ const SignedVerdictOverviewForm: React.FC<Props> = (props) => {
               ) : user?.role === UserRole.JUDGE ||
                 user?.role === UserRole.REGISTRAR ? (
                 <Button
+                  variant="ghost"
```
Arnarkari93, line 397: sneek in making both buttons ghoast 🤫

charts/judicial-system/values.prod.yaml
```diff
@@ -19,7 +19,7 @@ judicial-system-api:
     AUTH_AUDIENCE: 'rettarvorslugatt.island.is'
     BACKEND_URL: 'http://web-judicial-system-backend'
     CONTENTFUL_HOST: 'cdn.contentful.com'
-    HIDDEN_FEATURES: ''
```
gudjong, line 22: Consider also hiding on staging, so the hiding can be tested there.

gudjong: Note, that the yaml files should not be modified directly. The judicial-system-api.ts file in the infra folder of the judicial-system-api application should be modified and the yaml files generated from it.

gudjong (changes requested):

charts/judicial-system/values.prod.yaml
Arnarkari93: Addressed here: https://github.com/island-is/island.is/pull/7545/commits/847862f2d6978e1e5be581917b13cbf7628580b5

gudjong (approved):

Snojo (approved):
Looks good on DevOps side

PR #7546, opened 2022-05-30
feat(fa): Confirmation emails when application is submitted from Application System

davidpxn:
# ... 👆

https://app.asana.com/0/1202167941440767/1202189107644215

## What

- New endpoint in backend to send spouse email.
- Hook to send spouse email when applicant(with spouse) submits.
- Add new email format to notify applicant that he needs to wait for spouse.

## Checklist:


apps/financial-aid/backend/src/app/modules/application/dto/spouseEmail.dto.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  @IsOptional()
+  @IsString()
+  @ApiProperty()
+  readonly spouseName: string
+
+  @IsNotEmpty()
+  @IsString()
+  @ApiProperty()
+  readonly spouseEmail: string
```
MargretFinnboga, line 25: why is spousename optional but not spouse email?

libs/api/domains/municipalities-financial-aid/src/lib/dto/spouseEmail.input.ts
```diff
@@ -0,0 +1,34 @@
...
+  @Field()
+  readonly name!: string
+
+  @Allow()
+  @Field()
+  readonly email!: string
+
+  @Allow()
+  @Field()
+  readonly spouseName!: string
```
MargretFinnboga, line 17: here spouse name is not optional?

libs/application/templates/financial-aid/src/fields/Summary/SummaryForm.tsx
```diff
@@ -52,6 +56,19 @@ const SummaryForm = ({ application, goToScreen }: FAFieldBaseProps) => {
...
+    setBeforeSubmitCallback &&
+      setBeforeSubmitCallback(async () => {
+        const response = await sendSpouseEmail()
+        if (response) {
+          return [true, null]
+        }
+        return [false, 'Failed to send emails']
+      })
+  }
+
```
MargretFinnboga, line 71: Just an overall question why are we not handling sendSpouseEmail just in the backend? When creating a application, check if applicant has spouse and send email there from, why here in the application system?

MargretFinnboga (commented):
questionss ❓

libs/application/templates/financial-aid/src/fields/Summary/SummaryForm.tsx
davidpxn: We need to do it like this because the application is not sent to our backend until both the applicant and spouse have submitted. When an applicant (that has spouse) submits it is only saved in the AS database, but we need to send email to the spouse when this happens to notify the spouse that the application is waiting.

MargretFinnboga: Ahh of course I see, however are we sure we want to not allow applicant to send application if the email fails?

MargretFinnboga: Maybe just have a state, and if it fails to give the opportunity to send again or send a link himself or some?

MargretFinnboga (commented):
Just one pæling 🤔

davidpxn:
> Just one pæling 🤔

I changed this so that now the application is sent even if the email sending fails. If it fails then we display message to tell the applicant to share the link with the spouse.

MargretFinnboga (approved):
B-E-A-Utiful 🤩

PR #7554, opened 2022-05-31
feat(passport): Passport Application

albinagu:
WIP - first phase of Passport Application with mocked data, behind feature flag

## Checklist:


darrikonn (approved):
Core files look good.

libs/application/template-api-modules/src/lib/modules/templates/passport/passport.service.ts
```diff
@@ -0,0 +1,125 @@
...
+import { TemplateApiModuleActionProps } from '../../../types'
+import { getValueViaPath } from '@island.is/application/core'
+import {
+  PASSPORT_CHARGE_CODES,
+  YES,
+  NO,
+  YesOrNo,
+  DiscountCheck,
+} from './constants'
+import * as kennitala from 'kennitala'
```
kksteini, line 14: Why not 
```javascript
import { info } from 'kennitala'
```
?

libs/application/templates/passport/src/lib/PassportTemplate.ts
```diff
@@ -6,84 +6,109 @@ import {
...
   stateMachineConfig: {
-    initial: 'draft',
+    initial: States.DRAFT,
     states: {
       draft: {
         meta: {
-          name: 'Umsókn um vegabréf',
+          name: m.formName.defaultMessage,
           progress: 0.33,
           lifecycle: DefaultStateLifeCycle,
```
obmagnusson, line 32: All the lifecycle settings are on default lifecycle which means the data persists forever. I recommend you update those if they are not correct.

obmagnusson (changes requested):

libs/application/templates/passport/src/lib/PassportTemplate.ts
```diff
@@ -6,84 +6,113 @@ import {
...
+          [DefaultEvents.SUBMIT]: { target: States.DONE },
         },
       },
-      approved: {
+      [States.DONE]: {
         meta: {
-          name: 'Approved',
+          name: 'Done',
           progress: 1,
           lifecycle: DefaultStateLifeCycle,
```
obmagnusson, line 102: Should we not prune the applicatation at some point when we are done. Is there a reason for us to hold on to the datal.

obmagnusson (changes requested):
Forgot to tag these lines.

libs/application/templates/passport/src/lib/PassportTemplate.ts
```diff
@@ -6,84 +6,113 @@ import {
...
       draft: {
         meta: {
-          name: 'Umsókn um vegabréf',
+          name: m.formName.defaultMessage,
           progress: 0.33,
-          lifecycle: DefaultStateLifeCycle,
+          lifecycle: {
+            shouldBeListed: false,
+            shouldBePruned: true,
+            whenToPrune: 24 * 3600 * 1000,
```
Thorrik58, line 35: You could use the EphemeralStateLifeCycle

libs/application/templates/passport/src/lib/PassportTemplate.ts
```diff
@@ -92,7 +121,7 @@ const PassportTemplate: ApplicationTemplate<
     },
   },
   mapUserToRole(_id: string, _application: Application): ApplicationRole {
-    return 'applicant'
+    return Roles.APPLICANT
```
Thorrik58, line 124: This is going to assign every user the role Applicant even if that user does not own the application. You'll need to check the applicant.id to avoid this like its done in most other applications.

Thorrik58 (changes requested):
One issue with the map user to role.

libs/application/templates/passport/src/lib/PassportTemplate.ts
Thorrik58, line 124: Can you check the applicant id and return the role and return undefined if it doesnt match the user?
```
mapUserToRole(
    nationalRegistryIdOfAuthenticatedUser: string,
    application: Application,
  ): ApplicationRole | undefined {
    if (nationalRegistryIdOfAuthenticatedUser === application.applicant) {
      return Roles.APPLICANT
    }
    return undefined
  },
```

obmagnusson (approved):

PR #7556, opened 2022-05-31
feat(j-s): Iteration on the modify ruling page

oddsson:
# Iteration on the modify ruling page

[Asana](https://app.asana.com/0/1199153462262248/1201976501087692/f)

## What

Iteration on the modify ruling page. In this iteration we

- Disable inputs that should never be allowed to change after a verdict is made.
- Remove the functionality that make the back button say "Cancel" instead of "Back"
- Add a alert message to the top of the modify ruling page that alerts the user that changes are indeed saved as they are typed
- Force regeneration of the rulingPDF on the modify ruling screen.

## Why

These are UX issues that came up during acceptance testing.

## Checklist:


apps/judicial-system/api/src/app/modules/file/file.service.ts
```diff
@@ -33,7 +34,9 @@ export class FileService {
     headers.set('cookie', req.headers.cookie as string)
 
     const result = await fetch(
-      `${this.config.backendUrl}/api/case/${id}/${route}`,
+      `${this.config.backendUrl}/api/case/${id}/${route}${
+        forceRegeneration ? '?forceRegeneration=true' : ''
```
oddsson, line 38: Having this as a query parameter seems a bit iffy to me 🤔 Being able to control whether or not the pdf is re-generated or not by changing the URL seems a bit weird. What do you think @gudjong, @Arnarkari93?

gudjong: I think it is fine. We could flip the logic and have a parameter `useSigned` with default true on the backend. So by setting that to false the server would not look for a signed version of the ruling.

The issue is not really about generating or not generating. It is about getting the signed ruling or not.

apps/judicial-system/backend/src/app/modules/case/test/caseController/getRulingPdf.spec.ts
```diff
@@ -52,15 +63,20 @@ describe('CaseController - Get ruling pdf', () => {
     const theCase = { id: caseId } as Case
     const res = {} as Response
 
-    beforeEach(async () => {
+    it('should lookup pdf', async () => {
```
gudjong, line 66: Please preserve the test pattern used and only call givenWhenThen in a beforeEach clause.

PR #7558, opened 2022-05-31
fix(service-portal): Use correct attribute for svg image

thordurhhh:
## What

Update stop-color attribute

## Why

Previous one was wrong and causing an error

## Checklist:


thorkellmani (approved):
wow

PR #7560, opened 2022-05-31
fix(driving-license-servive): Handle temporary license resubmit

alexdiljar:
It has happened due to payment service bugs or some other hick-ups in third party services that temporary driving license applications have need to been resubmitted since the 200 response got lost in translation. The resubmit returns a 400 error with a needed error code that the generated api does not map correctly. So if the submit returns a 400 error we recheck the canApplyFor function to look for the error code "HAS_B_CATEGORY".

And minor text fix.


libs/clients/driving-license/src/lib/drivingLicenseApi.service.ts
```diff
@@ -222,27 +221,40 @@ export class DrivingLicenseApi {
...
+    } catch (e) {
+      // In cases where first submit fails then resubmission gets 400 response
+      // The generated api does not map the error correctly so we check if the canApply status has changed to "HAS_B_Category"
+      if ((e as { status: number })?.status === 400) {
+        const hasTemp = await this.getCanApplyForCategoryTemporary({
+          nationalId: params.nationalIdApplicant,
+        })
+        return hasTemp.errorCode === 'HAS_B_CATEGORY'
+      }
+      return false
```
kristofer-lp8f, line 256: Rethrow error

PR #7595, opened 2022-06-02
E2e/login command

kristofer-lp8f:
## What

- Add login command for `system-e2e` project
- Add basic example of usage

## Why
Simplifies logging in, and saves network/testing time with `cy.session`.

## Checklist:

- [] I have performed a self-review of my own code


package.json
```diff
@@ -374,7 +374,7 @@
     "cross-spawn": "7.0.3",
     "csstype": "2.6.13",
     "csv-parse": "4.16.0",
-    "cypress": "8.4.1",
+    "cypress": "9.6",
```
busla, line 377: looks like there is a list of breaking changes when upgrading to 9.x, we might want to review those.

https://docs.cypress.io/guides/references/changelog#9-0-0

kristofer-lp8f: I don't see anything here that's relevant to us

apps/system-e2e/src/support/commands.ts
```diff
@@ -1,28 +1,45 @@
+const authDomain = 'identity-server.dev01.devland.is/'
 const testEnviron = Cypress.env('testEnvironment')
```
saevarma, line 2: `testEnviron` is unused

apps/system-e2e/src/support/index.d.ts
```diff
@@ -8,6 +8,10 @@ declare namespace Cypress {
      * Custom command to log in using cognito
      * @example cy.loginViaCognito()
```
saevarma, line 9: Outdated function name in comment

saevarma (changes requested):
The missing binary issue in service-portal-e2e was caching related issue. But please address the broken custom command failure in financial-aid-web-osk-e2e.

package.json
oddsson: I might be wrong here, but I believe that Cypress is bundled together with NX and in order for us to update out Cypress version we need to update the `@nrwl/cypress` package instead of the one you are updating. In fact, I think the only reason Cypress is a dependency is to extract some Cypress types 🤔 

Screen Shot 2022-06-02 at 13 51 37

kristofer-lp8f: There are also some depending on `@nrwl/cypress` for preprocessing, and for running the Cypress tests in nx.

apps/system-e2e/src/support/commands.ts
```diff
@@ -1,28 +1,44 @@
-const testEnviron = Cypress.env('testEnvironment')
+const authDomain = 'identity-server.dev01.devland.is/'
```
pshomov, line 1: Hardcoded to Dev env as is, needs to support Staging as well

saevarma (approved):

scripts/run-cypress-tests.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
+set -euo pipefail
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+CYPRESS_BIN=$(realpath "${DIR}/../node_modules/.bin/cypress")
+ENV_FILE="$(realpath "$DIR"/../.env.secret)"
```
pshomov, line 6: realpath is not present on default macOS setup

workspace.json
```diff
@@ -9535,7 +9535,7 @@
             "exclude": ["**/node_modules/**", "!apps/system-e2e/**/*"]
           }
         },
-        "system-test": {
+        "e2e": {
```
pshomov, line 9538: That's change in the docker image name as well so we should rather not do it.

apps/system-e2e/src/integration/service-portal/homepage.spec.ts
```diff
@@ -1,10 +1,10 @@
+const phoneNumber = Cypress.env('PHONE_NUMBER')
+
 describe('Home page', () => {
-  before(() => {
-    cy.ensureLoggedIn({ url: '/minarsidur/' })
```
pshomov, line 3: I think we still need to ensure the Cognito login is taken care of in environments where that is present.

pshomov (changes requested):
Very good, really close now 🏁

busla:
ensureLoggedIn is gone, I will put it back

apps/system-e2e/src/integration/service-portal/homepage.spec.ts
kristofer-lp8f: Swapped `ensureLoggedIn` out for `cognitoLogin` and made it run before each test.

kristofer-lp8f:
> ensureLoggedIn is gone, I will put it back

Removed places where it was used instead.

pshomov (approved):
LGTM

apps/system-e2e/src/plugins/index.js
```diff
@@ -1,10 +1,13 @@
...
 module.exports = (on, config) => {
-  config.env.testEnvironment = testEnvironment
-  config.env.cognito_username = process.env.AWS_COGNITO_USERNAME ?? 'system-e2e'
-  config.env.cognito_password = process.env.AWS_COGNITO_PASSWORD
-  if (testEnvironment) {
-    config.baseUrl = `https://beta.${config.env.testEnvironment}01.devland.is`
-  }
+  const testEnvironment = process.env.TEST_ENVIRONMENT || 'local'
+  config.baseUrl = config.env[testEnvironment].baseUrl
+  console.log(config)
```
kksteini, line 4: Hebb ebb ebb, console log?

albinagu (approved):

apps/system-e2e/src/plugins/index.js
busla: fixed

gudjong (approved):

package.json
Arnarkari93: Looking at the yarn.lock `nrwl/cypress` is not even dependent on cypress at all. So with this, we are upgrading the cypress version.

PR #7557, opened 2022-05-31
chore(service-portal): Remove vehicle detail title. Update some default strings

thordurhhh:
## What

* Remove vehicle detail subtitle
* Update few default strings

## Why

* No subtitle should be displayed on the vehicle detail page.

## Checklist:

- [x] I have performed a self-review of my own code
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] Formatting passes locally with my changes
- [x] I have rebased against main before asking for a review

PR #7548, opened 2022-05-30
feat(service-portal): Vehicle update

thordurhhh:
## What

* Update vehicle API
* Update vehicle UI
* Add vehicle Scope

## Why

* Updates to design
* Updates to API
* Missing scope

## Checklist:

- [x] I have performed a self-review of my own code
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] Formatting passes locally with my changes
- [x] I have rebased against main before asking for a review

PR #7496, opened 2022-05-23
feat(service-portal): vehicle additions

thordurhhh:
## What

Additions and updates to the vehicle api and ui.
This feature is behind a feature-flag.

## Why

API updated. More information added.

## Checklist:

- [x] I have performed a self-review of my own code
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] Formatting passes locally with my changes
- [x] I have rebased against main before asking for a review

PR #7597, opened 2022-06-02
feat(service-portal): Hotfix vehicle additions.

thordurhhh:
I am aware this looks more like a feature-release than a hotfix.

As the vehicle API was updated after the last release. A request has come from Island.is to update the vehicle section on /minarsidur so the feature can be released (it's currently feature-flagged). A hotfix was requested because there are a couple of weeks until the next release.

Hotfix:

https://github.com/island-is/island.is/pull/7496
https://github.com/island-is/island.is/pull/7548
https://github.com/island-is/island.is/pull/7557
https://github.com/island-is/island.is/pull/7589
https://github.com/island-is/island.is/pull/7599


disaerna (approved):
LGTM 💯

PR #7606, opened 2022-06-02
refactor(service-portal): move delegation + logout buttons

disaerna:
# Service Portal - Sidebar & User Menu changes

## What
* Move delegation button to sidebar
* Add logout button to bottom of sidebar
* Add logout button to bottom of mobile menu

## Why
* Users were having trouble finding delegation button in user testing.

## Screenshots

Sidebar: 
Screenshot 2022-06-07 at 10 47 15

Mobile menu:
Screenshot 2022-06-07 at 10 47 54

User menu:
Screenshot 2022-06-09 at 11 27 32

## Checklist:


libs/service-portal/settings/access-control/src/index.ts
```diff
@@ -12,11 +12,16 @@ export const accessControlModule: ServicePortalModule = {
   name: 'Aðgangsstýring',
   widgets: () => [],
   routes: ({ userInfo }) => {
+    const isCompany = userInfo.profile.subjectType === 'legalEntity'
+    const isDelegation = Boolean(userInfo.profile.actor)
```
osk, line 16: These checks are done more than once and are not excactly the same. Would recommend creating a helper that ensures consistency, eases refactoring, and allows for a centralized place for documenting behavior (i.e. it's not obvious why `isDelegation = Boolean(userInfo.profile.actor)`)

osk (approved):

libs/service-portal/settings/access-control/src/index.ts
disaerna: Agree 🚀 Will refactor in another PR

raRaRa:
Looks good! There's definitely room for refactoring the `NavItem` & `LogOutItem` components since they share so much code. There was another thing I noticed which can be improved as well, which is to remove the `onMouseOver` & `onMouseLeave` events to style the hover effect. Instead we should add one style to the parent `Box` that has a style called `styles.navItem`, that handles the hover effect with a selector, and a globalStyle to target the svg.

Here's the code for the styles:

```
export const navItem = style({
  ':hover': {
    color: theme.color.blue400,
    backgroundColor: theme.color.blue100,
  },
})

globalStyle(`${navItem}:hover svg`, {
  color: theme.color.blue400,
})
```

Using `globalStyle` is great when you need to target a children, e.g. the svg in this case.


The Box component would end up with these styles:

```
      className={[
        styles.navItem,
        styles.navItemActive[navItemActive],
      ]}
```

We could just create a refactoring task for later, but if you feel like it then I challenge you to try the style improvement :)

raRaRa (approved):
Looks good! There's definitely room for refactoring the `NavItem` & `LogOutItem` components since they share so much code. There was another thing I noticed which can be improved as well, which is to remove the `onMouseOver` & `onMouseLeave` events to style the hover effect. Instead we should add one style to the parent `Box` that has a style called `styles.navItem`, that handles the hover effect with a selector, and a globalStyle to target the svg.

Here's the code for the styles:

```
export const navItem = style({
  ':hover': {
    color: theme.color.blue400,
    backgroundColor: theme.color.blue100,
  },
})

globalStyle(`${navItem}:hover svg`, {
  color: theme.color.blue400,
})
```

Using `globalStyle` is great when you need to target a children, e.g. the svg in this case.


The Box component would end up with these styles:

```
      className={[
        styles.navItem,
        styles.navItemActive[navItemActive],
      ]}
```

We could just create a refactoring task for later, but if you feel like it then I challenge you to try the style improvement :)

PR #7635, opened 2022-06-03
feat(j-s): SQS Setup

gudjong:
# SQS Setup

https://app.asana.com/0/1199153462262248/1201777996978577/f

## What

- Creates a new Nest application for processing messages from a dedicated judicial system SQS queue.
- Updates Helm charts to deploy the new Nest application.

## Why

This will facilitate the decoupling of backend processing, such as sending notifications and uploading files.

## Checklist:


apps/judicial-system/backend/infra/judicial-system-backend.ts
```diff
@@ -54,6 +54,10 @@ export const serviceSetup = (): ServiceBuilder<'judicial-system-backend'> =>
         staging: 'https://judicial-system.staging01.devland.is/verjandi/',
         prod: 'https://rettarvorslugatt.island.is/verjandi/',
       },
+      SQS_QUEUE_NAME: 'sqs-judicial-system',
+      SQS_DEAD_LETTER_QUEUE_NAME: 'sqs-judicial-system-dlq',
+      SQS_REGION: 'eu-west-1',
+      SQS_ENABLED: { dev: 'true', staging: 'false', prod: 'false' },
```
pshomov, line 60: We already have server-side feature toggles that seem should fit your usage - https://docs.devland.is/technical-overview/devops/service-setup#server-side-feature-flags

apps/judicial-system/message-handler/infra/judicial-system-message-handler.ts
```diff
@@ -0,0 +1,20 @@
...
+    .env({
+      SQS_QUEUE_NAME: 'sqs-judicial-system',
+      SQS_DEAD_LETTER_QUEUE_NAME: 'sqs-judicial-system-dlq',
+      SQS_REGION: 'eu-west-1',
+      BACKEND_URL: ref((h) => `http://${h.svc(services.backend)}`),
+    })
+    .secrets({
+      BACKEND_ACCESS_TOKEN: '/k8s/judicial-system/BACKEND_ACCESS_TOKEN',
+    })
+    .replicaCount({ min: 1, max: 1, default: 1 })
```
pshomov, line 18: Why only one replica?

apps/judicial-system/backend/infra/judicial-system-backend.ts
gudjong: Switched to using server-side feature toggles. Not sure its correctly done thought.

apps/judicial-system/message-handler/infra/judicial-system-message-handler.ts
gudjong: No real reason. I will remove this.

apps/judicial-system/backend/src/app/modules/case/case.controller.ts
```diff
@@ -222,6 +235,15 @@ export class CaseController {
       state !== CaseState.DELETED,
     )
 
+    if (
+      updatedCase &&
+      completedCaseStates.includes(updatedCase.state) &&
+      ServerSideFeatureClient.isOn('judicial-system-sqs')
```
saevarma, line 241: You should be using the `FeatureFlagService` from the `nest/feature-flags` library, it also takes in the user object to check if a specific user has access to the feature.

infra/src/uber-charts/judicial-system.ts
```diff
@@ -11,9 +12,10 @@ const jsApi = jsApiSetup({ backend: jsBack })
 const jsWeb = jsWebSetup({ api: jsApi })
 const jsXrdApi = jsXrdApiSetup({ backend: jsBack })
 const jsScheduler = jsSchedulerSetup({ backend: jsBack })
+const jsMessageHndler = jsMessageHandlerSetup({ backend: jsBack })
```
saevarma, line 15:
```suggestion
const jsMessageHandler = jsMessageHandlerSetup({ backend: jsBack })
```

saevarma (changes requested):

Arnarkari93 (approved):

apps/judicial-system/backend/src/app/modules/case/case.controller.ts
gudjong: I am using the `@island.is/feature-flags`. Why is that not good enough?

pshomov (approved):
LGTM

saevarma (approved):

PR #7639, opened 2022-06-05
feat(web): Hide the Watson chat bot launcher if the locale is not 'is'

RunarVestmann:
# Hide the Watson chat bot launcher if the locale is not 'is'

## What

* Now the chat bubble for the watson chat bot won't appear when locale is not 'is'

## Why

* The chat bot is only in Icelandic right now so we want to not show it when the locale is set to english

## Checklist:


kksteini (approved):
LGTM

PR #7661, opened 2022-06-08
feat(j-s): Log when a defender opens a case

oddsson:
# Log when a defender opens a case

[Asana](https://app.asana.com/0/1199153462262248/1202086011710896/f)

## What

Log when a defender opens a case and display an alert for court users and prosecutors.

## Screenshots / Gifs

Screenshot 2022-06-08 at 10 49 31

## Checklist:


apps/judicial-system/web-e2e/src/integration/Court/InvestigationCases/overview.spec.ts
```diff
@@ -60,6 +61,10 @@ describe(`${IC_OVERVIEW_ROUTE}/:id`, () => {
     intercept(caseDataAddition)
   })
 
+  it('should let the user know if the assigned defender has viewed the case', () => {
+    cy.getByTestid('alertMessage').should('not.match', ':empty')
```
Arnarkari93, line 65: can we be more specific here with the alert message? Are we allowed to pass in a custom data-testid into the alert?

oddsson: There is currently no testid prop in the Alert component, but we could change that. How about adding an `id` prop that sets both an id and a data-testid? Or maybe just passing in a specific testid prop? What do you think?

Arnarkari93: It would be nice to add the option so we know what `alertMessage` is being rendered. I think having a specific `data-testid` prop is the better option.

Arnarkari93 (approved):

PR #7679, opened 2022-06-09
feat(service-portal): remove "eldri utgafa"

disaerna:
# Service portal - Remove "eldri útgáfa"

## What

Remove "Eldri útgáfa" link from service portal home/greetings page

## Why

New "Mínar síður" have taken over and no need to link older pages 🥳 

## Checklist:


thordurhhh (approved):
Nice. 

I actually think we should follow the best practice for cypress testing and give elements that don't necessarily need CSS attributes for other than testing, a `data-testid` or `data-cy` attributes instead.

https://docs.cypress.io/guides/references/best-practices#Selecting-Elements

PR #7697, opened 2022-06-10
feat(web): Rank increase for articles in search results

RunarVestmann:
# Rank increase for articles in search results

## What

* News are appearing above articles in search results so I boosted articles so they'll appear at the top of search results

## Why

* This was a requested feature (to not have news appearing at the top and giving articles a bigger role when it comes to search results)
* I wasn't able to only decrease the weight for news so my solution was to boost articles instead (perhaps we should consider in the future to give "weights" to the different types)

### Before 
image

### After
image

## Checklist:


libs/content-search-toolkit/src/queries/search.ts
```diff
@@ -4,6 +4,13 @@ import { TagQuery, tagQuery } from './tagQuery'
 import { typeAggregationQuery } from './typeAggregation'
 import { processAggregationQuery } from './processAggregation'
 
+const getBoostForType = (type: string, defaultBoost: string | number = 1) => {
+  if (type === 'webArticle') {
+    return 55
```
kristofer-lp8f, line 9: Could you add a comment on why you chose the number `55`?

kristofer-lp8f (approved):
LGTM 🍦

PR #7715, opened 2022-06-13
fix(skuldleysisvottord): Add possibility to hide submit error for specific

johannaagma:
## What

Hide the message "Eitthvað fór úrskeiðis" here for the No Debt Certificate application:

image

if the only error we got from providers is that the user is not debtless (its not really an error, but a message from the FJS api that user cannot receive a No Debt Certificate, since he is not debtless).

Will look like this after:

image

But if there is any other error (from another provider) it will look like this:

image

## Why

Request from Vésteinn to hide this field, so it is more user friendly if user is not debtless.

## Checklist:


obmagnusson (commented):
Small change

obmagnusson (approved):

unnursolingimars (approved):

PR #7716, opened 2022-06-13
fix(fa): Only show second part of approved comment

davidpxn:
# ... 👆

https://app.asana.com/0/1202244723867117/1202244723867128

## What

Only show second part of approved comment

## Why

The first part of the comment is the final amount, which is already displayed in the breakdown below on the status page.

## Checklist:


libs/application/templates/financial-aid/src/fields/Status/ApprovedAlert/ApprovedAlert.tsx
```diff
@@ -21,7 +21,7 @@ const ApprovedAlert = ({ events }: Props) => {
       return events.find((el) => el.eventType === ApplicationEventType.APPROVED)
         ?.comment
     }
-  }, [events])
+  }, [events])?.split('\n')[1]
```
MargretFinnboga, line 24: why only show part of the approved comment?

MargretFinnboga (commented):
🤔

libs/application/templates/financial-aid/src/fields/Status/ApprovedAlert/ApprovedAlert.tsx
MargretFinnboga: I want to challenge this, I think we should just show the full comment, not all employees are going to remember to put a new line after approving, I think this can easily wind up showing the wrong comment or no comment. And also what if they put two new lines, then you will only be showing the one? 

If they really do insist on removing the first line, than I rather see that the you split the array and remove the first element and show the rest

davidpxn: Nice, well spotted! I forgot to think about the case where the user enters multiple lines 🤕

MargretFinnboga (approved):
🌻

PR #7718, opened 2022-06-13
fix(api): Provide getIdentityWithFallback method in identity service

saevarma:
## What

Provides a new method `getIdentityWithFallback` to use to always get identity object response, instead of null or error.

## Why

When getting delegations, the `to` and `from` fields are resolved using the identity resolver. When National Registry is down it causes error to happen, so instead we want to use fallback values to always get an identity object.

## Screenshots / Gifs

## Checklist:


eirikurn (approved):
LGTM 🎉

PR #7719, opened 2022-06-13
feat(web): Added support for fetching english related articles

RunarVestmann:
# Added support for fetching english related articles

## What

* Added a way to fetch related articles for different locales

## Why

* Before this change you couldn't see related articles on english locale

### Before
image

### After
image

## Checklist:

### DIFF
--- a/libs/cms/src/lib/cms.resolver.ts
+++ b/libs/cms/src/lib/cms.resolver.ts
@@ -85,6 +85,7 @@ import { GetSupportCategoriesInput } from './dto/getSupportCategories.input'
 import { GetSupportCategoriesInOrganizationInput } from './dto/getSupportCategoriesInOrganization.input'
 import { GetPublishedMaterialInput } from './dto/getPublishedMaterial.input'
 import { EnhancedAssetSearchResult } from './models/enhancedAssetSearchResult.model'
+import { Locale } from '@island.is/shared/types'
 
 const { cacheTime } = environment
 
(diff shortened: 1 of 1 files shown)


libs/cms/src/lib/cms.resolver.ts
```diff
@@ -347,13 +348,15 @@ export class CmsResolver {
 
   @Directive(cacheControlDirective())
   @Query(() => Article, { nullable: true })
-  getSingleArticle(
-    @Args('input') { lang, slug }: GetSingleArticleInput,
-  ): Promise<Article | null> {
-    return this.cmsElasticsearchService.getSingleDocumentTypeBySlug<Article>(
+  async getSingleArticle(@Args('input') { lang, slug }: GetSingleArticleInput) {
```
kristofer-lp8f, line 351: Why remove the return type?

libs/cms/src/lib/cms.resolver.ts
```diff
@@ -497,7 +500,10 @@ export class ArticleResolver {
 
   @Directive(cacheControlDirective())
   @ResolveField(() => [Article])
-  async relatedArticles(@Parent() article: Article) {
-    return this.cmsContentfulService.getRelatedArticles(article.slug, 'is')
+  async relatedArticles(@Parent() article: Article & { lang?: Locale }) {
```
kristofer-lp8f, line 503: Could you add a return type?

kristofer-lp8f (changes requested):
Good job 😄 

Just minor type annotation comments

libs/cms/src/lib/cms.resolver.ts
RunarVestmann: The type is inferred but ok, I've added a type annotation now

kristofer-lp8f (approved):

PR #7742, opened 2022-06-14
chore(j-s): Judicial System Message Library

gudjong:
# Judicial System Message Library

[Vista gögn sjálfkrafa í Auði eftir afgreiðslu ef þau hafa ekki verið vistuð](https://app.asana.com/0/1199153462262248/1201777996978577/fO

## What

Creates a judicial system message library in preparation for adding various messaging utilities.

## Why

The elements defined in this library will be shared between judicial system applications.

## Checklist:


apps/judicial-system/backend/src/app/modules/case/case.controller.ts
```diff
@@ -26,11 +26,11 @@ import {
   SigningServiceResponse,
 } from '@island.is/dokobit-signing'
 import { InjectQueue, QueueService } from '@island.is/message-queue'
+import { MessageType } from '@island.is/judicial-system/message'
```
oddsson, line 29: `MessageType` and `completedCaseStates` are not being used in this file. Please remove them.

oddsson (changes requested):
There are a couple of unused imports

apps/judicial-system/backend/src/app/modules/case/case.controller.ts
gudjong: They are not used because we commented out the queuing functionality. I will comment these out as well, but not remove them.

oddsson (approved):

apps/judicial-system/backend/src/app/modules/case/case.controller.ts
```diff
@@ -26,11 +26,11 @@ import {
   SigningServiceResponse,
 } from '@island.is/dokobit-signing'
 import { InjectQueue, QueueService } from '@island.is/message-queue'
+// import { MessageType } from '@island.is/judicial-system/message'
```
saevarma, line 29: Remove redundant code?

apps/judicial-system/backend/src/app/modules/case/case.controller.ts
```diff
@@ -26,11 +26,11 @@ import {
...
 import {
   CaseState,
   CaseType,
-  completedCaseStates,
-  MessageType,
+  // completedCaseStates,
```
saevarma, line 33: Remove?

saevarma (changes requested):

apps/judicial-system/backend/src/app/modules/case/case.controller.ts
gudjong: These imports are temporarily commented out along with some queuing code that is not working because the queuing config on dev is incomplete. We don't want to remove the code because it will be uncommented as soon as Andes has fixed the config.

apps/judicial-system/backend/src/app/modules/case/case.controller.ts
gudjong: See above.

saevarma (approved):

PR #7750, opened 2022-06-15
refactor(service-portal): refactor delegation check

disaerna:
# Refactor delegation check in service portal

## What

Remove redundant checks and move it to shared utils.

(Was first move to service portal core but shared components, as in all User menu stuff, cannot depend on service portal!) 

## Why

More maintainable and readable to have it in one centralizes area.

## Screenshots / Gifs

## Checklist:


Valur (approved):

libs/shared/components/src/auth/UserMenu/UserButton.tsx
```diff
@@ -10,19 +10,17 @@ import { User } from '@island.is/shared/types'
 import { useLocale } from '@island.is/localization'
 import { userMessages } from '@island.is/shared/translations'
 import * as styles from './UserMenu.css'
+import { checkDelegation } from '@island.is/service-portal/core'
```
eirikurn, line 13: `shared/components` should not depend on `service-portal/*`, can we move this utility somewhere else?

eirikurn (changes requested):

libs/shared/components/src/auth/UserMenu/UserButton.tsx
disaerna: Woops! I'll fix and move it to shared utils instead.

PR #7752, opened 2022-06-15
feat(ads): Polling every minute instead of every ten seconds

kksteini:
# What
Users poll the graphql api every 10 seconds once on Air Discount Scheme.

# Why the change
Reduce request load by sixfold.

# Impact on users
Some users may figure out that they either have to wait or manually refresh themselves in the case where they are trying to use multiple discount codes in quick succession.

## Checklist:


rafnarnason (approved):
lgtm

PR #7757, opened 2022-06-15
fix(service-portal): hide access control if delegation

disaerna:
# Service portal - Access control

## What

If user is an actor for someone else or company - hide Access control from sidebar

## Why

Otherwise, it is confusing.

## Screenshots / Gifs

## Checklist:

### DIFF
--- a/libs/service-portal/settings/access-control/src/index.ts
+++ b/libs/service-portal/settings/access-control/src/index.ts
@@ -30,17 +30,13 @@ export const accessControlModule: ServicePortalModule = {
       {
         name: m.accessControlGrant,
         path: ServicePortalPath.SettingsAccessControlGrant,
-        enabled:
-          showDelegation &&
-          userInfo.scopes.includes(AuthScope.writeDelegations),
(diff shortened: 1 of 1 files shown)


libs/service-portal/settings/access-control/src/index.ts
```diff
@@ -14,26 +14,29 @@ export const accessControlModule: ServicePortalModule = {
   routes: ({ userInfo }) => {
     const isCompany = userInfo.profile.subjectType === 'legalEntity'
     const isDelegation = Boolean(userInfo.profile.actor)
+    const showDelegation =
+      isCompany || isDelegation
```
eirikurn, line 18: Why are we checking this?

The scope check should be enough. If it is not correct in some case then that indicates a security issue we need to fix.

disaerna: I agree - This has been done a little bit differently in another PR that has been merged already, so i will delete this PR. 
I'll double check that we can trust the scopes only and refactor these checks.

PR #7763, opened 2022-06-15
fix(j-s): temporarly disable queue

Arnarkari93:
## What
- Temporarily disable queue
## Why
- It is causing crash loop

## Checklist:


apps/judicial-system/backend/src/app/modules/case/case.controller.ts
```diff
@@ -77,8 +69,7 @@ export class CaseController {
     private readonly caseService: CaseService,
     private readonly userService: UserService,
     private readonly eventService: EventService,
-    @Inject(LOGGER_PROVIDER) private readonly logger: Logger,
-    @InjectQueue(caseModuleConfig().sqs.queueName) private queue: QueueService,
+    @Inject(LOGGER_PROVIDER) private readonly logger: Logger, // @InjectQueue(caseModuleConfig().sqs.queueName) private queue: QueueService,
```
Arnarkari93, line 72: Removing constructor call to QueueService.

gudjong (approved):

PR #7766, opened 2022-06-15
chore(application-login-service): Implement new company search in login service

egill15:
# Company search in login service

[https://app.clickup.com/t/2fkekvu](https://app.clickup.com/t/2fkekvu)
## What
Make the old company search code obsolete by implementing the new company search feature into the application.

## Why
Better code practice to use reusable functions. And it keeps things from being bloated.

## Checklist:


libs/application/templates/login-service/src/fields/IsatInvalid/IsatInvalid.tsx
```diff
@@ -0,0 +1,48 @@
...
+  Text,
+} from '@island.is/island-ui/core'
+import { useLocale } from '@island.is/localization'
+import { errorMessages } from '../../lib/messages'
+
+export const IsatInvalid = ({ application }: FieldBaseProps) => {
+  const { formatMessage } = useLocale()
+  return (
+    <>
+      <GridColumn span={['1/1', '1/1', '1/1']} paddingTop={2}>
```
saevarma, line 17: `span` accepts a single value and as the value is always the same you can do instead
```suggestion
      <GridColumn span='1/1' paddingTop={2}>
```

libs/application/templates/login-service/src/shared/types.ts
```diff
@@ -0,0 +1,7 @@
+export type selectCompanySearchField = {
+  searchField: {
+    isat: string
+    nationalId: string
+    label: string
+  }
+}
```
saevarma, line 7: This type is duplicated here and in the `CompanySearchFromField` we should try to reuse the type.

libs/shared/form-fields/src/lib/CompanySearchController/CompanySearchController.tsx
```diff
@@ -94,6 +111,34 @@ export const CompanySearchController: FC<Props> = ({
...
+  const getIsatNumbers = (nationalId: string) => {
+    if (!shouldIncludeIsatNumber) return ''
+
+    const companies: Array<any> = data?.companyRegistryCompanies?.data ?? []
+    if (companies.length === 0) return ''
+
+    const filteredCompany = companies.filter((company) => {
+      if (company.nationalId === nationalId) return true
+      return false
+    })
```
saevarma, line 123: This can be simplified to
```suggestion
    const filteredCompany = companies.filter((company) => company.nationalId === nationalId )
```

libs/shared/form-fields/src/lib/CompanySearchController/CompanySearchController.tsx
```diff
@@ -94,6 +111,34 @@ export const CompanySearchController: FC<Props> = ({
     return options || []
   }
 
+  const getIsatNumbers = (nationalId: string) => {
+    if (!shouldIncludeIsatNumber) return ''
+
+    const companies: Array<any> = data?.companyRegistryCompanies?.data ?? []
```
saevarma, line 117: You could let gql generate typed hooks for the query to get typed responses.

libs/shared/form-fields/src/lib/CompanySearchController/CompanySearchController.tsx
```diff
@@ -94,6 +111,34 @@ export const CompanySearchController: FC<Props> = ({
     return options || []
   }
 
+  const getIsatNumbers = (nationalId: string) => {
```
saevarma, line 114: The function name indicates return of string array of ISAT numbers, but it only returns a single string with the main ISAT number
```suggestion
  const getIsatNumber = (nationalId: string) => {
```

libs/shared/form-fields/src/lib/CompanySearchController/CompanySearchController.tsx
```diff
@@ -94,6 +111,34 @@ export const CompanySearchController: FC<Props> = ({
...
+          if (c.type === 'Aðal') {
+            isat = `${c.number} - ${c.name}`
+          }
+          return c
+        })
+      }
+      return v
+    })
+    return isat
+  }
```
saevarma, line 140: The search for the main ISAT can be improved to return when you find it
```suggestion
    for (let v of vats) {
      if (!v.dateOfDeregistration) {
        const classification = v.classification.find(c => c.type === 'Aðal')
        if (classification) {
          return `${c.number} - ${c.name}`
        }
      }
    }
    return ''
  }
```

saevarma (changes requested):

darrikonn (changes requested):
Can you clean up unused variables that this PR introduces?

libs/shared/form-fields/codegen.yml
```diff
@@ -0,0 +1,27 @@
+schema:
```
darrikonn, line 1: IMO: It does not make sense to add a codegen here. This is a shared lib, used by all applications, and it should not have any graphql fields.

libs/shared/form-fields/src/lib/CompanySearchController/CompanySearchController.tsx
```diff
@@ -3,21 +3,10 @@ import { Controller, useFormContext } from 'react-hook-form'
 import { AsyncSearch, Box, AlertMessage } from '@island.is/island-ui/core'
```
darrikonn, line 3: This component is probably the root issue. It should not be in this library, since it contains side effects that around bound to all applications.

libs/shared/form-fields/src/lib/CompanySearchController/companySearchController.graphql.ts
```diff
@@ -0,0 +1,24 @@
+import { gql } from '@apollo/client'
```
darrikonn, line 1: ditto

darrikonn (changes requested):

libs/application/ui-components/src/components/CompanySearchController/CompanySearchController.tsx
```diff
@@ -94,6 +87,35 @@ export const CompanySearchController: FC<Props> = ({
...
+      return ''
+    }
+
+    const vats = filteredCompany.companyInfo.vat
+
+    for (const v of vats) {
+      if (!v.dateOfDeregistration && v.classification) {
+        const c = v.classification.find((c) => c.type === 'Aðal')
+        if (c) {
+          return `${c.number} - ${c.name}`
```
darrikonn, line 112: This looks like the job of a field resolver that should be handled in the api.

darrikonn (approved):
Approved, but can be improved by moving the isat logic to the api in a field resolver

saevarma (approved):
LGTM, but please consider Darri's recommendation on improvement.

PR #7280, opened 2022-04-29
feat(application-system): Template Api Data providers

obmagnusson:
## What

Move data providers to the template api.

## Why

Less boilerplate and modularity. 

## Screenshots / Gifs
Should be none, because impact on end-user is none.

## Checklist:

- [ ] I have performed a self-review of my own code
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] Formatting passes locally with my changes
- [ ] I have rebased against main before asking for a review

PR #7791, opened 2022-06-16
WIP: Implement getUserProfile

gunnar-sendirad:
We weren't told what to return from this function, so it seems to make sense to just return the object we got from the client call.

That client has more functions. We had talked about only implementing the read ones, that is, the ones that use HTTP GET. Those are:

- userProfileControllerFindOneByNationalId
- userTokenControllerFindOneByNationalId
- userTokenControllerGetDeviceTokens

This commit implements the first one. We didn't (yet) do the second two because the module this gets used in is UserProfileModule.  As far as I can tell, that module is configured in SharedDataProviders, where it appears as an instance of ApplicationTemplateAPIAction. That interface has a single field called apiModuleAction, and our original module is only configured with getUserProfile as that action. So... we're done?

### DIFF
--- a/libs/application/template-api-modules/src/lib/modules/shared/data-providers/user-profile/user-profile.module.ts
+++ b/libs/application/template-api-modules/src/lib/modules/shared/data-providers/user-profile/user-profile.module.ts
@@ -11,14 +11,23 @@ import { BaseTemplateAPIModuleConfig } from '../../../../types'
 import { UserProfileService } from './user-profile.service'
 
 import { Configuration, UserProfileApi } from '@island.is/clients/user-profile'
+import { IslyklarApi, IslykillApiModule } from '@island.is/clients/islykill'
 
 export class UserProfileModule {
   static register(config: BaseTemplateAPIModuleConfig): DynamicModule {
     return {
       module: UserProfileModule,
-      imports: [SharedTemplateAPIModule.register(config)],
+      imports: [
+        SharedTemplateAPIModule.register(config),
+        IslykillApiModule.register({
+          cert: config.islykill.cert,
+          passphrase: config.islykill.passphrase,
+          basePath: config.islykill.basePath,
+        }),
+      ],
       providers: [
         UserProfileService,
+        IslyklarApi,
         {
           provide: UserProfileApi,
           useFactory: () =>
--- a/libs/application/template-api-modules/src/lib/modules/shared/data-providers/user-profile/user-profile.service.ts
+++ b/libs/application/template-api-modules/src/lib/modules/shared/data-providers/user-profile/user-profile.service.ts
@@ -1,45 +1,149 @@
 import { Auth, AuthMiddleware } from '@island.is/auth-nest-tools'
 import { UserProfileApi } from '@island.is/clients/user-profile'
-import { ProblemError } from '@island.is/nest/problem'
-import { ProblemType } from '@island.is/shared/problem'
-import { isRunningOnEnvironment } from '@island.is/shared/utils'
+import { IslyklarApi } from '@island.is/clients/islykill'
+import { logger } from '@island.is/logging'
 import { Injectable } from '@nestjs/common'
 import { SharedTemplateApiService } from '../..'
 import { TemplateApiModuleActionProps } from '../../../../types'
+import { isRunningOnEnvironment } from '@island.is/shared/utils'
+import differenceInMonths from 'date-fns/differenceInMonths'
+
+export const MAX_OUT_OF_DATE_MONTHS = 6
+
+const handleError = (error: any) => {
+  logger.error(JSON.stringify(error))
+  throw new Error(`Failed to resolve request (${error.status})`)
+}
+
+enum DataStatus {
+  NOT_DEFINED = 'NOT_DEFINED',
+  NOT_VERIFIED = 'NOT_VERIFIED',
+  VERIFIED = 'VERIFIED',
+  EMPTY = 'EMPTY',
+}
+
 @Injectable()
 export class UserProfileService {
   constructor(
     private readonly sharedTemplateAPIService: SharedTemplateApiService,
     private readonly userProfileApi: UserProfileApi,
-  ) {}
+    private readonly islyklarApi: IslyklarApi,
+  ) { }
 
-  userProfileApiWithAuth(auth: Auth) {
+  userProfileApiWithAuth(auth: Auth): UserProfileApi {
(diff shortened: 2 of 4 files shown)


libs/api/domains/user-profile/src/index.ts
```diff
@@ -1,2 +1,3 @@
 export * from './lib/userProfile.module'
 export * from './lib/userProfile.resolver'
+export { IslykillService } from './lib/islykill.service'
```
obmagnusson, line 3: Remove

libs/application/template-api-modules/src/lib/modules/shared/data-providers/user-profile/user-profile.module.ts
```diff
@@ -11,14 +11,24 @@ import { BaseTemplateAPIModuleConfig } from '../../../../types'
 import { UserProfileService } from './user-profile.service'
 
 import { Configuration, UserProfileApi } from '@island.is/clients/user-profile'
+import { IslykillService } from '@island.is/api/domains/user-profile'
```
obmagnusson, line 14: We don't want to import from the API Layer. Should be purely using the Client libs.

libs/application/template-api-modules/src/lib/modules/shared/data-providers/user-profile/user-profile.module.ts
```diff
@@ -11,14 +11,24 @@ import { BaseTemplateAPIModuleConfig } from '../../../../types'
...
+        SharedTemplateAPIModule.register(config),
+        IslykillApiModule.register({
+          cert: config.islykill.cert,
+          passphrase: config.islykill.passphrase,
+          basePath: config.islykill.basePath,
+        }),
+      ],
       providers: [
         UserProfileService,
+        IslykillService,
```
obmagnusson, line 31: Should just be able to import the Islykillmodule and inject the service. Add it to export within the module if not exported.

libs/application/template-api-modules/src/lib/modules/shared/data-providers/user-profile/user-profile.service.ts
```diff
@@ -1,45 +1,135 @@
...
-import { ProblemError } from '@island.is/nest/problem'
-import { ProblemType } from '@island.is/shared/problem'
-import { isRunningOnEnvironment } from '@island.is/shared/utils'
+import { UserProfileApi, UserProfile } from '@island.is/clients/user-profile'
+import { IslykillService } from '@island.is/api/domains/user-profile'
+import { logger } from '@island.is/logging'
 import { Injectable } from '@nestjs/common'
 import { SharedTemplateApiService } from '../..'
 import { TemplateApiModuleActionProps } from '../../../../types'
+import { ApolloError } from 'apollo-server-express'
```
obmagnusson, line 8: I think we can safely remove the Appollo Gql dependencies in here.

libs/application/template-api-modules/src/lib/modules/shared/data-providers/user-profile/user-profile.service.ts
```diff
@@ -1,45 +1,135 @@
...
+        const outOfDateEmailMobile = (emptyMail || emptyMobile) && diffOutOfDate
 
-      return {
-        email: profile.email,
-        emailVerified: profile.emailVerified,
-        mobilePhoneNumber: profile.mobilePhoneNumber,
-        mobilePhoneNumberVerified: profile.mobilePhoneNumberVerified,
-      }
-    } catch (e) {
-      if (isRunningOnEnvironment('local')) {
```
obmagnusson, line 36: We dont want everyone working on the application system having to set up and running the userprofile service and or islykill.

libs/application/template-api-modules/src/lib/modules/shared/data-providers/user-profile/user-profile.service.ts
```diff
@@ -1,45 +1,135 @@
...
   }
 
-  async getUserProfile({ auth }: TemplateApiModuleActionProps) {
-    try {
-      //TODO finish this
-      const profile = await this.userProfileApiWithAuth(
-        auth,
-      ).userProfileControllerFindOneByNationalId({
-        nationalId: auth.nationalId,
+  // eslint-disable-next-line
```
obmagnusson, line 37: Why is this needed?

obmagnusson (changes requested):

libs/api/domains/user-profile/src/index.ts
gunnar-sendirad: Gone

libs/application/template-api-modules/src/lib/modules/shared/data-providers/user-profile/user-profile.module.ts
gunnar-sendirad: Done

libs/application/template-api-modules/src/lib/modules/shared/data-providers/user-profile/user-profile.service.ts
gunnar-sendirad: Done

libs/application/template-api-modules/src/lib/modules/shared/data-providers/user-profile/user-profile.module.ts
gunnar-sendirad: Done.

libs/application/template-api-modules/src/lib/modules/shared/data-providers/user-profile/user-profile.service.ts
```diff
@@ -1,45 +1,135 @@
...
+         * Even if userProfileApiWithAuth does not exist.
+         * Islykill data might exist for the user, so we need to get that, with default values in the userprofile data.
+         */
+        return this.getIslykillProfile(auth)
+      }
+      handleError(error)
+    })
   }
-}
+}
```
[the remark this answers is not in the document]
gunnar-sendirad: \n

libs/api/domains/user-profile/src/index.ts
```diff
@@ -1,2 +1,2 @@
 export * from './lib/userProfile.module'
-export * from './lib/userProfile.resolver'
+export * from './lib/userProfile.resolver'
```
obmagnusson, line 2: New line

obmagnusson:
Implemented within #7280

PR #7685, opened 2022-06-09
fix(api): setting feature flag for INAO_client

Snojo:
# ...
Removing secret from prod that was not supposed to go there during last release. 

Regarding this one https://islandis.slack.com/archives/C03FT47CFEW/p1653401619510379

## What

Specify what you're trying to achieve

## Why

Specify why you need to achieve this

## Screenshots / Gifs

Attach Screenshots / Gifs to help reviewers understand the scope of the pull request

## Checklist:

- [ ] I have performed a self-review of my own code
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] Formatting passes locally with my changes
- [ ] I have rebased against main before asking for a review

PR #7823, opened 2022-06-20
feat(financial-statements-inao): Add application api template module

fjandakornid:
# Add application api template module for Financial Statements IANO

## What

- Added application api template module for Financial Statements INAO
- Moved client from api-module to it's own lib
- Changed connection for Dataverse from sandbox to RE environment

## Why

Needed to get current user's client type from Dataverse in applications x-state onEntry function.

## Checklist:


apps/application-system/api/infra/application-system-api.ts
```diff
@@ -74,6 +74,17 @@ export const serviceSetup = (services: {
   service('application-system-api')
     .namespace(namespace)
     .serviceAccount(serviceAccount)
+    .features({
+      INAO_Client: {
```
pshomov, line 78: I cannot find where this flag is used in the code. Unconditional use of feature-specific environment variable can be dangerous. For more info on server-side features, please see - https://docs.devland.is/libs/nest/config#server-side-features

fjandakornid: The flag isn't used in the code. This is because these variables doesn't exist in the AWS parameter store and this is to prevent them to be configured there. I was copying Snorri's work in this PR: https://github.com/island-is/island.is/pull/7685

obmagnusson (approved):
Application System files look good. Although you might want to update the codeowners file.

apps/application-system/api/infra/application-system-api.ts
pshomov: That was the part where we configure the deployment and the feature. But The code needs to respect the flag and not throw because of missing env variables, which is what happened during the last release to prod.

fjandakornid:
> @fjandakornid Is this PR "on hold"? I am not sure if I am supposed to review something here after our last conversion.

I just made the changes we talked about and it can be reviewed now. I will send you updated secret values before merging this.

saevarma (changes requested):
Please rebase with main and review the upgrade guide https://www.notion.so/Nx-13-10-5-a77cd802539e468596da6265700ff3cf.

FridrikMProgramm (approved):

andes-it:

andes-it:

PR #7831, opened 2022-06-21
feat(service-portal): Update inbox document filter

thordurhhh:
## What

* Add group to table.
* Add filter tags

## Why

Improved filtering.

## Checklist:


thordurhhh:
> Approved ✅ - wondering why renaming document -> doc though 🤔

My IDE was confusing it with the [DOM](https://developer.mozilla.org/en-US/docs/Web/API/Document_Object_Model/Introduction) at some point so I just changed it to avoid confusion 💯

PR #7851, opened 2022-06-21
fix(application): Extract array validation for arrays that can be optional

kksteini:
## What

Dataschemas with validations of the type: `z.object({example: z.string()}).array().optional()` will fail since the schema `_def` will not contain a type, but rather an array of options from which a type can be extracted.

This is because arrays marked as optional are parsed as a union type between the array and an undefined type. The partial validation was specifically extracting against simple array types.

## Why

We want to be able to validate arrays that can be undefined (in the case where a repeater does not have to contain elements).

## Checklist:


libs/application/core/src/validation/validators.ts
```diff
@@ -92,10 +94,20 @@ function partialSchemaValidation(
           try {
             trimmedSchema.parse({ [key]: [el] })
           } catch (e) {
+            let schemaShape = trimmedSchema?.shape[key]?._def?.type
```
Thorrik58, line 97: Feel like some tests would be in order for this before we go any further showing that this solves the issue mentioned in the why.

Thorrik58 (changes requested):

libs/application/core/src/validation/validators.ts
kksteini: Hehe,
Good point, on it :)

Thorrik58 (approved):
LGTM

PR #7863, opened 2022-06-22
fix(service-portal): Enabling finance menu item if the user has access to one of the nested item

saevarma:
## What

User can give delegation to either finance overview or finance schedule (or both). 
Currently the finance menu item is only looking at the finance overview scope to become enabled.

## Why

So users with access to finance schedule through delegations can navigate to the page in the menu.

## Screenshots / Gifs

## Checklist:


thordurhhh (approved):
👍

PR #7869, opened 2022-06-22
feat(test): Make proxy script cancellable

kristofer-lp8f:
## What
Make it possible to cancel running proxies using `./scripts/run-test-proxies.sh` and `Ctrl+C`.

## Why
Otherwise, killing proxies requires searching for processes running with the used ports and killing them by process IDs.

## Checklist:

### DIFF
--- a/scripts/run-test-proxies.sh
+++ b/scripts/run-test-proxies.sh
@@ -2,4 +2,8 @@
 
 set -euo pipefail
 
-./scripts/run-es-proxy.sh "$@" & ./scripts/run-soffia-proxy.sh "$@" & ./scripts/run-xroad-proxy.sh "$@" &
+./scripts/run-es-proxy.sh "$@" &
+./scripts/run-soffia-proxy.sh "$@" &
+./scripts/run-xroad-proxy.sh "$@" &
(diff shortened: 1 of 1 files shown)


scripts/run-test-proxies.sh
```diff
@@ -2,4 +2,8 @@
 
 set -euo pipefail
 
-./scripts/run-es-proxy.sh "$@" & ./scripts/run-soffia-proxy.sh "$@" & ./scripts/run-xroad-proxy.sh "$@" &
+./scripts/run-es-proxy.sh "$@" &
+./scripts/run-soffia-proxy.sh "$@" &
+./scripts/run-xroad-proxy.sh "$@" &
+
+wait
```
darrikonn, line 9: This won't do anything for the background processes.
The other commands are suffixed with `&`, so the shell executes the command in the background in a subshell.
If you kill the current shell with `ctrl+C`, the subshells will continue to run.

darrikonn (changes requested):

kristofer-lp8f:
Naive, won't work, stale. Closing.

PR #7876, opened 2022-06-23
fix(fa): Add english translation to months

davidpxn:
# ... 👆

https://app.asana.com/0/1202167941440767/1202488702554254

## What

Check on which localization the user is using and display correct months with correct language.

## Why

So the months fit in when the project will be translated in Contentful.

## Checklist:


MargretFinnboga (approved):
🥖

PR #7888, opened 2022-06-24
feat(ui-fields): Adding tag option to radio button field

albinagu:
Screenshot 2022-06-24 at 14 15 21

## Checklist:


eirikurn (approved):
LGTM. Would be good to change codeowner for shared/form-fields to either island-ui or sendiradid.

Thorrik58 (approved):

libs/island-ui/core/src/lib/RadioButton/RadioButton.tsx
```diff
@@ -129,6 +136,17 @@ export const RadioButton = ({
             <Tooltip text={tooltip} />
           </div>
         )}
+        {tag && large && (
+          <Box
+            display="flex"
+            justifyContent="flexEnd"
+            style={{ width: '-webkit-fill-available' }}
```
oddsson, line 143: This is probably not a good idea, because this will only work on Chrome. I think something like `width="full"` would be better.

oddsson (changes requested):
One thing regarding a `-webkit-` style

libs/island-ui/core/src/lib/RadioButton/RadioButton.tsx
albinagu: yes, ofcourse! good spot :)

oddsson (approved):

PR #7900, opened 2022-06-27
refactor(j-s): Replace setAndSendToServer and setAndSendDateToServer with autofill

oddsson:
# Replace setAndSendToServer and setAndSendDateToServer with autofill

[Asana](https://app.asana.com/0/1199153462262248/1202253499513154/f)

## What

- Replace all instances of `setAndSendToServer` and `setAndSendDateToServer` with `autofill`
- Simplify `UpdateCase` and `createCase` types

## Why

This was a technical dept.

## Checklist:


oddsson:
I'm thinking `force` should default to true and you'd have to turn it off, rather than `force` defaulting to false. The reasoning is that forcing is way more common. What do you think?

gudjong (changes requested):
I would like to make a suggestion for changes.

gudjong (approved):

PR #7905, opened 2022-06-27
test(system-e2e): refactoring IDS login

pshomov:
# ...

## What

## Why

## Screenshots / Gifs

## Checklist:

### DIFF
--- a/apps/system-e2e/src/support/commands.ts
+++ b/apps/system-e2e/src/support/commands.ts
@@ -23,21 +23,21 @@ const cognitoLogin = (creds: CognitoCreds) => {
   cy.url().should('contain', baseUrl)
 }
 
-const idsLogin = (phoneNumber: string) => {
+const idsLogin = (phoneNumber: string, url: string) => {
   const sentArgs = {
     args: {
       phoneNumber: phoneNumber,
       authUrl: authUrl,
-      minarSidur: `${Cypress.config().baseUrl}/minarsidur/`,
+      minarSidur: `${Cypress.config().baseUrl}${url}`,
     },
   }
   cy.patchSameSiteCookie(`${authUrl}/login/app?*`)
-  cy.visit('/minarsidur/', { log: true })
+  cy.visit(url, { log: true })
   cy.origin(authUrl, sentArgs, ({ phoneNumber }) => {
     cy.get('input[id="phoneUserIdentifier"]').type(phoneNumber)
     cy.get('button[id="submitPhoneNumber"]').click()
   })
-  cy.url().should('contain', `${Cypress.config().baseUrl}/minarsidur/`)
+  cy.url().should('contain', `${Cypress.config().baseUrl}${url}`)
 }
 
 Cypress.Commands.add('patchSameSiteCookie', (interceptUrl) => {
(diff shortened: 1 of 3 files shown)


busla (commented):
should we maybe use `urlPath` to denote that its not a full url?

apps/system-e2e/src/support/commands.ts
```diff
@@ -23,21 +23,21 @@ const cognitoLogin = (creds: CognitoCreds) => {
...
 }
 
-const idsLogin = (phoneNumber: string) => {
+const idsLogin = (phoneNumber: string, url: string) => {
   const sentArgs = {
     args: {
       phoneNumber: phoneNumber,
       authUrl: authUrl,
-      minarSidur: `${Cypress.config().baseUrl}/minarsidur/`,
+      minarSidur: `${Cypress.config().baseUrl}${url}`,
```
kristofer-lp8f, line 31: Unused `minarSidur`

apps/system-e2e/src/support/commands.ts
```diff
@@ -60,18 +60,18 @@ Cypress.Commands.add('patchSameSiteCookie', (interceptUrl) => {
   }).as('sameSitePatch')
 })
 
-Cypress.Commands.add('idsLogin', ({ phoneNumber }) => {
+Cypress.Commands.add('idsLogin', ({ phoneNumber, url }) => {
```
kristofer-lp8f, line 63: Maybe have some sane default?
```suggestion
Cypress.Commands.add('idsLogin', ({ phoneNumber, url='/' }) => {
```

apps/system-e2e/src/support/index.d.ts
```diff
@@ -33,7 +33,7 @@ declare namespace Cypress {
      * Custom command to log in using cognito and island.is SSO
      * @example cy.login()
      */
-    idsLogin({ phoneNumber: string }): Chainable<void>
+    idsLogin({ phoneNumber: string, url: string }): Chainable<void>
     cognitoLogin(): Chainable<void>
     patchSameSiteCookie(interceptUrl: string): void
```
kristofer-lp8f, line 38: Maybe remove these and make them private functions? Should the even be accessible outside of the custom commands?

apps/system-e2e/src/support/commands.ts
```diff
@@ -60,18 +60,18 @@ Cypress.Commands.add('patchSameSiteCookie', (interceptUrl) => {
...
       )
-      idsLogin(phoneNumber)
+      idsLogin(phoneNumber, url)
     })
   } else {
     cy.session('idsLogin', () => {
-      idsLogin(phoneNumber)
+      idsLogin(phoneNumber, url)
     })
   }
```
kristofer-lp8f, line 76: Could be simplified to have `cy.session` wrapping the `if`-statement

kristofer-lp8f (changes requested):
Looks nice :)

apps/system-e2e/src/support/commands.ts
pshomov: That might be a different refactoring ;). This one was getting rid of the hardcode path, so I think it is better to stick to that purpose.

kristofer-lp8f (approved):

PR #7916, opened 2022-06-28
feat(service-portal): Add institution logos to action cards for applications

egillben:
# Institution logos to action cards for applications

[https://app.clickup.com/t/2ewv26q](https://app.clickup.com/t/2ewv26q)
## What
Add logos to action cards. 
Reused same code from Documents so code had to be moved to Shared.
Refactored how we get the institution names so we get them from Contentful.

The new clock icon is always present on all cards.
The logo is optional.

## Why
By design

## Screenshots / Gifs
image

OLD DESIGN..
image

## Checklist:


libs/shared/utils/src/lib/getOrganizationLogoUrl.ts
```diff
@@ -28,5 +18,3 @@ const getOrganizationLogoUrl = (
   }
   return ''
```
eirikurn, line 19: Do you know why this utility returns "" instead of the fallback image if there are 0 orgs? I think that's a bad fallback since it might cause extra requests to the HTML page.

eirikurn (approved):

libs/shared/utils/src/lib/getOrganizationLogoUrl.ts
egill15: I am not sure why it was implemented this way. But good catch! :) I will change it to the fallback image.

Thorrik58 (approved):

disaerna (approved):

oddsson (approved):

barabrian (approved):
LGTM (reviewed island-ui)

PR #7935, opened 2022-06-29
feat(content-search-indexer): Lock search indexer while sync is in progress

Ingimarsson:
# Lock search indexer while sync is in progress

## What

Every time an entry is auto-saved in Contentful, the search-indexing web hook is called. When multiple people are editing articles, this hook is called frequently. The service starts too many sync operations and it clogs up the pipes. When this happens, content is not synced for up to 20 minutes.

This PR adds a lock. When the sync operations starts, a special document is written to ES containing `running: true`. Every time a sync starts, it checks the lock. This ensures that only one sync process is running at a time.

## Screenshots / Gifs

Below is a screenshot of logs that shows over 30 sync tasks that were all running concurrently.

Screenshot from 2022-06-28 13-48-01

## Checklist:


kksteini (approved):
LGTM

PR #7950, opened 2022-06-30
fix(application-system-payment): Incorporate FJS claim (fella niður kröfu)

johannaagma:
## What

* Add possibility to delete Criminal Record Certificate and Mortgage Certificate
* Fix delete function, so it also deletes entry in Payment table if applies (otherwise you will get a FK error)
* When deleting/pruning application that is connected to a FJS payment, we need to call the FJS API to incorpoare the claim (fella kröfu)

## Why

If the claim is not incorporated, then the user will be "in debt to the state", and will e.g. not be able to fetch their No Debt Certificate (skuldleysisvottorð).

## Checklist:


apps/application-system/api/src/app/modules/application/application.controller.ts
```diff
@@ -1128,6 +1139,48 @@ export class ApplicationController {
...
+    this.deleteApplicationCharge(existingApplication)
+
+    // delete the entry in Payment table to prevent FK error
+    await this.paymentService.delete(existingApplication.id, user)
+
     await this.applicationService.delete(existingApplication.id)
   }
+
+  private async deleteApplicationCharge(application: BaseApplication) {
+    try {
```
obmagnusson, line 1153: You should move this to a separate service within and so you can use it here and for the pruning process to avoid code duplication.

apps/application-system/api/src/app/modules/application/application.controller.ts
```diff
@@ -1128,6 +1139,48 @@ export class ApplicationController {
         'Users role does not have permission to delete this application in this state',
       )
     }
+
+    // delete charge in FJS
+    this.deleteApplicationCharge(existingApplication)
```
obmagnusson, line 1144: this could fail and we would still delete the payment and application on our side.

obmagnusson (changes requested):
Nice! minor remarks on the application-system files

obmagnusson (changes requested):

apps/application-system/api/src/app/modules/application/lifecycle/application-lifecycle.service.ts
```diff
@@ -102,6 +105,14 @@ export class ApplicationLifeCycleService {
     }
   }
 
+  private async pruneApplicationCharge() {
+    for (const prune of this.processingApplications) {
+      await this.applicationChargeService.deleteApplicationCharge(
```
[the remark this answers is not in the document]
johannaagma: @obmagnusson If we continue here, then we will prune the externalData (which contains the chargeId), and then the FJS charge will never be deleted. Should I make this process keep going despite that?

obmagnusson (approved):
Awesome nice work!

apps/application-system/api/src/app/modules/application/charge/application-charge.service.ts
```diff
@@ -0,0 +1,61 @@
...
+        | undefined
+        | null
+      if (!externalData?.createCharge?.data) {
+        return
+      }
+
+      // Delete the charge, using the ID we got from FJS
+      const { id: chargeId } = externalData.createCharge.data as {
+        id: string
+      }
```
saevarma, line 48: This is duplicated in the `application-lifecycle.service`. 
You could consider creating a `getChargeId`  method in the `application-charge.service` that can do this, so it could be reused in the `application-lifecycle.service`.

libs/clients/charge-fjs-v2/src/lib/chargeFjsV2Client.config.ts
```diff
@@ -0,0 +1,30 @@
...
+    return {
+      xRoadServicePath: env.required(
+        'XROAD_CHARGE_FJS_V2_PATH',
+        'IS-DEV/GOV/10021/FJS-Public/chargeFJS_v2',
+      ),
+      fetchTimeout: env.optionalJSON('XROAD_CHARGE_FJS_V2_TIMEOUT') ?? 20000,
+      tokenExchangeScope: env.optionalJSON('XROAD_CHARGE_FJS_V2_SCOPE') ?? [
+        '@fjs.is/finance',
+        // TODO: Remove when fjs has migrated to the scope above.
+        'api_resource.scope',
```
saevarma, line 24: Is this a new API from FJS that is still using the `api_resource.scope`?

apps/application-system/api/src/app/modules/payment/payment.service.ts
```diff
@@ -135,4 +135,28 @@ export class PaymentService {
     }
     return application[0] as ApplicationModel
   }
+
+  async delete(applicationId: string, @CurrentUser() user: User) {
```
saevarma, line 139: You don't use the decorator in a service method, as you are always sending the user in as a parameter. 
The decorator is used in a top level  controller method, to read the user from the context.
```suggestion
  async delete(applicationId: string, user: User) {
```

saevarma (changes requested):

libs/clients/charge-fjs-v2/src/lib/chargeFjsV2Client.config.ts
johannaagma: I talked with Ágúst Leó and he told me to use @fjs.is/finance, but is soon going to go over these scopes.
So I just assumed I should set it up the same as for Finance service for now?

apps/application-system/api/src/app/modules/application/charge/application-charge.service.ts
```diff
@@ -0,0 +1,61 @@
...
+      if (chargeId) {
```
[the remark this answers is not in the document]
johannaagma: If an error occurs calling the function, createCharge will be added to externalData, but with an error message and not the id

albinagu (approved):

apps/application-system/api/src/app/modules/application/lifecycle/application-lifecycle.service.ts
```diff
@@ -7,6 +7,8 @@ import { AwsService } from '@island.is/nest/aws'
 import AmazonS3URI from 'amazon-s3-uri'
 import { LOGGER_PROVIDER } from '@island.is/logging'
 import type { Logger } from '@island.is/logging'
+import { ApplicationChargeService } from '../charge/application-charge.service'
+import { ExternalData } from '@island.is/application/types'
```
saevarma, line 11: GH comment about unused import

apps/application-system/api/src/app/modules/payment/payment.service.ts
```diff
@@ -8,7 +8,7 @@ import type {
   PaymentServiceOptions,
   Item,
 } from '@island.is/clients/payment'
-import type { User } from '@island.is/auth-nest-tools'
+import { CurrentUser, User } from '@island.is/auth-nest-tools'
```
saevarma, line 11: GH comment about unused import

libs/clients/charge-fjs-v2/src/lib/chargeFjsV2Client.config.ts
saevarma: Okay sounds like they have at least started to use the new scope for new APIs. So we should remove the `api_resource.scope`

saevarma (changes requested):

saevarma (approved):

PR #7952, opened 2022-06-30
feat(service-portal): Passport client

thordurhhh:
## What

* Add client for passport service.

## Checklist:


SUMMARY.md
```diff
@@ -169,6 +169,7 @@
   - [National Registry](libs/clients/national-registry/README.md)
     - [V1](libs/clients/national-registry/v1/README.md)
     - [V2](libs/clients/national-registry/v2/README.md)
+  - [Clients Passports](libs/clients/passports/README.md)
```
saevarma, line 172: The title should preferably be overriden using gitbook comments in the library's readme to be:
```suggestion
  - [Passports](libs/clients/passports/README.md)
```

saevarma (changes requested):
Looks good, just minor fix for GitBook.

Snojo (approved):
LGTM

SUMMARY.md
thordurhhh: Makes sense!

Fixed. 💯

libs/clients/passports/src/lib/passports.config.ts
```diff
@@ -0,0 +1,28 @@
...
+  schema,
+  load(env) {
+    return {
+      xRoadServicePath: env.required(
+        'XROAD_PASSPORT_LICENSE_PATH',
+        'IS-DEV/GOV/10001/SKRA-Protected/Forskraning-V1',
+      ),
+      fetch: {
+        timeout: 10000,
+        scope: [NationalRegistryScope.individuals],
```
eirikurn, line 24: Are they using the same scope on this passport service? 🤔

eirikurn (commented):
LGTM, just one question

libs/clients/passports/src/lib/passports.config.ts
thordurhhh: Yes according to the information we got from Þjóðskrá it would be the same scope as for the individual service.
Do you feel like we should require a different scope for the passport service?

apps/api/src/app/app.module.ts
```diff
@@ -267,6 +268,7 @@ const autoSchemaFile = environment.production
         FishingLicenseClientConfig,
         DrivingLicenseBookClientConfig,
         DrivingLicenseApiConfig,
+        PassportsClientConfig,
```
eirikurn, line 271: Should configure the required environment variables in `api/infra` files?

libs/clients/passports/src/lib/passports.config.ts
```diff
@@ -0,0 +1,28 @@
...
+  schema,
+  load(env) {
+    return {
+      xRoadServicePath: env.required(
+        'XROAD_PASSPORT_LICENSE_PATH',
+        'IS-DEV/GOV/10001/SKRA-Protected/Forskraning-V1',
+      ),
+      fetch: {
+        timeout: 10000,
+        scope: [NationalRegistryScope.passport],
```
eirikurn, line 24: What's the status of this change on their side? Will they perform it in all environments soon, or do we need to include the individual scope for compatibility?

libs/clients/passports/src/lib/passports.config.ts
eirikurn: Discussed offline that they should use the @skra.is/passport scope

libs/clients/passports/src/lib/passports.config.ts
thordurhhh: I did not really get a timeline on that, only the confirmation that it could be changed to /passport.

Perhaps it would be best to change it back and have both until I get the confirmation that it will be ok to remove the /individuals scope. So we can be safe in all environments.

I'll do that.

apps/api/src/app/app.module.ts
eirikurn: Looks like this is all configured correctly. My bad! 😆

eirikurn (approved):

PR #7954, opened 2022-06-30
feat(service-portal): new firearm license service

thorkellmani:
## What

Implenet a firearm license that enables fetching of a user's firearm license

## Why

So users can see their firearm license

## Checklist:


apps/api/src/app/app.module.ts
```diff
@@ -250,6 +251,7 @@ const autoSchemaFile = environment.production
       load: [
         AdrAndMachineLicenseClientConfig,
         AssetsClientConfig,
+        FirearmLicenseClientConfig,
```
osk, line 254: Since the client returns more than just the license (firearm ownership at least, also some application endpoints which we probably wont use?), should it have a more generic name?

libs/api/domains/license-service/src/lib/licenseService.module.ts
```diff
@@ -104,11 +126,13 @@ export class LicenseServiceModule {
...
               case GenericLicenseType.MachineLicense:
                 return new GenericMachineLicenseApi(logger, machineApi)
+              case GenericLicenseType.FirearmLicense:
+                return new GenericFirearmLicenseApi(logger, firearmApi)
               default:
                 return null
             }
           },
-          inject: [AdrApi, VinnuvelaApi],
+          inject: [AdrApi, VinnuvelaApi, FirearmApplicationApi],
```
osk, line 135: and here it's called ApplicationApi

thorkellmani: That's the generated Api name

libs/api/domains/license-service/src/lib/client/firearm-license-client/firearmLicenseMapper.ts
```diff
@@ -0,0 +1,72 @@
...
+) => {
+  if (value) {
+    return {
+      type: type,
+      label: 'Nafn',
+      value: value,
+    }
+  }
+  return null
+}
```
saevarma, line 21: This seems to be unused

saevarma (changes requested):
Please rebase your branch with main and review this guide on the upgrade: https://www.notion.so/Nx-13-10-5-a77cd802539e468596da6265700ff3cf

PR #7968, opened 2022-07-01
fix(j-s): disable next step in stepper on hearing arrangements

Arnarkari93:
[Asana - Mál færist ekki úr NEW í DRAFT ef hliðarstika er notuð til að halda áfram eftir Óskir um fyrirtöku skjá hjá sækjanda](https://app.asana.com/0/1199153462262248/1202531295457509/f)
## What

- Disable the next step in stepper on HearingArrangement Page

## Why

- To make sure users click on the continue button that triggers the correct transition of the case from `NEW -> DRAFT`
- Only disable when case state is `NEW`

## Screenshots / Gifs
Screenshot 2022-07-01 at 13 34 28

## Checklist:


oddsson:
Why not transition the case on the next screen instead?

Arnarkari93:
> Why not transition the case on the next screen instead?

Good idea, did not think of that.

oddsson (approved):

PR #7971, opened 2022-07-01
feat(service-portal): add detail finance schedule .. again

disaerna:
# Service portal - Finance schedule detail vol.2

Detail added for finance schedule
Text changes
Props added to expandable row

## Screenshots / Gifs

Screenshot 2022-07-01 at 14 35 26


libs/service-portal/core/src/utils/amountFormat.ts
```diff
@@ -1,4 +1,5 @@
 export const amountFormat = (value: number): string => {
+  if (value === 0) return '0 kr.'
```
thordurhhh, line 2: Is this line unnecessary perhaps?
The previous version would return 0 kr. still if the value is `0` no?

thordurhhh (approved):
nice 🤑

PR #7997, opened 2022-07-06
feat(service-portal): document services sort

disaerna:
# Service Portal - Update filter for mailbox

## What

Refactor to make the backend service filter, search and sort documents instead of doing it in the frontend.
Fetching only pagesizes items (15) at once. 

## Why

To ensure performance when more documents are in the mailbox.

## Checklist:


libs/api/domains/documents/src/lib/document.service.ts
```diff
@@ -33,21 +41,26 @@ export class DocumentService {
     }
   }
 
-  async listDocuments(nationalId: string): Promise<Document[]> {
+  async listDocuments(
```
thordurhhh, line 44: We might want to publish these functions as version2 and keep the others in, while the app still uses them so we don't catch them by surprise when we change the responses. 😃

thordurhhh (changes requested):
This is nice!
Only request for change I see is for a [second version of the document service](https://github.com/island-is/island.is/pull/7997/files#r948904425)

thordurhhh (approved):

oddsson (approved):

PR #8005, opened 2022-07-06
fix(ads): National registry client rewrite

kksteini:
## National Registry Client Rewrite
National Registry Client V2 had been injected into ADS for the purpose of using Forsjá.
This PR rewrites the rest of the logic and shifts around some logic such that the whole of the V1 client is replaced.

We're taking advantage of the fact that an authenticated user creates their own codes and from there we cache the validation instead of making the airlines validate the same thing again later.

There is no issue with a cache miss since the codes have been connected to users from the very beginning of this project.

## Validation rewrite
Discount codes generated for users who are not eligible for Loftbrú.
The following was used to determine `meetsADSRequirements`:
```javascript
if (user.fund) {
  return user.fund.credit === user.fund.total - user.fund.used
}
return false
```
Moved this to a filter for creation of discount codes in its resolver instead.
This allows for removing validation checks in the public facing API for the airline companies 

## Checklist:


eirikurn (approved):
Reviewed core files, but this PR sounds awesome 👍

PR #8010, opened 2022-07-07
fix(j-s): Missing DOB Regression

gudjong:
# Missing DOB Regression

[Regression - tóm kennitala representuð sem bandstrik í infopanel](https://app.asana.com/0/1199153462262248/1202505428021043/f)

## What

- No longer displays '-' in info card when date of birth / national id is missing.
- No longer disploys '-' for missing date of birth / national id in autofilled police demands in investigation cases.

## Why

This was a regression.

## Screenshots / Gifs

image

image

## Checklist:

### DIFF
--- a/apps/judicial-system/backend/src/app/modules/case/case.controller.ts
+++ b/apps/judicial-system/backend/src/app/modules/case/case.controller.ts
@@ -25,12 +25,12 @@ import {
   DokobitError,
   SigningServiceResponse,
 } from '@island.is/dokobit-signing'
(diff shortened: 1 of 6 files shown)


oddsson:
Consider moving this logic into the `formatDOB` function. We already do a 

```  
if (!nationalId) {
    return '-'
}
```

You could do 

```  
if (!nationalId) {
    return ''
}
```

instead and skip the null check where we call the function

oddsson (approved):

PR #8035, opened 2022-07-11
feat(web): Organization pages, external script support

RunarVestmann:
# Organization pages, external script support

## What

* Added a way for organization pages to load an external script

## Why

* This was a requested feature

## Checklist:

### DIFF
--- a/apps/web/screens/Organization/Home.tsx
+++ b/apps/web/screens/Organization/Home.tsx
@@ -53,16 +53,16 @@ const Home: Screen<HomeProps> = ({ organizationPage, namespace }) => {
   )
 
   useEffect(() => {
-    if (!organizationPage.organization?.statusMonitoringScriptUrl) return
+    if (!organizationPage.organization?.externalScriptUrl) return
     const element = document.createElement('script')
     element.type = 'text/javascript'
-    element.src = organizationPage.organization.statusMonitoringScriptUrl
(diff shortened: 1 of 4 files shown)


Ingimarsson (approved):

kksteini:
I am worried about this.
Who requested this and for what purposes?
How will this be moderated?

I think we should take issue with allowing any script to be loaded on the basis of tracking.
As far as I know we do not track personal information across our website. We are using plausible for anonymous heatmap and query tracking purposes. There is no "Allow Cookies" banner on island.is for example.

kksteini:
Also, will this not clash with the viewer-response ruleset, see here: https://github.com/island-is/infrastructure/blob/main/modules/islandis-cloudfront/viewer-response.js?

RunarVestmann:
This was requested by Sjúkratryggingar originally. They want to embed a script that allows the web to automatically show if there is an error for one of their web services. So basically the script shows a pop up in the bottom right corner if it is appropriate. And regarding the viewer-response ruleset, I believe that Snorri from Andes changed the ruleset for this as he was also in the e-mail communication regarding this.

RunarVestmann:
Alright, let's put this on hold for now, I might be going for another solution to solve this issue: https://islandis.slack.com/archives/C02P06N03T3/p1658395848757129

PR #8046, opened 2022-07-14
feat(application-system): Only allow only one application of each type as default behaviour

Thorrik58:
# Only allow creation of one application of each type in draft as default behaviour

## What

Only allow the user to create one application of each type in draft as the default behaviour.

## Why

Default behaviour of an application should be to only allow the user to create one for each type.
Adding a flag that lets the team decide whether there should be allowed more than one.

## Screenshots / Gifs

When the application has allowMultipleApplicationsInDraft as undefined or as true the button appears.
image

When the user has created 2 applications in the draft state with the flag set as false the button disappears
image

## Checklist:


apps/application-system/form/src/routes/Applications.tsx
```diff
@@ -41,9 +49,18 @@ export const Applications: FC = () => {
...
+        ApplicationContext,
+        ApplicationStateSchema<EventObject>,
+        EventObject
+      >
+    | undefined
+  >(undefined)
   const checkDelegation = useCallback(() => {
     setDelegationsChecked((d) => !d)
   }, [])
+  console.log(template)
```
Thorrik58, line 63:
```suggestion
```

obmagnusson (approved):
LGTM

Thorrik58:
Going into next release as we need to make sure all applications that allow multiple applications in draft are defined as such.

PR #8051, opened 2022-07-15
refactor(service-portal): use IntroHeader on all screens

disaerna:
# Service Portal - Use IntroHeader 

## What

Refactor a lot of screens to use IntroHeader instead of each component using h1 and text.

... plus some accidental import sort 😬 

## Why

All screens should have the same spacing and heading size. Minimises future work if anything changes.

## Checklist:


Thorrik58 (approved):
LGTM

thorkellmani (approved):

kksteini (approved):
Reviewed CodeOwner files only, LGTM.

Did notice a warning for `Array.prototype.map()` expecting a return value in `ApplicationList.tsx` and since nothing is being return, or stored; may I suggest `.forEach` instead since that file is being edited anyway :)

PR #8056, opened 2022-07-21
fix(web): Fix category page when article belongs to multiple groups

Ingimarsson:
# Fix category page when article belongs to multiple groups

## What

Articles can be shown in more than one group by adding to the `otherCategory` and `otherGroup` fields. Currently, this only works when an article's main category is different from the other category.

This PR fixes it so that articles can appear in the main group and also other groups on the same category page.

## Checklist:


rafnarnason (approved):
lgtm

PR #8064, opened 2022-07-25
feat(service-portal): Add application status subpages overview

Thorrik58:
# Add application incomplete overview

## What

Add application incomplete/in progress/finished overview and refactor of overview (decided to postpone as we need to standardise application states)
Refactor of code to simplify and avoid rerenders.
Focus added when a user arrives at the overview after finishing an application.

## Why

Allow the user to access unfinished directly through link.

## Screenshots / Gifs

image

## Checklist:


libs/service-portal/applications/src/components/ApplicationGroup.tsx
```diff
@@ -0,0 +1,69 @@
...
+                onClick={() => {
+                  handlePageChange(page)
+                }}
+              >
+                {children}
+              </button>
+            )}
+          />
+        </Box>
+      ) : null}
```
egill15, line 66: You can clean it up by using Inline If with logical && operator so you don't have to use ":null"

egill15 (approved):

obmagnusson (approved):
LGTM

thordurhhh (approved):

PR #8068, opened 2022-07-26
feat(web): Add Mailchimp module for email signup forms

Ingimarsson:
# Add Mailchimp module for email signup forms

https://islandis.slack.com/archives/C01C7C3M2J3/p1657014545357239

## What

We currently have the Mailing List Signup Slice for allowing users to subscribe to the Stafrænt Ísland mailing list and also to register for the Tengjum Ríkið conference. They currently make a direct GET call to Mailchimp, which does not work in Firefox when strict mode is enabled.

This is a more permanent solution for Mailchimp subscriptions. We create an API module that talks to Mailchimp, and in the frontend we just call the mutation.

Slice can be seen on:

https://island.is/s/stafraent-island/frettabref
https://island.is/s/stafraent-island/tengjum-rikid-2022

## Why

Mailchimp calls sometimes get blocked in user's browsers.

## Checklist:


apps/web/components/NameSignupForm/NameSignupForm.tsx
```diff
@@ -66,24 +77,18 @@ export const NameSignupForm = ({ namespace, slice }: NameSignupFormProps) => {
...
-              type: 'error',
-              text: n(
-                'formEmailAlreadyRegistered',
-                'Þetta netfang er þegar á skrá.',
-              ),
-            })
-          } else if (data) {
+      })
+        .then((result) => {
+          if (result.data.mailchimpSubscribe.subscribed) {
```
RunarVestmann, line 91: Perhaps have an else clause here like you do in the MailingListSignup component submit handle function

libs/cms/src/lib/cms.contentful.service.ts
```diff
@@ -744,4 +749,23 @@ export class CmsContentfulService {
       null
     )
   }
+
+  async getMailingListSignupSlice({
+    id,
```
RunarVestmann, line 754: Perhaps add a lang parameter in case we'd like to reuse this for english things

RunarVestmann (approved):
Looks good to me, I left 2 comments which are not show stoppers but would be nice to have

libs/api/domains/mailchimp/src/lib/dto/mailchimpSubscribe.input.ts
```diff
@@ -0,0 +1,23 @@
...
+
+@InputType()
+export class MailchimpSubscribeInput {
+  @Field()
+  @IsString()
+  signupID!: string
+
+  @Field()
+  @IsString()
+  email!: string
```
rafnarnason, line 12: @IsEmail() to validate email addresses :)

apps/web/components/NameSignupForm/NameSignupForm.tsx
kristofer-lp8f, line 91: No `?` chaining? Will this chaining never fail?

rafnarnason (approved):
lgtm with one comment

libs/api/domains/mailchimp/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+Module for Mailchimp subscription endpoints.
+
+## Running unit tests
+
+Run `nx test api-domains-mailchimp` to execute the unit tests via [Jest](https://jestjs.io).
+
+## Running lint
+
+Run `nx lint api-domains-mailchimp` to execute the lint via [ESLint](https://eslint.org/).
```
kristofer-lp8f, line 13:
```suggestion
Run `yarn nx test api-domains-mailchimp` to execute the unit tests via [Jest](https://jestjs.io).

## Running lint

Run `yarn nx lint api-domains-mailchimp` to execute the lint via [ESLint](https://eslint.org/).
```

apps/web/components/MailingListSignup/MailingListSignup.tsx
```diff
@@ -80,9 +65,19 @@ export const MailingListSignup: React.FC<MailingListSignupProps> = ({
               type: 'success',
               message: successMessage,
             })
+          } else {
```
kristofer-lp8f, line 68: Is there no case for "Already signed up"?

kristofer-lp8f (changes requested):
I don't see any changes to test cases. Can you modify/add test cases?

apps/web/components/MailingListSignup/MailingListSignup.tsx
Ingimarsson: Mailchimp is not configured to return "already signed up", it always just returns "success". So I decided to omit this.

apps/web/components/NameSignupForm/NameSignupForm.tsx
Ingimarsson: Could fail, added optional chaining :)

libs/api/domains/mailchimp/src/lib/dto/mailchimpSubscribe.input.ts
Ingimarsson: Didn't know about that one, thanks :)

Ingimarsson:
> I don't see any changes to test cases. Can you modify/add test cases?

Added a test case and also fixed what you commented on :)

kristofer-lp8f (approved):
Well done 🎉

libs/api/domains/mailchimp/src/lib/dto/mailchimpSubscribe.input.ts
```diff
@@ -0,0 +1,23 @@
...
+
+  @Field({ nullable: true })
+  @IsOptional()
+  @IsString()
+  name?: string
+
+  @Field({ nullable: true })
+  @IsOptional()
+  @IsBoolean()
```
darrikonn, line 21: Don't need IsBoolean + IsOptional + IsString for GraphQL inputs. Creates unnecessary noise for the reader.

Only valid validation is `IsEmail`. Other basic validation will be picked up by the GraphQL Schema.

apps/web/components/NameSignupForm/NameSignupForm.tsx
```diff
@@ -56,6 +59,14 @@ export const NameSignupForm = ({ namespace, slice }: NameSignupFormProps) => {
     return errors
   }
 
+  const [
+    submit,
```
darrikonn, line 63: This comment should be treated as a feedback (not requiring code-changes, but it would improve it)

Same goes for the other file, a bit nit-picky; but improves code readability.
It would improve the code if this would be explicitly named, i.e. if this would be e.g. `subscribeToMailchimp`.
`submit` is too broad, since this is explicitly calling a mutation with side-effects. This component could e.g. have more forms/mutations.

Ingimarsson:
@darrikonn Thank you, pushed changes :)

Also added us as code owners for this module.

darrikonn (approved):

PR #8072, opened 2022-07-26
feat(devx): Allow multiple projects and --reset anywhere

kristofer-lp8f:
## What

Improve developer experience
- Allow fetching multiple project secrets in a single invocation
- Allow `--reset` option anywhere, not just at the end
- Add `-r` shortform for `--reset`
- Feedback when no secrets found
- Feedback on total number of secrets found

## Why

When interacting with the `get-secrets` command developers need to manually verify that they found _any_ secrets. There was no feedback at all; whether typo, incorrect usage, trying to use multiple projects.

## Screenshots / Gifs

Example invocation
Screenshot 2022-07-26 at 17 07 08

## Checklist:


busla (commented):
Personally I would like to discontinue the persistent .env.secrets file with creds in plain text, and instead export them pre-runtime to the corresponding environments.

busla:
something like this (not tested)

```bash
function secret() {
  local param
  param="$1"

  # We split the ssm key on / and keep the last item as ENV
  aws ssm get-parameter \
    --name "$param" \
    --with-decryption \
    --query 'Parameter.{Name:Name,Value:Value}' | jq -r '. | "\(.Name|split("/")|.[-1])=\(.Value)"'
}

function set_secrets() {
  while read -r line ; do
    info "getting ${line?} from ssm and setting env"
    export "$(secret "${line?}")"
  done < "${SCRIPT_DIR}/../.env.ssm"
}

```

This approach assumes that all projects would need a `.env.ssm` file with required ssm keys for that project only. The DSL could generate those files.
 
```bash
/project/a/ENV_FOR_A
/project/a/ANOTHER_ENV_FOR_A
```

kristofer-lp8f:
I agree that storing secrets in plaintext is bad.

Until we get proper dependency startup and something similar to what you propose, can we add this, at least until we deprecate the `get-secrets` script?

busla (approved):
LGTM

pshomov (approved):
LGTM

PR #8077, opened 2022-07-29
feat(web): Add a way to embed stepper in article body

RunarVestmann:
# Add a way to embed stepper in article body

## What

* Instead of having the stepper process entry always at the top you can now place it where you want (in between text for example)
* Made the step option value for the select component actually reference the same memory address as exists in the list of options, meaning that when you now open the select box you'll actually see the selected value highlighted instead of just the top one in the list
* 

## Why

* Before this change a stepper process entry would only be at the top of the page, we want to be able to choose where it appears

### Before
image

(Highlight was off due to the way the Select component compares values)
image

### After (now we can actually choose where the process entry is displayed)
image

(Highlight is correct now)
image

## Checklist:


albinagu (approved):
lgtm

PR #8083, opened 2022-07-29
feat(web): Allow Process entry asset reference instead of url

RunarVestmann:
# Allow Process entry asset reference instead of url

## What

* Now users can select an asset in Contentful instead of having to use a url
* Removed the type field from Process Entry since it's no longer used
* Only render a process entry if there is a process link provided, since the field is no longer required in Contentful and we don't want to show a process entry if there is no link

(Now there's a new field in Contentful for an asset)
image

## Checklist:


albinagu (approved):
lgtm

PR #8122, opened 2022-08-09
fix(application-system): Reverted tell us a story and added reply to application emails

Thorrik58:
# Reverted tell us a story and added reply to application emails

## What

Reverted tell us a story and added reply to application emails

## Why

Wrong approach to a fix for application system and should not be needed for the form.
The email service doesn't allow us to assign any email as sender, which makes perfect sense when you think about it. 😅 

## Screenshots / Gifs

## Checklist:


egill15 (approved):

rafnarnason (approved):
lgtm

PR #8129, opened 2022-08-10
fix(j-s): make case type required to continue for investigation cases

Arnarkari93:
[Asana - Það er hægt að stofna nýja rannsóknarkröfu án þess að setja tegund kröfu í viðmóti](https://app.asana.com/0/1199153462262248/1202735884071838/f)
## What

- Make sure users selects a option for case type in the form before allowing them to create the case

## Why

- The field is required

## Checklist:


apps/judicial-system/web-e2e/src/integration/Prosecutor/InvestigationCases/defendant.spec.ts
```diff
@@ -8,14 +8,15 @@ describe(NEW_IC_ROUTE, () => {
     cy.visit(NEW_IC_ROUTE)
   })
 
-  it('should require valid data', () => {
+  it.only('should require valid data', () => {
```
oddsson, line 11:
```suggestion
  it('should require valid data', () => {
```

oddsson (approved):

PR #8131, opened 2022-08-10
fix(island-ui): Update dialogprompt to match other island.is designs 

egill15:
# Dialog modal update

## What
Change styling and added option to add image to dialog card

## Why
Consistency in theme for island.is, for example
image

## Screenshots / Gifs
With logo
image

Without logo
image

With one button
image

Without buttons
image

Mobile
image

## Checklist:


libs/island-ui/core/src/lib/DialogPrompt/DialogPrompt.tsx
```diff
@@ -89,60 +89,70 @@ export const DialogPrompt = ({
...
+            </Box>
+            <GridRow align="flexStart" alignItems="flexStart">
+              <GridColumn span={img ? ['8/8', '5/8'] : '1/1'}>
+                <Text variant="h2" as="h2" marginBottom={1}>
+                  {title}
                 </Text>
+                {description && <Text paddingTop={2}>{description}</Text>}
+              </GridColumn>
+
+              {img && (
```
Arnarkari93, line 115: I think we should instead allow passing in the image component instead. That way this component is more flexible.

libs/island-ui/core/src/lib/DialogPrompt/DialogPrompt.tsx
```diff
@@ -89,60 +89,70 @@ export const DialogPrompt = ({
...
-                <Icon icon="close" color="blue400" size="medium" />
-              </FocusableBox>
-              <Text variant="h2" as="h3" paddingBottom={2}>
-                {title}
-              </Text>
-              {description && (
-                <Text variant="intro" paddingBottom={2}>
-                  {description}
+            <Box className={styles.closeButton}>
+              <Button
```
Arnarkari93, line 97: For better accessibility, we should add an aria label since this button has no text.

Arnarkari93 (changes requested):

Arnarkari93 (approved):

PR #8141, opened 2022-08-11
refactor(auth-react): Refactor Check IDP Session

saevarma:
# ...

https://www.notion.so/SessionInfo-bug-on-computer-sleep-3628b88e47aa46258114d8a78a14b2ec

## What

Refactor Check IDP Session to handle computer sleep and temporary loss of network connection.

## Why

For more reliably user session timeout monitoring.

## Screenshots / Gifs

## Checklist:

### DIFF
--- a/libs/auth/react/src/lib/Authenticator/CheckIdpSession.tsx
+++ b/libs/auth/react/src/lib/Authenticator/CheckIdpSession.tsx
@@ -82,7 +82,7 @@ export const CheckIdpSession = () => {
 
       if (!isActive() && retryCount > MAX_RETRIES && hasBeenActive) {
         window.location.reload()
-      } else if (!hasMessage) {
+      } else if (!hasMessage && retryCount < MAX_RETRIES) {
         userSession.current.retryCount += 1
         reloadIframe()
       }
--- a/libs/auth/react/src/lib/Authenticator/CheckIdpSession.tsx
+++ b/libs/auth/react/src/lib/Authenticator/CheckIdpSession.tsx
@@ -1,21 +1,13 @@
-import React, {
-  SyntheticEvent,
-  useCallback,
-  useEffect,
-  useRef,
-  useState,
-} from 'react'
+import addSeconds from 'date-fns/addSeconds'
+import { useCallback, useEffect, useReducer, useRef } from 'react'
 
 import { getAuthSettings, getUserManager } from '../userManager'
 
-enum SessionInfoMessageType {
-  SessionInfoRequest = 'SessionInfoRequest',
-  SessionInfoResponse = 'SessionInfoResponse',
-}
+const SessionInfoMessageType = 'SessionInfo'
 
-export interface SessionInfoResponse {
+interface SessionInfo {
   // Type to use to filter postMessage messages
-  type: SessionInfoMessageType.SessionInfoResponse
+  type: typeof SessionInfoMessageType
 
   // Message detailing if the request was processed OK, no session detected or with failure.
   message: string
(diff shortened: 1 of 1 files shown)


libs/auth/react/src/lib/Authenticator/CheckIdpSession.tsx
```diff
@@ -30,58 +22,108 @@ export interface SessionInfoResponse {
...
-    }
-  }
+  const checkActiveSession = useCallback(() => {
+    setTimeout(() => {
+      const hasBeenActive = !!userSession.current.sessionEnd
+      const { retryCount, hasMessage } = userSession.current
+
+      if (!isActive() && retryCount > MAX_RETRIES && hasBeenActive) {
+        window.location.reload()
+      } else if (!hasMessage && retryCount < MAX_RETRIES) {
```
eirikurn, line 85: Would be nice to reuse the `isActive()` and get rid of `hasMessage`, but I understand if you were worried about a race condition with `sessionInfo.isExpired`.

If that's a concern you should also toggle `hasMessage` if there is "No user session".

eirikurn: When can a session info message be active but not have an `expiresIn` value?

libs/auth/react/src/lib/Authenticator/CheckIdpSession.tsx
```diff
@@ -30,58 +22,108 @@ export interface SessionInfoResponse {
...
+
+          userSession.current.intervalHandle = setInterval(() => {
+            const now = new Date()
+
+            if (
+              userSession.current.sessionEnd &&
+              now > userSession.current.sessionEnd
+            ) {
+              resetUserSession()
+              reloadIframe()
```
eirikurn, line 124: Maybe explain what this means. Why we reload the iframe instead of signing in.

eirikurn, line 124: Maybe explain what this condition means. Why we reload the iframe instead of signing in.

libs/auth/react/src/lib/Authenticator/CheckIdpSession.tsx
```diff
@@ -30,58 +22,108 @@ export interface SessionInfoResponse {
...
+  /* The handle of the setInterval function. If set this indicates the user session active state. */
+  intervalHandle: ReturnType<typeof setInterval> | null
+
+  /* Number of times we have tried to load the iframe to receive a new session info message. */
+  retryCount: number
 }
 
-const messageEventName = 'message'
+const MAX_RETRIES = 2
+const MESSAGE_EVENT_NAME = 'message'
```
eirikurn, line 40: IMO this is unneccessary obfuscation when dealing with a built in browser event.

eirikurn (approved):
I've gone through this with a fine-toothed comb and only thing I got was readability comments and a head ache 😅

libs/auth/react/src/lib/Authenticator/CheckIdpSession.tsx
```diff
@@ -22,38 +22,45 @@ interface SessionInfo {
...
 }
 
+/**
+ * This component monitors if the user session is active on the Identity Provider (IDP).
+ * When it detects that the user session is expired it redirects to the sign in page on the IDP.
+ *
+ * It loads a script from the IDP's 'connect/sessioninfo' endpoint into an iframe.
+ * The script uses the postMessage API to post SessionInfoMessage, which contains
+ * details if the session is expired or after how many seconds it will expire.
+ * We use these details to register a interval to monitor the session expiration.
```
eirikurn, line 56: 💯 !

libs/auth/react/src/lib/Authenticator/CheckIdpSession.tsx
```diff
@@ -30,88 +22,148 @@ export interface SessionInfoResponse {
...
-    }
-  }
+  const checkActiveSession = useCallback(() => {
+    setTimeout(() => {
+      const hasBeenActive = !!userSession.current.sessionEnd
+      const { retryCount } = userSession.current
+
+      if (!isActive() && retryCount > MAX_RETRIES && hasBeenActive) {
+        // We are unable to retrieve a message from the IDP after max retries,
+        // so we reload the window to check if the user session is still valid.
```
eirikurn, line 91:
```suggestion
        // We were unable to retrieve a message from the IDP after max retries and have a reason
        // to believe that the session is expired (an earlier UserSessionMessage has expired).
        // So we reload the window just to be safe. This causes one of three things to happen:
        // - If the iframe is broken and the user does have a valid IDP session, they'll generally reload where they were.
        // - If the iframe is broken and the user does not have a valid IDP session, they're sent to the login page.
        // - If the user has a network problem, then they'll see a browser error screen, but at least any sensitive information is not visible any more.
```

eirikurn (commented):
💯

PR #8158, opened 2022-08-16
fix(j-s): dont render invalid dates in case list

Arnarkari93:
[Asana - Mobile útgáfa af málalista](https://app.asana.com/0/1199153462262248/1202329595345230/f)
## What

- Fix invalid dates being rendered in the mobile case list view

## Why

- We want to skip rendering if dates are not relevant

## Screenshots / Gifs
Screenshot 2022-08-16 at 14 25 38

## Checklist:


apps/judicial-system/web/src/routes/Shared/Cases/MobileCase.tsx
```diff
@@ -66,7 +66,7 @@ const MobileCase: React.FC<Props> = ({
       heading={capitalize(caseTypes[theCase.type])}
       onClick={onClick}
       tags={[
-        <Tag variant={tag.color} outlined disabled>
+        <Tag variant={tag.color} outlined disabled key={tag.text}>
```
oddsson, line 69: I think there are going to be non-unique keys if we do this. Consider using something else, like a uuid or a timestamp.

oddsson (changes requested):
One thing regarding unique keys

apps/judicial-system/web/src/routes/Shared/Cases/MobileCase.tsx
Arnarkari93: It is not a problem here since we explicitly say only a single tag is rendered.

Arnarkari93: > The key only has to be unique among its siblings, not globally unique.
https://reactjs.org/docs/reconciliation.html#keys

oddsson (approved):

PR #8162, opened 2022-08-17
feat(auth-public-api): Add scopes value to audit log on create and update on delegation.

saevarma:
https://www.notion.so/B-ta-vi-gildum-scope-um-Audit-Log-c22a7bdd8ed945ea96f7969e05ddb3d1

## What

Update audit on create and update on delegation to include scope names, validFrom and validTo values.

## Why

For audit purpose when multiple users can update the same delegation, i.e. for a company.

## Screenshots / Gifs

## Checklist:


apps/services/auth-public-api/src/app/modules/delegations/meDelegations.controller.ts
```diff
@@ -167,7 +174,7 @@ export class MeDelegationsController {
         namespace,
         action: 'update',
         resources: (delegation) => delegation?.id ?? '',
-        meta: { fields: Object.keys(delegation) },
+        meta: { scopes: delegation.scopes },
```
eirikurn, line 177: Can the update audit include which scopes were deleted in addition to which were saved?

saevarma: Excellent idea. I will add it as two sperate keys: `added` and `deleted`.

saevarma: After rethinking this, it would introduce a new DB query to get the delegation as is before the update. The input or the result of the update only contains how the scopes should be now.

I'm hesitant to do this, if we simply audit the values from the DTO, you would be able to query the audit log of the last two updates and reviews those list to see new and deleted scopes.

eirikurn (approved):
LGTM 🙌

PR #8170, opened 2022-08-18
feat(web): Form - use field from content type instead of from UI configuration

RunarVestmann:
# Form: use field from content type instead of from UI configuration

## What

* Created two fields on the form content type instead of having those fields globally set via a UI config
* The dropdown showed the wrong selected item, so I fixed that by having the value be the same reference as the options list item

## Why

* Some forms won't be using the determined text so we want to be able to change it per form instead of having to change every single form

## Checklist:


albinagu (approved):
lgtm

PR #8173, opened 2022-08-18
feat(service-portal): Add feature flag option for single pages

thordurhhh:
## What

* Add feature flag option for single pages

## Why

There are some pages that are a part of a module that have, and will be added behind a feature flag. We have had to make some adjustments like creating a new module for that case, since we have the option today to flag a module, but not a single page.

## Checklist:


disaerna (approved):
Love it 🥇

PR #8175, opened 2022-08-18
fix(auth-react): Disable user session monitoring on localhost

saevarma:
## What

We make sure that when we receive 'No Session' message from IDS we check that the user has been active before reloading the window.

## Why

Due to 3rd party cookie blocking, it causes redirect loops on locahost when using IDS in dev, because the cookie is blocked and IDS sends 'No Session' message.

## Screenshots / Gifs

## Checklist:


eirikurn (commented):
I thought CheckIdpSession was never supposed to reload the page unless the iframe worked at least once, and then only up to an hour later?

saevarma:
> I thought CheckIdpSession was never supposed to reload the page unless the iframe worked at least once, and then only up to an hour later?

This is only issue on localhost. The iframe is working correctly. Due to the 3rd party cookie blocking, the IDS sees the request as unauthenticated so the iframe sends a message about no user session, which should cause a redirect to Sign In.

We could use `hasBeenActive` in combination with the `No Session` to not trigger redirect to Sign In.

eirikurn (approved):
LGTM!

PR #8179, opened 2022-08-19
fix(p-sign-application): Legal Guardian contact info

alexdiljar:
Broke down form sections. Added applicant with actor role, if there is a legal guardian actor on the application we need to get their contact info. This was the simplest way to change the string value on two fields since it is not possible to get the whole application in condition or use dynamic textfields. 

## Checklist:


albinagu (approved):
i like it

PR #8184, opened 2022-08-20
feat(web): Contentful role permissions app

RunarVestmann:
# Contentful role permissions app

## What

* A Next.js app that helps us manage contentful role permissions
* This app will not be deployed but rather ran locally (think of this as a dev tool)
* I updated the contentful-management library to a more later version so I could access the plain client API (it hopefully won't break anything but if someone could double check I'd appreciate it).
* I also changed the DropdownMenu component a bit by having a boolean prop which enabled a scroll view

## Why

* The Contentful default UI for managing roles is very difficult to work with and since we intend to start managing what Contentful users can see and do we went for the approach of utilising Contentful roles but doing so with our own UI so we can customize how we manage those roles.
* We want this code to be version controlled here in the monorepo because perhaps in the future more teams would like to use this app and the fact that this app is highly related to Contentful

image

https://user-images.githubusercontent.com/43557895/192320529-7860ddb5-63dd-433d-b4dd-1c34b85bc6ab.mov

## Checklist:


package.json
```diff
@@ -81,6 +81,7 @@
     "@nestjs/sequelize": "8.0.0",
     "@nestjs/swagger": "4.8.2",
     "@nestjs/terminus": "7.0.1",
+    "@nrwl/next": "13.10.5",
```
eirikurn, line 84: Hm. Did the NX generator add this as `dependencies` rather than a `devDependencies`?

It should at least only be listed once.

package.json
```diff
@@ -139,7 +140,7 @@
     "configcat-js": "5.4.0",
     "configcat-node": "6.7.0",
     "contentful": "7.14.6",
-    "contentful-management": "6.0.5",
+    "contentful-management": "10.12.0",
```
eirikurn, line 143: Need to make sure the translation script still works. Try this:

```
yarn nx extract-strings shared-translations
```

apps/tools/contentful-role-permissions/pages/api/update-role-permissions.ts
```diff
@@ -0,0 +1,110 @@
...
+  const data = JSON.parse(req.body) as {
+    checkboxState: ReturnType<
+      typeof extractInitialCheckboxStateFromRolesAndContentTypes
+    >
+    readonlyCheckboxState: ReturnType<
+      typeof extractInititalReadonlyCheckboxStateFromRolesAndContentTypes
+    >
+    roleNamesThatCanReadAllAssets: string[]
+    tags: TagProps[]
+  }
```
eirikurn, line 30: Would consider refactoring this into explicit types and interfaces.

apps/tools/contentful-role-permissions/.eslintrc.json
```diff
@@ -0,0 +1,31 @@
+{
```
eirikurn, line 1: Should add this folder to CODEOWNERS.

eirikurn (commented):
LGTM! A few minor comments 🙌

package.json
RunarVestmann: Yep, the NX generator added this

apps/tools/contentful-role-permissions/.eslintrc.json
RunarVestmann: Done, thanks for reminding me :)

package.json
RunarVestmann: Seems to work 
image

package.json
RunarVestmann: I removed it from the dependencies object

apps/tools/contentful-role-permissions/pages/api/update-role-permissions.ts
RunarVestmann: Done :)

eirikurn (approved):
Lgtm

libs/island-ui/core/src/lib/DropdownMenu/DropdownMenu.css.ts
```diff
@@ -9,6 +9,11 @@ export const menu = style({
   },
 })
 
+export const scroll = style({
```
oddsson, line 12: I think this could be a bit error prone. Let's say you didn't know how many `items` you were rendering in this menu for whatever reason (maybe you're fetching some data from an api or you are conditionally rendering items where some users have a lot of items and others have a few). If you set the `hasScroll` prop on a menu that has a few `items`, the menu would be way to big, i.e.

Screenshot 2022-09-06 at 08 43 27

Also, this is not very mobile friendly.

I think a better solution would be to add the ability to add class names instead of hasScroll, like we are doing in the `Box` component. That way, the person using the component would have control over how it acted under certain conditions.

oddsson (changes requested):
One comment regarding the DropdownMenu change.

oddsson (approved):

siggifv (approved):

PR #8192, opened 2022-08-22
chore(core): Fix hot reloading of node servers in development.

eirikurn:
## What

Make sure our local dev servers watch files properly and restart when they change.

## Why

For DX.

We're using esbuild for faster and smaller production builds. Before this summer's NX update, NX would ignore the configuration and use the internal node webpack builder during development. After the update, it tries to use the esbuild executor during development, but it turns out its watch implementation is quite broken.

It only watches app folder and restarts a couple times. Possible fixes would have issues of their own; watch the entire monorepo, cost more CPU resources and/or include large delays.

So we decided to create a custom executor which dynamically invokes the webpack executor when `watch: true`, otherwise it runs the esbuild executor.

## Checklist:


saevarma (approved):
🥳

HelgiHH (approved):

obmagnusson (approved):
Nice!

disaerna (approved):

tools/executors/node/impl.js
```diff
@@ -0,0 +1,53 @@
+"use strict";
```
Arnarkari93, line 1: Why do we need to check in the generated `.js` file and `js.map`?

Arnarkari93 (changes requested):
One question

tools/executors/node/impl.js
eirikurn: Good question. It's because this is loaded as an NX executor, which doesn't support TypeScript at runtime.

Our choices were to:

1. Build and publish this as a stand-alone NPM package which we install in the monorepo.
2. Build the JS on every developer machine and CI.
3. Commit the JS.

We decided on #3 as the simplest option for this "one-off" thing, with the possibility to improve this later if we keep changing it.

Arnarkari93: Ahh, I see. Understandable.

Arnarkari93 (approved):

Snojo (approved):
DevOps looks good

PR #8193, opened 2022-08-22
feat(j-s): Mask filenames

oddsson:
# Mask filenames

[Asana](https://app.asana.com/0/1199153462262248/1202798217733967/f)

## What

Mask filenames so that everything between the first and last letter of the filename is replaced with a *. 

Example
foobar.gif becomes f****r.gif. 
foobar.baz.gif becomes f********z.gif

## Why

Filenames sometimes contain personal info.

## Checklist:


apps/judicial-system/backend/src/app/modules/court/court.service.ts
```diff
@@ -242,7 +253,7 @@ export class CourtService {
             courtId,
             courtCaseNumber,
             subject,
```
gudjong, line 255: Consider also masking subject

gudjong (changes requested):

gudjong (approved):

PR #8223, opened 2022-08-24
feat(island-ui): add image to action card + refactor

disaerna:
# Action card refactor

## What

Refactor images on action card. 

Image has 3 types
1. Avatar, displays heading initials 
2. Logo, displays a small image leftside of Heading and above Text 
3. Image, displays a larger image leftside of both Heading and Text

## Why

Instead of having these types as three seperate props, we merge them together in one. We never want to display more then one of these types in the same card.

## Screenshots / Gifs
Screenshot from Storybook 📖 

Screenshot 2022-08-24 at 10 19 42

## Checklist:

### DIFF
--- a/libs/application/ui-components/src/components/ApplicationList/ApplicationList.tsx
+++ b/libs/application/ui-components/src/components/ApplicationList/ApplicationList.tsx
@@ -148,7 +148,7 @@ const ApplicationList = ({
 
             return (
               <ActionCard
-                logo={getLogo(application.typeId)}
+                image={{ type: 'logo', url: getLogo(application.typeId) }}
                 key={`${application.id}-${index}`}
                 date={format(new Date(application.modified), formattedDate)}
                 tag={{
--- a/libs/application/ui-shell/src/components/DelegationsScreen.tsx
+++ b/libs/application/ui-shell/src/components/DelegationsScreen.tsx
@@ -204,7 +204,7 @@ export const DelegationsScreen = ({
               {screenData.authDelegations?.map((delegation: Delegation) => (
                 <ActionCard
                   key={delegation.from.nationalId}
-                  avatar
+                  image={{ type: 'avatar' }}
                   heading={delegation.from.name}
                   text={
                     formatMessage(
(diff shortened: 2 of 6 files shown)


Arnarkari93 (commented):
This Action Card component is getting really bloated. In theory, it is a Box with a border, some padding, optional background, and then content. There should not be any limits to the contents of the card. 

If some application has a usecase where the content is fixed it is cleaner to define a local component that fits those specific needs.

I have no authority to enforce the refactoring effort needed, I just wanted to raise my concern.

disaerna:
> This Action Card component is getting really bloated. In theory, it is a Box with a border, some padding, optional background, and then content. There should not be any limits to the contents of the card.
> 
> If some application has a usecase where the content is fixed it is cleaner to define a local component that fits those specific needs.
> 
> I have no authority to enforce the refactoring effort needed, I just wanted to raise my concern.

@Arnarkari93 - I got a comment from @barabrian when I was adding image to the action card, to refactor it and this is the result. (The original PR with these changes got reverted)

I agree, it is really bloated. Would you recommend to refactor both service portal and applications to use their own local components instead?

Arnarkari93:
> I agree, it is really bloated. Would you recommend to refactor both service portal and applications to use their own local components instead?

Yes. I believe we should aim to have a more general approach when implementing these components and try to not have a specific use-case in mind. The variations of components in Figma and Storybook should only be ideas on how these components can be used together to create interfaces. They should not be a guide on how to implement the components.

disaerna:
> > I agree, it is really bloated. Would you recommend to refactor both service portal and applications to use their own local components instead?
> 
> Yes. I believe we should aim to have a more general approach when implementing these components and try to not have a specific use-case in mind. The variations of components in Figma and Storybook should only be ideas on how these components can be used together to create interfaces. They should not be a guide on how to implement the components.

Alright, I will close this PR and but the refactoring on the backlog. Thanks for this.

Arnarkari93:
My recommendation is to depreciate the ActionCard component and make a version 2 that only defines the border, padding and then renders the children elements.

PR #8239, opened 2022-08-25
chore: added local e2e bash script

busla:
## What
Bash script for developers to simplify testing e2e both from source, dist and container. 

## Why

## Screenshots / Gifs

## Checklist:

### DIFF
--- a/apps/system-e2e/entrypoint.sh
+++ b/apps/system-e2e/entrypoint.sh
@@ -3,5 +3,5 @@
 set -euo pipefail
 
 echo "Current test environment: ${TEST_ENVIRONMENT}"
-
-cypress run "$@"
+echo "Cypress args: $*"
+cypress run --browser chrome "$@"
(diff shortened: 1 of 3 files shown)


apps/system-e2e/entrypoint.sh
```diff
@@ -3,5 +3,5 @@
 set -euo pipefail
 
 echo "Current test environment: ${TEST_ENVIRONMENT}"
-
-cypress run "$@"
+echo "Cypress args: $*"
+cypress run --browser chrome "$@"
```
kristofer-lp8f, line 7:
```suggestion
${TEST_BROWSER-chrome}
cypress run --browser $TEST_BROWSER "$@"
```

busla: we are experiencing timeouts still when using Chrome for some reason. We are thinking about removing the `--browser` flag and revert back to `electron` for now.

The flag will be set explicitly in the pipeline.

pshomov (approved):
LGTM

PR #8241, opened 2022-08-25
feat(national-registry-v1): Add child registration service back in

thordurhhh:
## What

* Add child registration service back in.

## Why

* Allows parents to send a message to Þjóðskrá regarding their children's registration.

## Checklist:

### DIFF
--- a/libs/api/domains/national-registry/src/lib/dto/FamilyCorrectionInput.input.ts
+++ b/libs/api/domains/national-registry/src/lib/dto/FamilyCorrectionInput.input.ts
@@ -5,10 +5,6 @@ import { IsString } from 'class-validator'
 export class FamilyCorrectionInput {
   @Field()
   @IsString()
-  ssn!: string
-
-  @Field()
-  @IsString()
   ssnChild!: string
 
   @Field()
--- a/libs/api/domains/national-registry/src/lib/dto/FamilyCorrectionInput.input.ts
+++ b/libs/api/domains/national-registry/src/lib/dto/FamilyCorrectionInput.input.ts
@@ -5,7 +5,7 @@ import { IsString } from 'class-validator'
 export class FamilyCorrectionInput {
   @Field()
   @IsString()
-  ssnChild!: string
+  nationalIdChild!: string
 
   @Field()
   @IsString()
--- a/libs/api/domains/national-registry/src/lib/dto/FamilyCorrectionInput.input.ts
+++ b/libs/api/domains/national-registry/src/lib/dto/FamilyCorrectionInput.input.ts
@@ -9,10 +9,6 @@ export class FamilyCorrectionInput {
 
   @Field()
   @IsString()
-  name!: string
-
-  @Field()
-  @IsString()
   phonenumber!: string
 
   @Field()
--- a/libs/clients/national-registry/v1/src/lib/nationalRegistryApi.ts
+++ b/libs/clients/national-registry/v1/src/lib/nationalRegistryApi.ts
@@ -1,8 +1,4 @@
-import {
-  InternalServerErrorException,
-  NotFoundException,
-  ForbiddenException,
-} from '@nestjs/common'
+import { InternalServerErrorException, NotFoundException } from '@nestjs/common'
 import Soap from 'soap'
 
 import isEmpty from 'lodash/isEmpty'
@@ -142,7 +138,9 @@ export class NationalRegistryApi {
     )
 
     if (!response) {
-      throw new ForbiddenException(`User correction not sent. Unknown error`)
+      throw new InternalServerErrorException(
+        'User correction not sent. Unknown error',
+      )
     }
     return {
       success: response?.success,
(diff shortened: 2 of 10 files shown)


disaerna (approved):

libs/api/domains/national-registry/src/lib/dto/FamilyCorrectionInput.input.ts
```diff
@@ -3,10 +3,6 @@ import { IsString } from 'class-validator'
 
 @InputType()
 export class FamilyCorrectionInput {
-  @Field()
-  @IsString()
-  ssn!: string
-
   @Field()
   @IsString()
   ssnChild!: string
```
darrikonn, line 8: ssn should be nationalID (the common name for kennitala in island.is)

libs/clients/national-registry/v1/src/lib/nationalRegistryApi.ts
```diff
@@ -123,6 +123,33 @@ export class NationalRegistryApi {
...
+  public async postUserCorrection(
+    values: FamilyCorrection,
+  ): Promise<{ success: boolean; message?: string }> {
+    const response = await this.signal(
+      'CreateAndUpdateMS_Leidretting',
+      {
+        S5RequestID: '',
+        Kennitala: values.ssn,
+        Barn: values.ssnChild,
+        Nafn: values.name,
```
darrikonn, line 135: Should the user name not be fetched from Þjóðskrá?
Might be safer in terms of identifying the correct requester.

libs/api/domains/national-registry/src/lib/dto/FamilyCorrectionInput.input.ts
darrikonn: Additionally, what is the scope of the project here?

1. Can a person file a correction for any child?
2. Can the person only file a correction for a child that is signed as his/her family member?

darrikonn (commented):
Minor comments.

libs/clients/national-registry/v1/src/lib/nationalRegistryApi.ts
thordurhhh: In the front end we do not allow the user to set their own name, it's using the name from the Þjóðskrá response.
It feels kinda like doing the same thing twice if I were to add a lookup for the name to the same endpoint here as we do in the front end.

libs/api/domains/national-registry/src/lib/dto/FamilyCorrectionInput.input.ts
thordurhhh: * Yes good point, I will change it to nationalId.

The scope should be that only a parent/legal guardian of a child can request a correction for their child's info.

libs/clients/national-registry/v1/src/lib/nationalRegistryApi.ts
darrikonn: The web front-end is not the only client that can send in the response; so in terms of "securing" the correct name, it needs to be done on the back-end.

I can e.g. call the mutation with cURL and set any name I want.

thordurhhh: That's true, then where do we want to stop regarding the security of said information?

I can fetch the information from Þjóðskrá to verify the name, then there is another endpoint where I can verify if the logged in user is a legal guardian/parent of this child (like we do on the front end as well when displaying the children).

So I could technically add those both so we can be sure of the name and be sure that this child's SSN is part of the sender's family number.

darrikonn: Yeah, according to your answer above.. then yeah, we need to verify that the childNationalId is indeed one of the child nationalIDs of the requestor.

You can see `getFamilyMemberDetails`, which is doing the same thing.

thordurhhh: Indeed. I will change it.

libs/clients/national-registry/v1/src/lib/nationalRegistryApi.ts
```diff
@@ -123,6 +123,33 @@ export class NationalRegistryApi {
...
+        Simanumer: values.phonenumber,
+        Netfang: values.email,
+        Athugasemd: values.comment,
+        Stada: 'Skráð',
+      },
+      true,
+    )
+
+    if (!response) {
+      throw new ForbiddenException(`User correction not sent. Unknown error`)
```
darrikonn, line 145: Throw a different type of error, since this isn't really forbidden exception, and we most always take different actions based on forbidden exceptions in the apollo's error link.

darrikonn (approved):
Nice, approved.. just the different exception for better front-end error handling

libs/clients/national-registry/v1/src/lib/nationalRegistryApi.ts
thordurhhh: yes, makes sense 💯

PR #8242, opened 2022-08-25
feat(service-portal): new empty screen for accesses

disaerna:
# New empty screen

## What

Access screen was a bit lame before so we added a beautiful empty screen instead, with two buttons so user can switch access or grant access.

## Screenshots / Gifs

Screenshot 2022-08-25 at 10 11 09

Screenshot 2022-08-25 at 10 10 52

## Checklist:


libs/service-portal/core/src/screens/NoDataScreen/NoDataScreen.tsx
```diff
@@ -7,98 +7,140 @@ import {
...
+  IconProps,
 } from '@island.is/island-ui/core'
 import * as styles from '../ErrorScreen/ErrorScreen.css'
 import { Link } from 'react-router-dom'
 import { ServicePortalPath } from '../../lib/navigation/paths'
 
+interface ButtonType {
+  link?: ServicePortalPath | string
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  onClick?: any
```
thordurhhh, line 19:
```suggestion
  onClick?: () => void
```

😎

thordurhhh (approved):

libs/service-portal/settings/access-control/src/screens/AccessControl/AccessControl.tsx
```diff
@@ -1,20 +1,69 @@
...
 const AccessControl: ServicePortalModuleComponent = ({ userInfo, client }) => {
   useNamespaces('sp.settings-access-control')
+  const { data, loading } = useQuery<Query>(AuthDelegationsQuery)
+  const { switchUser } = useAuth()
+  const { formatMessage } = useLocale()
 
+  if (!loading && data?.authDelegations.length === 0) {
+    return (
+      <NoDataScreen
+        title="Umboð"
```
eirikurn, line 37: Translation?

libs/service-portal/core/src/screens/NoDataScreen/NoDataScreen.tsx
```diff
@@ -7,98 +7,139 @@ import {
...
 } from '@island.is/island-ui/core'
 import * as styles from '../ErrorScreen/ErrorScreen.css'
 import { Link } from 'react-router-dom'
 import { ServicePortalPath } from '../../lib/navigation/paths'
 
+interface ButtonType {
+  link?: ServicePortalPath | string
+  onClick?: () => void
+  icon?: Pick<IconProps, 'icon' | 'type'>
+  type: 'internal' | 'external' | 'click'
```
eirikurn, line 20: This is a lot of boilerplate for common logic. Would be nice to upgrade the design system so buttons and links can support all of this automatically.

libs/service-portal/core/src/screens/NoDataScreen/NoDataScreen.tsx
```diff
@@ -7,98 +7,139 @@ import {
...
+                  {renderButton(secondaryButton)}
+                </Box>
               )}
             </Box>
-          )}
-        </Box>
-      </GridColumn>
-
-      <GridColumn span={['1/1', '4/12']} offset={['0', '3/12']}>
-        <Box display="flex" justifyContent="center">
```
eirikurn, line 94: Can get rid of the fragment, GridRows and GridColumns since it's just one column now?

eirikurn (commented):
🙌

libs/service-portal/core/src/screens/NoDataScreen/NoDataScreen.tsx
disaerna: yes 💯  ✅

libs/service-portal/settings/access-control/src/screens/AccessControl/AccessControl.tsx
disaerna: whoops - thanks ✅

libs/service-portal/core/src/screens/NoDataScreen/NoDataScreen.tsx
disaerna: Yeah well Im getting different comments from the UI team regarding what the purpose of this UI library is. Some say it should only be a skeleton and the developer should handle more detailed actions and UI, while other want to refactor to handle all possible versions 🤷‍♀️ So I'm not gonna refactor this as well 😅

eirikurn (approved):
Lgtm

PR #8245, opened 2022-08-25
chore(core): Adding local ESLint rule to disallow async module initialisation

### LINKED ISSUE #8240: Create ESLint rule to ban async module initialisation

saevarma:
Closes #8240 

## What

* Creating a local ESLint rule that catches when developer creates a async `register` function or uses async `useFactory` for a provider in module setup.
* Fixing lint errors raised.
* Updating soap package.

## Why

* Having async code in module initialisation can cause unexpected errors or timeouts on app startup.
* The soap package had newer version including vulnerability fixes.

## Screenshots / Gifs

## Checklist:


eslint-local-rules.js
```diff
@@ -33,4 +33,42 @@ module.exports = {
...
+        noAsyncRegister:
+          'Disallowing async register functions in modules to prevent unexpected startup failures or timeouts.',
+        noAsyncProviderFactory:
+          'Disallowing async useFactory in module providers to prevent unexpected startup failures or timeouts.',
+      },
+    },
+    schema: [],
+    create: function (context) {
+      return {
+        "MethodDefinition[static=true][key.name='register'][value.async=true]": (
```
eirikurn, line 55: Did not know you could implement an eslint rule like this. It's really cool!

I wonder about the selectors though. This one might have false positives (async register functions not related to Nest) as well as false negatives. "register" is just a naming convention, "forRootAsync" is another common name for async module imports. 

Maybe we should select for what they return instead: `Promise<DynamicModule>` - though that depends on the user specifying the correct return type.

Your call. Maybe report all three selectors for now and loosen up if it's too annoying?

eslint-local-rules.js
```diff
@@ -33,4 +33,42 @@ module.exports = {
...
+          node,
+        ) => {
+          context.report({
+            node,
+            messageId: 'noAsyncRegister',
+          })
+        },
+        "ObjectExpression[properties.length=2] > Property[key.name='provide'] + Property[key.name='useFactory'][value.async=true]": (
```
eirikurn, line 63: Think this selector is too narrow. It will not catch async factories where the provide property comes after, or has more properties like `scope` and `inject`.

It's probably ok to ban all async "useFactory" properties.

eirikurn (commented):
Nice job!

Look forward to improving the soap logic. Or getting rid of soap completely 🙈

eslint-local-rules.js
saevarma: Ahh of course, fixed.

eslint-local-rules.js
```diff
@@ -33,4 +33,67 @@ module.exports = {
...
+                node,
+                messageId: 'noAsyncRegister',
+                data: {
+                  name,
+                },
+              })
+            },
+          }
+        }, {}),
+        "MethodDefinition[static=true][value.async=true] TSTypeReference[typeName.name='Promise'] TSTypeReference[typeName.name='DynamicModule']": (
```
eirikurn, line 80: Can skip the async check here.

```suggestion
        "MethodDefinition[static=true] TSTypeReference[typeName.name='Promise'] TSTypeReference[typeName.name='DynamicModule']": (
```

eirikurn (approved):
Sweet! Super impressive eslint ast wrangling. Learning a lot :)

obmagnusson (approved):
LGTM

disaerna (approved):
🚀

kksteini (approved):

PR #8248, opened 2022-08-25
feat(j-s): multiple police case numbers

Arnarkari93:
[Asana - Ákærði/Varnaraðili: Mörg LÖKE-númer](https://app.asana.com/0/1199153462262248/1202805727084389/f)

## What

- Add support for having multiple police case numbers in a Case

## Why

- Required for indictment cases, nice to have for other cases

## Screenshots / Gifs

Large GIF (850x514)

## Checklist:


apps/judicial-system/backend/src/app/formatters/formatters.ts
```diff
@@ -141,19 +141,19 @@ export function formatCourtResubmittedToCourtSmsNotification(
...
-  policeCaseNumber: string,
+  policeCaseNumbers: string[],
   overviewUrl: string,
   courtName?: string,
 ) {
   const subject = formatMessage(
     notifications.defenderResubmittedToCourt.subject,
     {
-      policeCaseNumber,
+      policeCaseNumber: policeCaseNumbers.join(', '),
```
Arnarkari93, line 151: There are many changes like this one. I decided not to change the parameter name to the contentful string to avoid having to change the string.

apps/judicial-system/web-e2e/src/integration/Prosecutor/Indictments/defendant.spec.ts
```diff
@@ -29,31 +27,9 @@ describe(CREATE_INDICTMENT_ROUTE, () => {
     cy.getByTestid('inputErrorMessage').should('not.exist')
     cy.getByTestid('continueButton').should('be.disabled')
 
-    // Birthday
-    cy.get('[type="checkbox"]').check()
```
Arnarkari93, line 33: Removed in favor of nationalid lookup

apps/judicial-system/xrd-api/src/app/app.dto.ts
```diff
@@ -11,9 +17,11 @@ export class CreateCaseDto {
   readonly type!: CaseType
 
   @IsNotEmpty()
-  @IsString()
+  @IsArray()
+  @ArrayMinSize(1)
+  @IsString({ each: true })
   @ApiProperty()
-  readonly policeCaseNumber!: string
+  readonly policeCaseNumbers!: string[]
```
gudjong, line 24: This cannot be changed at this time as the change will break our contract with LÖKE.

apps/judicial-system/backend/src/app/modules/case/dto/internalCreateCase.dto.ts
```diff
@@ -11,9 +17,11 @@ export class InternalCreateCaseDto {
   readonly type!: CaseType
 
   @IsNotEmpty()
-  @IsString()
+  @IsArray()
+  @ArrayMinSize(1)
+  @IsString({ each: true })
   @ApiProperty()
-  readonly policeCaseNumber!: string
+  readonly policeCaseNumbers!: string[]
```
gudjong, line 24: This should probably not be changed as this is called form the xrd-api which in turn is called from LÖKE. We must not break our contract with LÖKE.

gudjong (changes requested):

apps/judicial-system/xrd-api/src/app/app.dto.ts
Arnarkari93: True, forgot about that. I will update so that the endpoint still accepts a single number and call it internally with a list.

apps/judicial-system/api/src/app/modules/case/dto/updateCase.input.ts
```diff
@@ -29,8 +29,8 @@ export class UpdateCaseInput implements UpdateCase {
   readonly description?: string
 
   @Allow()
-  @Field({ nullable: true })
-  readonly policeCaseNumber?: string
+  @Field(() => [String], { nullable: true })
```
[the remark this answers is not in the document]
Arnarkari93: Looks like it is not needed. The API generates the field with `[String!]!` 
Screenshot 2022-08-26 at 11 01 13

apps/judicial-system/backend/src/app/modules/case/dto/internalCreateCase.dto.ts
Arnarkari93: Fixed in `xrd-api` see https://github.com/island-is/island.is/pull/8248/commits/ccbd32566913b4755f7fcb377ec9a3cec5f876f5

gudjong (approved):

PR #8272, opened 2022-08-29
fix(fa): Add size and type restriction to upload files

davidpxn:
# ... 👆

https://app.asana.com/0/1202167941440767/1202882546493116

## What

- Add constants to restrict the size and types of attachments
- Handle when too big files are uploaded and add error message for that

## Why

So users can not send problematic file types or really big files to the database.
This problem was spotted after a security review.

## Checklist:


MargretFinnboga (approved):
cool 🦈

PR #8279, opened 2022-08-30
feat(get-secrets): Better secret rendering

kristofer-lp8f:
## What
Render secrets better, removing newlines and apostrophe, wrapping the secret in a non-expanding `'literal string`.

## Why
Several projects use problematic characters like `$`, `'`, `"`, or `{`. This causes an error when sourcing the `.env.secret` file.

## Checklist:


kristofer-lp8f (commented):
Replace all

infra/src/cli/render-secrets.ts
```diff
@@ -66,7 +66,10 @@ export const renderSecrets = async (service: string) => {
   const values = await getParams(secretRequests.map(([_, ssmName]) => ssmName))
 
   secretRequests.forEach(([envName, ssmName]) => {
-    console.log(`export ${envName}=${values[ssmName]}`)
+    const escapedValue = values[ssmName]
+      .replace(/\s+/g, ' ')
+      .replace(/'/g, "'\\''")
```
pshomov, line 71: I am not sure about this but let's see what happens

pshomov (approved):
LGTM

PR #8287, opened 2022-08-30
feat(j-s): Upload Case to Police

gudjong:
# Upload Case to Police

[Vista úrskurð í LÖKE gegnum vefþjónustu](https://app.asana.com/0/1199153462262248/1201599798492597/f)

## What

- Uploads a case to police after ruling signature.

## Why

- To better serve prosecutors.

## Checklist:


apps/judicial-system/backend/src/app/modules/case/case.service.ts
```diff
@@ -472,6 +474,23 @@ export class CaseService {
       })
   }
 
+  private async deliverCaseToPolice(theCase: Case): Promise<boolean> {
+    const pdf = await getCourtRecordPdfAsString(theCase, this.formatMessage)
```
Arnarkari93, line 478:
```suggestion
    await this.refreshFormatMessage()
    const pdf = await getCourtRecordPdfAsString(theCase, this.formatMessage)
```

apps/judicial-system/backend/src/app/modules/case/case.service.ts
```diff
@@ -472,6 +474,23 @@ export class CaseService {
...
+    const defendantNationalIds = theCase.defendants
+      ?.filter((defendant) => !defendant.noNationalId && defendant.nationalId)
+      .map((defendant) => defendant.nationalId ?? '')
```
Arnarkari93, line 481: Consider using `flatMap` instead to get an array with guaranteed `nationalid` values.
```suggestion
      ?.flatMap((defendant) => !defendant.noNationalId && defendant.nationalId ? [defentant.nationalId] : [])
```

oddsson, line 481: Is the `?? ''` needed here, because you've already filtered out all defendants that don't have a value for nationalId. I know typescript is sometimes a bit weird when it comes to these kinds of checks, but I just wanted to make sure.

apps/judicial-system/backend/src/app/modules/police/police.service.ts
```diff
@@ -10,44 +10,71 @@
...
+      config.tlsBasePathWithEnv,
+      XRoadMemberClass.GovernmentInstitution,
+      config.policeMemberCode,
+      config.policeApiPath,
+    )
+    this.agent = new Agent({
+      cert: config.clientCert,
+      key: config.clientKey,
+      ca: config.clientPem,
+      rejectUnauthorized: false,
```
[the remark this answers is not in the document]
oddsson: Is this something we need to worry about?

oddsson (changes requested):
Two questions

Arnarkari93 (changes requested):
Looks good, just one suggestion.

apps/judicial-system/backend/src/app/modules/police/police.service.ts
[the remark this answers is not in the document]
gudjong: No, this warning pops up every time I modify xRoad code, but the xRoad configuration requires this setting.

apps/judicial-system/backend/src/app/modules/case/case.service.ts
gudjong: Well spotted and you would think this was unnecessary. However, the type inference is not smart enough to figure this out and that is why added this redundant code.

apps/judicial-system/backend/src/app/modules/case/case.service.ts
gudjong: The refresh is actually done a bit earlier in the call chain, i.e. in the public method `deliver`. That method calls a number of `deliver...` methods some of which use `formatMessage`. So I don't think a refresh is necessary here.

apps/judicial-system/backend/src/app/modules/case/case.service.ts
gudjong: I did not know about that method. Will try it out.

gudjong: The compiler complains that `flatMap` does not exist and suggests changing the 'lib' compiler option to 'es2019' or later. Probably not something we can do here.

gudjong: I managed to simplify using `reduce` instead and the type inference is correct.

oddsson (approved):

Arnarkari93 (approved):

PR #8289, opened 2022-08-30
feat(web): Organization page sidebar card update

RunarVestmann:
# Organization page sidebar card update

## What

* Added support for mailing list signup in sidebar of organization pages
* Added support for connected components in sidebar of organization pages (now we're using a connected component to render a newscard for the latest news of an organization)
* I added box props to the newsletter signup component to be able to make the content fit inside the sidebar
* Removed default_with_image theme from contentful types since it's not used and was updated in Contentful

## Why

* This was requested by Stafrænt Ísland

## Screenshots / Gifs

image
## Checklist:


apps/web/components/Organization/Wrapper/OrganizationWrapper.tsx
```diff
@@ -434,18 +439,76 @@ export const OrganizationWrapper: React.FC<WrapperProps> = ({
...
+                            inputAndButtonContainerBoxProps={{
+                              display: 'flex',
+                              flexDirection: 'column',
+                              alignItems: 'flexStart',
+                            }}
+                            buttonBoxProps={{
+                              marginTop: 3,
+                            }}
+                            variant="white"
+                            namespace={namespace ?? (({} as unknown) as any)}
```
Ingimarsson, line 504: Is there no better type that can be used?

libs/cms/src/lib/models/sidebarCard.model.ts
```diff
@@ -24,11 +25,15 @@ export class SidebarCard {
...
+export const mapSidebarCard = ({
+  sys,
+  fields,
+}: ISidebarCard): SystemMetadata<SidebarCard> => ({
+  typename: 'SidebarCard',
   id: sys.id,
   type: fields.type ?? '',
   title: fields.title ?? '',
-  content: fields.content ?? '',
+  contentString: fields.content ?? '',
```
Ingimarsson, line 36: Any reason for this?

Ingimarsson (approved):

libs/cms/src/lib/models/sidebarCard.model.ts
RunarVestmann: Yes, we can't have two fields with the same name in fragments.ts sadly so I had to rename this

apps/web/components/Organization/Wrapper/OrganizationWrapper.tsx
RunarVestmann: I changed the namespace type so this has been addressed :)

apps/web/components/MailingListSignup/MailingListSignup.tsx
```diff
@@ -21,7 +20,7 @@ interface FormProps {
 }
 
 interface MailingListSignupProps {
-  namespace: GetNamespaceQuery['getNamespace']
+  namespace: Record<string, string>
```
kksteini, line 23: Hmmm... are the generated types failing?
I'd be interested to know how

RunarVestmann: We are JSON parsing the object before it is passed in as props so the typing is wrong. I could've changed this in another PR but I changed it here.

kksteini (approved):

PR #8292, opened 2022-08-31
fix(ads): Void national ids handled against invalid json

kksteini:
The National Registry Client typing indicates that address/postalcode will always have a value but sometimes this is not the case. That can happen with system national ids (is: kerfiskennitölur) and some national ids that do not have address/postalcode attached.

This causes the generated OpenAPI client itself to fail with 'invalid-json'.

Instead of the backend failing (500) we can return a null user with a postalcode outside of the ADS region to indicate that the user doesn't have any rights.

## Checklist:


apps/air-discount-scheme/backend/src/app/modules/nationalRegistry/nationalRegistry.service.ts
```diff
@@ -343,4 +344,24 @@ export class NationalRegistryService {
     }
     throw error
   }
+
+  private handleInvalidJson(error: { type: string }): Einstaklingsupplysingar {
+    // OpenAPI client fails when national ids do not contain the promised information,
+    // (f.x. kerfiskennitala). The client fails with 'invalid-json'.
```
kristofer-lp8f, line 350:
```suggestion
    // (e.g. kerfiskennitala). The client fails with 'invalid-json'.
```

apps/air-discount-scheme/backend/src/app/modules/nationalRegistry/nationalRegistry.service.ts
```diff
@@ -343,4 +344,24 @@ export class NationalRegistryService {
...
+    // Instead of the backend failing we return a null user with no ADS rights.
+    if (error.type === 'invalid-json') {
+      return {
+        bannmerking: false,
+        faedingardagur: new Date(),
+        kennitala: '0000000000',
+        kynkodi: '1',
```
kristofer-lp8f, line 357:
```suggestion
        kynkodi: '-1',
```

apps/air-discount-scheme/backend/src/app/modules/nationalRegistry/nationalRegistry.service.ts
```diff
@@ -343,4 +344,24 @@ export class NationalRegistryService {
...
+        nafn: '',
+        adsetur: {
+          heiti: '',
+          postnumer: '100',
```
kristofer-lp8f, line 361:
```suggestion
          postnumer: '-100',
```

kristofer-lp8f (changes requested):
LGTM :+1: 

f.x. not valid in English, though.

kristofer-lp8f (approved):

PR #8296, opened 2022-08-31
chore(j-s): Court API Availability

gudjong:
# Court API Availability

[Ekki tala við Auði í umhverfum þar sem hún er ekki aðgengileg](https://app.asana.com/0/1199153462262248/1202882386999270/f)

## What

- Avoid calling the court API in environments where it is not available.

## Why

We want to prevent unnecessary delays when responding to user requests and we want to avoid spamming error logs.

## Checklist:


oddsson (approved):

apps/judicial-system/backend/src/app/modules/court/court.service.ts
```diff
@@ -117,12 +129,15 @@ export class CourtService {
     fileName: string,
     content: Buffer,
   ): Promise<string> {
-    return this.uploadStream(
-      courtId,
-      `${fileName}.pdf`,
-      'application/pdf',
-      content,
-    )
+    return this.confirmApiAvailability()
```
Arnarkari93, line 132: I feel like there should be a safer way of doing this. One could easily forget to call this method first before trying to access `this.courtClientService`. When this config is set, could we inject a striped-out implementation of the `CourtClientService`?

Arnarkari93 (changes requested):
One thought.

apps/judicial-system/backend/src/app/modules/court/court.service.ts
gudjong: You are right of course, but I am not sure it is worth the effort. I use the same approach in the police service when calling LÖKE. I'm going to ponder this a bit and see if I can easily come up with a more elegant solution.

Arnarkari93: I see, I will take the auto merge label off and accept. Then you can let this through if you want.

Arnarkari93 (approved):

Arnarkari93 (approved):

PR #8300, opened 2022-08-31
fix(p-sign-application): Email and phone from user profile and answers

alexdiljar:
Showing email and phone from userprofile for applicant in info section and from answers in overview.


albinagu (approved):
aaa, right! nice! ✨

PR #8309, opened 2022-09-01
feat(financial-statements-inao): Client Financial Limit

fjandakornid:
# Client Financial Limit

## What

Method to check what spending amount is allowed before candidates have to return financial statements.

## Why

Need to know if candidates need to return financial statement or just a statement that they didn't exceed spending limits.

## Checklist:


programm-ingovals (approved):
lgtm

PR #7189, opened 2022-04-20
chore(core): Monorepo upgrade

saevarma:
# Upgrade monorepo dependencies

## What

This PR includes major version upgrades our most important packages:
- Jest 27
- Nx 13
- NestJS 8
- NextJS 12
- RxJS 7
- Sequelize 6
- Sequelize-typescript 2
- Webpack 5

See further description in [Notion](https://www.notion.so/Nx-13-0-1-a77cd802539e468596da6265700ff3cf)

## Why

The repo was falling behind on version upgrades. So it was necessary to upgrade our most important dependencies.

## Screenshots / Gifs

Attach Screenshots / Gifs to help reviewers understand the scope of the pull request

## Checklist:

- [ ] I have performed a self-review of my own code
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] Formatting passes locally with my changes
- [ ] I have rebased against main before asking for a review

PR #8317, opened 2022-09-02
chore(core): Fix circular dependency dev regression in node apps

eirikurn:
## What

Configure typescript in node-based webpack builds to use commonjs module syntax.

Fixes circular dependency crashes in air-discount-scheme-api, air-discount-scheme-backend, financial-aid and probably more.

## Why

- Workaround for a regression in webpack 5 (#7189) which does not "deal with" circular dependencies as well as it used to with es/harmony based imports.
- The regression became active after #8192 which started using webpack 5 again for local development to workaround esbuild watch issues.
- New NX monorepos use webpack 5 and this same tsconfig configuration for node-based apps.
- However, telling esbuild to emit commonjs modules causes more circular dependency errors. So we override this config for webpack only.

The actual errors happen because typescript reflection metadata tries to access exports from modules through barrel files which have not been required yet. [Barrel files](https://github.com/basarat/typescript-book/blob/master/docs/tips/barrel.md) make circular dependencies more likely to cause issues so we recommend avoiding barrel files in general except in the root of libraries.


saevarma (approved):
🙏

PR #8328, opened 2022-09-05
chore(infra): upgrade nestjs/graphql to 8.0.2

Snojo:
There shouldn't be any breaking changes, but we're doing a feature deployment to make sure. 

This should fix the issue regarding random 500 errors on prod we've been having. 
Context: https://islandis.slack.com/archives/C01C7C3M2J3/p1662116560466819

## What

## Why

## Screenshots / Gifs

## Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,7 +76,7 @@
     "@nestjs/common": "8.4.4",
     "@nestjs/config": "0.5.0",
     "@nestjs/core": "8.4.4",
-    "@nestjs/graphql": "7.9.11",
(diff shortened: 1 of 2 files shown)


eirikurn:
It's weird that the nx migration left our @nestjs/graphql. We currently have peer dependency issues since they're on different versions, which this PR fixes.

Can you review the other nestjs packages we depend on and see if there are more peer dependency issues which we can fix?

eirikurn (approved):

PR #8340, opened 2022-09-06
chore(ci): Build schemas only for affected.

peturgq:
# ...

[Issue](https://github.com/island-is/infrastructure/issues/853)

## What
Make schema builds faster

## Why
Inefficient use of nx commands causing additional overhead in schema generation.

#### Locally on my T14s:
Before:
image

After:
image

## Checklist:

### DIFF
--- a/scripts/schemas.js
+++ b/scripts/schemas.js
@@ -39,14 +39,11 @@ const main = async () => {
     console.log(`--> Running command for ${target}\n`)
 
     try {
-      await exec(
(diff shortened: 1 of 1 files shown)


darrikonn (approved):

darrikonn (changes requested):
This change can only be merged if our dependency tree has been revised.
e.g. `air-discount-scheme-web` should depend on `air-discount-scheme-api`. So making changes to the GraphQL schema in the `api` will break `web` with these changes, since the schema won't be up-to-date since web wasn't targeted.

PR #8350, opened 2022-09-06
feat(web): Form field can decide form email recipient

RunarVestmann:
# Form field can decide form email recipient

## What

* Feature that allows us to dynamically pick what email gets sent the form message

## Why

* We want to be able to decide what email gets the form message depending on what the user selects in the form
* For example if the user selects "Akureyri" from a dropdown we'd like the recipient to be some e-mail related to Akureyri.

## Checklist:


apps/web/components/Form/Form.tsx
```diff
@@ -318,6 +324,7 @@ export const Form = ({ form, namespace }: FormProps) => {
           <Stack space={4}>
             {form.fields.map((field) => {
               const slug = slugify(field.title)
```
Ingimarsson, line 328: Is this unused?

Ingimarsson (approved):

libs/api/domains/communications/src/lib/communications.service.ts
```diff
@@ -135,6 +135,19 @@ export class CommunicationsService {
...
+    if (input.recipientDeciderValue && form.recipientFormFieldDecider) {
+      const recipientAccordingToCMS =
+        form.recipientFormFieldDecider.recipientDecider?.[
+          input.recipientDeciderValue
+        ]
+      if (recipientAccordingToCMS) {
+        recipient = recipientAccordingToCMS
+      }
+    }
+
```
kristofer-lp8f, line 150: Would something like this also work? How does `input.recipientDeciderValue` play into this? I'm just wondering whether we can remove a nesting level without resorting to long ternaries.
```suggestion
    const recipientAccordingToCMS =
        form.recipientFormFieldDecider?.recipientDecider?.[
          input.recipientDeciderValue
        ]
    let recipient = recipientAccordingToCMS ?? form.recipient
```

kristofer-lp8f (commented):
The new variables have repetitive names and it's a bit unclear what they do/contain.

I like the change :)

Someone else with more front-end skills than me must approve.

libs/api/domains/communications/src/lib/communications.service.ts
RunarVestmann: Nice suggestions, I'll change this :)

RunarVestmann: Your suggestion seems to cause a lint error since we are potentially indexing with undefined
image

RunarVestmann: This however works 
image

kristofer-lp8f (changes requested):
I'm having a hard time understanding how the end-user sees the form.

Can you please change the `recipientDeciderFormLongObscureThing` to something more legible? Like `recipientDeciderValue -> recipientName` or `recipientDecider -> cmsRecipients`?

libs/cms/src/lib/generated/contentfulTypes.d.ts
```diff
@@ -816,6 +819,9 @@ export interface IFormFieldFields {
 
   /** Options */
   options?: string[] | undefined
+
+  /** Recipient Decider */
+  recipientDecider?: Record<string, any> | undefined
```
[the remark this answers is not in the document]
RunarVestmann: This is generated from contentful, so I'm not sure how to fix that. Even if I do change the file, the next time someone generates contentful types it'll be overwritten

apps/web/screens/queries/fragments.ts
```diff
@@ -500,10 +500,19 @@ export const slices = gql`
...
     aboutYouHeadingText
     questionsHeadingText
+    recipientFormFieldDecider {
+      title
+      placeholder
+      type
+      required
+      options
+      recipientDecider
+    }
```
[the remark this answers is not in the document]
RunarVestmann: We do, but i'll omit the recipientDecider (now called emailConfig cuz I changed it) since there's no need for the frontend to want that info.

libs/api/domains/communications/src/lib/communications.service.ts
```diff
@@ -135,6 +135,16 @@ export class CommunicationsService {
       return false
     }
 
+    let recipient = form.recipient
+
+    const emailConfig = form.recipientFormFieldDecider?.emailConfig
+    const key = input.recipientFormFieldDeciderValue as string
```
kristofer-lp8f, line 141: This might still become the key `"undefined"`, which might make debugging _very_ hard if for any reason `emailConfig` unintentionally has the `"undefined"` key.

libs/api/domains/communications/src/lib/communications.service.ts
```diff
@@ -135,6 +135,16 @@ export class CommunicationsService {
...
+
+    // The CMS might have a form field which decides what the recipient email address is
+    if (emailConfig && emailConfig[key]) {
```
kristofer-lp8f, line 144:
```suggestion
    const emailConfig = form.recipientFormFieldDecider?.emailConfig
    const key: string|undefined = input.recipientFormFieldDeciderValue

    // The CMS might have a form field which decides what the recipient email address is
    if (!!key && emailConfig && emailConfig[key as string]) {
```

kristofer-lp8f (changes requested):
Minor safety check to prevent future headaches.

Otherwise, LGTM.

libs/api/domains/communications/src/lib/communications.service.ts
```diff
@@ -135,6 +135,12 @@ export class CommunicationsService {
       return false
     }
 
+    // The CMS might have a form field which decides who the recipient is
+    const recipientAccordingToCMS =
+      form.recipientFormFieldDecider?.recipientDecider?.[
+        input?.recipientDeciderValue as string
+      ]
```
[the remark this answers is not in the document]
kristofer-lp8f: Fix suggested below

libs/api/domains/communications/src/lib/communications.service.ts
RunarVestmann: I've changed it

kristofer-lp8f (approved):

PR #8355, opened 2022-09-07
feat(web): Don't show the FAQ List title in navigation if it's hidden on Life Event page

RunarVestmann:
# Don't show the FAQ List title in navigation if it's hidden on Life Event page

## What

* You can choose whether the title of a FAQ List gets displayed or not but even though it's hidden it still gets displayed in the navigation on the right
* This change addresses that issue
* I also added a check for sectionwithimage slice since we don't want other slices (unless they have a showTitle) field to appear in the navigation

## Why

* We don't want hidden titles to appear in the navigation (this includes titles of images and faqlist titles.

### Before (Um pósthólfið is a hidden title)
image

### After
image

## Checklist:


apps/web/utils/navigation.ts
```diff
@@ -12,9 +12,12 @@ interface Navigatable {
...
 
-const isNavigatable = (slice: any): slice is Navigatable => {
-  return typeof slice === 'object' && slice.id && slice.title
-}
+const isNavigatable = (slice: Slice) =>
+  typeof slice === 'object' &&
+  slice['id'] &&
+  slice['title'] &&
+  slice.__typename !== 'Image' &&
+  (slice['showTitle'] === undefined || slice['showTitle'])
```
kristofer-lp8f, line 20: I find this clearer. Maybe add a comment on the intended behaviour.
```suggestion
  (slice['showTitle'] ?? true)
```

kristofer-lp8f (changes requested):

apps/web/utils/navigation.ts
RunarVestmann: I've added a comment and changed the line

apps/web/utils/navigation.ts
```diff
@@ -13,7 +13,15 @@ interface Navigatable {
 }
 
 const isNavigatable = (slice: any): slice is Navigatable => {
```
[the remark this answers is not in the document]
RunarVestmann: I changed the name

kristofer-lp8f (approved):

PR #8362, opened 2022-09-08
chore(web): Increase number of articles fetched on category page

RunarVestmann:
# Increase number of articles fetched on category page

## What

* We've reached the currently set limit of articles fetched for some categories (meaning that new articles will not appear on the category page)
* I set the limit to be a 1000 (which is the same number used for the island.is/s page)

## Why

* We want to be able to see new articles on category pages even though there are more than 150 articles

## Screenshots / Gifs

## Checklist:


kksteini (approved):
LGTM

P.S:
> We want to be able to see new articles on category pages even though there are more than 150 articles

Increasing the limit will fix that for a long time.
Maybe sorting by popular or by latest is more future proof? We may want to look into that as a continuation.

PR #8368, opened 2022-09-08
fix(service-portal): Display delegated applications on overview if permitted.

Thorrik58:
# Display delegated applications on overview if permitted.

## What

## Why

## Screenshots / Gifs

## Checklist:


libs/application/api/core/src/lib/application/application.service.ts
```diff
@@ -128,8 +129,13 @@ export class ApplicationService {
         [Op.and]: [
           {
             [Op.or]: [
-              { applicant: nationalId },
-              { assignees: { [Op.contains]: [nationalId] } },
+              ...(actor
+                ? [
+                    { applicant: nationalId },
+                    { assignees: { [Op.contains]: [nationalId] } },
```
alexdiljar, line 135: assignees should probably be actors and checking if it contains the actor nationalId.. right?

Thorrik58: Oops!

obmagnusson (approved):
LGTM

PR #8376, opened 2022-09-09
chore: Implement more methods in NationalRegistryClientService

eirikurn:
## What

- Implement more methods in NationalRegistryClientService and create related interfaces
- Replace the use of the (Icelandic-language) generated code from across the monorepo.
- Unify national registry error handling.

## Why

- Because error handling involved a lot of boilerplate which was out of date in many many places.
- The generated api classes are in Icelandic and are a bit difficult to work with.
- To simplify the use of the NationalRegistry API.

## Screenshots / Gifs

## Checklist:


libs/api/domains/national-registry-x-road/src/lib/nationalRegistryXRoad.resolver.ts
```diff
@@ -54,10 +46,10 @@ export class NationalRegistryXRoadResolver {
     @Context('req') { user }: { user: User },
     @Parent() person: NationalRegistryPerson,
   ): Promise<NationalRegistryPerson[] | undefined> {
-    return this.nationalRegistryXRoadService.getChildrenCustodyInformation(
-      user,
-      person.nationalId,
-    )
+    if (person.nationalId !== user.nationalId) {
```
darrikonn, line 49: Wondering if we should handle this in GraphQL fields using the filterFields Extension. Check identity.model in api/domains/identity for reference.

Since this can apply to more fields resolvers than this one.

eirikurn: As it is, users can't query anyone except their-self from this domain. This code is not concerned with filtering fields, as I don't know which fields should be filtered should this domain be opened up more.

This code is just dealing with the fact that - with/without field filtering - it is impossible to query children for another person.

eirikurn: I think it makes sense to use the filterFields extension if/when we open this domain up. It would be part of that project to decide which fields are available.

kksteini (approved):

apps/air-discount-scheme/backend/src/app/modules/user/user.service.ts
```diff
@@ -29,52 +18,15 @@ export class UserService {
   constructor(
     private readonly flightService: FlightService,
     private readonly nationalRegistryService: NationalRegistryService,
-    private readonly nationalRegistryIndividualsApi: EinstaklingarApi,
```
kksteini, line 32: The EinstaklingarAPI came from the NationalRegistryClientModule if I'm not mistaken. If that isn't used anymore then I think the following changes should happen to `user.module.ts`

```diff
...
-import { NationalRegistryClientModule } from '@island.is/clients/national-registry-v2'
import { CacheModule } from '../cache'
import { PrivateUserController } from './user.controller'

@Module({
  imports: [
    FlightModule,
    NationalRegistryModule,
-    NationalRegistryClientModule,
    CacheModule,
  ],
...
```

eirikurn: Great catch!

Snojo (approved):
DevOps LGTM

libs/clients/national-registry/v2/src/lib/nationalRegistryClient.service.ts
```diff
@@ -1,33 +1,130 @@
...
+        .withMiddleware(new AuthMiddleware(parentUser))
+        .einstaklingarGetForsjaForeldriRaw({
+          id: parentUser.nationalId,
+          barn: childId,
+        }),
+    )
+    return response || []
+  }
+
+  async getCohabitationInfo(
```
saevarma, line 57: I would use Maritial Status instead of cohabitation as the english for "hjúskaparstaða".
https://www.skra.is/english/forms/single-product/?productid=5d5773a3-137d-11e6-943e-005056851dd2
```suggestion
  async getMaritialStatusInfo(
```

libs/clients/national-registry/v2/src/lib/types/cohabitation.dto.ts
```diff
@@ -0,0 +1,40 @@
+import { Hjuskapur } from '../../../gen/fetch'
+import { logger } from '@island.is/logging'
+
+export interface CohabitationDto {
```
saevarma, line 4: See maritial status comment.

libs/clients/national-registry/v2/src/lib/types/family.dto.ts
```diff
@@ -0,0 +1,21 @@
...
+export function formatFamilyDto(
+  family: Fjolskyldumedlimir | null,
+): FamilyDto | null {
+  if (family == null || family.fjolskyldunumer == null) {
+    return null
+  }
+  return {
+    familyId: family.fjolskyldunumer,
+    individuals: (family.einstaklingar ?? []).map(
+      (individual) => formatIndividualDto(individual)!,
```
saevarma, line 18: Could you use the `is` operator to skip the non-null assertion, if not please add es-lint ignore for the line.

package.json
```diff
@@ -409,6 +409,7 @@
     "mailparser": "3.5.0",
     "next-secure-headers": "2.1.0",
     "nx": "13.10.3",
+    "openapi-typescript-codegen": "0.23.0",
```
saevarma, line 412: 🙌

saevarma (changes requested):
Really nice, I'm so happy to finally get the national registry client wrapped in english.

package.json
eirikurn: Whoops. This was accidental. 😅

libs/clients/national-registry/v2/src/lib/nationalRegistryClient.service.ts
eirikurn: Hm ... this is the translation Þjóðskrá uses in their API, which I assume follows their formal [ontology](https://en.wikipedia.org/wiki/Ontology_(information_science)): 

https://island.is/throun/vefthjonustur/SVNfR09WXzY1MDM3NjA2NDlfU0tSQS1Qcm90ZWN0ZWRfRWluc3Rha2xpbmdhcg#operation/Einstaklingar_GetHjuskapur

There are examples of both naming conventions in our monorepo. I would lean toward using their naming if I hadn't already renamed some properties to fit OIDC naming. 😇

libs/api/domains/national-registry-x-road/src/lib/nationalRegistryXRoad.service.ts
```diff
@@ -16,205 +14,172 @@ import { NationalRegistryCitizenship } from '../models/nationalRegistryCitizensh
...
-        id: parentNationalId,
-      })
-      .catch(this.handle404)
-
-    if (!childrenNationalIds) {
+    const childrenNationalIds = await this.nationalRegistryApi.getCustodyChildren(
+      parentUser,
+    )
+
+    const parentAFamily = await this.nationalRegistryApi.getFamily(
```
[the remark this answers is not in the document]
eirikurn: Oh this looks like a regression. Great catch!

libs/clients/national-registry/v2/src/lib/nationalRegistryClient.service.ts
saevarma: 🙈

saevarma (approved):

davidpxn (approved):

PR #8378, opened 2022-09-09
fix(fa): Add safer-buffer to build

davidpxn:
# ... 👆

## What

The financial-aid apps (Veita and Ósk) are not working on DEV and it looks like they are having trouble with the financial-aid-api. The api is crashing and complaining that it is missing the safer-buffer package. 
It is not certain that this change will fix this but it should not break anything. 
The weird thing is that we only get an error on the dev environment but not locally. 

## Checklist:


MargretFinnboga (approved):
osom 🚀

PR #8397, opened 2022-09-12
chore(ci): Upgrade dependencies and change package managers

peturgq:
####  To start off we're going to upgrade a few things:
- Upgrade Yarn to berry(2+)
- Upgrade Node to 18
- Upgrade Typescript to 4.4.2
- Upgrade graphql to 16.6.0
- Upgrade esbuild to 0.15.7
- Upgrade jest to 28.1.3
- Upgrade apollo-server-express to 3.10.2

Positive side effect of upgrading node and typescript is you are forced to update some other packages. 

Yarn is being upgraded to berry because this allows us to utilize some new methods to optimize things surrounding node_modules in our builds.

Typescript is being upgraded because of: https://github.com/yarnpkg/berry/issues/3275 
The changes introduced some more type safety enforcing in the compiler, so some additional changes in code were necessary.

## Issues pending fixes
- Running code coverage finds broken code which is usually caused by the new typescript compiler being stricter
## Checklist:


disaerna (approved):

.gitignore
```diff
@@ -15,6 +15,12 @@ libs/contentful-extensions/parent-slug/.cache
 
 # dependencies
 node_modules/
+.yarn/*
+!.yarn/patches
+!.yarn/releases
+!.yarn/plugins
+!.yarn/sdks
+!.yarn/versions
```
eirikurn, line 23: Why are we committing yarn releases and more in the monorepo?

apps/air-discount-scheme/web-e2e/src/integration/app.spec.ts
```diff
@@ -1,4 +1,4 @@
-describe('air-discount-scheme-web', () => {
+xdescribe('air-discount-scheme-web', () => {
```
eirikurn, line 1: Whaaat? Remove instead?

apps/air-discount-scheme/web/i18n/withLocale.tsx
```diff
@@ -1,6 +1,6 @@
 import React from 'react'
 import { NextComponentType } from 'next'
-import { BaseContext, NextPageContext } from 'next/dist/next-server/lib/utils'
+import { BaseContext, NextPageContext } from 'next/dist/shared/lib/utils'
```
eirikurn, line 3: Why are these changes needed if we didn't change the nextjs dependency?

apps/air-discount-scheme/web/tsconfig.json
```diff
@@ -7,7 +7,8 @@
     "forceConsistentCasingInFileNames": true,
     "noEmit": true,
     "resolveJsonModule": true,
-    "isolatedModules": true
+    "isolatedModules": true,
+    "incremental": true
```
eirikurn, line 11: Exciting :)

apps/api/src/app/graphql.middleware.ts
```diff
@@ -1,4 +1,5 @@
 import { FieldMiddleware, NextFn, MiddlewareContext } from '@nestjs/graphql'
+import { any } from 'zod'
```
eirikurn, line 2: Accidental?

apps/github-actions-cache/src/app/health.ts
```diff
@@ -6,7 +6,7 @@ export const healthCheck = async (_: any, res: express.Response) => {
 }
 
 export const livenessCheck = async (_: any, res: express.Response) => {
-  if (cache.ping()) {
+  if (await cache.ping()) {
```
eirikurn, line 9: Yay typescript!

apps/judicial-system/backend/src/app/modules/case/case.service.ts
```diff
@@ -993,7 +993,10 @@ export class CaseService {
   }
 
   async addCompletedCaseToQueue(caseId: string): Promise<string> {
-    return this.queueService.add({ type: MessageType.CASE_COMPLETED, caseId })
+    return await this.queueService.add({
+      type: MessageType.CASE_COMPLETED,
+      caseId,
+    })
```
eirikurn, line 999: Neccessary? Some recommend skipping await when you can return a "compatible" promise.

apps/judicial-system/web/src/routes/Prosecutor/components/PoliceCaseFiles/PoliceCaseFiles.tsx
```diff
@@ -128,6 +128,7 @@ const PoliceCaseFiles: React.FC<Props> = ({
 
   const [policeCaseFiles, setPoliceCaseFiles] = useState<PoliceCaseFilesData>()
   useEffect(() => {
+    //if (typeof(policeDataError?.graphQLErrors[0].extensions?.code) !== "unknown")
```
eirikurn, line 131: Remove?

apps/judicial-system/web/src/routes/Prosecutor/components/PoliceCaseFiles/PoliceCaseFiles.tsx
```diff
@@ -151,7 +152,7 @@ const PoliceCaseFiles: React.FC<Props> = ({
         files: policeData ? policeData.policeCaseFiles : [],
         isLoading: false,
         hasError: true,
-        errorCode: policeDataError?.graphQLErrors[0].extensions?.code,
+        errorCode: policeDataError?.graphQLErrors[0].extensions.code as string,
```
eirikurn, line 155: Possible regression. Is `extensions` always an object?

libs/clients/middlewares/src/lib/e2e/enhancedFetch.spec.ts
```diff
@@ -1,3 +1,4 @@
+import 'node-fetch'
```
eirikurn, line 1: Why remove test-cert.pfx? The e2e test depends on it.

libs/infra-nest-server/src/lib/buildOpenApi.ts
```diff
@@ -33,7 +33,7 @@ export const buildOpenApi = async ({
     // https://github.com/configcat/common-js/issues/36
     // TODO: Remove this when it's been fixed.
     process.exit(0)
-  } catch (e) {
+  } catch (e: any) {
```
eirikurn, line 36: Should use `Error` instead?

libs/infra-next-server/src/lib/bootstrap.ts
```diff
@@ -30,7 +30,7 @@ type BootstrapOptions = {
   /**
    * Proxy configuration. Ignored in production (according to NODE_ENV).
    */
-  proxyConfig?: { [context: string]: Config }
+  proxyConfig?: { [context: string]: any }
```
eirikurn, line 33: Doesn't the new version of http-proxy-middleware return a new type for this?

libs/infra-next-server/src/lib/bootstrap.ts
```diff
@@ -84,7 +84,7 @@ export const bootstrap = async (options: BootstrapOptions) => {
     readyPromise,
     options.externalEndpointDependencies,
   )
-  expressApp.all('*', (req, res) => handle(req, res))
+  expressApp.all('*', (req: any, res: any) => handle(req, res))
```
eirikurn, line 87: We should have better express types 😬

libs/infra-next-server/src/lib/proxy.ts
```diff
@@ -1,17 +1,13 @@
...
-  proxyConfig: { [context: string]: Config } | undefined,
+  proxyConfig: { [context: string]: any } | undefined,
   dev: boolean,
 ) => {
   if (!proxyConfig || !dev) {
     return
   }
-  const { default: createProxyMiddleware } = await import(
-    'http-proxy-middleware'
-  )
```
eirikurn, line 14: The idea here was to code-split this in production builds to not evaluate unused code on boot.

libs/shared/babel/src/exportFinder/exportFinder.ts
```diff
@@ -133,7 +133,7 @@ function getExportDeclarations(
 }
 
 const getExports = (modulePath: string): ExportDefinition[] => {
-  const ast = parse(readFileSync(modulePath, 'utf-8'), {
+  const ast: any = parse(readFileSync(modulePath, 'utf-8'), {
```
eirikurn, line 136: Should not be needed. Might be an old babel duplication issue.

libs/shared/mocking/src/graphql/createResolvers.ts
```diff
@@ -13,6 +13,7 @@ export type ResolverMap = Record<
   | Record<string, ((...args: any) => any) | undefined>
   | GraphQLScalarType
   | undefined
+  | any
```
eirikurn, line 16: I don't like that we're adding `any` types across the codebase. I understand why, but maybe we can take a minute and find better fixes for some of them.

package.json
```diff
@@ -4,8 +4,8 @@
   "license": "MIT",
   "engines": {
     "npm": "please-use-yarn",
-    "node": "^14.17",
-    "yarn": "^1.22"
+    "node": "18.8.0",
+    "yarn": "3.2"
```
eirikurn, line 8: 🙌🎉

package.json
```diff
@@ -185,10 +186,10 @@
     "is-url": "1.2.4",
     "isomorphic-fetch": "3.0.0",
     "isomorphic-unfetch": "3.0.0",
+    "jest-environment-jsdom": "28.1.3",
```
eirikurn, line 189: devDependencies?

package.json
```diff
@@ -268,41 +270,40 @@
...
-    "@graphql-codegen/typescript-operations": "1.17.15",
-    "@graphql-codegen/typescript-react-apollo": "1.17.7",
-    "@graphql-codegen/typescript-resolvers": "1.17.9",
+    "@graphql-codegen/cli": "2.12.0",
+    "@graphql-codegen/fragment-matcher": "3.3.1",
+    "@graphql-codegen/schema-ast": "2.5.1",
+    "@graphql-codegen/typescript": "2.7.3",
+    "@graphql-codegen/typescript-operations": "2.5.3",
+    "@graphql-codegen/typescript-react-apollo": "3.3.3",
+    "@graphql-codegen/typescript-resolvers": "2.7.3",
```
eirikurn, line 306: These are used by our schema generation script. No regressions? 🙈

Thankfully TypeScript should catch most of it.

apps/api/src/app/app.module.ts
```diff
@@ -87,14 +89,17 @@ const autoSchemaFile = environment.production
...
       },
       plugins: [
-        responseCachePlugin({
-          shouldReadFromCache: ({ request: { http } }) => {
-            const bypassCacheKey = http?.headers.get('bypass-cache-key')
-            return bypassCacheKey !== process.env.BYPASS_CACHE_KEY
-          },
-        }),
+        // This was causing problems since graphql upgrade, gives us issues like:
+        // Error: overallCachePolicy.policyIfCacheable is not a function
```
eirikurn, line 93: Sounds like a node_modules resolution/deduplication issue. 

Not sure how important this cache is but would feel better if we fix it :)

eirikurn (commented):
Great job!

apps/auth-admin-web/specs/index.spec.tsx
```diff
@@ -1,6 +1,6 @@
 import React from 'react'
 import { render } from '@testing-library/react'
-
+import 'whatwg-fetch'
```
oddsson, line 3: Is this accidental?

apps/judicial-system/web/src/routes/Prosecutor/components/PoliceCaseFiles/PoliceCaseFiles.tsx
oddsson, line 131:
```suggestion
```

apps/judicial-system/web/src/routes/Prosecutor/components/PoliceCaseFiles/PoliceCaseFiles.tsx
oddsson, line 155:
```suggestion
        errorCode: policeDataError?.graphQLErrors[0].extensions?.code as string,
```

Extensions is of type `readonly extensions: { [key: string]: any } | undefined;`. It can be `undefined` so I'd think we need a `?` after `extensions`

apps/judicial-system/backend/src/app/modules/case/case.service.ts
gudjong: Agree, there is no need to wait.

gudjong (changes requested):
One thing

apps/judicial-system/web/src/routes/Prosecutor/components/PoliceCaseFiles/PoliceCaseFiles.tsx
Arnarkari93, line 155: I think adding a few `?.` should do the trick here. `extensions` if defined is always an object/map https://spec.graphql.org/draft/#sel-FAPHRPDZAADbBAzwM
```suggestion
        errorCode: policeDataError?.graphQLErrors[0]?.extensions?.code as string,
```

apps/judicial-system/web/src/routes/Shared/SignedVerdictOverview/Components/ModifyDatesModal/ModifyDatesModal.spec.ts
```diff
@@ -5,7 +5,7 @@ import { createCaseModifiedExplanation } from './ModifyDatesModal'
 describe('createCaseModifiedExplanation', () => {
   const formatMessage = createIntl({ locale: 'is', onError: jest.fn() })
     .formatMessage
-  beforeAll(() => jest.useFakeTimers('modern'))
+  beforeAll(() => jest.useFakeTimers())
```
Arnarkari93, line 8: It seems like the new way of setting the time in jest tests is setting the `now` field in the `fakeTimersConfig` 
```suggestion
  beforeAll(() => jest.useFakeTimers({ now: new Date('2022-06-13T13:37:00Z') }))
```
https://jestjs.io/docs/jest-object#fake-timers

apps/judicial-system/web/src/utils/validate.spec.ts
```diff
@@ -185,7 +185,7 @@ describe('Validate phonenumber format', () => {
 
 describe('Validate court case number', () => {
   beforeAll(() => {
-    jest.useFakeTimers('modern').setSystemTime(new Date('2020-01-01'))
+    jest.useFakeTimers().setSystemTime(new Date('2020-01-01'))
```
Arnarkari93, line 188:
```suggestion
    jest.useFakeTimers({ now: new Date('2020-01-01') })
```

Arnarkari93 (changes requested):

obmagnusson (approved):

.gitignore
peturgq: `yarn/releases` contains the executable for yarn, the other ones I don't know so much, but I took this from here:
https://yarnpkg.com/getting-started/qa#which-files-should-be-gitignored

apps/api/src/app/graphql.middleware.ts
peturgq: Absolutely accidental, good catch!

package.json
peturgq: We are generating schemas successfully but I'm not sure how to test them specifically beyond that, would love your input in that area.

apps/auth-admin-web/specs/index.spec.tsx
peturgq: No, seemed like the only way to get fetching to work inside the test, on fetch the test is using relative paths where the default global fetch API introduced in v18 was failing, I tried `node-fetch` too but this was also failing.

peturgq:
@gudjong 
> One thing 
> Agree, there is no need to wait.

If we did not use await there we hit the following error:
'Can not use queue until nest application bootstrapping has finished'

gudjong:
> @gudjong
> 
> > One thing
> > Agree, there is no need to wait.
> 
> If we did not use await there we hit the following error: 'Can not use queue until nest application bootstrapping has finished'

That is surprising, but I will accept the change. This code has change and we may need to revisit this when you merge with main.

gudjong (approved):

apps/air-discount-scheme/web/i18n/withLocale.tsx
peturgq: I don't think this ever worked, its broken on main f.ex, `lib/utils` simply does not exist in `next-server` in this version of next.
I just took the liberty of fixing it here.

apps/judicial-system/web/src/routes/Shared/SignedVerdictOverview/Components/ModifyDatesModal/ModifyDatesModal.spec.ts
peturgq: Are we ok with using something arbitrary like '2022-06-13T13:37:00Z' here?

Arnarkari93: yes, for this test it is fine.

peturgq: Seeing as these tests are succeeding, I think we should let it slip for now, this PR is not supposed to go into changing the functionality of tests.

My assumption was that the faketimers now use 'modern' by default, I will let this one be while I fixed the other one where a date was being used.

apps/api/src/app/app.module.ts
peturgq: We tried fixing this, but weren't successful - Any ideas on what this cache actually does?

Arnarkari93 (approved):

peturgq:
We're already seeing failures now after merging latest main, here.. We'll fix this and merge this in ASAP

PR #8401, opened 2022-09-12
feat(j-s): Indictments - Court record route

oddsson:
# Indictments - Court record route

[Asana](https://app.asana.com/0/1199153462262248/1202861834321045)

## What

Implement the court record screen in indictments

## Screenshots / Gifs

screencapture-localhost-4200-domur-akaera-thingbok-f61f235f-49f8-486f-af47-642758d2e900-2022-09-12-14_46_55

## Checklist:


apps/judicial-system/web/src/routes/Shared/Cases/Cases.tsx
```diff
@@ -105,26 +107,20 @@ export const Cases: React.FC = () => {
...
-            : !c.rulingDate
-        }),
-      )
+      const casesByState = partition(casesWithoutDeleted, (c: Case) => {
+        return isIndictmentCase(c.type) && c.state === CaseState.ACCEPTED
+          ? false
+          : isPrisonAdminUser || isPrisonUser
+          ? !c.isValidToDateInThePast && c.rulingDate
+          : !c.rulingDate
+      })
```
gudjong, line 120: Consider not returning indictments to the client for STAFF (prison and prison admin users).

gudjong (changes requested):

Arnarkari93 (approved):
LGTM

apps/judicial-system/web/src/routes/Shared/Cases/Cases.tsx
oddsson: I looked into it and we're already doing this, only custodies, travel bans and admission to facilities are being sent to STAFF users.

Heres the code

```
function getStaffCasesQueryFilter(
  institutionType?: InstitutionType,
): WhereOptions {
  return institutionType === InstitutionType.PRISON_ADMIN
    ? {
        [Op.and]: [
          { isArchived: false },
          { state: CaseState.ACCEPTED },
          {
            type: [
              CaseType.ADMISSION_TO_FACILITY,
              CaseType.CUSTODY,
              CaseType.TRAVEL_BAN,
            ],
          },
        ],
      }
    : {
        [Op.and]: [
          { isArchived: false },
          { state: CaseState.ACCEPTED },
          { type: [CaseType.CUSTODY, CaseType.ADMISSION_TO_FACILITY] },
          {
            decision: [
              CaseDecision.ACCEPTING,
              CaseDecision.ACCEPTING_PARTIALLY,
            ],
          },
        ],
      }
}
```

| Court | Staff |
|-------|-------|
|Screenshot 2022-09-13 at 15 04 08|Screenshot 2022-09-13 at 15 04 38|

gudjong: This is correct. However, there is also logic that blocks access to individual cases. This logic is used in guards. The method `getBlockedTypes` in the file `case.filters.ts` needs to change to reflect this. Line 74 reads `blockedTypes.push(...investigationCases)`. Adding the line `blockedTypes.push(...indictmentCases)` immediately before line 74 will do the trick. We should consider modifying the logic to allow access to case types rather than block access. That way, the logic will hold even when adding new case types. The same applies to other blocking logic in this file.

oddsson: Nice! I agree we should flip the logic to grant access instead of restricting access. I'll create a task in Asana.

Instead of having two `blockedTypes.push` I added `blockedTypes.push(...indictmentCases, ...investigationCases)`.

gudjong (approved):

PR #8402, opened 2022-09-12
feat(auth-api): Disable delegations for deceased individuals

snaerseljan:
[Task](https://www.notion.so/Disable-delegations-for-deceased-individuals-6d91ef169f9a455da1f3cf2a1648b859)

## What

- Add a feature to disable delegations for deceased individuals in `delegationsService.findAllIncoming` 
- Update audit service to support system-specific events.

## Why

- We do not want to display delegations for deceased people to the user when loading the delegation picker.

## Checklist:


libs/auth-api-lib/src/lib/services/delegations.service.ts
```diff
@@ -352,23 +363,126 @@ export class DelegationsService {
...
     }
     if (
       (!client || client.supportsPersonalRepresentatives) &&
       (!hasDelegationTypeFilter ||
         delegationTypes?.includes(DelegationType.PersonalRepresentative))
     ) {
-      delegationPromises.push(this.findAllRepresentedPersonsIncoming(user))
+      needVerifyDelegationPromises.push(
+        this.findAllRepresentedPersonsIncoming(user),
+      )
```
eirikurn, line 375: Afaik, representedPersons delegations are not stored in the delegation table, so we need different logic to "disable" them. 

I see that the invalidateDelegations function skips over those delegations (without `id`), and they get removed from the response, which is maybe a fine MVP. But there is still the matter of the audit message which would sound repeatedly if there is a deceased represented person.

libs/auth-api-lib/src/lib/services/delegations.service.ts
```diff
@@ -352,23 +363,126 @@ export class DelegationsService {
...
+    if (delegations.length === 0) {
+      return {
+        aliveDelegations: [],
+        deceasedDelegations: [],
+      }
+    }
+
+    const delegationsPromises = delegations.map(({ fromNationalId }) =>
+      this.nationalRegistryClient.getIndividual(fromNationalId),
+    )
```
eirikurn, line 461: A few suggestions:

1. Should not try and look up company ids in national registry.
2. If an error is thrown during lookup, we may want to swallow it and assume the delegation is alive.

apps/services/auth/public-api/src/app/modules/delegations/actorDelegations.controller.spec.ts
```diff
@@ -108,8 +118,33 @@ describe('ActorDelegationsController', () => {
     })
 
     describe('GET /actor/delegations', () => {
+      let nationalRegistryApiSpy: jest.SpyInstance
       const path = '/v1/actor/delegations'
       const query = '?direction=incoming'
+      const deceasedNationalIds = generateItemFromCnt(3, getFakeNationalId)
```
eirikurn, line 124: Can use existing utility function from `@island.is/testing/fixtures` to generate better fake national ids.

apps/services/auth/api/src/app/modules/delegations/delegations.controller.spec.ts
```diff
@@ -32,6 +38,10 @@ import {
   getScopePermission,
   personalRepresentativeType,
 } from '../../../../test/stubs/personalRepresentativeStubs'
+import { isDefined } from '@island.is/shared/utils'
+
+export const generateItemFromCnt = <T>(cnt: number, objFn: () => T) =>
+  Array.from(Array(cnt).keys()).map(() => objFn())
```
eirikurn, line 44: Maybe use `lodash/times`?

apps/services/auth/api/src/app/modules/delegations/delegations.controller.spec.ts
```diff
@@ -286,6 +354,30 @@ describe('DelegationsController', () => {
...
+
+              const allDelegationsScopes = delegationsModels
+                ?.map((delegation) => delegation?.scopes)
+                .filter(isDefined)
+                .flat()
+
+              // Assert
+              allDelegationsScopes?.forEach((scope) => {
+                expect(scope.validTo).toEqual(subDays(today, 1))
+              })
```
eirikurn, line 379: Are there any delegation scopes in this array? Since this is testing deceased represented people and not custom delegations.

apps/services/auth/api/src/app/modules/delegations/delegations.controller.spec.ts
```diff
@@ -191,6 +212,34 @@ describe('DelegationsController', () => {
...
+                createDelegation({
+                  fromNationalId: deceasedNationalIds[i],
+                  toNationalId: user.nationalId,
+                  scopes: [defaultScopes.testUserHasAccess.name],
+                  today,
+                }),
+                {
+                  include: [{ model: DelegationScope, as: 'delegationScopes' }],
+                },
+              )
```
eirikurn, line 240: I don't think there is a delegation in personal representative relationships.

eirikurn (changes requested):
I really like how you're cleaning up code as you go. 🙌

Commented on some small and large things to consider for this tricky task.

apps/services/auth/api/src/app/modules/delegations/delegations.controller.ts
```diff
@@ -14,10 +14,12 @@ import {
   ScopesGuard,
 } from '@island.is/auth-nest-tools'
 import type { User } from '@island.is/auth-nest-tools'
+import { Audit } from '@island.is/nest/audit'
 
 @UseGuards(IdsUserGuard, ScopesGuard)
 @ApiTags('delegations')
 @Controller('delegations')
+@Audit({ namespace: '@island.is/api/delegations' })
```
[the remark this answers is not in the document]
snaerseljan: Hmm okey maybe I am misunderstanding.
However I put this in delegations.controller since one of its methods is using
```
return this.delegationsService.findAllIncoming(user)
```
which uses it like so
```
      this.auditService.auditSystem({
        action: 'deceasedDelegation',
        resources: deceasedDelegations.map(({ id }) => id).filter(isDefined),
      })
```
The logic I wrote in the `auditSystem` will error if namespace is missing, i.e. in this.checkNamespace()
```

```
  private formatSystemMessage({
    namespace = this.defaultNamespace,
    action,
    resources,
    meta,
  }: AuditSystemMessage) {
    this.checkNameSpace(namespace)
    const namespacePostfix = `${namespace}/system`
    const message = {
      namespace: namespacePostfix,
      action: `${namespacePostfix}#${action}`,
      resources: isString(resources) ? [resources] : resources,
      meta,
      appVersion: process.env.APP_VERSION,
    }

    return this.useDevLogger ? { message: 'Audit record', ...message } : message
  }
```

Would you rather want me to pass in a namespace here?

```
    this.auditService.auditSystem({
        action: 'deceasedDelegation',
        resources: deceasedDelegations.map(({ id }) => id).filter(isDefined),
        namespace: 'Some namespace'
      })
```

File: apps/services/auth/public-api/src/app/modules/delegations/actorDelegations.controller.spec.ts [hunk 8]
```diff
@@ -108,8 +118,33 @@ describe('ActorDelegationsController', () => {
...
+
+      beforeAll(async () => {
+        const persons = generateItemFromCnt(10, getFakePerson)
```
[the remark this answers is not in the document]
snaerseljan: No this is correct, but I get your concern

apps/services/auth/public-api/test/utils.ts
```diff
@@ -39,3 +40,19 @@ export const sortDelegations = (delegations: DelegationDTO[]) => {
...
+export const getFakeNationalId = () =>
+  faker.helpers.replaceSymbolWithNumber('##########')
+
+export const getFakeName = () =>
+  faker.fake('{{name.firstName}} {{name.lastName}}')
+
+export const getFakePerson = (): NameIdTuple => [
+  getFakeName(),
+  getFakeNationalId(),
+]
```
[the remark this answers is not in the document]
snaerseljan: These methods are copy/paste from  [genericStubs](https://github.com/island-is/island.is/blob/c16cd3aa40efc06ef05716eb162cc43ef23b2990/apps/services/auth/api/test/stubs/genericStubs.ts)

I wanted to reuse these method, but we discussed that it would be better in some cases just to copy paste between projects.

Where are these fixture methods?

libs/shared/utils/src/lib/date.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * Gets yesterday's date from current date, in UTC format.
+ */
+export const getYesterday = (date: Date) => {
+  date.setDate(date.getDate() - 1)
+
+  return date
+}
```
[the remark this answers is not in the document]
snaerseljan: Good point will update

apps/services/auth/api/src/app/modules/delegations/delegations.controller.spec.ts
```diff
@@ -286,6 +354,30 @@ describe('DelegationsController', () => {
...
+                      as: 'delegationScopes',
+                    },
+                  ],
+                })
+              ).map((delegation) => delegation.toDTO())
+
+              const allDelegationsScopes = delegationsModels
+                ?.map((delegation) => delegation?.scopes)
+                .filter(isDefined)
+                .flat()
```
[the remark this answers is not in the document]
snaerseljan: Yes the magic is the return value, i.e. `value is TValue` then TS infers the correct type :)

apps/services/auth/api/src/app/modules/delegations/delegations.controller.spec.ts
snaerseljan: Ohh nice. Will use

libs/auth-api-lib/src/lib/services/delegations.service.ts
```diff
@@ -352,23 +363,126 @@ export class DelegationsService {
...
+   */
+  private async invalidateDelegations(
+    user: User,
+    deceasedDelegations: DelegationDTO[],
+  ) {
+    const delegations = await Promise.all(
+      // We need to refetch the delegation to make sure we get all delegation.scopes
+      deceasedDelegations.map(({ id }) =>
+        id ? this.findById(user, id, true) : undefined,
+      ),
```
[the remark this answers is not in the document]
snaerseljan: I have now updated the code to delete the delegation instead of updating the validTo delegationScope field

apps/services/auth/api/src/app/modules/delegations/delegations.controller.spec.ts
snaerseljan: This has now been updated to delete delegation instead.

apps/services/auth/public-api/test/utils.ts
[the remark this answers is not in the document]
snaerseljan: I updated these methods to use methods from fixture

libs/nest/audit/src/lib/audit.service.ts
```diff
@@ -119,6 +126,25 @@ export class AuditService {
...
   }
 
+  private formatSystemMessage({
+    namespace = this.defaultNamespace,
+    action,
+    resources,
+    meta,
+  }: AuditSystemMessage) {
+    this.checkNameSpace(namespace)
+    const namespacePostfix = `${namespace}/system`
```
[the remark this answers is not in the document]
saevarma: Discussed offline: Do not do the automatic naming convention, but instead add a property `system: true`

libs/auth-api-lib/src/lib/services/delegations.service.ts
snaerseljan: I divided the logic up to make things more clear.
now these two methods, i.e. `findAllRepresentedPersonsIncoming` and `findAllValidCustomIncoming` encapsulate the logic. That is they handle deleting the appropriate models

apps/services/auth/api/src/app/modules/delegations/delegations.controller.spec.ts
snaerseljan: this has been removed

libs/auth-api-lib/src/lib/services/delegations.service.ts
snaerseljan: Great call out, here is the updated version
```
 /**
   * Divides delegations into alive and deceased delegations
   * - Makes calls for every delegation to NationalRegistry to check if the person exists.
   * - Divides the delegations into alive and deceased delegations, based on
   *   1. All companies will be divided into alive delegations.
   *   2. If the person exists in NationalRegistry, then the delegation is alive.
   */
  private async getLiveStatusFromDelegations(
    delegations: DelegationDTO[],
  ): Promise<{
    aliveDelegations: DelegationDTO[]
    deceasedDelegations: DelegationDTO[]
  }> {
    if (delegations.length === 0) {
      return {
        aliveDelegations: [],
        deceasedDelegations: [],
      }
    }

    const delegationsPromises = delegations
      // Filter out companies, since we do not need to make a natinonal registry call for them.
      .filter(({ fromNationalId }) => !kennitala.isCompany(fromNationalId))
      .map(({ fromNationalId }) =>
        this.nationalRegistryClient.getIndividual(fromNationalId),
      )

    try {
      // Check if delegations is linked to a person, i.e. not deceased
      const persons = await Promise.all(delegationsPromises)
      const personsValues = persons.filter(isDefined)

      // Divide delegations into alive or deceased delegations.
      const [aliveDelegations, deceasedDelegations] = partition(
        delegations,
        ({ fromNationalId }) =>
          // All companies will be divided into aliveDelegations
          kennitala.isCompany(fromNationalId) ||
          // Make sure we can match the person to the delegation, i.e. not deceased
          personsValues.find((person) => person?.nationalId === fromNationalId),
      )

      return {
        aliveDelegations,
        deceasedDelegations,
      }
    } catch (error) {
      this.logger.error(
        `Error getting live status from delegations. Delegations: ${delegations.map(
          (d) => d.id,
        )}`,
        error,
      )

      // We do not want to fail the whole request if we cannot get the live status from delegations.
      // Therefore we return all delegations as alive delegations.
      return {
        aliveDelegations: delegations,
        deceasedDelegations: [],
      }
    }
  }
```

2.

apps/services/auth/api/src/app/v1/delegations/delegations.controller.spec.ts
```diff
@@ -204,10 +228,18 @@ describe('DelegationsController', () => {
...
                 const user = nationalRegistryUsers.find(
-                  (u) => u.nationalId === id,
+                  (u) =>
+                    u?.nationalId === id &&
+                    // Make sure we don't return a user that has been marked as deceased
+                    !deceasedNationalIds.includes(u?.nationalId),
                 )
-                return user ? Promise.resolve(user) : Promise.reject()
+
+                return user ? Promise.resolve(user) : Promise.reject(null)
```
eirikurn, line 242: Should resolve with null. This whole function would probably be simpler as an `async` function.

apps/services/auth/api/src/app/v1/delegations/delegations.controller.spec.ts
```diff
@@ -284,6 +318,24 @@ describe('DelegationsController', () => {
                 ),
               ).toBeTruthy()
             })
+
+            it('should have deleted prModels and prRightModels for deceased persons', async () => {
```
eirikurn, line 322: Let's discuss if we want to delete personal representative data for individuals that we assume are "deceased". Might want to just "hide" these delegations without deleting.

On one hand, we do not "own" these rows. Réttindagæslan should do this. But we don't know if they will, and if these will collect up and cause weird behaviour later on.

/cc @hrefnalind

eirikurn (changes requested):

apps/services/auth/api/src/app/v1/delegations/delegations.controller.spec.ts
hrefnalind: Yes I think it makes more sens to hide them

libs/nest/audit/src/lib/audit.service.ts
```diff
@@ -97,23 +100,25 @@ export class AuditService {
...
       client: this.getClients(auth),
       action: `${namespace}#${action}`,
-      resources:
-        resources && (typeof resources === 'string' ? [resources] : resources),
+      resources: isString(resources) ? [resources] : resources,
       meta,
       ip: auth.ip,
       userAgent: auth.userAgent,
       appVersion: process.env.APP_VERSION,
+      system: system ?? false,
```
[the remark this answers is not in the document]
snaerseljan: @eirikurn do you mean to skip it entirely so maybe instead do 

```
appVersion: process.env.APP_VERSION,
...(system && { system }),
```

libs/auth-api-lib/src/lib/delegations/delegations.service.ts
```diff
@@ -519,27 +612,68 @@ export class DelegationsService {
...
+  ) {
+    // Delete all personal representatives and their rights
+    const deletePromises = personalRepresentatives
+      .map(({ id }) => (id ? this.prService.delete(id) : undefined))
+      .filter(isDefined)
+
+    await Promise.all(deletePromises)
+
+    this.auditService.audit({
+      auth: user,
```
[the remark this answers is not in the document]
snaerseljan: @eirikurn I updated the audit service types and logic a bit. Now you do not need to pass the user for system logs nor system in other logs. Let me know what you think.

apps/services/auth/api/src/app/v1/delegations/delegations.controller.spec.ts
snaerseljan: good point. Old code as well

libs/auth-api-lib/src/lib/delegations/delegations.service.ts
```diff
@@ -519,27 +612,68 @@ export class DelegationsService {
...
   }
 
+  private async deletePersonalRepresentatives(
+    personalRepresentatives: PersonalRepresentativeDTO[],
+    user: User,
+  ) {
+    // Delete all personal representatives and their rights
+    const deletePromises = personalRepresentatives
+      .map(({ id }) => (id ? this.prService.delete(id) : undefined))
+      .filter(isDefined)
```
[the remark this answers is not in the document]
snaerseljan: Yes weird. But there must be a reason. Right?

libs/nest/audit/src/lib/audit.service.ts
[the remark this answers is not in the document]
eirikurn: Fyi, your suggestion can be useful in some specific scenarios, but when dealing with JSON (responses, requests, audit), `undefined` values have the useful quality of not being serialised at all. So my suggested change works and is simpler. 

Not relevant with your refactor.

libs/auth-api-lib/src/lib/delegations/delegations.service.ts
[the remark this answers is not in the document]
eirikurn: Love it!

libs/auth-api-lib/src/lib/delegations/delegations.service.ts
```diff
@@ -616,6 +743,32 @@ export class DelegationsService {
...
+      const deletePromises = deceasedDelegations
+        .map(({ id }) =>
+          id ? this.delete(user, id, DelegationDirection.INCOMING) : undefined,
+        )
+        .filter(isDefined)
+
+      await Promise.all(deletePromises)
+
+      this.auditService.audit({
+        auth: user,
```
eirikurn, line 764: Can you still pass both `auth` and `system: true`? 😮

libs/auth-api-lib/src/lib/delegations/delegations.service.ts
```diff
@@ -538,27 +628,66 @@ export class DelegationsService {
         provider: DelegationProvider.PersonalRepresentativeRegistry,
       })
 
-      const rp = await this.prService.getByPersonalRepresentative(
+      const personalRepresentatives = await this.prService.getByPersonalRepresentative(
```
eirikurn, line 631: We should filter inactive personal representatives so we're not constantly rechecking them. This should be inside `getByPersonalRepresentative`, maybe a flag to skip or include inactive personal representatives.

libs/auth-api-lib/src/lib/delegations/delegations.service.ts
```diff
@@ -538,27 +628,66 @@ export class DelegationsService {
...
 
-      const resultPromises = rp.map(async (representative) =>
-        this.nationalRegistryClient
-          .getIndividual(representative.nationalIdRepresentedPerson)
-          .then((person) =>
-            toDelegationDTO(person?.name ?? 'Óþekkt nafn', representative),
+      const personPromises = personalRepresentatives.map(
+        ({ nationalIdRepresentedPerson }) =>
+          this.nationalRegistryClient.getIndividual(
+            nationalIdRepresentedPerson,
```
eirikurn, line 639: One more (big) thing... The IDS needs to be as resilient as possible from outside dependencies like Þjóðskrá. If any of these `getIndividual` calls throw an Error (eg network issue, 500, some xroad error), we need to continue without crashing the request and try to keep as much functionality as possible.

Ideally we assume the person is still around and continue serving the delegation until proven otherwise.

It may complicate these functions some. Not sure what's the smartest way to tell apart "null" and the Error. Maybe just `.catch(error => return error)`, and work with an array of `Individual | null | Error`.

Sorry about mentioning this soo late ... I remember spotting this way earlier but somehow forgot to add it to my earlier reviews. 😭

eirikurn (commented):
Aaaaaalmost there!

libs/auth-api-lib/src/lib/delegations/delegations.service.ts
snaerseljan: Ahh forgot never on property field, This does the trick
image

libs/auth-api-lib/src/lib/delegations/delegations.service.ts
snaerseljan: good point

libs/auth-api-lib/src/lib/delegations/delegations.service.ts
snaerseljan: No problemo. I have update the code. Let's hope it does the trick.
Now we swallow the error and return appropriate items although the api call to Þjóðskrá errors. The fromName field is then populated as `"Óþekkt nafn"`

snaerseljan: These three commits are fixed to the newest comments.

https://github.com/island-is/island.is/pull/8402/commits/c7e6a20bb06a9a9bc26941f27c2d6aae0e7bcee1

https://github.com/island-is/island.is/pull/8402/commits/4f5276f35aab9cf9d9096a2a2d8ae4112c1325a5

https://github.com/island-is/island.is/pull/8402/commits/c7ead4539c7b95078d58f6461780cdc122022774

and the last commit is just a small refactor.

https://github.com/island-is/island.is/pull/8402/commits/33f46498ae7c3b933629d49132331a3956a45677

apps/services/auth/api/src/app/v1/delegations/delegations.controller.spec.ts
```diff
@@ -341,6 +382,18 @@ describe('DelegationsController', () => {
                 )
               })
             })
+
+            it('should return delegation if national registry api getIndividual throws and error', async () => {
```
eirikurn: What is this testing?

libs/auth-api-lib/src/lib/delegations/delegations.service.ts
```diff
@@ -411,6 +411,27 @@ export class DelegationsService {
     )
   }
 
+  private async handlerGetIndividualError(error: null | Error) {
```
eirikurn, line 426: Unneccessarily `async`?

libs/auth-api-lib/src/lib/delegations/delegations.service.ts
```diff
@@ -435,26 +456,49 @@ export class DelegationsService {
...
+        ({ fromNationalId }, index) =>
           // All companies will be divided into aliveDelegations
           kennitala.isCompany(fromNationalId) ||
+          // Pass persons that are of Error instance, since we want do return the delegation.
+          persons[index] instanceof Error ||
           // Make sure we can match the person to the delegation, i.e. not deceased
-          personsValues.find((person) => person?.nationalId === fromNationalId),
+          personsValuesNoError.some(
+            (person) => person?.nationalId === fromNationalId,
+          ),
```
eirikurn, line 494: Can use `persons[index]` here as well?

libs/auth-api-lib/src/lib/delegations/delegations.service.ts
```diff
@@ -435,26 +456,49 @@ export class DelegationsService {
...
+      // If Þjóðskrá throws and error for some fromNationalId, then we still want to return the delegation.
+      // however we change the delegation.fromName to unknown name.
+      const modifiedAliveDelegations = aliveDelegations
+        .map((aliveDelegation) => {
+          const person = this.getPersonByNationalId(
+            personsValuesNoError,
+            aliveDelegation.fromNationalId,
+          )
+
+          aliveDelegation.fromName = person?.name ?? 'Óþekkt nafn'
```
eirikurn: `aliveDelegation.fromName = person?.name ?? aliveDelegation.fromName ?? 'Óþekkt nafn'`?

libs/auth-api-lib/src/lib/delegations/delegations.service.ts
```diff
@@ -435,26 +456,49 @@ export class DelegationsService {
...
-          personsValues.find((person) => person?.nationalId === fromNationalId),
+          personsValuesNoError.some(
+            (person) => person?.nationalId === fromNationalId,
+          ),
       )
 
+      // If Þjóðskrá throws and error for some fromNationalId, then we still want to return the delegation.
+      // however we change the delegation.fromName to unknown name.
+      const modifiedAliveDelegations = aliveDelegations
+        .map((aliveDelegation) => {
```
eirikurn, line 498: Consider `aliveDelegations.forEach` to skip the extra array since we're mutating the objects anyways?

libs/auth-api-lib/src/lib/delegations/delegations.service.ts
```diff
@@ -435,26 +456,49 @@ export class DelegationsService {
...
+          const person = this.getPersonByNationalId(
+            personsValuesNoError,
+            aliveDelegation.fromNationalId,
+          )
+
+          aliveDelegation.fromName = person?.name ?? 'Óþekkt nafn'
+
+          return aliveDelegation
+        })
+        .filter(isDefined)
```
eirikurn, line 508: Neccessary?

libs/auth-api-lib/src/lib/delegations/delegations.service.ts
```diff
@@ -636,22 +680,25 @@ export class DelegationsService {
...
         personalRepresentatives,
-        ({ nationalIdRepresentedPerson }) =>
-          findPersonByNationalId(nationalIdRepresentedPerson),
+        ({ nationalIdRepresentedPerson }, index) =>
+          // Pass persons that are of Error instance, since we want do return the personal representative.
+          persons[index] instanceof Error ||
+          // Make sure we can match the person to the personal representatives, i.e. not deceased
+          personsValuesNoError.some(
+            (person) => person?.nationalId === nationalIdRepresentedPerson,
+          ),
```
eirikurn, line 711: Same?

eirikurn (approved):
Nice! Just nitpicks at this point. 💩

Approving so we can get this quickly to main :)

apps/services/auth/api/src/app/v1/delegations/delegations.controller.spec.ts
snaerseljan: This is making sure that delegations get return although the national registry throws an error.

libs/auth-api-lib/src/lib/delegations/delegations.service.ts
snaerseljan: Ohh yeah jesus, was apparently too fixated on the writing the `partitionWithIndex` 😅

libs/auth-api-lib/src/lib/delegations/delegations.service.ts
snaerseljan: hmm weird . I am not sure why you where seeing this line because the newest line looks like
```
aliveDelegation.fromName = person?.name ?? aliveDelegation.fromName ?? UNKNOWN_NAME
```

Did you forget to pull or something?

libs/auth-api-lib/src/lib/delegations/delegations.service.ts
snaerseljan: Good point but I rather update the logic inside map to
```
      const modifiedAliveDelegations = aliveDelegations
        .map((aliveDelegation) => {
          const person = this.getPersonByNationalId(
            personsValuesNoError,
            aliveDelegation.fromNationalId,
          )

          return {
            ...aliveDelegation,
            fromName: person?.name ?? aliveDelegation.fromName ?? UNKNOWN_NAME,
          }
        })
        .filter(isDefined)
```

libs/auth-api-lib/src/lib/delegations/delegations.service.ts
snaerseljan: Nice spot

libs/auth-api-lib/migrations/20221012094608-add-personal-representative-inactive.js
```diff
@@ -2,18 +2,21 @@
...
 module.exports = {
   async up(queryInterface, Sequelize) {
-    return Promise.all([
-      queryInterface.addColumn('personal_representative', 'inactive', {
-        type: Sequelize.BOOLEAN,
-        allowNull: false,
-        defaultValue: false,
-      }),
-      queryInterface.addColumn('personal_representative', 'inactiveReason', {
+    await queryInterface.addColumn('personal_representative', 'inactive', {
```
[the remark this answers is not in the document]
snaerseljan: Hmm I don't understand

PR #8408, opened 2022-09-13
fix(service-portal): axle data, timeout, null values

disaerna:
# Service portal - Fixes

## What

* Add axle table and remove from technical data
* Add timeout 
* Remove null values
* Remove name if name should be secret

## Why

Comments from Samgöngustofa

## Screenshots / Gifs

Screenshot 2022-09-13 at 12 56 41

## Checklist:


libs/service-portal/vehicles/src/components/DetailTable/AxleTable.tsx
```diff
@@ -0,0 +1,70 @@
...
+            </T.HeadData>
+            <T.HeadData>
+              <Text variant="medium" fontWeight="semiBold">
+                {formatMessage(messages.axleWheel)}
+              </Text>
+            </T.HeadData>
+          </T.Row>
+        </T.Head>
+        <T.Body>
+          {axles?.map((item: VehiclesAxle, index) => {
```
thordurhhh, line 47: Should this map the `tyreArray` or is it a case of where the axles will always match the tyreArray?

disaerna: Yes axles and tyre size will always be hand in hand. I changed it a little to prevent null errors, put empty string if no value in tyre array so the table will still be rendered correctly.

thordurhhh (approved):
💪  🎉

PR #8421, opened 2022-09-14
fix(service-portal): Calendar display on finance screens

thordurhhh:
## What

Update display for calendars in filter on finance screen.

## Why

Calendar/Input tends to be cut off when there is not too much content on page, and the screen is too small.

## Screenshots 

Is currently:
Screenshot 2022-09-14 at 14 14 43

Becomes with this PR:
Screenshot 2022-09-14 at 14 13 41

## Checklist:


disaerna (changes requested):
Looking at the image in PR description, I'm pretty sure the picker should be inline with the input box, as in no space between and the left/right borders should be flowing from the input and to the picker container.

libs/service-portal/finance/src/components/DocumentScreen/DocumentScreen.tsx
```diff
@@ -162,10 +168,11 @@ const DocumentScreen: FC<Props> = ({
               labelClearAll={formatMessage(m.clearAllFilters)}
               labelOpen={formatMessage(m.openFilter)}
               labelClose={formatMessage(m.closeFilter)}
+              popoverFlip={false}
               filterInput={
                 <FilterInput
                   placeholder={formatMessage(m.searchPlaceholder)}
-                  name="rafraen-skjol-input"
+                  name="finance-document-input"
```
disaerna, line 175: 💯

libs/service-portal/finance/src/components/DocumentScreen/DocumentScreen.tsx
```diff
@@ -162,10 +168,11 @@ const DocumentScreen: FC<Props> = ({
               labelClearAll={formatMessage(m.clearAllFilters)}
               labelOpen={formatMessage(m.openFilter)}
               labelClose={formatMessage(m.closeFilter)}
+              popoverFlip={false}
```
disaerna, line 171: Just curious, what does this do? Prevent the calendar from popping up or down? 🤔

thordurhhh:
> Yes, thank makes sense 🙈

libs/service-portal/finance/src/components/DocumentScreen/DocumentScreen.tsx
thordurhhh: The default state is popping down. So it will prevent it from popping up.

thordurhhh: changed the code comment a bit to reflect more on what this does.

oddsson (approved):

PR #8424, opened 2022-09-14
fix(national-registry): Transgender adults can be parents

kksteini:
From the Einstaklingar endpoint on gender code

```
{
    "kodi": "7",
    "lysing": "Kynsegin/annað fullorðinn"
}
```

## Checklist:


kksteini:
> Good catch! 💯 We need to remove all usage of this function though. As it will display "children" over the age of 18 as spouse.

Oh wow, I didn't consider the return type.
I think it's good to change the check anyway since we don't want the logic precedent regardless of the function's final outcome.

PR #8435, opened 2022-09-15
feat(web): Fiskistofa - calculators and ship search

RunarVestmann:
# Fiskistofa - calculators and ship search

## What

* Add 2 calculator applications (the calculations are performed by external web services which I call via an OpenAPI generated client)
* Add ship search to the sidebar on the Fiskistofa organization pages
* Add ship search page (by using a connected component slice embedded in rich text of an organization subpage)
* Add a high z-index value to Select menu due to it otherwise going behind the table (I had to give some columns z-indexes since they were fixed)

## Why

* To be able to fully move Fiskistofa over to Ísland.is and shut down the old web we need to have the calculators and a way to search for ships.

## Screenshots / Gifs

https://user-images.githubusercontent.com/43557895/191465105-eecf881f-aa9e-4e02-b0ba-bb1e04ee0872.mov

## Checklist:


pshomov (approved):
LGTM(infra files)

libs/clients/fiskistofa/src/lib/fiskistofaClient.service.ts
```diff
@@ -0,0 +1,143 @@
...
+  constructor(private api: FiskistofaApi) {}
+
+  async wrapper(callback: (api: FiskistofaApi) => any) {
+    try {
+      this.api.initialize()
+      return callback(this.api)
+    } catch (error) {
+      if (error instanceof FetchError) {
+        if (error.status === 401) {
+          this.api.initialize(true)
```
siggifv, line 37: It'd be good to see some comments here, why this is needed.

libs/shared/connected/src/lib/fiskistofa/ShipSearch/ShipSearch.tsx
```diff
@@ -0,0 +1,157 @@
...
+        }}
+      />
+      <Box marginTop={3}>
+        <Button
+          disabled={
+            nameInput === nameInputDuringLastSearch && nameInput.length > 0
+          }
+          onClick={() => handleShipSearch(nameInput)}
+        >
+          Leita
```
siggifv, line 94: i18n missing in this component.

libs/shared/connected/src/lib/fiskistofa/SidebarShipSearchInput/SidebarShipSearchInput.tsx
```diff
@@ -0,0 +1,70 @@
...
+  shipSearchHref?: string
+  placeholder?: string
+  label?: string
+}
+
+export const SidebarShipSearchInput = ({
+  shipDetailsHref = '/s/fiskistofa/skip',
+  shipSearchHref = '/s/fiskistofa/skipaleit',
+  placeholder = 'Skipaskrárnúmer eða nafn',
+  label = 'Skoða skip',
```
siggifv, line 16: i18n missing.

libs/shared/connected/src/lib/fiskistofa/ShipSearch/ShipSearch.tsx
RunarVestmann: I've addressed this issue by having component props (which are directly read from Contentful since we are using the Connected Component slice)

libs/shared/connected/src/lib/fiskistofa/SidebarShipSearchInput/SidebarShipSearchInput.tsx
RunarVestmann: I'm using the connected component content type in Contentful and the translations are passed throught as props via the JSON field in Contentful.

libs/clients/fiskistofa/src/lib/fiskistofaClient.service.ts
RunarVestmann: I added a comment
`If we are unauthorized, then we want to re-initialize the api by getting a new access token and then calling the callback again`

apps/web/screens/Organization/Fiskistofa/ShipDetails/ShipDetails.tsx
```diff
@@ -0,0 +1,197 @@
...
+                  {n(tab, `Reiknivél ${tab}`)}
+                </Link>
+              )
+            })}
+          </Inline>
+        </Box>
+      </GridContainer>
+
+      <Box className={styles.container}>
+        {selectedTab === 'aflamarks' && shipNumber && (
```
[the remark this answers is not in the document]
RunarVestmann: TypeScript already provides a fail-safe here for typos

siggifv (approved):

apps/web/screens/Organization/Fiskistofa/ShipDetails/ShipDetails.tsx
```diff
@@ -0,0 +1,199 @@
...
+        </Box>
+      </GridContainer>
+
+      <Box className={styles.container}>
+        <GridContainer>
+          {selectedTab === 'aflamarks' && shipNumber && (
+            <AflamarkCalculator shipNumber={shipNumber} namespace={namespace} />
+          )}
+          {selectedTab === 'deilistofna' && shipNumber && (
+            <DeilistofnaCalculator
```
kksteini, line 137: Just fish food for thought :fish: :fish: 
Question regarding how much we want to keep our code conventions for english.

Aflamark is "Catch quota" in english.
Deilistofn, when related to fish, is "Straddling stock".

kksteini (approved):
Júní files OK :+1:

apps/web/screens/Organization/Fiskistofa/ShipDetails/ShipDetails.tsx
RunarVestmann: Aah ok, thanks for that. I'll change the names :D

libs/api/domains/fiskistofa/src/lib/fiskistofa.resolver.ts
```diff
@@ -0,0 +1,97 @@
...
+const cacheControlDirective = (ms = cacheTime) => `@cacheControl(maxAge: ${ms})`
+
+@Resolver()
+export class FiskistofaResolver {
+  constructor(
+    private readonly fiskistofaClientService: FiskistofaClientService,
+  ) {}
+
+  @Directive(cacheControlDirective())
+  @Query(() => ExtendedShipStatusInformation)
```
eirikurn, line 32: Please follow our [GraphQL naming conventions](https://docs.devland.is/technical-overview/api-design-guide/graphql-naming-conventions) for all GraphQL queries, mutations, inputs and objects.

E.g. this could be `fiskistofaShipInformation(input: ...)` or `fiskistofaShip(shipNumber: ...) { catchQuotaCategories(timePeriod|calendarYear) { ... } }`

libs/api/domains/fiskistofa/src/lib/fiskistofa.resolver.ts
```diff
@@ -0,0 +1,97 @@
...
+  @Mutation(() => ShipStatusInformation)
+  updateShipStatusForCalendarYear(
+    @Args('input') input: UpdateShipStatusForCalendarYearInput,
+  ) {
+    return this.fiskistofaClientService.updateShipStatusForCalendarYear(input)
+  }
+
+  @Directive(cacheControlDirective())
+  @Query(() => [QuotaType])
+  getQuotaTypesForTimePeriod(
```
eirikurn, line 72: Same, `fiskistofaQuotaTypes(timePeriod|calendarYear)`

libs/api/domains/fiskistofa/src/lib/fiskistofa.resolver.ts
```diff
@@ -0,0 +1,97 @@
...
+
+  @Directive(cacheControlDirective())
+  @Query(() => [ShipBasicInfo])
+  getShips(@Args('input') input: GetShipsInput) {
+    return this.fiskistofaClientService.getShips(input)
+  }
+
+  @Directive(cacheControlDirective())
+  @Query(() => SingleShip)
+  getSingleShip(@Args('input') input: GetSingleShipInput) {
```
eirikurn, line 94: `fiskistofaShip(id)`

libs/api/domains/fiskistofa/src/lib/models/quotaStatus.ts
```diff
@@ -0,0 +1,40 @@
+import { Field, ObjectType } from '@nestjs/graphql'
+
+@ObjectType()
```
eirikurn, line 3:
```suggestion
@ObjectType('FiskistofaQuotaStatus`)
```

same for all of the other types.

libs/clients/fiskistofa/src/lib/api.ts
```diff
@@ -0,0 +1,219 @@
...
+} from '../../gen/fetch'
+import { FiskistofaClientConfig } from './fiskistofaClient.config'
+import {
+  mapShipStatusForTimePeriod,
+  mapShipStatusForCalendarYear,
+  mapQuotaType,
+} from './utils'
+import { FetchError } from '@island.is/clients/middlewares'
+
+@Injectable()
```
eirikurn, line 27: By doing the following, other devs can start the API without having all of the required fiskistofa secrets. It'll just crash the queries which depend on fiskistofa. But in production it will stop the release if the secrets are missing.

```suggestion
import { LazyDuringDevScope } from '@island.is/nest/config'

@Injectable({ scope: LazyDuringDevScope })
```

libs/clients/fiskistofa/src/lib/api.ts
```diff
@@ -0,0 +1,219 @@
...
+  }
+
+  async updateShipQuotaStatusForTimePeriod(
+    params: V1StadaskipsSkipnumerFiskveidiarFiskveidiarKvotiBreyttPostRequest,
+  ) {
+    const data = await this.stadaSkipsApi?.v1StadaskipsSkipnumerFiskveidiarFiskveidiarKvotiBreyttPost(
+      params,
+    )
+    return {
+      nextYearCatchQuota: data?.aNaestaArAflamark,
```
eirikurn, line 117: Awesome job wrapping their API in the client library with english interfaces 👏

libs/island-ui/contentful/src/lib/RichTextRC/defaultRenderComponents.tsx
```diff
@@ -30,10 +31,13 @@ const renderConnectedComponent = (slice) => {
...
       if (typeof data === 'object') {
         const { graphqlLink } = data
-
         return <CompanyListConnected graphqlLink={graphqlLink} />
       }
       break
+    case 'Fiskistofa/ShipSearch':
+      return <ShipSearch {...data} />
+    case 'Fiskistofa/ShipSearchSidebarInput':
+      return <SidebarShipSearchInput {...data} />
```
eirikurn, line 40: Feel like all of this should be in `libs/web/richtext` or something, rather than inside the island-ui folder.

Know this isn't your fault but maybe you'll feel an itch to refactor this someday 😆

eirikurn (changes requested):
This looks really cool (nice video in the description). Feel free to reach out if any of my comments are unclear :D

libs/island-ui/core/src/lib/Select/Select.css.ts
```diff
@@ -287,6 +287,7 @@ export const menu = style(
         borderBottomLeftRadius: 8,
         borderBottomRightRadius: 8,
         boxSizing: 'border-box',
+        zIndex: 9999,
```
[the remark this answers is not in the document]
RunarVestmann: I don't think we have something like that at a "global" level. As you can see from the screenshot, the Toast and Navigation ui components just have large z indexes since they are used globally
image

libs/api/domains/fiskistofa/src/lib/fiskistofa.resolver.ts
```diff
@@ -0,0 +1,97 @@
...
+
+  @Directive(cacheControlDirective())
+  @Query(() => ExtendedShipStatusInformation)
+  getShipStatusForTimePeriod(
+    @Args('input') input: GetShipStatusForTimePeriodInput,
+  ) {
+    return this.fiskistofaClientService.getShipStatusForTimePeriod(input)
+  }
+
+  @Mutation(() => ExtendedShipStatusInformationUpdate)
```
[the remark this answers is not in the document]
RunarVestmann: The mutations call a POST endpoint that calculates new values given an input, the endpoint doesn't do anything other than calculate so I wasn't sure what to do here. I ended up using a mutation but honestly this should probably be a query, right?

libs/clients/fiskistofa/src/lib/api.ts
```diff
@@ -0,0 +1,219 @@
...
+          'Content-Type': 'application/json',
+        },
+        body: JSON.stringify({
+          grant_type: 'client_credentials',
+          client_id: this.clientConfig.accessTokenServiceClientId,
+          client_secret: this.clientConfig.accessTokenServiceClientSecret,
+          audience: this.clientConfig.accessTokenServiceAudience,
+        }),
+      },
+    )
```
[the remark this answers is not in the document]
RunarVestmann: To be honest I'm quite new to authentication and the way we are doing GraphQL backend stuff (at least in this project). I thought I could use the createEnhancedFetch autoAuth thing but it required me to provide a scope, so I wasn't sure what to do there. Could the scope be just what I expect to be returned?
 
Screenshot 2022-09-23 at 10 37 51

libs/api/domains/fiskistofa/src/lib/fiskistofa.resolver.ts
RunarVestmann: Ok, sorry for not doing that. I'll change the names.

libs/clients/fiskistofa/src/lib/api.ts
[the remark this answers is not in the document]
RunarVestmann: I think this is OIDC, I wasn't sure what to put as the scope for the createEnhancedFetch autoAuth param so I just made my own wrapper that gets a new token on 401. I'll look into the autoAuth option and see if I can't use that

libs/clients/fiskistofa/src/lib/api.ts
RunarVestmann: Ok, I had no idea that existed. Thanks for letting me know :)

libs/api/domains/fiskistofa/src/lib/fiskistofa.resolver.ts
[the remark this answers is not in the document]
eirikurn: I see. If there is no change in server state and this is just to do some server calculations then let's map it to GraphQL queries instead.  👍🏻

libs/clients/fiskistofa/src/lib/api.ts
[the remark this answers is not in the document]
eirikurn: If they don't use scopes you can try passing an empty array (skipping the argument in the token endpoint call)

libs/island-ui/core/src/lib/Select/Select.css.ts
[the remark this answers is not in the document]
eirikurn: Haha. At least there's room for improvement, which would be quite easy to do imo. I can also see that the service portal already did some work like this for their code.

libs/clients/fiskistofa/src/lib/api.ts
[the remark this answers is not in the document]
RunarVestmann: They are using scopes here's the scope I get back when calling their backend: "scope": "read:skip read:stodtoflur read:stadaSkips",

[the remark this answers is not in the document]
eirikurn: Ah, I seem to remember when getting a token with client credentials, that if you don't pass the scope parameter you may get all of the scopes which the client has access to.

Still, I'm curious if you need the audience parameter. Haven't seen that used before in this way.

PR #8436, opened 2022-09-16
feat(auth): Add domain column to delegations.

valurefugl:
https://www.notion.so/Update-and-migrate-DB-to-have-domain-specific-delegations-78005ad7e5df473ea279c5af3cf3ed21
https://www.notion.so/Update-current-delegation-API-to-support-new-domain-specific-delegations-8032cb15e5104c2abae5bf306673ae53

## What

Add a domain_name column to delegations in a backwards compatible way.

## Why

So we can have domain specific delegations.

## Screenshots / Gifs

## Checklist:


eirikurn (changes requested):
Is there a unique index on fromNationalId and toNationalId? Should we have one on those columns plus domainName?

I guess this makes the existing API save new delegations with the correct domainName, but there are a couple of things to consider before these tasks are done:

- Should add this field to the Domain ORM class to facilitate further development and syncing test databases.
- Do we need to update any auth-api or auth-admin-api delegation controllers for this? I.e:
  - What happens when we create a second delegation between the same users which belong to different domains, having a different set of scopes?
  - Would the IDS handle it correctly?
  - I assume authPublicApi actorDelegations returns the correct delegation based on the access token.
  - How about meDelegation CRUD, does that also function "correctly" thanks to the access token? Would be good to have a set of API Test around these assumptions.

libs/auth-api-lib/migrations/20220915153658-add-delegation-domain.js
```diff
@@ -0,0 +1,43 @@
...
+            ADD CONSTRAINT fk_delegation_domain FOREIGN KEY (domain_name) REFERENCES domain (name);
+        
+        CREATE INDEX delegation__domain_name__index
+            ON delegation (domain_name);
+        
+        ALTER TABLE delegation
+          DROP CONSTRAINT IF EXISTS unique_from_to;
+
+        ALTER TABLE delegation ADD CONSTRAINT unique_domain_from_to
+          UNIQUE (domain_name, from_national_id, to_national_id);
```
eirikurn, line 20: Nice!

libs/auth-api-lib/migrations/20220915153658-add-delegation-domain.js
```diff
@@ -0,0 +1,43 @@
...
+          ADD COLUMN domain_name VARCHAR DEFAULT '@island.is' NOT NULL;
+    
+        ALTER TABLE delegation
+            ADD CONSTRAINT fk_delegation_domain FOREIGN KEY (domain_name) REFERENCES domain (name);
+        
+        CREATE INDEX delegation__domain_name__index
+            ON delegation (domain_name);
+        
+        ALTER TABLE delegation
+          DROP CONSTRAINT IF EXISTS unique_from_to;
```
eirikurn, line 17: Should be mirrored in `down` to not break `up => down => up` flow?

libs/auth-api-lib/src/lib/delegations/models/delegation.model.ts
```diff
@@ -52,6 +55,14 @@ export class Delegation extends Model {
   })
   toName!: string
 
+  @Column({
+    type: DataType.STRING,
+    allowNull: true,
```
eirikurn, line 60: Should this be nullable? It's not in the migration.

libs/auth-api-lib/src/lib/delegations/delegations.service.ts
```diff
@@ -282,13 +283,17 @@ export class DelegationsService {
     })
 
     // Make sure when using the otherUser filter that we only find one delegation
-    if (otherUser && delegations && delegations.length > 1) {
+    if (
+      otherUser &&
+      delegations &&
+      delegations.filter((d) => d.domainName == DEFAULT_DOMAIN).length > 1
+    ) {
```
eirikurn, line 290: Fyi @saevarma, I guess you'll take this further for multi-domain support.

apps/services/auth/public-api/src/app/modules/delegations/actorDelegations.controller.ts
```diff
@@ -88,6 +89,10 @@ export class ActorDelegationsController {
       )
     }
 
-    return this.delegationsService.findAllIncoming(actor, delegationTypes)
+    return (
+      await this.delegationsService.findAllIncoming(actor, delegationTypes)
+    ).filter(
+      (d) => d.type != DelegationType.Custom || d.domainName == DEFAULT_DOMAIN,
+    )
```
eirikurn, line 96: Let's keep it like this for now, but this (public) API should support delegations for other domains as well (through the domain of the authenticated client, or fancy delegation merging).

We do plan to eventually move the meDelegations controller to an internal API but keep this one public for other orgs, so hopefully we can fix this soon.

libs/auth-api-lib/src/lib/delegations/delegations.service.ts
```diff
@@ -361,7 +366,7 @@ export class DelegationsService {
 
     return uniqBy(
       ([] as DelegationDTO[]).concat(...delegationSets),
-      'fromNationalId',
+      (delegation) => [delegation.domainName, delegation.fromNationalId].join(),
```
eirikurn, line 369: Let's consider if someone has a natural delegation and a custom delegation from the same national ID.

Before, the natural one would override the custom delegation. With this change it will show up twice in the list with unknown behaviour. We have a task to merge delegations properly, but I would consider this a regression until then.

Can you think of a way to keep the old behaviour for now?

eirikurn (changes requested):
Nice!

apps/services/auth/public-api/src/app/modules/delegations/meDelegations.controller.spec.ts
```diff
@@ -534,7 +542,36 @@ describe('MeDelegationsController', () => {
         )
 
         // Act
-        const res = await server.get(`${path}/${mockDelegations.otherUsers.id}`)
+        const res = await server.get(
+          `${path}/${mockDelegations.otherDomain.id}`,
```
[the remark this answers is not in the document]
valurefugl: Yes, this was accidental. Fixed.

apps/services/auth/public-api/src/app/modules/delegations/meDelegations.controller.ts
```diff
@@ -98,7 +100,11 @@ export class MeDelegationsController {
       )
     }
 
-    return this.delegationsService.findAllOutgoing(user, validity, otherUser)
+    return (
+      await this.delegationsService.findAllOutgoing(user, validity, otherUser)
+    ).filter(
+      (d) => d.type != DelegationType.Custom || d.domainName == DEFAULT_DOMAIN,
```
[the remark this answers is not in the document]
valurefugl: Good point, fixed.

apps/services/auth/public-api/src/app/modules/delegations/meDelegations.controller.ts
```diff
@@ -130,7 +136,11 @@ export class MeDelegationsController {
       delegationId,
     )
 
-    if (!delegation) {
+    if (
+      !delegation ||
+      (delegation.type == DelegationType.Custom &&
```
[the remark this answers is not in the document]
valurefugl: Fixed.

libs/auth-api-lib/src/lib/delegations/models/delegation.model.ts
valurefugl: Fixed.

libs/auth-api-lib/src/lib/delegations/delegations.service.ts
valurefugl: Reverted this change, limit the custom delegations to the default domain for now.

We probably need to sort out the delegation merge-ing to do this properly.

libs/auth-api-lib/migrations/20220915153658-add-delegation-domain.js
valurefugl: So delete rows with domain different from the default and then add constraint?

libs/auth-api-lib/src/lib/delegations/delegations.service.ts
eirikurn: Our partner organisation is interested to start testing this on dev as soon as next sprint. Happy to merge this to unblock other work, but we need to either filter the delegations based on the client domain or finish the delegation merging before next sprint. 👍

libs/auth-api-lib/migrations/20220915153658-add-delegation-domain.js
eirikurn: Ah sorry, I actually missed the "IF EXISTS". I just don't want the flow to crash the release but sounds like this should work (and better to not remove data).

apps/services/auth/public-api/src/app/modules/delegations/actorDelegations.controller.ts
```diff
@@ -8,6 +8,7 @@ import {
 import { ApiTags } from '@nestjs/swagger'
 
 import {
+  DEFAULT_DOMAIN,
```
eirikurn, line 11: Can revert this and the redundant `await` below.

eirikurn (approved):
Approving with minor comment (can fix and then add automerge tag).

Let's get this to main and then follow up with improvements to "findAllIncoming" so we can start testing multi domain delegations next sprint 👍

apps/services/auth/public-api/src/app/modules/delegations/actorDelegations.controller.ts
valurefugl: Fixed.

PR #8443, opened 2022-09-16
fix(national-registry): Remove family relation from national registry v1 getFamily

thordurhhh:
## What

* Remove `familyRelation` from from national registry v1 `getFamily`.
* Remove all usage of `familyRelation`from `getFamily`

## Why

The `getFamilyRelation` will not always return correct information regarding the familyRelation.
It can in some cases, if you read the [removed code](https://github.com/island-is/island.is/pull/8443/files#diff-32c7fdb08f542ca01fb7f75566d8df63c7ac83a16817e83b135833c8a26667e8L262), return a child that's over 18 as a spouse.

The correct usage to get the spouse in v1 would be to use [spouse object](https://github.com/island-is/island.is/blob/main/libs/api/domains/national-registry/src/lib/nationalRegistry.service.ts#L60) in `getUser`.
The way to get the [children information](https://github.com/island-is/island.is/blob/main/libs/api/domains/national-registry/src/lib/nationalRegistry.service.ts#L140) would be to use `getChildren`.

I am removing the only current usage of `familyRelation` which is in `parentalLeaveUsage` where it seems to be being used as a fallback for the [correct way](https://github.com/island-is/island.is/blob/31d07b7d10297d651944abcb340b6e89e6d882d8/libs/application/templates/parental-leave/src/lib/parentalLeaveUtils.ts#L205) to get the spouse info.

## Checklist:


Thorrik58 (approved):

Thornvik (approved):

disaerna (approved):
lgtm 💯

eirikurn (approved):
Love this: 
image

Change sounds good but only taking responsibility for api mocks ;)

PR #8459, opened 2022-09-19
feat(j-s): Deliver Indictments

gudjong:
# Deliver Indictments

[Vista dómstólaskjöl S-mála í Auði og LÖKE](https://app.asana.com/0/1199153462262248/1202974314193390/f)

## What

- Uploads court documents to Auður and LÖKE.

## Why

- Part of automating case completion.

## Checklist:

### DIFF
--- a/apps/judicial-system/backend/src/app/modules/case/case.service.ts
+++ b/apps/judicial-system/backend/src/app/modules/case/case.service.ts
@@ -27,6 +27,7 @@ import { MessageType } from '@island.is/judicial-system/message'
 import { EmailService } from '@island.is/email-service'
 import { SIGNED_VERDICT_OVERVIEW_ROUTE } from '@island.is/judicial-system/consts'
 import {
+  CaseFileCategory,
   CaseFileState,
   CaseOrigin,
   CaseState,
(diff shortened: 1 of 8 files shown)


apps/judicial-system/backend/src/app/modules/case/case.service.ts
```diff
@@ -414,12 +415,63 @@ export class CaseService {
     }
   }
 
+  private async deliverIndictmentRulingToCourt(
```
oddsson, line 418: If I'm not mistaken, the only difference between this function and `deliverIndictmentCourtRecortToCourt` is the selection of which documents to deliver. Consider combining these function and add a parameter for the selection.

oddsson (changes requested):
One thing..

apps/judicial-system/backend/src/app/modules/case/case.service.ts
gudjong: I would rather not, because that would result in a lot of conditionals within the function: which document to fetch, which error messages to display when something go wrong, which title to use, which file name to use, etc.

PR #8461, opened 2022-09-19
fix(service-portal): wrong key

thorkellmani:
## What

Fix wrong key

## Why

It's currently using the wrong api key

## Checklist:


darrikonn (approved):

apps/api/src/app/environments/environment.ts
```diff
@@ -119,7 +119,7 @@ const prodConfig = () => ({
   },
   firearmLicense: {
     firearmLicensePassTemplateId: process.env.FIREARM_LICENSE_PASS_TEMPLATE_ID,
-    pkPassApiKey: process.env.VE_PKPASS_API_KEY,
+    pkPassApiKey: process.env.RLS_PKPASS_API_KEY,
```
eirikurn, line 122: Fyi, we should move most configs like these from environment files and `ClientModule.register` functions to use `@island.is/nest/config`.

eirikurn (approved):

apps/api/src/app/environments/environment.ts
thorkellmani: Yeah, I agree, there just was an issue with config injections into a manually created client

PR #8470, opened 2022-09-20
feat(j-s): Give defenders access to S-cases

oddsson:
# Give defenders access to S-cases

[Asana](https://app.asana.com/0/1199153462262248/1202978837143953/f)

## What

Give defenders access to S-cases.

## Why

Defenders need to be able to familiarize themselves with a case before going into court.

### Closed cases
screencapture-localhost-4200-verjandi-akaera-623b95f2-d9d9-4feb-8c1e-594d818757ea-2022-09-23-10_13_00

### Active cases
screencapture-localhost-4200-verjandi-akaera-623b95f2-d9d9-4feb-8c1e-594d818757ea-2022-09-23-10_11_38

## Checklist:


apps/judicial-system/web/pages/verjandi/akaera/[id].ts
```diff
@@ -0,0 +1,3 @@
+import Overview from '@island.is/judicial-system-web/src/routes/Court/components/Overview/Overview'
```
Comment (Arnarkari93) [inline apps/judicial-system/web/pages/verjandi/akaera/[id].ts:1]: Should this not be `routes/Defender/CaseOverview`?

apps/judicial-system/web/src/utils/hooks/useSections/index.ts
```diff
@@ -479,54 +479,60 @@ const useSections = () => {
...
-            isSubpoenaStepValid(workingCase) &&
-            isprosecutorAndDefenderStepValid(workingCase)
-              ? `${constants.INDICTMENTS_COURT_RECORD_ROUTE}/${workingCase.id}`
-              : undefined,
-        },
-      ],
+      children:
+        activeSubSection === undefined
+          ? subsections.map((subsection) => {
+              return { name: '', type: '', href: subsection.href }
```
Arnarkari93, line 533: Why is this needed? Is it to remove the subsections?

Arnarkari93 (changes requested):

apps/judicial-system/web/pages/verjandi/akaera/[id].ts
Comment (oddsson) [reply apps/judicial-system/web/pages/verjandi/akaera/[id].ts:1]: No, we are using the Overview screen for defenders in indictments. The Defender/CaseOverview screen is a completely different design. See 👇 

Screenshot 2022-09-27 at 11 12 31

Comment (Arnarkari93) [reply apps/judicial-system/web/pages/verjandi/akaera/[id].ts:1]: Should we when Create a new version for `Defenders` or move the `Court/Overview` to shared?

apps/judicial-system/web/src/utils/hooks/useSections/index.ts
oddsson: Initially, I set this up so that this function would return an empty array if there was no activeSubsection and by doing that we could render a form stepper with an active section but no subsections, i.e. something like this (in this case the active section would be section nr2, but whatever.)

Screenshot 2022-09-27 at 11 15 34

There was however an issue with this approach because it made [this line](https://github.com/island-is/island.is/blob/main/apps/judicial-system/web/src/routes/Shared/Cases/Cases.tsx#L181) fail because there is no href.

This returns an array with subsections with empty names and type but includes the hrefs.

apps/judicial-system/web/pages/verjandi/akaera/[id].ts
Comment (oddsson) [reply apps/judicial-system/web/pages/verjandi/akaera/[id].ts]: Yea! I've now moved Court/Overview to shared. How do you like this setup?

apps/judicial-system/web/src/routes/Court/components/index.ts
```diff
@@ -1,4 +1,4 @@
 export { default as CourtCaseNumber } from './CourtCaseNumber/CourtCaseNumber'
 export { default as DraftConclusionModal } from './DraftConclusionModal/DraftConclusionModal'
-export { default as Overview } from './Overview/Overview'
+export { default as Overview } from '../../Shared/Overview/Overview'
```
Arnarkari93, line 3:
```suggestion
```

apps/judicial-system/web/pages/verjandi/akaera/[id].ts
Comment (Arnarkari93) [reply apps/judicial-system/web/pages/verjandi/akaera/[id].ts]: Looks good

apps/judicial-system/web/src/utils/hooks/useSections/index.ts
Arnarkari93: Would it be more readable to check if subsections are empty and then define a fallback route? Like done here https://github.com/island-is/island.is/blob/473606d590a4c4799c7a9ea7bb4bc116303e7438/apps/judicial-system/web/src/routes/Shared/Cases/Cases.tsx#L196-L198

oddsson: Totally! I'll get right on that 💪

apps/judicial-system/backend/src/app/modules/case/case.service.ts
```diff
@@ -1124,7 +1125,7 @@ export class CaseService {
 
   async archive(): Promise<ArchiveResponse> {
     const theCase = await this.caseModel.findOne({
-      include: [...includes, { model: CaseFile, as: 'caseFiles' }],
+      include: [...includes],
```
Arnarkari93, line 1128:
```suggestion
      include: includes,
```

apps/judicial-system/web/src/routes/Shared/Cases/Cases.tsx
```diff
@@ -176,9 +176,13 @@ export const Cases: React.FC = () => {
           getInvestigationCaseCourtSections(caseToOpen, user),
         ).href
       } else {
-        // Route to Indictment Overivew section since it always a valid step and
+        // Route to Indictment Overview section since it always a valid step and
         // would be skipped if we route to the last valid step
-        routeTo = getIndictmentsCourtSections(caseToOpen).children[0].href
+        const routeToOpen =
+          getIndictmentsCourtSections(caseToOpen).children[0].href ||
```
Arnarkari93, line 182: Weird that typescript dose not give an error here.
```suggestion
          getIndictmentsCourtSections(caseToOpen).children[0]?.href ||
```

PR #8473, opened 2022-09-20
feat(j-s): fail signature confirmation if upload S3 fails

Arnarkari93:
[Asana - Úrskurður í Auði heitir "Úrskurður leiðrétt" þó að hann sé sá fyrsti (R-223 í test)](https://app.asana.com/0/1199153462262248/1202929403207278/f)
## What

- Fix ruling being named "Ruling modified" initially
- Fail the ruling signature confirmation if upload to S3 fails

## Why

- Trying to simplify flow and edge cases in this flow: Ruling signed -> Upload S3 -> Update Case -> Upload to court
- This has the disatvantage that we reliant on AWS S3 and I think we are ok with that.

## Checklist:


apps/judicial-system/backend/src/app/modules/case/case.service.ts
```diff
@@ -372,49 +372,6 @@ export class CaseService {
     }
   }
 
-  private async uploadSignedRulingPdf(
```
Arnarkari93, line 375: This function is no longer being used

apps/judicial-system/backend/src/app/modules/case/case.service.ts
```diff
@@ -473,14 +430,12 @@ export class CaseService {
...
       .then((pdf) => this.uploadSignedRulingPdfToCourt(theCase, pdf))
       .catch(() => {
-        // The signed ruling should have been delivered to the court
-        // either directly to the court system or via email
         this.logger.info(
           `The ruling for case ${theCase.id} was not found in AWS S3`,
           { caseId: theCase.id },
         )
 
-        return true
```
gudjong, line 483: Consider logging an error now since we are failing the ruling delivery

gudjong (changes requested):
One coment

apps/judicial-system/backend/src/app/modules/case/case.service.ts
Arnarkari93: I decided to return `false` so that an email would be send out (Ekki tókst að vista þingbók og/eða úrskurð í máli {courtCaseNumber} í Auði....). If we throw an errror that email wont be sent out.  

What does that mean for the current implementation to fail the ruling delivery? Here I would like to try again if this fails but we have no support for that right?

gudjong: Right, we will get that soon I hope.

gudjong (approved):

PR #8477, opened 2022-09-20
feat(j-s): Prosecutor Documents

gudjong:
# Prosecutor Documents

[Vista sækjendaskjöl S-mála í Auði](https://app.asana.com/0/1199153462262248/1203000767767668/f)

## What

- Uploads prosecutor documents for indictments in Auður when a court case number is assigned to a case.
- Moves the upload to queue.

## Why

The documents must be delivered to court.

## Checklist:


apps/judicial-system/backend/src/app/messages/courtUpload.ts
```diff
@@ -1,22 +1,52 @@
...
+    id: 'judicial.system.backend:court_upload.indictment',
+    defaultMessage: 'Ákæra {courtCaseNumber}',
+    description: 'Notaður sem nafn á ákæru í Auði.',
+  },
+  criminalRecord: {
+    id: 'judicial.system.backend:court_upload.criminal_record',
+    defaultMessage: 'Sakavottorð {courtCaseNumber}',
+    description: 'Notaður sem nafn á sakavottorði í Auði.',
+  },
+  coustBreakdown: {
```
oddsson, line 24:
```suggestion
  costBreakdown: {
```

apps/judicial-system/backend/src/app/modules/case/case.service.ts
```diff
@@ -416,69 +426,46 @@ export class CaseService {
...
-          .then(() => true)
-      })
-      .catch((reason) => {
-        // Tolerate failure, but log error
-        this.logger.error(
-          `Failed to deliver the court record for case ${theCase.id} to court`,
-          { reason },
-        )
+    return isIndictmentCase(theCase.type)
+      ? (await this.deliverCaseFileToCourt(
```
oddsson, line 433: Consider refactoring `deliverCaseFileToCourt` function to accept `CaseFileCategory[]` instead of a single `CaseFileCategory` and then run operations on each of the CaseFileCategory.

apps/judicial-system/backend/src/app/modules/file/file.service.ts
```diff
@@ -71,11 +92,74 @@ export class FileService {
...
+          courtCaseNumber: theCase.courtCaseNumber,
+        })
+        break
+      default:
+        courtDocumentFolder = CourtDocumentFolder.CASE_DOCUMENTS
+        subject = file.name
+    }
+
+    const fileName = file.category
+      ? file.name.replace(/^.+\./, `${subject}.`)
```
oddsson, line 154: Consider moving the regex into a variable with a descriptive name for better code readability

apps/judicial-system/message-handler/src/app/prosecutorDocumentsDelivery.service.ts
```diff
@@ -0,0 +1,63 @@
...
+          if (!response.requestDeliveredToCourt) {
+            logger.error(
+              `Failed to deliver the request for case ${caseId} to court`,
+            )
+          }
+
+          return
+        }
+
+        logger.error(
```
oddsson, line 47: I think it's a bit weird to have this code here and in the .catch. Consider throwing an error here instead.

oddsson (changes requested):
Just a couple of comments

Arnarkari93 (changes requested):
Nice work! Just one nitpick regarding random enums

apps/judicial-system/backend/src/app/messages/courtUpload.ts
gudjong: On it.

apps/judicial-system/backend/src/app/modules/case/case.service.ts
gudjong: I intended to do this in a future PR as we are rewriting/refactoring queuing as a whole. But I will take a look.

gudjong: Done, managed to delete some code 🥇

apps/judicial-system/backend/src/app/modules/file/file.service.ts
gudjong: Makes sense. Will do.

apps/judicial-system/message-handler/src/app/prosecutorDocumentsDelivery.service.ts
gudjong: I can understand where you are coming from, but it is not exactly the same code. Here we are logging the failure response from the backend. In the catch clause we are logging an exception thrown during the network call or the json parse. I will not make this change.

libs/judicial-system/message/src/lib/test/postMessageToQueue.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+      } catch (error) {
+        then.error = error as Error
+      }
+
+      return then
+    }
+  })
+
+  describe('message posted to queue', () => {
+    const message = { type: randomEnum(MessageType), caseId: uuid() }
```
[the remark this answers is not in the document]
gudjong: I kind of felt it would be an over kill, but you are right. There is no harm making the test more detailed and consistent. Will do.

libs/judicial-system/message/src/lib/test/postMessageToQueue.spec.ts
```diff
@@ -0,0 +1,58 @@
+import { uuid } from 'uuidv4'
+import each from 'jest-each'
+
+import { QueueService } from '@island.is/message-queue'
+
+import { MessageType, Message } from '../message'
+import { randomEnum } from './random'
```
oddsson, line 7:
```suggestion
```

oddsson (changes requested):
You forgot to remove the `randomEnum` import. Otherwise, it's awesome :)

Arnarkari93 (approved):
+1 on removing the randomEnum. 👍  LGMT,

oddsson (approved):

PR #8486, opened 2022-09-22
feat(service-portal): Delegation V2 UI

snaerseljan:
# Delegation V2 UI

[Task](https://www.notion.so/service-portal-Setup-new-project-for-delegations-75cb14f959aa43a682df56e9c8be61b8)

## What

This is the initial project setup for our updated delegation system.
The plan is to have the new module behind a feature flag until it is ready since it is a fairly big feature.

- Feature flag has been created. `outgoingdelegationsv2`. This flag will be used throughout all services/endpoints that are connected to delegations.
- Both modules will be active on dev. The new module behind a feature flag.
- At first Module `/libs/service-portal/settings/access-control` was copied  to a new location `/libs/service-portal/access-control/delegations`. The work on the new delegations UI version has been trying to reuse as much as possible from the old module as well as refactor and restructure the module.
- Routes in the new  module are a little different from the old module:
   - `/adgangsstyring/umbod`, 
   - `/adgangsstyring/umbod/veita`
   - `/adgangsstyring/umbod/:delegationId`

## Why

We want to set up the initial structure of the of delegations v2.

## Screenshots / Gifs

Here below you can see images from the new module which is behind a feature flag.

Delegations list
image

Delete modal
image

Create new delegation
image

Grant access to delegation screen
image

Confirm grant access modal
image

## Checklist:


libs/service-portal/access-control/delegations/src/screens/AccessControl/AccessControl.tsx
```diff
@@ -0,0 +1,127 @@
...
+        <Box marginTop={8}>
+          <IntroHeader
+            title={m.accessControl}
+            intro={formatMessage({
+              id: 'sp.settings-access-control:home-intro',
+              defaultMessage:
+                'Hérna kemur listi yfir þau umboð sem þú hefur gefið öðrum. Þú getur eytt umboðum eða bætt við nýjum.',
+            })}
+          />
+          <Accesses />
```
disaerna, line 108: Isn't this displaying the same thing? Accesses given to others?

disaerna:
@snaerseljan - I recommend that you update the codeowner file and set this module as your (Aranja) project, just like the "old" access control 😊

libs/service-portal/access-control/delegations/src/screens/AccessControl/AccessControl.tsx
snaerseljan: Yes, this PR is just to get the show running so you can just ignore this.
Modified API endpoints have not yet been created and the design is well on its way. The plan was to basically just start with a copy-paste project and then gradually add updates on screen bases.
I started adding dummy tabs but this will all be updated accordingly in time :)

disaerna: Ah okey I see, thanks for the info :)

snaerseljan:
@disaerna Thanks for letting me know, I will do that.

SUMMARY.md
```diff
@@ -223,6 +223,7 @@
 - [Regulations](libs/regulations/README.md)
 - [Residence History](libs/residence-history/README.md)
 - [Service Portal](libs/service-portal/README.md)
+    - [Service Portal Access Control Delegations](libs/service-portal/access-control/delegations/README.md)
```
eirikurn, line 226: Should mark the library as hidden from gitbook (see README.md in other service portal modules).

apps/service-portal/src/store/modules.ts
```diff
@@ -45,11 +46,13 @@ export type ModuleKeys =
   | 'licenses'
   | 'petitions'
   | 'vehicles'
+  | 'delegations'
 
 export const featureFlaggedModules: ModuleKeys[] = [
   'documentProvider',
   'icelandicNamesRegistry',
   'petitions',
+  'delegations',
```
eirikurn, line 55: Fyi, we recently encountered an issue in production when cleaning up this kind of magic feature flag which is shared with the API. For this reason I would like to change this `featureFlaggedModules` functionality to depend on feature flags exported from `@island.is/feature-flags`.

libs/service-portal/access-control/delegations/src/screens/AccessControl/AccessControl.tsx
```diff
@@ -0,0 +1,152 @@
...
+  const isDelegation = Boolean(actor)
+
+  const isCompany = userInfo.profile.subjectType === 'legalEntity'
+  const personDelegation = isDelegation && !isCompany
+
+  if (personDelegation) {
+    return <AccessDenied userInfo={userInfo} client={client} />
+  }
+
+  const creatTab = (
```
eirikurn, line 85: Typo? `createTab`

libs/service-portal/access-control/delegations/src/screens/AccessControl/AccessControl.tsx
```diff
@@ -0,0 +1,152 @@
...
+              },
+              {
+                id: 'sp.access-control-delegations:from-me-title',
+                defaultMessage: 'Umboð sem ég hef veitt',
+              },
+            ),
+            creatTab(m.accessControlDelegationsToMe, {
+              id: 'sp.access-control-delegations:to-me-title',
+              defaultMessage: 'Umboð sem ég hef fengið',
+            }),
```
eirikurn, line 143: I imagine the `content` of the tabs to grow more different over time. Often a bit of code duplication is easier to read:

```
tabs={[
  {
    label: m.accessControlDelegationsFromMe,
    content: <DelegationsFromMeTab />
  },
  {
    label: m.accessControlDelegationsToMe,
    content: <DelegationsToMeTab />
  },
]}
```

libs/service-portal/access-control/delegations/src/screens/AccessControl/AccessControl.tsx
```diff
@@ -0,0 +1,152 @@
...
+
+  return (
+    <Box>
+      {!loading && data && data?.authDelegations?.length > 0 && (
+        <Tabs
+          selected={
+            isDelegationToMe ? delegationToMeTabId : delegationFromMeTabId
+          }
+          onClick={tabClickHandler}
+          label="This is used as the aria-label as well"
```
eirikurn, line 128: Need a proper translated value.

libs/service-portal/access-control/delegations/src/screens/AccessControl/AccessControl.tsx
```diff
@@ -0,0 +1,152 @@
...
+              'Hérna kemur listi yfir þau umboð sem þú hefur gefið öðrum. Þú getur eytt umboðum eða bætt við nýjum.',
+          })}
+        />
+        <Accesses />
+      </Box>
+    ),
+  })
+
+  const delegationFromMeTabId = '0'
+  const delegationToMeTabId = '1'
```
eirikurn, line 106: Should be constants outside function?

libs/service-portal/access-control/delegations/src/screens/AccessControl/AccessControl.tsx
```diff
@@ -0,0 +1,152 @@
...
+    // Make sure not to add to history stack the same route twice in a row
+    if (url !== location.pathname) {
+      history.push(url)
+    }
+  }
+
+  return (
+    <Box>
+      {!loading && data && data?.authDelegations?.length > 0 && (
+        <Tabs
```
eirikurn, line 123: Should skip the tabs for now since they come way later in M2?

libs/service-portal/access-control/delegations/src/screens/AccessControl/AccessControl.tsx
```diff
@@ -0,0 +1,152 @@
...
+  }
+  const actor = userInfo.profile.actor
+  const isDelegation = Boolean(actor)
+
+  const isCompany = userInfo.profile.subjectType === 'legalEntity'
+  const personDelegation = isDelegation && !isCompany
+
+  if (personDelegation) {
+    return <AccessDenied userInfo={userInfo} client={client} />
+  }
```
eirikurn, line 83: Both of these early returns should probably be inside the tab screen component? Also the data fetching logic? And we should return access denied before no data screen (probably not your code i know).

libs/service-portal/core/src/lib/navigation/masterNavigation.ts
```diff
@@ -332,6 +332,26 @@ export const servicePortalMasterNavigation: ServicePortalNavigationItem[] = [
         },
         description: m.accessControlDescription,
       },
+      // Aðgangsstýring umboð
+      {
+        name: m.accessControl,
```
eirikurn, line 337: Maybe add something like "V2" to this label for now while users might see both at the same time? :)

libs/island-ui/core/src/lib/Tabs/Tabs.tsx
```diff
@@ -30,6 +31,7 @@ export const Tabs: FC<TabInterface> = ({
   tabs,
   contentBackground = 'purple100',
   size = 'md',
+  onClick: onClickHandler,
```
eirikurn, line 34: This onClick event probably doesn't handle all the ways that the tabs can change, and annoying to deal with double clicks outside.

Should look into exposing an `onChange` property instead. Maybe by tracking changes to `tab.selectedId` using a ref?

eirikurn (changes requested):
Nice! 🙌

Also, to avoid confusion, we may want to disable the feature flag on dev and only allow our national ids for now.

apps/service-portal/src/store/modules.ts
snaerseljan: Discussed locally. I will merge wait with the merge until the feature flagged modules functionality is merged

libs/service-portal/access-control/delegations/src/screens/AccessControl/AccessControl.tsx
snaerseljan: This has been refactored and changed

libs/service-portal/access-control/delegations/src/screens/AccessControl/AccessControl.tsx
snaerseljan: I have updated the code a bit and the content for to me delegation tab is now null. I think it won't be long until @saevarma has updated the query to support direction outgoing and incoming.
Let me know if you still want me to remove until then

libs/island-ui/core/src/lib/Tabs/Tabs.tsx
snaerseljan: Good idea since the library does not expose listeners

libs/service-portal/access-control/delegations/src/screens/AccessControl/AccessControl.tsx
snaerseljan: fixed

oddsson (approved):

libs/application/core/src/lib/messages.ts
```diff
@@ -11,6 +11,21 @@ export const coreMessages = defineMessages({
     defaultMessage: 'Til baka',
     description: 'Back',
   },
+  buttonCancel: {
```
disaerna, line 14: Should these translations strings be in applications core? 
I would recommend using service portal core messages instead if those strings are only used there.

libs/service-portal/core/src/lib/messages.ts
```diff
@@ -535,6 +535,26 @@ export const m = defineMessages({
     id: 'service.portal:accessControl',
     defaultMessage: 'Aðgangsstýring',
   },
+  accessControlV2: {
+    id: 'service.portal:accessControlV2',
+    defaultMessage: 'Aðgangsstýring V2',
+  },
+  accessControlDelegations: {
```
disaerna, line 542: These should not be in service portal core messages. Instead use a seperate file for translation strings with the namespace: **sp.access-control-delegations**. 
If there are any other namespaces inside core the extract string command will delete other 😞

apps/service-portal/src/components/Layout/Layout.tsx
```diff
@@ -57,7 +57,7 @@ const Layout: FC = ({ children }) => {
         <Box as="main" component="main">
           <GridContainer className={styles.layoutContainer}>
             <GridRow>
-              <GridColumn span={'11/12'} className={styles.layoutGrid}>
+              <GridColumn span={'12/12'} className={styles.layoutGrid}>
```
disaerna, line 60: This should be 11/12 according to design. But we are discussing changes for the sidebar so the layout might change in the near future.

snaerseljan: Okey good point regarding desktop, but we still need to fix it for mobile to expand to the container else it looks the following in mobile. Notice the right margin.
image

I will update the span to `{['12/12', '12/12', '12/12', '11/12']}`
Good with that? :)

libs/application/core/src/lib/messages.ts
snaerseljan: I assumed that the cancel button translation was common. In fact it is more common than many translations in the core messages. So is this maybe old convention to put in the coreMessages instead messages on project bases?

apps/service-portal/src/components/Layout/Layout.tsx
disaerna: Yes thanks! Definitely, didn't spot the mobile part before 🙈

libs/service-portal/core/src/lib/messages.ts
disaerna: I see the comment came out pretty confusing. I was talking about what a few strings beneath the navigation string.

libs/application/core/src/lib/messages.ts
disaerna: Yeah I agree it is common but the application system is not right for that either. All translations we have in service portal we keep inside service portal. That way we can control and translate more easily each module.

libs/service-portal/core/src/lib/messages.ts
snaerseljan: Ahhh yes I misunderstood. Thx I will move it :)

snaerseljan: Or on second though as I look at it better. I see that one of the string is unused so I removed it. However the other three are all used in masterNavigation, i.e.

```
// Aðgangsstýring umboð
      {
        name: m.accessControlV2, // Here
        path: ServicePortalPath.AccessControlDelegations,
        icon: {
          icon: 'lockClosed',
        },
        description: m.accessControlDescription,
        children: [
          {
            name: m.accessControlDelegations, // Here
            path: ServicePortalPath.AccessControlDelegations,
            children: [
              {
                name: m.accessControlGrant, // Here
                path: ServicePortalPath.AccessControlDelegationsGrant,
                navHide: true,
              },
            ],
          },
          {
            name: m.accessControlDelegationsToMe, // Here
            path: ServicePortalPath.AccessControlDelegationsToMe,
            navHide: true,
          },
          {
            name: m.accessControlAccess, // here
            path: ServicePortalPath.AccessControlDelegationAccess,
            navHide: true,
          },
        ],
      },
```

libs/application/core/src/lib/messages.ts
snaerseljan: Sure I will move them. They have been moved.

libs/service-portal/core/src/lib/messages.ts
disaerna: Okey great, just make sure they have the same namespace as other core messages 👍

snaerseljan: Ahh yes thx 🤦

libs/api/domains/auth/src/lib/resolvers/domain.resolver.ts
```diff
@@ -28,4 +34,18 @@ export class DomainResolver {
   ): Promise<Domain[]> {
     return this.domain.getDomains(user, input)
   }
+
+  @ResolveField('organisationLogoUrl', () => String, { nullable: true })
+  async resolveOrganisationLogoKey(
```
eirikurn, line 39:
```suggestion
  async resolveOrganisationLogoUrl(
```

libs/api/domains/auth/src/lib/resolvers/domain.resolver.ts
```diff
@@ -28,4 +34,18 @@ export class DomainResolver {
...
+    @Loader(OrganizationLogoLoader)
+    organizationLogoLoader: OrganizationLogoDataLoader,
+    @Parent() domain: Domain,
+    @Args('lang', { type: () => String, nullable: true, defaultValue: 'is' })
+    lang: string,
```
eirikurn, line 44: Organisation logos should not be translated. Can get rid of the lang argument and the object below. Just pass in the logo key directly.

libs/cms/src/lib/cms.contentful.service.ts
```diff
@@ -155,6 +163,40 @@ export class CmsContentfulService {
     }
   }
 
+  async getOrganizationsLogo({
```
eirikurn, line 166:
```suggestion
  async getOrganizationLogos({
```

libs/cms/src/lib/cms.contentful.service.ts
```diff
@@ -155,6 +163,40 @@ export class CmsContentfulService {
...
+    lang = 'is-IS',
+    organizationTitles,
+  }: {
+    lang?: string
```
eirikurn, line 170: Can skip the lang here as well?

libs/cms/src/lib/dto/getOrganizationsLogo.input.ts
```diff
@@ -0,0 +1,14 @@
+import { Field, InputType } from '@nestjs/graphql'
+import { IsOptional, IsString } from 'class-validator'
+
+@InputType()
+export class OrganizationLogoInput {
```
eirikurn, line 5: Don't think we need this type.

libs/cms/src/lib/loaders/organizationLogo.loader.ts
```diff
@@ -0,0 +1,40 @@
...
+
+    const organizationLogos = await this.cmsContentfulService.getOrganizationsLogo(
+      { organizationTitles, lang },
+    )
+
+    return organizationLogos
+  }
+
+  generateDataLoader(): OrganizationLogoDataLoader {
+    return new DataLoader(this.loadOrganizationLogo.bind(this))
```
eirikurn, line 38: Should define loadOrganizationLogo as an arrow function instead of using bind here since we're not binding any arguments.

libs/service-portal/core/src/lib/messages.ts
```diff
@@ -805,4 +821,19 @@ export const m = defineMessages({
     id: 'service.portal:error-screen-title',
     defaultMessage: 'Villa',
   },
+  buttonCancel: {
+    id: 'application.system:button.cancel',
```
eirikurn, line 825: Fix the namespace in these keys.

libs/service-portal/settings/access-control/src/screens/AccessControl/AccessControl.tsx
```diff
@@ -12,25 +12,20 @@ import {
...
-        validTo
-      }
-    }
-  }
-`
 const AccessControl: ServicePortalModuleComponent = ({ userInfo, client }) => {
   useNamespaces('sp.settings-access-control')
-  const { data, loading } = useQuery<Query>(AuthDelegationsQuery, {
-    variables: { input: { domain: ISLAND_DOMAIN } },
+  const { data, loading } = useAuthDelegationsQuery({
```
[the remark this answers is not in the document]
snaerseljan: hmm okey, but just comes down to the exact same thing :)

libs/service-portal/graphql/src/lib/queries/getOrganizations.ts
```diff
@@ -14,8 +14,8 @@ const OrganizationFragment = gql`
 `
 
 export const GET_ORGANIZATIONS_QUERY = gql`
-  query getOrganizations {
-    getOrganizations(input: { lang: "is-IS", perPage: 200 }) {
+  query getOrganizations($input: GetOrganizationsInput) {
+    getOrganizations(input: $input) {
```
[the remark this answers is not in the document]
snaerseljan: no it does not complain and that is intentional. The fields in the input are preset in the resolver so in my mind it is good to make the input optional. Do you disagree?

libs/service-portal/graphql/src/lib/queries/delegations.ts
```diff
@@ -0,0 +1,48 @@
...
+        name
+      }
+      ... on AuthCustomDelegation {
+        ...AuthCustomDelegationFragment
+      }
+    }
+  }
+`
+
+export const AUTH_DELEGATIONS_LIST_QUERY = gql`
```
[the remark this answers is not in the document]
snaerseljan: no it is not use. I will remove

libs/service-portal/graphql/src/lib/queries/apiScopes.ts
```diff
@@ -0,0 +1,47 @@
+import { gql } from '@apollo/client'
+import { authApiScopeFragment } from '../fragments/scope'
+
+export const AUTH_API_SCOPES_QUERY = gql`
+  query AuthApiScopes($input: AuthApiScopesInput!) {
```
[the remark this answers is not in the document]
snaerseljan: nope

libs/service-portal/graphql/src/lib/queries/apiScopes.ts
```diff
@@ -0,0 +1,47 @@
...
+        displayName
+        description
+        group {
+          name
+          displayName
+          description
+          children {
+            ...AuthApiScopeFragment
+          }
+        }
```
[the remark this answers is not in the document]
snaerseljan: Yes good idea

libs/service-portal/access-control/delegations/src/components/DelegationsAccessGuard/index.ts
```diff
@@ -0,0 +1 @@
+export { DelegationsAccessGuard } from './DelegationsAccessGuard'
```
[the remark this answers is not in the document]
snaerseljan: Good point. I updated the component folder structure to the following tree and got rid of barrel files on component level.
image

libs/service-portal/access-control/delegations/src/components/DelegationsFromMe/DelegationsFromMe.tsx
```diff
@@ -0,0 +1,121 @@
...
+        domain: domainName,
+      },
+    })
+  }
+
+  return (
+    <>
+      <GridContainer>
+        <GridRow>
+          <GridColumn paddingBottom={4} span="12/12">
```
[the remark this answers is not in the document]
snaerseljan: No definitely not copy paste from older code

libs/service-portal/access-control/delegations/src/components/IdentityCard/IdentityCard.tsx
```diff
@@ -0,0 +1,66 @@
...
+      justifyContent="center"
+      background={color ? `${color}100` : 'white'}
+      paddingX={[3, 4]}
+      paddingY={3}
+      borderRadius="standard"
+      width="full"
+      {...(!color ? { borderColor: 'blue200', borderWidth: 'standard' } : {})}
+    >
+      {imgSrc ? (
+        <Box display="flex" alignItems="center" columnGap={3}>
```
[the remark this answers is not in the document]
snaerseljan: in my mind yes it helps with alignment and gap between columns.

libs/service-portal/access-control/delegations/src/components/access/access.utils.ts
```diff
@@ -0,0 +1,108 @@
...
+  // Scope parent of the children
+  {
+    ...apiScopeGroup,
+    model: `${GROUP_PREFIX}.${startIndex}`,
+  },
+  // Scope children
+  ...(apiScopeGroup.children?.map((scope, childIndex) => ({
+    ...scope,
+    model: `${SCOPE_PREFIX}.${startIndex + childIndex}`,
+  })) || []),
```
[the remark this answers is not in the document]
snaerseljan: I would say yes and no. I tried to reuse the older code as we discussed. The AccessItem component expects data in this format and that is why I did it in this way. I remember when doing this that I thought to my self. Hmmm this is not that more simpler 😅  
I would clearly like to refactor this properly and do this a lot simpler. For now I think we should do it as a separate task, because it could take a little time which we don't have right? 
What do you think?

libs/service-portal/access-control/delegations/src/hooks/useDomains.ts
```diff
@@ -0,0 +1,108 @@
...
+  /**
+   * Updates the domain name in state, session storage and query string.
+   * If no domain exist in query string then we skip it.
+   */
+  const updateDomainName = (value: string | null) => {
+    const option = getOptionByName(value) ?? allDomainsOption
+
+    const newDomainName = option.value
+    setDomainName(newDomainName)
+    sessionStorage.setItem('domain', newDomainName)
```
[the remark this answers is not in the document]
snaerseljan: Agree and disagree. I would be against using global variable or storing it on the window.
I would like to keep the sessionStorage but with a check to make sure it is supported by the browser.
But then again still, are we aiming for IE 6 and 7 again? Because it is supported by all browsers some time ago. Or at least all the known ones.
https://caniuse.com/?search=sessionstorage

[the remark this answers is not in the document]
snaerseljan: Also we are using sessionStorage all over in the monorepo without the guard.

libs/service-portal/access-control/delegations/src/hooks/useDomains.ts
```diff
@@ -0,0 +1,108 @@
...
+        }
+      } else {
+        updateDomainName(ISLAND_DOMAIN)
+      }
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [data?.authDomains])
+
+  return {
+    domainName: domainName === ALL_DOMAINS ? null : domainName,
```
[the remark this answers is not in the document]
snaerseljan: sure

libs/service-portal/access-control/delegations/src/index.ts
```diff
@@ -0,0 +1,56 @@
...
+} from '@island.is/service-portal/core'
+
+export const delegationsModule: ServicePortalModule = {
+  name: 'Aðgangsstýring',
+  featureFlag: Features.outgoingDelegationsV2,
+  widgets: () => [],
+  routes: ({ userInfo }) => {
+    const isCompany = userInfo.profile['subjectType'] === 'legalEntity'
+    const isDelegation = Boolean(userInfo.profile.actor)
+    const personDelegation = isDelegation && !isCompany
```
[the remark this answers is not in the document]
snaerseljan: Nice will update. Old code again :)

libs/service-portal/access-control/delegations/src/components/access/AccessCard.tsx
```diff
@@ -0,0 +1,154 @@
...
+}
+
+export const AccessCard = ({ delegation, onDelete }: AccessCardProps) => {
+  const { formatMessage } = useLocale()
+  const history = useHistory()
+  const { pathname } = useLocation()
+  const tags = delegation.scopes.map((scope) => ({
+    name: scope.displayName,
+    isExpired: isDateExpired(scope.validTo),
+  }))
```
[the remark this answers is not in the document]
snaerseljan: 👍 https://www.notion.so/Delegations-access-cards-tags-update-665c49843bc14cda915d0f96a38fa2d8

libs/service-portal/access-control/delegations/src/components/access/AccessConfirmModal.tsx
```diff
@@ -0,0 +1,222 @@
...
+  const [error, setError] = useState(false)
+  const { md } = useBreakpoint()
+
+  const onConfirmHandler = async () => {
+    if (!delegation.id || !scopes) {
+      setError(true)
+      return
+    }
+
+    onConfirm()
```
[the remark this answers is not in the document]
snaerseljan: I added logic to support error if update delegation fails

libs/service-portal/access-control/delegations/src/components/access/AccessForm.tsx
```diff
@@ -0,0 +1,222 @@
...
+        domain: delegation.domain.name,
+        lang,
+      },
+    },
+  })
+
+  const { authScopeTree } = scopeTreeData || {}
+
+  const methods = useForm<{
+    [SCOPE_PREFIX]: AccessFormScope[]
```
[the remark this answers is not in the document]
snaerseljan: are you talking about doing ` const { authScopeTree: scope } = scopeTreeData || {}` instead?

apps/service-portal/src/components/ContentBreadcrumbs/ContentBreadcrumbs.tsx
```diff
@@ -82,7 +82,7 @@ const ContentBreadcrumbs: FC = () => {
   if (items.length < 2) return null
 
   return (
-    <Box paddingTop={[0, 3]} paddingBottom={[2, 3]}>
+    <Box paddingTop={3} paddingBottom={[2, 3]}>
```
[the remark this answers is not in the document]
snaerseljan: Not sure I will revert. I think the header in mobile is also incorrect regarding spacing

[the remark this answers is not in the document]
snaerseljan: also just in general the padding bottom and top should be controlled in this case by the parent.

libs/service-portal/access-control/delegations/src/hooks/useDomains.ts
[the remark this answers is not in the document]
snaerseljan: Discussed offline

libs/service-portal/access-control/delegations/src/components/access/access.utils.ts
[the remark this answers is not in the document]
snaerseljan: Discussed offline. Decide to refactor later, see task
https://www.notion.so/Refactor-AccessItem-and-AccessForm-components-e7e276d9eca2423abb4b343be7c14c90

kksteini (approved):
Codeowner files LGTM

libs/service-portal/access-control/delegations/src/components/delegations/DelegationsHeader.tsx
```diff
@@ -0,0 +1,103 @@
...
+        {loading ? (
+          <SkeletonLoader height={71} />
+        ) : (
+          <Select
+            label={formatMessage(m.accessControl)}
+            size="xs"
+            name="domain"
+            backgroundColor="blue"
+            id="domain"
+            noOptionsMessage="Enginn valmöguleiki"
```
disaerna, line 62: Translation missing?

libs/service-portal/access-control/delegations/src/hooks/useDomains.ts
```diff
@@ -0,0 +1,114 @@
...
+  const defaultLabel = formatMessage({
+    id: 'sp.access-control-delegations:all-domains',
+    defaultMessage: 'Öll kerfi',
+  })
+  const allDomainsOption = {
+    label: defaultLabel,
+    value: ALL_DOMAINS,
+  }
+
+  const { data, loading } = useAuthDomainsQuery({
```
disaerna, line 44: We received comment from Core team about graphql queries and not storing them in one central place:

*I know this is done almost everywhere, but we really shouldn't store queries in a central location and query everything. This takes the power away from GraphQL, and treats it like a Rest API.*

*Components should query the needed data. GraphQL will take care of merging those requests into a batch operation and remove redundant fields.*

libs/service-portal/access-control/delegations/src/screens/Access/Access.tsx
```diff
@@ -0,0 +1,177 @@
...
+                  <DatePicker
+                    id="validityPeriod"
+                    size="sm"
+                    label={formatMessage(m.date)}
+                    backgroundColor="blue"
+                    minDate={new Date()}
+                    selected={validityPeriod ? validityPeriod : defaultDate}
+                    locale={lang}
+                    handleChange={setValidityPeriod}
+                    placeholderText="Veldu dagsetningu"
```
disaerna, line 152: Translation missing?

disaerna (approved):
Minor comments otherwise nice 💯

libs/service-portal/access-control/delegations/src/hooks/useDomains.ts
snaerseljan: Good and valid point. 
There have been some debates in the past as I understood it on what would be the preferred way of doing the queries.
There are ups and downs to both ways that where discusses, i.e. let queries coexist inside component files or generate hooks in one or more places.

We evaluated that we would prefer typed queries and responses over having the queries where the components live.
In a perfect world we would like to run codegens on module bases so that we don't have all the hooks in one specific lib, but that dosen't scale well when generating types from the GraphQL API.

libs/service-portal/graphql/src/lib/queries/getOrganizations.ts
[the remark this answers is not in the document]
eirikurn: No just curious and had forgotten the syntax for "required" arguments/fields in GraphQL.

libs/cms/src/lib/loaders/organizationLogo.loader.ts
```diff
@@ -0,0 +1,28 @@
...
+
+@Injectable()
+export class OrganizationLogoLoader implements NestDataLoader<string, LogoUrl> {
+  constructor(private readonly cmsContentfulService: CmsContentfulService) {}
+
+  async loadOrganizationLogo(
+    organizationTitles: readonly string[],
+  ): Promise<Array<LogoUrl>> {
+    const organizationLogos = await this.cmsContentfulService.getOrganizationLogos(
+      organizationTitles.map((title) => title),
```
eirikurn, line 19: Redundant map.

libs/cms/src/lib/cms.contentful.service.ts
```diff
@@ -155,6 +163,36 @@ export class CmsContentfulService {
     }
   }
 
+  async getOrganizationLogos(
+    organizationTitles: string[],
+  ): Promise<Array<string | null>> {
+    const params = {
+      ['content_type']: 'organization',
+      select: 'fields.logo,fields.title',
+      'fields.title[in]': organizationTitles.join(','),
```
eirikurn, line 172: Does the library not support arrays? :(

I can't find great documentation around this but we might need to urlencode each title somehow ... or hope there's no organisation with "#" or "," in its name.

eirikurn (approved):
Nice!

.github/workflows/pullrequest.yml
```diff
@@ -81,7 +81,7 @@ jobs:
...
+          export HASH=${{ hashFiles('scripts/schemas.js', 'libs/cms/src/lib/generated/contentfulTypes.d.ts', 'apps/air-discount-scheme/web/i18n/withLocale.tsx', 'apps/air-discount-scheme/web/components/AppLayout/AppLayout.tsx', 'apps/air-discount-scheme/web/components/Header/Header.tsx', 'apps/air-discount-scheme/web/screens/**.tsx', 'apps/**/codegen.yml', 'libs/**/codegen.yml', 'apps/**/*.model.ts', 'libs/**/*.model.ts', 'apps/**/*.enum.ts', 'libs/**/*.enum.ts', 'apps/**/queries/**/*.tsx?', 'libs/**/queries/**/*.tsx?', 'libs/**/mutations/**/*.tsx?', 'libs/**/fragments/**/*.tsx?' , 'apps/**/*.resolver.ts', 'libs/**/*.resolver.ts', 'apps/**/*.service.ts', 'libs/**/*.service.ts', 'apps/**/*.dto.ts', 'libs/**/*.dto.ts', 'apps/**/*.input.ts', 'libs/**/*.input.ts', 'apps/**/*.module.ts', 'libs/**/*.module.ts', 'apps/**/*.controller.ts', 'libs/**/*.controller.ts', 'apps/**/*.union.ts', 'libs/**/*.union.ts', 'apps/**/*.graphql.ts', 'libs/**/*.graphql.ts') }}
```
brynjarorng, line 84: Did files move from apps to libs? When looking at the source I cannot see this being done

eirikurn: Probably not. The diff is a bit funky but the only thing that's changing is that we're removing a redundant app queries path, and adding these two: `'libs/**/mutations/**/*.tsx?', 'libs/**/fragments/**/*.tsx?',`

brynjarorng: Ok, if it's unused it should not affect the hash and is therefore fine

brynjarorng (approved):
LGTM

libs/cms/src/lib/loaders/organizationLogo.loader.ts
snaerseljan: its because its readonly which is required when instantiating new DataLoader. 
image

Do you have a better suggestion? @eirikurn

libs/cms/src/lib/cms.contentful.service.ts
snaerseljan: What do you mean somehow? Are you thinking something else that this maybe
```
      'fields.title[in]': organizationTitles
        .map((title) => encodeURIComponent(title))
        .join(','),
```

libs/cms/src/lib/loaders/organizationLogo.loader.ts
snaerseljan: I just updated it to use type assertion instead.
```
    const organizationLogos = await this.cmsContentfulService.getOrganizationLogos(
      organizationTitles as string[],
    )
```

obmagnusson (approved):

PR #8497, opened 2022-09-23
fix(parental-leave): yes/no fix in personal allowance

karenbjorg:
User couldn't go forward if first selected yes in 'Viltu nýta persónuafsláttinn þinn?' and then selected no without selecting anything in the second question, because of validation error from the second question.

Screenshot 2022-09-23 at 15 36 47

## Checklist:


Thornvik (approved):
Looks good

PR #8503, opened 2022-09-26
chore(user-profile): Refactor user-profile verification service.

thordurhhh:
# Update verification service + update test

## What

* Add extra test for verification service
* Remove unused code for isVerified.

## Why

`isPhoneNumberVerified` and `isEmailVerified` are no longer used as we use instant verification instead of allowing the user to verify later as it previously was set up.

## Checklist:


disaerna (approved):
💯

PR #8511, opened 2022-09-26
chore(j-s): default to block access too new states

Arnarkari93:
[Asana - Snúa við blocking logic í controller guards](https://app.asana.com/0/1199153462262248/1202974314193379/f)
## What

- Default to block access to cases with a given role or institution in a given state 

## Why

- More secure if we add new institution types, user roles and case states

## Checklist:


gudjong (approved):
Nice start

PR #8513, opened 2022-09-26
refactor(j-s): default to block access for case type check

Arnarkari93:
[Asan - Snúa við blocking logic í controller guards](https://app.asana.com/0/1199153462262248/1202974314193379/f)
## What

- Use allow list instead of blocking list when checking for CaseType

## Why

- Default to block access to cases when adding new case typs, user roles and institution types

## Checklist:


oddsson (approved):
Awesome!

apps/judicial-system/backend/src/app/modules/case/filters/case.filters.spec.ts
```diff
@@ -453,51 +469,37 @@ describe('isCaseBlockedFromUser', () => {
...
-    ${CaseType.BODY_SEARCH}
-    ${CaseType.INTERNET_USAGE}
-    ${CaseType.RESTRAINING_ORDER}
-    ${CaseType.EXPULSION_FROM_HOME}
-    ${CaseType.ELECTRONIC_DATA_DISCOVERY_INVESTIGATION}
-    ${CaseType.VIDEO_RECORDING_EQUIPMENT}
-    ${CaseType.OTHER}
-  `.describe('given an accepted $type case', ({ type }) => {
-    each`
+  each([...indictmentCases, ...investigationCases]).describe(
```
gudjong, line 472: This does not seem to work. When logging out the type parameter below, I get undefined.

apps/judicial-system/backend/src/app/modules/case/filters/case.filters.ts
```diff
@@ -61,38 +61,32 @@ function isStateHiddenFromRole(
...
   forUpdate: boolean,
   institutionType?: InstitutionType,
 ): CaseType[] {
-  const blockedTypes: CaseType[] = []
-
-  if (role !== UserRole.STAFF) {
-    return blockedTypes
+  if (
+    [
+      UserRole.ADMIN,
```
gudjong, line 71: Admins should not have access to any cases. They should only handle user management. The reason they were not explicitly blocked before was that they are prevented from accessing cases through guards.

gudjong (changes requested):

apps/judicial-system/backend/src/app/modules/case/filters/case.filters.ts
Arnarkari93: Are you saying that historically we needed Admins to have access to cases and that has been addressed? So now we should remove admin here and return an empty list of allowed case types?

gudjong: No, admins have never had access, but they were blocked in guards not here. Now we are reversing the logic to give access rather than to block and should not grant any access to ADMINs.

gudjong (approved):

PR #8517, opened 2022-09-27
feat(application-system): Add dynamic application name

Thorrik58:
# Add dynamic application name

## What

Give devs the options on changing the name of the application by a function defined in the template where we conditionally return translations strings from defined logic in the function.

## Why

## Screenshots / Gifs

Default overview when an example application has been created
image

Default when in the application
image

User selects no on having worked before
image

User submits his answer no on to having worked before on the example application and the name changes within the application like so
image

Application in overview also changes the name then.
image

## Checklist:


obmagnusson (approved):
LGTM

PR #8524, opened 2022-09-28
fix(web): Remove life event tag from Digital Iceland services in search results

RunarVestmann:
# Remove life event tag from Digital Iceland services in search results

## What

* We recently reused the life event pages to write about the Digital Iceland services but we forgot to tweak the search results page (as of right now you see "Innskráning fyrir alla" tagged as a life event, which isn't quite right)

## Why

* We want to remove the tag from life event pages that are used as a Digital Iceland service since it's wrong to talk about the services as life events

### Before
image

### After
image

## Checklist:

### DIFF
--- a/apps/web/screens/Search/Search.tsx
+++ b/apps/web/screens/Search/Search.tsx
@@ -295,6 +295,16 @@ const Search: Screen<CategoryProps> = ({
     return false
   }
 
+  const getItemLink = (item: SearchType) => {
+    if (
+      item.__typename === 'LifeEventPage' &&
+      item.pageType === 'Digital Iceland Service'
+    ) {
+      return linkResolver('digitalicelandservicesdetailpage', [item.slug])
+    }
+    return linkResolver(item.__typename, item?.url ?? item.slug.split('/'))
+  }
+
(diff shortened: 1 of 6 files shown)


siggifv (approved):

apps/web/screens/Search/Search.tsx
```diff
@@ -292,14 +295,24 @@ const Search: Screen<CategoryProps> = ({
     return false
   }
 
+  const getItemLink = (item: SearchType) => {
+    if (
+      item.__typename === 'LifeEventPage' &&
+      item.pageType === 'Digital Iceland Service'
```
albinagu, line 301: wondering whether 'Life Event' and 'Digital Iceland Service' should be listed as constants?

albinagu (approved):

apps/web/screens/Search/Search.tsx
RunarVestmann: The GraphQL type for pageType is a string (instead of being 'Digital Iceland Service' | 'LifeEventPage'). But yeah having a constant might be the best middle ground :)

PR #8534, opened 2022-09-28
feat(service-portal): move feature flags to module definition

eirikurn:
## What

Refactor service portal feature flags to the module definition, using `@island.is/feature-flags` enum.

Benefits:

- Allows us to refer to specific Features enum values with TypeScript checking.
- All active features are in one place.
- More control over the name of the feature flag which enables a service portal module.
- When cleaning up a feature flag, TypeScript will tell you everywhere it's used. No string magic.
- More consistent with how other code is feature flagged (application system, API routes).

This PR also modifies our feature flag client wrappers to include the user's subjectType when getting features. This allows us to target features based on wether the user is a person or legal entity.

## Why

Sometimes we use the same feature flag to enable logic in the service portal, API (and potentially other systems). If the feature flag has "servicePortal" in the name, it sounds like it is only used in the service portal.

Recently there was an incident where someone was cleaning up one of these feature flags in the service portal but forgot to check if it was used in other places. This took down the feature in production.

With this PR we can have a broader name, and TypeScript is likely to yell if you don't remove the feature flag everywhere in the repo.

## Checklist:


apps/service-portal/src/hooks/useModules/useModules.ts
```diff
@@ -6,72 +6,37 @@ import { useFeatureFlagClient } from '@island.is/react/feature-flags'
...
+      ModuleKeys,
+      ServicePortalModule,
+    ][]
+    const filteredModules: Record<string, ServicePortalModule> = {}
+
+    for (const [moduleKey, module] of moduleEntries) {
+      let enabled = true
+
+      if (enabled && isCompany && !companyModules.includes(moduleKey)) {
+        enabled = false
```
darrikonn, line 28: redundant enabled check in the statement above, and might be better to `continue` here to skip `enabled` checks everywhere else.

darrikonn: Just a thought regarding if reduce improves readability? :
```typescript
const filteredModules = moduleEntries.reduce((acc, [moduleKey, module]) => {
  const isMissingFromCompanyModules = isCompany && !companyModules.includes(moduleKey)
  const isFeatureFlagged = !isMissingFromCompanyModules && module.featureFlag && Boolean(await featureFlagClient.getValue(module.featureFlag, false))
  if (isFeatureFlagged) {
    return {
      ...acc,
      [moduleKey]: module,
    }
  }
  return acc
}, {})
```

libs/nest/feature-flags/src/lib/feature-flag.guard.spec.ts
```diff
@@ -15,7 +15,8 @@ import {
 } from '../'
 import { GraphQLModule, Query, Resolver } from '@nestjs/graphql'
 
-const testUser = createCurrentUser()
+const testUser = createCurrentUser({ nationalIdType: 'person' })
+const testCompany = createCurrentUser({ nationalIdType: 'company' })
```
darrikonn, line 19: nice 👍 ... like these fixtures `nationalIdType`

albinagu (approved):

snaerseljan (approved):
LGTM. Great work!

apps/service-portal/src/hooks/useModules/useModules.ts
eirikurn: In general I like verbose variable names to help with readability but some of the other suggestions I disagree with:

- `continue` _could_ make sense here, but I was conditioned to avoid the keyword years ago, since it can make code harder to read or follow, similar to old `goto` and multiple returns. 🤓
- Regarding `reduce`, I agree strongly with [Jake Archibald](https://twitter.com/jaffathecake/status/1213077702300852224) on this one. I would especially like to have an eslint rule banning reduce-based object/array construction because its bad performance (`O(n^2)`) is ... at best, unnecessarily slow ... at worst, causing dangerous CPU-bottlenecks on single-threaded JS runtime running on slow mobile devices or overloaded server processes. 😬

I kept the redundant `enabled` check for consistency, idea being that a module is enabled unless any condition makes it not enabled. Sure I could implement this somehow fancier but I like the KISS principle.

Also, I must admit I stared at your suggestion for over 5 minutes to try and understand if it was correct, only to finally figure out that it's not...

<details>
<summary>Feel free to compare the implementations on your own, or expand this...</summary>

It does not return modules for normal users or company modules for companies unless the modules also have a configured featureFlag which is true.

Here's my stab to fix it in the GitHub UI. I always ended up with a ton of double negations, which are really confusing. Still not 100% sure if the logic is correct or not. 😆

```
const filteredModules = moduleEntries.reduce((acc, [moduleKey, module]) => {
  const isMissingFromCompanyModules = isCompany && !companyModules.includes(moduleKey)
  const isMissingFeatureFlag = !isMissingFromCompanyModules && module.featureFlag && !Boolean(await featureFlagClient.getValue(module.featureFlag, false))
  if (!isMissingFromCompanyModules && !isMissingFeatureFlag) {
    return {
      ...acc,
      [moduleKey]: module,
    }
  }
  return acc
}, {})
```

</details>

Thanks for taking the time, but I think I'll keep this as is. ❤️

libs/feature-flags/src/lib/features.ts
```diff
@@ -29,6 +29,11 @@ export enum Features {
 
   // Application System Delegations active
   applicationSystemDelegations = 'applicationSystemDelegations',
+
+  // Service portal modules
```
thordurhhh, line 33: Good change 👍 

One question, would it make sense to split these into separate files where each project has it's own `features.ts` file?

for exaple service-portal would go into `libs/feature-flags/src/lib/features-service-portal.ts` or some variation of that. And the codeowners for the service portal would be set as codeowners on that file as well.

My thinking is, so that we avoid always needing a code review from core just to add a feature flag string variable, and less noise for the core team as well.

thordurhhh (approved):
Nice 👍

libs/feature-flags/src/lib/features.ts
eirikurn: I don't mind the noise, but we can do it if it's causing delays or annoyances.

My main thinking is that many of the service-portal flags we're creating shouldn't just affect the UI. They should be checked on the backend as well, eg for data integrations which are "not ready". Putting them in a "service-portal-features" file or having service-portal in the name of the field can cause mistakes in this regard.

thordurhhh: That makes sense, I don't think we're adding too many feature flagged modules anyways. 💯

PR #8493, opened 2022-09-23
fix(syslumenn-client): Improve Asset fetching

kristofer-lp8f:
- Add `IsAssetId` property check
- Improve code legibility
- Improve documentation for `filterFields`

## Checklist:

- [ ] I have performed a self-review of my own code
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] Formatting passes locally with my changes
- [ ] I have rebased against main before asking for a review

PR #8538, opened 2022-09-29
fix(propertyDetail): Add guard on which fields to return

kristofer-lp8f:
- Add filtering for which fields to return
- Add test for ownership

Fork from #8493 

## Checklist:


thordurhhh:
Nice! 🤝 

So the idea is that because the values in the field array are public information, that can be returned even if the requester is not the owner?
`fields: ['defaultAddress', 'land', 'propertyNumber']`

kristofer-lp8f:
> So the idea is that because the values in the field array are public information, that can be returned even if the requester is not the owner? `fields: ['defaultAddress', 'land', 'propertyNumber']`

Yes, exactly. In the original PR (#8493) I plan to add good documentation on how to use the `filterFields` decorator. This current change should prevent future data leakage by accidentally over-exposing non-public field.

PR #8566, opened 2022-10-03
feat(search-indexer): Add env variable for chunk size of contentful fetch

RunarVestmann:
# Add env variable for chunk size of contentful fetch

## What

* The search-indexer is calling Contentful but getting back a "Response too big" response
* Changing the chunk size of the fetch locally helped with indexing so having an env variable to change in the deployed environment might not be such a bad idea

## Checklist:


siggifv (approved):

pshomov (approved):
LGTM

PR #8584, opened 2022-10-04
chore(search-indexer): Change default chunk size for contentful entry fetch

RunarVestmann:
# Change default chunk size for contentful entry fetch

* Changed default value due to `Response size too big` error occurring during indexing.


Ingimarsson (approved):

pshomov (approved):
LGTM

PR #8586, opened 2022-10-04
feat(financial-statements-inao): Submit audit data

HelgiHH:
## What

Submission for individual elections statement 

## Checklist:


libs/application/template-api-modules/src/lib/modules/templates/financial-statements-inao/financial-statements-inao.service.ts
```diff
@@ -1,22 +1,178 @@
...
+        'election.incomeLimit',
+      ) as string
+
+      const noValueStatement = electionIncomeLimit === LESS ? true : false
+
+      // actor is undefined until we add the delegate functionality to the frontend
+      const result: DataResponse = await this.financialStatementsClientService
+        .postFinancialStatementForPersonalElection(
+          nationalId,
+          undefined,
```
fjandakornid, line 78: The actor should also come from the auth object... and should be undefined if auth.actor.nationalId is undefined

libs/application/template-api-modules/src/lib/modules/templates/financial-statements-inao/financial-statements-inao.service.ts
```diff
@@ -1,22 +1,178 @@
...
+        )
+        .then((data) => {
+          if (data === true) {
+            return { success: true }
+          } else {
+            return { success: false }
+          }
+        })
+        .catch((e) => {
+          console.error('dataverse error', e)
```
fjandakornid, line 92: console.log... should be able to import logger and use it instead

fjandakornid (changes requested):
Few change requests

fjandakornid (approved):
LGTM

PR #8594, opened 2022-10-05
feat(web): RealEstateAgentsList and LawyersList - Connected components

siggifv:
# RealEstateAgentsList connected component

## What

Implements renderers for Contentful Connected Components (components renderable within rich text values from Contentful), for listing **Real Estate Agents** and **Lawyers**, fetched from Syslumenn API.
Note that currently, these two lists look very much alike, but I've decided to keep the separate in code (i.e. not to create a parent component that these to inherit) because they are likely to develop differently in the near future.

## Why

Requested by Syslumenn.

## Screenshots / Gifs

Screenshot 2022-10-05 at 11 18 54

Screenshot 2022-10-09 at 16 57 42

## Checklist:


oddsson (approved):

albinagu (approved):
lgtm

RunarVestmann (approved):

libs/shared/connected/src/utils/sortAlpha.ts
```diff
@@ -0,0 +1,47 @@
+/**
+ * NOTE:
+ *  The contents of this file are the same as in the file apps/web/utils/sortAlpha.ts
+ *  We cannot import from that file, as it will create circular dependency.
+ *  Therefor, at least for now, we copy this code here. Ideally, this code would be
+ *  moved into a dedicated utils library.
+ */
```
eirikurn, line 7: Should move this to `libs/shared/utils` as a first step.

Though this shared/connected library (and way more "web"-related code) should be in `libs/web/*` instead of under shared, since they won't really be used by any other projects.

eirikurn (approved):
Some comments. Feel free to address or refactor in follow-up.

PR #8601, opened 2022-10-05
feat(j-s): Case Files PDF

gudjong:
# Case Files PDF

[Velja tól til að setja saman gagnapakka](https://app.asana.com/0/1199153462262248/1203089836240621/f)

## What

- Implements an HTTP endpoint that generates a case files pdf for a case.

## Why

Proof of concept.

## Checklist:


Arnarkari93 (approved):

package.json
```diff
@@ -212,6 +212,7 @@
     "opossum": "6.1.0",
     "passport": "0.4.1",
     "passport-jwt": "4.0.0",
+    "pdf-lib": "1.17.1",
     "pdfkit": "0.11.0",
```
eirikurn, line 216: I see you are already using pdfkit a lot in j-s, so why add a new pdf generation tool for this?

Generally we should favour one "tool" for each job. If there are good arguments for switching tools, it should be shared with the discipline and over time it should replace the old tool.

Personally, I dream of island.is creating its own pdf generation API powered by headless chrome and "@island.is/island-ui/print" components. Took less then a sprint in another project we worked on.

gudjong: In this case, the old tool cannot handle the task we are trying to accomplish. We are working on indictments for criminal cases and need to merge a number of existing pdf documents into one and add a table of contents with "clickable" links to the merged document, plus some other things. We plan to replace the old library with the new one and this PR is the first step in that journey. Replacing the old library is a large task and we would like to accomplish that gradually.

eirikurn: Ok.

If `pdf-lib` is functionally equivalent or better than `pdfkit`, then please write a short ADR about this in `handbook/technical-overview/adr` and share it in #dev-faghopur. Basically your findings from comparing the two. That way we can document our preference for all teams and hopefully replace the old library from all projects at some point.

Otherwise, approving this PR to unblock your work. 👍

eirikurn (approved):

PR #8621, opened 2022-10-06
fix(j-s): S-Case Court Documents

gudjong:
# S-Case Court Documents

[Ekki vista þingbók og dóm í Auði í S-málum](https://app.asana.com/0/1199153462262248/1203114095779785/f)

## What

- Stops uploading the court record and ruling to court for S-cases.
- Stops calling file service to get all case files as the case files are already included in the case.
- Uses `findByPk` where possible.

## Why

Verified bug and useful rewrites.

## Checklist:


Arnarkari93 (approved):
LGTM!

PR #8630, opened 2022-10-07
feat(auth): prepare supporting custom delegations for the delegation scope

eirikurn:
## What

Get rid of the read scope, just use write instead.
Get rid of a manual UI check, just check the scope.

## Why

So procuring holders can give employees access to the access control page.

## Checklist:


saevarma (approved):
💯

thordurhhh (approved):
✅

PR #8634, opened 2022-10-10
refactor(service-portal): Documents fetched only on click

disaerna:
# Service portal - Documents fetch fix

## What

To check if document is HTML, I previously fetched every document on load (only first page though) to read the filetype. 
Better solution is clearly to only fetch document if user clicks on it! 💯 

## Why

To avoid bunch of unnecessary calls to Advania 👍 

## Screenshots / Gifs

## Checklist:


thordurhhh (approved):
🎉

PR #8643, opened 2022-10-10
fix(auth-api): Set ApiScope include to be required in findAllValidCustomScopesTo

saevarma:
## What

After removing identity resources from delegation scopes we need to set ApiScopes to be required in `findAllValidCustomScopesTo`.

## Why

To prevent errors when scopes are no longer supports delegations or are expired.

## Screenshots / Gifs

## Checklist:


eirikurn (approved):
LGTM

PR #8644, opened 2022-10-10
feat(auth): Support delegations one level deep

eirikurn:
## What

- Remove extra checks in service-portal.
- Filter out actor delegation.
- Block POST/PUT on delegation to actor.

## Why

Requested functionality.

For now we hide and disable any edits to the actor delegation since it has some unexpected edge cases with the delegation scope and the user's active access token.

## Checklist:

### DIFF
--- a/libs/auth-api-lib/src/lib/delegations/delegations.service.ts
+++ b/libs/auth-api-lib/src/lib/delegations/delegations.service.ts
@@ -92,7 +92,10 @@ export class DelegationsService {
     user: User,
     createDelegation: CreateDelegationDTO,
   ): Promise<DelegationDTO | null> {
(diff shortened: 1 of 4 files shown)


libs/auth-api-lib/src/lib/delegations/delegations.service.ts
```diff
@@ -309,6 +321,13 @@ export class DelegationsService {
       })
   }
 
+  private isDelegationToActor(
```
saevarma, line 324: Great rename, this reads better 💯

saevarma (approved):
🎉

thordurhhh (approved):

PR #8646, opened 2022-10-11
fix(j-s): Defender Resubmit Notifications

gudjong:
# Defender Resubmit Notifications

[Verjandi getur fengið póst um endursenda kröfu áður en hann fær fyrirtökupóst](https://app.asana.com/0/1199153462262248/1202978837144050/f)

## What

- Does not notify the defender of the resubmission of a request to court unless the defender has previously been notified of a court date.
- Rewrites the resubmit notification received by the defender, identifying the case with the court case number rather than the police case number and only providing a link to Réttarvörslugátt if the defender has access (i.e. the defender's national id is recorded in the case).

## Why

Verified bug.

### Notification when defender national is is missing:
image

### Notification when defender national id is known:
image

## Checklist:


apps/judicial-system/backend/src/app/messages/notifications.ts
```diff
@@ -122,19 +122,27 @@ export const notifications = {
   defenderResubmittedToCourt: defineMessages({
     subject: {
       id:
-        'judicial.system.backend:notifications.defender_resubmitted_to_court.subject_v2',
-      defaultMessage: 'Krafa um {caseType} send aftur',
+        'judicial.system.backend:notifications.defender_resubmitted_to_court.subject',
```
Arnarkari93: Should this be a new id? You mentioned that strings are not removed from contentful and we need to make new ids each time.
```suggestion
        'judicial.system.backend:notifications.defender_resubmitted_to_court.subject_v3',
```

apps/judicial-system/backend/src/app/messages/notifications.ts
```diff
@@ -122,19 +122,27 @@ export const notifications = {
...
+      defaultMessage: 'Gögn í máli {courtCaseNumber}',
       description:
         'Notaður sem titil í pósti til verjanda þegar sækjandi breytir kröfunni og sendir aftur á héraðsdómstól',
     },
     body: {
       id:
-        'judicial.system.backend:notifications.defender_resubmitted_to_court.body_v2',
+        'judicial.system.backend:notifications.defender_resubmitted_to_court.body',
```
Arnarkari93:
```suggestion
        'judicial.system.backend:notifications.defender_resubmitted_to_court.body_v3',
```

apps/judicial-system/backend/src/app/modules/notification/notification.service.ts
```diff
@@ -439,6 +438,35 @@ export class NotificationService {
...
+    try {
+      const notifications = await this.notificationModel.findAll({
+        where: {
+          caseId: theCase.id,
+          type: NotificationType.COURT_DATE,
+        },
+      })
+
+      exists = notifications.some(async (notification) => {
+        JSON.parse(notification.recipients ?? '[]').some(
```
Arnarkari93, line 454: Looks like there is a missing return here
```suggestion
        return JSON.parse(notification.recipients ?? '[]').some(
```

apps/judicial-system/backend/src/app/messages/notifications.ts
oddsson, line 125: Does this not need to be `subject_v3`, based on what you were saying earlier about CF storing ids indefinitely?

oddsson (commented):
One question

apps/judicial-system/backend/src/app/modules/notification/test/createTestingNotificationModule.ts
```diff
@@ -36,7 +36,14 @@ export const createTestingNotificationModule = async () => {
     ],
     controllers: [NotificationController, InternalNotificationController],
     providers: [
-      CourtService,
+      {
+        provide: CourtService,
+        useValue: {
+          createDocument: jest.fn(async () => {
```
Arnarkari93, line 42: Do we need the mock implementation here? Is `jest.fn()` not enough?

apps/judicial-system/backend/src/app/messages/notifications.ts
gudjong: Yes

apps/judicial-system/backend/src/app/modules/notification/test/notificationController/sendReadyForCourtNotifications.spec.ts
```diff
@@ -0,0 +1,233 @@
...
+    })
+
+    it('should send ready for court email notification to prosecutor', () => {
+      expect(mockEmailService.sendEmail).toHaveBeenCalledWith({
+        from: { name: 'Réttarvörslugátt', address: 'ben10@omnitrix.is' },
+        replyTo: { name: 'Réttarvörslugátt', address: 'ben10@omnitrix.is' },
+        to: [{ name: 'Derrick', address: 'derrick@dummy.is' }],
+        subject: 'Krafa um gæsluvarðhald send',
+        text: `Þú hefur sent kröfu á Héraðsdóm Reykjavíkur vegna LÖKE máls ${policeCaseNumber}. Skjalið er aðgengilegt undir málinu í Réttarvörslugátt.`,
+        html: `Þú hefur sent kröfu á Héraðsdóm Reykjavíkur vegna LÖKE máls ${policeCaseNumber}. Skjalið er aðgengilegt undir <a href="http://localhost:4200/krafa/stadfesta/${caseId}">málinu í Réttarvörslugátt</a>.`,
```
Arnarkari93, line 120: Consider taking `http://localhost:4200` from the config.

Arnarkari93 (changes requested):
Looks good! Just some minor points from me

apps/judicial-system/backend/src/app/messages/notifications.ts
gudjong: Already fixed

apps/judicial-system/backend/src/app/messages/notifications.ts
gudjong: Also fixed

apps/judicial-system/backend/src/app/modules/notification/notification.service.ts
gudjong: Good catch. However, the code works as is, probably because javascript returns the result of the last expression in the block. Will fix.

apps/judicial-system/backend/src/app/modules/notification/test/createTestingNotificationModule.ts
gudjong: You are right, since we are not promise chaining call to createDocument, we do not need the async mock.

apps/judicial-system/backend/src/app/modules/notification/test/notificationController/sendReadyForCourtNotifications.spec.ts
gudjong: I guess the test would be more robust by simply referring to the config rather than hard coding the text. I will make that change.

This will not remove `http://localhost:4200` from the config however.

gudjong: Done

Arnarkari93 (approved):

PR #8661, opened 2022-10-11
chore(j-s): Add fields to case_file table

oddsson:
# Add fields to case_file table

[Asana](https://app.asana.com/0/1199153462262248/1203127878754336/f)

## What

Add fields `policeCaseNumber`, `userGeneratedFilename`, `chapter`, and `orderWithinChapter` to the case_file table in DB.

## Why

This is needed for the caseFile feature we are building.

## Checklist:


apps/judicial-system/backend/migrations/20221011135323-update-case.js
```diff
@@ -0,0 +1,73 @@
...
+          },
+          {
+            transaction,
+          },
+        ),
+      ]),
+    )
+  },
+
+  down: (queryInterface, Sequelize) => {
```
gudjong, line 55: Consider removing `Sequelize` as it is not used

gudjong (changes requested):

libs/judicial-system/types/src/lib/file.ts
```diff
@@ -63,6 +63,10 @@ export interface CaseFile {
   key?: string
   size: number
   category?: CaseFileCategory
+  policeCaseNumber?: string
+  userGeneratedFilename?: string
+  chapter?: number
+  orderWithinChapter?: number
 }
 
 export interface CreateFile {
```
Arnarkari93, line 72: `policeCaseNumber` missing here

Arnarkari93 (commented):
Looks good, I think here is only one place left.

Arnarkari93 (approved):

libs/judicial-system/types/src/lib/file.ts
```diff
@@ -63,10 +63,17 @@ export interface CaseFile {
...
+  caseId: string
```
Arnarkari93, line 73: caseId is not part of the CreateFile.dto. It can be removed here. 
```suggestion
```

gudjong (approved):

PR #8679, opened 2022-10-13
feat(auth): New GraphQL schema for DelegationsV2

eirikurn:
## What

- Feature flagged `authDomains` query with mocked data.
- Feature flagged `authScopeTree` query.
- Feature flagged `patchAuthDelegation` mutation.
- Feature flagged `domain` field resolver on delegation (allows querying translated domain information directly in delegation list).
- New `domain` filter to `delegations` query.
- New `domainName` input to `createDelegation` mutation.
- New `apiScope` field resolver on delegationScope (allows querying translated scope and group names directly in delegation lists).
- Using `dataloader` for batched loading in `domain` and `apiScope` field resolvers.
- Forked [`nestjs-dataloader`](https://www.npmjs.com/package/nestjs-dataloader) as `@island.is/nest/dataloader` with GraphQL context support and correct types.
- Wrapped current MeDelegationService and ApiScopeService with feature flagging to support v2 services.
- Update current access control interface to pass `@island.is` as domain in queries/mutations for forward compatibility with the new API.

## Why

These changes are needed for DelegationsV2 functionality which we're implementing in a new UI and REST API. Note that we're not versioning the GraphQL schema. All of the changes are designed to be backwards and forwards compatible, with only the GraphQL service layer being feature flagged to integrate with different REST backends.

The new schemas in this PR are only implemented against mock data and the current REST service (auth-public-api). This unblocks UI integration work. We'll integrate the new REST service in a follow up PR when it is ready.

## Checklist:


libs/api/domains/auth/src/lib/resolvers/delegationScope.resolver.ts
```diff
@@ -15,4 +18,18 @@ export class DelegationScopeResolver {
   resolveName(@Parent() delegationScope: DelegationScopeDTO): string {
     return delegationScope.scopeName
   }
+
+  @ResolveField('apiScope', () => ApiScope, { nullable: true })
+  resolveApiScope(
+    @Loader(ApiScopeLoader) apiScopeLoader: ApiScopeDataLoader,
+    @Parent() delegationScope: DelegationScope,
+    @Args('lang', { type: () => String, nullable: true, defaultValue: 'is' })
```
snaerseljan, line 26: I realize it might be a big refactor change throughout the project, but maybe in the future it would be nice to have typed language codes

libs/api/domains/auth/src/lib/services/types.ts
```diff
@@ -0,0 +1,88 @@
...
+  DeleteDelegationInput,
+  PatchDelegationInput,
+  UpdateDelegationInput,
+} from '../dto'
+import { DelegationByOtherUserInput } from '../dto/delegationByOtherUser.input'
+import { ApiScope } from '../models'
+import { ScopeTreeNode } from '../models/scopeTreeNode.model'
+import { ApiScopesInput } from '../dto/apiScopes.input'
+
+export interface MeDelegationsServiceI {
```
snaerseljan, line 15: Why ending with I? Version 1?

snaerseljan (approved):
This looks great. Well done!

libs/api/domains/auth/src/lib/resolvers/delegationScope.resolver.ts
eirikurn: Agree. I think we have it in some places. Just kept it like this for internal consistency.

libs/api/domains/auth/src/lib/models/scopeTreeNode.model.ts
```diff
@@ -0,0 +1,8 @@
+import { createUnionType } from '@nestjs/graphql'
+import { ApiScopeGroup } from './apiScopeGroup.model'
+import { ApiScope } from './apiScope.model'
+
+export const ScopeTreeNode = createUnionType({
+  name: 'AuthScopeTreeNode',
+  types: () => [ApiScope, ApiScopeGroup],
```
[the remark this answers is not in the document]
eirikurn: Yeah I started with `as const` like in the docs, but TypeScript doesn't allow it for some reason...

image

Might be an issue with our TypeScript or NestJS version. Haven't experienced any problems with this so far. Maybe since we're not using this between projects.

libs/api/domains/auth/src/lib/models/delegation.model.ts
```diff
@@ -78,4 +78,6 @@ export class CustomDelegation extends Delegation {
 
   @Field(() => [DelegationScope])
   scopes!: DelegationScope[]
+
+  domainName?: string
```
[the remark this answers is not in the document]
eirikurn: I need this property in our field resolver but I don't want to expose it in our GraphQL schema. I'll put a comment explaining it.

libs/api/domains/auth/src/lib/models/scopeTreeNode.model.ts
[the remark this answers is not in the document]
snaerseljan: Yes seams to be versioning problem
https://github.com/nestjs/graphql/issues/2122

libs/api/domains/auth/src/lib/dto/apiScope.input.ts
```diff
@@ -0,0 +1,13 @@
+import { Field, InputType } from '@nestjs/graphql'
+
+@InputType('AuthApiScopeInput')
+export class ApiScopeInput {
+  @Field(() => String)
+  lang = 'is'
```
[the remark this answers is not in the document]
eirikurn: Not in this project, no. I feel there are pros and cons for having strict locales. Not great to require many code changes to support a new language, when we have the capability to dynamically load translations from the IDS and Contentful.

But I guess that's still a rare event for Island.is.

libs/api/domains/auth/src/lib/services/types.ts
snaerseljan: I think it is redundant to use prefix or suffix on interfaces with ´I´. Thats more the C# way. 
We get proper type and context from things like intellesense and language servers.

But in the end it is just a preference :)

eirikurn: Yeah. I agree, that's what I meant and what I strive for, but in this case I couldn't really find a way around it.

This is my problem:

```
// The interface
interface MeDelegationsService {}

// The main wrapper service
class MeDelegationsService implements MeDelegationsService {}

// Feature flagged service.
class MeDelegationsServiceV1 implements MeDelegationsService {}
class MeDelegationsServiceV2 implements MeDelegationsService {}
```

The interface and the main service can't have the same name in the same file, so I need to differentiate them somehow. Elsewhere, I want to use dependency injection to inject "MeDelegationsService", the actual class instead of the interface. And I kind of do not want to name it something like "MeDelegationsServiceWrapper". So it is better to rename the interface ... and "Interface" is the logical differentiator.

"The exception makes the rule"? Not happy about this, but here we are :D

libs/api/domains/auth/src/lib/resolvers/delegationScope.resolver.ts
```diff
@@ -15,4 +18,18 @@ export class DelegationScopeResolver {
...
+    lang: string,
+  ): Promise<ApiScope | null> {
+    return apiScopeLoader.load({
+      lang,
+      domain: delegationScope.domainName,
```
saevarma, line 31: Same argument about domainName being unnecessary on apiScope, as the scopeName is unique primary key.

libs/api/domains/auth/src/lib/services/apiScope.service.ts
```diff
@@ -0,0 +1,41 @@
...
+  ) {}
+
+  private async service(user: User): Promise<ApiScopeServiceI> {
+    const newDelegations = await this.featureFlagService.getValue(
+      Features.outgoingDelegationsV2,
+      false,
+      user,
+    )
+    // TODO: Implement ApiScopeServiceV2.
+    return newDelegations ? this.apiScopeServiceV1 : this.apiScopeServiceV1
```
saevarma, line 24: Nice versioning based on feature flag 💯

libs/api/domains/auth/src/lib/services/types.ts
```diff
@@ -0,0 +1,88 @@
...
+    input: UpdateDelegationInput,
+  ): Promise<DelegationDTO>
+
+  patchDelegation(
+    user: User,
+    input: PatchDelegationInput,
+  ): Promise<DelegationDTO>
+}
+
+export interface ApiScopeServiceI {
```
saevarma, line 46: Would prefer `ApiScopeServiceInterface` over `I` postfix. And maybe as a general naming convention when we are actually using interface as a service contract instead of data modeling.

libs/service-portal/settings/access-control/src/constants.ts
```diff
@@ -0,0 +1,3 @@
+// To keep this deprecated UI future compatible with graphql schema changes, we're
+// hard-coding the domain it should manage.
+export const ISLAND_DOMAIN = '@island.is'
```
saevarma, line 3: You are defining the same constant in two different libraries. Put it in a shared constant library?

saevarma (commented):
Nice job, I'm way too late just few comments on possible follow up improvements.

libs/api/domains/auth/src/lib/services/types.ts
eirikurn: Agreed!

libs/service-portal/settings/access-control/src/constants.ts
eirikurn: Good idea. Didn't want to create a new library but I see we already have a shared constant library.

libs/api/domains/auth/src/lib/resolvers/delegationScope.resolver.ts
eirikurn: We haven't planned a global scope REST resource. The API is domain scoped at the moment, and my instinct is to batch load them by domain instead of batch loading all scopes from all domains. Unless we create a custom batch load endpoint.

But this is basically implemented based on the current APIs and plans.

libs/api/domains/auth/src/lib/dto/apiScope.input.ts
```diff
@@ -0,0 +1,13 @@
...
+
+  @Field(() => String)
+  name!: string
+
+  @Field(() => String, { nullable: true })
+  domain?: string
```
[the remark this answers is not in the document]
eirikurn: See my other comment about how scopes are domain scoped at the moment :D

But this query is actually not exposed in GraphQL at the moment. It is basically a DTO for the service layer used by dataloader, so I should probably have skipped all the decorators in this case.

PR #8682, opened 2022-10-13
feat(j-s): bulk case file updates

Arnarkari93:
[Asana - Skjalaskrá - ákærandi raðar skjölum í kafla](https://app.asana.com/0/1199153462262248/1203046355378030/f)

## What

- Add endpoint to bulk update case files

## Why

- So we can update the order and chapter for case files
- So we can update the user-generated name for a file/files

## Checklist:


oddsson (approved):

apps/judicial-system/api/src/app/data-sources/backend.ts
```diff
@@ -93,6 +95,14 @@ export class BackendApi extends DataSource<{ req: Request }> {
     })
   }
 
+  private patch<TBody, TResult>(route: string, body: TBody): Promise<TResult> {
```
gudjong, line 98: Nice. in fact, all of our updates should use PATCH rather than PUT, but that is a different issue.

gudjong (approved):
Look really good.

PR #8739, opened 2022-10-20
feat(j-s): Case Files PDF

gudjong:
# Case Files PDF

[Smíða einn PDF gagnapakka úr málsgögnum](https://app.asana.com/0/1199153462262248/1203046355378049/f)

## What

- Generates a PDF document from case files in s-cases.

## Why

## Screenshots / Gifs

https://user-images.githubusercontent.com/795382/197073680-167150de-a490-4181-9093-69d7fcdce7ac.mov

## Checklist:


apps/judicial-system/backend/src/app/formatters/caseFilesRecordPdf.ts
```diff
@@ -0,0 +1,177 @@
...
+} from '@island.is/judicial-system/formatters'
+
+import { caseFilesRecord } from '../messages'
+import { Defendant } from '../modules/defendant'
+import { Case } from '../modules/case'
+import { Alignment, PageLink, PdfDocument } from './pdf'
+
+function formatDefendant(defendant: Defendant) {
+  let nationalId = ''
+  if (defendant.noNationalId) {
```
oddsson, line 15: Isn't this a bit strange? You first check for noNationalId and then nationalId. How can nationalId be set if noNationalId is set?

apps/judicial-system/backend/src/app/formatters/caseFilesRecordPdf.ts
```diff
@@ -0,0 +1,177 @@
...
+    .addText(capitalize(caseTypes[theCase.type]), textFontSize, {
+      position: { x: defendantIndent },
+    })
+    .addText(
+      formatMessage(caseFilesRecord.tableOfContentsHeading),
+      subtitleFontSize,
+      { alignment: Alignment.Center, bold: true, marginTop: 9 },
+    )
+
+  for (const chapter of [0, 1, 2, 3, 4, 5]) {
```
oddsson, line 137: Consider defining constants or an enum for the chapters for better readability

apps/judicial-system/backend/src/app/formatters/caseFilesRecordPdf.ts
gudjong: `nationalId` is also used for date of birth, so it can be set whether `noNationalId` is `true` or `false`. However, date of birth is optional.

oddsson (approved):

Arnarkari93 (approved):
LGTM

PR #8742, opened 2022-10-20
fix(service-portal): Update html lang

thordurhhh:
## What

Update `lang` attribute to account for the current language.

## Why

Even though we do not need to account for SEO in the service portal..

We have some comments regarding the `<html lang="">` tag on the service-portal. It can be confusing for some assisted technologies like screen readers to go to a page in one language when the lang attr is showing a different language.

Details here: https://accessibility.psu.edu/foreignlanguages/langtaghtml/

## Checklist:


disaerna (approved):
Francisko Corto 🤭

eirikurn (approved):

PR #8753, opened 2022-10-20
feat(auth): GraphQL integration for DelegationsV2

eirikurn:
## What

- Integrate with `services-auth-delegation-api` as v2 of the delegation domain.
- Refactor auth clients into client subfolder.
- Use server feature flag while we don't have production env values.

## Why

Allows us to manage delegation across multiple systems.

## Checklist:


charts/identity-server/values.dev.yaml
```diff
@@ -6,7 +6,7 @@ auth-admin-web:
     NEXTAUTH_URL: 'https://identity-server.dev01.devland.is/admin/api/auth'
     NEXT_PUBLIC_BACKEND_URL: '/backend'
     NODE_OPTIONS: '--max-old-space-size=208'
-    SERVERSIDE_FEATURES_ON: ''
+    SERVERSIDE_FEATURES_ON: 'delegation-api'
```
snaerseljan, line 9: Out of curiosity what how do you add more features to SERVERSIDE_FEATURES_ON?
Is it a comma separated list or an array or just not supported?

libs/api/domains/auth/src/lib/services-v2/meDelegations.service.ts
```diff
@@ -0,0 +1,170 @@
...
+    if (!delegation) {
+      throw new NotFoundException()
+    }
+    const oldScopes =
+      delegation.scopes?.map((scope) => ({
+        name: scope.scopeName,
+        validTo: scope.validTo,
+      })) ?? []
+
+    const updateScopes = differenceWith(
```
snaerseljan, line 122: Nice and clean these compares 🥇

snaerseljan (approved):
Awesome work!

charts/identity-server/values.dev.yaml
eirikurn: This is generated automatically from here: https://github.com/island-is/island.is/pull/8753/files/c017d3ccf78585496a0b7351f495a6a23fff102d#diff-f7be9b6be4760aab04ee07d7a80b2ded5b8338cbdb39c4e77ed852749b7c8a4e

Note that there it is an array so I assume it gets converted to comma seperated string or something like that.

libs/api/domains/auth/src/lib/services-v2/meDelegations.service.ts
```diff
@@ -0,0 +1,170 @@
...
+
+  async getDelegationByOtherUser(
+    user: User,
+    { toNationalId, domain }: DelegationByOtherUserInput,
+  ): Promise<DelegationDTO | null> {
+    const delegations = await this.delegationsApiWithAuth(
+      user,
+    ).meDelegationsControllerFindAll({
+      direction: MeDelegationsControllerFindAllDirectionEnum.Outgoing,
+      xQUERYOTHERUSER: toNationalId,
```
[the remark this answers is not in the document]
eirikurn: This is generated like this from OpenAPI "X-QUERY-OTHER-USER". Though I guess it's worth trying to register the header with different casing (eg "X-Query-Other-User"). It shouldn't matter when accessing it since that's case insensitive.

PR #8792, opened 2022-10-25
feat(j-s): change notification recipients to json array

Arnarkari93:
[Asana - Svæði fyrir viðtakendur tilkynninga í gagnagrunni er of lítið](https://app.asana.com/0/1199153462262248/1203150123978292/f)

## What

- migrate column recipients to json[] in notification table
- use type for notification recipients instead of json-string

## Why

- Column `recipients` was to small (varchar(255))

## Checklist:


apps/judicial-system/backend/migrations/20221020110402-update-notification.js
```diff
@@ -0,0 +1,118 @@
...
+            { transaction },
+          )
+          .catch((err) => {
+            console.error(
+              'Error adding column recipients_json to table notification',
+              err,
+            )
+            throw err
+          }),
+        queryInterface.sequelize
```
gudjong, line 28: It would be safer to chain the database updates with `then` to make sure they are executed in the right order.

apps/judicial-system/backend/migrations/20221020110402-update-notification.js
```diff
@@ -0,0 +1,118 @@
...
+      Promise.all([
+        // temp column for rollback
+        queryInterface.addColumn(
+          'notification',
+          'recipients_json_string',
+          { type: Sequelize.STRING },
+          { transaction },
+        ),
+        // copy data from new column to temp column
+        queryInterface.sequelize
```
gudjong, line 90: It would be safer to chain the database updates with `then` to make sure they are executed in the right order.

gudjong (changes requested):

PR #8822, opened 2022-10-27
fix(marriage-conditions): Final fixes & touchups

albinagu:
## Checklist:


libs/clients/national-registry/v2/src/lib/types/religion.dto.ts
```diff
@@ -0,0 +1,46 @@
+import { Kodar } from '../../../gen/fetch'
+
+export interface ReligionDto {
+  code: string
+  name: string | null
+}
+
+const order =
```
eirikurn, line 8: There is similar code in multiple places in the repo now. Please move/refactor this to libs/shared/utils.

libs/api/domains/national-registry-x-road/src/lib/nationalRegistryKeysXRoad.resolver.ts
```diff
@@ -0,0 +1,34 @@
...
+  IdsUserGuard,
+  Scopes,
+  ScopesGuard,
+} from '@island.is/auth-nest-tools'
+import { Audit } from '@island.is/nest/audit'
+
+import { NationalRegistryXRoadService } from './nationalRegistryXRoad.service'
+import { NationalRegistryReligion } from '../models/nationalRegistryReligion.model'
+
+@UseGuards(IdsAuthGuard, IdsUserGuard, ScopesGuard)
```
eirikurn, line 15: Should not have both IdsAuthGuard and IdsUserGuard. Should generally use IdsUserGuard in our graphql api.

eirikurn: In fact, let's skip all of the guards and the Scopes decorator since we're not returning user data in this resolver.

eirikurn (changes requested):

libs/clients/national-registry/v2/src/lib/types/religion.dto.ts
```diff
@@ -0,0 +1,46 @@
...
+  '0123456789aAáÁbBcCdDeEéÉfFgGhHiIíÍjJkKlLmMnNoOóÓpPqQrRsStTuUúÚvVwWxXyYýÝzZþÞæÆöÖ'
+
+const charOrder = (a: string) => {
+  const ix = order.indexOf(a)
+  return ix === -1 ? a.charCodeAt(0) + order.length : ix
+}
+
+const sortIsAlphabetic = (a: ReligionDto, b: ReligionDto) => {
```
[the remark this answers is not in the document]
albinagu: neeeeh, but will do what Eiki says :)

libs/shared/utils/src/lib/sortAlpha.ts
```diff
@@ -2,7 +2,7 @@
  * NOTE:
  *  The contents of this file are the same as in the file apps/web/utils/sortAlpha.ts
  *  We cannot import from that file, as it will create circular dependency.
- *  Therefor, at least for now, we copy this code here. Ideally, this code would be
+ *  Therefore, at least for now, we copy this code here. Ideally, this code would be
```
eirikurn, line 5: Yay! Let's get rid of this comment though :)

eirikurn (approved):

kksteini (approved):

PR #8831, opened 2022-10-31
feat(auth): Add an internal procuring scope

programm-ingovals:
## What

Adds an enum for the new scope
Requests the new scope in the application client
Allows the scope in FinancialStatementInao applications api
Allows the scope in NoDebt request in domain api
Remove the NoDebt scope for being requested by application client
Seed data for the new scope

## Why

There is a need for an scope similar to the internal scope but for Procuring holders.
This would allow delegation holders access to subsystems like applications conditionally.
There has already been implemented a custom scope in one such instance, the no-debt-certificate scope.
This could replace that scope and used in new instances f.e. the financial statement Inao application.

## Checklist:


programm-ingovals:
Should we remove NoDebtCertificate scope now, or assign it to that code owner to ensure we aren't messing with anything?

libs/auth-api-lib/seeders/data/scope-internal-procuring.ts
```diff
@@ -0,0 +1,17 @@
+import { createScope } from './helpers'
+
+export const up = createScope({
+  // Required:
+  name: '@island.is/internal:procuring',
+  displayName: 'Generic internal API scope for procuring holders',
+  description:
+    'Shared scope for interal usage in cases of delegation for procuring holders',
```
eirikurn, line 8:
```suggestion
    'Shared scope for internal usage in cases of delegation for procuring holders',
```

libs/auth-api-lib/seeders/data/scope-internal-procuring.ts
```diff
@@ -0,0 +1,17 @@
...
+
+  // Optional:
+  delegation: {
+    procuringHolders: true,
+  },
+  accessControlled: false,
+  addToResource: '@island.is',
+  addToClients: ['@island.is/clients/application-system'],
```
eirikurn, line 16: This is probably supposed to be `"@island.is/web"`, which is used in both service-portal and application-system-form.

```suggestion
  addToClients: ['@island.is/web'],
```

eirikurn (commented):
Love it!

libs/auth-api-lib/seeders/data/scope-internal-procuring.ts
programm-ingovals: Ok, the other client seems to exist on DEV though. Maybe some legacy we could remove.

Thorrik58 (approved):

fjandakornid (approved):
LGTM

apps/application-system/form/src/auth.ts
```diff
@@ -35,13 +35,13 @@ if (userMocked) {
...
       EndorsementsScope.main,
       ApiScope.internal,
+      ApiScope.internalProcuring,
       ApiScope.meDetails,
       ApiScope.fishingLicense,
       MunicipalitiesFinancialAidScope.read,
       MunicipalitiesFinancialAidScope.write,
       MunicipalitiesFinancialAidScope.applicant,
       ApiScope.assets,
-      ApiScope.noDebtCertificate,
```
disaerna, line 44: Is this scope being used anywhere else? If not, can it be deleted?

disaerna (approved):
Looks good to me, just one question 😊

apps/application-system/form/src/auth.ts
programm-ingovals: It is not being used elsewhere, plan is to delete from IDS in the future, but keep it there until after this changes has gone into effect. Will make reverting these changes easier if needed.

Also I guess it will ensure tokens already released before will work until they are refreshed for the new scope as we still allow this scope until we delete it completely.

valurefugl (approved):

libs/auth-api-lib/seeders/data/scope-internal-procuring.ts
eirikurn: It is used on the backend to perform token exchange with client credentials.

eirikurn (approved):
LGTM!

PR #8833, opened 2022-10-31
fix(circular-deps): Fix circular dependency error in auth-api-lib

snaerseljan:
# Fix error for circular dependency 

## What

 Fix error for circular dependency in auth-api-lib

## Checklist:


eirikurn (approved):
LGTM

PR #8836, opened 2022-10-31
feat(service-portal): license service refactor

thorkellmani:
## What

Refactor License service to stop using manual environments and instantiations, instead leveraging the config module and DI.

## Why

The old way is error prone and not very scalable.

## Checklist:


thordurhhh (approved):
🙌

libs/api/domains/license-service/src/lib/client/machine-license-client/genericMachineLicense.module.ts
```diff
@@ -0,0 +1,34 @@
...
+          passTemplateId: config.passTemplateId,
+        }
+        return smartConfig
+      },
+      inject: [GenericFirearmLicenseConfig.KEY],
+    },
+    SmartSolutionsApi,
+  ],
+  exports: [GenericMachineLicenseService],
+})
```
eirikurn, line 33: Ah, I see that you're instantiating multiple instances of SmartSolutionApi, each with different configuration. I guess it is not the "easy" way of using our ConfigModule, but this is still better than before.

We should join this with the [proper way](https://docs.nestjs.com/fundamentals/dynamic-modules#community-guidelines) of "instantiating multiple instances" of modules.

Currently you're importing SmartSolutionsApiClientModule, but then re-providing its options and api. Instead we should do something like this:

```
  imports: [
    SmartSolutionsApiClientModule.registerAsync({
      useFactory: (config: ConfigType<typeof GenericMachineLicenseConfig>) => {
        const smartConfig: SmartSolutionsConfig = {
          apiKey: config.apiKey,
          apiUrl: config.apiUrl,
          passTemplateId: config.passTemplateId,
        }
        return smartConfig
      },
      inject: [GenericFirearmLicenseConfig.KEY],
    }),
  ]
```

It would be nice to have NestJS 9's new `ConfigurableModuleBuilder` utility to get `registerAsync` for free. Until then, you basically need something like this:

```
@Module(...)
class SmartSolutionsApiClientModule {
  register(config: SmartSolutionsConfig) {
    return {
      module: SmartSolutionsApiClientModule,
      providers: [{ provide: SMART_SOLUTIONS_API_CONFIG, value: config }],
    }
  }

  registerAsync(configProvider: ?) {
    return {
      module: SmartSolutionsApiClientModule,
      providers: [{ ...configProvider, provide: SMART_SOLUTIONS_API_CONFIG }],
    }
  }
}
```

libs/api/domains/license-service/src/lib/client/machine-license-client/genericMachineLicense.config.ts
```diff
@@ -0,0 +1,18 @@
+import { defineConfig } from '@island.is/nest/config'
+import { SmartSolutionsApiConfigSchema as schema } from '@island.is/clients/smartsolutions'
+import * as z from 'zod'
+
+export const GenericMachineLicenseConfig = defineConfig<z.infer<typeof schema>>(
+  {
+    name: 'GenericMachineLicenseConfig ',
+    schema,
+    load: (env) => ({
+      apiKey: env.required('VE_PKPASS_API_KEY', ''),
```
eirikurn, line 10: Having "''" as a devFallback hurts DX since the developer doesn't get notified about which environment variables they are missing. An alternative is to have no fallback but use `scope: LazyDuringDevScope`. This way the whole server doesn't crash when the environment is missing but the individual requests will fail with descriptive errors. Only during development.

Just add it to providers which depend on these partial configs (in this case SMART_SOLUTIONS_API_CONFIG).

libs/api/domains/license-service/src/lib/licenseService.module.ts
```diff
@@ -96,88 +73,78 @@ export const AVAILABLE_LICENSES: GenericLicenseMetadata[] = [
...
+        type: GenericLicenseType,
+        cacheManager: CacheManager,
+      ): Promise<GenericLicenseClient<unknown> | null> => {
+        switch (type) {
+          case GenericLicenseType.DriversLicense:
+            return new GenericDrivingLicenseApi(
+              logger,
+              drivingLicenseConfig,
+              cacheManager,
+            )
```
eirikurn, line 129: Why is this instantiated on-demand instead of being singleton like the others?

libs/api/domains/license-service/src/lib/client/driving-license-client/genericDrivingLicense.config.ts
```diff
@@ -0,0 +1,51 @@
...
+  {
+    name: 'GenericDrivingLicenseConfig ',
+    schema,
+    load: (env) => ({
+      xroad: {
+        baseUrl: env.required('XROAD_BASE_PATH', 'http://localhost:8081'),
+        clientId: env.required(
+          'XROAD_CLIENT_ID',
+          'IS-DEV/GOV/10000/island-is-client',
+        ),
```
eirikurn, line 31: We generally import the xroad base path and client id from the shared XRoadConfig.

You may consider creating new shared configuration for other stuff if it makes sense (pkpass?), but no stress.

eirikurn (commented):
This task was definitely trickier then I thought but great job!

libs/api/domains/license-service/src/lib/client/machine-license-client/genericMachineLicense.module.ts
thorkellmani: Ah ok, will do

libs/api/domains/license-service/src/lib/client/machine-license-client/genericMachineLicense.config.ts
thorkellmani: Yes, good point, didn't know.

thorkellmani: Spoke to devops, there is some weirdness in the graphql schemas generation which requires a fallback on the env.required parameters. So the fallback will stay temporarily, until they fix it at least.

libs/api/domains/license-service/src/lib/licenseService.service.ts
```diff
@@ -420,8 +421,9 @@ export class LicenseServiceService {
     passTemplateId: string,
   ): GenericLicenseType | null {
     for (const [key, value] of Object.entries(this.config)) {
+      console.log(value)
```
eirikurn, line 424: Remove

libs/clients/smartsolutions/src/lib/smartsolutions.module.ts
```diff
@@ -1,8 +1,44 @@
...
 
-@Module({
-  providers: [SmartSolutionsApi],
-  exports: [SmartSolutionsApi],
-})
-export class SmartSolutionsApiClientModule {}
+export type SmartSolutionsModuleAsyncOptions = {
+  useFactory: (config: ConfigType<never>) => SmartSolutionsConfig
+  inject: Array<string>
+}
```
eirikurn, line 9: This is closer in line to the SequelizeModule.forRootAsync types:

```suggestion
export type SmartSolutionsModuleAsyncOptions = {
  useFactory: (...args: any[]) => Promise<SmartSolutionsConfig> | SmartSolutionsConfig;
  inject?: any[];
}
```

Note how inject and the useFactory args are optional and under the consumer's control.

eirikurn (approved):
Nice! Approving with two minor comments.

PR #8848, opened 2022-11-01
feat(j-s): Add support for a defender per defendant

oddsson:
# Add support for a defender per defendant

[Asana](https://app.asana.com/0/1199153462262248/1203242082057161/f)
[Asana](https://app.asana.com/0/1199153462262248/1203225278132964/f)

## What

Add support for a defender per defendant

## Why

In S-cases, defendants can have a defender of their own choosing, as well as denying the right to counsel. 

## Screenshots / Gifs

## Checklist:


apps/judicial-system/api/src/app/modules/defendant/models/defendant.model.ts
```diff
@@ -36,4 +36,19 @@ export class Defendant implements TDefendant {
...
+  readonly defenderNationalId?: string
+
+  @Field({ nullable: true })
+  readonly defenderEmail?: string
+
+  @Field({ nullable: true })
+  readonly defenderPhoneNumber?: string
+
+  @Field({ nullable: true })
+  readonly defendantWaivesRightToCounsel?: boolean
```
Arnarkari93, line 53: I think this flag should be required. We don't have 3 states here. It is good to have it optional in the input and default to false.

apps/judicial-system/backend/src/app/modules/defendant/models/defendant.model.ts
```diff
@@ -88,4 +88,39 @@ export class Defendant extends Model {
...
+  })
+  @ApiProperty()
+  defenderPhoneNumber?: string
+
+  @Column({
+    type: DataType.BOOLEAN,
+    allowNull: true,
+  })
+  @ApiProperty()
+  defendantWaivesRightToCounsel?: boolean
```
Arnarkari93, line 125: This should be required and default to false.

Arnarkari93 (changes requested):
Looks good! I have one concern about the `defendantWaivesRightToCounsel`

Arnarkari93 (approved):
LGTM

PR #8852, opened 2022-11-01
chore(service-portal): allow component in category cards

thordurhhh:
## What

* allow component in category cards

## Why

We are using react-router in service portal (and it's used in application system as well). So we have to work around the nextjs specific components in the component lib. We need to be able to pass in our own components to be able to use the components with react router instead of next-link for example.

## Checklist:


barabrian (approved):
LGTM

disaerna (approved):

PR #8861, opened 2022-11-01
feat(j-s): filter draft indictmentCases for court users

Arnarkari93:
[Asana - Ekki opna ákæru fyrir dómstólsnotendum fyrr en hún er komin í stöðuna Send](https://app.asana.com/0/1199153462262248/1202978837144009/f)
## What

- Filter out indicitment cases in draft state for `Judge` and `Registerar`

## Why

- We only want to show indicitments for court users that have been submitted

## Screenshots / Gifs
Generated sql
```sql
WHERE 
("Case"."is_archived" = false AND NOT ("Case"."state" IN ('NEW', 'DELETED')) AND 
("Case"."court_id" IS NULL OR "Case"."court_id" = 'd1e6e06f-dcfd-45e0-9a24-2fdabc2cc8bf') AND NOT
(("Case"."state" = 'DRAFT' AND ("Case"."type" = 'CHILD_PROTECTION_LAWS' OR "Case"."type" = 'PROPERTY_DAMAGE' OR "Case"."type" = 'NARCOTICS_OFFENSE' OR "Case"."type" = 'EMBEZZLEMENT' OR "Case"."type" = 'FRAUD' OR "Case"."type" = 'DOMESTIC_VIOLENCE' OR "Case"."type" = 'ASSAULT_LEADING_TO_DEATH' OR "Case"."type" = 'MURDER' OR "Case"."type" = 'MAJOR_ASSAULT' OR "Case"."type" = 'MINOR_ASSAULT' OR "Case"."type" = 'RAPE' OR "Case"."type" = 'UTILITY_THEFT' OR "Case"."type" = 'AGGRAVATED_ASSAULT' OR "Case"."type" = 'TAX_VIOLATION' OR "Case"."type" = 'ATTEMPTED_MURDER' OR "Case"."type" = 'TRAFFIC_VIOLATION' OR "Case"."type" = 'THEFT' OR "Case"."type" = 'OTHER_CRIMINAL_OFFENSES' OR "Case"."type" = 'SEXUAL_OFFENSES_OTHER_THAN_RAPE' OR "Case"."type" = 'OTHER_OFFENSES')))) 
ORDER BY "defendants"."created" ASC
```

Screenshot 2022-11-01 at 13 22 56

## Checklist:


apps/judicial-system/backend/src/app/modules/case/filters/case.filters.ts
```diff
@@ -328,12 +328,29 @@ export function getCasesQueryFilter(user: User): WhereOptions {
         ]
       : []
 
+  const blockDraftIndictmentsForCourt = [
```
gudjong, line 331: This look fine. However, this only excludes from the returned list of cases. It does not block direct access to individual cases. To block direct access, you would need to modify the function `isCaseBlockedFromUsera`.

You could accomplish blocking for both case list and direct access in one go, by passing the case type to `getAllowedStates`. I think that would be a much simpler change and would keep the change in one place.

gudjong (changes requested):

apps/judicial-system/backend/src/app/modules/case/filters/case.filters.ts
Arnarkari93: It looks like that is what we want. [See the description in asana](https://app.asana.com/0/1199153462262248/1202978837144009/f):  

> Ekki öryggis-issue heldur bara filtering í bakenda.
> R-mál eiga að sjást áfram á vinnslustigi en S-mál eiga ekki að sjást í listanum fyrr en þau eru komin með stöðuna Send (ákærandi er búinn að senda á dómstólinn)

gudjong: I think the wording in the ticket was intended to make the implementation easier, i.e. we don't have to block direct access if it complicates matters. However, I feel the code is more complicated if we don't block both and it may cause confusion down the road if case list and direct access blocking are not identical.

However, I will accept this as is, but please consider making the two behave in the same way.

gudjong (approved):
Accepting with comments.

apps/judicial-system/backend/src/app/modules/case/filters/case.filters.ts
Arnarkari93: Addressed in https://github.com/island-is/island.is/pull/8861/commits/a0135471f0d4fdb5f8bef31a71977ed2c74c392b

PR #8886, opened 2022-11-03
feat(air-discount-scheme): Make Discounts accessible in the API

kksteini:
# ...

## What

## Why

## Screenshots / Gifs

## Checklist:


apps/api/src/app/environments/environment.ts
```diff
@@ -89,6 +89,9 @@ const prodConfig = () => ({
     username: process.env.SYSLUMENN_USERNAME,
     password: process.env.SYSLUMENN_PASSWORD,
   },
+  airDiscountScheme: {
+    backendUrl: process.env.BACKEND_URL ?? 'http://localhost:4248',
+  },
```
kksteini, line 94: TODO: Get BACKEND_URL (and perhaps change name to ADS_BACKEND_URL) with the help of devops

apps/service-portal/src/auth.ts
```diff
@@ -59,6 +59,7 @@ if (userMocked) {
       ApiScope.licensesVerify,
       ApiScope.company,
       ApiScope.vehicles,
+      '@vegagerdin.is/air-discount-scheme-scope',
```
kksteini, line 62: TODO: Figure out whether scopes should be extended/augmented for privilige finetuning. As it stands this scope may be too permissive

libs/api/domains/air-discount-scheme/src/lib/models/flight.model.ts
```diff
@@ -0,0 +1,35 @@
+import { Field, ObjectType, ID } from '@nestjs/graphql'
+
+import { UserInfo as TUserInfo } from '@island.is/air-discount-scheme/types'
+import { User } from './user.model'
+import { FlightLeg } from './flightLeg.model'
+
+@ObjectType()
+export class UserInfo implements TUserInfo {
+  @Field()
+  gender!: 'kk' | 'kvk' | 'hvk'
```
kksteini, line 10: This will be fixed upon later database migration and type interaction with Loftbrú

PR #8890, opened 2022-11-03
fix(auth): Delegation service build

eirikurn:
## What

Configure esbuild for the new delegation service.

## Why

To fix node module issues on dev.


snaerseljan (approved):
LGTM!

PR #8895, opened 2022-11-03
fix(service-portal): add plausible event urls

thordurhhh:
## What

Send custom url with plausible events within the service-portal.

## Why

We are already tracking pageviews by non identifiable url's. We need to do the same thing with events. 
When a plausible event fires the url will display as `window.location.href` by default. We don't want that. We want to replace the identifiable path with a non-identifiable one.

## Checklist:


libs/service-portal/core/src/utils/formatPlausiblePathToParams.ts
```diff
@@ -0,0 +1,35 @@
...
+  const routes = Object.values(ServicePortalPath)
+  const currentPath = matchPath(path, {
+    path: routes,
+    exact: true,
+    strict: true,
+  })?.path
+
+  const pageOrigin = window.location.origin
+  const rootPath = ServicePortalPath.MinarSidurPath
+  const absoluteUrl = `${pageOrigin}${rootPath}${currentPath ?? ''}`
```
saevarma, line 28: Isn't possible for `currentPath` to have identifiable info like nationalId? If the route has a `:nationalId` segment? For example the `Child`, `FamilyMember` and `Spouse` routes have it. Wouldn't that result in the nationalId being sent to plausible if they are using events?

thordurhhh: Yes! That's exactly what's happening today and what I want to prevent with this change..

So here the `currentPath` coming from line 20 will go through matchPath from `react-router` where it will match the current path, /family/123, with the corresponding `ServicePortalPath` /family/:nationalId and return the `ServicePortalPath`.

So this:

`formatPlausiblePathToParams("/minar-sidur/123")`

should return:
```
{
    url: "https://www.islandis/minar-sidur/:nationalId",
    location: "minar-sidur/:nationalId",
  }
  ```

I was sending the wrong path in the location though, I have fixed it and added one more test case.

saevarma: Perfect. I was mixing the `path` and `url` properties from `matchPath` together.
Thanks for the clarification.

saevarma (approved):
💯

PR #8901, opened 2022-11-03
feat(auth-api): Incoming delegations and merged delegations for actors.

valurefugl:
https://www.notion.so/Incoming-delegations-REST-APIs-9d06fa15385545738444792c5302642a

https://www.notion.so/Update-findAllIncoming-to-have-unique-by-fromNationalId-domain-b194970387194dbea1f5714e101c6531

https://www.notion.so/Merge-duplicate-delegations-a035c076029e4c69bd9200f5ec8de8ae

## What

Incoming delegations for delegation UI 2.0.
Merged actor delegations.

## Why

To support the new delegation UI with multiple domains.

## Checklist:


libs/auth-api-lib/src/lib/delegations/delegations-incoming-company.service.ts
```diff
@@ -0,0 +1,49 @@
...
+@Injectable()
+export class IncomingDelegationsCompanyService {
+  constructor(
+    private featureFlagService: FeatureFlagService,
+    private rskProcuringClient: RskProcuringClient,
+    @Inject(LOGGER_PROVIDER)
+    private logger: Logger,
+  ) {}
+
+  async findAllIncoming(user: User): Promise<DelegationDTO[]> {
```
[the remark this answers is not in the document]
eirikurn: I see now that you deprecated that code. Is the plan to remove it along with the `/v1/delegations` ids code?

libs/auth-api-lib/src/lib/delegations/delegations-incoming-representative.service.ts
```diff
@@ -0,0 +1,137 @@
...
+  async findAllIncoming(user: User): Promise<DelegationDTO[]> {
+    try {
+      const feature = await this.featureFlagService.getValue(
+        Features.personalRepresentativeDelegations,
+        false,
+        user,
+      )
+      if (!feature) {
+        return []
+      }
```
eirikurn, line 38: Guess we can keep this feature flag around for a second.

libs/auth-api-lib/src/lib/delegations/delegations-incoming-custom.service.ts
```diff
@@ -0,0 +1,308 @@
...
+      .filter(
+        (d) =>
+          // The requesting client must have access to at least one scope for the delegation to be relevant.
+          d.delegationScopes && d.delegationScopes.length > 0,
+      )
+
+    const mergedDelegationDTOs = uniqBy(
+      allowedDelegations.map((d) => d.toMergedDTO()),
+      'fromNationalId',
+    )
```
eirikurn, line 115: Should we also merge these delegations by `fromNationalId`?

libs/auth-api-lib/src/lib/delegations/delegations-incoming-custom.service.ts
```diff
@@ -0,0 +1,308 @@
...
+  }
+
+  private checkIfScopeAllowed(
+    scope: DelegationScope,
+    allowedScopes: string[],
+  ): boolean {
+    return allowedScopes.includes(scope.scopeName)
+  }
+
+  private filterCustomScopeRule(scope: string, user: User): boolean {
```
eirikurn, line 193: Not used. Think we don't need this here.

libs/auth-api-lib/src/lib/delegations/delegations-incoming-custom.service.ts
```diff
@@ -0,0 +1,308 @@
...
+      const person = this.getPersonByNationalId(fromNameInfo, d.fromNationalId)
+
+      return {
+        ...d,
+        fromName: person?.name ?? d.fromName ?? UNKNOWN_NAME,
+      }
+    })
+  }
+
+  async findAllAvailableIncoming(user: User): Promise<MergedDelegationDTO[]> {
```
eirikurn, line 82: Maybe this should return DelegationDTOs and let the parent function handle delegation merging instead of doing uniqBy here?

Then this function is exactly the same as above, except it filters delegations/scopes by a client. Could be named or refactored based on that.

eirikurn (changes requested):
Exciting!!

libs/auth-api-lib/src/lib/delegations/delegations-incoming-custom.service.ts
valurefugl: Not sure if we need that?

libs/api/domains/auth/src/lib/resolvers/delegation.resolver.ts
```diff
@@ -45,7 +46,7 @@ export class DelegationResolver {
     @CurrentUser() user: User,
     @Args('input', { type: () => ActorDelegationInput, nullable: true })
     input?: ActorDelegationInput,
-  ): Promise<DelegationDTO[]> {
+  ): Promise<MergedDelegationDTO[]> {
```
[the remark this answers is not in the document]
valurefugl: Fixed?

libs/auth-api-lib/src/lib/delegations/delegations-incoming-custom.service.ts
eirikurn: We need this if/when we start supporting (third party?) clients which get consented access to scopes from more than one domain.

libs/api/domains/auth/src/lib/models/delegation.model.ts
```diff
@@ -53,6 +53,16 @@ export abstract class Delegation {
...
 }
+export abstract class MergedDelegation {
+  @Field(() => Identity)
+  from!: Identity
+
+  @Field(() => Identity)
+  to!: Identity
+
+  @Field(() => DelegationType)
+  type!: DelegationType
```
eirikurn, line 64: Maybe like this?

```suggestion
  @Field(() => [DelegationType])
  type!: DelegationType[]
```

libs/api/domains/auth/src/lib/models/delegation.model.ts
```diff
@@ -53,6 +53,16 @@ export abstract class Delegation {
   @Field(() => DelegationProvider)
   provider!: DelegationProvider
 }
+export abstract class MergedDelegation {
```
eirikurn, line 56:
```suggestion
@ObjectType('AuthMergedDelegation')
export class MergedDelegation {
```

libs/api/domains/auth/src/lib/resolvers/delegation.resolver.ts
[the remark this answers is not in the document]
eirikurn: We need a new merged-delegation.resolver.ts with field resolvers for the MergedDelegation from and to fields. Basically same implementation as in this file.

We also need to update the usage of authActorDelegations [here](https://github.com/island-is/island.is/blob/0992722b4f7a0bbcfa351ae2cc4db1dac343b71b/libs/application/ui-shell/src/components/DelegationsScreen.tsx#L120) and [here](https://github.com/island-is/island.is/blob/1b586399cee31c5e81369f14fd67b6faed1ecf5d/libs/application/graphql/src/lib/queries/actorDelegations.ts#L6).

eirikurn (changes requested):
Almost there!

eirikurn (approved):
Done!

apps/services/auth/ids-api/src/app/delegations/delegations.controller.spec.ts
```diff
@@ -295,6 +296,106 @@ describe('DelegationsController', () => {
             })
           })
 
+          describe('when user calls GET /delegations/v2', () => {
+            const path = '/delegations/v2'
```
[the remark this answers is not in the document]
saevarma: Looking at the swagger generated for the IDS api, it should be:

image

PR #8930, opened 2022-11-07
feat(service-portal): Feature flag for outgoing delegations V1

snaerseljan:
#  Feature flag for outgoing delegations V1 (older project)

## What

I added a feature flag in Configcat for older outgoing delegations. The default is "ON" in all systems for all users.

## Why

We need to be able two switch between newer and older outgoing delegations in all envs.

## Screenshots / Gifs

image

## Checklist:


saevarma (approved):
Aren't we also going to remove v2 from the menu item name for the v2 UI?

snaerseljan:
> Aren't we also going to remove v2 from the menu item name for the v2 UI?

Good spot. Yes thats probably for the best

disaerna (approved):
Nice!

PR #8934, opened 2022-11-07
fix(j-s): Fixes notification field size issue

gudjong:
# Fixes notification field size issue

Svæði fyrir viðtakendur tilkynninga í gagnagrunni er of lítið](https://app.asana.com/0/1199153462262248/1203150123978292/f)

## What

- Changes the notification recipient field to a JSON list.
- This is the second attempt to perform this update.

## Why

- The field used to be a too short STRING.

## Checklist:


oddsson (approved):

apps/judicial-system/backend/migrations/20221107150202-update-notification.js
```diff
@@ -0,0 +1,103 @@
...
+        .then(async ([notifications]) => {
+          for (const notification of notifications) {
+            const fallback = [{ success: false }]
+
+            let parsedRecipients = JSON.parse(notification.recipients)
+            if (!parsedRecipients || parsedRecipients.length === 0) {
+              parsedRecipients = fallback
+            }
+
+            await queryInterface.bulkUpdate(
```
Arnarkari93, line 31: If we don't wait for these updates, we will remove the old column and rename the new column to `recipients`. It is probably fine since this is all in a transaction and you already have all the data in the old table in memory.

Arnarkari93 (approved):
I have one concern, but I think it will not matter so I am approving.

apps/judicial-system/backend/migrations/20221107150202-update-notification.js
gudjong: The old table and data is also in the database until you commit the transaction. The database server takes care of this completely and don't even need to think about it if we use transactions.

PR #8937, opened 2022-11-08
test(system-e2e): Fix failing inbox test

kristofer-lp8f:
Current system-e2e test is failing since there are now two links to the inbox. This PR makes the selection more specific, so only one is selected.

## Checklist:


apps/system-e2e/src/tests/service-portal/smoke/service-portal.spec.ts
```diff
@@ -21,8 +21,8 @@ test.describe('Service portal', () => {
   test('should have clickable navigation bar', async () => {
     const page = await context.newPage()
     await page.goto('/minarsidur')
-    await expect(
-      await page.locator('a[href^="/minarsidur/"]:has(svg):visible').nth(4),
+    expect(
```
pshomov, line 24: This does not feel right, at least one of these should be awaited, otherwise it cannot possible be a retried operation. Can you please double check that?

pshomov (changes requested):
Potential flaky-ness introduced

apps/system-e2e/src/tests/service-portal/smoke/service-portal.spec.ts
kristofer-lp8f: My editor tells me that both `expect` and `page.locator` don't return promises, and therefore awaiting is redundant.

kristofer-lp8f: `page.locator` returns a non-promise `Locator` ([docs](https://playwright.dev/docs/api/class-page#page-locator))

Since `page.locator` is synchronous `expect` is also synchronous.

pshomov (approved):
LGTM

PR #8939, opened 2022-11-08
feat(j-s): Send notifications to all defenders in S-cases

oddsson:
# Send notifications to all defenders in S-cases

[Asana](https://app.asana.com/0/1199153462262248/1203242082057169/f)

## What

In S-cases, each defendant can have his or her own defender. Notifications need to be sent to each and every defender.

These notifications are

- `DEFENDER_ASSIGNED`: A notification sent to defenders, letting them know they have been assigned to a case.
-  `RULING`: A notification sent to defenders when a ruling has been made in a case they are assigned as defender in.

## Checklist:


Arnarkari93 (commented):
One thought

apps/judicial-system/backend/src/app/modules/notification/notification.service.ts
```diff
@@ -1224,39 +1245,67 @@ export class NotificationService {
...
+    if (isIndictmentCase(theCase.type) && theCase.defendants) {
+      theCase.defendants.forEach((defendant) => {
+        const { defenderEmail, defenderNationalId, defenderName } = defendant
 
-    const { subject, body } = formatDefenderAssignedEmailNotification(
-      this.formatMessage,
-      theCase,
-      `${this.config.clientUrl}${DEFENDER_ROUTE}/${theCase.id}`,
-    )
+        if (!defenderNationalId || !defenderEmail) {
```
gudjong, line 1257: Actually, I think the intention is to send a notification to a defender even if no national id is stored. This is done for ruling notifications. Then difference between defenders with a national id and without a national id is the contents of the email. A defender with national id will get a link to RVG, other defenders get an email and instructed to ask the court for documents.

apps/judicial-system/backend/src/app/modules/notification/notification.service.ts
```diff
@@ -1224,39 +1245,67 @@ export class NotificationService {
...
+        )
+
+        promises.push(
+          this.sendEmail(subject, body, defenderName, defenderEmail),
+        )
+      })
+    } else {
+      const { defenderEmail, defenderNationalId, defenderName } = theCase
+
+      if (!defenderNationalId || !defenderEmail) {
```
gudjong, line 1282: Same comment as above. I realise that the previous implementation did not send an email to defenders without national ids, but I think that was incorrect.

gudjong (changes requested):
One idea for change.

apps/judicial-system/backend/src/app/messages/notifications.ts
```diff
@@ -535,11 +535,19 @@ export const notifications = {
...
+    // TODO: REMOVE body
     body: {
       id: 'judicial.system.backend:notifications.defender_assigned_email.body',
       defaultMessage:
         '{court} hefur skipað þig verjanda í máli {linkStart}{courtCaseNumber}{linkEnd}. Gögn málsins eru aðgengileg í Réttarvörslugátt með rafrænum skilríkjum.',
       description: 'Texti í pósti til verjanda þegar hann er skráður á mál.',
     },
+    bodyV2: {
+      id:
+        'judicial.system.backend:notifications.defender_assigned_email.body_v2',
```
Arnarkari93, line 547: Where is this being used? I don´t see it in this diff.

apps/judicial-system/backend/src/app/modules/notification/notification.service.ts
```diff
@@ -1221,17 +1241,18 @@ export class NotificationService {
...
+  private async shouldSendDefenderAssignedNotification(
     theCase: Case,
-  ): Promise<SendNotificationResponse> {
-    const defenderEmail = theCase.defenderEmail
-    if (!theCase.defenderNationalId || !defenderEmail) {
-      return Promise.resolve({ notificationSent: false })
-    }
-
+    defenderEmail?: string,
+  ): Promise<true | { notificationSent: boolean }> {
```
Arnarkari93, line 1247: Seems like you can get away with having the return value only `Promise<boolean>`

apps/judicial-system/backend/src/app/messages/notifications.ts
oddsson: It's here 👉  https://github.com/island-is/island.is/pull/8939/files#diff-b2788662cb5317437f7ba480f353c0d321e3181497eb8d1accfed00b066ca25eR663

apps/judicial-system/backend/src/app/modules/notification/notification.service.ts
oddsson: 😬 😬 😬 

Screen Shot 2022-11-10 at 10 39 20

Arnarkari93: Yes, we will need to change this to false. I did not see any code using this `{notificationSent: false}` object, no?

apps/judicial-system/backend/src/app/modules/notification/internalNotification.controller.ts
```diff
@@ -35,7 +35,7 @@ export class InternalNotificationController {
     @Param('caseId') caseId: string,
     @CurrentCase() theCase: Case,
     @Body() notification: SendNotificationDto,
-  ): Promise<SendNotificationResponse> {
+  ): Promise<SendNotificationResponse | void> {
```
gudjong, line 38: Please consider modifying the code so you do not need "void".

apps/judicial-system/backend/src/app/modules/notification/notification.controller.ts
```diff
@@ -62,7 +62,7 @@ export class NotificationController {
     @CurrentHttpUser() user: User,
     @CurrentCase() theCase: Case,
     @Body() notification: SendNotificationDto,
-  ): Promise<SendNotificationResponse> {
+  ): Promise<SendNotificationResponse | void> {
```
gudjong, line 65: Please consider modifying the code so you do not need "void".

apps/judicial-system/backend/src/app/modules/notification/notification.service.ts
```diff
@@ -1221,17 +1241,18 @@ export class NotificationService {
...
     const pastNotifications = await this.notificationModel.findAll({
       where: { caseId: theCase.id, type: NotificationType.DEFENDER_ASSIGNED },
     })
+
+    if (!defenderEmail) {
+      return Promise.resolve({ notificationSent: false })
```
gudjong, line 1253: Please consider returning `false`.

apps/judicial-system/backend/src/app/modules/notification/notification.service.ts
```diff
@@ -1240,23 +1261,74 @@ export class NotificationService {
       return Promise.resolve({ notificationSent: false })
```
gudjong, line 1261: Please consider returning `false`.

apps/judicial-system/backend/src/app/modules/notification/notification.service.ts
```diff
@@ -1240,23 +1261,74 @@ export class NotificationService {
...
+      for (const defendant of theCase.defendants) {
+        const { defenderEmail, defenderNationalId, defenderName } = defendant
+
+        const shouldSend = await this.shouldSendDefenderAssignedNotification(
+          theCase,
+          defenderEmail,
+        )
+
+        if (shouldSend !== true) {
+          break
```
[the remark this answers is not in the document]
gudjong: Should this be `continue` instead of `break`.

gudjong (changes requested):

apps/judicial-system/backend/src/app/modules/notification/notification.service.ts
oddsson: Yea, thats correct. I've now made changes to this. Thanks 🙏

Arnarkari93 (approved):

gudjong (approved):

PR #8941, opened 2022-11-08
refactor(island-ui): pagination totalPages calculation

disaerna:
# Island-UI Pagination

## What

Update **Pagination** component to calculate totalPages. 

## Why

Every component that uses Pagination has to calculate totalPages from totalDataSize and itemsPerPage. 
Adding the ability for pagination component to calculate the totalPages to ensure correct totalPages.

If **itemsPerPage** and **totalItems** are not undefined then we calculate, otherwise use totalPages param that defaults to zero.

## Checklist:


thordurhhh (approved):
🙌

oddsson (approved):

PR #8947, opened 2022-11-08
feat(financial-statement-inao):  input error for negative numbers

HelgiHH:
## What

show error after a user has entered negative number

## Checklist:


libs/application/templates/financial-statements-inao/src/lib/utils/dataSchema.ts
```diff
@@ -15,6 +15,14 @@ const FileSchema = z.object({
   url: z.string().optional(),
 })
 
+const checkIfNegative = (inputNumber: string) => {
+  if (Number(inputNumber) < 0) {
+    return false
+  } else {
+    return true
```
programm-ingovals, line 22: Naming here is confusing a bit. `checkIfNegative `returns `false `if it is negative. Also code could be simplified.

`const isNotNegative = (inputNumber: string) => Number(inputNumber) >= 0`

PR #8973, opened 2022-11-09
feat(auth-api): Add endpoint for updating claims.

valurefugl:
## What

Add endpoint to ids-api for updating user claims.

## Why

So that the IDS can keep them up to date.

## Screenshots / Gifs

## Checklist:


saevarma (approved):
LGTM

libs/auth-api-lib/src/lib/user-identities/user-identities.service.ts
```diff
@@ -214,4 +217,30 @@ export class UserIdentitiesService {
...
+          value: c.value,
+          valueType: c.valueType,
+          issuer: c.issuer,
+          originalIssuer: c.originalIssuer,
+        })),
+        { transaction: t },
+      )
+    })
+
+    return this.claimModel.findAll({ where: { subjectId: subjectId } })
```
saevarma, line 244: As we are writing and reading in the same service call we should add `useMaster: true`, we @eirikurn  saw issue when creating delegations as the query read from read replica which was not updated.

saevarma (changes requested):

libs/auth-api-lib/src/lib/user-identities/user-identities.service.ts
valurefugl: Good point, fixed.

libs/auth-api-lib/src/lib/user-identities/user-identities.service.ts
```diff
@@ -241,6 +241,8 @@ export class UserIdentitiesService {
       )
     })
 
-    return this.claimModel.findAll({ where: { subjectId: subjectId } })
+    return this.claimModel.findAll({
+      where: { subjectId: subjectId, useMaster: true },
```
eirikurn, line 245: `useMaster: true` should be outside the `where` object.

eirikurn (changes requested):

libs/auth-api-lib/src/lib/user-identities/user-identities.service.ts
valurefugl: Yikes. Fixed.

eirikurn (approved):
💯

PR #8974, opened 2022-11-09
feat(service-portal): Assets service version 2

thordurhhh:
## What

* Add assets service version 2
* Add feature flag to switch between v1 and v2

## Why

* Version 1 of the assets service will be discontinued on November 21st.
* We will add version 2 behind a feature-flag, where we can try v2 before the cut-over.
* Once we get the go-ahead from the service-providers we can turn the flag ON for everyone without having to hotfix.
* And we'll clean up the flag, and remove v1 once v2 has taken over.

## Checklist:

- [ ] ~Formatting passes locally with my changes~

### DIFF
--- a/libs/clients/assets-v2/src/lib/assetsV2.config.ts
+++ b/libs/clients/assets-v2/src/lib/assetsV2.config.ts
@@ -1,0 +1,53 @@
+import { NationalRegistryScope } from '@island.is/auth/scopes'
+import { defineConfig } from '@island.is/nest/config'
+import { z } from 'zod'
+
+const schema = z.object({
+  xRoadServicePath: z.string(),
+  fetch: z.object({
+    timeout: z.number().int(),
+    autoAuth: z
+      .object({
+        mode: z.enum(['token', 'tokenExchange', 'auto']),
+        issuer: z.string(),
+        clientId: z.string(),
+        clientSecret: z.string(),
+        scope: z.array(z.string()),
+      })
+      .optional(),
+  }),
+})
+
+export const AssetsV2ClientConfig = defineConfig<z.infer<typeof schema>>({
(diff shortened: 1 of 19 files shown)


disaerna (approved):
Like it 💯

libs/clients/assets-v2/src/lib/assetsV2.config.ts
```diff
@@ -0,0 +1,53 @@
...
+              clientId:
+                env.optional('XROAD_PROPERTIES_CLIENT_ID') ??
+                '@island.is/clients/national-registry',
+              clientSecret,
+              scope: env.optionalJSON('XROAD_PROPERTIES_SCOPE') ?? [
+                NationalRegistryScope.properties,
+                'api_resource.scope',
+              ],
+            }
+          : undefined,
```
saevarma, line 49: You should reference the `charge-fjs-v2` client for the auto auth setup.

In short:
Instead of defining autoAuth here you should do it in your `PropertiesV2ApiProvider` and use the `IdsClientConfig` from `@island.is/nest/config`. This simplifies your config schema as it only needs to contain the `autoAuth.scope` config and optionally the `fetch.timeout`.

The v1 `assets` client is using the "old" version where we initially were going to start with machine client per organisation integration. But it should be using the `api` machine client for token exchange, as we are going forward with one machine client per service, and the `IdsClientConfig` is a utility to minimise the boilerplate.

And as always feel free to ping me on slack to if anything is unclear :)

saevarma (changes requested):
Reviewed core files and they look good. Just minor refactor needed on the `autoAuth` setup.

saevarma (approved):
Core files are 💯

brynjarorng (approved):
LGTM

PR #8982, opened 2022-11-09
fix(service-portal): Delegation UI V2 fixes

snaerseljan:
#  Delegation UI V2 fixes

## What

- Fixes sorting delegation by name in access control view.
- Sorts out expired delegations on the API GraphQL client
- Removes background colour on row list in scope rights.
- Disables confirm button if no delegations have been selected

## Checklist:

### DIFF
--- a/libs/api/domains/auth/src/lib/services-v2/meDelegations.service.ts
+++ b/libs/api/domains/auth/src/lib/services-v2/meDelegations.service.ts
@@ -17,11 +17,7 @@ import {
   UpdateDelegationInput,
 } from '../dto'
 import { DelegationByOtherUserInput } from '../dto/delegationByOtherUser.input'
-import {
-  DelegationDTO,
-  DelegationScopeDTO,
-  MeDelegationsServiceInterface,
-} from '../services/types'
+import { DelegationDTO, MeDelegationsServiceInterface } from '../services/types'
 
 @Injectable()
 export class MeDelegationsServiceV2 implements MeDelegationsServiceInterface {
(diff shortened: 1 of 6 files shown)


libs/api/domains/auth/src/lib/services-v2/meDelegations.service.ts
```diff
@@ -190,12 +187,15 @@ export class MeDelegationsServiceV2 implements MeDelegationsServiceInterface {
...
     }
+
     return {
       ...delegation,
-      scopes: delegation.scopes?.map((scope) => ({
-        ...scope,
-        domainName: delegation.domainName,
-      })),
+      scopes: delegation.scopes
+        ?.filter((scope) => scope?.validTo && scope.validTo > new Date())
```
eirikurn, line 194: We also need to show scopes which are valid to and including "today". These are saved as the start of today, so maybe use date-fns `const today = startOfDay(new Date())` and check if `scope.validTo >= today`.

libs/service-portal/access-control/delegations/src/components/delegations/DelegationsFromMe.tsx
```diff
@@ -39,7 +40,11 @@ export const DelegationsFromMe = () => {
   })
 
   const delegations = useMemo(
-    () => (data?.authDelegations as AuthCustomDelegation[]) ?? [],
+    () =>
+      sortBy(
+        data?.authDelegations as AuthCustomDelegation[],
```
eirikurn, line 45: Can skip some of these explicit types now that we have hooks?

eirikurn (changes requested):
One mustfix :)

eirikurn (approved):
🙌

PR #8985, opened 2022-11-09
feat(web): Email Signup content type refactor

RunarVestmann:
# Email Signup content type refactor

## What

* The Email Signup (mailingListSignupSlice) content type in Contentful is very limited and "hardcoded". It wasn't possible to have more input fields (you could only have a name and email field) and also there wasn't any way to choose in what order the fields would appear
* I created a new content type which will replace the old one (I will clean up the code once this change goes live and I can safely do so).
* This new content type allows us to setup the form fields in a more generic manner (so we can now choose fields and in what order they appear in)

## Why

* The current solution for email signup on the web wasn't at all scalable, so this change was made because of the needed feature of being able to have more input fields.

## Screenshots / Gifs

image
## Checklist:


pshomov (approved):
LGTM

siggifv (approved):

PR #8997, opened 2022-11-10
chore(service-portal): Update path of start service worker for minarsidur

thordurhhh:
## What

* Update path of service-worker registration URL.
https://mswjs.io/docs/api/setup-worker/start#serviceworker

## Why

* The mocking in service-portal was no longer working after some update, not sure what the cause is but with a default setting it can not access `mockServiceWorker.js`

## Checklist:


eirikurn (approved):
Not ideal to have SP specific code here but fine until we find another solution :)

PR #9041, opened 2022-11-15
fix(j-s): Indictment Creation

gudjong:
# Indictment Creation

[Stofna ný S-mál í DRAFT state](https://app.asana.com/0/1199153462262248/1203363337546688/f)

## What

- Creates indictment cases in DRAFT state instead of NEW.

## Why

- The states NEW and DRAFT are handled identically for indictments and thus no need to transition from NEW to DRAFT.

## Checklist:


apps/judicial-system/backend/src/app/modules/case/test/caseController/create.spec.ts
```diff
@@ -7,6 +7,9 @@ import {
   CaseOrigin,
   CaseState,
   CaseType,
```
oddsson, line 9: CaseType is not used, please remove it.

```suggestion
```

oddsson (approved):

PR #9054, opened 2022-11-16
feat(delegation-api): Add domains and scope direction support for filtering

saevarma:
# ...

[Extend scopes API to support incoming delegations UI](https://www.notion.so/Extend-scopes-API-to-support-incoming-delegations-UI-c8ab6d2436df4543ad26d03e948706e4)
[Extend domains API to support incoming delegations UI](https://www.notion.so/Extend-domains-API-to-support-incoming-delegations-UI-f0e87875e03f4234a5abc54a80e5869b)

## What

Adding `direction` query parameter to `/domains`, `/ domains/:domainId/scope-tree` and `/domains/:domainId/scopes` endpoints.

## Why

To change default case to return all domains and all scopes in domain when not filtered by direction. Then support `direction=outgoing` to filter the list to filtered scopes for only allowed for outgoing delegations.

## Screenshots / Gifs

N/A

## Checklist:

### DIFF
--- a/apps/services/auth/delegation-api/test/access-test-cases.ts
+++ b/apps/services/auth/delegation-api/test/access-test-cases.ts
@@ -5,7 +5,7 @@ import {
 import { AuthScope } from '@island.is/auth/scopes'
 import { User } from '@island.is/auth-nest-tools'
 import { ConfigType } from '@island.is/nest/config'
-import { DelegationConfig } from '@island.is/auth-api-lib'
+import { DelegationConfig, DelegationDirection } from '@island.is/auth-api-lib'
 import { CreateCustomDelegation, CreateDomain } from './fixtures/types'
 
 export interface DomainAssertion {
@@ -20,12 +20,19 @@ export interface TestCase {
   accessTo?: string[]
   domains: CreateDomain[]
   expected: DomainAssertion[]
+  direction?: DelegationDirection
 }
+
+const personUser = createCurrentUser({
+  nationalIdType: 'person',
+  scope: [AuthScope.delegations],
+})
 
 export const accessTestCases: Record<string, TestCase> = {
   // Normal user should be able to grant delegations for scopes allowing explicit delegation grants.
   happyCase: {
-    user: createCurrentUser({ scope: [AuthScope.delegations] }),
+    direction: DelegationDirection.OUTGOING,
+    user: personUser,
     domains: [
       {
         name: 'd1',
@@ -50,7 +57,8 @@ export const accessTestCases: Record<string, TestCase> = {
   },
   // Should not see scopes unless they allow explicit delegation grants.
   noExplicitDelegationGrant: {
-    user: createCurrentUser({ scope: [AuthScope.delegations] }),
+    direction: DelegationDirection.OUTGOING,
+    user: personUser,
     domains: [
       {
         name: 'd1',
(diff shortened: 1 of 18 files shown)


apps/services/auth/delegation-api/test/access-test-cases.ts
```diff
@@ -179,10 +189,16 @@ export const accessTestCases: Record<string, TestCase> = {
         ],
       },
     ],
-    expected: [],
+    expected: [
+      {
+        name: 'd1',
+        scopes: [{ name: 's1' }],
+      },
+    ],
```
eirikurn, line 197: Did you mean to change this test case? Isn't it valid as it was, for outgoing?

apps/services/auth/delegation-api/test/access-test-cases.ts
```diff
@@ -302,4 +318,80 @@ export const accessTestCases: Record<string, TestCase> = {
     ],
     expected: [],
   },
+  // Should get list of all domains if not filtered by direction.
+  allDomains: {
```
eirikurn, line 322: Arguably these test cases should only focus on outgoing filtering. We can create separate test cases or stand-alone tests for this. Otherwise we're running a lot of invalid tests on this test-cases.

libs/auth-api-lib/src/lib/delegations/delegations-outgoing.service.ts
```diff
@@ -292,6 +308,7 @@ export class DelegationsOutgoingService {
 
   private async findOneInternal(
     user: User,
+    direction: DelegationDirection,
```
eirikurn, line 311: Not used?

libs/auth-api-lib/src/lib/resources/delegation-resources.service.ts
```diff
@@ -137,36 +154,59 @@ export class DelegationResourcesService {
...
+        ...this.skipScopeFilter(user, prefix),
+        ...this.accessControlFilter(user, prefix),
+        ...this.delegationTypeFilter(user, prefix),
+      )
+    }
+
+    return apiScopeFilter
   }
 
   apiScopeInclude(user: User) {
```
eirikurn, line 212: Can also skip a bunch of includes if we're not filtering outgoing.

eirikurn (approved):
LGTM. Mostly nice to have comments I think :)

apps/services/auth/delegation-api/test/access-test-cases.ts
saevarma: This change is on the `unrelatedCustomDelegations` not `customDelegationScopes`, and as I didn't apply any direction on the `unrelatedCustomDelegations` as it is not related to the user, the domain and scope is returned in the corresponding endpoints.

eirikurn: I know, but the purpose of this test case is to make sure that unrelated custom delegations don't affect `direction=outgoing` filtering of the current user.

saevarma: The `unrelatedCustomDelegations` is important test case for the delegations controller but causes trouble in the general `it.each` approach for the domains. So I introduced `expectedDomains` property to solve this test case for the domains.

PR #9062, opened 2022-11-16
fix(parental-leave): Other parent data still shown when opting out of providing it

markburk:
## What

Do not show other parent data when, after having first provided it, the user goes back and opts to not provide it

## Why

If a user enters information about the other parent (like email and phone number) as well as stating they want to request days from that parent but then changes their mind and goes back to change their selection, the previously entered data is still visible.

## Screenshots / Gifs

UUFV-177_other_parent_data_01

## Checklist:


libs/application/templates/parental-leave/src/fields/OtherParent/index.tsx
```diff
@@ -8,38 +8,126 @@ import {
...
+import { SPOUSE, NO, SINGLE, TransferRightsOption } from '../../constants'
 import { useFormContext } from 'react-hook-form'
 
 export const OtherParent: FC<FieldBaseProps> = ({ application, field }) => {
   const { id, title } = field
+  const { errors, setValue, register } = useFormContext()
   const { formatMessage } = useLocale()
-  const { errors, setValue } = useFormContext()
+
+  // const {
```
arnarleifs, line 23: This commented out code, can be removed.

libs/application/templates/parental-leave/src/fields/OtherParent/index.tsx
```diff
@@ -8,38 +8,126 @@ import {
...
+            component: FieldComponents.RADIO,
+            title,
+            children: undefined,
+            options: (application) =>
+              getOtherParentOptions(application, formatMessage),
+            onSelect: async (s: string) => {
+              if (s === SPOUSE || s === NO || s === SINGLE) {
+                setValue('otherParentObj.otherParentName', '')
+                setValue('otherParentObj.otherParentId', '')
+                // if (s === NO || s === SINGLE) {
```
arnarleifs, line 62: This as well, and on other places.

libs/application/templates/parental-leave/src/lib/answerValidators.ts
```diff
@@ -178,7 +179,14 @@ export const answerValidators: Record<string, AnswerValidator> = {
     }
     return undefined
   },
-  // TODO: should we add validation for otherParent's email?
+  [OTHER_PARENT_EMAIL]: (newAnswer: unknown, application: Application) => {
+    const email = newAnswer as string
+    const { otherParent } = getApplicationAnswers(application.answers)
+    if (otherParent !== NO && otherParent !== SINGLE && !isValidEmail(email)) {
```
arnarleifs, line 185: I would consider reversing this boolean for clarity:

`if (otherParent === YES && otherParent !== SINGLE && !isValidEmail(email)) { ... }`

libs/application/templates/parental-leave/src/lib/answerValidators.ts
```diff
@@ -178,7 +179,14 @@ export const answerValidators: Record<string, AnswerValidator> = {
...
+      return buildValidationError(OTHER_PARENT_EMAIL)(errorMessages.email)
+    }
+    return undefined
```
arnarleifs, line 188: Explicitly returning undefined is unnecessary.

libs/application/templates/parental-leave/src/fields/OtherParent/index.tsx
qdd1: I have them there so if someone want to take this feature could continue to work with it

libs/application/templates/parental-leave/src/fields/OtherParent/index.tsx
qdd1: Same here ( I have them there so if someone want to take this feature could continue to work with it )

arnarleifs (approved):

PR #9073, opened 2022-11-17
fix(web): Check circularity after mapping models

RunarVestmann:
# Check circularity after mapping models

## What

* Data is allowed to be circular and the mapping should make sure to handle the circularity correctly
* This change checks for circularity after mapping the model instead of beforehand

## Why

* Some perfectly valid articles weren't getting indexed by the search-indexer since they contained entry-hyperlink fields that had circular references (the mapping would take care of that)

## Checklist:


libs/cms/src/lib/search/importers/article.service.ts
```diff
@@ -111,11 +111,19 @@ export class ArticleSyncService implements CmsSyncProvider<IArticle> {
...
+          const mappedEntry = mapArticle(processedEntry)
+          if (!isCircular(mappedEntry)) {
+            processedEntries.push(processedEntry)
+          } else {
+            logger.warn('Circular reference found in article', {
+              id: entry?.sys?.id,
+            })
+          }
+        } catch (error) {
+          logger.warn('Failed to map article', {
```
siggifv, line 124: Consider logging this as error. Same for the `Circular reference found in article` above.
Apparently, there are many places like this. We should discuss what we should consider as error vs. warning, keeping in mind that we want to know when some data is not being synced/updated, as there is highly likely that someone is making changes in Contentful and waiting for these changes to be appear on the web.

siggifv (changes requested):

libs/cms/src/lib/search/importers/supportQNA.service.ts
```diff
@@ -30,17 +30,19 @@ export class SupportQNASyncService implements CmsSyncProvider<ISupportQna> {
...
-            })
-          ) {
-            processedEntries.push(entry)
-          } else {
-            logger.warn('Circular reference found in question', {
-              id: entry.sys.id,
+          try {
+            const mappedEntry = mapSupportQNA(entry)
+            if (!isCircular(mappedEntry)) {
+              processedEntries.push(entry)
```
kristofer-lp8f, line 36: Is this really supposed to be the non-mapped `entry`?
```suggestion
              processedEntries.push(mappedEntry)
```

libs/cms/src/lib/search/importers/news.service.ts
```diff
@@ -13,13 +13,12 @@ export class NewsSyncService implements CmsSyncProvider<INews> {
   processSyncData(entries: processSyncDataInput<INews>) {
```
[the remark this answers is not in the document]
RunarVestmann: Right, this function could be called something else.

However for example in the article.service.ts file this function is actually processing something (it's removing fields) so one might argue that this function name might be right.

libs/cms/src/lib/search/importers/article.service.ts
siggifv: As per our discussions on Slack, we will not be logging these as errors, at least for now. Errors will call the operations team to action, which we don't want to do until we're confident that the indexer will not throw "too many" of those. In the meantime, we'll keep an eye on the logged warnings while we stabilize the indexer code.

siggifv (approved):

PR #9075, opened 2022-11-17
fix(j-s): Case Files Record Page Numbers

gudjong:
# Case Files Record Page Numbers

[Laga blaðsíðutöl í málaskrár pdf](https://app.asana.com/0/1199153462262248/1203265017403971/f)

## What

- Starts counting from page one.

## Why

- This is the preferred way.

## Screenshots / Gifs

image

## Checklist:


apps/judicial-system/backend/src/app/modules/file/dto/updateFile.dto.ts
```diff
@@ -27,6 +27,7 @@ export class UpdateFileDto {
   @ApiPropertyOptional()
   readonly chapter?: number | null
 
+  @IsOptional()
```
oddsson, line 30: Please add this to `chapter` as well

oddsson (changes requested):
One thing

oddsson (approved):

PR #9097, opened 2022-11-21
chore(j-s): Case Type Guard

gudjong:
# Case Type Guard

[Blokka endapunkta á server fyrir sakamálum þar sem við á](https://app.asana.com/0/1199153462262248/1202974314193375/f)

## What

- Adds a case type guard.
- Applies the new guard to select controller end points.

## Why

Improved access control.

## Checklist:


apps/judicial-system/backend/src/app/modules/case/guards/test/caseTypeGuard.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+interface Then {
+  result: boolean
+  error: Error
+}
+
+type GivenWhenThen = () => Then
+
+describe('Case Type Guard', () => {
+  const mockRequest = jest.fn()
+  const [selectedTypes, remainingTypes] = randomListSplit(CaseType)
```
Arnarkari93, line 21: Consider not using random values in tests. It does not matter what types are put here as long as they are not duplicated, so I am just going to approve.

Arnarkari93 (approved):

PR #9112, opened 2022-11-22
fix(estate): Application/Estate - fixing after tests

albinagu:
## Checklist:


libs/application/templates/estate/src/forms/ResidencePermit/overviewSection.ts
```diff
@@ -154,17 +159,23 @@ export const overview = buildSection({
...
-        }),
+        buildCustomField(
+          {
+            title: '',
+            id: 'bankAccountsCards',
+            component: 'Cards',
+            doesNotRequireAnswer: true,
+          },
+          {
+            cards: ({ answers }: Application) =>
```
kksteini, line 170: The answers are validated by the dataschema.
The dataschema can be used to infer types.

Maybe we could use it to type the answers rather than using any

```typescript
import { estateSchema } from '@island.is/application/templates/estate'
import { infer as zinfer } from 'zod'
type EstateSchema = zinfer<typeof estateSchema>
type EstateAnswers = EstateSchema['estate']

// and then later
cards: ({ answers }: EstateAnswers) =>
  (answers?.bankAccounts ?? []).map((account) => .....
```

kksteini (approved):

PR #9113, opened 2022-11-22
feat(api): PowerBi embed tokens

RunarVestmann:
# PowerBi embed tokens

## What

* If applied, this change will allow users to embed Power BI reports that require an embed token inside rich text fields in Contentful (they do so by using the Power BI Slice content type)

## Why

* Many organizations have reports in Power BI which they'd like to show on the web and as of right now we aren't able to embed report that require an embed token

## Checklist:


libs/api/domains/powerbi/src/lib/powerbi.resolver.ts
```diff
@@ -0,0 +1,176 @@
...
+    let clientId = ''
+    let clientSecret = ''
+    let tenantId = ''
+
+    switch (owner) {
+      case 'Fiskistofa':
+        {
+          clientId = process.env.FISKISTOFA_POWERBI_CLIENT_ID as string
+          clientSecret = process.env.FISKISTOFA_POWERBI_CLIENT_SECRET as string
+          tenantId = process.env.FISKISTOFA_POWERBI_TENANT_ID as string
```
eirikurn, line 29: Should use [ConfigModule](https://github.com/island-is/island.is/tree/main/libs/nest/config) to get runtime config validation.

These variables also need to be defined in the infra file and parameter store for all environments.

libs/api/domains/powerbi/src/lib/powerbi.resolver.ts
```diff
@@ -0,0 +1,176 @@
...
+
+    const clientApplication = new ConfidentialClientApplication(msalConfig)
+
+    try {
+      const tokenResponse = await clientApplication.acquireTokenByClientCredential(
+        {
+          scopes: ['https://analysis.windows.net/powerbi/api/.default'],
+        },
+      )
+      return tokenResponse?.accessToken
```
eirikurn, line 56: I wouldn't be surprised if you could use our enhanced fetch autoAuth logic to automatically get access token using client credentials.

No need to delay the project to explore that though.

libs/api/domains/powerbi/src/lib/powerbi.resolver.ts
```diff
@@ -0,0 +1,176 @@
...
+const BASE_URL = 'https://api.powerbi.com/v1.0/myorg'
+
+@Resolver()
+export class PowerBiResolver {
+  constructor(
+    @Inject(LOGGER_PROVIDER)
+    private readonly logger: Logger,
+  ) {}
+
+  private async getAccessToken(owner: Owner) {
```
eirikurn, line 19: Should keep the resolver class light and move the logic to a service class.

libs/api/domains/powerbi/src/lib/powerbi.resolver.ts
```diff
@@ -0,0 +1,176 @@
...
+      }
+      const report = await response.json()
+      return report
+    } catch (err) {
+      this.logger.error(`Error occurred when trying to fetch Power Bi report`, {
+        workspaceId,
+        reportId,
+        err,
+      })
+      return null
```
eirikurn, line 92: Not recommended to do custom error handling and swallow all errors. Better to use standard error handling and handle errors as close to the UI as possible.

You may differentiate client errors from server errors. For example ok to return null if some "id" doesn't exist, but unexpected errors (missing config, 500 responses, network errors) should be thrown and handled higher up.

Enhanced Fetch can help with this and includes consistent error logging.

eirikurn (changes requested):
Need to add the variables to infra files and should validate configuration. Rest of comments are suggestions.

libs/api/domains/powerbi/src/lib/powerbi.resolver.ts
RunarVestmann: Ok, that's something I've been wondering about. How we think about handling errors.

I'll change it

libs/api/domains/powerbi/src/lib/powerbi.resolver.ts
RunarVestmann: Done

libs/api/domains/powerbi/src/lib/powerbi.resolver.ts
RunarVestmann: I agree, hopefully I was able to move things over to a service class in a "correct manner"

libs/api/domains/powerbi/src/lib/powerbi.service.ts
```diff
@@ -0,0 +1,89 @@
...
+
+    const response = await fetch(url, {
+      method: 'GET',
+      headers: {
+        'Content-Type': 'application/json',
+        Authorization: `Bearer ${accessToken}`,
+      },
+    })
+
+    return response.json()
```
eirikurn, line 55: This is too little error handling, where non-200 responses will cause hard to trace errors (either `response.json()` fails, or graphql complaints about missing attributes).

I recommend switching to enhanced fetch. It will track request metrics in datadog, automatically log request errors and non-200 responses, and throw fancy FetchError for non-200 responses.

Just add this class member:

```
fetch = createEnhancedFetch({ name: 'PowerBiService' })
```

Then use it here instead:

```suggestion
    const response = await this.fetch(url, {
      method: 'GET',
      headers: {
        'Content-Type': 'application/json',
        Authorization: `Bearer ${accessToken}`,
      },
    })

    return response.json()
```

You don't need to change anything else unless you want to `catch` and handle "some" non-200 responses differently.

libs/api/domains/powerbi/src/lib/powerbi.service.ts
```diff
@@ -0,0 +1,89 @@
...
+        },
+      ],
+      datasets: datasetIds.map((datasetId) => ({ id: datasetId })),
+      targetWorkspaces: [{ id: targetWorkspaceId }],
+    }
+
+    const embedTokenApiUrl = `${BASE_URL}/GenerateToken`
+
+    // Generate Embed token for single report, workspace, and multiple datasets. Refer to: https://aka.ms/MultiResourceEmbedToken
+    const response = await fetch(embedTokenApiUrl, {
```
eirikurn, line 77:
```suggestion
    const response = await this.fetch(embedTokenApiUrl, {
```

eirikurn (changes requested):

libs/api/domains/powerbi/src/lib/powerbi.service.ts
RunarVestmann: Thanks for the tips, I appreciate it.

I for some reason haven't been able to grasp how exactly it'd be best to handle potential errors. Whether to do nothing at all or try and catch everything.
I guess the answer is that it's good to only try and catch errors if they need to be handled in a specific way or to write out more specific and detailed log messages but then after that it's best to just throw the error instead of letting the code continue doing it's thing by returning null or something instead.

I guess from the client's point of view there either was an error or there wasn't an error so there might be no need to try and return something that indicates what kind of error happened either.

libs/api/domains/powerbi/src/lib/powerbi.service.ts
RunarVestmann: Done

eirikurn (approved):
Lgtm ❤️

brynjarorng (approved):
LGTM(Devops)

libs/api/domains/powerbi/src/lib/powerbi.service.ts
eirikurn: Yeah. It can be good to distinguish "expected" errors from "unexpected" errors, as well as client errors (bad input data) vs "internal server errors".

Expected errors should be returned in a way that the client can easily tell them a part and handle them. In REST terms this can be fancy ProblemJSON responses. In GraphQL this may be returning "null" for missing data or having a special Payload schema with typed problem details.

Everything else should be thrown.

The problem I was commenting on in this thread is that when using normal fetch in this way, unexpected errors will be thrown allright, but they won't be very descriptive. Either "string is not json" or "can not read property of undefined". The standard fetch api requires a lot of error handling boilerplate, which we don't want. Thus we use Enhanced Fetch to handle all of the boilerplate so we can keep the code simple like this. 👍

siggifv (approved):

libs/api/domains/powerbi/src/lib/powerbi.service.ts
RunarVestmann: Totally :)

PR #9127, opened 2022-11-23
fix(financial-statements-inao): Use object as input for p election

fjandakornid:
# Use object as input instead of many parameters

## What

Use object instead of many parameters for personal election

## Why

Trying to fix the "function not found" error on dev

## Checklist:


HelgiHH (approved):
lgtm

PR #9128, opened 2022-11-23
feat(application-system): if user has role in currstate that allows user access return application

alexdiljar:
General Petition application has a state that is open to all.
Currently the application system does not return applications if the national id of the user is not found in applicant, assignee og applicantActor fields on the application. This changes checks if the user has a role in the current state of the application if the application is not returned for nationalId. 

## Checklist:


apps/application-system/api/src/app/modules/application/tools/applicationAccess.service.ts
```diff
@@ -49,16 +55,36 @@ export class ApplicationAccessService {
     return existingApplication as BaseApplication
   }
 
-  async canDeleteApplication(
+  async isUserActorOnApplication(id: string, user: User) {
```
obmagnusson, line 58: I would expect a function named isUserActorOnApplication to return boolean true if the user is an actor on the application. Or atleast I dont quite get why it throws a bad subject if the user is an actor.

alexdiljar: Yeah, that makes sense. I'll move the code back, I was aiming for code readability but it's maybe just better kept intact.

obmagnusson (approved):
LGTM

PR #9131, opened 2022-11-23
fix(service-portal): Block vehicle lookup when personal delegation is active

thordurhhh:
## What

* Block vehicle lookup when personal delegation is active. 
* Add delegation helpers

## Why

Vehicle is kind of a special case where you only get 5 lookups per day. It should only be available for the individual themselves. But should be available to companies. So we can't really block all delegations. 

## Checklist:


disaerna (approved):
💯

thordurhhh:
Closing this.
Will be solved another way.

PR #9133, opened 2022-11-23
chore(infra): cleanup npm upgrade spamm

brynjarorng:
# ...

## What

## Why

## Screenshots / Gifs

## Checklist:


apps/air-discount-scheme/api/infra/api.ts
```diff
@@ -36,6 +36,7 @@ export const serviceSetup = (services: {
         staging: 'cdn.contentful.com',
         prod: 'cdn.contentful.com',
       },
+      NO_UPDATE_NOTIFIER: 'true',
```
pshomov, line 39: Maybe set this in the global env section?

brynjarorng: Are you talking about file in `infra/src/environments.ts`? It is already set there but it did not work in every location which is really strange

pshomov: That's weird. Let's fix that issue after this fix.

pshomov (approved):
LGTM

PR #9138, opened 2022-11-23
fix(auth): Tweak system-e2e delegation tests.

eirikurn:
## What

- Make login url matching more lenient to deal with trailing slashes.
- Move access control test to smoke test folder so it runs more often.

## Why

Better E2E tests.

## Checklist:


apps/system-e2e/src/tests/service-portal/smoke/company-delegation.spec.ts
```diff
@@ -39,7 +39,7 @@ test.describe('Service portal', () => {
       .textContent()
     expect(companyName).toBeTruthy()
     await firstCompany.click()
-    await page.waitForURL(homeUrl, {
+    await page.waitForURL(new RegExp(homeUrl), {
```
pshomov, line 42: A bit uncomfortable with converting a URL to a regex, but should work in this case anyway. A better way maybe to use simply `/minarsidur*`?

pshomov (approved):

PR #9173, opened 2022-11-28
fix(financial-statements-inao): Fix undefined methods on production builds

obmagnusson:
Functions where missing only on production 
Error : this.financialStatementsClientService.postFinancialStatementForPersonalElection is not a function

## What

Removed `<>` type casting

## Why

The application-system-api needs to have `"jsx": "react"`  set in tsconfig  because the templates have both server side code and client side code (React components).  The <> type cast was causing the esbuild to bundle up the rest of the file in som jsx elment rendering the rest of the function in the file inaccessable.

## Checklist:


HelgiHH (approved):
lgtm

PR #9204, opened 2022-11-29
chore(j-s): Case File Name

gudjong:
# Case File Name

[Vista sækjenda skjöl í ákærum undir upprunalegu nafni í Auði](https://app.asana.com/0/1199153462262248/1203447613452301/f)

## What

- Stores case files under original name in the court system.

## Why

- That is correct.

## Checklist:


oddsson (changes requested):
One unused import

PR #9206, opened 2022-11-30
fix(e2e): Fix brittle homepage test

kristofer-lp8f:
- Add `data-testid` to the shared `Box` component
- Move `data-testid` on homepage image to container instead of graphic

## Checklist:

### DIFF
--- a/libs/island-ui/core/src/lib/Box/types.ts
+++ b/libs/island-ui/core/src/lib/Box/types.ts
@@ -5,5 +5,4 @@ export interface BoxProps
   extends Omit<UseBoxStylesProps, 'component'>,
     Omit<AllHTMLAttributes<HTMLElement>, 'width' | 'height' | 'className'> {
   component?: ElementType
-  dataTestId?: string
 }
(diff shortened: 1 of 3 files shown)


rafnarnason (approved):
lgtm

libs/island-ui/core/src/lib/Box/types.ts
```diff
@@ -5,4 +5,5 @@ export interface BoxProps
   extends Omit<UseBoxStylesProps, 'component'>,
     Omit<AllHTMLAttributes<HTMLElement>, 'width' | 'height' | 'className'> {
   component?: ElementType
+  dataTestId?: string
```
pshomov, line 8: Not sure if it makes sense, but could we maybe use [this type](https://github.com/island-is/island.is/blob/main/libs/island-ui/utils/src/lib/testSupport.ts) instead of specifying this explicitly?

pshomov (approved):
Approved, minor suggestion for/question about improvement.

libs/island-ui/core/src/lib/Box/types.ts
kristofer-lp8f: I thought we'd removed it, since I couldn't find it. Using it now.

PR #9207, opened 2022-11-30
feat(charge-service): Add shouldDeleteCharge field to application state

johannaagma:
## What

Add possibility to delete charge even though payment has been fulfilled depending on state.
Check before we delete charge if it has already been deleted.

## Why

To make it possible to refund charge (if already paid) in certain application states.
To prevent error if charge was already deleted while deleting/pruning application.

## Checklist:


sigruntg (approved):

apps/application-system/api/src/app/modules/application/charge/application-charge.service.ts
```diff
@@ -17,21 +18,44 @@ export class ApplicationChargeService {
     this.logger = logger.child({ context: 'ApplicationChargeService' })
   }
 
-  async deleteCharge(application: Pick<Application, 'id' | 'externalData'>) {
+  async deleteCharge(
+    application: Pick<Application, 'id' | 'externalData' | 'typeId' | 'state'>,
```
obmagnusson, line 22: Should we remove external data here and use the Id from `findPaymentByApplicationId`.

apps/application-system/api/src/app/modules/application/charge/application-charge.service.ts
```diff
@@ -17,21 +18,44 @@ export class ApplicationChargeService {
...
+      // No need to delete charge if already paid (and should not be refunded)
+      if (payment.fulfilled) {
+        const template = await getApplicationTemplateByTypeId(
+          application.typeId,
+        )
+
+        const stateConfig =
+          template.stateMachineConfig.states[application.state]
+
+        if (!stateConfig.meta?.lifecycle?.shouldDeleteCharge) {
```
obmagnusson, line 43: And instead of checking external data if id exists check `shouldDeleteCharge` beforehand an then try to delete a charge.

obmagnusson (changes requested):
I have some optimization thoughts. As things are we are fetching external data for all applications that should be pruned and that has proven to scale badly if many applications need pruning at once

libs/application/types/src/lib/StateMachine.ts
```diff
@@ -71,12 +71,14 @@ export type StateLifeCycle =
       // Controls visibility from my pages + /umsoknir/:type when in current state
       shouldBeListed: boolean
       shouldBePruned: false
+      shouldDeleteCharge?: boolean | null
```
obmagnusson, line 74: Readme update?

obmagnusson:
> @obmagnusson
> 
> I've made changes, so we always get the chargeID using `findPaymentByApplicationId` instead of looking at externalData (which is actually I should've done from the start I think...).
> 
> But regarding your comment "And instead of checking external data if id exists check shouldDeleteCharge beforehand an then try to delete a charge." - right now I'm calling `findPaymentByApplicationId` for every application we are pruning. I'm wondering which would scale better to do for **every** application we are pruning (to check if application had a payment):
> 
> * Check the template's state config if the application/applicationState has gone through payment. (Note: would need to mark every application state that uses payment, so I need to add a new field to the template state config: "shouldDeleteCharge"). After confirming that the application has gone through payment, then call `findPaymentByApplicationId` to get the chargeID
> * Right away call `findPaymentByApplicationId` to check if there exists a chargeID for the application

Yeah we should go for the "mark as payment state" direction to save db calls. But I think I will include it in another pr with refactoring of the Payment system. For now I think this will suffice.

apps/application-system/api/src/app/modules/application/lifecycle/application-lifecycle.service.ts
```diff
@@ -12,10 +12,9 @@ export interface ApplicationPruning {
   pruned: boolean
   application: Pick<
     Application,
-    'id' | 'attachments' | 'answers' | 'externalData'
+    'id' | 'attachments' | 'answers' | 'externalData' | 'typeId' | 'state'
```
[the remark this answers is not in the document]
obmagnusson: Not for now it is needed for some tests

obmagnusson (approved):

PR #9259, opened 2022-12-05
feat(j-s): FormStepper refactoring

oddsson:
# FormStepper refactoring

[Asana](https://app.asana.com/0/1199153462262248/1203449306021105/f)

## What

Refactor FormStepper functionality. This includes deprecating the existing `FormStepper` component in island-ui.

## Why

The existing FormStepper component in island-ui violates [the open-closed principle](https://en.wikipedia.org/wiki/Open%E2%80%93closed_principle). You could only use links in the old form stepper but we feel that limitation should not be set by the component (you should be able to render whatever component you want in the form stepper). It is the developer's responsibility implement the component according to the design.

In our case, we needed a button in the stepper instead of links to be able to run some code when using the FormStepper.

## Screenshots / Gifs

https://user-images.githubusercontent.com/3789875/207574126-21f93cc2-ba1f-4916-82df-7191153ff200.mov

## Checklist:


libs/island-ui/core/src/lib/FormStepper/SubSections/SubSections.tsx
```diff
@@ -53,6 +55,7 @@ export const SubSections: FC<{
               }
               showIcon={showSubSectionIcon}
               href={subSection.href}
+              onClick={subSection.onClick}
```
oddsson, line 58:
```suggestion
```

apps/judicial-system/web/src/routes/Court/InvestigationCase/HearingArrangements/HearingArrangements.tsx
```diff
@@ -96,38 +97,45 @@ const HearingArrangements = () => {
...
-            ? formatDateForServer(new Date(courtDate))
-            : undefined,
-          force: true,
-        },
-      ],
+      if (hasSentNotification && !courtDateHasChanged) {
+        router.push(`${destination}/${workingCase.id}`)
+      } else {
+        setNextRoute(`${destination}/${workingCase.id}`)
+        setModalVisible(true)
```
Arnarkari93, line 123: Consider refactoring the use of these `nextRoute` states to be a part of setting the modal to visible. That way you will never render a modal that expects a destination route without one. This applies to all use-cases of this `nextRoute` states.

apps/judicial-system/web/src/utils/formHelper.ts
```diff
@@ -169,3 +171,145 @@ export const hasDateChanged = (
...
+export const stepValidations = (theCase: Case): stepValidationsType => {
+  return {
+    [constants.CREATE_RESTRICTION_CASE_ROUTE]: validations.isDefendantStepValidRC(
+      theCase,
+      theCase.policeCaseNumbers,
+    ),
+    [constants.CREATE_TRAVEL_BAN_ROUTE]: validations.isDefendantStepValidRC(
+      theCase,
+      theCase.policeCaseNumbers,
+    ),
```
Arnarkari93, line 219: We do not need to run all the validation functions until the `caller` of this function has decided which routes to validate.
```suggestion
export const stepValidations = (): stepValidationsType => {
  return {
    [constants.CREATE_RESTRICTION_CASE_ROUTE]: (theCase: Case) => validations.isDefendantStepValidRC(
      theCase,
      theCase.policeCaseNumbers,
    ),
    [constants.CREATE_TRAVEL_BAN_ROUTE]: (theCase: Case) validations.isDefendantStepValidRC(
      theCase,
      theCase.policeCaseNumbers,
    ),
```

apps/judicial-system/web/src/utils/formHelper.ts
```diff
@@ -169,3 +171,145 @@ export const hasDateChanged = (
...
+  const validationEntries = Object.entries(validations)
+  const stepsToCheck = validationEntries.filter(([key]) => steps.includes(key))
+
+  if (stepsToCheck.every(([, value]) => value === true)) {
+    return steps[steps.length - 1]
+  }
+
+  const [key] = stepsToCheck.find(([, value]) => value === false) || []
+  return key
+}
```
Arnarkari93, line 315:
```suggestion
export const findFirstInvalidStep = (steps: string[], theCase: Case) => {
  const validations = stepValidations()
  const validationFunctions = Object.entries(validations)
  const stepsToCheck = validationEntries.filter(([key]) => steps.includes(key))

  if (stepsToCheck.every(([, validationFn]) => validationFn(theCase) === true)) {
    return steps[steps.length - 1]
  }

  const [key] = stepsToCheck.find(([, validationFn]) => validationFn(theCase) === false) || []
  return key
}
```

libs/island-ui/core/src/lib/FormStepper/SubSectionItem/SubSectionItem.tsx
```diff
@@ -49,9 +53,20 @@ export const SubSectionItem: FC<SubSectionItemProps> = ({
         )}
       </SectionNumberColumn>
       {href && currentState !== 'active' ? (
-        <Link href={href} underline="small" underlineVisibility="hover">
+        <Link href={href} underline="small">
           {renderChildren()}
         </Link>
+      ) : onClick && currentState !== 'active' ? (
```
Arnarkari93, line 59: Do we need these changes after introducing the V2?

Arnarkari93 (changes requested):
Good work! I have some suggestions/nitpicks.

Arnarkari93 (approved):
🙌

PR #9276, opened 2022-12-07
feat(web): Zenter email signup configuration

RunarVestmann:
# Zenter email signup configuration

## What

* Added a way to subscribe to zenter mailing lists

## Why

* This was a feature needed by the directorate of fisheries

## Checklist:


brynjarorng (approved):
LGTM(Devops)

libs/api/domains/email-signup/src/lib/serviceProvider.ts
```diff
@@ -0,0 +1,13 @@
+import { Provider } from '@nestjs/common'
+import { ConfigType, LazyDuringDevScope } from '@island.is/nest/config'
+import { EmailSignupConfig } from './emailSignup.config'
+import { EmailSignupService } from './emailSignup.service'
+
+export const EmailSignupServiceProvider: Provider<EmailSignupService> = {
```
eirikurn, line 6: We use this "Provider" pattern to wrap classes that have been generated for us, like openapi clients.

In this case, you can mark EmailSignupService directly with `@Injectable()` and  `@Scope(LazyDuringDevScope)` (writing from memory).

libs/api/domains/email-signup/src/lib/emailSignup.resolver.ts
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    if (emailSignupModel.signupType === 'zenter') {
+      return this.emailSignupService.subscribeToZenter(
+        emailSignupModel,
+        inputFields,
+      )
+    }
+
+    return { subscribed: false }
```
eirikurn, line 49: Could move most/all of this to `this.emailSignupService.subscribe(input)`. Might even refactor mailchimp and zenter specific logic to their own services (eg ZenterSignupService) which EmailSignupService depends on.

eirikurn (approved):
Approving with minor suggestions.

libs/api/domains/email-signup/src/lib/emailSignup.resolver.ts
RunarVestmann: I did it 🙌

libs/api/domains/email-signup/src/lib/serviceProvider.ts
RunarVestmann: I think I managed to do it in a recent commit I pushed just now

ingolfurs (approved):

PR #9303, opened 2022-12-12
feat(web): Fetch 5 levels of depth when getting tab section for the web

RunarVestmann:
# Fetch 5 levels of depth when getting tab section for the web

## What

* Currently we fetch data at 4 levels of depth which is not enough, so this solution specifically fetches the tab section on the client with a depth of 5
* Now the Tab Section content is full width

## Why

* We want to be able to render accordions and faq lists inside Tab Sections

### Before
image

### After
image

## Checklist:


apps/web/components/Organization/Slice/TabSection/TabSection.tsx
```diff
@@ -23,37 +29,49 @@ interface SliceProps {
...
+  useQuery<GetTabSectionQuery, QueryGetTabSectionArgs>(GET_TAB_SECTION_QUERY, {
+    variables: {
+      input: {
+        id: tabSection?.id,
+        lang: activeLocale,
+      },
+    },
+    onCompleted(data) {
+      if (data?.getTabSection?.tabs?.length) {
+        setTabSection(data.getTabSection as TabSection)
```
Ingimarsson, line 58: Is putting the data in a state better than using the output from `useQuery` ? e.g.
```
const { loading, error, data } = useQuery(...
```

libs/island-ui/contentful/src/lib/FaqList/FaqList.tsx
```diff
@@ -25,7 +25,7 @@ export const FaqList: FC<FaqListProps> = ({
 }) => {
   return (
     <Stack space={6}>
-      {showTitle !== false && (
+      {title && showTitle !== false && (
```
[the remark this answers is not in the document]
RunarVestmann: showTitle could also be null (or undefined) in that case we'd still want to show the title so that's why this seemingly redundant way was taken

apps/web/components/Organization/Slice/TabSection/TabSection.tsx
RunarVestmann: I did it this way because I'm replacing what's already present after I've gotten data with more depth.
So the site is still fully populated and will only re-render the TabSection once I've gotten the data from contentful

Ingimarsson (approved):

oddsson (approved):

PR #9308, opened 2022-12-12
feat(portals): New shareable admin layout

snaerseljan:
# New layout

This PR introduces updates on the `Layout.tsx` component and two new properties on the PortalModule, i.e. 
```
{
  /**
   * The layout type of the module
   */
  layout?:
    | 'none' // Full screen
    | 'full' // Full grid, i.e 12 cols
    | 'default' // Narrow grid, i.e. 8 cols

  /**
   * Module layout wrapper component, that wraps all the module content.
   * This can be convenient to keep state between module routes.
   * @example
    moduleLayoutWrapper({ children }) {
      return (
        <SomeProvider>
          <h1>I am module wrapper</h1>
          {children}
        </SomeProvider>
      )
    },
   */
  moduleLayoutWrapper?: React.FC<PortalModuleProps>
}
```

Property `layout` enables us to define what kind of layout we want to wrap our module with, should it be `none` (full width and height), full or default. See comments on property.
Property `moduleLayoutWrapper` enables us to set a specific wrapper around the module, i.e. if we need to share state for example.

We entend to move `my-pages (service-portal)`  to use this Layout, once we start to transition it over to the new file structure.

## Screenshots / Gifs

Default
image

Full
image

None
image

With Wrapper and layout default
image

image

## Checklist:


libs/portals/core/src/utils/modules.ts
```diff
@@ -82,3 +82,31 @@ export const arrangeRoutes = async ({
...
+  userInfo,
+  modules,
+  apolloClient,
+}: GetActiveModuleArgs) =>
+  modules.find((module) =>
+    module
+      // Get all routes for the module
+      .routes({
+        userInfo,
+        client: apolloClient,
```
eirikurn, line 104: This `client` argument was added to support async dynamic routes, but it is not used currently and the types ban it. Only in the `global` function returns a Promise.

Should get rid of the GraphQL client until we need it again.

libs/portals/core/src/utils/modules.ts
```diff
@@ -82,3 +82,31 @@ export const arrangeRoutes = async ({
...
+      })
+      // Extract the path from each route
+      .map(({ path }) => path)
+      // Flatten the array of arrays, since route path can be string or array of strings
+      .flat()
+      // Find the route path that matches the current pathname
+      .find((path) => path === pathname),
```
eirikurn, line 111: This needs to use `matchPath` since path might be "`/adgangsstyring/umbod/:delegationId`".

libs/portals/core/src/utils/modules.ts
```diff
@@ -82,3 +82,31 @@ export const arrangeRoutes = async ({
...
+  apolloClient: ApolloClient<NormalizedCacheObject>
+  pathname: string
+}
+
+export const getActiveModule = ({
+  pathname,
+  userInfo,
+  modules,
+  apolloClient,
+}: GetActiveModuleArgs) =>
```
eirikurn, line 98: We should move all of this logic into ModuleProvider. It can use `useLocation` and handle everything in one render. This `useEffect` implementation involves two (pretty large) renders and may even flash the wrong Layout for a ms.

eirikurn (approved):
Love it. I'm ok with getting this on main and improving it in a follow up. 👍

PR #9316, opened 2022-12-13
fix(j-s): remove duplicates when displaying indictment subtypes

Arnarkari93:
[Asana - Endurtekning á undirtegund ákæru](https://app.asana.com/0/1199153462262248/1203532045215674/f)
## What

- Remove duplicates when displaying indictment subtypes

## Why

- It is not relevant information to see the same subtype appear more than once in an overview

## Screenshots / Gifs
Screenshot 2022-12-13 at 10 28 30
Screenshot 2022-12-13 at 10 28 40

## Checklist:


oddsson (changes requested):
Could you please do something similar for slack notifications? (see screenshot on Asana task)

Arnarkari93:
> Could you please do something similar for slack notifications? (see screenshot on Asana task)

Its used by both the frontend and backend code. Here is the call from the `event.service.ts`
https://github.com/island-is/island.is/blob/f44bdffb22335fcb771d6474a6b4f9efce1ac048/apps/judicial-system/backend/src/app/modules/event/event.service.ts#L76-L94

oddsson:
> > Could you please do something similar for slack notifications? (see screenshot on Asana task)
> 
> Its used by both the frontend and backend code. Here is the call from the `event.service.ts`
> 
> https://github.com/island-is/island.is/blob/f44bdffb22335fcb771d6474a6b4f9efce1ac048/apps/judicial-system/backend/src/app/modules/event/event.service.ts#L76-L94

Ah! Ofcorse

oddsson (approved):

PR #9321, opened 2022-12-13
feat(Portals): Refactor delegations queries into module

snaerseljan:
# Move delegations queries into to the module.

## What
This moves the queries as close as possible to the components in the delegation module. By doing that we ensure encapsulation of the module queries.

## Why

We do not want to have `service-portal` references in the `libs/portals/shared-modules/delegations` module and we would like to keep this convention on when generating graphql hooks

## Checklist:


libs/portals/shared-modules/delegations/project.json
```diff
@@ -25,6 +25,13 @@
       "options": {
         "command": "yarn ts-node -P libs/localization/tsconfig.lib.json libs/localization/scripts/extract 'libs/portals/shared-modules/delegations/src/{components,screens,hooks}/**/*.{js,ts,tsx}'"
       }
+    },
+    "schemas/codegen": {
+      "executor": "@nrwl/workspace:run-commands",
+      "options": {
+        "output": "libs/portals/shared-modules/delegations/src/**/*.codegen.ts",
```
eirikurn, line 32: Isn't the default `*.generated.ts`?

libs/portals/shared-modules/delegations/src/components/access/AccessConfirmModal.tsx
```diff
@@ -54,8 +55,9 @@ export const AccessConfirmModal = ({
     setError(formError)
   }
 
-  const toName = delegation?.to?.name
-  const toNationalId = delegation?.to?.nationalId
+  const toName = (delegation as AuthCustomDelegationOutgoing)?.to?.name
```
eirikurn, line 58: Can you change the component props to use the right type if this modal is only for outgoing delegations?

eirikurn (approved):
Nice! Wish this required fewer type assertions, but I don't want us to spend time on it :)

libs/portals/shared-modules/delegations/src/components/access/AccessConfirmModal.tsx
snaerseljan: Good catch

PR #9329, opened 2022-12-13
feat(fetch): HOTFIX Default to log error response bodies

eirikurn:
This is to make devops life easier. Error responses should generally
not include user PII and we already filter out national ids and more
from our logs.

We can always toggle this off on case-by-case basis.


snaerseljan (approved):
Neat! Well done

PR #9333, opened 2022-12-13
chore: Making feature-deployment with mocks an explicit decision

pshomov:
Making feature deployments with mocks an explicit option. Massaging the local environment generation.
# ...

## What

## Why

## Screenshots / Gifs

## Checklist:

### DIFF
--- a/apps/system-e2e/global.d.ts
+++ b/apps/system-e2e/global.d.ts
@@ -1,4 +1,6 @@
 // global.d.ts
+export {};
+
 declare global {
   namespace PlaywrightTest {
     interface Matchers<R, T> {
--- a/apps/system-e2e/global.d.ts
+++ b/apps/system-e2e/global.d.ts
@@ -1,5 +1,5 @@
 // global.d.ts
-export {};
+export {}
 
 declare global {
   namespace PlaywrightTest {
--- a/infra/src/dsl/types/output-types.ts
+++ b/infra/src/dsl/types/output-types.ts
@@ -181,7 +181,7 @@ export type HelmValueFile = {
 }
 export type LocalrunValueFile = {
   services: Services<LocalrunService>
-  mocks: Services<LocalrunService>
+  mocks: string
 }
 
 export interface OutputFormat<T extends ServiceOutputType> {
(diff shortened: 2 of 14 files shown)


infra/src/dsl/types/output-types.ts
```diff
@@ -181,7 +181,7 @@ export type HelmValueFile = {
 }
 export type LocalrunValueFile = {
   services: Services<LocalrunService>
-  mocks: Services<LocalrunService>
+  mocks: string
```
brynjarorng, line 184: Is this so that new services can be added without changing the interface?

brynjarorng (approved):
LGTM(Devops)

apps/system-e2e/global.d.ts
```diff
@@ -1,4 +1,6 @@
 // global.d.ts
+export {};
+
```
kristofer-lp8f, line 3: What's this for?

infra/src/feature-env.ts
```diff
@@ -117,7 +117,7 @@ yargs(process.argv.slice(2))
         env,
       )
       await writeToOutput(
-        await renderHelmValueFileContent(env, habitat, featureYaml, 'no-mocks'),
+        await renderHelmValueFileContent(env, habitat, featureYaml, argv.withMocks ?? false ? 'with-mocks' : 'no-mocks'),
```
kristofer-lp8f, line 120:
```suggestion
        await renderHelmValueFileContent(env, habitat, featureYaml, argv.withMocks ? 'with-mocks' : 'no-mocks'),
```

`undefined` is also falsey

kristofer-lp8f (approved):

infra/src/dsl/types/output-types.ts
pshomov: No, I am just adjusting the output for the mocks section - it is a command string to be copy/pasted into the terminal.

apps/system-e2e/global.d.ts
pshomov: So that TS stops complaining about `.toHaveCountGreaterThan()`

infra/src/feature-env.ts
pshomov: True, still I prefer to make the default explicit.

PR #9354, opened 2022-12-15
fix(web): Add fail-safe to slug field in search results

RunarVestmann:
# Add fail-safe to slug field in search results

## What

* If a search result does not have a url or slug field it'll cause the web to show a 500 error
* This change addresses that issue

## Checklist:


rafnarnason (approved):
lgtm

PR #9357, opened 2022-12-15
fix(web): Hotfix - Add fail-safe to slug field in search results (#9354)

RunarVestmann:
#  Hotfix - Add fail-safe to slug field in search results (#9354)


rafnarnason (approved):
lgtm

PR #9358, opened 2022-12-15
fix(qa): Run service-portal tests with locale=en

eirikurn:
## What

Add support in react SPA apps to configure locale with query parameter (`?locale=is`).

Implementation details:

- The locale kicks in fairly late, after fetching the user locale (overriding it). This was to keep the change simple and make it work in all of our SPA apps. Ideally we'd override the locale from the beginning in @island.is/localization, but that library is used by both NextJS and SPA apps so there's no simple way to get the query string.
- The query string does not actually mutate the user locale permanently. This needs more discussion, if we want outside links to be able to change the user profile.

## Why

We use this in System E2E tests to specify a locale for the test.

## Checklist:


pshomov (approved):
LGTM

saevarma (approved):

PR #9361, opened 2022-12-15
fix(web): Add fail-safe to slug field in search results

RunarVestmann:
# Add fail-safe to slug field in search results
This PR is related to #9354


rafnarnason (approved):
lgtm

PR #9365, opened 2022-12-16
fix(parental leave): verification for the national id of the employer

arnarleifs:
Fix verification for the national id of the employer

## What

Add a validation for an employer national id and make sure it is a valid company.

## Why

Because only valid companies should be able to approve as an employer.

## Checklist:


karenbjorg (approved):
looks good :)

PR #9369, opened 2022-12-16
fix(ads): Add explicit code and flight legs to domains

kksteini:
Added flight legs as a query in user space.
Gives the registered flight legs of a person this year.

Simplified the discount validation logic.

## Screenshots / Gifs

## Checklist:


apps/air-discount-scheme/backend/src/app/modules/flight/flight.controller.ts
```diff
@@ -373,6 +373,7 @@ export class PrivateFlightController {
   }
 
   @Get('users/:nationalId/flights')
+  @Scopes(AirDiscountSchemeScope.admin, AirDiscountSchemeScope.default)
```
eirikurn, line 376: Is this API designed for admins or normal users?

I would guess it's for admins since it looks up PII of other users. So it should allow  the admin scope only.

libs/api/domains/air-discount-scheme/src/lib/flight-leg/flight-leg.resolver.ts
```diff
@@ -0,0 +1,38 @@
...
+import { FlightLegsInput } from './dto/flight-leg.input'
+import { FlightLeg } from '../models/flightLeg.model'
+import { FlightLegService } from './flight-leg.service'
+import { Audit } from '@island.is/nest/audit'
+
+@UseGuards(IdsUserGuard, ScopesGuard)
+@Scopes(ApiScope.internal)
+@Audit({ namespace: '@island.is/air-discount-scheme' })
+@Resolver(() => FlightLeg)
+export class FlightLegResolver {
```
eirikurn, line 23: Shouldn't this be in the flight-leg-admin folder and require the admin scope, since it uses the admin api?

eirikurn (commented):
Not sure what this PR is for since there's no UI in it.

libs/api/domains/air-discount-scheme/src/lib/flight-leg/flight-leg.resolver.ts
kksteini: Yeah, or a different approach entirely.
I want to populate flightlegs of a user and I should be using a resolveField for that instead of writing a whole new resolver, bit of a brainfart there.

I want that lookup for a public facing user.

On facing away from the admin api: 
I'm having a problem with assigning an endpoint multiple tags because the generated api export file has errors stemming from something like
```typescript
export * from ./adminApi
export * from ./publicApi
```
will result in a type error upon exporting the same endpoint.

I'll close this for now and come back with a different approach but all insights are welcome.

PR #9371, opened 2022-12-16
fix(ap): Update ADS modal.

eirikurn:
## What

Update ADS modal:

- To match island-ui better.
- To fix zIndex issues in Admin Portal.

## Why

All modals are basically coded from scratch as island-ui only provides a base. The ADS modal was out of date. 

We would suggest implementing moving some of these styles/logic to island-ui. Either a "Modal", "ConfirmModal" or both could make sense.

## Screenshots / Gifs

image
image

## Checklist:


libs/portals/admin/air-discount-scheme/src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,96 @@
...
+                    {t.buttons.continue}
+                  </Button>
+                  <Button fluid variant="ghost" onClick={onCancel}>
+                    {t.buttons.cancel}
+                  </Button>
+                </Inline>
+              </Stack>
+
+              <Box position="absolute" top={4} right={4} aria-hidden>
+                <button onClick={onCancel}>
```
snaerseljan, line 84: Are there some native button styles that should perhaps be removed, i.e. border or focus outline?
Or is this maybe set in css globally 🤔

snaerseljan (approved):
Great refactor. Nice job

libs/portals/admin/air-discount-scheme/src/components/Modal/Modal.tsx
eirikurn: It seems to be reset globally. I took this code directly from the delegation modal. Unfortunately it doesn't match the official modal design perfectly but there was no way to fix it without writing more custom CSS which I didn't feel like doing this time.

oddsson (approved):

PR #9375, opened 2022-12-19
feat(portals): Admin portal dashboard screen

snaerseljan:
# Admin portal dashboard screen

This PR add Admin portal dashboard screen

## Screenshots / Gifs
Desktop 3 columns per row
image

Mobile
image

## Checklist:


barabrian (approved):
LGTM

apps/portals/admin/src/components/Layout/Layout.tsx
```diff
@@ -100,7 +100,9 @@ export const Layout: FC = ({ children }) => {
 
   return (
     <LayoutOuterContainer>
-      <LayoutModuleContainer layout={layout}>{children}</LayoutModuleContainer>
+      <LayoutModuleContainer layout={!activeModule ? 'none' : layout}>
```
eirikurn, line 103: The layout prop should be handled the same when ModuleLayoutWrapper is true'ish (see above).

Maybe you can default `layout` to 'non'` when destructuring?

libs/portals/core/src/lib/messages.ts
```diff
@@ -54,6 +54,19 @@ export const m = defineMessages({
...
+    id: 'portals:dashboard-title',
+  },
+  dashboardTitle: {
+    defaultMessage: 'Stjórnborð Ísland.is',
+    id: 'portals:dashboard-title',
+  },
+  dashboardImageAlt: {
+    defaultMessage: 'Mynd af byggingu sem táknar stjórnborð',
+    id: 'portals:dashboard-img-alt',
+  },
```
eirikurn, line 69: Don't we have an admin portal specific translation namespace?

apps/portals/admin/src/screens/Dashboard/Dashboard.tsx
```diff
@@ -0,0 +1,94 @@
...
+  const [topNavigation, bottomNavigation] = partition(
+    navigation?.children || [],
+    (navItem) =>
+      formatMessage(navItem.name) !==
+      formatMessage(delegationsMessages.accessControl),
+  )
+
+  const renderNavItem = (item: PortalNavigationItem, index: number) => {
+    return (
+      !item.navHide && (
```
eirikurn, line 33: I thought our navigation system automatically filtered these out?

apps/portals/admin/src/screens/Dashboard/Dashboard.tsx
```diff
@@ -0,0 +1,94 @@
...
+export const Dashboard = () => {
+  const { formatMessage } = useLocale()
+  const navigation = useNavigation()
+  const { md } = useBreakpoint()
+
+  const [topNavigation, bottomNavigation] = partition(
+    navigation?.children || [],
+    (navItem) =>
+      formatMessage(navItem.name) !==
+      formatMessage(delegationsMessages.accessControl),
```
eirikurn, line 28: It would be nice to have a better way to partition these rather than comparing translated text.

One is to have separate navigations for top and bottom.

```
const top = useNavigation(TOP_NAVIGATION)
const bottom = useNavigation(BOTTOM_NAVIGATION)
```

apps/portals/admin/src/screens/Dashboard/Dashboard.tsx
```diff
@@ -0,0 +1,94 @@
...
+                src="./assets/images/adminOverview.svg"
+                alt={formatMessage(m.dashboardImageAlt)}
+                className={styles.img}
+              />
+            )}
+          </Box>
+          <Box display="flex" flexDirection="column" rowGap={6} marginTop={5}>
+            <GridRow rowGap={3}>{topNavigation.map(renderNavItem)}</GridRow>
+            <Divider />
+            <GridRow rowGap={3}>{bottomNavigation.map(renderNavItem)}</GridRow>
```
eirikurn, line 88: We should hide these if there is nothing here, eg if the user does not have access to delegations.

eirikurn (approved):
Commented on a few minor things. Otherwise LGTM 🙌

apps/portals/admin/src/components/Layout/Layout.tsx
snaerseljan: Good point on the former one. But no then we will have to add the `layout: 'default' to every module.`

libs/portals/core/src/lib/messages.ts
snaerseljan: Ahh yeah good point. We have not created yet `libs/portals/admin/core` library. I will add it and have the translations as the first thing defined there

apps/portals/admin/src/screens/Dashboard/Dashboard.tsx
snaerseljan: copy/paste form service-portal. I will remove.

apps/portals/admin/src/screens/Dashboard/Dashboard.tsx
snaerseljan: Yes I thought about that, but I think in the end the `id` field on at least the navigation item would be great and also on the portal module.

The way you are suggesting would mean that we would have to either manually provide the top and bottom navigation or if you are thinking as a constants we would still have to filter out the desired navItems.

What about optional id field and then when service-portal moves over to the new portal structure, then we can make it required.

kristofer-lp8f (approved):

apps/portals/admin/src/screens/Dashboard/Dashboard.tsx
snaerseljan: Extracted into `useAdminNavigation()` hook and added required id field to all modules and optional id field to PortalNavigationItem.

obmagnusson (approved):

apps/portals/admin/src/screens/Dashboard/Dashboard.tsx
snaerseljan: Nevermind the above went you way :)

PR #9377, opened 2022-12-19
fix(portals-delegations): Portal specific default domain option

snaerseljan:
# Portal specific default domain option

Delegations is used both in service-portal and admin-portal. Queries must set the right default options to domain options.
This introduces default options by portal type

## Checklist:


eirikurn (approved):
LGTM 👍

PR #9382, opened 2022-12-19
feat(service-portal): smart solutions refactor

thorkellmani:
## What
Improves the overall quality of the service

* Simplifies types
* More generic
* Shared fetch handling

## Why
It's better to make the services more generic

## Screenshots / Gifs

## Checklist:


libs/clients/smartsolutions/src/lib/smartSolutions.types.ts
```diff
@@ -1,88 +1,68 @@
...
   | 3
-  /** Request contains some field errors */
+  /** External service error */
   | 4
-  /** Unknown error */
+  /** Incomplete service response */
+  | 5
+  /** Request contains some field errors */
+  | 6
+  /** Generic error code */
```
osk, line 67: I'm pretty sure this is a breaking change against the app, we can't mix the `VerifyPassStatusCode` with a new error code

libs/clients/smartsolutions/src/lib/smartSolutions.api.ts
```diff
@@ -49,12 +39,76 @@ export class SmartSolutionsApi {
...
+        category: LOG_CATEGORY,
+      })
+
+      return {
+        error: {
+          serviceCode: 2,
+          message: 'Response object not ok',
+          httpStatus: {
+            status: res.status,
+            statusText: res.statusText,
```
osk, line 79: would it make sense to try and log out the text body here?

libs/clients/smartsolutions/src/lib/smartSolutions.api.ts
```diff
@@ -49,12 +39,76 @@ export class SmartSolutionsApi {
...
+    } catch (e) {
+      this.logger.warn('Unable to parse JSON', {
+        exception: e,
         category: LOG_CATEGORY,
       })
-      return null
+      return {
+        error: {
+          serviceCode: 3,
+          message: 'Json parse failure',
```
osk, line 96:
```suggestion
          message: 'JSON parse failure',
```

libs/clients/smartsolutions/src/lib/smartSolutions.api.ts
```diff
@@ -84,75 +138,42 @@ export class SmartSolutionsApi {
...
+        serviceCode: 99,
+        message: 'Unknown error',
+      },
+    }
   }
 
   async verifyPkPass(
     payload: DynamicBarcodeDataInput,
-  ): Promise<VerifyPassResult | null> {
+  ): Promise<Result<Pass & { valid: boolean }>> {
```
osk, line 176: It might be more clear to split this into it's own type, e.g.
`type PassWithValidation = Pass & { valid: boolean }`

libs/clients/smartsolutions/src/lib/smartSolutions.types.ts
thorkellmani: Yeah, this was typed up while developing, so will fix in a second run through

thorkellmani: The return objects will change as well so the app will have to update anyway

osk (approved):
LGTM 👍

PR #9392, opened 2022-12-20
fix(parental-leave): Make different text for extra rights for single parent in review screen

ylfahfa:
## What

Make sure extra rights for single parent is not called "common" rights (like for two parents)

## Why

So that it reflects that it's extra rights for single parent and not common right that can be divided between two parents

## Screenshots / Gifs

Screenshot 2022-12-20 at 10 35 29

Screenshot 2022-12-20 at 10 36 27

## Checklist:


karenbjorg (approved):
lgtm

PR #9399, opened 2022-12-20
feat: samradsgat project

diljath:
## What

Initial deployment of samradsgatt to dev with infra setup.
Has the beginings of design on frontpage but otherwise an empty project.

## Why

To start devolping samradsgatt

## Checklist:


eirikurn (commented):
Initial review of core files and setup. 

Let me know when you feel like this is ready for main. 👍

apps/samradsgatt/proxy.config.json
```diff
@@ -0,0 +1,6 @@
+{
+  "/app/samradsgatt/api/graphql": {
```
eirikurn, line 2: Should copy this file from the web project. It basically maps "/api" to the graphql api url.

That way you can call the API locally like in production where this project is on /samradsgatt and the api is on the same domain /api.

apps/samradsgatt/server.ts
```diff
@@ -0,0 +1,11 @@
+import { bootstrap } from '@island.is/infra-next-server'
+import proxyConfig from './proxy.config.json'
+
+bootstrap({
+  name: 'samradsgatt',
+  appDir: 'apps/samradsgatt',
+  proxyConfig,
+  externalEndpointDependencies: process.env.BASE_URL
+    ? [process.env.BASE_URL]
+    : [],
```
eirikurn, line 10: BASE_URL is not defined in the infra file. I don't think we need to do a DNS health check against the deployed BASE_URL (or why auth-admin-web does it). Probably fine to remove.

eirikurn (approved):
LGTM with minor comments 🙌

apps/samradsgatt/infra/samradsgatt.ts
```diff
@@ -0,0 +1,65 @@
...
+      },
+      NEXTAUTH_URL: {
+        dev: ref(
+          (ctx) =>
+            `https://${
+              ctx.featureDeploymentName ? `${ctx.featureDeploymentName}-` : ''
+            }loftbru.dev01.devland.is`,
+        ),
+        staging: 'https://loftbru.staging01.devland.is',
+        prod: 'https://loftbru.island.is',
```
brynjarorng, line 32: Is this app supposed to reference the loftbru app here?

apps/samradsgatt/infra/samradsgatt.ts
```diff
@@ -0,0 +1,65 @@
+import { ref, service, ServiceBuilder } from '../../../infra/src/dsl/dsl'
+
+export const serviceSetup = (): ServiceBuilder<'samradsgatt'> =>
+  service('samradsgatt')
+    .namespace('samradsgatt')
+    .liveness('/liveness')
+    .readiness('/readiness')
```
brynjarorng, line 7: I recommend using /liveness here

apps/samradsgatt/infra/samradsgatt.ts
brynjarorng, line 32: Is this also supposed to reference loftbru?

apps/samradsgatt/infra/samradsgatt.ts
```diff
@@ -0,0 +1,65 @@
...
+      max: 30,
+      min: 5,
+    })
+    .resources({
+      limits: { cpu: '400m', memory: '512Mi' },
+      requests: { cpu: '200m', memory: '256Mi' },
+    })
+    .env({
+      BASEPATH: '/samradsgatt',
+      IDENTITY_SERVER_ISSUER_URL: {
```
brynjarorng, line 19: Are you using the IDS? I can't see it being used in the code

brynjarorng (approved):
LGTM(Devops)

PR #9404, opened 2022-12-20
fix(driving-licence): Inconsistent UI state

kristofer-lp8f:
UI goes into inconsistent state in some cases, which makes it appear as a student without permission could get assigned a school.

- Fix inconsistent UI state when going back & forth
- Add e2e tests for application

## Checklist:


eirikurn (commented):
Whoops, did I forget adding the codegen dependency  😅

eirikurn (approved):

apps/system-e2e/src/playwright.config.ts
```diff
@@ -1,11 +1,7 @@
-import type { PlaywrightTestConfig } from '@playwright/test'
-import { devices } from '@playwright/test'
-import './addons'
-/**
- * Read environment variables from file.
- * https://github.com/motdotla/dotenv
- */
-// require('dotenv').config();
+import { PlaywrightTestConfig } from '@playwright/test'
```
pshomov, line 1: I think this `type` thing was important for the bundler to work. Probably to best leave it be.

apps/system-e2e/src/playwright.config.ts
```diff
@@ -15,10 +11,6 @@ const config: PlaywrightTestConfig = {
   /* Maximum time one test can run for. */
   timeout: 90 * 1000,
   expect: {
-    /**
```
pshomov, line 18: I am thinking this file is all about the configuration so it might be a good idea to leave the comments and commented out sections around for people to easily disable/enable them locally instead of needing to dig through the documentation for those.

apps/system-e2e/src/playwright.config.ts
```diff
@@ -42,79 +34,15 @@ const config: PlaywrightTestConfig = {
         ],
         ['html', { open: 'never' }],
       ]
-    : 'html',
+    : 'line',
```
pshomov, line 37: I think at this early stage of the adoption of the testing, it might be better to opt for the more verbose 'html' setup to be more helpful troubleshoot issues.

apps/system-e2e/global.d.ts
```diff
@@ -5,6 +5,8 @@ declare global {
   namespace PlaywrightTest {
     interface Matchers<R, T> {
       toHaveCountGreaterThan(a: number): R
+      toBeApplication(): R
```
pshomov, line 8: Missing parameter

pshomov (approved):

albinagu (approved):

kksteini (approved):
LGTM

PR #9430, opened 2022-12-22
chore(user-profile): More specific handle error logging for userprofile domain

thordurhhh:
## What

* Be more specific in handle error logging for userprofile domain service

## Why

* All the errors fall into the generic "Userprofile error" as things stand now. This will group specific errors within `@category:userprofile-service` 

## Checklist:


libs/api/domains/user-profile/src/lib/userProfile.service.ts
```diff
@@ -30,8 +30,8 @@ export const MAX_OUT_OF_DATE_MONTHS = 6
 const LOG_CATEGORY = 'userprofile-service'
 
 // eslint-disable-next-line
-const handleError = (error: any) => {
-  logger.error('Userprofile error', {
+const handleError = (error: any, details?: string) => {
+  logger.error(details || 'Userprofile error', {
     error: JSON.stringify(error),
     category: LOG_CATEGORY,
   })
```
kristofer-lp8f, line 37:
```suggestion
  logger.error(details || 'Userprofile error', error)
```

kristofer-lp8f (changes requested):
I would have thought the logger shows stack trace and more if we just log the error object directly. If so, the extra detail messages you added might be redundant.

libs/api/domains/user-profile/src/lib/userProfile.service.ts
thordurhhh: But then we will not be able to filter the logs in DD by category. In this case I can go in, and filter with `@category:userprofile-service` to see all the errors coming from this service, which has proven to be a nice debugging tool.

Valur (approved):

kristofer-lp8f (approved):

PR #9438, opened 2022-12-22
feat(admin-portal): Static module switcher with dashboard redirect on single item

saevarma:
# Static module switcher with dashboard redirect on single item

https://www.notion.so/Make-module-switcher-static-if-user-only-has-access-to-one-admin-module-e9151c35c77449c1a0caf2e7aaa84db3
https://www.notion.so/Automatic-redirect-if-user-only-has-access-to-one-admin-module-9089ec7d49fc4bf4971911fbc1f5967e

## What

* Static module switcher with single nav item.
* Redirect from dashboard screen when only single item.
* Fixing font weight on active item in desktop.

## Why

For better UX when only having access to single admin module.

## Screenshots / Gifs

N/A

## Checklist:


apps/portals/admin/src/screens/Dashboard/Dashboard.tsx
```diff
@@ -10,17 +12,32 @@ import {
...
   const { md } = useBreakpoint()
   const topNavigation = useNavigation(TOP_NAVIGATION)
   const bottomNavigation = useNavigation(BOTTOM_NAVIGATION)
+  const navigationItem = getNavigationItemIfSingle(
+    TOP_NAVIGATION,
+    BOTTOM_NAVIGATION,
+  )
+
+  if (navigationItem) {
+    if (navigationItem.path) {
```
snaerseljan, line 37: can be done in a single if

snaerseljan (approved):
Really nice job. Maybe change getNavigationItemIfSingle into a hook as in useNav...

PR #9439, opened 2022-12-22
refactor(admin-portal): Remove dependency on user profile scope.

saevarma:
# Remove dependency on user profile read scope

https://www.notion.so/Remove-user-profile-scope-dependency-20a39240e0ba4bf194c6d8949952ad57

## What

Removing dependency on the user profile read scope.
Remove the `<UserProfileLocale />` to not use any user profile related queries.

## Why

Access to the admin portal depends on access controlled scopes. The user profile scope is not access controlled and thus breaks the authorization in IDS checking if the user has any scope to the client.

We need to address the usage of the locale in our apps to make it actor specific and an automatic scope grant in delegations.

## Screenshots / Gifs

N/A

## Checklist:


snaerseljan (approved):
LGTM! 👍

PR #9449, opened 2022-12-27
fix(air-discount-scheme): Resolve fields and explicit discount code simplification

kksteini:
## What

Resolve flightlegs and travel for discounts.
Introduce, simplify and fix explicit code logic in the resolver/service space.

## Why

Finish the relevant GraphQL contract.
Allow admin measures to reach end user.

## Screenshots / Gifs
image

## Checklist:


rafnarnason (approved):
lgtm

apps/air-discount-scheme/backend/src/app/modules/flight/flight.controller.ts
```diff
@@ -373,6 +373,7 @@ export class PrivateFlightController {
   }
 
   @Get('users/:nationalId/flights')
+  @Scopes(AirDiscountSchemeScope.admin, AirDiscountSchemeScope.default)
```
eirikurn, line 376: Should refactor this controller like I did with PrivateDiscountController, so we have `PrivateFlightController` with this endpoint, and then `PrivateFlightAdminController` with the other endpoints. Then you can have different ApiTags on them and avoid depending on the Admin API in the discount resolver.

libs/api/domains/air-discount-scheme/src/lib/discount/discount.resolver.ts
```diff
@@ -11,16 +11,37 @@ import {
...
+  async resolveFlightLegs(
+    @CurrentUser() user: User,
+    @Parent() relation: Relation,
+  ): Promise<TFlightLeg[]> {
+    // AirDiscountSChemeDiscounts yields discounts for the authenticated user
+    // and any eligible wards (relations).
+    return this.discountService.getFlightLegsByNationalId(
+      user,
+      relation.nationalId,
+    )
```
eirikurn, line 40: I have concerns about this endpoint which can query flight legs for any arbitrary national id. I know the service is inside our VPN and the endpoint is only exposed like this for "child" national ids. But it still unnecessarily opens us up for data leaks in case of VPN breaches.

What's the usage pattern like? Is this heavy? Does the client sometimes request this data and sometimes not? Does the client fetch this data one at a time or for all national ids at the same time?

Depending on the answers, we either might want to add this data to the original discount request and/or update the controller to check if CurrentUser has access to the specified `nationalId`.

apps/air-discount-scheme/backend/src/app/modules/flight/flight.controller.ts
kksteini: Oh yeah, good point. I incorrectly assumed the endpoint was being used.
Will do :+1:

libs/api/domains/air-discount-scheme/src/lib/discount/discount.resolver.ts
kksteini: I'm opting for including this in the response instead. Good call.

apps/air-discount-scheme/backend/src/app/modules/flight/flight.controller.ts
kksteini: Done

libs/api/domains/air-discount-scheme/src/lib/discount/discount.resolver.ts
kksteini: Ok, I included this in the discount response.

There may be a case where flights and flight-legs are present in the system but a person isn't eligible for discounts so I added another endpoint, tied to the user, which only displays that person's flight legs, which is consistent with the current implementation of Air Discount Scheme.

eirikurn (changes requested):
I dove deeper into how this is used in the current ADS. There it is a "flightLegs" field resolver on the user model, and it is queried in a "Usage" component, which also queries the name of each flight leg.

With that knowledge I think it might be a mistake to have flightLegs as part of the Discount resolver, since these are two unrelated models. As you said, the user might not have a discount but they can still query their flight legs, or see their "usage".

I recommend getting rid of the field resolver and expanding the new resolver to return all flight legs for all your relations like it did before. This logic could be in a new endpoint in air-discount-scheme-backend (to return flight legs for all relations).

libs/api/domains/air-discount-scheme/src/lib/flight-leg/flight-leg.service.ts
```diff
@@ -38,35 +41,81 @@ export class FlightLegService {
...
+        user: constructedUser,
+        flightLegs: [],
+      } as Flight
+
+      // We loop through the flightLegs and attach the extra information needed
+      // as well as attaching a reference to the new constructed flight
+      for (const flightLeg of flight.flightLegs ?? []) {
+        const constructedFlightLeg: FlightLeg = {
+          ...flightLeg,
+          travel: `${flightLeg.origin} - ${flightLeg.destination}`,
```
eirikurn, line 92: This makes sense as a field resolver.

libs/api/domains/air-discount-scheme/src/lib/flight-leg/flight-leg.service.ts
```diff
@@ -38,35 +41,81 @@ export class FlightLegService {
...
+          flight: constructedFlight,
```
eirikurn, line 93: We actually only have flightLeg queries so we could in fact get rid of the "flightLegs" field from the flight model to get rid of this circular mess. Or make it nullable. But this is also fine :)

libs/api/domains/air-discount-scheme/src/lib/flight-leg/flight-leg.service.ts
```diff
@@ -38,35 +41,81 @@ export class FlightLegService {
...
+  private async getUserRelations(auth: User): Promise<TUser[]> {
+    const getRelationsResponse = await this.getADSWithAuth(auth)
+      .privateUserControllerGetUserRelations({ nationalId: auth.nationalId })
+      .catch((e) => {
+        this.handle4xx(e)
+      })
+
+    if (!getRelationsResponse) {
+      return []
+    }
```
eirikurn, line 117: Does this endpoint ever return 404?

libs/api/domains/air-discount-scheme/src/lib/flight-leg/flight-leg.service.ts
```diff
@@ -38,35 +41,81 @@ export class FlightLegService {
...
+        ...relation,
+        name: `${relation.firstName} ${relation.lastName}`,
+      }
+      const constructedFlightLegs: FlightLeg[] = []
+
+      // UserInfo in flight.userInfo has gender as string in the generated schema
+      // but is a string union type, hence the `as Flight` coercion
+      const constructedFlight: Flight = {
+        ...flight,
+        user: constructedUser,
```
eirikurn, line 83: Annoying to have to manually query and fill the user here. Could be included in original request or implemented as a field resolver with the dataloader pattern.

eirikurn (approved):
LGTM. Left some nitpicky comments for knowledge sharing and/or future improvement.

libs/api/domains/air-discount-scheme/src/lib/flight-leg/flight-leg.resolver.ts
```diff
@@ -11,16 +11,31 @@ import {
...
+    name: 'airDiscountSchemeUserAndRelationsFlights',
+  })
+  async getFlightLegs(
+    @CurrentUser() user: User,
+    @Info() info: GraphQLResolveInfo,
+  ): Promise<FlightLeg[]> {
+    // See Info()
+    // https://docs.nestjs.com/graphql/resolvers#graphql-argument-decorators
+
+    const depth = detectGraphqlDepthFromInfo(info)
```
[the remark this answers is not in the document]
kksteini: Yeah, then I would like to remove this :)
I did think this was a practical concern and therefore cooked this up.

libs/api/domains/air-discount-scheme/src/lib/flight-leg/flight-leg.service.ts
kksteini: Yeah, I agree.

I tried including this in the original response but didn't like the semantic changes to the user controller by querying flight details from there. 
Neither did I like the whole ADS backend becoming `forwardRef() =>` dependent upon introducing the UserModule to the FlightModule if I remember correctly.

Possibly I could do some code duplication but there is a hefty block of logic associated with getting a user with their calculated benefits.

Or just move this endpoint to the User Controller?

libs/api/domains/air-discount-scheme/src/lib/flight-leg/flight-leg.service.ts
kksteini: If memory serves, yes. When `/forsja` doesn't find any wards it returns a 404 instead of an empty array.
I'll double check just in case.

kksteini: Memory didn't serve. It's a 400.
Changed to a simple handleError.

libs/api/domains/air-discount-scheme/src/lib/flight-leg/flight-leg.service.ts
kksteini: You're right. Made it a field resolver.

PR #9460, opened 2022-12-29
chore(j-s): Remove function handleRemoveFromS3

oddsson:
# Remove function handleRemoveFromS3

[Asana](https://app.asana.com/0/1199153462262248/1203374349779618/f)

## What

Remove function handleRemoveFromS3 and use a new `remove` function instead.

## Why

handleRemoveFromS3 has been deprecated

## Checklist:


apps/judicial-system/backend/src/app/modules/file/file.controller.ts
```diff
@@ -167,7 +167,7 @@ export class FileController {
     CaseNotCompletedGuard,
     CaseFileExistsGuard,
   )
-  @RolesRules(prosecutorRule, representativeRule)
+  @RolesRules(prosecutorRule, representativeRule, registrarRule, judgeRule)
```
oddsson, line 170: This is needed because judges and registrars can upload files in indictments and therefore they should be able to delete files as well

Arnarkari93: You probably need to update the tests for this.

apps/judicial-system/web/src/routes/Court/Indictments/CourtRecord/CourtRecord.tsx
```diff
@@ -38,21 +40,21 @@ import * as constants from '@island.is/judicial-system/consts'
...
   const { transitionCase } = useCase()
 
-  const {
-    files,
-    handleS3Upload,
-    handleRemoveFromS3,
-    handleRetry,
-    allFilesUploaded,
-  } = useS3Upload(workingCase)
+  const { files, handleS3Upload, handleRetry, allFilesUploaded } = useS3Upload(
```
Arnarkari93, line 54: Dose the functionality break if we use the `useS3UploadV2` here?

Arnarkari93 (commented):
One question

apps/judicial-system/web/src/routes/Court/Indictments/CourtRecord/CourtRecord.tsx
oddsson: UseS3UploadV2 is not a drop-in replacement for useS3Upload. You need some extra code for the upload functionality to work if we use v2 but the remove functionality does not break. This PR is just about stop using the remove functionality from the old hook.

oddsson: NB. this pr is one step in a bigger story that is to completely remove useS3Upload from our codebase and use the newer useS3UploadV2 instead. I'm just cutting that story down into smaller pieces.

Arnarkari93: I see, sounds good.

Arnarkari93 (approved):
LGTM

PR #9463, opened 2022-12-30
chore(j-s): Refactoring (renaming)

oddsson:
# Refactoring (renaming)

[Asana](https://app.asana.com/0/1199153462262248/1203621685545204/f)

## What

Change step names from StepOne, StepTwo, etc to Defendant, HearingArrangements etc

## Why

If we ever needed to add a step in the middle of the current steps, having names like StepTwo would not be ideal

## Checklist:


Arnarkari93 (approved):
🦄 😍 🌟 🏩 🥇 💯 👍 🤩

PR #9476, opened 2023-01-02
test(Portals): Tests for filter navigation tree method

snaerseljan:
# Unit tests for filter navigation tree method

## What

Test cases for filterNavigationTree which is used in useNavigation hook

## Why

This filter method is really important for our module and navigation system in our Portals.

## Checklist:


saevarma (approved):
👏 amazing test setup to test a really important nav filtering.

PR #9477, opened 2023-01-02
feat(eslint): Allow fake national IDs

kristofer-lp8f:
- Allow e.g. `101030-*`
- Allow common place-holders such as `111111-*`

Depends on #9484 

## Checklist:


eslint-local-rules.js
```diff
@@ -1,5 +1,6 @@
 'use strict'
 const kennitala = require('kennitala')
+const fakeNationalIdPrefixes = ['010130', /(\d+)\1{6,}/]
```
eirikurn, line 3: The first prefix may match a real person and these do not match some of the newer fake ids which we might need to use in some tests. 

I don't mind this change at all, but I also don't mind using eslint-ignore. Additionally, we could have something like this:

```
import { FakePerson } from '@island.is/testing/fixtures'

// enum access:
FakePerson.gervimadurUtlond // 0101307789
```

Also, these fake ids should only be used in system-e2e tests. I think all other tests should use `createNationalId()`.

eirikurn (approved):
Approving with comment

PR #9505, opened 2023-01-04
fix(parental-leave): new data provider

karenbjorg:
## What

Add new data provider.

## Why

Add new data provider so the application would work correctly.

## Checklist:


obmagnusson (changes requested):
There already is a National registry spouse provider.
[definition](https://github.com/island-is/island.is/blob/main/libs/application/types/src/lib/template-api/shared-api/shared-api-definitions/national-registry-spouse.ts)

 This would just be a duplicate of that one.  Is there some reason you can't use the existing one?

qdd1 (approved):

karenbjorg (commented):
new PR for this issue: https://github.com/island-is/island.is/pull/9548.
Will close this one.

PR #9515, opened 2023-01-05
fix(qa): Fix breaking admin-portal test

kristofer-lp8f:
- Conditionally select delegation IDS login
- Refactoring utilities

## Checklist:


apps/system-e2e/src/support/session.ts
```diff
@@ -90,11 +91,11 @@ async function ensureIDSsession(
 
 export async function session({
   browser,
-  storageState,
   homeUrl,
   phoneNumber,
   authUrl,
   idsLoginOn,
+  storageState = `${homeUrl}-${phoneNumber}`,
```
eirikurn, line 98: Nice!

We might want to add the delegationNationalId parameter here as well, and include it in the default storageState.

Also, do you need to escape homeUrl somehow since it might contain "/", and if I understand this right, it is used to create files on the filesystem.

apps/system-e2e/src/support/login.ts
```diff
@@ -0,0 +1,69 @@
...
+  await btn.click()
+  await page.waitForURL(new RegExp(`${home}|${urls.authUrl}/delegation`), {
+    waitUntil: 'domcontentloaded',
+  })
+
+  // Handle delegation on login
+  if (await page.url().startsWith(urls.authUrl)) {
+    console.log('Still on auth site')
+    const delegations = page.locator('button[name="SelectedNationalId"]')
+    await expect(delegations).toHaveCountGreaterThan(3)
```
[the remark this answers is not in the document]
kristofer-lp8f: No reason.

In the past, in some cases a selector finds more than you intended; in which case, expecting at least 1 element would pass the test, even though the list was empty.

This selector is accurate enough to only require 2 delegations (if 1 or 0, the delegation screen shouldn't appear).

apps/system-e2e/src/tests/admin-portal/smoke/admin-portal.spec.ts
```diff
@@ -11,7 +11,7 @@ test.describe('Admin portal', () => {
     context = await session({
       browser: browser,
       storageState: 'service-portal-afrika.json',
-      homeUrl: `${urls.islandisBaseUrl}/stjornbord`,
+      homeUrl: `/stjornbord`,
       phoneNumber: '0103019',
```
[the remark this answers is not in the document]
kristofer-lp8f: Indeed, would be nice to have the same test data in both environments. Mocking could help here.

You're only commenting on the user, not the change, right?

apps/system-e2e/src/support/login.ts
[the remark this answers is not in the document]
kristofer-lp8f:
```suggestion
    await expect(delegations).toHaveCountGreaterThan(2)
```

apps/system-e2e/src/support/session.ts
```diff
@@ -90,11 +91,11 @@ async function ensureIDSsession(
...
 }: {
   browser: Browser
   storageState?: string
```
kristofer-lp8f, line 101:
```suggestion
  idsLoginOn,
  delegation,
  storageState = `${homeUrl}-${phoneNumber}`,
}: {
  browser: Browser
  delegation?: string,
  storageState?: string
```

apps/system-e2e/src/support/session.ts
kristofer-lp8f: Added delegation to the default

kristofer-lp8f: If a `storageState` contains `/`, new directories are created. This doesn't matter, since the lookup resolves into the directory. I.e. functionally, there is no difference.

There is no mention in the [Playwright docs](https://playwright.dev/docs/api/class-browsercontext#browser-context-storage-state) on whether the directories are created on save. However, in my testing, sub-directories are created and the browser context is indeed stored as expected.

apps/system-e2e/src/addons.ts
```diff
@@ -1,28 +1,24 @@
 import { expect, Locator, Page } from '@playwright/test'
 
-const sleep = (ms: number) => new Promise((r) => setTimeout(r, ms))
-
```
pshomov, line 4: Removing the sleep now leaves this operation polling the DOM potentially thousands of times per second. This is not a good idea in general and I would suspect will have a very real impact when running in k8s. What is the problem with the current code?

apps/system-e2e/src/support/session.ts
eirikurn: Mainly mentioning this because homeUrl usually starts with /, like "/stjornbord". Depending on the fs logic it could cause problems ("folder//file", or resolved to a filesystem root folder because of path.resolve). 

The docs say this: "If path is a relative path, then it is resolved relative to current working directory.". And you're basically passing in an absolute path, which could cause issues now or later, ie trying to create the sessionStore in a root folder on your drive.

apps/system-e2e/src/tests/admin-portal/smoke/admin-portal.spec.ts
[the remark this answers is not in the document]
eirikurn: Mocking is difficult because this data is fetched from the shared identity server. But yeah, the change is solid.

apps/system-e2e/src/support/login.ts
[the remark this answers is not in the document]
eirikurn: That sounds like a weird issue. 

Will this check break for users which only have one delegation (two buttons)?

apps/system-e2e/src/addons.ts
kristofer-lp8f: Indeed, not a good idea. I was getting `execution context changed` errors, and thought this might be it. It wasn't, but I didn't add it back.

Sleeping is back, defaulting to 100 ms

apps/system-e2e/src/support/session.ts
kristofer-lp8f: Good catch. Made some changes in https://github.com/island-is/island.is/pull/9515/commits/f584511fabcd322e98299ab56b18e672647180e4 to counteract this. "Paths" like `folder//file` should resolve to `folder/file`.

Using folders gives clear scoping of who is doing what.

eirikurn (approved):

pshomov (approved):
LGTM

PR #9537, opened 2023-01-06
chore(j-s): Ready For Court Notifications

gudjong:
# Ready For Court Notifications

[Setja READY_FOR_COURT tilkynningar á biðröð](https://app.asana.com/0/1199153462262248/1203640056170084/f)

## What

- Posts "ready for court" notifications on queue.
- Posts "police request" upload to court on queue when a "ready for court" notification is sent for a received case. Previously, the "police request" was uploaded at the same time notifications were being sent in email and sms.
- Email and sms notifications are still sent in one big batch. If at leas one notification succeeds, other failing notifications will not be retried. If all notifications fail, then a retry will be scheduled. Breaking up "ready for court" notifications per recipient would be a good improvement in the future.

## Why

- Less waiting time for the client.
- Notifications and uploads will be retried if they fail.

## Checklist:


Arnarkari93 (approved):
LGTM!

PR #9538, opened 2023-01-06
feat(auth-admin): Add name field to ApiScopeUser

snaerseljan:
# Add name field to apiScopeUser in ids admin

## What

This adds additional field to apiScopeUser, i.e. name field.
- Add name field to model, i.e. db.
- Fix failing server requests in service.
- Added client code to display name field in APIScopeUser.
- Updated search in backend to support case insensitive search for 3 fields, i.e. name, email and nationalId.
- Swapped out scope name for scope display name in client for readability.
- Created migrations.

## Why

NationalId and emails tells us little about the apiScopeUser. Name field makes it easier to manage and read.
This applies as well for scope displayName.

## Screenshots / Gifs
image

image

image

image

## Checklist:


libs/auth-api-lib/migrations/1672932758825-add-api-scope-user-name-field.js
```diff
@@ -0,0 +1,29 @@
...
+          'api_scope_user',
+          'name',
+          {
+            type: Sequelize.STRING,
+            allowNull: true,
+          },
+          { transaction },
+        ),
+      ]),
+    )
```
saevarma, line 17: You can simplify your migration a bit, and also just join those two into a single file. And modify the field to not allow null when you have set default values.
```suggestion
    return queryInterface.sequelize.transaction(async (transaction) => {
        await queryInterface.addColumn(
          'api_scope_user',
          'name',
          {
            type: Sequelize.STRING,
            allowNull: true,
          },
          { transaction },
        )
        
        await queryInterface.sequelize.query(
            `
            UPDATE api_scope_user
            SET name = 'NAME_NOT_SET'
            WHERE name IS NULL`,
            { transaction },
        )
        
        await queryInterface.changeColumn(
          'api_scope_user',
          'name',
          {
              type: Sequelize.STRING,
              allowNull: false,
          },
          { transaction },
        )
        
    })
```

libs/auth-api-lib/src/lib/resources/dto/api-scope-user-update.dto.ts
```diff
@@ -3,6 +3,13 @@ import { IsNotEmpty, IsOptional, IsString } from 'class-validator'
 import { ApiScopeUserAccessDTO } from './api-scope-user-access.dto'
 
 export class ApiScopeUserUpdateDTO {
+  @IsOptional()
+  @IsString()
+  @ApiProperty({
+    example: 'Jane Doe',
+  })
+  name!: string
```
saevarma, line 11: Shouldn't this be undefined if missing? Same with the email property below.
```suggestion
  name?: string
```

libs/auth-api-lib/src/lib/resources/dto/api-scope-user.dto.ts
```diff
@@ -3,6 +3,13 @@ import { IsNotEmpty, IsOptional, IsString } from 'class-validator'
 import { ApiScopeUserAccessDTO } from './api-scope-user-access.dto'
 
 export class ApiScopeUserDTO {
+  @IsOptional()
```
saevarma, line 6: Not optional?

saevarma (approved):
Nice job!
Just few nice to have improvement comments.

PR #9544, opened 2023-01-09
chore(j-s): Optimize cases query

oddsson:
# Optimize cases query

[Asana](https://app.asana.com/0/1199153462262248/1203654762273366/f)

## What

Optimize cases query by creating a new query for the cases list. The old query was fetching whole cases and filtering them, which was _very_ slow. The new query only fetches the data the cases list needs.

## Why

Fetching the list of all cases has become very slow. This PR speeds things up.

### Before 

Screenshot 2023-01-09 at 15 03 25

### After

Screenshot 2023-01-09 at 15 05 57

## Checklist:


libs/judicial-system/types/src/lib/case.ts
```diff
@@ -243,6 +243,35 @@ export interface Case {
   crimeScenes?: CrimeSceneMap
 }
 
+export type CaseListEntry = Pick<
```
Arnarkari93, line 246: I was thi

apps/judicial-system/web/src/utils/hooks/useCase/index.ts
```diff
@@ -271,7 +272,7 @@ const useCase = () => {
 
   const transitionCase = useMemo(
     () => async (
-      workingCase: Case,
+      workingCase: Case | CaseListEntry,
```
Arnarkari93, line 275: consider only accepting `id` from `workingCase`

libs/judicial-system/types/src/lib/case.ts
Arnarkari93, line 246: I think we should consider using the generated type. Maybe we can discuss that with the team?

apps/judicial-system/web/src/routes/Court/components/ReceptionAndAssignment/ReceptionAndAssignment.tsx
```diff
@@ -75,11 +75,11 @@ const ReceptionAndAssignment = () => {
     },
   )
 
-  const receiveCase = async (workingCase: Case, courtCaseNumber: string) => {
+  const receiveCase = async (workingCase: Case) => {
     if (workingCase.state === CaseState.SUBMITTED && !isTransitioningCase) {
       // Transition case from SUBMITTED to RECEIVED when courtCaseNumber is set
       const received = await transitionCase(
-        { ...workingCase, courtCaseNumber },
```
oddsson, line 82: This confuses me, I don't think this is needed. @Arnarkari93 @gudjong do you remember why we did this or is this just unneeded code?

Arnarkari93: I am not sure. Could it be that the new `courtCaseNumber` is not updated for the `workingCase` state? Since `transitionCase` sends a mutation call to the server and updates the local state for `workingCase`

oddsson: That was what I thought but `transitionCase` does not do anything with the courtCaseNumber. I'mma say this is a mistake on our half

Arnarkari93 (approved):

apps/judicial-system/backend/src/app/modules/case/case.controller.ts
```diff
@@ -250,27 +250,6 @@ export class CaseController {
...
-    assistantRule,
-    staffRule,
-  )
-  @Get('cases')
-  @ApiOkResponse({
-    type: Case,
-    isArray: true,
-    description: 'Gets all existing cases',
-  })
-  getAll(@CurrentHttpUser() user: User): Promise<Case[]> {
```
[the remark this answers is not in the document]
gudjong: Nice

gudjong (approved):
Nice

PR #9551, opened 2023-01-09
fix(admin-portal): Add access denied if user has no access

saevarma:
## What

If user managed to be routed to the Admin dashboard, we make sure to render a access denied screen.

Set enabled property of ADS module to be disabled if user doesn't have the scope.

## Why

To handle if user is routed to the admin dashboard with valid login but no access to any admin module.

Note: The module switcher will be updated in a separate PR to handle access denied and single module access better.

## Screenshots / Gifs

## Checklist:


snaerseljan (approved):
Splendid nice work and spot!

PR #9558, opened 2023-01-10
chore(ci): Code coverage - take 2

pshomov:
### Goal 
One of the goals of the Quality Assurance project is to find a way to get a high-level metric of the quality of important parts of our codebase. One such metric, commonly accepted by the industry, is code coverage. Although it is not a perfect metric, we feel it is indicative enough of the trend of internal quality. 

### Current status
Our current code coverage is (besides being broken right now) giving us code coverage at each individual NX workspace/project level. This is helpful in PRs since it provides automated feedback if the author of the PR is lowering or raising the quality of the component they are working on. Overall though it is hard to aggregate what these individual components' quality metrics mean in the context of services that we are running. For example, we have no idea what the code coverage for service `api` or service `web` is.

### Challenges
Code coverage is pretty straightforward in multi-repo setup:
 * All code external to the individual repository is excluded from the quality metrics
 * At each build of the repository, full code coverage is collected for the repo, and that's all
 
In a monorepo, we are however faced with a different situation - libraries are part of the monorepo. Additionally, complexity stems from the fact that it is not practical to build the full set of components at every change - we build only what has changed. Thus the tooling and services we use need to be aware of these monorepo challenges and handle them. That reduces our pool of possible candidates significantly.
To address these issues we have to overcome a few challenges:
 1. Gather code coverage information not only from the code of a component, but also from the code of the libraries in the monorepo it is using
 2. Aggregate the code coverage of a component by combining coverage of test runs from its own tests as well as test runs of other components using it directly on indirectly(obtained using method above)
 3. Provide a way to define aggregates of components' coverage that will be a proxy metric for a service
 4. Provide a way to present the information of code coverage of the individual components as well as the aggregates

### Experiments

After some experimentation we managed to get our `jest` executions to produce code coverage for libraries used during the test run. That required changes to the default jest configuration but seems to be quite doable.
To maintain a long term view on the code coverage information we turned to third-party providers. 
The first candidate was [Sonar Cloud](https://www.sonarsource.com/products/sonarcloud/). It is attractive since it is a complete software quality as a service package that is quite good and would allow us to adopt other quality metrics beyond code coverage - code complexity, maintainability, duplication, security scanning, etc. Unfortunately, Sonar Cloud gathering and processing of data is incredibly slow and hard to configure. We need quite large computational resources to be able to process each individual component. Additionally, we cannot easily aggregate the individual code coverage report to provide a combined number. All in all, Sonar Cloud processing is quite impractical and would not give us what we need anyway.
Our other candidate is [Codecov](https://about.codecov.io). They are a SaaS which specializes in code coverage and seem to have best-in-class monorepo support - carry-forward flags, as well as good Github integration. They do support aggregation of code coverage in something they call "component" and that's what we will be using. The user interface supporting these "components" is quite sparse at the moment, but there are indications that they are building it up.

### Limitations
The UI for components in codecov is pretty minimal at the time of this writing. It consists of Github PR checks that show this quality. 

chore_ci___Code_coverage_-_take_2_by_pshomov_·_Pull_Request__9558_·_island-is_island_is

We are betting on a future update that will make this feature much more prominent UI-wise, which seems to be indicated in the discussion [here](https://community.codecov.com/t/components-feature-ui/4050).

### Next steps

This PR is covering just the infrastructure work and defines the code coverage aggregation for a couple of services. This is just the beginning and it is probably best that service owners or @island-is/core create more and better aggregate definitions.


gudjong (approved):
LGTM

scripts/ci/30_test.sh
```diff
@@ -1,4 +1,12 @@
 #!/bin/bash
 set -euxo pipefail
 
-SERVERSIDE_FEATURES_ON=\"\" yarn run test "${APP}" --runInBand
\ No newline at end of file
+EXTRA_OPTS=""
+
+projects_uncollectible_coverage=("contentful-translation-extension" "application-templates-no-debt-certificate" "api-domains-email-signup" "skilavottord-web" "shared-babel")
```
pshomov, line 6: These projects refused to provide code coverage information and generated an error when tried to do so. I attempted downgrading and upgrading `jest` & co. but did not help. Leaving those as a future task hoping that upgrades in the future might reveal what the issue is.

brynjarorng (approved):
LGTM

busla (approved):
LGTM

HelgiHH (approved):

eirikurn:
Sounds great.

As a developer I'll probably continue to look at the flag-based results, but can you explain why those numbers usually don't make any sense? Like for this PR, there are huge changes in libs which you aren't touching:

image

My main concern with components is that many of our larger components aren't a big focus for testing effort.

- The service-portal is mostly tested with system-e2e tests instead of jest tests.
- The graphql api is composed of a ton of domains, most of which are just passing data directly from other services.
- The application system API should have excellent test coverage, but it also loads a ton of template libs which are tested differently (eg like parental leave with system-e2e). Can we exclude the template libs from the component for codecov reasons?
- The web project is mostly one workspace right now, but also tricky to test with jest.

If I'm not mistaken we can't get system-e2e test coverage into this codecov (if it even makes sense). For the jest tests the focus is on our REST API tests and shared libraries. These should have high code coverage but they're usually in individual NX workspaces.

Still interesting if we can easily create components which combine codecov results for stuff like "Microservices", "React libraries", "NestJS libraries", "Island-UI", etc.

apps/air-discount-scheme/api/jest.config.js
```diff
@@ -1,8 +1,10 @@
...
-  preset: '../../../jest.preset.js',
-  coverageDirectory: '../../../coverage/apps/air-discount-scheme/api',
-  setupFiles: ['./test/environment.jest.ts'],
-  globals: { 'ts-jest': { tsconfig: '<rootDir>/tsconfig.spec.json' } },
+  preset: './jest.preset.js',
+  rootDir: '../../..',
+  roots: [__dirname],
+  coverageDirectory: '<rootDir>/coverage/apps/air-discount-scheme/api',
+  setupFiles: [`${__dirname}/test/environment.jest.ts`],
+  globals: { 'ts-jest': { tsconfig: `${__dirname}/tsconfig.spec.json` } },
```
eirikurn, line 7: Can you explain why we need to move the root?

This will definitely conflict with future NX migrations (not a big issue though).

codecov.yml
```diff
@@ -1,9 +1,35 @@
...
+        target: auto
+        informational: true
+        branches:
+          - '!main'
+  individual_components:
+    - component_id: api-graphql # this is an identifier that should not be changed
+      name: Portal GraphQL
+      paths:
+        - apps/api/**
+        - libs/api/domains/**
```
eirikurn, line 25: Ah, so we can easily list which workspaces to combine in a single component? That's nice!

Can we have the same workspace in more than one component?

codecov.yml
```diff
@@ -1,9 +1,35 @@
...
+    - component_id: application-system-api-graphql # this is an identifier that should not be changed
+      name: Application System API
+      paths:
+        - apps/application-system/api/**
+        - libs/application/api/**
+        - libs/application/core/**
+        - libs/application/graphql/**
+        - libs/application/template-api-modules/**
+        - libs/application/template-loader/**
+        - libs/application/templates/**
```
eirikurn, line 35: So we can just skip the templates folder from the AS API component

apps/air-discount-scheme/api/jest.config.js
eirikurn: Worse yet ... new apps/libs will not be correctly configured...

Do we need to create our own templates or lint this somehow?

eirikurn (approved):
Only reviewed the files outside the apps and libs folder. Great job!

Remember to check for new projects.

apps/air-discount-scheme/api/jest.config.js
pshomov: Moving the root so jest sees and generates code coverage across workspaces/projects. Not sure how these come to be but yes - we need future ones to be similar.

codecov.yml
pshomov: Yes.

pshomov:
> Looks awesome! Thanks for digging deep into this topic 🙏
> 
> As a developer I'll probably continue to look at the flag-based results, but can you explain why those numbers usually don't make any sense? Like for this PR, there are huge changes in libs which you aren't touching:
> 
> image
> 
> My main concern with components is that many of our larger components aren't a big focus for testing effort.
> 
> * The service-portal is a UI project which is mostly tested with system-e2e tests instead of jest tests.
> * The graphql api is composed of a ton of domains, most of which are just passing data directly from other services.
> * The application system API should have excellent test coverage, but it also loads a ton of template libs which are tested differently (eg like parental leave with system-e2e). Can we exclude the template libs from the component for codecov reasons?
> * The web project is mostly one workspace right now, but is another UI project which is tricky to test with jest. More value in using system-e2e.
> 
> If I'm not mistaken we can't get system-e2e test coverage into this codecov (if it even makes sense). For the jest tests the focus is on our REST API tests and shared libraries. These should have high code coverage but they're usually in individual NX workspaces.
> 
> Still interesting if we can easily create components which combine codecov results for stuff like "Microservices", "IDS/Auth services", "React libraries", "NestJS libraries", "Island-UI", etc.

I think the numbers currently you see are out of wack due to the fact we have not submitted code coverage info to codecov.io in quite a while. I hope the number will become a lot more meaningful when we start using the service continously.

The code coverage for codecov "components" is precisely to address our code coverage aggregation needs.

PR #9567, opened 2023-01-11
feat(auth): Enable filtering delegations with requested scopes

valurefugl:
## What

Add an optional parameter with requested scopes to get available delegations.

## Why

So we can show only those delegations that are available to the client.

## Screenshots / Gifs

## Checklist:

### DIFF
--- a/libs/auth-api-lib/src/lib/delegations/delegations-incoming.service.ts
+++ b/libs/auth-api-lib/src/lib/delegations/delegations-incoming.service.ts
@@ -12,7 +12,7 @@ import { DelegationsIncomingRepresentativeService } from './delegations-incoming
 import { DelegationType } from './types/delegationType'
 import { DelegationsIncomingWardService } from './delegations-incoming-ward.service'
 import { ApiScope } from '../resources/models/api-scope.model'
-import { Op } from 'sequelize'
+import { WhereOptions } from 'sequelize'
 import { ClientAllowedScope } from '../clients/models/client-allowed-scope.model'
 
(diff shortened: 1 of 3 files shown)


libs/auth-api-lib/src/lib/delegations/delegations-incoming.service.ts
```diff
@@ -101,10 +101,14 @@ export class DelegationsIncomingService {
   async findAllAvailable(
     user: User,
     delegationTypes?: DelegationType[],
+    requestedScopes?: string[],
```
saevarma, line 104: ~When working with few optional parameters it can be nice to switch to object, so when the function is called you don't have to pass undefined as a parameter, i.e.:
But it requires updating the code where the delegationTypes are used but it should be just one or two places.

```
interface FindAvailableInput {
  user: User
  delegationTypes?: DelegationType[]
  requestedScopes?: string[]
}

async findAvailable({
  user.
  delegationTypes
  requestedScopes
}: FindAvailableInput): Promise<MergedDelegationDTO[]> {
  ....
}

// Then call it in the controller without delegationTypes.
findAllAvailable({user, requestedScopes})
```

saevarma (approved):
Nice!
Just on nice to have feedback.

libs/auth-api-lib/src/lib/delegations/delegations-incoming.service.ts
valurefugl: Fixed.

PR #9568, opened 2023-01-11
fix(ads): Display Discount Model in requests instead of empty array

kksteini:
# What changes?
The `airDiscountSchemeDiscounts` response was being filtered out based on credits linked to the normal discount. The connectionDiscountCodes are outside of this credit system as they are considered a continuation of a credited flight leg.
Therefore, the discount response was being filtered out completely, prematurely.

I propose an empty code response on the discount instead.

# Screenshots
## How the system acted before
A user who is eligible for connectionDiscounts has used all of their 6 normal legs. The logic currently filters out the discount response which is incorrect.
image

## When all discounts used but there are still discount codes
This PR addresses the fact that if all discounts where used the connectionDiscountCode array would get dropped in the filtering which is entirely not what we want. Instead we erase the generated discount code and return it as an empty string (easily checked and verifiable against the flightLegs query on 6 legs).

This example has more flight legs than currently possible but the fact still remains that there is no credit but there are loads of flights eligible for a connecting flight, shown in the connectionDiscountCodes array 
image

## When no rights
When there are no rights we simply return an empty discount code, something that can easily be checked against:
image

## Checklist:


libs/api/domains/air-discount-scheme/src/lib/discount/discount.service.ts
```diff
@@ -39,6 +39,13 @@ export class DiscountService {
     return credit >= 1
   }
 
+  processDiscount(discount: TDiscount): TDiscount {
+    if (!this.discountIsValid(discount)) {
+      discount.discountCode = ''
```
eirikurn, line 44: I'd suggest setting discountCode to null instead. Feels more explicit then a magic string ""

kksteini: I agree. Done
image

libs/api/domains/air-discount-scheme/src/lib/models/discount.model.ts
```diff
@@ -9,8 +9,8 @@ export class Discount {
   @Field(() => ID)
   nationalId!: string
 
-  @Field()
-  discountCode!: string
+  @Field(() => String, { nullable: true })
+  discountCode!: string | null
```
kristofer-lp8f, line 13: "I assure you, it's never not there... except when it is :sweat_smile: "

libs/api/domains/air-discount-scheme/src/lib/discount/discount.service.ts
```diff
@@ -39,6 +39,13 @@ export class DiscountService {
     return credit >= 1
   }
```
kristofer-lp8f, line 41: In lint warnings above:
Maybe change to `Error` or `unknown`?

libs/api/domains/air-discount-scheme/src/lib/models/discount.model.ts
kksteini: I thought `?` in the typing system meant `undefined`.
I'd have it either be string or null.
Am I mistaken? :sweat_smile:

libs/api/domains/air-discount-scheme/src/lib/discount/discount.service.ts
kksteini: :eyes:

libs/api/domains/air-discount-scheme/src/lib/models/discount.model.ts
kksteini: Alternatively, do you have suggestions on how to represent it better?

albinagu (approved):
lgtm

libs/api/domains/air-discount-scheme/src/lib/discount/discount.service.ts
kksteini: Changed the types to `(FetchError | ApolloError): void`

libs/api/domains/air-discount-scheme/src/lib/models/discount.model.ts
kristofer-lp8f: Here  I was pointing out the contradiction between the [non-null operator](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#non-null-assertion-operator-postfix-) and typing as optional `null`.

If the type-checker doesn't complain and the tests covering this passes, then I don't really mind it.

kksteini: I made the default value null instead of typing with `?`
I also fixed the tests when I realized they were a bit off on what they represented.

Tests and an additional smoketest working as expected.

Thanks for the tip :+1:

PR #9571, opened 2023-01-11
chore(j-s): Move RECEIVE transition from frontend to backend

oddsson:
# Move RECEIVE transition from frontend to backend

[Asana](https://app.asana.com/0/1199153462262248/1203665733665694/f)

## What

Move RECEIVE transition from frontend to backend

## Why

Users were sometimes experiencing issues where cases were not being put in correct state. Transitioning cases on the backend makes the transition more stable.

## Checklist:


apps/judicial-system/backend/src/app/modules/case/case.controller.ts
```diff
@@ -27,11 +27,13 @@ import {
 } from '@island.is/dokobit-signing'
 import {
   CaseState,
+  CaseTransition,
   CaseType,
   completedCaseStates,
   indictmentCases,
   investigationCases,
   isIndictmentCase,
+  NotificationType,
```
gudjong, line 36: Consider removing unused import

apps/judicial-system/backend/src/app/modules/case/case.controller.ts
```diff
@@ -179,7 +181,22 @@ export class CaseController {
...
+    return Promise.all([
+      update,
+      ...(caseToUpdate.courtCaseNumber && theCase.state === CaseState.SUBMITTED
+        ? [
+            this.transition(caseId, user, theCase, {
+              transition: CaseTransition.RECEIVE,
+            }),
+          ]
+        : []),
+    ]).then((values) => values[0])
```
gudjong, line 199: This implementation will not perform both updates in a single transaction. Although less likely to cause problems, this does not completely fix the issue we are trying to prevent.

Consider modifying the code to perform the two updates in a single database transaction in the case service.

apps/judicial-system/web/src/routes/Court/components/CourtCaseNumber/CourtCaseNumber.tsx
```diff
@@ -61,7 +63,7 @@ const CourtCaseNumber: React.FC<Props> = (props) => {
 
     await updateCase(id, update)
     if (update.courtCaseNumber) {
-      receiveCase(workingCase, update.courtCaseNumber)
+      sendNotification(workingCase.id, NotificationType.RECEIVED_BY_COURT)
```
gudjong, line 66: This will also send a notification when the court case number is changed which we do not want to do.

Since the transition of the case has moved to the server, it would make sense to send the notification from there as well.

Consider moving the notification to the server. Alternatively, we could send the notification only when the state was SUBMITTED before the court case number update.

gudjong (changes requested):

apps/judicial-system/web/src/routes/Court/components/CourtCaseNumber/CourtCaseNumber.tsx
```diff
@@ -1,7 +1,11 @@
 import React from 'react'
 import { useIntl } from 'react-intl'
 
-import { CaseState, isIndictmentCase } from '@island.is/judicial-system/types'
+import {
+  CaseState,
+  isIndictmentCase,
+  NotificationType,
```
gudjong, line 7: Consider removing unused import.

gudjong (changes requested):

Arnarkari93 (approved):

apps/judicial-system/backend/src/app/modules/case/case.service.ts
```diff
@@ -559,6 +568,10 @@ export class CaseService {
         ) {
           await this.fileService.resetCaseFileStates(theCase.id, transaction)
         }
+
+        if ((update as { [key: string]: string }).state) {
```
gudjong, line 572: We forgot to check if state is RECEIVED

gudjong (changes requested):

gudjong (approved):

PR #9572, opened 2023-01-11
fix(parental leave): remove apply option personal allowance single parent

jhc-deloitte:
## What
If a parent is not a SPOUSE, that parent should not be allowed to get the other parents personal allowance.
So that input field has been removed. 

## Why
The see the reason above.

## Checklist:


karenbjorg (approved):
lgtm

PR #9577, opened 2023-01-11
chore(portals): Remove support for array of route paths

eirikurn:
## What

Remove support for arrays of paths in PortalRoute.

## Why

Removing to simplify code. Wasn't used anywhere.

Was probably typed this way to match react-router v4-5 but arrays are not supported in v6.

## Checklist:


saevarma (approved):
Happy to get rid of the array, it didn't make sense for me 🎉

PR #9579, opened 2023-01-12
chore(service-portal): Move regulations to admin portal

thordurhhh:
## What

* Move regulations over to the admin-portal

## Why

* We have been waiting for a permanent home for the regulation lib. It's here 😃 

## Note

* This is a PR to `regulations/admin` which will eventually be merged to main. But not yet.

I removed `/libs/service-portal/settings/access-control` from codeowners since it's not available on the base branch `regulations/admin` and the PR will fail. Will be fixed when merging to main eventually.

## Checklist:


Valur (approved):

apps/portals/admin/src/auth.ts
```diff
@@ -23,6 +23,8 @@ if (userMocked) {
       'profile',
       AdminPortalScope.delegations,
       AdminPortalScope.airDiscountScheme,
+      AdminPortalScope.regulationAdmin,
+      AdminPortalScope.regulationAdminManage,
```
eirikurn, line 27: Do they need two "roles" for this admin?

eirikurn (approved):
Amazing job!

Don't forget to create the scopes in all environments.

apps/portals/admin/src/auth.ts
thordurhhh: Yes, good question, we actually do need two roles.

* One is to create new regulations, and can not see other regulations. (`AdminPortalScope.regulationAdmin`)
* The other can create, manage, see all regulations and publish regulations. (`AdminPortalScope.regulationAdminManage`)

PR #9589, opened 2023-01-12
feat(core): Enforce monorepo boundaries

GunnlaugurG:
# Added Nx tags to every project in the repo

## What

I added a tag in every project.json file that indicates the scope of the project. 
Lint will now break when a project imports a library that it shouldn't be able to depend on ("@nrwl/nx/enforce-module-boundaries")

## Why

This is done so projects can not import projects which they shouldn't depend on (Like react project importing a Nestjs library).

## Screenshots / Gifs

image

## Checklist:


eirikurn (changes requested):
Nice job! I'll push some fixes, including the one mentioned here.

PR #9602, opened 2023-01-13
feat: add metrics to license-service and improve docs

osk:
# Add metrics to license service and improve documentation

## What

Add counters and timers to calls in license service.

## Why

Get stats about actual usage from service portal and apps along.

## Checklist:


libs/infra-metrics/src/lib/dogStatsD.ts
```diff
@@ -1,4 +1,8 @@
-import { StatsD, ClientOptions } from 'hot-shots'
+import { StatsD, ClientOptions, Tags } from 'hot-shots'
+
+export const METRICS_PROVIDER = 'Metrics'
```
osk, line 3: Is this the correct place for this?

libs/infra-metrics/src/lib/dogStatsD.ts
```diff
@@ -7,4 +11,13 @@ export class DogStatsD extends StatsD {
...
   }
+
+  static timer() {
+    return process.hrtime.bigint()
+  }
+
+  static duration(start: bigint): number {
+    const diff = process.hrtime.bigint() - start
+    return Number(diff) / 1e6
+  }
```
osk, line 22: Is this the correct place for this? Or some `util`?

osk:
Unsure about the metric naming, would be great to get input from devops.

Same with timing metrics ([example](https://github.com/island-is/island.is/pull/9602/files#diff-d4266a6668a7c159a8190ecea782542b289fd85ccfeaba8c6cbd099d5f847fbfR232-R242)), could this help? In the end the http request (which is tracked) would be reponsible for most of the time, but this would time getting a license with all that it entails.

libs/api/domains/license-service/src/lib/licenseService.service.ts
```diff
@@ -213,8 +228,20 @@ export class LicenseServiceService {
...
-              await licenseService.getLicense(user, locale, licenseLabels),
+            async () => {
+              const start = DogStatsD.timer()
+              const result = await licenseService.getLicense(
+                user,
+                locale,
+                licenseLabels,
+              )
+              this.metrics.timing(
+                `all_licenes.license.duration`,
```
peturgq, line 239: all_licences*

handbook/technical-overview/devops/metrics.md
```diff
@@ -20,4 +16,70 @@ We are collecting metrics at three distinct levels:
...
+
+### Naming metrics
+
+Metrics should be named with ASCII letters, underscores and periods. Periods are used to create a hierarchy that always starts with `islandis.`, provided by metrics client. Following that there should be a prefix for the given service and more detailed name for the metric in question. See more in ["What best practices are recommended for naming metrics and tags?" from DataDog](https://docs.datadoghq.com/developers/guide/what-best-practices-are-recommended-for-naming-metrics-and-tags/).
+
+### Example
+
+```typescript
+// Given this client
+const metrics = new DogStatsD({ prefix: `my-service.` })
```
eirikurn, line 40: Should show the import of DogStatsD.

handbook/technical-overview/devops/metrics.md
```diff
@@ -20,4 +16,70 @@ We are collecting metrics at three distinct levels:
...
+- counters
+- timers
+
+**Counters** are used to count the number of times an event has occurred. For example: "number of successful HTTP requests", "number of failed HTTP requests", etc.
+
+**Timers** are used to measure the time it takes to perform an operation in milliseconds. For example: "time it takes to generate a response", "time it takes to generate a PDF", etc.
+
+### Naming metrics
+
+Metrics should be named with ASCII letters, underscores and periods. Periods are used to create a hierarchy that always starts with `islandis.`, provided by metrics client. Following that there should be a prefix for the given service and more detailed name for the metric in question. See more in ["What best practices are recommended for naming metrics and tags?" from DataDog](https://docs.datadoghq.com/developers/guide/what-best-practices-are-recommended-for-naming-metrics-and-tags/).
```
eirikurn, line 34: A bit confusing that the example below does not follow the documented hierarchy, because the DogStatsD class automatically adds the "islandis" prefix. Worth explaining.

handbook/technical-overview/devops/metrics.md
```diff
@@ -20,4 +16,70 @@ We are collecting metrics at three distinct levels:
...
+} else {
+  metrics.increment('request.error', { my: 'tag' })
+}
+```
+
+### Tags
+
+Metrics can be tagged which can help categorizing and creating relevant filters in DataDog.
+
+Beware that each new tag will incur extra cost so use existing tags or create tags that will be used.
```
eirikurn, line 63:
```suggestion
{% hint style="warning" %}
Each combination of metric name and tag values is counted as a custom metric. If you have a lot (hundreds or thousands) of custom metrics they might incur extra hosting costs. Note that not every tag is counted as a new metric. See more in ["Custom Metrics Billing" from DataDog](https://docs.datadoghq.com/account_management/billing/custom_metrics).
{% endhint %}
```

libs/api/domains/license-service/src/lib/licenseService.service.ts
```diff
@@ -262,6 +289,13 @@ export class LicenseServiceService {
         fetch,
         payload: licenseDataFromService?.payload ?? undefined,
       }
+
+      if (combined.fetch.status === GenericUserLicenseFetchStatus.Error) {
+        this.metrics.increment(`all_licenes.license.error`, 1, metricsTags)
+      } else {
+        this.metrics.increment(`all_licenes.license.ok`, 1, metricsTags)
```
eirikurn, line 296: "licenses" typo in all of the metric names.

libs/api/domains/license-service/src/lib/licenseService.service.ts
```diff
@@ -353,21 +418,35 @@ export class LicenseServiceService {
       this.cacheManager,
     )
 
+    const metricsTags = this.getMetricsTags(licenseType)
+    this.metrics.increment(`pkpass_qrcode`, 1, metricsTags)
+
     if (licenseService) {
+      const start = DogStatsD.timer()
       pkpassQRCode = await licenseService.getPkPassQRCode(
```
eirikurn, line 426: I'd suggest having a timing helper which wraps promises:

```ts
pkpassQRCode = await this.metrics.measure(
  'pkpass_qrcode.duration',
  metricsTags,
  licenseService.getPkPassQRCode(...)
)
```

libs/infra-metrics/src/lib/dogStatsD.ts
eirikurn: Not really, since it is not provided by infra-metrics and this library is not (necessarily) a nest library.

You can move this to your other module or instantiate DogStatsD manually. But it would be awesome to create a new metric nest module which can be used like this:

```
import { MetricsModule, MetricsService } from '@island.is/nest/metrics'

@Module({
  imports: [
    MetricsModule({ prefix: 'license-service' }),
  ],
})
class MyModule {}

// elsewhere
class MyService {
  constructor(private metrics MetricsService) {}

  yo() {
    this.metrics.increment('yo')
  }
}
```

The MetricsService would abstract the StatsD client and just expose what we need.

libs/infra-metrics/src/lib/dogStatsD.ts
eirikurn: Fancy hrtime bigint logic! Does StatsD support this precision? :)

This is a fine location but might change depending on my other suggestions.

handbook/technical-overview/devops/metrics.md
```diff
@@ -1,14 +1,10 @@
 # Metrics
 
-Metric is one of the foundational pieces of telemetry we need to understand issues at runtime. It is theresfore deemed one of the core responsibilities of developers to provide metrics in their services.
+Metric is one of the foundational pieces of telemetry we need to understand issues at runtime. It is therefore deemed one of the core responsibilities of developers to provide metrics in their services.
```
eirikurn, line 3: Thanks for updating the docs! 🫡🙏

libs/api/domains/license-service/src/lib/licenseService.service.ts
```diff
@@ -198,6 +210,9 @@ export class LicenseServiceService {
         locale,
       )
 
+      const metricsTags = this.getMetricsTags(license.type)
+      this.metrics.increment('all_licenes.license', metricsTags)
```
eirikurn, line 214: Some of these metrics feel a bit redundant with the enhanced fetch metrics (timings and success/failure), but I guess they make it easier to assemble a focused "license dashboard".

libs/api/domains/license-service/src/lib/licenseService.service.ts
eirikurn: This reminds me that another fancy option is to track classes and function calls with APM:

https://github.com/DataDog/dd-trace-js/blob/master/docs/API.md#manual-instrumentation

This adds a span to the APM traces (and also supports all kinds of filtering and metrics). Main difference is that APM traces are expensive and you may lose data under load so they're not perfect for high-load metrics and probably not for this use case. But they can be very useful for debugging.

eirikurn (changes requested):
Awesome! Love the docs update.

libs/infra-metrics/src/lib/dogStatsD.ts
osk: No, we turn it into regular ol' number before sending it off.
[Node 18 deprecates `process.hrtime()`](https://nodejs.org/docs/latest-v18.x/api/process.html#processhrtimetime)

libs/infra-metrics/src/lib/dogStatsD.ts
osk: OK, did think about it and it's The Right Thing To Do™️ , I'll add it

handbook/technical-overview/devops/metrics.md
osk: Will update after adding metrics module

libs/api/domains/license-service/src/lib/licenseService.service.ts
osk: This makes it more explicit when we create&verify license and succeed/fail doing so. It will help in creating an explicit dashboard for the licenses which we've wanted for some time.

But I'm not experienced enough with this in production to know if this is the best way to do it, any suggestions welcome!

saevarma:
@osk is this still relevant?

andes-it:

andes-it:

PR #9608, opened 2023-01-13
fix(ap): Delegation scope should be 1 level deep.

eirikurn:
## What

The delegation scope used in the admin portal should work 1 level deep.

## Why

To match the existing scope.

## Checklist:


saevarma (approved):
😅 Nice catch!

PR #9609, opened 2023-01-13
fix(ap): Hide actor button in user menu.

eirikurn:
## What

Hide actor button in admin portal user menu.

Also fixes the location of the UserMenu on desktop.

## Why

The admin portal is designed for government organisations. Users are unable to log in as themself, so we don't want users to "switch" to themself while inside the admin portal.

## Screenshots / Gifs

image

## Checklist:


saevarma (approved):
💯

PR #9634, opened 2023-01-17
fix(core): Revert forbidUnknownValues

eirikurn:
## What

Revert the ValidationPipe forbidUnknownValues value to false.

## Why

Upgrading to class-validator 0.14 changed this property to true, to address a security report that wouldn't close even though the author had [tried for a year](https://github.com/typestack/class-validator/issues/1422#issuecomment-1317953863) to explain that the report was invalid and already fixed elsewhere. So it should be safe to change this value back.

## Checklist:


brynjarorng (approved):
LGTM

snaerseljan (approved):
Nice

PR #9650, opened 2023-01-17
fix(j-s): Conditional text for secondary button

unakb:
Text on secondary button of hearing arrangements notification modal should only display "send later" if the notification has never been sent before. Otherwise the modal only appears the first time the user continues after changing the time, so "send later" creates confusion as there is no for the user to trigger a resend

## What

Display different text on modal button according to whether hearing arrangements time was changed or not

## Why

Avoid confusion over ability to resend notification 

## Screenshots / Gifs

Screenshot 2023-01-17 at 22 08 51
Screenshot 2023-01-17 at 22 10 17

## Checklist:

### DIFF
--- a/apps/judicial-system/web/src/routes/Court/InvestigationCase/HearingArrangements/HearingArrangements.strings.ts
+++ b/apps/judicial-system/web/src/routes/Court/InvestigationCase/HearingArrangements/HearingArrangements.strings.ts
@@ -136,7 +136,8 @@ export const icHearingArrangements = {
     secondaryButtonText: {
       id:
         'judicial.system.core:investigation_cases_hearing_arrangements.modal.secondary_button_text',
-      defaultMessage: 'Nei',
+      defaultMessage:
+        'Nei{courtDateHasChanged, select, true {} other {, senda seinna}}',
       description:
         'Notaður sem texti í "Nei" takkann í tilkynningaglugganum á fyrirtöku skrefi í rannsóknarheimildum.',
     },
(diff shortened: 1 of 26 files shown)


apps/judicial-system/web/src/routes/Court/InvestigationCase/HearingArrangements/HearingArrangements.strings.ts
```diff
@@ -136,7 +136,8 @@ export const icHearingArrangements = {
     secondaryButtonText: {
       id:
         'judicial.system.core:investigation_cases_hearing_arrangements.modal.secondary_button_text',
-      defaultMessage: 'Nei',
+      defaultMessage:
```
Arnarkari93, line 139: Adding a param to a contentful string requires the id to be updated.

Arnarkari93 (changes requested):

apps/judicial-system/web/src/routes/Court/InvestigationCase/HearingArrangements/HearingArrangements.strings.ts
unakb: Even if the contentful string hadn't been added yet?

Arnarkari93: Discussed offline. The string is not yet contentful so it is fine to not update the id

Arnarkari93 (approved):

PR #9659, opened 2023-01-18
chore(j-s): Don't fetch case on users route

oddsson:
# Don't fetch case on users route

[Asana](https://app.asana.com/0/1199153462262248/1203747284457174/f)

## What

Don't fetch case on users route

## Why

The users routes (/notendur/*) use the `FormProvider`. `FormProvider` tries to fetch a case when it's rendered, which doesn't make sense for the users routes.

## Checklist:

### DIFF
--- a/apps/judicial-system/backend/src/app/messages/notifications.ts
+++ b/apps/judicial-system/backend/src/app/messages/notifications.ts
@@ -410,6 +410,20 @@ export const notifications = {
         'Texti í pósti til fangelsis sem tilgreinir hver verjandi sakbornings er',
     },
   }),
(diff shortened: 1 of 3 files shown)


Arnarkari93 (commented):
I think we should consider refactoring the code so we don't wrap the users routes with the `<FormProvider />`. We could create an small component what conditional wraps the `<FormProvider />' like:

```js
 const MaybeFormProvider = ({children}) => {
  const router = useRouter()
  return router.pathname.includes(USERS_ROUTE) ? children : (
    <FormProvider>{children}</FormProvider>
  )
}
```

PR #9677, opened 2023-01-19
feat(j-s): modified notification to queue

Arnarkari93:
# Modified Notification

[Setja MODIFIED tilkynningar á biðröð](https://app.asana.com/0/1199153462262248/1203640056170090/f)

## What

- Sends modified notifications from the server instead of the client after "to dates" are modified in restriction cases.
- Adds modified notifications to queue for later processing instead of sending them out immediately while the client waits.

## Why

- More robust notifications.

## Checklist:


apps/judicial-system/backend/src/app/modules/case/test/caseController/createCourtCase.spec.ts
```diff
@@ -145,7 +145,6 @@ describe('CaseController - Create court case', () => {
         where: {
           id: caseId,
           isArchived: false,
-          state: { [Op.not]: CaseState.DELETED },
```
oddsson, line 148: There is an unused import in this file, please consider removing it

apps/judicial-system/backend/src/app/modules/case/test/caseController/transition.spec.ts
```diff
@@ -245,7 +245,6 @@ describe('CaseController - Transition', () => {
               order,
               where: {
                 id: caseId,
-                state: { [Op.not]: CaseState.DELETED },
```
oddsson, line 248: There is an unused import in this file, please consider removing it

apps/judicial-system/web/src/routes/Shared/SignedVerdictOverview/SignedVerdictOverview.tsx
```diff
@@ -481,12 +481,7 @@ export const SignedVerdictOverview: React.FC = () => {
       return false
     }
 
-    await sendNotification(workingCase.id, NotificationType.MODIFIED)
```
oddsson, line 484: There are two unused import in this file, please consider removing them

oddsson (changes requested):
There are some unused imports. Otherwise, nice! 💪

oddsson (approved):

PR #9692, opened 2023-01-20
feat(transport authority change operator): Add validation + Cleanup

johannaagma:
## What

* (Update API version for VehicleOperators)
* Add validation to Change Operator of Vehicle (per vehicle and whole application before payment)
* Fix the way we call VehicleOperators API, so we wont get an error when we are (only) removing all operators
* Move GetCurrentVehicles from libs/api/vehicles to libs/api/transport-authority

## Why

* To be able to validation vehicle/application before user pays for application / better UI experience
* To be able to remove all operators of a vehicle, instead of getting an API error
* To begin with, GetCurrentVehicles was simple, and only used the VehicleSearch API and VehicleServiceFJS API (debt status), but now has changed so it is connected to VehicleOwnerchange API and VehicleOperator API. To make it cleaner (and not flood libs/api/vehicles with these custom endpoints), I have moved that logic into libs/api/transport-authority.

## Checklist:


disaerna (approved):
Nice refactor 🎉

sigruntg (approved):

PR #9713, opened 2023-01-23
fix(j-s): Wait for update to complete when updating CourtDate

oddsson:
# Wait for update to complete when updating CourtDate

[Asana](https://app.asana.com/0/1199153462262248/1203702896361389/f)

## What

Wait for update to complete when updating CourtDate

## Why

If we don't wait for the update, we could be sending a notification while CourtDate is not set. This should never happen.

## Checklist:


apps/judicial-system/web/src/routes/Court/Indictments/Subpoena/Subpoena.tsx
```diff
@@ -60,11 +60,13 @@ const Subpoena: React.FC = () => {
 
   const handleNavigationTo = useCallback(
     async (destination: keyof stepValidationsType) => {
-      const hasSentNotification = workingCase?.notifications?.find(
-        (notification) => notification.type === NotificationType.COURT_DATE,
+      const hasSentNotification = workingCase.notifications?.find(
+        (notification) =>
+          notification.type === NotificationType.COURT_DATE &&
+          notification.recipients.every((recipient) => recipient.success),
```
gudjong, line 66: Checking for successful notifications is a bit more subtle than this. Over time, court date notifications may have been sent multiple times for the same case, some may have succeeded, some failed and some succeeded partially.

It would probably be more accurate to check if the latest notification was partially successful. To find the latest notification you can use the field `created` in the `notifications` objects. To check for partial succes, you can exchange `every` for `some` in the check.

The reason for checking for partial success is the way we handle notifications in the backend. Any given notification may be sent to multiple recipients. We keep retrying the notification until we succeed in sending the notification out to a single recipient. Later, we may start retrying individual recipients and label the notifications with something more useful than an email address or phone number. For instance, we could store that the prosecutor was successfully sent a court date notification, but a defender was not. Until then, partial success is the best we can do.

apps/judicial-system/web/src/routes/Court/InvestigationCase/HearingArrangements/HearingArrangements.tsx
```diff
@@ -98,11 +98,13 @@ const HearingArrangements = () => {
 
   const handleNavigationTo = useCallback(
     async (destination: keyof stepValidationsType) => {
-      const hasSentNotification = workingCase?.notifications?.find(
-        (notification) => notification.type === NotificationType.COURT_DATE,
+      const hasSentNotification = workingCase.notifications?.find(
+        (notification) =>
+          notification.type === NotificationType.COURT_DATE &&
+          notification.recipients.every((recipient) => recipient.success),
```
gudjong, line 104: See above.

gudjong (changes requested):

gudjong (approved):

PR #9714, opened 2023-01-23
fix(inheritance-report): adding data schema

albinagu:
## Checklist:


Thorrik58 (approved):

libs/application/templates/inheritance-report/src/lib/dataSchema.ts
```diff
@@ -2,4 +2,161 @@ import * as z from 'zod'
...
 export const inheritanceReportSchema = z.object({
   approveExternalData: z.boolean().refine((v) => v),
+
+  /* assets */
+  realEstate: z.object({
+    data: z
+      .object({
+        assetNumber: z.string(),
+        description: z.string(),
+        propertyValuation: z.string().refine((v) => v),
```
kksteini, line 12: Just a thought.

Is this, and other instances of `z.string().refine((v) => v)` a `non-empty` assertion?
I don't know what is a good practice here but maybe `z.string().min(1)` is more clear if that's the case?
Does that maybe create weird error messages in the validation on the frontend?

kksteini (approved):

libs/application/templates/inheritance-report/src/lib/dataSchema.ts
albinagu: thought of that as well.. i dont really know which one is better

albinagu: ill do the research :)

PR #9715, opened 2023-01-23
feat(j-s): Queue notifications that are sent when a case is revoked

unakb:
# Queue notifications that are sent when a case is revoked

[Setja REVOKED tilkynningar á biðröð](https://app.asana.com/0/1199153462262248/1203640056170094/f)

## What

Move "revoked" notification to backend queue so that the web is not responsible for triggering notifications 

## Why

So they can be retried if necessary, and it's more reliable

## Screenshots / Gifs

## Checklist:


apps/judicial-system/web/src/routes/Shared/Cases/Cases.tsx
```diff
@@ -204,7 +204,6 @@ export const Cases: React.FC = () => {
       caseToDelete.state === CaseState.SUBMITTED ||
       caseToDelete.state === CaseState.RECEIVED
     ) {
-      await sendNotification(caseToDelete.id, NotificationType.REVOKED)
```
gudjong, line 207: Consider removing `sendNotification` and `NotificationType` from the imports above as they are no longer used.

gudjong (changes requested):

gudjong (approved):

PR #9719, opened 2023-01-24
fix(financial-statement-inao): fsi don't overwriet fetched data on return

HelgiHH:
## Why

data previously field and possibley changes was being overwritten when re entering the application 

## Checklist:


fjandakornid (approved):
LGTM

PR #9727, opened 2023-01-24
fix(operating-license): Criminal record validate and submit

alexdiljar:
No endpoint to check if user has clean criminal record, so we check if its valid and then  submit it to sýslumenn with the application.

## Checklist:


libs/application/template-api-modules/src/lib/modules/templates/operating-license/operatingLicense.service.ts
```diff
@@ -256,23 +251,48 @@ export class OperatingLicenseService extends BaseTemplateApiService {
...
+
+    return sealedDocument
   }
 
   private async getAttachments(
     application: ApplicationWithAttachments,
   ): Promise<Attachment[]> {
     const attachments: Attachment[] = []
 
+    const dateStr = new Date(Date.now()).toISOString().substring(0, 10)
```
kristofer-lp8f, line 287: Why not use `date-fmt`?

libs/application/template-api-modules/src/lib/modules/templates/operating-license/operatingLicense.service.ts
```diff
@@ -76,16 +76,16 @@ export class OperatingLicenseService extends BaseTemplateApiService {
...
       }
 
-      return { success: true }
+      throw new TemplateApiError(
+        {
+          title: coreErrorMessages.dataCollectionCriminalRecordTitle,
+          summary: coreErrorMessages.errorDataProvider,
+        },
+        400,
+      )
```
kristofer-lp8f, line 88: The error is swallowed anyway in the duplicate code below
```suggestion
      throw new Error()
```

kristofer-lp8f (approved):
LGTM

PR #9750, opened 2023-01-26
test(j-s): Implement first playwright test

oddsson:
# Implement first playwright test

[Asana](https://app.asana.com/0/1199153462262248/1203659220898784/f)

## What

This is the first test for the Judicial-system written in Playwright. This test does not do much, but it serves as something we can use as we are setting up support for e2e tests in the deployment pipe before creating more useful tests.

## Checklist:


apps/system-e2e/src/tests/islandis/judicial-system/smoke/first-test.spec.ts
```diff
@@ -0,0 +1,27 @@
...
+      phoneNumber: '0103019',
+      idsLoginOn: false,
+    })
+  })
+
+  test.afterAll(async () => await context.close())
+
+  test('should work', async () => {
+    const page = await context.newPage()
+    await page.goto('/api/auth/login?nationalId=0000000009')
```
kristofer-lp8f, line 24: This looks strange. Is this how login works in the Judicial System Kröfur?

kristofer-lp8f (changes requested):
Great first test!

Some pointers to make the output clearer and more robust. I would also like to see the filename be something meaningful, like `krofur.spec.ts`.

apps/system-e2e/src/tests/islandis/judicial-system/smoke/first-test.spec.ts
oddsson: Yeah, this is strange, I should have explained.

We've created a way to bypass the login in our system everywhere but in PROD and that's how we do it.

apps/system-e2e/src/support/session.ts
```diff
@@ -147,3 +147,16 @@ export async function session({
...
+  const page = await context.newPage()
+  const authUrlPrefix = urls.authUrl
+  await ensureCognitoSessionIfNeeded(
+    page,
+    JUDICIAL_SYSTEM_HOME_URL,
+    authUrlPrefix,
+  )
+  await page.close()
+  return context
+}
```
krrrrrrrrrrr, line 162: Why is this needed? Can't you use this in your test?
```
session({
  browser,
  homeUrl: '/api/auth/login?nationalId=9999999999',
  phoneNumber: null,
  authUrl: '/',
  idsLoginOn: false,
  delegation: false,
})
```

krrrrrrrrrrr (changes requested):
Test LGTM :+1: 

I don't see why you need a new session function that's different from the one already in place?

apps/system-e2e/src/support/session.ts
oddsson: We use a way to bypass the login that's different than in other projects. What we do is when we go to `localhost/api/auth/login?nationalId=some_national_id` we bypass the login and are redirected to our homepage, i.e. `/krofur`.

`session` checks if `homeurl` is correct and that doesn't happen in our case, because we use this bypass mechanism. 

Screenshot 2023-03-08 at 13 57 16

This was created after a talk with @pshomov, who suggested this.

Screenshot 2023-03-08 at 13 58 11

krrrrrrrrrrr: I don't know whether I agree. I would prefer you used our session mechanism, and instead updated it to allow your workflow. Surely, we will have to test other services in the future with similar requirements, which would benefit from a common `session` mechanism.

This would also reduce maintenance :crossed_fingers:

krrrrrrrrrrr (approved):

kristofer-lp8f (approved):

PR #9759, opened 2023-01-27
fix: amount format browser support

disaerna:
# Service Portal - Finance formatting amount

## What

Replace formatting because of users on old Safari browsers could not view finance pages. 

## Why

Safari 12.1.2 doesnt support intl.PluralRules 🤷 

## Checklist:


thordurhhh (approved):
🙌

PR #9763, opened 2023-01-30
fix(erfdafjarskyrsla): Efs improvements

albinagu:
## Checklist:


libs/application/templates/inheritance-report/src/fields/CalculateTotalAssets/index.tsx
```diff
@@ -0,0 +1,42 @@
...
+  const [total] = useState(
+    (getValueViaPath(answers, 'assets.otherAssets.total') as number) +
+      (getValueViaPath(answers, 'assets.money.total') as number) +
+      (getValueViaPath(answers, 'assets.stocks.total') as number) +
+      (getValueViaPath(answers, 'assets.claims.total') as number) +
+      (getValueViaPath(answers, 'assets.bankAccounts.total') as number) +
+      (getValueViaPath(answers, 'assets.inventory.total') as number) +
+      (getValueViaPath(answers, 'assets.vehicles.total') as number) +
+      (getValueViaPath(answers, 'assets.realEstate.total') as number),
+  )
```
kksteini, line 24: If one of these is undefined the result will be `NaN`. Is that okay?
Is all of this required before reaching this state?

libs/application/templates/inheritance-report/src/fields/CalculateTotalBusiness/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { useFormContext } from 'react-hook-form'
+import { m } from '../../lib/messages'
+
+export const CalculateTotalBusiness: FC<FieldBaseProps> = ({ application }) => {
+  const { answers } = application
+  const { formatMessage } = useLocale()
+  const { setValue } = useFormContext()
+
+  const [total] = useState(
+    (getValueViaPath(answers, 'business.businessAssets.total') as number) -
```
kksteini, line 16: getValueViaPath<number> ?

kksteini (approved):
LGTM but maybe use the generic type in getValueViaPath.

That is, use `getValueViaPath<T>(a, path)` instead of `getValueViaPath(a, path) as T`

PR #9765, opened 2023-01-30
fix(applications): Url parameter for delegation

egill15:
# Url parameter fix

## What
Adding correct url parameter to delegation checked application

## Why
In the new react router package when calling
useNavigation
and instantly
useLocation

the useNavigation is not fast enough to update itself so the uselocation gets the old value.

Fixed by sending the value itself into the function from parent.

## Checklist:

### DIFF
--- a/libs/application/ui-shell/src/components/DelegationsScreen.tsx
+++ b/libs/application/ui-shell/src/components/DelegationsScreen.tsx
@@ -20,7 +20,7 @@ import { getApplicationTemplateByTypeId } from '@island.is/application/template-
 import { LoadingShell } from './LoadingShell'
 import { format as formatKennitala } from 'kennitala'
 import { useLocale } from '@island.is/localization'
-import { useNavigate } from 'react-router-dom'
+import { useNavigate, useLocation } from 'react-router-dom'
 import { ScreenType, DelegationsScreenDataType, Delegation } from '../types'
 import { FeatureFlagClient, Features } from '@island.is/feature-flags'
 import { useFeatureFlagClient } from '@island.is/react/feature-flags'
@@ -44,6 +44,7 @@ export const DelegationsScreen = ({
   const { switchUser, userInfo: user } = useAuth()
   const navigate = useNavigate()
   const featureFlagClient: FeatureFlagClient = useFeatureFlagClient()
+  const location = useLocation()
 
   // Check for user delegations if application supports delegations
   const { data: delegations, error } = useQuery(ACTOR_DELEGATIONS, {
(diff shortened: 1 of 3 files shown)


libs/auth/react/src/lib/Authenticator/Authenticator.tsx
```diff
@@ -94,7 +94,10 @@ export const Authenticator: FC<Props> = ({ children, autoLogin = true }) => {
       return userManager.signinRedirect({
         state:
           authSettings.switchUserRedirectUrl ??
-          getReturnUrl(locationRef.current, authSettings),
+          getReturnUrl(
+            location != null ? location : locationRef.current,
```
snaerth, line 98: Instead of passing a new location object to the getReturnUrl we need to make sure the `locationRef.current` is also updated.

So I would instead update the ref instead of passing a conditional to the getReturnUrl. So something like.

```ts
 const switchUser = useCallback(
    async function switchUser(nationalId?: string, newLocation?: Location) {
      if (newLocation) {
        locationRef.current = newLocation
      }

      const args =
        nationalId !== undefined
          ? {
              login_hint: nationalId,
              /**
               * TODO: remove this.
               * It is currently required to switch delegations, but we'd like
               * the IDS to handle login_required and other potential road
               * blocks. Now OidcSignIn is handling login_required.
               */
              prompt: 'none',
            }
          : {
              prompt: 'select_account',
            }

      dispatch({
        type: ActionType.SWITCH_USER,
      })
      return userManager.signinRedirect({
        state:
          authSettings.switchUserRedirectUrl ??
          getReturnUrl(locationRef.current, authSettings),
        ...args,
      })
      // Nothing more happens here since browser will redirect to IDS.
    },
    [userManager, dispatch, authSettings, locationRef],
  )
```

libs/application/ui-shell/src/components/DelegationsScreen.tsx
```diff
@@ -170,10 +171,10 @@ export const DelegationsScreen = ({
   ])
 
   const handleClick = (nationalId?: string) => {
-    if (screenData.screenType !== ScreenType.ONGOING) {
-      navigate('?delegationChecked=true')
-    }
-    if (nationalId) {
+    if (screenData.screenType !== ScreenType.ONGOING && nationalId) {
+      location.search = '?delegationChecked=true'
```
snaerth, line 175: This seems hacky 😨 . But does the trick for now I guess

snaerth (commented):
Nice spot 👍 ! 
It seems weird there is some race condition with the updating on the location.
I had a comment that would be nice to change

snaerth (approved):
Nice job!

snaerseljan (approved):

PR #9772, opened 2023-01-30
fix(parental-leave): Applicant does not give other parent consent but can still request for days

veronikasif:
# ...

## What

[UUFV-239] - If applicant do not want to give his consent that the non-custodial parent is allowed right of access during the parental leave, we should show the transferal of rights page BUT we need to disable the third option (I want to give days to the other parent)

[UUFV-257] - Fixed personal allowance so can't select percentage below 1 (could select 0.1%)

## Why

## Screenshots / Gifs

## Checklist:


karenbjorg (approved):
lgtm 👍

PR #9774, opened 2023-01-30
feat(j-s): Add traffic violation screen

oddsson:
# Add traffic violation screen

[Asana](https://app.asana.com/0/1199153462262248/1203818100039903/f)

## What

Add traffic violation screen. This is the first step towards creating a special handling of traffic violations.

## Checklist:


apps/judicial-system/api/infra/judicial-system-api.ts
```diff
@@ -35,8 +35,8 @@ export const serviceSetup = (services: {
       },
       HIDDEN_FEATURES: {
         dev: '',
-        staging: 'INDICTMENTS',
-        prod: 'INDICTMENTS',
+        staging: 'INDICTMENTS, INDICTMENT_ROUTE',
```
busla, line 38: Looking at the DSL, I cant see how the comma separated string will be parsed. Is it guaranteed that it will work as expected?

busla (commented):
LGTM

busla (approved):

apps/judicial-system/web/src/components/CaseDates/CaseDates.tsx
```diff
@@ -1,11 +1,12 @@
 import React from 'react'
 import { useIntl } from 'react-intl'
 
-import { CaseDecision, CaseType } from '@island.is/judicial-system/types'
+import { CaseDecision } from '@island.is/judicial-system/types'
 import { formatDate } from '@island.is/judicial-system/formatters'
 import { Box, Button, IconMapIcon, Text } from '@island.is/island-ui/core'
 import { TIME_FORMAT } from '@island.is/judicial-system/consts'
 import { caseDates } from '@island.is/judicial-system-web/messages'
+import { CaseType } from '@island.is/judicial-system-web/src/graphql/schema'
```
unakb, line 9: Just wondering, was there a specific reason for switching to a generated CaseType in relation to the traffic violation screen task ? Or could they have been separate PRs?

apps/judicial-system/web/src/components/FeatureProvider/FeatureProvider.tsx
```diff
@@ -6,7 +6,10 @@ interface FeatureProvider {
   features: Feature[]
 }
 
-const availableFeatures: Feature[] = [Feature.INDICTMENTS]
+const availableFeatures: Feature[] = [
```
gudjong, line 9: We may be able to simplify this by using something like `Object.keys(Feature)`. That will make this less error prone.

gudjong (changes requested):
Awesome, some minor comments.

apps/judicial-system/web/src/components/CaseDates/CaseDates.tsx
oddsson: They could very well be separate PRs. My thinking is that we do this refactoring as we are working on other stories, that way we don't need to plan and estimate them, but rather we just refactor as we go along. We can discuss other strategies for the future.

apps/judicial-system/web/src/components/FeatureProvider/FeatureProvider.tsx
oddsson: Excellent point! I do think, however that belongs in a separate PR since this is not straight forward. The type of Object.keys will always be string[], and therefore we'd miss the typing on this. I'd like to dig a little into typescript to do this properly. I've added a task to our Inbox.

gudjong (approved):

brynjarorng (approved):
LGTM (Devops)

PR #9790, opened 2023-02-01
feat(application-system): Better DX when adding payments to an application.

obmagnusson:
## What

- Shared payment form for payment states.
- Create charge Shared template api 
- VerifyPayment Shared template api
- Documentation updated

## Why

Less code duplication for payments. Less implementation time to add payments for applications 

## Checklist:


egill15 (approved):
LGTM

kksteini (approved):
CODEOWNER files LGTM

libs/application/types/src/lib/template-api/shared-api/shared-api-definitions/payment-catalog.ts
```diff
@@ -6,5 +6,8 @@ export interface PaymentCatalogParameters {
 
 export const PaymentCatalogApi = defineTemplateApi<PaymentCatalogParameters>({
   action: 'paymentCatalog',
-  namespace: 'PaymentCatalog',
+  namespace: 'Payment',
+  params: {
+    organizationId: '6509142520',
```
saevarma, line 11: Is this default `organizationId` value or is the application template always required to provide the id when defining their paymentapi? Just wondering if we could accidentally have Syslumadur on some payments that should belong to another organisation?

saevarma (approved):
Approved, just one question about the `PaymentCatalogApi` `organizationId`.

libs/application/types/src/lib/template-api/shared-api/shared-api-definitions/payment-catalog.ts
obmagnusson: yeah good point. Ill remove the this id as the default and just throw an error if it is missing.

PR #9791, opened 2023-02-01
feat(web): Add terms and conditions to footer links

RunarVestmann:
# Add terms and conditions to footer links

## Why

* This was requested by Digital Iceland

## Screenshots / Gifs

image

## Checklist:


libs/island-ui/core/src/lib/Footer/Footer.tsx
```diff
@@ -6,7 +6,8 @@ import { Logo } from '../Logo/Logo'
 import { Tiles } from '../Tiles/Tiles'
 import { Text } from '../Text/Text'
 import { Inline } from '../Inline/Inline'
-import { Icon } from '../Icon/Icon'
+import { Icon as IconDeprecated } from '../Icon/Icon'
```
oddsson, line 9: Why do we need to use the deprecated icon component?

oddsson (changes requested):
One thing about using a deprecated component

libs/island-ui/core/src/lib/Footer/Footer.tsx
RunarVestmann: Because it contains the facebook icon, perhaps we could add that icon to the other Icon component then I can replace it

RunarVestmann: I added the facebook icon from your PR and stopped using the deprecated Icon component in the Footer

oddsson (approved):

PR #9805, opened 2023-02-01
feat(portals): React router object routes

snaerseljan:
# Update React router (Phase 2)

## Intro

We divided the router migration into 3-4 phases.

### Phases
1. Migrate to newest version of React Router, backward compatibility needed.
2. **(current): Update all SPA's routes to support the new router object tree and move the authentication outside the router..**
3. Start to implement few screens in service-portal to utilize new features in React Router.
4. Update all screens.

## What

Update All SPA's to new react-router > 6.4 object tree syntax. This PR includes.

- Update all SPA's to object routes (service-portal, admin-portal and application-system)
- Authentication decoupled from the router. Now router does not start rendering unless user is authenticated.
- Fixed UserOnboarding which was accidentally deleted a while ago.
- Updated tests and did some refactor

## Why

To utilize the full power of React router, mainly the new route loader, then we need to update to the new object tree syntax, which is the recommended way of defining routes by the React router team.

## TODO

- [ ] Verify that there is no regression, since this is a fairly big change.

## Checklist:


thordurhhh (approved):
Nice work!

Good solution with the onboarding modal 🙌

Thorrik58 (approved):

libs/auth/react/src/lib/auth/AuthProvider.tsx
```diff
@@ -0,0 +1,260 @@
...
+    }
+
+    return user
+  }, [userManager, dispatch])
+
+  const switchUser = useCallback(
+    async function switchUser(nationalId?: string, newLocation?: Location) {
+      if (newLocation) {
+        urlRef.current = `${newLocation.pathname}${newLocation.search}`
+      }
```
eirikurn, line 87: This `newLocation` argument was added to deal with an issue with out of date React Router useLocation data.  We might want to clean it up and just use window.location directly.

We might have an issue with basePath here as well, since the Location object comes from react router, probably without a basepath.

libs/auth/react/src/lib/auth/AuthProvider.tsx
```diff
@@ -0,0 +1,260 @@
...
+  autoLogin = true,
+  basePath,
+}: AuthProviderProps) => {
+  const [state, dispatch] = useReducer(reducer, initialState)
+  const [hasError, setHasError] = useState(false)
+  const userManager = getUserManager()
+  const authSettings = getAuthSettings()
+  const monitorUserSession = !authSettings.scope?.includes('offline_access')
+  const urlRef = useRef<string>()
+  urlRef.current = `${window.location.pathname}${window.location.search}`
```
eirikurn, line 54: Since we're not depending on `useLocation` anymore, we can get rid of `urlRef` and always get the latest location from `window.location`.

libs/auth/react/src/lib/auth/AuthProvider.tsx
```diff
@@ -0,0 +1,260 @@
...
+    async function signOut() {
+      dispatch({
+        type: ActionType.LOGGING_OUT,
+      })
+      await userManager.signoutRedirect()
+    },
+    [userManager, dispatch],
+  )
+
+  const checkLogin = useCallback(async () => {
```
eirikurn, line 129: Fyi, using naming functions inside hooks (and any place they're not assigned directly to a variable) can help when debugging, eg the names show up in stack traces.

libs/auth/react/src/lib/auth/AuthProvider.tsx
```diff
@@ -0,0 +1,260 @@
...
+      signOut,
+    }),
+    [state, signIn, signInSilent, switchUser, signOut],
+  )
+
+  const isLoading =
+    !state.userInfo ||
+    // We need to display loading screen if current route is the redirect path.
+    // This is because the redirect path in not part of our React Router routes.
+    isCurrentRoute(authSettings?.redirectPath)
```
eirikurn, line 244: How about redirectPathSilent? Is userInfo ever set on these routes?

The redirectPathSilent stuff happens in an iframe. Should make sure we don't render the whole app in the iframe.

libs/auth/react/src/lib/auth/AuthProvider.tsx
```diff
@@ -0,0 +1,260 @@
...
+
+const getReturnUrl = (returnUrl: string, { redirectPath }: AuthSettings) => {
+  if (redirectPath && returnUrl.startsWith(redirectPath)) {
+    return '/'
+  }
+
+  return returnUrl
+}
+
+export const AuthProvider = ({
```
eirikurn, line 43: I like this name change.

~Can you think of any way to split up this component in an elegant way?

libs/auth/react/src/index.ts
```diff
@@ -1,7 +1,14 @@
...
+
+// Lib
 export * from './lib/userManager'
 export * from './lib/authLink'
 export { getAccessToken } from './lib/getAccessToken'
+
+// Types
 export type { MockUser } from './lib/createMockUser'
+export type { AuthSettings } from './lib/AuthSettings'
+export * from './lib/createMockUser'
```
eirikurn, line 14: Need to update docs in README.md?

libs/portals/core/src/components/PortalRouter.tsx
```diff
@@ -0,0 +1,67 @@
...
+        modules,
+      })
+        .then((data) => setRouterData(data))
+        .catch((err) => setError(err))
+    }
+  }, [userInfo, modules, featureFlagClient])
+
+  if (error) {
+    throw error
+  }
```
eirikurn, line 47: Interesting 👀

libs/portals/core/src/components/PortalRouter.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+  if (!(userInfo && routerData)) {
+    return null
+  }
```
eirikurn, line 51: Should render loading ui here?

eirikurn (changes requested):
Amazing job Snær!

Most of my comments are fyis, nitpicky or nice-to-have. Should mainly focus on the possible regressions.

apps/application-system/form/src/app/App.tsx
```diff
@@ -1,43 +1,25 @@
...
-import { Application } from '../routes/Application'
-import { Applications } from '../routes/Applications'
-import { AssignApplication } from '../routes/AssignApplication'
-import { Layout } from '../components/Layout/Layout'
 import { environment } from '../environments'
-import { FeatureFlagProvider } from '@island.is/react/feature-flags'
-import { UserProfileLocale } from '@island.is/shared/components'
+import { createRoutes } from '../lib/createRoutes'
+
+const BASE_PATH = '/umsoknir'
```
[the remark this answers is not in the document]
snaerseljan: Good point, will update

libs/auth/react/src/lib/auth/AuthProvider.tsx
```diff
@@ -0,0 +1,260 @@
...
+    if (isCurrentRoute(authSettings?.redirectPath)) {
+      try {
+        const user = await userManager.signinRedirectCallback(
+          window.location.href,
+        )
+
+        dispatch({ type: ActionType.SIGNIN_SUCCESS, payload: user })
+
+        const url = typeof user.state === 'string' ? user.state : '/'
+        window.location.replace(url)
```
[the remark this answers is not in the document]
snaerseljan: You meant location instead of history right? Since history does not have a replace method

libs/auth/react/src/lib/auth/AuthProvider.tsx
snaerseljan: Good spot

libs/auth/react/src/lib/auth/AuthProvider.tsx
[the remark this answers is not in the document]
snaerseljan: Discussed offline

libs/auth/react/src/lib/auth/AuthProvider.tsx
snaerseljan: Either split it up into a specific hook or helper functions. But that should be a part of another PR in my mind.

eirikurn (approved):

PR #9808, opened 2023-02-01
feat(delegation-api): Add clients controller

saevarma:
# Adding clients controller

[Endpoint task](https://www.notion.so/Add-clients-clientId-GET-endpoint-in-delegation-api-to-get-client-name-6289a884e9f04d6d83d3bf4aabef3b13)
[Filter task](https://www.notion.so/Add-clientIds-filter-for-clients-in-delegation-api-to-filter-set-of-clients-5fd9c980a97342edbc0ea04dfdade996?pvs=4)

## What

Adding clients controller to be able to query clients for name and domainName.

## Why

To use for sessions gql resolver, to provider data for session history.

## Screenshots / Gifs

## Checklist:


apps/services/auth/delegation-api/src/app/clients/clients.controller.ts
```diff
@@ -0,0 +1,64 @@
...
+
+    if (!clients) {
+      return []
+    }
+
+    return Promise.all(
+      clients.map(async (client) => {
+        // Todo: This is a temporary solution until we have linked clients to domains
+        const clientDomain = client.clientId.split('/')[0]
+        const domain = await this.resourcesService.findDomainByPk(clientDomain)
```
eirikurn, line 54: I'm worried about performance and scaling. We already have a lot of clients. It loads all of them, including all attributes even though only a few are used, and does N+1 domain look up as well.

I'd like to see one or more of the following:

- Add a filter query which supports a list of client ids to load. (For the log  we only need a few clients at a time)
- Only select client attributes which we're going to use.
- Load domains in one query. Can also return an unverified domainName and let the data-loader query the domains.

apps/services/auth/delegation-api/src/app/clients/client.dto.ts
```diff
@@ -0,0 +1,22 @@
...
+    description:
+      'This is the name of the client that is displayed to the user.',
+  })
+  clientName!: string
+
+  @ApiProperty({
+    example: 'Stafrænt Ísland',
+    description: 'This key is used to look up the organisation in Contentful.',
+  })
+  organisationLogoKey?: string
```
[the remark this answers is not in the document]
saevarma: Yes that is a good point. Will return the domainName instead and let the GQL use the data loader to query it.

apps/services/auth/delegation-api/src/app/clients/clients.controller.ts
saevarma: Yes I shared this concern, was going to include the query param filter in second iteration. But that was maybe a bit to much shape up 🤪

I updated this PR to:
* Add the query param to filter by clientIds
* Select only the id and the name attributes
* Return domainName to be queried in the GQL layer.

PR #9821, opened 2023-02-02
feat(sessions): Add get endpoint for sessions.

valurefugl:
https://www.notion.so/Add-session-GET-paginated-endpoint-040fe3686ce343709934d73781424ebc
https://www.notion.so/Add-session-GET-query-param-to-support-nationalId-filtering-df6072880cd24a76bc44f76b1ae78c45
https://www.notion.so/Index-Session-Model-db-table-baa8463fc44d4af893bdff5d0994dd0f

## What

## Why

## Screenshots / Gifs

## Checklist:


apps/services/sessions/src/app/sessions/sessions-result.dto.ts
```diff
@@ -0,0 +1,16 @@
...
+
+import { PageInfo } from '@island.is/nest/pagination'
+
+import { Session } from './session.model'
+
+export class SessionsResultDto {
+  @ApiProperty({ required: true })
+  totalCount!: number
+
+  @ApiProperty({ type: () => [Session], required: true })
```
saevarma, line 11:
```suggestion
  @ApiProperty({ type: [Session] })
```
The type property don't need to be lazy loaded with a function, unless there is a circular dependency.

apps/services/sessions/src/app/sessions/sessions-result.dto.ts
```diff
@@ -0,0 +1,16 @@
...
+  data!: Session[]
+
+  @ApiProperty({ required: true })
```
saevarma, line 14:
```suggestion
  @ApiProperty()
```

apps/services/sessions/src/app/sessions/sessions.controller.spec.ts
```diff
@@ -1,34 +1,46 @@
 import { getQueueToken } from '@nestjs/bull'
+import { AuthDelegationType } from 'delegation'
```
saevarma, line 2: I was surprised that this worked, but saw the reason in the first `paths` definition in `tsconfig.base.json`. But I would rather suggest to use the `@island.is` import path, as this can be confused as some npm package.
```suggestion
import { AuthDelegationType } from '@island.is/shared/types'
```

apps/services/sessions/src/app/sessions/sessions.controller.spec.ts
```diff
@@ -61,16 +281,103 @@ describe('SessionsController', () => {
...
       // Assert
       expect(res.status).toEqual(403)
     })
   })
 
+  describe('with personal delegation', () => {
+    let app: TestApp
+    let server: request.SuperTest<request.Test>
+    const user = createCurrentUser({
+      scope: [ApiScope.internal, SessionsScope.sessionsWrite],
```
saevarma, line 295: Personal delegation should never have the `internal` scope. So this is probably a unnecessary test case and with the `if` statement in the service being unnecessary as well.

apps/services/sessions/src/app/sessions/sessions.controller.spec.ts
```diff
@@ -61,16 +281,103 @@ describe('SessionsController', () => {
...
+    })
+  })
+
+  describe('with company delegation', () => {
+    let app: TestApp
+    let server: request.SuperTest<request.Test>
+    let factory: FixtureFactory
+    let sessions: Session[]
+    const user = createCurrentUser({
+      scope: [ApiScope.internal, SessionsScope.sessionsWrite],
```
saevarma, line 332: Procuring holder would have the `internalProcuring` instead of the `internal`.

apps/services/sessions/src/app/sessions/sessions.controller.ts
```diff
@@ -43,14 +46,18 @@ export class SessionsController {
...
-  @Audit<Session[]>({
-    resources: (sessions) => sessions.map((session) => session.id),
+  @Audit<SessionsResultDto>({
+    resources: (result) => result.data.map((session) => session.id),
   })
-  findAll(@CurrentUser() user: User) {
-    return this.sessionsService.findAll(user)
+  findAll(
+    @CurrentUser() user: User,
+    @Headers('X-Query-OtherUser') otherUser: string,
```
saevarma, line 57: Nice!

apps/services/sessions/src/app/sessions/sessions.service.ts
```diff
@@ -1,8 +1,14 @@
-import { User } from '@island.is/auth-nest-tools'
-import { Injectable } from '@nestjs/common'
+import { ForbiddenException, Injectable } from '@nestjs/common'
 import { InjectModel } from '@nestjs/sequelize'
+import * as kennitala from 'kennitala'
+import { Op, WhereOptions } from 'sequelize'
+
+import { User } from '@island.is/auth-nest-tools'
+import { PageInfo, paginate } from '@island.is/nest/pagination'
```
saevarma, line 7: GH flagged `PageInfo` as unused.

apps/services/sessions/src/app/sessions/sessions.service.ts
```diff
@@ -11,11 +17,65 @@ export class SessionsService {
...
+  async findAll(
+    user: User,
+    otherUser: string,
+    query: SessionsQueryDto,
+  ): Promise<SessionsResultDto> {
+    if (user.actor && !kennitala.isCompany(user.nationalId)) {
+      throw new ForbiddenException(
+        'Personal delegations are not allowed to get session data.',
+      )
+    }
```
saevarma, line 29: This is unnecessary as only authenticated user gets the `internal` scope, and procuring holders get the `internalProcuring` scope.

apps/services/sessions/src/app/sessions/sessions.service.ts
```diff
@@ -11,11 +17,65 @@ export class SessionsService {
...
+            },
+            { subjectNationalId: user.nationalId },
+          ],
+        }
+      }
+    }
+
+    return paginate({
+      Model: this.sessionModel,
+      limit: query.limit || 10,
```
saevarma, line 69: Wondering if we should have some upper limit on the user input, to prevent slow queries that could be harmful if the user tries to be funny and query million rows.

apps/services/sessions/test/session.fixture.ts
```diff
@@ -1,15 +1,25 @@
 import faker from 'faker'
 
+import { createNationalId } from '@island.is/testing/fixtures'
+
 import { CreateSessionDto } from '../src/app/sessions/create-session.dto'
 
-export const createRandomSession = (
-  actorNationalId: string,
+export interface CreateSessionDtoOptions {
+  subjectType?: any
```
saevarma, line 8: We should reuse the type form the `createNationalId` for this property. Helps with DX when adding tests later on.

apps/services/sessions/src/app/sessions/sessions.controller.spec.ts
```diff
@@ -45,11 +57,219 @@ describe('SessionsController', () => {
...
+          .map((session) => ({
+            id: session.id,
+            actorNationalId: session.actorNationalId,
+            subjectNationalId: session.subjectNationalId,
+            clientId: session.clientId,
+            timestamp: session.timestamp.toISOString(),
+            userAgent: session.userAgent,
+            ip: session.ip,
+          }))
+          .slice(0, 10),
```
saevarma, line 90: We should create some utility to do the sorting, filtering and slicing as this is repeated through out the tests.
We could also do some other utility for the expect above.

Also the `map` can be simplified if you only need to change the `timestamp`
```
.map((session) => ({
  ...session,
  timestamp: session.timestamp.toISOString()
}))
```

saevarma (changes requested):
Really nice work, and nice job with thorough testing!
Just few comments on improvements.

apps/services/sessions/src/app/sessions/sessions-query.dto.ts
```diff
@@ -0,0 +1,10 @@
+import { ApiProperty } from '@nestjs/swagger'
+import { IsOptional } from 'class-validator'
+
+import { PaginationDto } from '@island.is/nest/pagination'
+
+export class SessionsQueryDto extends PaginationDto {
+  @IsOptional()
+  @ApiProperty({ required: false })
```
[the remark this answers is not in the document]
valurefugl: Fixed.

apps/services/sessions/src/app/sessions/sessions-result.dto.ts
valurefugl: Fixed.

apps/services/sessions/src/app/sessions/sessions-result.dto.ts
```diff
@@ -0,0 +1,16 @@
+import { ApiProperty } from '@nestjs/swagger'
+
+import { PageInfo } from '@island.is/nest/pagination'
+
+import { Session } from './session.model'
+
+export class SessionsResultDto {
+  @ApiProperty({ required: true })
```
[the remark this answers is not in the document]
valurefugl: Fixed.

apps/services/sessions/src/app/sessions/sessions-result.dto.ts
valurefugl: Fixed.

apps/services/sessions/src/app/sessions/sessions.controller.spec.ts
valurefugl: Fixed.

apps/services/sessions/src/app/sessions/sessions.service.ts
valurefugl: Wonder if any user would have a million associated logins in a year.. Any suggestions what the limit should be?

apps/services/sessions/src/app/sessions/sessions.controller.spec.ts
valurefugl: The map is needed since we are comparing properties of the sequelize model to the response.

Don´t think repetition is necessarily bad in test code.  It can make it easier to understand an individual test at a glance without having to navigate utility functions.

apps/services/sessions/src/app/sessions/sessions.controller.spec.ts
valurefugl: Fixed.

apps/services/sessions/src/app/sessions/sessions.controller.spec.ts
valurefugl: Removed.

apps/services/sessions/src/app/sessions/sessions.service.ts
valurefugl: Removed.

apps/services/sessions/test/session.fixture.ts
valurefugl: Fixed.

apps/services/sessions/src/app/sessions/sessions.service.ts
saevarma: I would just start with 100. I don't think the UI is going to query for more than 100 at a time.

apps/services/sessions/src/app/sessions/sessions.controller.spec.ts
saevarma: Ahh ofc the sequelize props.
Okay leave this as is for now.

saevarma (approved):
💯

apps/services/sessions/src/app/sessions/sessions.service.ts
valurefugl: Fixed.

PR #9826, opened 2023-02-03
feat(sessions): Configure session url for ids.

valurefugl:
# ...

## What

## Why

## Screenshots / Gifs

## Checklist:

### DIFF
--- a/apps/services/auth/ids-api/infra/identity-server.ts
+++ b/apps/services/auth/ids-api/infra/identity-server.ts
@@ -72,9 +72,9 @@ export const serviceSetup = (services: {
         (h) => `http://${h.svc(services.authIdsApi)}`,
       ),
       PersistenceSettings__SessionsBaseAddress: {
-        dev: 'http://web-services_sessions.services_sessions.svc.cluster.local',
+        dev: 'http://web-services-sessions.services-sessions.svc.cluster.local',
         staging:
-          'http://web-services_sessions.services_sessions.svc.cluster.local',
+          'http://web-services-sessions.services-sessions.svc.cluster.local',
         prod: 'https://sessions-api.internal.island.is',
       },
       Application__MinCompletionPortThreads: '10',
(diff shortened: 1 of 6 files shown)


apps/services/auth/ids-api/infra/identity-server.ts
```diff
@@ -71,8 +71,12 @@ export const serviceSetup = (services: {
       PersistenceSettings__BaseAddress: ref(
         (h) => `http://${h.svc(services.authIdsApi)}`,
       ),
-      PersistenceSettings__UserProfileBaseAddress:
-        'http://web-service-portal-api.service-portal.svc.cluster.local',
+      PersistenceSettings__SessionsBaseAddress: {
+        dev: 'http://web-services_sessions.services_sessions.svc.cluster.local',
```
busla, line 75: is there a particular reason for using underscores in the URL?

valurefugl: Not that I'm aware of, do you know anything about that @saevarma ?

saevarma: As I understood the internal ingress urls the pattern for dev and staging is  `http://web-<service name>.<namespace>.svc.cluster.local` and the service name and namespace for the sessions service is set with underscore [here](https://github.com/island-is/island.is/blob/262fd4604f58c371bedbb2a75739242b7cacf469/apps/services/sessions/infra/sessions.ts#L3)

valurefugl: @busla should we change the service name and namespace so that they don´t have underscores?

brynjarorng: Yes please change it to have a dash instead of an underscore for the `namespace`, `serviceName` and `workerName`. I should have spotted this during the original PR, sorry about that @valurefugl

valurefugl: No prob, will fix, thanks.

valurefugl: Fixed.

brynjarorng (approved):
LGTM(Devops)

PR #9840, opened 2023-02-06
chore(applications): Update conclusion screen for all norda applications

egill15:
# New conclusion screen

https://app.clickup.com/t/861m7ft45

## What
Add new conclusion screen to applications

## Why
Standardize the look for all applications

## Screenshots / Gifs
image
image

## Checklist:


libs/application/ui-forms/src/lib/formConclusionSection/formConclusionSection.ts
```diff
@@ -28,15 +44,26 @@ export const formConclusionSection = (type: props) =>
         children: [
           buildAlertMessageField({
             id: 'uiForms.conclusionAlert',
-            title: type.alertTitle,
+            title: props.alertTitle,
             alertType: 'success',
-            message: type.alertMessage,
+            message: props.alertMessage,
+          }),
+          buildS3PdfLinkField({
```
Thorrik58, line 51: So what im thinking after reviewing the code a bit, why not just make this field a link field that renders a button with a button to a link to whatever the dev defines?

It could then be buildLinkButtonField, that would accept whatever link (basically a string) we send and let the user open it on the conclusion form. This would help with attachment that could possibly be hosted on other websites. Like if sjukra.is would host their pdf on their site it could link to that from the id defined in the application or something.

We then just put in the hands of the dev to fetch the s3 file from the external data in their application and we in turn make it a lot more general.

What do you think?

Thorrik58 (changes requested):
I would like to see the s3pdfLinkFormField be changed to a general linkFormField or something similarly named that would be more general as it would accept an url instead of an s3 attachment key.

Thorrik58 (approved):

PR #9842, opened 2023-02-06
feat(application-system): PhoneInput component

Toti91:
# PhoneInput component

[Attach a link to issue if relevant](https://app.clickup.com/t/861m6rrk0)

## What

Add new input component `PhoneInput` which allows users to select a country calling code.

The component handles parsing the defaultValue into a country code and phone number.
For some countries the value is formatted correctly, if no formatting is found for a country then it defaults to no formatting.

F.x. `defaultValue="+455812345"` would automatically parse the calling code +45 and select it in the dropdown.
`defaultValue="5812345"` will fallback to a default calling code "+354"

## Why

So that users can select their country calling code if needed

## Screenshots / Gifs

image

image

https://user-images.githubusercontent.com/16031201/217035462-047636fe-d8ee-4a7b-97cd-953db7df64d2.mp4

### Different formats
image

### Focused
image

### Error
image

### Disabled
image

### Required
image

### allowedCountryCodes = ['IS', 'DK']
image

## Checklist:


libs/shared/form-fields/src/lib/PhoneInputController/PhoneInputController.tsx
```diff
@@ -0,0 +1,206 @@
...
+
+/**
+ * Gets default country code.
+ * This function tries to extract a country calling code from a phone number,
+ * and it assumes that the phone number itself is always 7 characters
+ * Might need to rethink this logic in the future when more country calling codes are allowed.
+ *
+ * Example outputs:
+ * getDefaultCountryCode("+3545812345") // +354
+ * getDefaultCountryCode("+455812345") // +45
```
barabrian, line 86: If I'm not mistaken a phone number from DK (+45) is 8 digits so this would probably fail. same with many other country codes

barabrian (commented):
Looks good, just some minor comments regarding phone number formats and country codes

libs/shared/form-fields/src/lib/PhoneInputController/PhoneInputController.tsx
Toti91: Yeah this will only work for icelandic phone numbers as it is

The problem i'm facing is how to correctly parse the country code without knowing how long the phone number is.
e.g. +45689912346 could be parsed as country code +45 or +456

I guess the best way would just be to hardcode this information in the data

eirikurn (commented):
Nice! Not reviewing but wanted to shoot some DoD questions:

- Should there be a story for the island-ui component?
- Has the component been a11y tested. E.g. for limited mobility and visibility impaired users (keyboard and screen reader).
- Should this component support a variant with no country selector? I can imagine many use cases which only allow Icelandic phone numbers but would like to use the "PhoneNumber" component, eg for masking and validation of Icelandic phone numbers.

libs/shared/form-fields/src/lib/PhoneInputController/PhoneInputController.tsx
```diff
@@ -0,0 +1,206 @@
...
+          readOnly={readOnly}
+          rightAlign={rightAlign}
+          countryCodes={countryCodes}
+          backgroundColor={backgroundColor}
+          data-testid={dataTestId}
+          placeholder={placeholder}
+          label={label}
+          type="tel"
+          value={value?.replace(cc, '')}
+          format="###-####"
```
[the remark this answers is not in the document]
eirikurn: Might include this placeholder in `countryCodes`. Not sure where to find it for all countries, but maybe better to have no formatting for some countries than the wrong one.

We can use `libphonenumber-js` to validate the final phone number.

libs/shared/form-fields/src/lib/PhoneInputController/PhoneInputController.tsx
eirikurn: Should loop through countryCodes to find a matching country code. The phone number system is designed so there won't be conflicts like +45 and +456.

eirikurn (commented):
A few more comments.

Also curious about the split between PhoneInput and PhoneInputController. Should NumberFormat and more of the logic be in the island-ui component, so PhoneInputController is only wrapping it with React Hook Form?

Toti91:
> Nice! Not reviewing but wanted to shoot some DoD questions:
>
> Should there be a story for the island-ui component?

Yes definitely, oversight on my part.

> Has the component been a11y tested. E.g. for limited mobility and visibility impaired users (keyboard and screen reader).

I made the component is keyboard accessible, I'll take a better look at visibility and screen reader accessibility.

> Should this component support a variant with no country selector? I can imagine many use cases which only allow Icelandic phone numbers but would like to use the "PhoneNumber" component, eg for masking and validation of Icelandic phone numbers.

At the moment you could add `allowedCountryCodes={['IS']}` which would still display the dropdown but with only "IS" available. But I don't think it's a bad idea to be able to hide the dropdown, I think it really comes down to the designer, i'll take it up with him!

eirikurn:
@barabrian can you provide some input on the PhoneNumber vs PhoneNumberController split?

Toti91:
Moved the logic from the controller to the core component. Makes more sense to do all the heavy lifting there.

libs/island-ui/core/src/lib/PhoneInput/PhoneInput.tsx
```diff
@@ -0,0 +1,340 @@
...
+                    countryCodeList.find(
+                      (x) => x.dial_code === cc && !!x.format,
+                    )?.format
+                  }
+                  onValueChange={({ value }) => {
+                    // Don't prefix value with country code it it's empty.
+                    value
+                      ? onFormatValueChange?.(cc + value)
+                      : onFormatValueChange?.(value)
+                  }}
```
eirikurn, line 281: It's not clear to me how PhoneInput combines the country code and the phone number into the onChange event handler. Does NumberFormat handle it with this hook?

Does it trim all extra symbols? Let's say I have US +1, (123) 32-12321 or something; Is the value +11233212321?

eirikurn (approved):

barabrian (approved):
LGTM

barabrian:
> 

I've been occupied for the last week and have not had the opportunity to review this until now. The current implementation looks to be consistent with other components vs controller splits

barabrian:
> @barabrian can you provide some input on the PhoneNumber vs PhoneNumberController split?

I've been occupied for the last week and have not had the opportunity to review this until now. The current implementation looks to be consistent with other components vs controller splits

egillben (approved):
LGTM

RunarVestmann (approved):

PR #9860, opened 2023-02-07
chore(applications): Error proof pruned applications.

egill15:
# Error handle pruned applications that are already open

https://app.clickup.com/t/861m6rr1m

## What
Stop user from updating / submitting pruned application

## Why
Pruned applications are not allowed to be updated. 

User is redirected back to the application overview screen when the application is pruned

## Checklist:


obmagnusson (changes requested):
We should rather throw a 404 or a custom problem object that the client knows how to deal with and Display a specialized s error screen for this case. This makes for a lot better user experience and a more natural flow.

PR #9865, opened 2023-02-08
feat(service-portal): add error handling to pkpass

disaerna:
# Service portal - License pkpass error handling

## What

* Add error handling if QR code or Pkpass link returns an error
* Safari users in mobile always had to press two times

## Why

* User's were having trouble if service was returning error

## Checklist:

### DIFF
--- a/libs/service-portal/licenses/src/components/QRCodeModal/PkPass.tsx
+++ b/libs/service-portal/licenses/src/components/QRCodeModal/PkPass.tsx
@@ -1,6 +1,7 @@
 import React, { useEffect, useState } from 'react'
 import * as styles from './QRCodeModal.css'
 import {
+  AlertMessage,
   Box,
   Button,
   LoadingDots,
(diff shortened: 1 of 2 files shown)


libs/service-portal/licenses/src/components/QRCodeModal/DriversLicensePkPass.tsx
```diff
@@ -0,0 +1,106 @@
+import React, { useEffect, useState } from 'react'
+import * as styles from './QRCodeModal.css'
```
thordurhhh, line 2: We should try to remove theses unused vars when we see em :D

libs/service-portal/licenses/src/components/QRCodeModal/PkPass.tsx
```diff
@@ -30,131 +24,87 @@ export const PkPass = ({
...
   const { data: userProfile } = useUserProfile()
   const [displayLoader, setDisplayLoader] = useState<boolean>(false)
-  const [modalOpen, setModalOpen] = useState(false)
   const [isMobile, setIsMobile] = useState(false)
   const [fetched, setFetched] = useState(false)
+  const [linkError, setLinkError] = useState(false)
   const locale = (userProfile?.locale as Locale) ?? 'is'
   const { formatMessage } = useLocale()
   const { width } = useWindowSize()
   const timeFetched = new Date() // Used to compare if license is expired
```
thordurhhh, line 37: Do we need this? It seems like we could go without it. It's not really serving much of a purpose is it?

PR #9892, opened 2023-02-09
fix(search-indexer): Increase performance of search-indexer

RunarVestmann:
# Increase performance of search-indexer

## What

* Increased sync speed of search-indexer by fetching only necessary data from Contentful and doing so in a more concurrent manner.
* Added an endpoint for document deletion which Contentful can call when an entry is either archived or deleted. This is to make sure that ES documents on dev actually get deleted if their Contentful counterpart are removed since the Contentful Sync API does not provide such information.

## Why

* The search-indexer is taking a while to sync data, especially if it needs to fetch nested entries from Contentful

## Checklist:


libs/cms/src/lib/search/contentful.service.ts
```diff
@@ -278,46 +296,65 @@ export class ContentfulService {
...
+
+        const responses = await Promise.all(promises)
+
+        let counter = 0
+        for (const linkedEntries of responses) {
+          for (const linkedEntry of linkedEntries) {
+            counter += 1
+            if (
+              environment.indexableTypes.includes(linkedEntry.sys.id) &&
+              !environment.nestedContentTypes.includes(linkedEntry.sys.id)
```
siggifv, line 339: This does not cover the "BabyArticle" case, which is both indexable and nested.

Maybe something like so:
```
if (environment.indexableTypes.includes(linkedEntry.sys.id)) { 
  indexableEntryMap.set(linkedEntry.sys.id, linkedEntry)
} 
if (environment.nestedContentTypes.includes(linkedEntry.sys.id)) {
  nextLevelOfNestedEntryIds.add(linkedEntry.sys.id)
}
```

libs/cms/src/lib/search/contentful.service.ts
```diff
@@ -278,46 +296,65 @@ export class ContentfulService {
...
+            ) {
+              indexableEntryMap.set(linkedEntry.sys.id, linkedEntry)
+            } else {
+              nextLevelOfNestedEntryIds.add(linkedEntry.sys.id)
+            }
+          }
+        }
+
+        // Next round of the loop will only find linked entries to the non indexable entries
```
siggifv, line 348: Update this comment, Indexable and nested are NOT mutually exclusive (which seems to have been a foundational assumption in the sync indexer, e.g. BabyArticle).

siggifv (changes requested):

libs/cms/src/lib/search/contentful.service.ts
```diff
@@ -19,13 +19,6 @@ import {
 } from '@island.is/content-search-index-manager'
 import { Locale } from 'locale'
 
-interface SyncerResult {
```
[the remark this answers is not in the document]
RunarVestmann: Agreed

libs/cms/src/lib/search/contentful.service.ts
```diff
@@ -52,6 +45,16 @@ export class ContentfulService {
       environment: environment.contentful.environment,
       host: environment.contentful.host,
       removeUnresolved: true,
+      logHandler(level, data) {
+        const logContainsRateLimitWarning =
+          level === 'warning' &&
+          typeof data === 'string' &&
+          data.includes('Rate limit')
+
+        if (logContainsRateLimitWarning) return
```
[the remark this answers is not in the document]
RunarVestmann: Agreed

libs/cms/src/lib/search/contentful.service.ts
RunarVestmann: Agreed

charts/islandis/values.dev.yaml
```diff
@@ -1640,6 +1640,7 @@ search-indexer-service:
       cpu: '100m'
       memory: '512Mi'
   secrets:
+    API_CMS_DELETION_TOKEN: '/k8s/search-indexer/API_CMS_DELETION_TOKEN'
```
busla, line 1643: has this param been added to SSM in all envs?

libs/cms/src/lib/search/contentful.service.ts
```diff
@@ -19,10 +20,22 @@ import {
...
+// Taken from here: https://github.com/contentful/contentful-sdk-core/blob/054328ba2d0df364a5f1ce6d164c5018efb63572/lib/create-http-client.js#L34-L42
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const defaultContentfulClientLogging = (level: ClientLogLevel, data: any) => {
+  if (level === 'error' && data) {
+    const title = [data.name, data.message].filter((a) => a).join(' - ')
+    console.error(`[error] ${title}`)
+    console.error(data)
+    return
+  }
+  console.log(`[${level}] ${data}`)
```
busla, line 32: wondering how this will comply with our logging  standards, maybe use the islandis logging lib?

busla (changes requested):
LGTM, minor comments

charts/islandis/values.dev.yaml
RunarVestmann: Yep

libs/cms/src/lib/search/contentful.service.ts
RunarVestmann: Agreed

siggifv (changes requested):

libs/content-search-indexer/src/environments/environment.ts
```diff
@@ -1,5 +1,6 @@
 export const environment = {
   syncToken: process.env.API_CMS_SYNC_TOKEN ?? '',
-  lockTime: 300,
+  deletionToken: process.env.API_CMS_DELETION_TOKEN ?? '',
+  lockTime: 1800,
```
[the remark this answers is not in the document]
RunarVestmann: Done

libs/cms/src/lib/search/cmsSync.service.ts
```diff
@@ -208,4 +208,24 @@ export class CmsSyncService implements ContentSearchImporter<PostSyncOptions> {
     }
     return true
   }
+
+  async handleDocumentDeletion(
+    elasticIndex: string,
+    document: Pick<Entry<unknown>, 'sys'>,
+  ) {
+    if (document.sys.contentType.sys.id !== 'article') {
```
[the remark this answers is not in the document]
RunarVestmann: Done

siggifv (approved):

brynjarorng (approved):
LGTM(Devops)

busla (approved):
LGTM

PR #9917, opened 2023-02-11
feat(api-sessions): Use data loader to load client name and logo

saevarma:
## What

Finish implementation of `ClientLoader` to load client to resolve `clientName` and resolve domain to get logo for display.

## Why

To use client name and logo in session history UI.

## Screenshots / Gifs

N/A

## Checklist:

### DIFF
--- a/apps/services/sessions/sequelize.config.js
+++ b/apps/services/sessions/sequelize.config.js
@@ -1,12 +1,12 @@
 /* eslint-env node */
 module.exports = {
   development: {
-    username: 'dev_db',
-    password: 'dev_db',
-    database: 'dev_db',
+    username: 'dev_db' ?? process.env.DB_USER_SESSIONS,
+    password: 'dev_db' ?? process.env.DB_PASS_SESSIONS,
+    database: 'dev_db' ?? process.env.DB_NAME_SESSIONS,
     host: 'localhost',
     dialect: 'postgres',
-    port: 5434,
+    port: 5434 ?? process.env.DB_PORT_SESSIONS,
   },
   production: {
     username: process.env.DB_USER,
--- a/apps/services/sessions/sequelize.config.js
+++ b/apps/services/sessions/sequelize.config.js
@@ -1,12 +1,12 @@
 /* eslint-env node */
 module.exports = {
   development: {
-    username: 'dev_db' ?? process.env.DB_USER_SESSIONS,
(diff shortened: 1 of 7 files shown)


apps/services/sessions/sequelize.config.js
```diff
@@ -1,12 +1,12 @@
 /* eslint-env node */
 module.exports = {
   development: {
-    username: 'dev_db',
-    password: 'dev_db',
-    database: 'dev_db',
+    username: 'dev_db' ?? process.env.DB_USER_SESSIONS,
+    password: 'dev_db' ?? process.env.DB_PASS_SESSIONS,
+    database: 'dev_db' ?? process.env.DB_NAME_SESSIONS,
```
GunnlaugurG, line 6: intentional or forgot to remove ?

saevarma: Intentional to help with DX. But this has to be the other way around so I fixed it 😅

libs/api/domains/auth/src/lib/loaders/client.loader.ts
```diff
@@ -1,29 +1,62 @@
...
+      throw new UnauthorizedException()
+    }
+
+    // Only support one language at a time.
+    const lang = inputs[0].lang
+    const authClients = await this.clientService.getClients(user, {
+      lang,
+      clientIds: inputs.map((input) => input.clientId),
+    })
+    const domains = await this.domainsService.getDomains(user, { lang })
```
eirikurn, line 43: ~Could use a field resolver and share the domain dataloader for this. Saves a request in case the query isn't requesting domain fields.

libs/api/domains/auth/src/lib/loaders/client.loader.ts
```diff
@@ -1,29 +1,62 @@
...
+    const clients: Client[] = authClients.map((client) => ({
+      clientId: client.clientId,
+      clientName: client.clientName,
+      domain: domains.find((domain) => domain.name === client.domainName),
     }))
+
+    return inputs.map(
+      (input) =>
+        clients.find((client) => client.clientId === input.clientId) ??
+        new NotFoundException(`Could not find client: ${input.clientId}`),
```
eirikurn, line 54: ~It's pretty likely that some clients will be deleted. Should make sure the UI doesn't show an error. Maybe display "Unknown" with the clientId on hover?

eirikurn (approved):
LGTM

PR #9919, opened 2023-02-12
feat(services-sessions): Add geoip parsing

saevarma:
[Notion task for geoip lookup](https://www.notion.so/Use-geoip-lite-for-location-lookup-of-ip-addresses-0300c6cafbeb456baecbc3ad14023b62?pvs=4)
[Notion taks for geoip update](https://www.notion.so/Update-geoip-lite-as-part-of-the-build-step-of-the-activities-service-15e86ac891154a93b1b25477fcf823b0?pvs=4)

## What

Adding geoip-lite lookup in sessions service for worker.
Adding `preBuildCommands` support in our custom build executor to run pre build step which updates the geoip data if running in CI.

@island-is/devops before merging we need to get island.is [sign up](https://www.maxmind.com/en/geolite2/signup?utm_source=kb&utm_medium=kb-link&utm_campaign=kb-create-account) on MaxMind webstie and [generate license key](https://support.maxmind.com/hc/en-us/articles/4407111582235-Generate-a-License-Key) and add as `GEOIP_LICENSE_KEY` environment variable for Github Actions.

## Why

To display location based on ip and update the geoip data.

## Screenshots / Gifs

## Checklist:


apps/services/sessions/src/app/sessions/sessions.service.ts
```diff
@@ -71,9 +72,15 @@ export class SessionsService {
     const device =
       browser || os ? `${browser}${browser && os ? ` (${os})` : os}` : undefined
 
+    const geoLocation = geoip.lookup(session.ip)
+    const ipLocation = geoLocation
+      ? `${geoLocation.city}, ${geoLocation.country}`
+      : undefined
```
eirikurn, line 78: Does this need to be more defensive or are both properties always returned?

tools/executors/node/schema.json
```diff
@@ -161,6 +161,14 @@
       "type": "string",
       "description": "Name of the main output file. (defaults to `main.js`)",
       "default": "main.js"
+    },
+    "preBuildCommands": {
```
eirikurn, line 165: Think we should use run-commands instead 👍

apps/services/sessions/src/app/sessions/sessions.service.ts
saevarma: Looking at the source the country could be returned with empty city. So I will make this more defensive on the string formatting.

tools/executors/node/schema.json
saevarma: Agreed

apps/services/sessions/project.json
```diff
@@ -4,6 +4,22 @@
   "projectType": "application",
   "targets": {
     "build": {
+      "executor": "@nrwl/workspace:run-commands",
+      "options": {
+        "outputPath": "dist/apps/services/sessions",
+        "commands": ["nx esbuild services-sessions"],
+        "parallel": false
```
eirikurn, line 11: These should be fine to run in parallel ... right?

eirikurn (approved):
LGTM 👍

apps/services/sessions/project.json
saevarma: esbuild copies the package from node_modules, so we need to be sure that the data is updated before the build.

PR #9943, opened 2023-02-13
fix(sessions): Add device field. Move domain to field resolver. Use Intl for date format.

saevarma:
## What

* Add device field in gql model and query
* Move domain on client to use a field resolver
* Display Óþekkt for client name if not found, with client Id on hover.
* Use Intl date and time format functions.

## Why

To improve session UI.

## Screenshots / Gifs

## Checklist:


libs/api/domains/sessions/src/lib/sessions.resolver.ts
```diff
@@ -17,6 +17,12 @@ import { Session } from './models/session.model'
 import { SessionsService } from './services/sessions.service'
 import { PaginatedSessionResponse } from './dto/paginated-session.response'
 import { SessionsInput } from './dto/sessions.input'
+import { CustomDelegation, Domain } from '../../../auth/src/lib/models'
+import {
+  DomainDataLoader,
+  DomainLoader,
+} from '../../../auth/src/lib/loaders/domain.loader'
+import { ISLAND_DOMAIN } from '../../../auth/src/lib/services/constants'
```
eirikurn, line 25: Should use absolute imports

eirikurn (approved):
Nice!

libs/api/domains/sessions/src/lib/sessions.resolver.ts
saevarma: This shouldn´t be there, copy paste import gone by unnoticed!

PR #10091, opened 2023-02-15
fix(estate): Einkaskipti tweaks

albinagu:
## Checklist:


libs/application/templates/estate/src/lib/dataSchema.ts
```diff
@@ -129,4 +129,6 @@ export const estateSchema = z.object({
 
   // is: Hefur umsækjandi forræði á búi?
   applicantHasLegalCustodyOverEstate: z.enum([YES, NO]),
+
+  readTerms: z.array(z.enum([YES, NO])).nonempty(),
```
kksteini, line 133: Just double checking. So this is
```
intented and valid: [YES]
valid:              [YES, YES, NO]
valid:              [NO, YES]
invalid:            []
```

This is surely nitpicking if that's the case but we can use `.length()` on arrays.
F.X:
`z.array(z.enum([YES,NO])).length(1)`

which would change it to
```
valid:   [YES]
valid:   [NO]
invalid: []
invalid: [YES, NO]

kksteini (approved):

PR #10105, opened 2023-02-16
feat(transport authority applications): Use shared MessageWithLinkButton component

johannaagma:
## What

* Use shared MessageWithLinkButtonFormField
* Move OpenServicePortal messages into core messages

## Why

To not have 3 duplicates of MessageWithLinkButtonFormField in transport authority applications. It is not possible to use the standard formConclusionSection in these 3 applications, because we are "faking" the status page (show conclusion screen right after payment, but on refresh there is a status screen)

## Checklist:


sigruntg (approved):

Toti91 (approved):
Looks good

PR #10113, opened 2023-02-17
feat(session): Adde date filter to sessions

GunnlaugurG:

-Filter to retrieve data between two dates

-so users can look up sessions for a specific date range

## Screenshots / Gifs

## Checklist:


libs/service-portal/core/src/lib/messages.ts
```diff
@@ -481,6 +481,10 @@ export const m = defineMessages({
     id: 'service.portal:search-placeholder',
     defaultMessage: 'Sláðu inn leitarorð',
   },
+  searchByNationalId: {
```
thordurhhh, line 484: Perhaps this should be placed in:
`libs/service-portal/sessions/src/lib/messages.ts` 
with the `portals-my-pages.session-history:` namespace. 

🙌

thordurhhh (approved):

libs/service-portal/core/src/lib/messages.ts
GunnlaugurG: You are right! :D

libs/service-portal/sessions/src/screens/Sessions/Sessions.tsx
```diff
@@ -21,32 +19,50 @@ import LogTable from '../../components/LogTable/LogTable'
...
+  const { lg } = useBreakpoint()
 
   const [sessionsData, setSessionsData] = useState<SessionsSession[]>([])
   const [searchNationalId, setSearchNationalId] = useState('')
   const [nextCursor, setNextCursor] = useState<string>('')
   const [customLoading, setCustomLoading] = React.useState(true)
-
-  const getQueryVariables = (next: string, nationalId: string) => {
+  const [fromDate, setFromDate] = useState<Date>(initialDates.fromDate)
+  const [toDate, setToDate] = useState<Date>(initialDates.toDate)
```
saevarma, line 40: ~You could have this as an object then instead of 

```
setFromDate(initialDates.fromDate)
setToDate(initialDates.toDate)
```
you would do

```
// Setting the to date
setFilterDates((prev) => ({...prev, to: toDate }))
```

libs/service-portal/sessions/src/components/SessionFilter/SessionFilter.tsx
```diff
@@ -0,0 +1,139 @@
...
+import * as styles from './SessionFilter.css'
+import cn from 'classnames'
+import React, { useState } from 'react'
+import { useLocale } from '@island.is/localization'
+import addYears from 'date-fns/addYears'
+
+const defaultCalState = { top: false, lower: false }
+
+interface IProps {
+  onFilterChange: (value: string) => void
```
saevarma, line 20: Maybe rename this to be explicit about handling the `nationalId` change

libs/service-portal/sessions/src/components/SessionFilter/SessionFilter.tsx
```diff
@@ -0,0 +1,139 @@
...
+  nationalId: string
+  fromDate?: Date
+  toDate?: Date
+  handleDateChange: (date: Date, type: 'from' | 'to' | 'clear') => void
```
saevarma, line 24: Wondering if we could simplify the handling a bit in the `Sessions.tsx` if we just had a specific clear handler that would reset filter date and nationalId states value and refetch. And this would just be handling date changes.

oddsson (approved):

saevarma (approved):
Nice job!

PR #10114, opened 2023-02-17
fix(application): Change button size to link field and make icon customizable.

egillben:
# Button size & Icon

## What
Change button size and make icon customizable

## Why
UX

## Screenshots / Gifs
image

## Checklist:


libs/island-ui/core/src/index.ts
```diff
@@ -114,6 +114,7 @@ export type {
 } from './lib/DatePicker/types'
 export type { ResponsiveProp } from './utils/responsiveProp'
 export type { GridColumns } from './lib/Grid/GridColumn/GridColumn.css'
+export type { IconMapIcon } from './lib/IconRC/Icon'
```
disaerna, line 117: Instead of exporting the map functionality you can use already exported type **IconProps** in the component. Then you can also control if the icon is outlined or filled. 
  ```

  import { IconProps } from '@island.is/island-ui/core'

  icon?: Pick<IconProps, 'icon' | 'type'>

```

Is this something you can use instead?

egillben: I cant import import { IconProps } from '@island.is/island-ui/core' into .ts file, i would have to change the props so they would be exported from '@island.is/island-ui/core/types'.

Would that be okay?

disaerna: This example is from a .ts file 🤔 see example here: 
/libs/portals/core/src/types/portalCore.ts

Toti91 (approved):

libs/island-ui/core/src/index.ts
egillben: Getting strange error then, wonder why I can't import it,  have restarted my VS code but it wont resolve the type, gonna push to github and check what happens here

egillben: tsconfig does not allow .tsx to be imported in the libary im working with, gonna ask the core team if I can change it

disaerna (approved):

PR #10122, opened 2023-02-17
fix(user-profile): Update how to get locale of current logged in user.

thordurhhh:
## What

* Implemented a change to retrieve the locale of the presently authenticated user, rather than the delegated user.
* Remove `getUserProfileStatus`

I am very much open to any suggestions if someone has a different view of a way to implement these changes. 💟 

## Why

* Currently, the saved locale of the user is retrieved regardless of whether the user was acting on their own behalf or on behalf of another individual or company. This approach did not align with the user's expectations, as the language preference should be specific to the current user's context, rather than to a different person or company.
* `getUserProfileStatus` is not used anywhere, it was used for a very brief period.

## Checklist:


disaerna (approved):
LGTM

apps/services/user-profile/src/app/user-profile/userProfile.controller.ts
```diff
@@ -76,9 +84,16 @@ export class UserProfileController {
     nationalId: string,
     @CurrentUser()
     user: User,
+    @Query('actorRequest') actorRequest?: boolean,
   ): Promise<UserProfile> {
-    if (nationalId != user.nationalId) {
-      throw new ForbiddenException()
+    if (actorRequest) {
```
saevarma, line 89: Another option would to make `locales` as sub-resource of `users`.

We have used `/me` and `/actor` as short hand for `/users/.nationalId`, where the `/me` uses `@CurrentUser()` while `/actor` uses `@CurrentActor()` (fallbacks to current user if no actor in token) decorators to get the corresponding nationalId. 

So we could create `GET /actor/locales` (ideally would return array of single locale as it is a collection endpoint) and then we could create `POST /actor/locales` to update the preferred locale.

I would be interested to hear what @eirikurn thinks about this.

saevarma (commented):
When the user changes the locale in the UI it is updated in the user profile. I would assume that needs also a refactor.

thordurhhh:
@saevarma - Did you get a chance to look at this after Máni's changes?
I was testing this and it works well, just wondering how you guys like the setup as it is now?

apps/services/user-profile/src/app/user-profile/userProfile.controller.ts
```diff
@@ -76,9 +86,16 @@ export class UserProfileController {
     nationalId: string,
     @CurrentUser()
     user: User,
+    @Query('actorRequest') actorRequest?: boolean,
```
saevarma, line 89: Are you still going to use actor request to this endpoint?

saevarma:
> @saevarma - Did you get a chance to look at this after Máni's changes? I was testing this and it works well, just wondering how you guys like the setup as it is now?

@thordurhhh I looked over it again and had two comments.

apps/services/user-profile/src/app/user-profile/userProfile.controller.ts
thordurhhh: No, good point. Will fix.

thordurhhh:
> > @saevarma - Did you get a chance to look at this after Máni's changes? I was testing this and it works well, just wondering how you guys like the setup as it is now?
> 
> @thordurhhh I looked over it again and had two comments.

Yes, good comments. Should be fixed now. 🙌

apps/services/user-profile/src/app/user-profile/userProfile.controller.ts
```diff
@@ -67,6 +70,12 @@ export class UserProfileController {
     description: 'The nationalId of the application to update.',
     allowEmptyValue: false,
   })
+  @ApiQuery({
+    name: 'actorRequest',
+    required: false,
+    type: 'boolean',
+    description: 'Is the user requesting the actor information?',
+  })
```
saevarma, line 78: Forgot to remove the swagger decorator? Don't forget to remove the `ApiQuery` import when you remove this.

saevarma (changes requested):

apps/services/user-profile/src/app/user-profile/userProfile.controller.ts
thordurhhh: @saevarma - This should be updated now. I was hurrying a little too much there.

thordurhhh:
@saevarma - If you have a chance to look at this, it should be good now. 🙌

saevarma (approved):
💯 Great Job!

PR #10143, opened 2023-02-21
fix(infra): Fix api infra for render-local-env

krrrrrrrrrrr:
## Checklist:


busla (approved):
LGTM

wondering if you could use try/catch?

```js
export function getMockName(to: string) {
  // URL constructor requires a protocol
  try {
    const url = new URL(to)
    url.protocol = 'proto://'
    url.pathname = ''
    ...
  } catch (e) {
    throw new Error(`getting mock name failed, to is not a valid url: ${e.message}`)
  }
}
```

krrrrrrrrrrr:
> wondering if you could use try/catch?

That wouldn't help. This change is specifically to prevent failure/crash for non-URLs. In this case, we don't actually need URLs, just the host-name, which is very easily extracted using the `URL` constructor.

PR #10159, opened 2023-02-22
feat(auth-admin-domain): New AuthAdmin GQL domain with initial schema

saevarma:
# New AuthAdmin domain

[Notion task](https://www.notion.so/Create-mocked-GraphQL-schema-6f109d58c5024905b5f27ee581e382da)

## What

Creating new AuthAdmin GQL domain to contain schema and resolvers for the new IDS admin module.
Creates initial schema for tenants and returns mocked data while we implement the REST api.

## Why

Serve the new IDS admin module.

## Screenshots / Gifs

## Checklist:


libs/api/domains/auth-admin/src/lib/auth-admin.module.ts
```diff
@@ -0,0 +1,10 @@
+import { Module } from '@nestjs/common'
+import { TenantResolver } from './resolvers/tenant.resolver'
+import { TenantsService } from './services/tenants.service'
```
eirikurn, line 3: I prefer grouping files by domain (tenant) rather than type (resolvers/services), but guess the monorepo is all over the place with this 🤷‍♂️🙃

libs/api/domains/auth-admin/src/lib/models/tenant-environment.model.ts
```diff
@@ -0,0 +1,22 @@
+import { Field, Int, ObjectType } from '@nestjs/graphql'
+
+import { Environment } from './environment'
+import { TranslatedValue } from './translated-value.model'
+
+@ObjectType('AuthAdminTenantEnvironment')
+export class TenantEnvironment {
+  @Field()
+  id!: string
```
eirikurn, line 9: Fyi, this probably needs a field resolver to be unique across environments.

libs/api/domains/auth-admin/src/lib/models/application.model.ts
```diff
@@ -0,0 +1,12 @@
...
+import { TranslatedValue } from './translated-value.model'
+
+@ObjectType('AuthAdminApplication')
+export class Application {
+  @Field(() => ID)
+  name!: string
+
+  @Field(() => [TranslatedValue])
+  displayName!: TranslatedValue[]
+}
```
eirikurn, line 12: Is this used? Should displayName be here or under environments?

libs/api/domains/auth-admin/src/lib/models/tenant.model.ts
```diff
@@ -0,0 +1,12 @@
...
+
+import { TenantEnvironment } from './tenant-environment.model'
+
+@ObjectType('AuthAdminTenant')
+export class Tenant {
+  @Field()
+  id!: string
+
+  @Field(() => [TenantEnvironment])
+  environments!: TenantEnvironment[]
```
eirikurn, line 11: ~We could have a "defaultEnvironment" field resolver on all of our models which just resolves the first (or last) environment to simplify list queries. 

Then list queries would only query `defaultEnvironment` properties and not receive duplicate displayName, applicationsCount etc for each environment.

Not sure if the "max(applicationsCount)" is important enough to warrant querying all environments, or if we should move it to the server somehow, ie `tenant.defaultEnvironment.applicationsCount = Math.max(tenant.environments.map(e => e.applicationsCount)`. Or call it something else like `tenant.mergedEnvironment`?

libs/api/domains/auth-admin/src/lib/models/tenant-environment.model.ts
```diff
@@ -0,0 +1,22 @@
...
+
+  @Field(() => Environment)
+  environment!: Environment
+
+  @Field(() => [TranslatedValue])
+  displayName!: TranslatedValue[]
+
+  @Field(() => Int)
+  applicationsCount!: number
```
eirikurn, line 18:
```suggestion
  applicationCount!: number
```

libs/api/domains/auth-admin/src/lib/models/tenant-environment.model.ts
```diff
@@ -0,0 +1,22 @@
...
+
+  @Field(() => Int)
+  apisCount!: number
```
eirikurn, line 21:
```suggestion
  apiCount!: number
```

libs/api/domains/auth-admin/src/lib/models/tenant-environment.model.ts
eirikurn: Feel free to disagree :)

eirikurn (approved):

libs/api/domains/auth-admin/src/lib/models/tenant.model.ts
saevarma: Yes that is a great suggestion.

PR #10182, opened 2023-02-23
chore(j-s): Add timestamp to court uploads

unakb:
# Add timestamp to court uploads

[Timestamp þegar úrskurður og þingbók eru send í Auði](https://app.asana.com/0/1199153462262248/1203883403486330/f)

## What

Added timestamp to court uploads 

## Why

To clarify when files were created and updated 

## Checklist:


apps/judicial-system/backend/src/app/messages/courtUpload.ts
```diff
@@ -12,14 +12,14 @@ export const courtUpload = defineMessages({
     description: 'Notaður sem nafn á skjalaskrá í Auði.',
   },
   courtRecord: {
-    id: 'judicial.system.backend:court_upload.court_record',
-    defaultMessage: 'Þingbók {courtCaseNumber}',
+    id: 'judicial.system.backend:court_upload.court_record_v2',
+    defaultMessage: 'Þingbók {courtCaseNumber}{date}',
```
oddsson, line 16: Why do we not have a space here and skip it in [this line](https://github.com/island-is/island.is/pull/10182/files#diff-267297348ad8b9d18e4497b2f05ceb8641595b17d28c39e63c1617db2e914774R163)?

oddsson (changes requested):
One minor thing

apps/judicial-system/backend/src/app/messages/courtUpload.ts
unakb: Good point! Will change

oddsson (approved):

PR #10189, opened 2023-02-24
fix(parental-leave): edit-or-add-periods-prev-state old one failed

jhc-deloitte:
# ...

## What

## Why

## Screenshots / Gifs

## Checklist:

### DIFF
--- a/libs/application/templates/parental-leave/src/lib/ParentalLeaveTemplate.ts
+++ b/libs/application/templates/parental-leave/src/lib/ParentalLeaveTemplate.ts
@@ -39,8 +39,9 @@ import { parentalLeaveFormMessages, statesMessages } from './messages'
 import {
   allEmployersHaveApproved,
   findActionName,
-  goToState,
   hasEmployer,
+  isPreviousStateApproved,
+  isPreviousStateVinnumalastofnunApproval,
   needsOtherParentApproval,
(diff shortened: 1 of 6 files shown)


libs/application/templates/parental-leave/src/lib/ParentalLeaveTemplate.ts
```diff
@@ -728,13 +742,27 @@ const ParentalLeaveTemplate: ApplicationTemplate<
...
+              cond: (application) =>
+                goToState(application, States.VINNUMALASTOFNUN_APPROVAL),
+              target: States.VINNUMALASTOFNUN_APPROVAL,
+            },
+            {
+              cond: (application) =>
+                goToState(application, States.VINNUMALASTOFNUN_APPROVE_EDITS),
+              target: States.VINNUMALASTOFNUN_APPROVE_EDITS,
+            },
           ],
```
qdd1, line 758: Maybe we need a default target here.
For example if all condition is false then ABORT doesn't go anywhere. Let's 'vinnumalastofnun_approve_edits' be the defaults one

jhc-deloitte: Ok 👍

qdd1 (approved):

PR #10192, opened 2023-02-24
feat(ids-admin): Ids-admin module

GunnlaugurG:
# ...

Adding a front end for the new IDS admin portal. WIP

## What

Creating the front end for the new Admin portal.

## Screenshots / Gifs

## Checklist:


libs/portals/core/src/components/PortalNavigation/PortalNavigation.tsx
```diff
@@ -19,25 +20,33 @@ const findActiveNav = (
...
-        item?.href ? <Link to={item.href}>{link}</Link> : link
-      }
+      renderLink={(link, item) => {
+        Object.keys(params).forEach((key) => {
+          if (item?.href) {
+            item.href = item.href.replace(`:${key}`, params[key] as string)
+          }
+        })
+        return item?.href ? <Link to={item.href}>{link}</Link> : link
+      }}
```
eirikurn, line 49: Dangerous to mutate argument objects like this. Possibly the `item.href` is only replaced once and then has incorrect url when navigating further.

libs/portals/core/src/components/PortalProvider.tsx
```diff
@@ -38,18 +48,15 @@ export const PortalProvider = ({
...
-              .find((path) => matchPath(path, pathname)),
-          )
+        ? modules.find((module) => {
+            return (
+              spreadRoutsChildren(module.routes({ userInfo, client }))
+                // Extract the path from each route
+                .map(({ path }) => path)
+                // Find the route path that matches the current pathname
+                .find((path) => path && matchPath(path, pathname))
+            )
```
eirikurn, line 58: Note that this (like other places using `matchPath` does not support relative route paths.

It's tempting to see if we can store extra information on React Router routes (like the portal module) and replace all of our route matching to use `matchRoutes` instead.

libs/portals/admin/ids-admin/src/module.tsx
```diff
@@ -0,0 +1,71 @@
...
+  },
+  routes() {
+    return [
+      {
+        name: m.idsAdmin,
+        path: IDSAdminPaths.IDSAdmin,
+        element: <IDSAdmin />,
+      },
+      {
+        name: m.applications,
```
eirikurn, line 31: Should wait for more design before we scaffold the subnavigation and all of these subpages. Focus on launching the tenant page and work on some application page components.

eirikurn (approved):

libs/portals/admin/ids-admin/src/module.tsx
GunnlaugurG: You are right, only added the Application because the cards where almost a copy paste of the tenant cards.

libs/portals/admin/ids-admin/src/lib/navigation.ts
```diff
@@ -0,0 +1,61 @@
...
+    {
+      name: m.apis,
+      path: IDSAdminPaths.IDSAdminDomainsAPIS,
+      description: m.idsAdmin,
+      activeIfExact: true,
+    },
+  ],
+}
+
+export const applicationNav: PortalNavigationItem = {
```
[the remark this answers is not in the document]
GunnlaugurG: Thanks for that info, it should be an easy change then when the decision is final.

PR #10214, opened 2023-02-27
fix(web): Organization page button link DOM nodes

RunarVestmann:
# Organization page button links DOM nodes

## What

* Right now the button links on organization pages are an a tag wrapped around a button tag. This is incorrect DOM hierarchy we'd like to change
* The fix is to correctly determine the `as` prop (since the code contained a bug there previously where the as prop was ignored due to incorrectly phrased conditional statements) for the Button component since it can allow us to render the button as an `a` tag

To further explain the `as` prop for the Button

Here is the change I made
image

Before the statement was `as || variant === 'text' ? 'span' : 'button'` but if an as prop would be provided it would not be used 

That's why I changed the statement to this: `as ? (as as string) : variant === 'text' ? 'span' : 'button'`since that's the functionality we want. If an as prop is provided then we use it, otherwise we determine a default one.

Here's an example for values in node
image

### Before
image

### After
image

## Checklist:


Toti91 (approved):
Nice 👍

Toti91:
Also something that might be a part of this PR is adding typing for optional `href` prop to `<Button>`

image

RunarVestmann:
@Toti91 adding an optional href prop might be a good idea but then we would lose the logic that the `Link` component has unless we simply duplicate it and make the `Button` component also have the same logic.

The `Link`component opens a link in a new tab if it's external, otherwise it opens the link in the same tab. Perhaps we would also add an optional target prop to the `Button` component or somehow make the `Button` component open in a new tab or not depending on whether the link is external.

siggifv (approved):

Toti91:
> @Toti91 adding an optional href prop might be a good idea but then we would lose the logic that the `Link` component has unless we simply duplicate it and make the `Button` component also have the same logic.
> 
> The `Link`component opens a link in a new tab if it's external, otherwise it opens the link in the same tab. Perhaps we would also add an optional target prop to the `Button` component or somehow make the `Button` component open in a new tab or not depending on whether the link is external.

Yup we would have to add target prop as well. We could do some typescript magic i.e. if the Button has a `as="a"` then we allow `href` and `target` props. But yeah some refactor on the Link component would be required

PR #10219, opened 2023-02-27
feat(passport-annulment): new template for Passport Annulment

alexdiljar:
still a WIP, going out on dev with FF 

## Checklist:

---
  - https://app.asana.com/0/0/1203956476179430


libs/clients/passports/src/lib/passportsApi.service.ts
```diff
@@ -54,8 +57,8 @@ export class PassportsService {
   }
 
   private resolvePassports(passportData: IdentityDocumentResponse[]) {
-    const passportArray = passportData.map((item) => {
-      const { productionRequestID, ...passport } = item
+    const passportArray = passportData.map((passport) => {
+      // const { productionRequestID, ...passport } = item
```
thordurhhh, line 61: This could be removed now, right?

thordurhhh (approved):
just a minor comment.
passport client lgtm.

libs/clients/passports/src/lib/passportsApi.service.ts
albinagu: yes 👍

libs/application/templates/passport-annulment/src/fields/DoneImage/index.tsx
```diff
@@ -0,0 +1,12 @@
...
+
+export const DoneImage = () => {
+  return (
+    <Box display={'flex'} justifyContent={'center'} marginY={3}>
+      <School />
+    </Box>
+  )
+}
+
+export default DoneImage
```
kksteini, line 12: Strange to use School for passport?

libs/application/templates/passport-annulment/src/fields/PassportSelection/index.tsx
```diff
@@ -0,0 +1,181 @@
...
+              disabled:
+                tag(identityDocumentData.userPassport).variant !== 'mint',
+            },
+          ],
+          onSelect: (e) => {
+            setValue('productionRequestID', e.split(',')[0])
+            setValue('passportNumber', e.split(',')[1])
+            setValue('passportName', e.split(',')[2])
+            setValue(childPassportRadio, '')
+          },
```
kksteini, line 124: Random nitpick. Is something like this cleaner?
```typescript
          onSelect: (e) => {
            let [productionRequestID, passportNumber, passportName] = e.split(',')
            setValue('productionRequestID', productionRequestID)
            setValue('passportNumber', passportNumber)
            setValue('passportName', passportName)
            setValue(childPassportRadio, '')
          },
```

kksteini: Especially if we add on to this functionality, it's nice to have some semantics around the positions of the split?

libs/application/templates/passport-annulment/src/fields/PassportSelection/index.tsx
```diff
@@ -0,0 +1,181 @@
...
+                tag: child.passports ? tag(child.passports?.[0]) : undefined,
+                disabled,
+              }
+            },
+          ),
+          onSelect: (e) => {
+            setValue('productionRequestID', e.split(',')[0])
+            setValue('passportNumber', e.split(',')[1])
+            setValue('passportName', e.split(',')[2])
+            setValue(userPassportRadio, '')
```
kksteini, line 175: So, the same thought here?

kksteini (approved):

libs/application/templates/passport-annulment/src/fields/DoneImage/index.tsx
albinagu: i need to double check which image we are going to use here, will come back with a replacement

saevarma (approved):
Core files LGTM

PR #10225, opened 2023-02-27
feat(island-ui): Change icon prop for Input component

unakb:
# Changed functionality of icon prop so it accepts an onClick function

## What

Changed the icon prop in the core Input component so it can receive an onClick function. 
Fixed usages of the old icon prop

## Why

We need a clickable icon in an input box for a new feature in the judicial-system app where users can press a 'close' icon in order to remove items from a list.

## Screenshots / Gifs

https://user-images.githubusercontent.com/4820859/221832889-870cff21-ef44-4322-b584-a4ef53dcab34.mov

## Checklist:


oddsson (approved):

disaerna (approved):
👍

Thorrik58 (approved):

saevarma (approved):
Aranja files LGTM.

FYI
Not sure about the UX of the ´x´ icon in your video. As a user I would expect to clear the input field, not remove it. Has this been discussed with designers? Should this maybe be raised in #hönnun_faghópur on Slack?

oddsson (approved):

kksteini (approved):

unakb:
> FYI Not sure about the UX of the ´x´ icon in your video. As a user I would expect to clear the input field, not remove it. Has this been discussed with designers? Should this maybe be raised in #hönnun_faghópur on Slack?

Yeah, arguably it could be misconstrued, good point! However there are already tags on the same page that have an X icon and when pressed are removed, so we think the users of the judicial-system will likely understand the functionality when they are within that context. We'll start with this implementation but are definitely up for discussion if there are any alternatives people have in mind 😊

PR #9591, opened 2023-01-12
feat(license-api): New license-api app 

thorkellmani:
## What

License-Api provides a way for external clients to connect to us.

## Why

If clients are to be able to update their digital licenses, i.e. update inputs, revoke licenses, etc, they need to way to initiate the connection.

## Checklist:

- [x] I have performed a self-review of my own code
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] Formatting passes locally with my changes
- [x] I have rebased against main before asking for a review

PR #10243, opened 2023-02-28
feat(nest): create Core library

thorkellmani:
## What

A new library under /nest that contains reusable NestJs decorators, validators, etc.

## Why
In a recent PR review it was mentioned [(link)](https://github.com/island-is/island.is/pull/9591#discussion_r1112751207) that it would be good to make the NationalId parameter decorator reusable by adding a new nest/core library. However, there already was a library for shared code called`/Validators`. It seemed better to rename `/Validators` to `/Core` and move the validators under `/Core`.

## Checklist:


libs/nest/core/.babelrc
```diff
@@ -0,0 +1,4 @@
+{
+  "presets": ["@nrwl/react/babel"],
```
saevarma, line 2: Looks like you have used some other Nx plugin generator to generate this library than Nest.
Nest library doesn't require babel and in `.eslintrc.json`, `jest.config.js`, `project.json` and `tsconfig.*` don't need to list `.tsx` or `.jsx` files.

But instead of creating a new library we should just rename the `nest/validators` library and add the `validators` folder.
`yarn nx g @nrwl/workspace:move --project nest-validators --destination nest/core`

saevarma (changes requested):

libs/nest/core/.babelrc
thorkellmani: ah, i just used the previous setup but change the paths to reflect but this seems better

PR #10260, opened 2023-03-01
Feat(financial-statements-inao): Restrict elections dropdown

fjandakornid:
# Add restrictions to elections

RES-92

## What

Add restrictions to elections dropdown, so that only open elections and those that the applicant has reached age for are shown.

## Why

Because Jóhannes said so!

## Checklist:


programm-ingovals (approved):
LGTM

Perhaps we could reuse this logic when we need to sell alcohol online to verify age :)

PR #10264, opened 2023-03-01
feat(ids-admin): Tenant loader and back button update

snaerseljan:
# Tenant loader and back button update.

## What

I added a tenant RR loader pattern
I also updated the routing on the back button to navigate back to parent route. 

## Why

Loader pattern is something we are going to use from now on.
Doing it this way we won't have to update state in every single child route, however we have to define it
with RR handle object in `modue.tsx` file instead.

## Checklist:


thordurhhh (approved):

libs/portals/admin/ids-admin/src/components/Applications/Applications.tsx
```diff
@@ -44,7 +43,7 @@ const Applications = () => {
             <GridRow key={item.id}>
               <Link
                 className={styles.fill}
-                to={`/innskraningarkerfi/${tenant}/forrit/${item.id}/`}
+                to={`/innskraningarkerfi/${tenant}/forrit/${item.id}`}
```
saevarma, line 46: ~Wondering if we should use `IDSAdminPaths` and use your new `replaceParams` method to prevent broken urls if we change `IDSAdminPaths`

libs/react-spa/shared/src/lib/replaceParams.ts
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * Replace params in path
+ * @example
+ * replaceParams('/users/:id', { id: 1 }) // '/users/1'
+ */
+export const replaceParams = ({
+  href,
+  params,
+  paramsPrefix = '/',
```
saevarma, line 17: ~The regex is replacing `/:paramName` and if this would be set to something else it would change this character so the url template `/users/:id` would be `/users+1` if the `paramsPrefix  = '+'`. But path params should always be separate path segments.

Is there a use case for this `paramsPrefix` that I'm not seeing or should we just remove it and have it fixed as `'/'`?

libs/react-spa/shared/src/lib/replaceParams.ts
```diff
@@ -0,0 +1,23 @@
...
+  params: Params<string>
+  paramsPrefix?: string
+}
+
+/**
+ * Replace params in path
+ * @example
+ * replaceParams('/users/:id', { id: 1 }) // '/users/1'
+ */
+export const replaceParams = ({
```
saevarma, line 14: This is really handy tool to interpolate our URL template strings.

saevarma (approved):
Nice, good job!

libs/react-spa/shared/src/lib/replaceParams.ts
snaerseljan: Good point!! Will update

PR #10269, opened 2023-03-01
chore(core): Upgrade TypeScript

eirikurn:
## What

Upgrade to latest TypeScript.

This caused issues with services-auth-admin-api because it had some non-strict overrides. This PR updates it to be strict like other projects.

## Why

To support TypeScript features like satisefies and more NPM packages like Contentful Apps.

## Checklist:


apps/services/auth/admin-api/src/environments/environment.ts
```diff
@@ -20,7 +20,8 @@ const prodConfig = {
   },
   auth: {
     audience,
-    issuer: process.env.IDENTITY_SERVER_ISSUER_URL,
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    issuer: process.env.IDENTITY_SERVER_ISSUER_URL!,
```
busla, line 24: out of scope for this PR perhaps, but we could probably typeguard this with a config/env loader

busla (approved):
LGTM, minor comment

apps/services/auth/admin-api/src/environments/environment.ts
eirikurn: Yeah, there's still plenty of legacy configuration in these environment files which I'd love to get rid off. 😅

But out of scope here.

saevarma (approved):
LGTM 👏💯
But I'm wondering if we need to update the response handling in the Admin Web for this 204 No Content responses?
Admin web uses a `BaseService` [here](https://github.com/island-is/island.is/blob/main/apps/auth-admin-web/services/BaseService.ts#L121) to `handleResponse` or `handleError`.  My guess is that it will do nothing and no error banner is shown when it receives 204, so could be a more confusing UX than it already is in the old IDS Admin.

eirikurn:
Not too concerned. The API didn't handle missing data with 404s anyways. Most of the endpoints just returned 200 "null" afaics. You shouldn't see this change of behaviour since it only affects singular gets/puts/deletes, and you only call these for data you see in the UI which already exists.

RunarVestmann (approved):
Woohoo!

gudjong (approved):

thordurhhh (approved):

PR #10272, opened 2023-03-01
fix(infra): add datadog session replay

busla:
- ci: added autoupdate label to kodiaq for auto rebasing source branch
- chore: upgrade browser-rum from 3 -> 4
- fix: added session replay to user-monitoring, fix tsconfig

# ...

## What

Session Replay is a Datadog feature that is part of our bundle but we have not been using it.

From the docs:
> Datadog’s Session Replay allows you to capture and visually replay the web browsing experience of your users.
>
> Combined with RUM performance data, Session Replay is beneficial for error identification, reproduction, and resolution, and provides insights into your web application’s usage patterns and design pitfalls.

https://docs.datadoghq.com/real_user_monitoring/

## Why

## Screenshots / Gifs

## Checklist:


busla:
closes [#10273](https://github.com/island-is/island.is/issues/10273)

eirikurn (changes requested):
Has this been discussed with DI and cleared with legal and our privacy policy?

Afaics this would be a huge change in strategy when it comes to PII. 

Our current strategy is to carefully scrub all production PII from datadog since a lot of developers have access there. Our main website uses a gdpr friendly analytics system to not track any user pii and avoid the need for a cookie banner.

The application system especially displays and inputs a lot of sensitive user info and PII. Sure we can probably configure tools like these to trim all pii, but to me it remains very risky since we're constantly developing these solutions.

IMO this deserves special consideration and handling with proper MÁP workflow.

busla:
The user-monitoring lib sets the PII masking for kennitala already.

The default masking for session replay is to mask all user input but we could go up one more strictness level to mask everything.

How I understand this feature is that it replays the session using log tracing from logs that we are already shipping to DD.

I am trying to make this work now on a feature deployment to see if this is exposing anything more then current logs already do.

eirikurn:
As I understand these features, they send way more than the standard RUM which is mainly telemetry, network metadata, console logs and some dom events. They send the entire dom (compressed) every X ms.

My experience with these tools is that the built in filters are super leaky and you need to manually filter tons of stuff, which is hard to scale. Even though they filter user input, most of our UIs (the DOM) is displaying a lot of PII, eg in user menu, tables, read-only fields, summary screens, confirmation modals, etc.

busla:
I researched this more yesterday and I think we would be inviting alot of unnecessary scrutiny by adapting this feature. I am sure that this would be helpful in some scenarios but would cause way more overhead then the actual benefit.

I consulted this with my team this morning and they brought up some good points,  e.g. 

- Are uploaded PDFs masked?
- What about images?
- Can we be sure that we are not rendering stuff in the DOM that DD cannot mask for some reason?

This was a good run, thanks for raising these concerns @eirikurn 

Closing this PR.

PR #10277, opened 2023-03-02
fix(judicial-administration-client): XROAD paths corrected

alexdiljar:
These are the correct Xroad paths for judicial system.

## Checklist:


busla (commented):
@brynjarorng can you validate these paths?

peturgq (approved):

saevarma (approved):
You should add your team as a codeowner for the `libs/clients/judicial-administration` project. You shouldn't require Core review for this change.

But otherwise core files LGTM ;)

PR #10280, opened 2023-03-02
fix(system-e2e): Disable breaking test

krrrrrrrrrrr:
Tests are all broken, disabling broken test for now.
## Checklist:


apps/system-e2e/src/support/utils.ts
```diff
@@ -1,14 +1,19 @@
-import PDFDocument from 'pdfkit'
+//import PDFDocument from 'pdfkit'
 import fs from 'fs'
 
 export const sleep = (ms: number) => new Promise((r) => setTimeout(r, ms))
 
 export function createMockPdf() {
```
busla, line 6: this could also be done async

```suggestion
export const createMockPdf = async () => {
  await fs.promises.writeFile('./mockPdf.pdf', 'test', (e: any) => {
    throw e
  })

  /*
  const doc = new PDFDocument()
  doc.text('test')
  doc.save()
  doc.pipe(fs.createWriteStream('./mockPdf.pdf'))
  doc.end()
  */
}
```

busla (changes requested):
Minor comment

apps/system-e2e/src/support/utils.ts
krrrrrrrrrrr: This shouldn't be a final version anyway, Deloitte is working on a true fix.

busla (approved):
LGTM

PR #10283, opened 2023-03-02
feat(license-api): scopes n guards

thorkellmani:
## What

Implements a guard that checks that the requestee has a matching scope and license type in the request.

## Why

Every organization that seeks to update their digital licenses needs to be authenticated and authorized. The license type that's being updated is specified in the url, so if there wasn't any authorization, every organization could update every license type, which is obviously bad.

## Checklist:


apps/services/license-api/src/test/setup.ts
```diff
@@ -0,0 +1,20 @@
+import { Type } from '@nestjs/common'
+import { TestApp, testServer, TestServerOptions } from '@island.is/testing/nest'
+
+export let app: TestApp
+
+export const setup = (
+  module: Type<any>,
```
osk, line 7: Can we use `unknown`? If not `eslint-disable`

osk (approved):
LGTM!

apps/services/license-api/src/test/setup.ts
thorkellmani: Tests are on the docket for the future, for now I'll disable the warning

libs/auth-api-lib/seeders/data/client-licenses-rls.ts
```diff
@@ -0,0 +1,21 @@
...
+      '@island.is/licenses:firearm',
+      '@island.is/licenses:verify',
+    ],
+  }),
+  createScope({
+    name: '@island.is/licenses:firearm',
+    displayName: 'RLS digital license update scope',
+    description:
+      'Veitir aðgang að uppfærsluvirkni á skírteinum útgefnum af RLS',
+  }),
```
eirikurn, line 20: Should do this before creating the client which refers to it (move it up).

apps/services/license-api/src/app/modules/license/license.controller.ts
```diff
@@ -108,6 +114,8 @@ export class UserLicensesController {
 }
 
 @Controller({ version: ['1'], path: 'licenses/' })
+@UseGuards(IdsAuthGuard, LicenseTypeScopesGuard)
```
eirikurn, line 117: Should use standard `ScopesGuard`?

apps/services/license-api/src/app/modules/license/license.controller.ts
```diff
@@ -21,12 +22,17 @@ import {
...
+import { LicenseApiScope, getLicenseTypeScopes } from '@island.is/auth/scopes'
+import { LicenseTypeScopesGuard } from './guards/licenseTypeScope.guard'
 
 @ApiHeader({
   name: 'X-Param-NationalId',
   description: "The user's national id",
 })
 @Controller({ version: ['1'], path: 'users/.nationalId/licenses/' })
+@UseGuards(IdsAuthGuard, LicenseTypeScopesGuard)
+@Scopes(...getLicenseTypeScopes())
```
eirikurn, line 35: LicenseTypeScopesGuard is quite specialised so I would suggest getting rid of the `@Scopes()` decorator and get the scope list directly.

apps/services/license-api/src/app/modules/license/guards/licenseTypeScope.guard.ts
```diff
@@ -0,0 +1,40 @@
...
+
+    const licenseType = this.getLicenseTypeFromUrl(request.url)
+    const scopeActions = this.getScopeActions(request.auth?.scope ?? [])
+
+    return scopeActions.includes(licenseType)
+  }
+
+  private hasScope(needScopes: string[], haveScopes: string[] = []): boolean {
+    return needScopes.some((scope) => haveScopes.includes(scope))
+  }
```
eirikurn, line 32: I don't like the string handling in guard logic. Might break in unexpected ways in the future.

I'd suggest changing it to:

1. Use a map of license types and scopes. The guard is can import this directly instead of getting it dynamically using Reflector.
2. Get the license type from the request by doing `request.params.licenseType` (better param name than `licenseId` imo).
3. Return false if the `licenseType` param is empty or if the user does not have the scope for it, eg `request.auth?.scope?.includes(licenseTypeToScope[licenseType])`

eirikurn (changes requested):
Would be good to get a unit test for the guard logic since it is security critical code.

Can look at some of the auth services for how we do API tests for scope handling.

libs/auth-api-lib/seeders/data/client-licenses-rls.ts
thorkellmani: Ah, I just followed the example in the Readme.ts which defines the client first

apps/services/license-api/src/app/modules/license/license.controller.ts
thorkellmani: sounds good

apps/services/license-api/src/app/modules/license/guards/licenseTypeScope.guard.ts
thorkellmani: ah yeah, that seems much better

libs/auth-api-lib/seeders/data/client-licenses-rls.ts
eirikurn: Ah yeah. The readme example works because it first creates the client and when creating the scopes it uses "addToClients" to assign them to the client (which has been created at that point).

apps/services/license-api/src/app/modules/license/license.controller.ts
eirikurn: No I mean this controller should just use the standard ScopesGuard and @Scopes decorator since it doesn't need any dynamic License Type scope magic. This also simplifies the LicenseTypeScopeGuard further.

apps/services/license-api/src/app/modules/license/e2e/licenseTypeScopeGuard.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+
+  it.each`
+    method      | endpoint                                    | scope
+    ${'PUT'}    | ${'/users/.nationalId/licenses/firearm'}    | ${LicenseApiScope.licensesDisability}
+    ${'PUT'}    | ${'/users/.nationalId/licenses/disability'} | ${LicenseApiScope.licensesFirearm}
+    ${'DELETE'} | ${'/users/.nationalId/licenses/firearm'}    | ${LicenseApiScope.licensesDisability}
+    ${'DELETE'} | ${'/users/.nationalId/licenses/disability'} | ${LicenseApiScope.licensesFirearm}
+    ${'POST'}   | ${'/licenses/verify'}                       | ${LicenseApiScope.licensesFirearm}
+  `(
+    '$method $endpoint should return 403 when scope is $scope',
```
eirikurn, line 73: Missing successful (200) tests. E.g. I think "/verify" is broken now.

apps/services/license-api/src/app/modules/license/guards/licenseTypeScope.guard.ts
```diff
@@ -1,40 +1,43 @@
...
-    const scopes = this.reflector.getAllAndOverride<string[]>(SCOPES_KEY, [
-      context.getHandler(),
-      context.getClass(),
-    ])
-
     const request = getRequest(context)
 
-    if (scopes && !this.hasScope(scopes, request.auth?.scope)) {
+    const licenseId = request.params.licenseId
+    const isVerify = licenseId === 'verify'
```
eirikurn, line 17: Afaics this variable is always false.

The 'verify' endpoint does not use a licenseId url param. It also doesn´t need any license type scope logic so it should be just a simple ScopeGuard + Scopes endpoint. That way we can remove all `isVerify ?` logic from this guard.

apps/services/license-api/src/app/modules/license/guards/licenseTypeScope.guard.ts
```diff
@@ -1,40 +1,43 @@
...
-    const scopeActions = this.getScopeActions(request.auth?.scope ?? [])
+    const scopes = isVerify
+      ? [LicenseApiScope.licensesVerify]
+      : getLicenseTypeScopes()
+
+    const authScope = request.auth?.scope
+
+    if (scopes && !this.hasScope(scopes, authScope)) {
+      return false
+    }
```
eirikurn, line 31: This extra check (and `getLicenseTypeScopes()` is strictly not neccessary as the check below should cover everything.

eirikurn (changes requested):
Great! Might be a bug with verify and can clean this even further. 👍

apps/services/license-api/src/app/modules/license/license.controller.ts
thorkellmani: ah ok, i understand

eirikurn (approved):
Great job!

PR #10314, opened 2023-03-06
feat(ids-admin): Connect ApplicationsList to GraphQL layer

GunnlaugurG:
## What

Connect ApplicationsList to the graphql layer

## Why

To get data from the server to the client

## Screenshots / Gifs

## Checklist:


libs/portals/admin/ids-admin/src/components/Applications/Applications.loader.ts
```diff
@@ -0,0 +1,29 @@
+import type { WrappedLoaderFn } from '@island.is/portals/core'
+import {
+  GetApplicationsByIdDocument,
+  GetApplicationsByIdQuery,
+} from './Applications.generated'
+
+export type AuthApplicationsList = GetApplicationsByIdQuery['authAdminApplications']['data']
+export const applicationsLoader: WrappedLoaderFn = ({ client }) => {
+  return async ({ params }): Promise<AuthApplicationsList> => {
+    if (!params) {
```
snaerseljan, line 10: check for `params['tenant']`

libs/portals/admin/ids-admin/src/components/Applications/Applications.loader.ts
```diff
@@ -0,0 +1,29 @@
+import type { WrappedLoaderFn } from '@island.is/portals/core'
+import {
+  GetApplicationsByIdDocument,
+  GetApplicationsByIdQuery,
+} from './Applications.generated'
+
+export type AuthApplicationsList = GetApplicationsByIdQuery['authAdminApplications']['data']
```
snaerseljan, line 7: ~nit: New line above function for readability :)

libs/portals/admin/ids-admin/src/components/Applications/Applications.tsx
```diff
@@ -1,91 +1,233 @@
...
-          <Box>
+  const handleSearch = (value: string) => {
+    setInputSearchValue(value)
+
+    if (value.length > 0) {
+      const filteredList = applications.filter((application) => {
+        return (
+          application?.defaultEnvironment.displayName[0].value
+            .toLowerCase()
+            .includes(value.toLowerCase()) ||
```
snaerseljan, line 45: ~nit: Maybe store in a variable `const val = value.toLowerCase()`

libs/portals/admin/ids-admin/src/components/Applications/Applications.tsx
```diff
@@ -1,91 +1,233 @@
...
+          application.applicationId.toLowerCase().includes(value.toLowerCase())
+        )
+      })
+      setApplications(filteredList)
+    } else {
+      setApplications(originalApplications)
+    }
+  }
+
+  const getHeader = ({ withCreateButton = true }) => {
```
snaerseljan, line 55: you should make it optional, i.e.
`const getHeader = ({ withCreateButton = true } = {}) => {`

or just better to have it as optional argument instead of an object :)

libs/portals/admin/ids-admin/src/components/Applications/Applications.tsx
```diff
@@ -1,91 +1,233 @@
...
+                  name="session-nationalId-input"
+                  value={inputSearchValue}
+                  onChange={handleSearch}
+                  backgroundColor="blue"
+                />
+              }
+              onFilterClear={() => {
+                setInputSearchValue('')
+              }}
+            ></Filter>
```
snaerseljan, line 135: ~you should close like `/>` instead of `></Filter>`

libs/portals/admin/ids-admin/src/components/Applications/Applications.tsx
```diff
@@ -1,91 +1,233 @@
...
           </Box>
+          <Box marginTop={'gutter'}>
+            <Button variant={'text'}>Learn more</Button>
+          </Box>
         </Box>
       </GridRow>
+    </GridContainer>
+  ) : (
+    <GridContainer className={styles.relative}>
+      {getHeader({})}
```
snaerseljan, line 110: Remove redundant {} object param ☝️. See former comment

libs/portals/admin/ids-admin/src/components/Applications/Applications.tsx
```diff
@@ -1,91 +1,233 @@
...
+          <Text variant={'h3'}>No application created</Text>
+          <Text paddingTop={'gutter'}>
+            Lorem ipsum dolor sit amet consectetur. A non ut nulla vitae mauris
+            accumsan at tellus facilisi.
+          </Text>
+          <Box marginTop={6}>
             <Button size={'small'}>Create Application</Button>
           </Box>
+          <Box marginTop={'gutter'}>
+            <Button variant={'text'}>Learn more</Button>
```
snaerseljan, line 103: We should have all values prepared for translation. So it would be good to translate every string in `messages.ts`

libs/portals/admin/ids-admin/src/components/Applications/Applications.tsx
```diff
@@ -1,91 +1,233 @@
...
+      </GridRow>
+    )
+  }
+
+  return applications.length === 0 ? (
+    <GridContainer>
+      {getHeader({ withCreateButton: false })}
+      <GridRow>
+        <Box
+          display={'flex'}
```
snaerseljan, line 86: ~I think it is cleaner to just do `display="flex"` for string values instead of `display={'flex'}` on attributes.
The curly braces in my mind should be used for none string values. It also makes it more easy to spot.

libs/portals/admin/ids-admin/src/components/Applications/Applications.css.ts
```diff
@@ -0,0 +1,20 @@
+import { globalStyle, style } from '@vanilla-extract/css'
+import { theme } from '@island.is/island-ui/theme'
+
+export const relative = style({
+  position: 'relative',
+})
+
+globalStyle(`${relative} a`, {
```
snaerseljan, line 8: Does this have to be global?

libs/api/domains/auth-admin/src/lib/application/application.resolver.ts
```diff
@@ -0,0 +1,43 @@
...
+
+@UseGuards(IdsUserGuard)
+@Resolver(() => Application)
+export class ApplicationResolver {
+  constructor(private readonly applicationsService: ApplicationsService) {}
+
+  @Query(() => ApplicationsPayload, { name: 'authAdminApplications' })
+  getApplications(@Args('input') input: ApplicationsInput) {
+    return this.applicationsService.getApplications({
+      tenantId: input.tenantId,
```
snaerseljan, line 19: this should just be tenantId argument if it is the only thing being used

libs/portals/admin/ids-admin/src/components/Applications/Applications.graphql
```diff
@@ -0,0 +1,19 @@
+# Get Applications by ID
+query GetApplicationsById($input: ApplicationsInput!) {
+  authAdminApplications(input: $input) {
+    data {
+      applicationId
```
snaerseljan, line 5: Will this be reused in `GetApplication` query? Just thinking if the content of data should be a fragment

snaerseljan (commented):
Great work Gulli! Some nits and some things that can be refactored.

libs/portals/admin/ids-admin/src/components/Applications/Applications.tsx
GunnlaugurG: We will add content in there for the filter dropdown, so I will leave it as is :D

libs/portals/admin/ids-admin/src/components/Applications/Applications.css.ts
GunnlaugurG: It is only "global" inside this "relative" class, so it will apply to all "a" tags inside of it.

libs/portals/admin/ids-admin/src/components/Applications/Applications.graphql
GunnlaugurG: It will not be reused, to GetApplication we will need another input model

snaerseljan (approved):
Awesome job man!!!

PR #10323, opened 2023-03-06
chore(j-s): Make workingCase non-optional in PageLayout component

oddsson:
# Make workingCase non-optional in PageLayout component

[Asana](https://app.asana.com/0/1199153462262248/1203855198443291/f)

## What

Make workingCase non-optional in PageLayout component

## Why

workingCase was non-optional only because our user management uses the PageLayout component and in user management, there is no workingCase. In this PR, we are removing the PageLayout component from the user management and making workingCase non-optional. The thinking behind this is that PageLayout is a really important component that wraps all steps in our system if workingCase is omitted, that would result in unwanted behaviour and bugs.

## Checklist:


apps/judicial-system/web/src/routes/Admin/Users/Users.tsx
```diff
@@ -84,7 +84,7 @@ export const Users: React.FC = () => {
   return (
     <div className={styles.userControlContainer}>
       <PageHeader title={formatMessage(titles.admin.users)} />
-      <div className={styles.logoContainer}>
+      <Box display="flex" marginBottom={9}>
```
unakb, line 87: 🙈 Curious George here, is Box > div in most cases? Is this just some clean up for consistency or is there a deeper reason?

unakb (approved):
Just the tiny namespace thing I think would be good to change. 
Everything else is just small stuff I'd like to understand better 😇

PR #10325, opened 2023-03-06
chore: Add new codecov components

eirikurn:
## What

Define new codecov components for microservices, auth services, react libs, nest libs and island-ui.

## Why

To track coverages over time for our main test-based focus.

## Checklist:


pshomov:
I think if you maybe make a dummy change that would trigger a unit test, then all the Codecov components would show up.

codecov.yml
```diff
@@ -18,12 +18,24 @@ component_management:
...
         - apps/api/**
         - libs/api/domains/**
-    - component_id: application-system-api-graphql # this is an identifier that should not be changed
+    - component_id: microservices
+      name: All REST APIs except Application System API.
+      paths:
+        - apps/services/**
+        - apps/air-discount-scheme/backend/**
+        - apps/download-service/**
+        - libs/auth-api-lib/**
```
saevarma, line 32: We talked about when rethinking the folder structure of our repo that if service requires a library it should have `libs/services/<app name>` to we want to list that path as well? 
Today it is only us using it for shared testing utilities for auth services.

saevarma (approved):
LGTM
Removed `automerge` so you can review my comment before you decide to merge this.

codecov.yml
eirikurn: Good point

PR #10326, opened 2023-03-06
feat(nest): Add a new nest/core library

thorkellmani:
## What

* Rename the nest/validators lib to nest/core, with validators as a directory in said lib
* Add the nationalId param decorator to /nest/core/decorators to be shared.

## Why
* The nest/core lib will contain reusable code for use in nest apps. Reusable validators fall under that specification so they'll be folded in.
* The nationalId decorator is neat and should be shared.

## Checklist:


thordurhhh (approved):
Nice 🙌

kksteini (approved):
LGTM

eirikurn (approved):
Nice refactor, though not the biggest fan of the name "core". Who cares 🤷‍♂️😂

PR #10370, opened 2023-03-07
feat(ids-admin): Create Application form

snaerseljan:
# Create Application Form

## What

This PR has logic for Create application form modal in new `ids-admin`

- Modal and Form for create application
- Added client side validation
- Form is using new React router pattern with action and loader.
- Form action validation architecture has been implemented for other to replicate in the future.
- Added GrahpQL endpoints with **mocked** data.
- The modal is actually a route :)

Note: One thing I was not happy with was my solution to translate error messages in zod outside of a React scope.
I ended up with passing strings which match keys in `messages.ts`.
This is the current solution.
```ts
const schema = z.object({
  displayName: z.string().nonempty('errorDisplayName'),
  ....
})
```

If you have a better one please let me know.

And another think the application type translations are not good. I did not want to spend to much time on it.
That needs to be updated to proper Icelandic as well.

## Screenshots / Gifs

https://user-images.githubusercontent.com/112904566/223486020-bdfad795-188c-4cc6-ade8-6f23cbe25a1c.mov

## Checklist:


libs/island-ui/core/src/lib/Input/Input.tsx
```diff
@@ -208,59 +267,14 @@ export const Input = forwardRef(
...
-              {...(required && { 'aria-required': true })}
-            />
+            {prefix ? (
+              <Box display="flex" alignItems="center">
+                <span className={styles.prefix}>{prefix}</span>
+                {input}
+              </Box>
+            ) : (
+              input
+            )}
```
eirikurn, line 277: The parent element already has display flex and alignItems center. Do we need this extra Box?

libs/island-ui/core/src/lib/Input/types.ts
```diff
@@ -61,6 +61,7 @@ export interface InputProps extends InputComponentProps {
   errorMessage?: string
   tooltip?: string
   backgroundColor?: ResponsiveProp<InputBackgroundColor>
+  prefix?: string
```
eirikurn, line 64: I wonder if this feature is generally useful enough to extend our Input component. It's also arguably confusing for screen readers.

An alternative implementation would be to automatically prefill the id based on the name INCLUDING the prefix. If the user changes the id manually and tries to remove the prefix, they get a validation error. It would not match the two colour design, but maybe that's fine for simplicity's sake?

eirikurn: I guess my main concern is the lack of oversight and direction in the design system / island-ui project. I'd want some collaboration and discussion to take place before adding features like this.

Otherwise our components may bloat up if everyone adds any new design idea to it.

snaerseljan: I started out that way. But I thought it would bring to much complexity for a thing that can be solved much simpler. That way you need to make sure that the prefix is always present and cannot be erased in my mind

libs/island-ui/core/src/lib/Input/Input.tsx
snaerseljan: Yes this is necessary the parent element is, ` <Box flexGrow={1}>` but the grandparent is what you where referring to

libs/island-ui/core/src/lib/Input/types.ts
oddsson: I agree with @eirikurn, I think there needs to be a discussion with the designers before implementing. things like this. Otherwise the components in Island-UI will become bloated and hard to maintain. Also, the designers should have a say in what we implement IMO.

You could implement this by using a React-Input-Mask (already included in the monorepo). You wouldn't get the difference in colour though.

https://user-images.githubusercontent.com/3789875/223674175-129b4c9f-260e-476e-817f-0fc5ecd73252.mov 

The code is 

```
<InputMask
        mask="PREFIX 999-9999"
        maskPlaceholder={null}
        alwaysShowMask={true}
        ...
```

oddsson (dismissed):
I think we should be careful about adding functionality to Island-UI components without discussing it with the designers first.

libs/island-ui/core/src/lib/Input/types.ts
snaerseljan: Agreed. Thats a great suggestion. Thanks

libs/portals/admin/ids-admin/src/components/forms/CreateApplication/CreateApplication.action.ts
```diff
@@ -0,0 +1,69 @@
...
+        displayName: data.displayName,
+        applicationId: `${data.tenant}/${data?.applicationId}`,
+        environments: data.environments,
+        applicationType: data?.applicationType,
+      },
+    },
+  })
+
+  if (res.errors) {
+    throw res.errors
```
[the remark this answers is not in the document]
snaerseljan: This will log in the console the error but React router error element will handle it and display the current route segment error or the parent one if not specified.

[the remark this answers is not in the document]
snaerseljan: ~Nice to have would be to have some error box above the form and then we would have to handle this differently

libs/portals/admin/ids-admin/src/screens/Tenant/Tenant.loader.ts
```diff
@@ -0,0 +1,23 @@
...
+      variables: {
+        id: params['tenant'],
+      },
+    })
+
+    if (tenant.error) {
+      throw tenant.error
+    }
+
+    return tenant.data?.authAdminTenant ?? ({} as TenantLoaderResult)
```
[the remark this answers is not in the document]
snaerseljan: Typescript dosen't know if data in tenant.data is present.
I will change it do the following then
```
    if (tenant.error || !tenant.data) {
      throw tenant.error
    }

    return tenant.data.authAdminTenant
```

libs/island-ui/core/src/lib/Input/types.ts
snaerseljan: Solved another way out side component since mask was not sufficient for this scenario

libs/portals/admin/ids-admin/src/components/forms/CreateApplication/CreateApplication.action.ts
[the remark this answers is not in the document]
snaerseljan: Now we have global error box

thordurhhh (approved):
Service portal LGTM 🚀

PR #10374, opened 2023-03-07
chore(ci): auto commit dirty charts

busla:
- chore: auto-commit dirty charts on PRs
- chore: generate chart in PR
- chore: generate charts
- ci: checkout github.event.pull_request.head.sha
- ci: fix typo
- chore: update dirty charts

# Auto-commit changes from code generators

## What

Changes made by code generators will only be auto-commited in PRs, **not on main.** Current checks that cause jobs to intentionally fail will continue to run but only on `main

Reasons
-  No one should push directly on main (currently restricted anyway)
- Prevent polluting commit history (since PRs are squashed)

## Why

Reduce manual tasks for developers that can be done by github bots.

PR #10380, opened 2023-03-08
infra(devx): Shared hooks

krrrrrrrrrrr:
- Add shared hooks for better DX

## Checklist:


.githooks/pre-push
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+set -euo pipefail
+
+: "${GIT_ROOT:="$(git rev-parse --show-toplevel)"}"
+BRANCH="$(git rev-parse --abbrev-ref HEAD)"
+PROTECTED="main"
```
eirikurn, line 6: Also release and pre-release branches?

.githooks/pre-push
```diff
@@ -0,0 +1,15 @@
...
+
+
+if [[ "$BRANCH" == "$PROTECTED" ]]; then
+	echo "HOOK: Pushing to protected branch ($PROTECTED) 🙊"
+	echo "HOOK: Use git push --no-verify to force this operation."
+	exit 1
```
eirikurn, line 12: Guess this only protects us devops/core admins since only we can push directly on protected branches.

.githooks/pre-commit
```diff
@@ -0,0 +1,35 @@
...
+charts() {
+  local ecode
+  if git diff --cached --name-only | grep -q infra; then
+    yarn charts &>/dev/null
+    ecode=$?
+    echo "🤖 Generated charts for you 💖"
+    echo "    So you don't have to 😘"
+    echo "       Just remember to commit them ✅"
+    echo "          (check git status)"
+    echo ""
```
eirikurn, line 27: This is kind of solved already in #10374

eirikurn (approved):
Good idea, though it's difficult to decide these and if they fit everyone's needs, or if there are better solutions for them. 

These are at least small and fast checks, but they don't do much. As soon as they become slow (like running lint, test or format) then many people will rip them out for a manual + CI workflow.

.githooks/pre-push
krrrrrrrrrrr: Indeed, which is also protected in our github config. Belts and suspenders 🙏

PR #10384, opened 2023-03-08
chore(j-s): Refactor feature functionality

oddsson:
# Make availableFeatures of FeatureProvider read directly from Features

[Asana](https://app.asana.com/0/1199153462262248/1203883441002744/f)

## What

Today we have a `Feature` type that defines all features that are feature hidden. We also have a variable, `availableFeatures`, in our `FeatureProvider` that holds the same values but as an array and uses that to provide value to the provider.

When a new feature is added to the Feature type, you used to have to remember to add it in availableFeatures array as well. We often forgot and so this was error prone. AvailableFeatures is now reading directly from the Features type (via Object.values) and therefore you only need to add it once.

## Why

See ☝️ 

## Checklist:


apps/judicial-system/web/src/components/FeatureProvider/FeatureProvider.tsx
```diff
@@ -6,10 +6,7 @@ interface FeatureProvider {
   features: Feature[]
 }
 
-const availableFeatures: Feature[] = [
-  Feature.INDICTMENTS,
-  Feature.INDICTMENT_ROUTE,
-]
+const availableFeatures = Object.values(Feature)
```
unakb, line 9: The only difference here is now the availableFeatures array will always include Feature.NONE, while with the previous code it only had INDICTMENTS and INDICTMENT_ROUTE. It doesn't seem to affect the functionality at all though, but we could also remove it from the availableFeatures list if we want.

gudjong: I think this is fine.

gudjong (approved):

PR #10387, opened 2023-03-08
feat(auth-nest-tools): Add support for multiple issuers

saevarma:
## What

Adding support for multiple token issuers in auth-nest-tools `JwtStrategy`.
Configuring multiple issuers for `services-auth-admin-api`.

## Why

So our `auth-admin-api` can configure multiple issuer to allow for cross environment communication. 

## Screenshots / Gifs

N/A

## Checklist:


libs/auth-nest-tools/src/lib/jwt.strategy.ts
```diff
@@ -6,20 +6,29 @@ import { passportJwtSecret } from 'jwks-rsa'
...
+      secretOrKeyProvider: Array.isArray(config.issuer)
+        ? multiIssuerKeyProvider({
+            ...keyProviderBaseOptions,
+            jwksUri: JWKS_URI,
+            issuers: config.issuer,
+          })
+        : passportJwtSecret({
+            ...keyProviderBaseOptions,
+            jwksUri: `${config.issuer}${JWKS_URI}`,
+          }),
```
eirikurn, line 31: Would be nice to move all of this to a function like `secretOrKeyProvider: createKeyProvider(config.issuer),` which has all of the options and jwks path logic.

libs/auth-nest-tools/src/lib/multi-issuer-key-provider.ts
```diff
@@ -0,0 +1,81 @@
+import { SecretCallback } from 'express-jwt'
+import { decode } from 'jsonwebtoken'
+import { ExpressJwtOptions, JwksClient } from 'jwks-rsa'
+import JwksRsa = require('jwks-rsa')
```
eirikurn, line 4: Are any of these missing from our package.json?

eirikurn (approved):
Nice!

libs/auth-nest-tools/src/lib/create-key-provider.ts
```diff
@@ -1,10 +1,36 @@
...
+import {
+  ExpressJwtOptions,
+  JwksClient,
+  passportJwtSecret,
+  SecretCallback,
+} from 'jwks-rsa'
 
 import { logger } from '@island.is/logging'
 
+const JWKS_URI = '/.well-known/openid-configuration/jwks'
```
eirikurn, line 11:
```suggestion
const JWKS_PATH = '/.well-known/openid-configuration/jwks'
```

eirikurn (approved):
LGTM

busla (changes requested):
minor infra env type comment

charts/identity-server/values.prod.yaml
```diff
@@ -201,6 +201,7 @@ services-auth-admin-api:
     DB_REPLICAS_HOST: 'postgres-ids.internal'
     DB_USER: 'servicesauth'
     IDENTITY_SERVER_ISSUER_URL: 'https://innskra.island.is'
+    IDENTITY_SERVER_ISSUER_URL_LIST: '"https://innskra.island.is"'
```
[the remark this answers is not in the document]
saevarma: Yes. My thinking is that we are adding opt-in multi issuer of JWT tokens for API authentication. But for auth-admin-api in production it will only allow production tokens, so using single issuer. [This new function](https://github.com/island-is/island.is/pull/10387/files#diff-f5f9c1932e53e1e81509e25f5341b2942d089f6a37855b6e98fdda530ce11aa1R17) introduced in this PR is always checking if issuer is provided as an `string` or `string[]`  on app startup. But the configuration for the app is just the same and is not bloated by some array checking. 

I could do a additional smarter thing, if `createKeyProvider` is provided with array of single item it can just use the default method. So then this doesn't matter if it is string or array for single item.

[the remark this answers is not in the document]
busla: ok, I dont mean to be annoying but `IDENTITY_SERVER_ISSUER_URL_LIST` looks confusing when given a string

busla (approved):
LGTM

obmagnusson (approved):
Application system file LGTM

PR #10403, opened 2023-03-09
fix(web): Broken pages on web due to input props change

RunarVestmann:
# Broken pages on web due to input props change

## What

* Input component props were recently changed and are causing 500 errors on the web

See https://github.com/island-is/island.is/pull/10402


disaerna (approved):

brynjarorng (approved):
LGTM

PR #10410, opened 2023-03-09
test(system-e2e): More tests

krrrrrrrrrrr:
## Checklist:


package.json
```diff
@@ -226,6 +226,7 @@
     "pdfkit": "0.11.0",
     "pg": "8.1.0",
     "pg-hstore": "2.3.3",
+    "playwright": "^1.31.2",
```
eirikurn, line 229: Fixed dep?

eirikurn (approved):

package.json
eirikurn: I mean the dependency should not have a "^".

Though do we need this dependency? Or just upgrade @playwright/test? I notice the playwright docs don't install it manually.

scripts/ci/Dockerfile
```diff
@@ -134,6 +134,7 @@ RUN apt-get -y install awscli zip
 
 COPY --chown=pwuser:pwuser ${APP_HOME}/package.json .
 RUN yarn install
+RUN yarn playwright install
```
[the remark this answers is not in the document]
krrrrrrrrrrr: This is `playwright` installing its dependencies. The version of `playwright` running the tests is defined in the base image `mcr.microsoft.com/playwright:v1.31.2-focal` (i.e. version `1.31.2`)

peturgq (approved):

package.json
krrrrrrrrrrr: I set the version with the playwright image we are using. Good catch :)

PR #10414, opened 2023-03-09
feat(island-ui): Support for openOnHover in DropdownMenu

eirikurn:
## What

Extend DropdownMenu to support `openOnHover`.

Went setTimeout route because of 8px gutter between button and menu.

## Why

Needed for login button on island.is.

## Screenshots / Gifs

https://drops.aranja.com/L1uDYvNJ

## Checklist:

### DIFF
--- a/libs/island-ui/core/src/lib/DropdownMenu/useMenuHoverProps.tsx
+++ b/libs/island-ui/core/src/lib/DropdownMenu/useMenuHoverProps.tsx
@@ -1,0 +1,40 @@
+import React from 'react'
+import { MenuStateReturn } from 'reakit/Menu'
+
+interface HoverPropsRefState {
+  hoverCount: number
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  hideTimeout?: any
+}
+
+const MENU_LEAVE_DELAY = 500
+export const useMenuHoverProps = (
+  menu: MenuStateReturn,
+  openOnHover: boolean,
+) => {
+  const hoverRef = React.useRef<HoverPropsRefState>({
+    hoverCount: 0,
+    hideTimeout: undefined,
+  })
+  const hoverState = hoverRef.current
+
+  return openOnHover
+    ? {
+        onMouseEnter: () => {
+          hoverState.hoverCount++
+          if (hoverState.hoverCount === 1) {
+            clearTimeout(hoverState.hideTimeout)
+            menu.show()
+          }
+        },
+        onMouseLeave: () => {
+          hoverState.hoverCount--
+          if (hoverState.hoverCount === 0) {
+            hoverState.hideTimeout = setTimeout(() => {
+              menu.hide()
+            }, MENU_LEAVE_DELAY)
+          }
+        },
+      }
+    : undefined
+}
--- a/apps/web/components/Header/LoginButton.tsx
+++ b/apps/web/components/Header/LoginButton.tsx
@@ -80,6 +80,7 @@ export function LoginButton(props: {
             </Button>
           }
           items={items}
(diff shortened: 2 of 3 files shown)


libs/island-ui/core/src/lib/DropdownMenu/useMenuHoverProps.tsx
```diff
@@ -0,0 +1,40 @@
+import React from 'react'
+import { MenuStateReturn } from 'reakit/Menu'
+
+interface HoverPropsRefState {
+  hoverCount: number
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  hideTimeout?: any
+}
```
Toti91, line 8: I went down this rabbit hole once of typing a timeout value, and learned that typescript has support for it with `NodeJS.Timeout`

```ts
hideTimeout?: string | number | NodeJS.Timeout
```

libs/island-ui/core/src/lib/DropdownMenu/useMenuHoverProps.tsx
```diff
@@ -0,0 +1,40 @@
+import React from 'react'
+import { MenuStateReturn } from 'reakit/Menu'
+
+interface HoverPropsRefState {
+  hoverCount: number
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  hideTimeout?: any
```
Toti91, line 7:
```suggestion
  hideTimeout?: string | number | NodeJS.Timeout
```

Toti91 (changes requested):
one comment

libs/island-ui/core/src/lib/DropdownMenu/useMenuHoverProps.tsx
eirikurn: Yeah I already saw this. 

Just annoying that typescript has a different type from browsers and to need NodeJS types in our react component library. But that's probably fine? 🤷‍♂️

I'll try it out.

RunarVestmann (changes requested):
I noticed that on organization pages (like /s/syslumenn) the page scrolls when you stop hovering the button. Is there something we can do about that?

libs/island-ui/core/src/lib/DropdownMenu/useMenuHoverProps.tsx
Toti91: Yeah thats true. Someone here mentions using `ReturnType` 
https://stackoverflow.com/a/56239226/9491688

eirikurn: Oooh, that's much cleaner!

libs/island-ui/core/src/lib/DropdownMenu/useMenuHoverProps.tsx
```diff
@@ -0,0 +1,40 @@
...
+  const hoverRef = React.useRef<HoverPropsRefState>({
+    hoverCount: 0,
+    hideTimeout: undefined,
+  })
+  const hoverState = hoverRef.current
+
+  return openOnHover
+    ? {
+        onMouseEnter: () => {
+          hoverState.hoverCount++
```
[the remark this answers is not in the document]
eirikurn: Yeah, it's defensive code since these event handlers is added to two elements and I don't want to encounter issues with event ordering now or later.

Also found a way to fix the page scrolling.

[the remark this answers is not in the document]
RunarVestmann: Cool

PR #10416, opened 2023-03-10
fix(regulations-admin): Update admin backend client

thordurhhh:
## What

* Change `RegulationsAdminClientService` from `RESTDataSource`.
* OpenAPI and generated client with Enhanced Fetch.
* Minor regulation-admin home screen change, display error message and empty screen message.

## Why

* This is the preferred way.


thordurhhh:
Moving over to: https://github.com/island-is/island.is/pull/11043

PR #10446, opened 2023-03-10
chore(j-s): Add defender assigned notifications to queue

unakb:
#  Add defender assigned notifications to queue

[Setja DEFENDER_ASSIGNED tilkynningar á biðröð](https://app.asana.com/0/1199153462262248/1203640056170092)

## What
Added defender assigned notifications to queue

## Why
Move the last notification to queue so that delivery is more reliable and the web is not in charge of triggering the delivery

## Checklist:


apps/judicial-system/backend/src/app/modules/notification/notification.controller.ts
```diff
@@ -81,6 +81,7 @@ export class NotificationController {
         NotificationType.HEADS_UP,
         NotificationType.READY_FOR_COURT,
         NotificationType.COURT_DATE,
+        NotificationType.DEFENDER_ASSIGNED,
```
gudjong, line 84: Please consider removing the call to sendCaseNotification below, as all notifications should now be queued.

apps/judicial-system/message-handler/src/app/messageHandler.service.ts
```diff
@@ -170,6 +170,15 @@ export class MessageHandlerService implements OnModuleDestroy {
           { type: NotificationType.REVOKED },
         )
         break
+      case MessageType.SEND_DEFENDER_ASSIGNED_NOTIFICATION:
```
gudjong, line 173: Consider adding a test for this case in `messageHandler.spec.ts`.

gudjong (changes requested):

apps/judicial-system/backend/src/app/modules/notification/notification.controller.ts
unakb: I cleaned up the user types. After doing so I feel like we should change the name of the notification controller method to "sendCaseNotificationToQueue" and adjust the return type, since that's a more accurate representation of what the method is actually doing now. I think that should be a different PR though because it requires quite a few more changes to more files.

gudjong (approved):

PR #10450, opened 2023-03-12
chore(deps): Upgrade dd-trace

eirikurn:
## What

Upgrade dd-trace to latest.

## Why

The version we were using has a known memory leak which might explain some of our memory issues in production.

The [2->3 migration](https://github.com/DataDog/dd-trace-js/blob/master/MIGRATING.md) should be safe. 

## Checklist:


saevarma (approved):
Hope this helps with our memory leak.

PR #10466, opened 2023-03-13
feat(auth-api-lib): Auth api lib migrate clients to add domain relation

saevarma:
[Notion taks](https://www.notion.so/DB-migrate-clients-model-to-link-to-domains-tenants-4b878dd6753e436f88bec3bf23056bb6?pvs=4)

## What

Add `domainName` column to `Client` model to add relation to domains (tenants).
A client must and can only belong to single domain.

## Why

Required for the new IDS Admin.

## Screenshots / Gifs

N/A

## Checklist:


libs/auth-api-lib/src/lib/clients/models/client.model.ts
```diff
@@ -50,6 +53,16 @@ export class Client extends Model {
...
   clientType!: string
 
+  @ForeignKey(() => Domain)
+  @Column({
+    type: DataType.STRING,
+    allowNull: true,
+  })
+  domainName?: string
+
+  @BelongsTo(() => Domain)
```
valurefugl, line 63: It would be convenient for the IDS if this was also exposed in the API.

saevarma: Do you want to do that in a follow up PR when the IDS is updated as well?

valurefugl: Ok.

valurefugl (approved):

PR #10472, opened 2023-03-13
chore: Add util-deprecate to esbuild.json files

eirikurn:
## What

Fix esbuild node_modules handling.

## Why

Because it's flaky and broke after a recent dependency change.

## Checklist:

### DIFF
--- a/apps/air-discount-scheme/api/esbuild.json
+++ b/apps/air-discount-scheme/api/esbuild.json
@@ -20,6 +20,7 @@
     "graphql",
     "pg",
     "winston",
+    "util-deprecate",
     "source-map-resolve",
     "atob",
     "logform",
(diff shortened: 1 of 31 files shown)


busla (approved):
Do we need this package as a dependency? What about borrowing the business logic and add to shared so we can customize, e.g. use our logger instead of `console` and other things that might come up?

https://github.com/TooTallNate/util-deprecate/blob/master/browser.js

eirikurn:
No this is a dependency of a dependency of a dependency. It's a nightmare of our esbuild node_modules build system which breaks our builds every few weeks. 😪

Need to figure out a proper fix for this. Can chat on slack.

busla:
ahh ok! sure lets slack.

PR #10486, opened 2023-03-14
feat: set opening hours on feature deployments

busla:
# ...

## What

## Why

## Screenshots / Gifs

## Checklist:


charts/islandis/values.dev.yaml
```diff
@@ -851,6 +851,7 @@ consultation-portal:
     ENVIRONMENT: 'dev'
     NODE_OPTIONS: '--max-old-space-size=464'
     SERVERSIDE_FEATURES_ON: ''
+    UNUSED: 'na'
```
brynjarorng, line 854: Is there not a better name available for this parameter?

busla: oops, I was triggering a feature-deploy to test, not meant to be merged.

brynjarorng: We should also change the feature deployment message to tell users that deployments are only available for 24 hours, possibly also mention when it goes away (Feature deployment is available until XXXX). It would probably also make sense to change how we display the message. Currently we just edit the original comment which can get lost in a large PR. I think it would make sense to post at the bottom of the PR and possibly also remove the deploy feature label from the PR to minimize possible confusion for the devs when the deployment expires.

busla: absolutely, that work will be done in infra, from the job that deletes the namespace.

busla:
test 01

andes-it:

andes-it:

PR #10491, opened 2023-03-14
chore(j-s): Use generated type for SessionArrangements in frontend

oddsson:
# Use generated type for SessionArrangements in frontend

[Asana](https://app.asana.com/0/1199153462262248/1204174811173675/f)

## What

Use generated type for SessionArrangements in frontend

## Why

We are moving towards using generated types in our frontend. This is a part of that.

## Checklist:


unakb (approved):
Nice, just the one import that should be moved up 👍

unakb (changes requested):
Wait ok the build seems to be broken ☹️

PR #10498, opened 2023-03-15
fix(j-s): Fix spaces in indictment description autofill

oddsson:
# Fix spaces in indictment description autofill

[Asana](https://app.asana.com/0/1199153462262248/1204113546823724/f)

## What

Fix spaces in indictment description autofill

## Why

It's a bug

## Screenshots / Gifs

Screenshot 2023-03-15 at 09 10 20
Screenshot 2023-03-15 at 09 11 00

## Checklist:


oddsson:
In the future, I think we should be careful about not putting spaces in CF strings. By doing that, we are putting the responsibility of having a correctly formatted string on the content editors. They should only need to think about what words should be used, not where spaces are used.

unakb (changes requested):

unakb (approved):
Great ☺️

PR #10518, opened 2023-03-15
fix(ci): feature-deploy url's missing

peturgq:
# ...

## What

## Why

## Screenshots / Gifs

## Checklist:


.github/workflows/push.yml
```diff
@@ -623,7 +623,7 @@ jobs:
         run: |
           ./scripts/build-docker-container.sh $DOCKER_TAG
           echo "COMMENT<<EOF" >> $GITHUB_ENV
-          printf "Affected services are: ${{needs.prepare.outputs.IMAGES}}\n" >> $GITHUB_ENV
+          printf "Affected services are: `${{needs.prepare.outputs.IMAGES}}`\n" >> $GITHUB_ENV
```
busla, line 626: I dont think backticks are applicable here, they are (were, deprecated now) used for command substition, e.g. `foo=$(echo "bar")`

The `${{needs.prepare.outputs.IMAGES}}` variable is a string.

krrrrrrrrrrr, line 626: Why not `echo`? Prevents breakage in case `IMAGES` ever includes `%`

busla (changes requested):
backticks in bash consideration

.github/workflows/push.yml
krrrrrrrrrrr: I like to use `'` for delimination in these cases

busla: that would end up as a literal string and would output the variable name instead of the value

busla: probably because of the `\n` ?

busla: backticks are still there

busla (approved):
LGTM

busla (approved):
LGTM

PR #10520, opened 2023-03-15
fix(system-e2e): CI failure

AndesKrrrrrrrrrrr:
- Fix failures in CI related to versioning
- Reduce image size
- Reduce build time
- Dynamic playwright versioning from `package.json`

## Checklist:


apps/system-e2e/package.json
```diff
@@ -4,19 +4,21 @@
...
-    "mailparser": "3.5.0",
-    "nodemailer": "6.7.2",
-    "playwright-tesults-reporter": "1.0.1",
-    "react-intl": "5.13.1",
-    "slack-notify": "2.0.3"
-  },
-  "scripts": {},
-  "volta": {
-    "node": "16.17.0"
+    "@anev/ts-mountebank": "*",
```
busla, line 7: not sure this is advisable, I think we would rather stick to pinned versions

scripts/ci/Dockerfile
```diff
@@ -112,34 +112,41 @@ USER runner
...
 
 FROM playwright-base as output-playwright
+# TODO: remove awscli dependency (157 MB extra)
+RUN apt-get update && apt-get install -y zip awscli && apt-get purge
 
 ARG APP_HOME
 ARG APP_DIST_HOME
 
 WORKDIR ${APP_DIST_HOME}
-ENV APP_DIST_HOME=${APP_DIST_HOME}
```
busla, line 125: are you sure `APP_DIST_HOME` as ENV can be scrapped for ARG? If I remember correctly, it's used as env at runtime.

scripts/ci/Dockerfile
```diff
@@ -112,34 +112,41 @@ USER runner
...
-RUN yarn install
+COPY ${APP_HOME}/package.json .
+COPY ${APP_HOME}/yarn.lock .
+RUN yarn install --immutable
 RUN yarn playwright install
-COPY --chown=pwuser:pwuser --from=builder /build/${APP_DIST_HOME} /${APP_DIST_HOME}/
-COPY --chown=pwuser:pwuser ${APP_HOME}/entrypoint.sh .
-USER pwuser
+# TODO: Avoid copying build apps
+COPY --from=builder /build/${APP_DIST_HOME} /${APP_DIST_HOME}/
```
busla, line 139: doesn't  `pwuser` need ownership of the dist dirs to be able to run the application?

scripts/ci/_docker.sh
```diff
@@ -11,21 +11,31 @@ APP_DIST_HOME=$(jq ".targets.build.options.outputPath" -r < "$PROJECT_ROOT"/"$AP
...
+    --build-arg="APP=${APP}" \
+    --build-arg="APP_HOME=${APP_HOME}" \
+    --build-arg="APP_DIST_HOME=${APP_DIST_HOME}" \
     -t "${DOCKER_REGISTRY}""${APP}":"${DOCKER_TAG}" \
+    --build-arg="PLAYWRIGHT_VERSION=${PLAYWRIGHT_VERSION}" \
+  )
+  for extra_arg in ${EXTRA_DOCKER_BUILD_ARGS:-}; do
+    build_args+=("$extra_arg")
+  done
+  if ! docker --version | grep -q podman; then
```
busla, line 31: consider adding a new file, e.g. `_podman.sh` that sources `_docker.sh` and add the extra build_args there, to prevent contaminating the file with tooling that is not part of our build flow.

busla (changes requested):

scripts/ci/Dockerfile
krrrrrrrrrrr: :eyes:

scripts/ci/Dockerfile
krrrrrrrrrrr: Will use `pwuser` again

krrrrrrrrrrr: Checking with/without `COPY`

scripts/ci/Dockerfile
krrrrrrrrrrr: Doesn't seem like it :shrug:

scripts/ci/_docker.sh
krrrrrrrrrrr: Agreed; done

scripts/ci/Dockerfile
krrrrrrrrrrr: Using `pwuser` now :smile:

scripts/ci/Dockerfile
krrrrrrrrrrr: I.e. I tried running without this set in my env when running the image locally without any error

apps/system-e2e/package.json
eirikurn: Even with yarn.lock file, there are many ways for devs to accidentally upgrade packages without realising it or checking for regressions.

I personally think ^ (eg ^3.145.0) is a fine compromise since:

- It simplifies deduplicating packages in node_modules.
- Guarantees that deps won't upgrade to a new backwards incompatible major without an explicit package.json change.

eirikurn (changes requested):

busla (approved):
LGTM

eirikurn (approved):

PR #10522, opened 2023-03-15
feat(auth-admin): Add GraphQL for tenants/:tenantsId endpoint

GunnlaugurG:
# ...

https://www.notion.so/Add-GraphQL-for-tenants-tenantsId-endpoint-946bba40bedf48f786cc03a8ffb1bddb?pvs=4

## What

Added a graphQL layer to get tenant by id for all environments

## Why

so it can be fetched by the client

## Screenshots / Gifs

## Checklist:


libs/api/domains/auth-admin/src/lib/tenant/tenant.resolver.ts
```diff
@@ -39,4 +39,9 @@ export class TenantResolver {
   resolveAvailableEnvironments(@Parent() tenant: Tenant): Environment[] {
     return tenant.environments.map((env) => env.environment)
   }
+
+  @Query(() => Tenant, { name: 'authAdminTenantById' })
```
saevarma, line 43: I think our gql naming convention is to have this as singular `authAdminTenant`.
```suggestion
  @Query(() => Tenant, { name: 'authAdminTenant' })
```

libs/api/domains/auth-admin/src/lib/tenant/tenants.service.ts
```diff
@@ -147,4 +147,46 @@ export class TenantsService {
...
+        .catch(this.handleError.bind(this)),
+      this.adminStagingApiWithAuth(user)
+        ?.meTenantsControllerFindById({ id })
+        .catch(this.handleError.bind(this)),
+      this.adminProdApiWithAuth(user)
+        ?.meTenantsControllerFindById({ id })
+        .catch(this.handleError.bind(this)),
+    ])
+
+    const tenantMap = new Map<string, TenantEnvironment[]>()
```
saevarma, line 164: This should just be an simple `TenantEnvironment[]` as we are just working with a single tenant and don't need a map of multiple tenant ids.

libs/api/domains/auth-admin/src/lib/tenant/tenants.service.ts
```diff
@@ -147,4 +147,46 @@ export class TenantsService {
       pageInfo: { hasNextPage: false },
     }
   }
+
+  async getTenantById(id: string, user: User): Promise<Tenant> {
```
saevarma, line 151: This function should replace `getTenant` above, so just remove `getTenant` and use this instead.

saevarma (approved):
LGTM

PR #10526, opened 2023-03-16
feat(license-api): add TR client/scope

thorkellmani:
## What
* Adds a machine client and scope for TR

## Why

They need access to the license-api

## Checklist:


eirikurn (approved):
This works, but someone will have to manually create the secret in each environment and send it to the right people. For staging/production this can be requested in #ids_changes.

PR #10529, opened 2023-03-16
feat(admin-portal): Add application system to admin portal

Thorrik58:
# Adding Application system overview admin to the admin portal

## What

Application system overview admin to the admin portal

## Why

To give island.is tools to help users with their applications.

## Screenshots / Gifs

image

## Checklist:


obmagnusson (changes requested):

libs/island-ui/core/src/lib/Drawer/Drawer.tsx
```diff
@@ -0,0 +1,75 @@
...
+   * Show immediately without clicking the disclosure button
+   */
+  initialVisibility?: boolean | undefined
+  /**
+   * Position of the drawer
+   */
+  position?: 'right' | 'left'
+}
+
+export const Drawer = ({
```
eirikurn, line 32: What does this do? Does it feature deploy somewhere? Figma link?

libs/island-ui/core/src/lib/Toast/Toast.css.ts
```diff
@@ -10,7 +10,7 @@ export const root = style(
 export const useMotion = style({}, 'useMotion')
 
 globalStyle(`${root} .Toastify__toast-container`, {
-  zIndex: 9999,
+  zIndex: 99999,
```
eirikurn, line 13: Would be nice to transition to the new zIndex constants in theme 🙈

eirikurn (commented):
Exciting! 🤩

eirikurn (approved):
Changing to approved since all of my comments are suggestions and "nice to have". 👍

thordurhhh (approved):
service-portal lgtm

libs/island-ui/core/src/lib/Drawer/Drawer.tsx
Toti91: It's basically a Modal on the side if the screen

See design here (when clicking on a table row): https://www.figma.com/proto/Nz4FjxCAhpaHqjyK8mrDEy/Ums%C3%B3knarkerfi%C3%B0?node-id=1995-44442&starting-point-node-id=1995%3A44442

Thought it would make sense to have as a shared UI component as it will probably be needed elsewhere in the future, we will get our designer to add this as a proper UI component in Figma

libs/island-ui/core/src/lib/Toast/Toast.css.ts
Toti91: The problem i faced here was that when opening a toast inside a modal, the toast always renders behind the modal because Reakit is forcing a z-index of some ridiculous (9999+) value on all it's modals

I can see that we have zIndex constants `belowModal: 19` `modal: 20` `aboveModal: 21`
These are only used when rendering a Reakit `Menu`, but I guess `Modal` is different

I guess we could add a new constant `aboveEverything: 99999` ?

bjornlarus (approved):

Toti91 (approved):

apps/application-system/api/src/app/modules/application/utils/applicationUtils.spec.ts
```diff
@@ -0,0 +1,114 @@
...
+              status: 'success',
+            },
+          },
+        }),
+      ),
+    ).toEqual('Test User')
+  })
+})
+
+it('Should return name of the application when a function isnt defined just a string', () => {
```
obmagnusson, line 65:
```suggestion
it('Should return the name of the application when defined with a string', () => {
```

obmagnusson (approved):
Nice !

PR #10530, opened 2023-03-16
fix(contentful-apps): grant external ingress

busla:
# ...

## What

## Why

## Screenshots / Gifs

## Checklist:


apps/contentful-apps/pages/_app.tsx
```diff
@@ -16,3 +16,5 @@ function CustomApp({ Component, pageProps }: AppProps) {
 }
 
 export default CustomApp
+
+// deleteme
```
krrrrrrrrrrr, line 20: `deleteme`

busla: yeah, this workflow needs to be improved.

We need to modify a file in the app that we want to deploy as feature so that NX triggers it as affected.

PR #10477, opened 2023-03-13
feat(ids-admin): ApplicationDetails screen and graphql layer

GunnlaugurG:
## What

Added screen to view application details, and a graphql layer that returns mock data. The screen is ready to POST to server with action pattern

Auth-admin front end
- Added screen to display application information with edit buttons and isDirty check.
- Connected it to the graphql layer.

Auth-admin GraphQL layer
- added application id to the getApplication to fetch an application by applicationId and tenantId

Island UI 
 - added type "password" to Input component props
 - made the button in Tabs by of type button so it does not post on tab change

## Why
To get and display application details information, and later on add edit functionality to that screen

## Screenshots / Gifs

Attach Screenshots / Gifs to help reviewers understand the scope of the pull request

## Checklist:

- [x] I have performed a self-review of my own code
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] Formatting passes locally with my changes
- [x] I have rebased against main before asking for a review

PR #10534, opened 2023-03-16
refactor(auth-admin): Refactored Application details screen

GunnlaugurG:
Made changes from PR https://github.com/island-is/island.is/pull/10477 

Few changes from previous PR that was merged to main
- Fixed Application schema
- Stopped fetching the default environment
- Moved getApplicaitonById to new Query
- Changed input of getApplications to only fetch by tenant id

Small refactor and fixes

## Screenshots / Gifs

## Checklist:


libs/portals/admin/ids-admin/src/components/Application/Application.loader.ts
saevarma, line 1: Should we also refactor `AuthApplicationList` to be `AuthApplication` ? As this is working with single application not list of applications?

GunnlaugurG: Yes you are right, this is a copy paste error from be.

saevarma (approved):
LGTM

PR #10535, opened 2023-03-16
chore(island-ui): Render tags as span if href and onClick props are not set

oddsson:
# Render tags as span if href and onClick props are not set

## What

Render tags as span if href and onClick props are not set

## Why

Tags should not be clickable by default, but can rather be made clickable via props. This is better for a11y reasons.

## Checklist:


libs/island-ui/core/src/lib/Tag/Tag.tsx
```diff
@@ -105,6 +105,8 @@ export const Tag = forwardRef<HTMLButtonElement & HTMLAnchorElement, TagProps>(
       >
         {content}
       </button>
+    ) : (
+      <span {...sharedProps}>{content}</span>
```
eirikurn, line 109: Also include `{...props}` which afaics only allows the "id" prop at the moment.

eirikurn (commented):
Great!

eirikurn (approved):
LGTM

RunarVestmann (approved):

PR #10537, opened 2023-03-16
chore(j-s): And in Legal Arguments

gudjong:
# And in Legal Arguments

Heimfærsla - logík varðandi og](https://app.asana.com/0/1199153462262248/1204127104755246/f)

## What

- Inserts and into legal arguments.

## Why

- That is what the prosecutors want.

### With article 95 and only one other article:
image

### With article 95 and more than one other article:
image

### Without article 95 and only one other article:
image

### Without article 95 and more than one other article:
image

## Checklist:


oddsson (changes requested):
Me and @unakb decided to put more emphasis on tests in the front-end so please consider writing tests for this in https://github.com/island-is/island.is/blob/main/apps/judicial-system/web/src/routes/Prosecutor/Indictments/Indictment/IndictmentCount.spec.tsx 🙏

oddsson (approved):

PR #10538, opened 2023-03-16
feat(license-service): generating updates pkpass

thorkellmani:
## What

* When generating a pkpass, if there is an existing pass it'll be updated according to the live license data.

## Why

Generating a pkpass should update the pass as well

## Checklist:


libs/clients/smartsolutions/src/lib/smartSolutions.api.ts
```diff
@@ -210,49 +210,92 @@ export class SmartSolutionsApi {
...
-        },
+        data: res.data.upsertPass,
       }
     }
 
     return res
   }
 
-  async upsertPkPass(payload: PassDataInput): Promise<Result<Pass>> {
+  private async updateExistingPass(
```
osk, line 241: Shouldn't this also update the `expiryDate` in Smart?

libs/clients/smartsolutions/src/lib/smartSolutions.api.ts
```diff
@@ -210,49 +210,92 @@ export class SmartSolutionsApi {
...
+      //if failure, return the response
+      return getPassRes
+    }
+    const passInputData = mapPassToPassDataInput(getPassRes.data.pass)
+
+    const inputFieldValues = mergeInputFields(
+      passInputData.inputFieldValues ?? undefined,
+      payload.inputFieldValues ?? undefined,
+    )
+    //now we finally have the updated pass data!
```
osk, line 264:
```suggestion
```

it's an easy process now, no need to be dramatic 😎

libs/clients/smartsolutions/README.md
```diff
@@ -1,5 +1,6 @@
 # Smart Solutions
 
+A library enabling interaction with of digital licenses.
```
osk, line 3: with what of?

osk (changes requested):

libs/clients/smartsolutions/README.md
thorkellmani: 😩

libs/clients/smartsolutions/src/lib/smartSolutions.api.ts
thorkellmani: it does, the expiryDate is part of the payload which the client sends over
image

thorkellmani: We just have to be careful to do the same with other clients, which is on the to-do list

libs/clients/smartsolutions/src/lib/smartSolutions.api.ts
thorkellmani: It wasn't until it was!

PR #10544, opened 2023-03-17
fix(j-s): Don't check for a change in courtDate if a notification has not been sent

oddsson:
# Don't check for a change in courtDate if a notification has not been sent

[Asana](https://app.asana.com/0/1199153462262248/1203752015941957/f)

## What

When court users set a court date they are asked if they want to send a COURT_DATE notification. If they choose to do so and then change the court date later they are again asked if they want to send a new notification. This PR fixes a bug where we are not checking if the notification had been sent before asking if the user wants to send a new notification. The bug could be reproduced in these steps.

1. Set a court date as a court user and don't send a notification
2. Go back and change the court date.
3. See that you are asked if you want to send a _new_ notification.

You should not be asked if you want to send a new notification because you had never sent a notification in the first place. 

## Why

It's a bug.

### Before 

https://user-images.githubusercontent.com/3789875/225880379-248f05b8-2538-4f84-bf61-df2e52052f44.mov

### After

https://user-images.githubusercontent.com/3789875/225879093-c1808169-9478-4c28-af58-204f322d05da.mov

## Checklist:


unakb (approved):
lgtm :)

apps/judicial-system/web/src/utils/stepHelper.ts
```diff
@@ -108,3 +111,22 @@ export const isTrafficViolationCase = (
...
+) => {
+  if (!notifications || notifications.length === 0) {
+    return false
+  }
+
+  const n = notifications.filter(
+    (notification) => notification.type === notificationType,
+  )
+
+  const latestN = n.sort((a, b) =>
```
gudjong, line 127: Notifications are returned in order, starting with the latest notification.

Consider removing the sort.

gudjong (changes requested):

gudjong (approved):

PR #10583, opened 2023-03-21
fix(island-ui): Add missing React import to DialogPrompt

hedinne:
## What

`<DialogPrompt>` was missing the React import. I changed the ReactElement to React.ReactElement so that React isn't shown as an unused import in the IDE (and someone might think it wasn't being used).

## Checklist:


RunarVestmann (approved):
Is the `import React from 'react'` import important for storybook or why is this more beneficial than importing { ReactElement } directly?

hedinne:
Depending on how React code is transpiled, sometimes you need to import React into the component file and sometimes you don't. However, it looks like, in island-ui, you do need it (or at least in Storybook). 
This component used to include this import, but it was removed in a resent refactor. I guess it was removed because it shows as unused in some editors (see picture below). 

The import statement in the picture is perfectly valid but I wanted to change it to minimize the risk that someone would accidentally remove it. Of course, people can use import destructuring if they wish to, but in this case, I thought it was unnecessary since I was already importing React.

image

PR #10589, opened 2023-03-21
feat(auth-admin): Graphql layer to create client

GunnlaugurG:
# ...

## What

Adding a graphql layer to create a client in db

## Why

to connect the front end to the graphql layer to create client

## Screenshots / Gifs

## Checklist:


libs/api/domains/auth-admin/src/lib/shared/services/multi-environment.service.ts
```diff
@@ -0,0 +1,40 @@
...
+import {
+  AdminApi,
+  AdminDevApi,
+  AdminProdApi,
+  AdminStagingApi,
+} from '@island.is/clients/auth/admin-api'
+import { Auth, AuthMiddleware } from '@island.is/auth-nest-tools'
+
+@Injectable()
+export class MultiEnvironmentService {
```
saevarma, line 13: We should have this as an `abstract` class
```suggestion
export abstract class MultiEnvironmentService {
```

libs/api/domains/auth-admin/src/lib/application/applications.service.ts
```diff
@@ -23,27 +24,52 @@ export class ApplicationsService {
...
+            )
+          case Environment.Staging:
+            return this.adminStagingApiWithAuth(
+              user,
+            )?.meClientsControllerCreate(applicationRequest)
+          case Environment.Production:
+            return this.adminProdApiWithAuth(user)?.meClientsControllerCreate(
+              applicationRequest,
+            )
+        }
```
saevarma, line 59: We should move this to the MultiEnvironment abstract class as `adminApiByEnvironmentWithAuth` helper so we can use it in other methods as update and delete.

```
adminApiByEnvironmentWithAuth(environment: Environment, user: User) {
       switch (environment) {
          case Environment.Development:
            return this.adminDevApiWithAuth(user)?
          case Environment.Staging:
            return this.adminStagingApiWithAuth(user)?
          case Environment.Production:
            return this.adminProdApiWithAuth(user)?
        }
}
```


```suggestion
        return this.adminApiByEnvironmentWithAuth(environment, user)?.meClientsControllerCreate(
          applicationRequest,
        )
```

libs/api/domains/auth-admin/src/lib/application/applications.service.ts
```diff
@@ -23,27 +24,52 @@ export class ApplicationsService {
...
-      const applicationEnv = new ApplicationEnvironment()
-      applicationEnv.name = input.applicationId
-      applicationEnv.environment = env
-      applicationEnv.displayName = [displayName]
-      applicationEnv.ApplicationUrls.callbackUrls = []
+    const createApplicationResponse = [] as CreateApplicationResponseDto[]
 
-      return applicationEnv
-    })
+    const created = await Promise.all(
```
saevarma, line 44: This will reject if there is an error in one of the environment. We should use `Promise.allSettled` instead which will return an object with the promise status and the value. Then we can check the status and map the response accordingly.

libs/api/domains/auth-admin/src/lib/application/applications.service.spec.ts
```diff
@@ -0,0 +1,173 @@
...
+      load: [AuthAdminApiClientConfig],
+    }),
+  ],
+  controllers: [],
+  providers: [ApplicationsResolver, ApplicationsService],
+  exports: [ApplicationsResolver],
+})
+class TestModule {}
+
+describe('ApplicationsService', () => {
```
saevarma, line 49: ~ A test case to test when one env throws an error.

saevarma (approved):
Nice job 💯

PR #10592, opened 2023-03-21
fix(estate): further bugfixing of the estate applications

albinagu:
## Checklist:


libs/application/templates/estate/src/lib/dataSchema.ts
```diff
@@ -3,22 +3,40 @@ import { m } from './messages'
...
+const checkIfFilledOut = (arr: Array<string | undefined>) => {
+  if (arr.every((v) => v === '')) {
+    return true
+  } else if (arr.every((v) => v !== '')) {
+    return true
+  } else {
+    return false
+  }
+}
+
```
kksteini, line 21: Can be made more concise if you want :)
```suggestion
const checkIfFilledOut = (arr: Array<string | undefined>) => {
  if (arr.every((v) => v === '')) {
    return true
  } else if (arr.every((v) => v !== '')) {
    return true
  }
  return false
}

```

kksteini (approved):

PR #10598, opened 2023-03-21
fix(phone-input): Hotfix: change feature flag functionality

Toti91:
# Change PhoneInput feature flag functionality

## What

Feature flag now disables the country code dropdown instead of rendering a normal text input.

## Why

It was causing major problems because of zod schema validation between the two different inputs.
The PhoneInput prefixes the value with a country code, so it needs a different validation than a normal text input, and using a feature-flag to differentiate between two zod-schemas seems impossible

Also, adding the option to disable the dropdown with a prop e.g. 
```ts
buildPhoneField({
  id: 'applicant.phoneNumber',
  title: "Phone",
  disableDropdown: true,
}),
```

## Screenshots / Gifs

https://user-images.githubusercontent.com/16031201/226647386-d57f61c7-08c2-4cb9-aed3-6fbf301e726d.mp4

### Dropdown disabled
image

### Dropdown enabled
image

## Checklist:

### DIFF
--- a/libs/island-ui/core/src/lib/PhoneInput/CountryCodeSelect/Components/index.tsx
+++ b/libs/island-ui/core/src/lib/PhoneInput/CountryCodeSelect/Components/index.tsx
@@ -49,12 +49,14 @@ export const IndicatorsContainer = (
   const { icon } = props.selectProps
   const size: CountryCodeSelectProps['size'] = props.selectProps.size || 'md'
   const hasLabel: boolean = props.selectProps.inputHasLabel
+  const disabled: boolean = props.selectProps.isDisabled ?? false
   return (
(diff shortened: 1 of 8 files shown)


egillben (approved):

libs/island-ui/core/src/lib/PhoneInput/CountryCodeSelect/Components/index.tsx
RunarVestmann, line 1: `libs/island-ui/core/src/lib/PhoneInput/CountryCodeSelect/Components/index.tsx`

I find it odd that the folder `Components` has a capital letter. But that's just a nitpick :)

RunarVestmann (approved):

PR #10603, opened 2023-03-21
feat(j-s): Screen for appeals to court of appeals

oddsson:
# Screen for appeals to court of appeals

[Asana](https://app.asana.com/0/1199153462262248/1204127104511335/f)

## What

Screen for appeals to court of appeals for prosecutors and defenders. 

**NOTE**

There are a couple of broken things in this PR. There are tickets in Asana for them all and they will be implemented later.

- Files are not sent to Auður.
- The form stepper is wrong.
- No one can upload anything because until now we assume that files are never uploaded after a case is done. 

## Screenshots / Gifs

Kæra til Landsréttar - Réttarvörslugátt 2023-03-24

## Checklist:


apps/judicial-system/backend/migrations/20230322092130-update-case.js
unakb, line 1: Change APPEAL_CASE_FILE to APPEAL_BRIEF_CASE_FILE and add PROSECUTOR_APPEAL_STATEMENT_CASE_FILE

libs/judicial-system/types/src/lib/user.ts
```diff
@@ -60,7 +60,11 @@ export const prosecutionRoles: string[] = [
   UserRole.REPRESENTATIVE,
 ]
 
-export function isProsecutionRole(role: string): boolean {
+export function isProsecutionRole(role?: string): boolean {
+  if (!role) {
```
unakb, line 64: Just wondering, why would we call this with an empty string? Wouldn't we then already know that no role is not a prosecution role?

unakb (approved):
Just one question but not a showstopper :)

libs/judicial-system/types/src/lib/user.ts
oddsson: You are correct, you wouldn't call this with an empty string. However, the way our api is set up, this fields comes to the frontend as potentially undefined. So instead of checking everywhere before we use this function if role is not undefined, I decided to do it just once, inside the function.

PR #10610, opened 2023-03-21
fix(system-e2e): Broken tests

AndesKrrrrrrrrrrr:
## Checklist:

### DIFF
--- a/apps/system-e2e/package.json
+++ b/apps/system-e2e/package.json
@@ -3,19 +3,7 @@
   "version": "1.0.0",
   "main": "index.js",
   "license": "MIT",
-  "dependencies": {
-    "@aws-sdk/client-ses": "^3.145.0",
-    "@formatjs/intl": "^1.8.1",
-    "@playwright/test": "^1.29.2",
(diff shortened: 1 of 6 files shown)


busla (approved):
LGTM

apps/system-e2e/package.json
```diff
@@ -3,19 +3,7 @@
...
-    "axios": "^0.21.4",
-    "imap-simple": "^5.1.0",
-    "mailparser": "^3.5.0",
-    "nodemailer": "^6.7.2",
-    "playwright-tesults-reporter": "^1.0.1",
-    "react": "^18.2.0",
-    "react-intl": "^6.3.0",
-    "slack-notify": "^2.0.3"
-  },
+  "dependencies": {},
```
eirikurn, line 6: This may work locally but I'm worried that the production build won't have the dependencies it needs to run, since this isn't using the same esbuild magic as other apps. Can you verify?

AndesKrrrrrrrrrrr: You're right. This didn't even build locally :disappointed:

PR #10612, opened 2023-03-22
fix(delegations): Fix regression when switching domains

eirikurn:
## What

It did not work to change domains on the grant access screen. The form would be silently invalid with an empty national id after reseting the form.

This is another RHF regression. :(

## Why

So we can launch.

## Checklist:


saevarma (approved):
LGTM

PR #10620, opened 2023-03-22
feat(core): Support overriding json body limit

eirikurn:
## What

Allow overriding the request body size limit in bodyParser.json(). 

Also refactors @island.is/infra-nest-server and adds tests.

## Why

The regulation resolvers need to handle larger json submissions than we supported before.

## Checklist:


brynjarorng (approved):
LGTM

PR #10645, opened 2023-03-24
chore(j-s): Update Case with Prosecutor

gudjong:
# Update Case with Prosecutor

Skila gögnum í Auði um sóknaraðila í R-máli](https://app.asana.com/0/1199153462262248/1204203804330809/f)

## What

- Adds 200 OK error handling to the court client service.
- Adds more details to 400 error handling.

## Why

- The API returns some errors as 200 OK responses with explanation.
- A more detailed handling of 400 errors helps with debugging and system operation.

## Checklist:


oddsson (approved):
I know the reasons why we're listening for strings in error messages here instead of status codes but it still feels soo wrong 🫠

PR #10648, opened 2023-03-25
fix: Better esbuild node_modules copy logic

eirikurn:
## What

Patch bugs in @anatine/esbuildnx when it comes to copying node_modules to the dist folder.

Will submit this upstream when I have time to write some unit tests.

## Why

These bugs were randomly breaking our builds (with `cannot require 'something'` errors) whenever someone installed or upgraded a dependency which modified the node_modules structure.

The reason was that it wasn't recursively checking for missing sub-dependencies to copy. It only checked for sub-dependencies one level deep.

This PR modifies the copy logic to be more recursive.

## Checklist:


saevarma (approved):
💯

PR #10667, opened 2023-03-27
feat(driving-license-book): Update getMostRecentStudentBook to filter out cancelled books

thordurhhh:
## What

Update getMostRecentStudentBook to filter out cancelled books

## Why

Sometimes new books will be created instead of updating the current one, and those newly created books can have a created-date set so it will look like the old book is the most recent one.
This is a request from Samgöngustofa, that we filter out all books that have the status of `9`

## Screenshots / Gifs

## Checklist:


kksteini (approved):
LGTM

PR #10688, opened 2023-03-28
feat(j-s): Transition handles case appeal state transition 

unakb:
# Handling appeal state transitions

[Asana](https://app.asana.com/0/1199153462262248/1204127104755193/f)

## What

- Adjust the existing transition functionality so it can handle appeal state transitions
- Added new transition method to the limited access case controller to handle transitions triggered by defenders
- Send notification e-mail to judge when prosecutor appeals a case

## Why

So we can transition appealed cases and trigger events related to said transition

## Checklist:


apps/judicial-system/backend/src/app/modules/case/case.controller.ts
```diff
@@ -231,6 +244,16 @@ export class CaseController {
       update.courtRecordSignatureDate = null
     }
 
+    if (appealState !== undefined) {
+      // TODO: What if user is neither prosecutor nor defender?
```
unakb, line 248: I haven't completely thought this through yet. I feel like I'm unsure how defenders work here since they aren't the same types of users as prosecutors, judges etc. Can a prosecutor even get here since they don't have a rolesRules transition guard?

apps/judicial-system/backend/src/app/modules/case/case.controller.ts
```diff
@@ -231,6 +244,16 @@ export class CaseController {
...
+      if (user.role === UserRole.PROSECUTOR) {
+        update.prosecutorPostponedAppealDate = nowFactory()
+      } else if (user.role === UserRole.DEFENDER) {
```
gudjong, line 251: I suggest putting the defender appeal transitions into the limitedAccessCase controller.

apps/judicial-system/backend/src/app/modules/case/case.controller.ts
gudjong: I think the safest thing would be to not handle defenders here.

apps/judicial-system/message-handler/src/app/messageHandler.service.ts
```diff
@@ -178,7 +178,14 @@ export class MessageHandlerService implements OnModuleDestroy {
           { type: NotificationType.DEFENDER_ASSIGNED },
         )
         break
-
+      case MessageType.SEND_APPEAL_TO_COURT_OF_APPEALS_NOTIFICATION:
```
gudjong, line 181: Consider adding a test for this case

apps/judicial-system/backend/src/app/modules/case/case.controller.ts
unakb: The appeal transition for defenders is now in the limitedAccessController 🙌

apps/judicial-system/message-handler/src/app/test/handleMessage.spec.ts
```diff
@@ -504,7 +504,10 @@ describe('MessageHandlerService - Handle message', () => {
             'Content-Type': 'application/json',
             authorization: `Bearer ${config.backendAccessToken}`,
           },
-          body: JSON.stringify({ type: NotificationType.MODIFIED, userId }),
+          body: JSON.stringify({
+            type: NotificationType.READY_FOR_COURT,
```
gudjong, line 508: Shouldn't this be `MODIFIED` notification?

gudjong (changes requested):

gudjong (approved):

oddsson (approved):
Awwwwwesome!

PR #10695, opened 2023-03-29
fix(regulations): Multiple minor adjustments

thordurhhh:
## What

Some adjustments to the regulation portal
* Updated strings
* Updated default display time
* Updated search so no longer needed to only search for reg name, search string can now include anything in the regulation that's being searched for.

## Why

Some bugs fixed, some features improved.

## Checklist:


libs/portals/admin/regulations-admin/src/utils/index.ts
```diff
@@ -61,6 +61,29 @@ export const getNextWorkday = (date: Date) => {
...
+    }
+    d = addDays(d, 1)
+  }
+
+  d = getNextWorkday(d)
+
+  return d
+}
+
+// ---------------------------------------------------------------------------
```
thorkellmani, line 85: perfect recursion opportunity not taken smh 🙈

thorkellmani (approved):

PR #10656, opened 2023-03-27
feat(island-ui): Input v2

hedinne:
## What

Inputs can now accept multiple buttons with icons. This PR also implements the new design for Inputs (see [Figma](https://www.figma.com/file/Rb6Ga0RzPJluRZJf4i7eEv/Sandbox?node-id=4495-30816&t=hTw16u4qWEcJUVSf-4)). The most notable change in addition to the buttons is the background color of the error state.

- The input border had to be moved to a separate element because the buttons' background color overflowed the border. So now it is on `.container::before`. I reviewed all other components using the Input styles to ensure they haven't changed.
- The `icon` prop should only display static icons. Clickable icons are added to the `buttons` prop.
- This PR adds "@vanilla-extract/recipes" as a dependency. This is part of vanilla-extra and makes writing styles with different variants easier to read and maintain.  

## Why
These changes are required for features in the admin panel. This also fixes accessibility errors in the previous implementation of buttons.

#### Normal input. Focused. Visually hasn't changed.
Normal input. Visually hasn't changed.

#### Password Input with buttons. Focused
Password Input with buttons.

#### Small input with a non-clickable icon and two buttons.
Small input with a non-clickable icon and two buttons.

#### Error state
Error state

#### Extra small input with a clear button.
Extra small input with a clear button

Attach Screenshots / Gifs to help reviewers understand the scope of the pull request

## Checklist:

- [x] I have performed a self-review of my own code
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] Formatting passes locally with my changes
- [x] I have rebased against main before asking for a review

PR #10699, opened 2023-03-29
fix(island-ui): Input disabled state

hedinne:
## What

There was a bug In #10656 where the disabled state was only active if the input had no value. In a previous version of the component, the color of the label only changed if the input was both disabled and empty. This was not according to the design so this PR removes this functionality.

## Screenshots / Gifs

The Disabled states

## Checklist:


Toti91 (approved):
Nice!

PR #10708, opened 2023-03-30
feat(island-ui): Larger tooltip option

thordurhhh:
## What

Add option for full width tooltips.

This a suggestion on how it could be, I'm open to any other ideas.

## Why

Would be nice to be able to include more text on a single line. Will break with current implementation.

## Screenshots / Gifs

Screenshot 2023-04-11 at 14 51 59


oddsson (commented):
I like the idea of being able to have different sized tooltips. I am a little concerned about boxing us too much into specific sizes. I'm mostly afraid about what happens if I need a bigger tooltip than 360px, I'd have to do a code change to add a XL size prop.

How would you feel about counting the characters in the tooltip and applying inline width styles based on that. The implementation could either be something like

```
if (nr_of_chars < 100) size === 250px
else if  (nr_of_chars >= 100 && nr_of_chars >= 300 ) size === 360px
...
else size === (biggest possible size)px
```

Or somehow calculating the width in a more dynamic way using some math formula (I'm really bad at math 🙈)

thordurhhh:
> 

@oddsson 

Yeah, that makes sense. But as I was testing this out I thought, why stop there?
An option to make this as long as possible could work just as well with a fullWidth option.

oddsson (approved):

thorkellmani (approved):
lgtm

PR #10711, opened 2023-03-30
feat(portals-shared-modules-delegations): Dynamic scroll shadows

hedinne:
## What

This adds a hook that determines whether a scroll area is fully in view. If not then display a shadow on the sticky "footer". 

## Why

This functionality helps the users to notice when they can scroll.

#### Here is the list is fully in view so there is no shadow on the "footer"
localhost_4200_minarsidur_adgangsstyring_umbod_9ca1228a-9d2e-4ede-8f53-083ae3bca041 (1)

#### Here the list is not fully in view, and the shadow is used to indicate to the user that they can scroll down to see more.
localhost_4200_minarsidur_adgangsstyring_umbod_9ca1228a-9d2e-4ede-8f53-083ae3bca041

## Checklist:


saevarma (approved):
Nice!

PR #10712, opened 2023-03-30
feat(auth-admin): Form validation for edit client

GunnlaugurG:
# ...

- Added a validation for the form for clients edit.
- Also added delegation to the form.

so invalid data can not be saved to database.
- Extending the button component with value and name to be able to have multiple forms on one page with the react-router-dom: 6.4.0^ action pattern. That pattern only allows one form action per page by default, but extending the submit button with name: 'intent' and value for the action (ex. 'create', 'update') allows us to have multiple forms per page.

## Screenshots / Gifs

## Checklist:


hedinne (approved):
The code looks good to me. But, I would have liked a better explanation in the PR description on why you are extending the <Button /> component.

oddsson (approved):

kksteini (approved):
CODEOWNER files look good

PR #10714, opened 2023-03-31
feat(web): IBM Watson Chatbot - Zendesk authentication

RunarVestmann:
# IBM Watson Chatbot - Zendesk authentication

## What

* Secured the ibm watson assistant chat for the directorate of immigration via JWT that gets forwarded to Zendesk and displays the users name and email there.
* Added a custom panel to the directorate of immigration chat that prompts the user to enter his name and email

## Why

* This was requested by both Digital Iceland and the Directorate of Immigration institution

## Screenshots / Gifs

image

image

## Checklist:


andes-it (approved):

libs/api/domains/watson-assistant-chat/project.json
```diff
@@ -0,0 +1,23 @@
...
+    "test": {
+      "executor": "@nrwl/jest:jest",
+      "outputs": ["coverage/libs/api/domains/watson-assistant-chat"],
+      "options": {
+        "jestConfig": "libs/api/domains/watson-assistant-chat/jest.config.js",
+        "passWithNoTests": true
+      }
+    }
+  },
+  "tags": []
```
saevarma, line 22: You need to provide the correct tag for the Nx module boundary rule. So this tag would be `lib:api` so the api app can depend on this library.

libs/api/domains/watson-assistant-chat/.babelrc
saevarma, line 1: Wondering what Nx generator you used to create the library. This should just be a nest library, which doesn't require the .bablerc.

libs/api/domains/watson-assistant-chat/src/lib/dto/watsonAssistantChatIdentityToken.input.ts
```diff
@@ -0,0 +1,13 @@
+import { Field, InputType } from '@nestjs/graphql'
+
+@InputType()
+export class WatsonAssistantChatIdentityTokenInput {
```
saevarma, line 4: FYI you can also create the class and the file with shorter names and use the GQL decorator to control the schema name like:
```
@InputType('WatsonAssistantChatIdentityTokenInput')
export class IdentityTokenInput {
```

This is possible for the `ObjectType`, `InputType`, `Mutation`.  Just up to you what you prefer.

libs/api/domains/watson-assistant-chat/tsconfig.spec.json
```diff
@@ -0,0 +1,19 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "outDir": "../../../../dist/out-tsc",
+    "module": "commonjs",
+    "types": ["jest", "node"]
+  },
+  "include": [
```
saevarma, line 8: FYI the include array is generated as `"include": ["**/*.test.ts", "**/*.spec.ts", "**/*.d.ts"]` from the nest library generator. As `.tsx` should be redundant in gql api library.

libs/api/domains/watson-assistant-chat/src/lib/watson-assistant-chat.service.ts
```diff
@@ -0,0 +1,38 @@
...
+
+export class WatsonAssistantChatService {
+  private rsaKey!: RSA
+
+  constructor(private config: ConfigType<typeof WatsonAssistantChatConfig>) {
+    this.rsaKey = new RSA(this.config.directorateOfImmigrationPublicIBMKey)
+  }
+
+  async createIdentityToken(input: WatsonAssistantChatIdentityTokenInput) {
+    const encryptedUserPayload = this.rsaKey.encrypt(
```
saevarma, line 15: I'm wondering if this could just be done using NodeJS built in `crypto.publicEncrypt` instead of introducing a npm package. I haven't tried it but from short google it seems to be possible. I would be open to look into that with you if you want.

saevarma (commented):
Looks good, just minor comments and a thought about alternative solution instead of introducing a new npm package.

libs/api/domains/watson-assistant-chat/.babelrc
RunarVestmann: image

I've been generating api domains using this command :/

libs/api/domains/watson-assistant-chat/src/lib/watson-assistant-chat.service.ts
RunarVestmann: Here's the guide I followed: https://cloud.ibm.com/docs/assistant?topic=assistant-deploy-zendesk

image

So this has been tested using the node-rsa library and we know that works.

However we could try using the crypto package instead but how would I encrypt the user payload in the same way?

libs/api/domains/watson-assistant-chat/tsconfig.spec.json
RunarVestmann: I tweaked the tsconfig file since I generated the project as a node project

libs/api/domains/watson-assistant-chat/.babelrc
saevarma: Interesting, I have always use the `@nrwl/nest:library` but I guess there is no harm in the `.babelrc` file.

libs/api/domains/watson-assistant-chat/src/lib/watson-assistant-chat.service.ts
RunarVestmann: I'll try this out: 
image

saevarma (approved):
💯

ingolfurs (approved):

PR #10724, opened 2023-03-31
feat(island-ui): VisuallyHidden

hedinne:
## What

This PR adds a `<VisuallyHidden />` component to island-ui. 
VisuallyHidden is a common technique used in web accessibility to hide content from the visual client, but keep it readable for screen readers.

In some cases, using this technique can be better than using aria-label. For example:
- There can be variations in how screen readers announce aria-label.
- Some users use automatic systems to translate pages (e.g., Google Translate), and these systems don't always translate aria-label.

## Checklist:

### DIFF
--- a/libs/island-ui/core/src/lib/VisuallyHidden/VisuallyHidden.tsx
+++ b/libs/island-ui/core/src/lib/VisuallyHidden/VisuallyHidden.tsx
@@ -1,0 +1,33 @@
+import React, { CSSProperties, forwardRef, ReactNode } from 'react'
+
+const style: CSSProperties = {
+  border: 0,
+  clip: 'rect(0 0 0 0)',
+  height: '1px',
+  overflow: 'hidden',
+  padding: 0,
+  position: 'absolute',
+  whiteSpace: 'nowrap',
+  width: '1px',
+}
+
+interface VisuallyHiddenProps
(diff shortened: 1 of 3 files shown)


libs/island-ui/core/src/lib/VisuallyHidden/VisuallyHidden.tsx
```diff
@@ -0,0 +1,33 @@
...
+  extends Pick<
+    React.HTMLAttributes<HTMLElement>,
+    'aria-labelledby' | 'aria-describedby' | 'id'
+  > {
+  children?: ReactNode
+}
+
+const VisuallyHidden = forwardRef<HTMLSpanElement, VisuallyHiddenProps>(
+  function VisuallyHiddenInner({ children, ...restProps }, ref) {
+    return (
```
RunarVestmann, line 24: Why use inline styles rather than a className and vanilla extract?

hedinne: Both ways have their minor pros and cons, so I had to choose either way. I chose inline because this kind of component is usually written with inline styles, so that felt more natural.

RunarVestmann (approved):

PR #10725, opened 2023-04-01
fix(auth-admin): Remove filter dropdown for tenants and cleints

GunnlaugurG:
# ...

## What

Removing the filter button for clients and tenants

## Why

Because we will only allow search by name or id  for those lists

## Screenshots / Gifs

## Checklist:


libs/portals/admin/ids-admin/src/components/Clients/Clients.tsx
```diff
@@ -41,14 +42,13 @@ const Clients = () => {
 
   const handleSearch = (value: string) => {
     setInputSearchValue(value)
-
     if (value.length > 0) {
-      const filteredList = clients.filter((client: any) => {
+      const filteredList = originalClients.filter((client: any) => {
```
saevarma, line 46: Shouldn´t we implicitly have the type of the client ?

saevarma (approved):
LGTM just minor comments.

PR #10731, opened 2023-04-03
feat(j-s): Court of appeal overview screen

oddsson:
# Court of appeal overview screen

[Asana ](https://app.asana.com/0/1199153462262248/1204298511753938/f)

## What

**ISLAND-UI CHANGES**

Adds the ability to have a tooltip in AlertBanner 

Screenshot 2023-04-05 at 13 10 15

**JUDICIAL-SYSTEM CHANGES**
Added a overview screen for court of appeal. 

## Why

So that judges at COA can see an overview of appealed case.

## Screenshots / Gifs

image

## Checklist:


apps/judicial-system/web/src/routes/Shared/SignedVerdictOverview/SignedVerdictOverview.tsx
```diff
@@ -729,8 +728,7 @@ export const SignedVerdictOverview: React.FC = () => {
             workingCase.prosecutorAppealDecision ===
               CaseAppealDecision.APPEAL) &&
             user?.role &&
-            isCourtRole(user.role) &&
-            user?.institution?.type !== InstitutionType.HighCourt && (
```
oddsson, line 733: One thing about this change is that I'm removing all if statements regarding if a user is a high court (court of appeal) judge from SignedVerdictOverview page because they now won't be using this page but rather a new court of appeal overview page.

libs/island-ui/core/src/lib/AlertBanner/AlertBanner.tsx
```diff
@@ -143,6 +145,11 @@ export const AlertBanner: FC<AlertBannerProps> = ({
                 {description}
                 {description && link && ` `}
                 {link && <a href={link.href}>{link.title}</a>}
+                {link && link.tooltip && (
```
RunarVestmann, line 148: The code looks fine so I'll approve. However I'd like to know if this change to the island-ui alert banner will also be kept up to date in Figma? Could we maybe add this as a story in storybook to show that this is possible?

RunarVestmann (approved):

libs/island-ui/core/src/lib/AlertBanner/AlertBanner.tsx
oddsson: Thanks! I've now added a story for this and pinged the designers to notify them of this change and ask if they want to update the design.

apps/judicial-system/api/src/app/modules/case/interceptors/case.transformer.ts
```diff
@@ -14,10 +13,17 @@ export function transformCase(theCase: Case): Case {
       ? Date.now() > new Date(theCase.validToDate).getTime()
       : theCase.isValidToDateInThePast,
     isAppealDeadlineExpired: theCase.courtEndTime
-      ? Date.now() >= new Date(theCase.courtEndTime).getTime() + threeDays
+      ? Date.now() >= new Date(theCase.courtEndTime).getTime() + getDays(3)
```
unakb, line 16: Nice!

apps/judicial-system/web/src/components/Conclusion/Conclusion.tsx
```diff
@@ -0,0 +1,39 @@
...
+interface Props {
+  conclusionText?: string
+  judgeName?: string
+}
+
+const Conclusion: React.FC<Props> = (props) => {
+  const { formatMessage } = useIntl()
+  const { conclusionText, judgeName } = props
+
+  return !conclusionText ? null : (
```
unakb, line 18: I feel like this would be more clear if turned around to have the affirmative action first, so conclusionText ? <BlueBox...> : null

apps/judicial-system/web/src/routes/CourtOfAppeal/Overview/Overview.tsx
```diff
@@ -0,0 +1,277 @@
...
+import RulingDateLabel from '@island.is/judicial-system-web/src/components/RulingDateLabel/RulingDateLabel'
+import { capitalize, formatDate } from '@island.is/judicial-system/formatters'
+import Conclusion from '@island.is/judicial-system-web/src/components/Conclusion/Conclusion'
+import { CaseFileCategory } from '@island.is/judicial-system/types'
+import { useFileList } from '@island.is/judicial-system-web/src/utils/hooks'
+import * as constants from '@island.is/judicial-system/consts'
+
+import { courtOfAppealOverview as strings } from './Overview.strings'
+
+export const getStatementDeadline = (appealDate?: string) => {
```
unakb, line 32: I wonder if we could move this logic to the transformer as well? Since we're already checking there whether the deadline has expired, to return the actual deadline as well since we also need to use it in other places

unakb (approved):
Just some thoughts but no showstoppers :)

libs/island-ui/core/src/lib/AlertBanner/AlertBanner.tsx
oddsson: @RunarVestmann FYI, I've decided to revert these changes for now and work a bit more with the designers on this.

apps/judicial-system/web/src/routes/CourtOfAppeal/Overview/Overview.tsx
oddsson: Yes, I agree with you. I also think that appealDeadline should be there as well. I added a task to the inbox to fix this.

unakb: Agreed, I also want to put a few more things there like appealedDate, I'll add to the asana task

PR #10740, opened 2023-04-03
chore(contentful-extensions): Remove contentful parent-slug extension from monorepo

RunarVestmann:
# Remove contentful parent-slug extension from monorepo

## What

* The contentful parent slug extension will no longer be needed since we have replaced it's logic by have a contentful app that handles the previously created subarticles that rely on this behaviour as well as modifying the model for subarticles so their url's are up to date

## Checklist:


saevarma (approved):
LGTM

PR #10742, opened 2023-04-03
feat(auth-admin): Added advanced settings

GunnlaugurG:
# ...

Advanced settings for super admins for auth-admin

so super users can edit advanced settings 

## Screenshots / Gifs

## Checklist:


libs/portals/admin/ids-admin/src/components/forms/EditApplication/EditApplication.action.ts
```diff
@@ -34,6 +35,23 @@ const checkIfStringIsListOfUrls = (urls: string) => {
   return true
 }
 
+const checkIfStringIsArrayOfValidClaims = (claims: string) => {
+  if (claims.length === 0) {
```
saevarma, line 39: I would check it as, as it checks for undefined, null and empty string.
```suggestion
  if (!claims) {
```

libs/portals/admin/ids-admin/src/components/forms/EditApplication/EditApplication.action.ts
```diff
@@ -34,6 +35,23 @@ const checkIfStringIsListOfUrls = (urls: string) => {
...
+    return true
+  }
+
+  const array = splitStringOnCommaOrSpaceOrNewLine(claims)
+
+  const regex = /^\w+=\w+$|^\w+=\d+$/
```
saevarma, line 45: As `/w` is Alphanumeric characters it also contains `/d` so this regex can be simplified to 
```suggestion
  const regex = /^\w+=\w+$/
```

libs/portals/admin/ids-admin/src/components/forms/EditApplication/EditApplication.action.ts
```diff
@@ -34,6 +35,23 @@ const checkIfStringIsListOfUrls = (urls: string) => {
   return true
 }
 
+const checkIfStringIsArrayOfValidClaims = (claims: string) => {
```
saevarma, line 38: ~Have test for this function

saevarma (approved):

PR #10761, opened 2023-04-04
fix: slight tuning to the cpu request values to reduce scaling pressu…

brynjarorng:
…re (#10755)

* fix: slight tuning to the cpu request values to reduce scaling pressure

* raised the target by one again since this had an effect on prod but could likely be one higher, otherwise we might need to look at setting this for individual apps

---------

# ...

## What

## Why

## Screenshots / Gifs

## Checklist:


busla (approved):
LGTM

PR #10764, opened 2023-04-04
fix(j-s): Only display alert banner if case appeal has been postponed

unakb:
# Only display alert banner if case appeal has been postponed

[Landsréttur - gulur borði kærufrestur á ekki að birtast nema annar aðili taki sér frest. ](https://app.asana.com/0/1199153462262248/1204298511753923/f)

## What

Hide alert banner if CaseAppealDecision is not POSTPONED

## Why

So that we don't display the appeal banner if the case cannot be appealed

## Checklist:


apps/judicial-system/web/src/routes/Shared/SignedVerdictOverview/SignedVerdictOverview.tsx
```diff
@@ -508,6 +508,8 @@ export const SignedVerdictOverview: React.FC = () => {
 
     const shouldDisplayAppealAlertBanner =
       workingCase.courtEndTime &&
+      (workingCase.accusedAppealDecision === CaseAppealDecision.POSTPONE ||
```
oddsson, line 511: Please use the `isAppealedCase` variable defined on line 282

oddsson (commented):
One comment about an unused variable

apps/judicial-system/web/src/components/AppealAlertBanner/AppealAlertBanner.strings.ts
```diff
@@ -0,0 +1,30 @@
+import { defineMessages } from 'react-intl'
+
+export const strings = defineMessages({
+  appealTitle: {
+    id: 'judicial.system.core:appeal_alert_banner.appeal_title',
+    defaultMessage: 'Kærufrestur rennur út {appealDeadline}',
```
oddsson, line 6: Since you can appeal after the appeal deadline has passed, I think we need to be able to say 

Kærufrestur rann út {appealDeadline}

oddsson (approved):

apps/judicial-system/web/src/components/AppealAlertBanner/AppealAlertBanner.strings.ts
unakb: Discussed offline - this will be handled in separate task

PR #10766, opened 2023-04-05
feat(j-s): Add a isStatementDeadlineExpired field to our API

oddsson:
# Add a isStatementDeadlineExpired field to our API

[Asana](https://app.asana.com/0/1199153462262248/1204298511753938/f)

## What

Add a new field, `isStatementDeadlineExpired`, to our API. This deadline represents whether or not the deadline to hand in the statement document in appealed cases (frestur til að skila inn greinargerð) has passed. This deadline is 24 hours after a case is appealed.

## Checklist:


apps/judicial-system/api/src/app/modules/case/interceptors/case.transformer.spec.ts
```diff
@@ -176,4 +178,55 @@ describe('transformCase', () => {
...
+
+      // Act
+      const res = transformCase(theCase)
+
+      // Assert
+      expect(res.isStatementDeadlineExpired).toBe(false)
+    })
+
+    it('should be false when more than one day has passed since the case was appealed', () => {
+      // Arrange
```
unakb, line 195: Shouldn't it be false "when lass than one day has passed"? and then in line 214 "when more than one day"

unakb (changes requested):
I just have a concern because I'm not sure if I'm misunderstanding the statement deadline or if the logic here might be wrong. I thought that the statement deadline started ticking down when the judge has marked the case as received? Not when the case is appealed

unakb (approved):
There's one linting warning in the transformer and then the failing tests, but otherwise looks good

apps/judicial-system/api/src/app/modules/case/interceptors/case.transformer.spec.ts
oddsson: Yes! Thanks

PR #10771, opened 2023-04-05
feat(sessions): Changes to display country but not region

GunnlaugurG:
# ...

## What

Display the country in sessions instead of region

## Why

The region was not accurate enough

## Screenshots / Gifs

## Checklist:


saevarma (approved):
LGTM, we should have maybe done the country lookup in the gql with a field resolver.
Will add it to our board as nice to have task, but lets merge this and test.

PR #10772, opened 2023-04-05
feat(web): Add price content type

RunarVestmann:
# Add price content type

## What

* We need a way to have prices at a single place in the CMS since as of right now multiple articles may talk about the price for the same thing but if for example that price would change then we'd have to remember every single article that references that price and update it. This change introduces a new content type which can be inline embedded inside of rich text. This way users in the CMS will be encouraged to create a price entry and inline embed it instead if the previously mentioned method.

## Checklist:

### DIFF
--- a/libs/island-ui/contentful/src/lib/RichTextRC/defaultRenderNode.tsx
+++ b/libs/island-ui/contentful/src/lib/RichTextRC/defaultRenderNode.tsx
@@ -169,6 +169,27 @@ export const defaultRenderNodeObject: RenderNode = {
       </Box>
     )
   },
+  [INLINES.EMBEDDED_ENTRY]: (node) => {
+    const amount = node?.data?.target?.fields?.amount
+    if (typeof amount !== 'number') return null
+
+    let postfix = 'krónur'
+
+    const amountString = String(amount)
+
+    if (amountString.endsWith('1') && !amountString.endsWith('11')) {
+      postfix = 'króna'
+    }
+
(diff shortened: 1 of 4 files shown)


libs/island-ui/contentful/src/lib/RichTextRC/defaultRenderNode.tsx
```diff
@@ -169,6 +169,27 @@ export const defaultRenderNodeObject: RenderNode = {
...
+    if (amountString.endsWith('1') && !amountString.endsWith('11')) {
+      postfix = 'króna'
+    }
+
+    // Format the amount so it displays dots (Example of a displayed value: 2.700 krónur)
+    const formatter = new Intl.NumberFormat('de-DE')
+
+    return (
+      <span>
+        {formatter.format(amount)} {postfix}{' '}
```
oddsson, line 189: Consider using a template string for this. I think that's clearer than having the space before and after the postfix
```suggestion
        {`${formatter.format(amount)} ${postfix} `}
```

oddsson (commented):
One minor comment

libs/island-ui/contentful/src/lib/RichTextRC/defaultRenderNode.tsx
RunarVestmann: I've removed the space after the posfix since we might want to have a comma or period right after "krónur" so we need to be able to freely choose what the next character is via the CMS rich text content.

oddsson (approved):

siggifv (approved):

PR #10781, opened 2023-04-11
J s/longer appeal grace period

oddsson:
# ...

## What

## Why

## Screenshots / Gifs

## Checklist:

### DIFF
--- a/apps/judicial-system/api/src/app/modules/case/interceptors/case.transformer.spec.ts
+++ b/apps/judicial-system/api/src/app/modules/case/interceptors/case.transformer.spec.ts
@@ -152,7 +152,7 @@ describe('transformCase', () => {
     it('should be false while the appeal window is open', () => {
       // Arrange
       const courtEndTime = new Date()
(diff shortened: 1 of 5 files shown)


libs/judicial-system/types/src/lib/case.ts
```diff
@@ -459,3 +467,68 @@ export function hasCaseBeenAppealed(theCase: Case): boolean {
       Boolean(theCase.prosecutorPostponedAppealDate))
   )
 }
+
+export function getAppealInfo(theCase: Case): Case {
```
gudjong, line 471: Consider moving this logic to the the transformer in the api project as it is only used there.

gudjong (approved):
One consideration

libs/judicial-system/types/src/lib/case.ts
unakb: Discussed offline, we'll keep it here for now as I'm using part of the logic in my next PR in the backend, and then move later if needed.

PR #10784, opened 2023-04-11
feat(island-ui): Hide spin buttons and add inputmode

hedinne:
## What

This PR does two things to increase the UX of the <Input /> component. 
1. Hides the spin buttons on input[type='number']. They are never used.
2. Add support for the `inputmode` ([MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inputmode)) on inputs. This allows us to control what kind of keyboard is displayed on mobile devices.

## Screenshots / Gifs

Before removing the spin buttons:
Screen Shot 2023-04-11 at 14 39 39

After:
Screen Shot 2023-04-11 at 14 40 31

With `inputMode="numeric"`:
Screenshot 2023-04-11 at 14 47 02

Without:

Screenshot 2023-04-11 at 14 47 20

## Checklist:

### DIFF
--- a/libs/island-ui/core/src/lib/Input/types.ts
+++ b/libs/island-ui/core/src/lib/Input/types.ts
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { HTMLAttributes } from 'react'
 
 import * as styles from './Input.css'
 import { Icon as IconType, Type } from '../IconRC/iconMap'
@@ -52,7 +52,7 @@ export interface InputComponentProps {
     on: boolean
     maxHeight?: number
   }
-  inputMode?: string
+  inputMode?: HTMLAttributes<HTMLInputElement>['inputMode']
 }
 
 export type InputButton = {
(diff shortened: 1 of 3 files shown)


libs/island-ui/core/src/lib/Input/types.ts
```diff
@@ -52,6 +52,7 @@ export interface InputComponentProps {
     on: boolean
     maxHeight?: number
   }
+  inputMode?: string
```
oddsson, line 55: Just a thought, do we want to define a list of possible values for this (i.e. inputMode: "numeric" | "decimal" | ...) or is that not a good idea because we'd have to update our code if this attribute ever changes 🤔 

What do you think?

oddsson (commented):
One question

libs/island-ui/core/src/lib/Input/types.ts
hedinne: That would be better for the consumer of the component, but I always try to mirror the HTML element typings, so I put just a `string`. But I just looked back at it and it turns out that I was reading the `interface HTMLInputElement {}` from Typescript and not `interface HTMLAttributes {}` from React. The element uses the React typings and they have the values specified. 
So I'm going to change it. It was a good thing that you flagged this.

oddsson (approved):

PR #10785, opened 2023-04-11
feat(portals-admin): Fix clients loading state and button loading state

snaerseljan:
# Fixes for ids admin

## What

Fix create client loading state. I added a special hook to handle loading state in RR form submissions.
Fixed refetching clients when navigating back.
Fix unused ref which was causing warning in UI.

## Checklist:


saevarma (approved):
💯

libs/island-ui/core/src/lib/AsyncSearch/shared/Label/Label.tsx
```diff
@@ -9,10 +9,11 @@ export interface LableProps extends LabelHTMLAttributes<HTMLLabelElement> {
 }
 
 export const Label = forwardRef<HTMLLabelElement, LableProps>(
-  ({ hasError, color, children, ...props }) => {
```
oddsson, line 12: Is this `ref` never used or am I missing something 🤔

oddsson (commented):
One question

libs/island-ui/core/src/lib/AsyncSearch/shared/Label/Label.tsx
snaerseljan: I image this is never used but the forwardRef was there initially without a ref. I could also just remove the HOC but this felt a little safer.

snaerseljan: @oddsson can you re review or approve if you are satisfied :)

oddsson (approved):

PR #10793, opened 2023-04-12
feat(j-s): Handle all alert banners for appeal cases

unakb:
# Handle all alert banners for appeal cases

[Asana](https://app.asana.com/0/1199153462262248/1204127104755276)

## What

Handle all the different alert banners for appeal cases. There are a few things that are still TODOs because we haven't implemented the necessary logic in the backend yet. I also intend to move most of the logic that calculates the appeal information into a transformer in the api so the information gets returned automatically when we retrieve the case. 

## Why

To confine the complicated display logic to one component instead of spreading it out across the components that need to display the banners

## Checklist:


apps/judicial-system/api/src/app/modules/case/interceptors/case.transformer.spec.ts
```diff
@@ -226,4 +228,86 @@ describe('transformCase', () => {
...
+      // Assert
+
+      expect(res.appealDeadline).toBeUndefined()
+      expect(res.appealedByRole).toBeUndefined()
+      expect(res.appealedDate).toBeUndefined()
+      expect(res.hasBeenAppealed).toBe(false)
+      expect(res.canBeAppealed).toBe(false)
+    })
+
+    it('should have appeal deadline and hasBeenAppealed set to false', () => {
```
oddsson, line 249: Please consider adding a more specific description here, i.e. "should have appeal deadline and hasBeenAppealed set to false **when something** "

apps/judicial-system/api/src/app/modules/case/interceptors/case.transformer.spec.ts
```diff
@@ -226,4 +228,86 @@ describe('transformCase', () => {
...
+
+      // Act
+      const res = transformCase(theCase)
+
+      // Assert
+      expect(res.appealDeadline).toBeDefined()
+      expect(res.hasBeenAppealed).toBe(false)
+    })
+
+    it('should have hasBeenAppealed set to true and appealed date set', () => {
```
oddsson, line 264: Same comment about a more clear description

apps/judicial-system/web/src/components/AppealAlertBanner/AppealAlertBanner.tsx
```diff
@@ -28,79 +27,61 @@ interface Props {
...
     canBeAppealed,
-    hasBeenAppealed,
     appealDeadline,
-  } = getAppealInfo(workingCase)
+  } = workingCase
 
   let alertTitle, alertLinkTitle, alertLinkHref, alertDescription
 
-  if (hasBeenAppealed) {
+  const isCourtRoleUser = isCourtRole(user?.role)
```
unakb, line 49: I was a little unsure about this, should this maybe only be the judge and not a court role user?

apps/judicial-system/api/src/app/modules/case/interceptors/case.transformer.spec.ts
unakb: Donezo!

apps/judicial-system/web/src/components/AppealAlertBanner/AppealAlertBanner.tsx
oddsson: I think this is correct. I'll accept it and you can ask Gudny about it and change it if it's not.

oddsson (approved):

apps/judicial-system/web/src/components/AppealAlertBanner/AppealAlertBanner.tsx
unakb: Confirmed 😊

oddsson:
Closing this in favour of #10781

PR #10814, opened 2023-04-13
feat(auth-admin): Auth admin client patch

GunnlaugurG:
# ...

## What

* Rest endpoint to patch client
* Graphql layer to patch client
* Connected frontend to the graphql layer

## Why

So clients can be edited by users

## Screenshots / Gifs

## Checklist:


libs/portals/admin/ids-admin/src/components/Client/AdvancedSettings.tsx
```diff
@@ -85,53 +75,81 @@ const AdvancedSettings = ({
...
-          value={`${consent}`}
+          defaultChecked={inputValues.requireConsent}
+          checked={inputValues.requireConsent}
+          onChange={(e) => {
+            setInputValues({
+              ...inputValues,
+              requireConsent: e.target.checked,
+            })
+          }}
+          value={'true'}
```
saevarma, line 86: ~When the value is just hardcoded string we can drop the brackets.
```suggestion
          value="true"
```

libs/portals/admin/ids-admin/src/components/Client/Translations.tsx
```diff
@@ -39,10 +35,11 @@ const Translations = ({ translations }: TranslationsProps) => {
   return (
     <ContentCard
       title={formatMessage(m.translations)}
-      onSave={(saveOnAllEnvironments) => {
-        console.log('saveOnAllEnvironments: ', saveOnAllEnvironments)
+      onSave={() => {
```
saevarma, line 38: Looks like the `onSave` function is redundant, and in the other components. What is the purpose of the function other than control rendering of the 'Save in all environments' checkbox?

saevarma (approved):
Nice job 💯
Just minor comments.

libs/portals/admin/ids-admin/src/components/Client/Translations.tsx
GunnlaugurG: I added it before I added the loader pattern to the request, I forgot to remove it once the loader pattern was implemented :D

PR #10821, opened 2023-04-14
feat(contentful-apps): Move public env variables into constants

RunarVestmann:
# Move public env variables into constants

## What

* It seems that the env variables aren't being exposed in the browser so here's an attempt to fix that
* I move the env variables to constants since there wasn't a need to have env variables to begin with
* The secret used to call the mideind translation web service now lives inside Contentful via instance parameter

## Checklist:


brynjarorng (approved):
LGTM(Devops)

siggifv (approved):

PR #10835, opened 2023-04-17
feat(j-s): Appeal Counter Party Email

gudjong:
# Appeal Counter Party Email

[Landsréttur - senda kærutilkynningu þegar verjandi kærir](https://app.asana.com/0/1199153462262248/1204367224389681/f)

## What

- Sends full user details with queued messages to support defender initiated messages.
- Sends appeal email to the counter party (defender or prosecutor) when a case is appealed.

## Why

- To keep the counter party informed of the appeal.

## Checklist:


apps/judicial-system/backend/src/app/modules/case/test/internalCaseController/deliverCaseToPolice.spec.ts
```diff
@@ -41,7 +41,7 @@ describe('InternalCaseController - Deliver case to police', () => {
       const then = {} as Then
 
       await internalCaseController
-        .deliverCaseToPolice(caseId, user, theCase, { userId })
+        .deliverCaseToPolice(caseId, theCase, { user })
```
oddsson, line 44: Does the user here need to be wrapped in `{}`? It's already an object, am I misunderstanding something?

apps/judicial-system/backend/src/app/modules/case/test/internalCaseController/deliverCourtRecordToCourt.spec.ts
```diff
@@ -44,7 +45,7 @@ describe('InternalCaseController - Deliver court record to court', () => {
       const then = {} as Then
 
       await internalCaseController
-        .deliverCourtRecordToCourt(caseId, user, theCase, { userId })
+        .deliverCourtRecordToCourt(caseId, theCase, { user })
```
oddsson, line 48: Same comment as ☝️

apps/judicial-system/backend/src/app/modules/case/test/internalCaseController/deliverCaseToPolice.spec.ts
gudjong: The deliver DTO contains one field, user, which happens to be an object, so this is correct. We could let the DTO list all the fields of the user object but that would be a bit strange and would require us to change the DTO every time we change the User type.

apps/judicial-system/backend/src/app/modules/case/test/internalCaseController/deliverCourtRecordToCourt.spec.ts
gudjong: Same response.

oddsson (approved):

PR #10838, opened 2023-04-17
feat(passports-client): add delivery address api

alexdiljar:
Updating clientConfig for passports to add deliveryAdressApi. 
Locations now provided from Þjóskrá. 

## Checklist:


thordurhhh (approved):

libs/application/template-api-modules/src/lib/modules/templates/passport/passport.service.ts
```diff
@@ -37,6 +37,20 @@ export class PassportService extends BaseTemplateApiService {
...
 
+  async deliveryAddress({ auth }: TemplateApiModuleActionProps) {
+    const res = await this.passportApi.getDeliveryAddress(auth)
+    if (!res) {
+      throw new TemplateApiError(
+        {
+          title: coreErrorMessages.failedDataProvider,
+          summary: coreErrorMessages.errorDataProvider,
+        },
+        400,
```
kksteini, line 48: Would this be easier to trace with `res.status` if that's available to us, instead of putting down 400 explicitly?

libs/application/templates/passport/src/forms/Draft.ts
```diff
@@ -189,14 +189,13 @@ export const Draft: Form = buildForm({
...
                 },
               }) => {
                 return (data as DistrictCommissionerAgencies[])?.map(
-                  ({ id, name, place, address }) => ({
-                    value: id,
-                    label: `${name}, ${place}`,
-                    tooltip: `${address}`,
+                  ({ key, name }) => ({
+                    value: `${key}`,
+                    label: `${name}`,
```
kksteini, line 198:
```suggestion
                    value: key,
                    label: name,
```
Is the template string needed?

kksteini (approved):

PR #10841, opened 2023-04-18
fix(island-ui): Update path to shared icon types

hedinne:
## What

In #10114, the types for the IconRC were moved from the component file to a separate file. However, all individual Icons referenced one of those interfaces, and the import statements weren't updated. This fixes those imports.
Also, the interface for a sub-component, Placeholder, was wrongly typed and, therefore, had to use non-null assertion. I fixed that also.

## Why

Because I don't like having Typescript errors in my editor.

## Checklist:


Toti91 (approved):
Nice

PR #10846, opened 2023-04-18
feat(j-s): Case list for court of appeal

oddsson:
# Case list for court of appeal

[Asana](https://app.asana.com/0/1199153462262248/1204127104755188/f)

## What

Implement a case list for court of appeal users. In this PR we

- Route COA users to a new route `/landsrettur/krofur` after login
- Add `appealState` to the cases endpoint
- Move all common table strings to a shared message document
- Move open case logic to useCase hook as well as using a new query for court of appeal cases.

## Why

The Cases screen has up until this point been shared across all users. Instead of adding to it for COA uses, we now create a new screen with custom logic and query. This is a much cleaner approach IMO.

## Checklist:


unakb (approved):
We have all contributed to this PR in some form at this point so we're going to accept it so we can continue with our other changes that depend on this branch

PR #10884, opened 2023-04-21
ci: run prod builds on PR label

busla:
# Build apps using NX native parallel processing

See #11220

## What
This PR uses NX native parallel processing and builds all apps with a single runner.

## Why
Current build flow uses `nx print-affected` to generate a list of affected apps, chunks that list down and distributes the chunks over _n_ number of runners with a max job of `3`. Since NX is designed to build apps in a parallel fashion, using a single runner could be used to build all applications, given that the runner is provisioned with enough CPU resources.

The PR also uses NX Cloud to compute and return cached artifacts. Using S3 for compressing and storing cached artifacts is therefore not needed.

## Checklist:

---
  - https://app.asana.com/0/0/1205215313716674


peturgq:
Issue: https://github.com/island-is/island.is/issues/11220

.github/workflows/pullrequest-nx-cloud.yml
```diff
@@ -0,0 +1,176 @@
...
+    runs-on: ec2-runners
+    container:
+      image: public.ecr.aws/m3u4c4h9/island-is/actions-runner-public:latest
+    timeout-minutes: 30
+    strategy:
+      fail-fast: false
+      matrix:
+        nx-target:
+          - build
+          - lint
```
eirikurn, line 47: Missing test?

.github/workflows/pullrequest-nx-cloud.yml
```diff
@@ -0,0 +1,176 @@
...
+  run:
+    shell: bash
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+# concurrency:
+#   group: '${{ github.workflow }} @ ${{ github.event.pull_request.head.label || github.head_ref || github.ref }}'
+#   cancel-in-progress: true
```
eirikurn, line 21: Commented code?

.github/workflows/pullrequest-nx-cloud.yml
```diff
@@ -0,0 +1,176 @@
...
+    steps:
+      - uses: actions/download-artifact@v3
+      - run: |
+          echo "Check matrix"
+          matrix="$(cat */matrix | jq -c --slurp .)"
+          echo "matrix: $matrix"
+          echo "::set-output name=matrix::$matrix"
+        id: matrix
+
+  # TODO: trying to dynamically get the the matrix result output, unfinished
```
eirikurn, line 159: Would be good to get this working.

eirikurn (commented):
Exciting!

Should we still have a prepare step to install dependencies and generate schemas before starting nx in multiple agents?

I also wonder if 3 beefy agents * 30 parallel NX jobs is going to be slower (or crash) compared to more agents and lower NX parallelization. CPU is one thing, but many of our commands need a lot of heap memory (8GB) which might run faster on more agents. 🤷‍♂️

Finally, we should upgrade to [Distributed Task Execution](https://nx.dev/core-features/distribute-task-execution) immediately after we finish upgrading to the latest NX and NX Cloud 2. Should be able to schedule NX jobs in the "prepare" step and then run 3+ parallel agents to consume those jobs using NX Cloud.

busla:
- yes, I am temporarily disabling all PR jobs while I am working on this

- What I would like to do is to dynamically provision runners based on affected projects. We could set a tag (or another attribute) in project.json per project where we could set cpu/mem (or just something that describes the project intensity) and run `print-affected` to gather affected projects (and their intensity) in the prepare job. From there we could provision correctly sized gh runners to run the distributed agents. Most of the time we are building only a few affected projects, but during full builds, more runners would be provisioned. I need to look at if we can assign project targets to particular runners.

eirikurn (approved):
Approving core files. Please don't forget to remove deleteme.ts files.

Didn't have a chance to look deeper into change. Hope it works great :D

busla:
ughh, you are getting review requests because of deleteme files 🤦

not done yet

eirikurn:
Ah. Can you not change it back into a draft PR?

busla:
let me check again, had workflow restrictions when it was a draft, thats why I changed it

PR #10888, opened 2023-04-24
chore(service-portal): Update translation loading for modules

thordurhhh:
## What

* Update translation loading for index, vehicles and license

## Why

* The translation strings will show a flash of the default translation string before replacing it with the one from CMS.
* Adding a loader to modules where there is a mismatch.

## Checklist:


thorkellmani (approved):
lgtm

Issue #10903, opened 2023-04-24
allow app db user to install postgres extensions in DSL


busla:
Postgres extensions need to be enabled per database which we can do in terraform but this does not apply to feature-deployment.

We need to enhance the DSL to allow developers to set extensions that should be enabled for feature deployments.

busla:
Since we are currently running postgres `11.17` users with create create permissions on a db cannot install extensions.

> For RDS for PostgreSQL version 13 and higher versions, users (roles) with create permissions on a given database instance can install and use any trusted extensions. For a list of trusted extensions, see [PostgreSQL trusted extensions](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/CHAP_PostgreSQL.html#PostgreSQL.Concepts.General.Extensions.Trusted).

https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Appendix.PostgreSQL.CommonDBATasks.Extensions.html

busla:
Waiting for us to upgrade to postgres v13.x or higher

busla:
related:
https://github.com/island-is/island.is/issues/10903
https://github.com/island-is/island.is/pull/10912

busla:
this has been implemented for feature deployment since Spinnaker uses postgres master-user to create feature-deploy dbs

see:
https://github.com/island-is/island.is/pull/10904

PR #10904, opened 2023-04-24
feat(dsl): added extensions field to PostgresInfoForEnv to allow setting pg extensions

busla:
# create postgres extensions with DSL

https://github.com/island-is/island.is/issues/10903

## What
Grant developers to create postgres extensions for their databases using the DSL.

This is partially done (feature deployments). We need to upgrade to postgres 13.x for feature completion, see attached issue.  

## Why
Extensions need to be enabled per database but could be part of the self-service platform where db users with create perms could enable _trusted extensions_.

## Checklist:


brynjarorng (approved):
LGTM

apps/services/sessions/infra/sessions.ts
```diff
@@ -16,6 +16,7 @@ const workerPostgresInfo = {
   username: 'services_sessions',
   name: dbName,
   passwordSecret: '/k8s/services-sessions/DB_PASSWORD',
+  extensions: ['uuid-ossp'],
```
saevarma, line 19: Nice!
We should also add this to `apps/services/auth/ids-api/infra`

busla: > Nice! We should also add this to `apps/services/auth/ids-api/infra`

done

busla:
> Deployed services: .
> Excluded services: ``

why isn't session and ids-api affect?

PR #10912, opened 2023-04-25
Dsl postgres extensions

busla:
# ...

## What

## Why

## Screenshots / Gifs

## Checklist:


busla:
related https://github.com/island-is/island.is/pull/10912

infra/src/dsl/dsl.ts
```diff
@@ -171,7 +172,10 @@ export class ServiceBuilder<ServiceType> {
    */
   initContainer(ic: Optional<InitContainers, 'envs' | 'secrets' | 'features'>) {
     if (ic.postgres) {
-      ic.postgres = this.withDefaults(ic.postgres)
+      ic.postgres = {
+        ...this.withDefaults(ic.postgres),
+        extensions: ic.postgres.extensions,
```
peturgq, line 177: Not sure this is required here.

apps/services/auth/ids-api/src/main.ts
```diff
@@ -11,3 +11,5 @@ bootstrap({
   port: environment.port,
   enableVersioning: true,
 })
+
+// cache-bust-01
```
peturgq, line 15: What is this?

infra/src/dsl/dsl.ts
peturgq, line 177: I feel like we shouldn't need to specify extenstions specifically when making the object, as it is set in withDefaults?

peturgq: Removing this breaks the tests though, so I'm confused.

PR #10935, opened 2023-04-26
fix(service-portal): Modal sticky footer

hedinne:
# ...

## What

Set a maximum height on modals in delecation to match the viewport height.

## Why

So that the sticky footer is always visible.

## Screenshots / Gifs

[Kapture 2023-04-26 at 13.52.15.webm](https://user-images.githubusercontent.com/7573694/234602569-fed00d0f-275c-45a6-a490-b7b1fdcb699e.webm)

## Checklist:


libs/portals/shared-modules/delegations/src/components/Modal/Modal.css.ts
```diff
@@ -21,6 +21,8 @@ export const modal = style({
 })
 
 export const modalInner = style({
+  maxHeight: '100vh',
```
saevarma, line 24: We should also add the shadow above the footer if there is content to scroll.

hedinne: Done

saevarma (approved):
💯

PR #10936, opened 2023-04-26
feat(admin-api): Add scopes option to update client

snaerseljan:
# Added two fields to patch client call, aka update client. 

## What

Now you can add addedScopes and removedScopes fields which  both are array of scopeNames

## Why

This is a preparation for permissions updates in ids-admin clients

## Checklist:


libs/auth-api-lib/src/lib/clients/admin/dto/admin-client.dto.ts
```diff
@@ -101,4 +102,9 @@ export class AdminClientDto {
     type: [AdminClientClaimDto],
   })
   customClaims?: AdminClientClaimDto[]
+
+  @ApiProperty({
+    description: 'Allowed scopes for the client',
+  })
+  allowedScopes?: Pick<ClientAllowedScope, 'scopeName' | 'clientId'>[]
```
saevarma, line 109: This unfortunately doesn't generate the correct openapi schema, which won't generate the correct signature on the client function in the GQL layer.
I see that we have already `ClientAllowedScopeDTO` that we can just reuse here:

```suggestion
  @ApiProperty({
    description: 'Allowed scopes for the client',
    type: [ClientAllowedScopeDTO]
  })
  allowedScopes?: ClientAllowedScopeDTO[]
```

saevarma (approved):
Nice job 💯 just nice to haves and one fix that wee need to do. I removed the `automerge` as you need to address the type of the `allowedScope` for the client generation based on the openapi.

libs/auth-api-lib/src/lib/clients/admin/dto/admin-client.dto.ts
snaerseljan: Are you talking about the Pick or that I have to add the type?

PR #10970, opened 2023-04-28
fix(portals-admin-ids-admin): Modal body scroll

hedinne:
## What
This modal is a copy (or the original) of the modal In #10965. This fixes the same bug, where the body was allowed to scroll underneath the modal on mobile.

I've tested all the components that depend on this modal.

## Checklist:


saevarma (approved):
Assuming that preventBodyScroll is default true.

PR #10984, opened 2023-05-02
feat(web): Resolve powerbi embed props at request time

RunarVestmann:
# Resolve powerbi embed props at request time

## What

* Instead of having an open endpoint to fetch a powerbi embed token we'd rather like to resolve the embed token at request time
* I also added a way for powerbi reports to load in mobile portrait layout

## Why

* This is more "secure" since now only reports that should be embedded can actually generate a token without the user seeing any network request in his browser

## Checklist:


libs/cms/src/lib/models/powerBiSlice.model.ts
```diff
@@ -22,6 +22,9 @@ export class PowerBiSlice {
 
   @Field({ nullable: true })
   owner?: 'Fiskistofa'
+
+  @Field(() => graphqlTypeJson, { nullable: true })
+  powerBiEmbedPropsFromServer?: Record<string, unknown> | null
```
eirikurn, line 27: It's actually not neccessary to define this here, if you don't need it (then you can skip it also in mapPowerBiSlice). The field resolver decorators adds it to the final graphql schema.

libs/cms/src/lib/cms.resolver.ts
```diff
@@ -609,3 +612,20 @@ export class FeaturedSupportQNAsResolver {
     )
   }
 }
+
+@Resolver(() => PowerBiSlice)
+@Directive(cacheControlDirective())
+export class PowerBiSliceResolver {
+  constructor(private powerBiService: PowerBiService) {}
+
+  @ResolveField(() => graphqlTypeJson)
```
eirikurn, line 621: Field name can be simpler like "embedProps" and have a proper GraphQL type instead of using graphqlTypeJson. I.e. two fields: accessToken and embedUrl.

libs/cms/src/lib/cms.module.ts
```diff
@@ -7,15 +7,24 @@ import {
...
+import { ConfigModule } from '@island.is/nest/config'
+import { PowerBiConfig } from './powerbi.config'
 
 @Module({
-  imports: [TerminusModule],
+  imports: [
+    TerminusModule,
+    ConfigModule.forRoot({
+      load: [PowerBiConfig],
+    }),
```
eirikurn, line 26: This should generally only be done in the app module, listing all config objects. But here you might have an issue since this is a large module which is imported in some other apps like the application system. Weird to configure powerbi there.

I can suggest two solutions:

1. Move the PowerBi config and resolver back to a powerbi domain, but don't expose any root level queries, just field resolvers like PowerBiSlice resolver. This domain/module is not imported in the application system.
2. Use `imports: [PowerBiConfig.registerOptional()],` here while also loading it in app.module.ts. Same pattern as [here](https://github.com/island-is/island.is/blob/9e53f8c204878c3a625e786ecfe33c35bff0470d/libs/clients/national-registry/v2/src/lib/nationalRegistryClient.module.ts#L10). It will be validated in api (because it's loaded in app.module.ts), but in application system, the config will have `config.isConfigured === false` and throw runtime exceptions if you try to read anything from it (be careful not to read anything from it at startup if `config.isConfigured === false`).

I kind of prefer #1 😅

eirikurn (changes requested):
Like this change. Though it may break application system.

libs/cms/src/lib/cms.module.ts
RunarVestmann: I'm struggling a lot with option #1 :/ 

The PowerBiSlice model is needed for the field resolver so I think I need to define that in the cms lib due to not having circular dependencies between the api domain and the cms (since the PowerBiSlice model is needed for the fields resolver)

RunarVestmann: There is no need for the api domain so I'm hoping that what I've done most recently can work out 🙃

libs/cms/src/lib/powerbi.service.ts
```diff
@@ -95,3 +131,12 @@ export class PowerBiService {
...
 }
+
+export const PowerBiServiceProvider: Provider<PowerBiService> = {
+  provide: PowerBiService,
+  scope: LazyDuringDevScope,
+  useFactory(config: ConfigType<typeof PowerBiConfig>) {
+    return new PowerBiService(config)
+  },
+  inject: [PowerBiConfig.KEY],
+}
```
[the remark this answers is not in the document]
RunarVestmann: If I don't do it this way then (at least locally) I get a warning regarding PowerBiService dependencies not being found

libs/cms/src/lib/cms.resolver.ts
RunarVestmann: Agreed

libs/cms/src/lib/powerbi.service.ts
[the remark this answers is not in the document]
eirikurn: Did you configure the dependency correctly? This should work:

```
constructor(
    @Inject(PowerBiConfig.KEY)
    private config: ConfigType<typeof PowerBiConfig>,
) {}
```

apps/api/src/app/app.module.ts
```diff
@@ -315,7 +311,6 @@ const autoSchemaFile = environment.production
         PassportsClientConfig,
         FileStorageConfig,
         FiskistofaClientConfig,
-        PowerBiConfig,
```
eirikurn, line 318: Let's continue loading the PowerBiConfig here to make sure it's valid when the api starts. (Otherwise it can be silently ignored because of registerOptional)

eirikurn (approved):
LGTM

eirikurn (approved):
LGTM

apps/api/src/app/app.module.ts
RunarVestmann: Done and done (regarding the first point, no I must've been too tunnel visioned and forgotten the write the @Inject(PowerBiConfig.KEY) inside the constructor param 😅

siggifv (approved):

PR #10990, opened 2023-05-02
fix(j-s): Fixing choice of words in police case files 

MargretFinnboga:
# Fixing choice of words

[Breyta kröfu í ákæru](https://app.asana.com/0/1199153462262248/1204389017250394)
## What

- changing Krafa to ákæra 

## Why

- to have the correct wording 

## Screenshots / Gifs

## Checklist:


apps/judicial-system/web/src/routes/Prosecutor/components/PoliceCaseFiles/PoliceCaseFiles.strings.ts
```diff
@@ -56,10 +56,10 @@ export const policeCaseFiles = defineMessages({
       'Notaður sem texti í "hlaða upp" takka í LÖKE gagnapakkanum á rannsóknargagna skrefi.',
   },
   originNotLokeTitle: {
-    id: 'judicial.system.core:police_case_files.origin_not_loke_title',
-    defaultMessage: 'Krafa ekki stofnuð í LÖKE',
+    id: 'judicial.system.core:police_case_files.origin_not_loke_title_v1',
+    defaultMessage: 'Ákæra ekki stofnuð í LÖKE',
```
gudjong, line 60: The string is used in the `PoliceCaseFiles` component which is used for all case types. This means that we need two different strings and conditionally used one or the other based on the case type. There exists a helper function `isIndictmentCase` which can be used for this purpose.

gudjong (changes requested):

gudjong (approved):

PR #10976, opened 2023-05-02
fix(delegations): iOS scroll bug

hedinne:
## What

This fixes a bug where users on iOS devices couldn't scroll within a modal. 

## Checklist:

- [x] I have performed a self-review of my own code
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] Formatting passes locally with my changes
- [x] I have rebased against main before asking for a review

PR #11013, opened 2023-05-03
feat(portals-shared-modules-delegations): Drawer modal

hedinne:
## What

New design for modals in mobile view. A bit of refactoring within the delegations modal and most importantly, fixes the iOS scroll bug. Previus attempts to fix this bug in #10976, #10970, #10965 and #10935

The scroll-locking strategy that Reakit modal uses doesn't work on iOS so I had to disable it and use a different package for it. 
See: https://github.com/ariakit/ariakit/issues/1085#issuecomment-1083044216

The new implementations can handle sticky footers.

## Screenshots / Gifs

Screenshot 2023-05-03 at 17 07 58
Screenshot 2023-05-03 at 17 14 26

## Checklist:


libs/portals/shared-modules/delegations/src/components/Modal/Modal.css.ts
```diff
@@ -1,29 +1,101 @@
...
+    left: 0,
+    right: 0,
+    maxHeight: ['85vh', `calc(100dvh - ${theme.spacing[8]}px)`],
+    width: '100%',
+    maxWidth: 888,
+
+    paddingTop: theme.spacing[3],
+
+    backgroundColor: theme.color.white,
+    boxShadow: '0px 4px 30px rgba(0, 97, 255, 0.16)',
```
snaerseljan, line 24: This value exists in the theme

libs/portals/shared-modules/delegations/src/components/Modal/Modal.css.ts
```diff
@@ -1,29 +1,101 @@
...
+
+        maxWidth: 'min(calc(100vw - 64px), 888px)',
+
+        paddingTop: theme.spacing[9],
+      },
+    }),
+  },
+  variants: {
+    noPaddingBottom: {
+      true: {},
```
snaerseljan, line 43: You can probably skip true

libs/portals/shared-modules/delegations/src/components/Modal/Modal.tsx
```diff
@@ -30,47 +49,39 @@ export const Modal = ({
...
-          tabIndex={-1}
-          ref={headingRef}
-        >
-          {label && (
-            <Text variant="small" fontWeight="semiBold">
-              {label}
+      <Box className={styles.header} ref={headingRef}>
+        <Box display="flex" flexDirection="column" rowGap={1}>
+          {eyebrow ? <Text variant="eyebrow">{eyebrow}</Text> : null}
+          {title ? (
```
snaerseljan, line 69: ~nit: same as above

snaerseljan (approved):
Nicely done!  👍  
This must have been frustration to solve.
I did not start up the branch and test it out, because I am sure you tested it thoroughly :)

libs/portals/shared-modules/delegations/src/components/Modal/Modal.tsx
```diff
@@ -30,47 +49,39 @@ export const Modal = ({
...
-          rowGap={2}
-          tabIndex={-1}
-          ref={headingRef}
-        >
-          {label && (
-            <Text variant="small" fontWeight="semiBold">
-              {label}
+      <Box className={styles.header} ref={headingRef}>
+        <Box display="flex" flexDirection="column" rowGap={1}>
+          {eyebrow ? <Text variant="eyebrow">{eyebrow}</Text> : null}
```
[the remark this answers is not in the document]
hedinne: I always try to use ternaries rather than Short-circuiting with && because if the value is a falsy, that value will be rendered. This is not a big problem, but I thater be precise in my logic, even if it is a bit uglier.

I have had cases where a 0 is rendered to the DOM.

PR #11039, opened 2023-05-04
refactor(j-s): Appeals Court Access

gudjong:
# Appeals Court Access

[Rewrite single case filter](https://app.asana.com/0/1199153462262248/1204522261372966/f)

## What

- Rewrites the read and write access checks to a single case. Access is controlled per role and institution meaning that the same role will not necessarily have access to the same cases at a district court and the appeals court.
- Allows two roles at the appeals court: judge and registrar. This is not a change and we will use the registrar role for assistants at the appeals court.
- Restricts the selection of appeals court assistants to registrars roles at the appeals court. This is a change in the client on the appeals court registration page.
- Allows all court roles to perform the same updates on cases they have access to. For example, if an appeals judge has access to a given custody case, then that judge can sign the district court ruling. In practice this should not be a problem. The client restricts what users can do and we do not expect court users to intentionally do stupid things on purpose such as forcing call to the backend.
- Updates case transition roles rules with regards to appeals.

## Why

- Corrected and improved access control.

## Checklist:


apps/judicial-system/backend/src/app/modules/case/filters/case.filter.ts
```diff
@@ -2,228 +2,207 @@ import {
...
+  // Check case type access
+  if (user.role === UserRole.PROSECUTOR) {
+    if (
+      !isRestrictionCase(theCase.type) &&
+      !isInvestigationCase(theCase.type) &&
+      !isIndictmentCase(theCase.type)
+    ) {
+      return false
+    }
+  } else if (!isIndictmentCase(theCase.type)) {
```
MargretFinnboga, line 35: Why are we checking !isIndictmentCase(theCase.type) twice?

MargretFinnboga (approved):
tiny nitpicks, otherwise looks good 🔐

apps/judicial-system/backend/src/app/modules/case/filters/case.filter.ts
gudjong: We are not really checking it twice, but rather doing different checks base on user role.

So basically:
- If you are a prosecutor and the case type is not restriction, investigation or indictment, then you do not have access. Currently, this means prosecutors have access to all cases, but if we add not types of cases, prosecutors will not automatically get access.
- If you are some other role (currently only representative) and the case type is not indictment, then you do not get access.

PR #11049, opened 2023-05-05
feat(application-system): Events used as log markers for application history

obmagnusson:
# ...
 A small refactor of the history functionality
## What

Using events to mark logs. 

## Why

Previous desing did not account for some states going back and forth with different actions.  Resulting in messages that would not apply.

## Screenshots / Gifs

## Checklist:


Toti91 (approved):
Sweet i like these changes

libs/application/api/history/src/lib/history.service.ts
```diff
@@ -22,8 +20,9 @@ export class HistoryService {
   async saveStateTransition(
```
Thorrik58, line 20: do these need to be documented?

libs/application/templates/data-protection-complaint/src/lib/DataProtectionComplaintTemplate.ts
```diff
@@ -39,6 +39,14 @@ const DataProtectionComplaintTemplate: ApplicationTemplate<
             shouldBePruned: true,
             whenToPrune: 5 * 60000, //5 minutes
           },
+          actionCard: {
+            historyLogs: [
```
Thorrik58, line 43: feel like these properties within the object could be more descriptive.

libs/application/templates/reference-template/src/lib/ReferenceApplicationTemplate.ts
```diff
@@ -142,7 +142,12 @@ const ReferenceApplicationTemplate: ApplicationTemplate<
 
           actionCard: {
             description: m.draftDescription,
-            onExitHistoryLog: 'Umsókn send inn',
+            historyLogs: [
+              {
+                event: DefaultEvents.SUBMIT,
+                log: 'Umsókn send inn',
```
Thorrik58, line 148: Werent these supposed to be translation strings?

Thorrik58 (changes requested):
questions 🗡️

apps/application-system/api/migrations/20230504150905-history-add-exit-event.js
```diff
@@ -0,0 +1,20 @@
+'use strict'
+
+module.exports = {
+  up: (queryInterface, Sequelize) => {
+    return queryInterface.sequelize.transaction((t) =>
+      Promise.all([
+        queryInterface.addColumn('state_history', 'exit_event', {
```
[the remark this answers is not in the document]
obmagnusson: I think tracking the exit event should be sufficient. As we are logging on events in states. I will add a previous state for future proofing.

libs/application/api/history/src/lib/history.service.ts
obmagnusson: Idk think they are pretty descriptive as is. Plus this shoulndt be used outside these use cases in the system.

PR #11051, opened 2023-05-05
fix(j-s): Fix Appeal Court Folder

gudjong:
# Fix Appeal Court Folder

[Vista greinargerð og meðfylgjandi gögn í réttri möppu í Auði](https://app.asana.com/0/1199153462262248/1204540555171599/f)

## What

- Corrects the folder where appeal statements and accompanying files are store in the court system.

## Why

- They were being stored in the wrong folder.

## Checklist:


gudjong:
> Only thing I'm wondering is whether a lot of this code is going to be redundant now that we are not supposed to save the appeal statements in Héraðs Auði?

Yes some of it will, but I would rather just let it through with the rest of the code and deal with deletions separately.

PR #11052, opened 2023-05-05
feat(j-s): Store Case Files Record in AWS S3

gudjong:
# Store Case Files Record in AWS S3

[Vista skjalaskrár í S3](https://app.asana.com/0/1199153462262248/1203551118784320/f)

## What

- Stores case files records in AWS S3 when a case is received by a district court.
- Gets case files records from AWS S3 once stored there.
- Moves case files records to an AWS S3 folder with a limited lifespan when a case is completed.

## Why

- Generating case files record PDFs takes a long time and uses a lot of memory, so storing them in AWS S3 improves user experience.

## Checklist:


unakb (approved):
Looks good, tried locally and files went to folders as expected

PR #11060, opened 2023-05-08
feat(react-components): Shared modal

hedinne:
## What

Move the delegations (and ids-admin) modal to a shared library (react-components). And add `scrollType` prop that controls if the scroll is inside the modal or on the overlay. The "outside" scroll has been the default in island.is but having the scroll bar inside the modal can be useful when we want to have a sticky footer on the modal.
On mobile, the scroll is always inside the modal so the close button is always visible.

## Why
It is easier to maintain one modal than two copies of the same modal.

## Screenshots / Gifs

[Both scroll types on desktop](https://user-images.githubusercontent.com/7573694/236821990-c4f11a23-b9af-45b1-a471-d8ed7e67d080.webm)

[Both scroll types on mobile](https://user-images.githubusercontent.com/7573694/236822206-db2ccc0d-337a-46b5-bdb0-dd5a6bbe04a1.webm)

## Checklist:


saevarma (approved):
Really nice refactor  🙌  great job!

PR #11093, opened 2023-05-09
feat(admin-api): Permissions Rest endpoints and GQL resolvers

snaerseljan:
# Get permissions and permission endpoints

This PR includes rest endpoint for both list of permissions and single permission, aka scope.
It also includes graphql resolvers for both of those endpoints.

## Screenshots / Gifs
**AuthAdminScopes**
image

**AuthAdminScope**
image

## Checklist:


libs/api/domains/auth-admin/src/lib/scope/models/scope-environment.model.ts
```diff
@@ -1,20 +1,64 @@
...
 import { TranslatedValue } from '../../models/translated-value.model'
 
 @ObjectType('AuthAdminScopeEnvironment')
 export class ScopeEnvironment {
   @Field(() => Environment)
   environment!: Environment
 
-  @Field(() => String, { nullable: false })
+  @Field(() => Boolean)
+  enabled!: boolean
```
saevarma, line 12: Shouldn't the service just always return enabled scopes so we don't have to use this field?

libs/api/domains/auth-admin/src/lib/client/clients.service.ts
```diff
@@ -143,21 +142,13 @@ export class ClientsService extends MultiEnvironmentService {
...
-          }
-        } else if (resp.status === 'rejected') {
-          this.logger.error(
-            `Failed to create application ${input.clientId} in environment ${input.environments[index]}`,
-            resp.reason,
-          )
-        }
-      })
-      .filter(isDefined)
+    return this.handleSettledPromises(settledPromises, {
```
saevarma, line 145: Nice! We should make a nice to have task to use this handler as well in the `getClients` and `getClientById`

libs/api/domains/auth-admin/src/lib/scope/models/scope-environment.model.ts
```diff
@@ -1,20 +1,64 @@
...
   description!: TranslatedValue[]
+
+  @Field(() => String)
+  domainName!: string
+
+  @Field(() => Number)
+  order?: number
+
+  @Field(() => String)
+  groupId?: string
```
saevarma, line 30: These should also be marked as `nullable: true` in the GQL schema

libs/api/domains/auth-admin/src/lib/scope/models/scope-environment.model.ts
```diff
@@ -1,20 +1,64 @@
...
+  allowExplicitDelegationGrant!: boolean
+
+  @Field(() => Boolean)
+  automaticDelegationGrant!: boolean
+
+  @Field(() => Boolean)
+  alsoForDelegatedUser!: boolean
+
+  @Field(() => Boolean)
+  isAccessControlled?: boolean
```
saevarma, line 63: Should always be set and available as the default value would just be `false`.
```suggestion
  isAccessControlled!: boolean
```

apps/services/auth/admin-api/src/app/v2/scopes/me-scopes.controller.ts
```diff
@@ -39,18 +39,36 @@ export class MeScopesController {
...
+    response: { status: 200, type: ApiScopeDTO },
+  })
+  @Audit<ApiScopeDTO>({
+    resources: (scope) => scope.name,
+  })
   create(
     @Param('tenantId') tenantId: string,
     @Body() input: ClientCreateScopeDTO,
-  ): Promise<ApiScopesDTO> {
+  ): Promise<ApiScopeDTO> {
```
saevarma, line 71: We should be returning `AdminScopeDTO`

apps/services/auth/admin-api/src/app/v2/scopes/me-clients-scopes.controller.ts
```diff
@@ -28,15 +28,15 @@ export class MeClientsScopesController {
...
   })
-  @Audit<AdminScopeDTO[]>({
+  @Audit<ApiScopeDTO[]>({
     resources: (scopes) => scopes.map((scope) => scope.name),
   })
   findAll(
     @Param('tenantId') tenantId: string,
     @Param('clientId') clientId: string,
-  ): Promise<AdminScopeDTO[]> {
+  ): Promise<ApiScopeDTO[]> {
```
saevarma, line 39: We should be returning `AdminScopeDTO`

libs/auth-api-lib/src/lib/resources/admin/admin-scope.service.ts
```diff
@@ -33,7 +36,86 @@ export class AdminScopeService {
...
+      {
+        locale: 'is',
+        value: defaultValueIS,
+      },
+      ...Array.from(translations || []).map(([locale, translation]) => ({
+        locale,
+        value: translation.get(key) ?? '',
+      })),
+    ]
+  }
```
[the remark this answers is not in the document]
snaerseljan: I created a new service for the utility

image

libs/api/domains/auth-admin/src/lib/client/clients.service.ts
snaerseljan: https://www.notion.so/Refactor-code-in-client-service-to-replicate-code-in-scope-service-e655090a677d44c59073fe6b55177251?pvs=4

saevarma (approved):

PR #11095, opened 2023-05-10
chore(core): Upgrade configcat-js and configcat-node

eirikurn:
## What

Upgrade configcat packages.
Make the feature flag types more generic to match the upgraded configcat types.

## Why

To stay up to date and lower our requests per month.

## Checklist:


libs/react/feature-flags/src/lib/useFeatureFlag.ts
```diff
@@ -28,7 +32,7 @@ export const useFeatureFlag = <T extends boolean | string>(
       .getValue(featureFlag, defaultValue, user)
       .then((value) => {
         if (mounted) {
-          setState({ value: value as T, loading: false })
+          setState({ value: value as SettingTypeOf<T>, loading: false })
```
snaerth, line 35: Does TS not infer the value?

libs/feature-flags/src/lib/configcat.ts
```diff
@@ -15,18 +22,20 @@ export class Client implements FeatureFlagClient {
...
-      this.configcat = eval('require')('configcat-node').createClient(
+      this.configcat = eval('require')('configcat-node').getClient(
         resolvedSdkKey,
+        null,
         ccConfig,
       )
     } else {
       // eslint-disable-next-line @typescript-eslint/no-var-requires
-      this.configcat = require('configcat-js').createClient(
+      this.configcat = require('configcat-js').getClient(
```
snaerth, line 36: Is the updated package, not an ES module? Just wondering if import() should be used instead

snaerth (approved):
Awesome work! 💯 
How does this save us requests per month?

libs/feature-flags/src/lib/configcat.ts
eirikurn: This module is dynamically importing web vs nodejs version of configcat. We're trying not to run or bundle both.

snaerseljan (approved):

snaerseljan (approved):

PR #11097, opened 2023-05-10
fix(application-driving-learners-permit): Validation and license fallbacks

kksteini:
# What
The eligibility check in the beginning of a driving-learners-permit application assumes that a disqualification/deprivation will be present on the driving license fetched for the current applicant (instructor).
This is not the case, instead we must react to a deprivation error upon request and then build the expected license object and return it with the deprivation attached. This corrects for the typing/data assumptions and ensure that we communicate back the correct eligibility error to the user.

Additionally we add TemplateApiErrors to the submission to communicate when the Driving License Registry (is: ökuskírteinaskrá) rejects it. This should not happen since validations take place before submission but it doesn't hurt to include it anyway.

# Fixes
* Error/validation handling for submission endpoint
* Driving licenses do not contain disqualification information even though the generated type for the driving license suggests otherwise. The license endpoint throws on disqualifications and as such we must catch that particular error and return a driving license with attached disqualification/deprivation to resume assumed behaviour.

## Checklist:


infra/src/dsl/xroad.ts
```diff
@@ -118,7 +118,7 @@ export const DrivingLicense = new XroadConf({
       prod: 'r1/IS/GOV/5309672079/Logreglan-Protected/Okuskirteini-v1',
     },
     XROAD_DRIVING_LICENSE_V5_PATH: {
-      dev: 'r1/IS-DEV/GOV/10005/Logreglan-Protected/Okuskirteini-v5',
+      dev: 'r1/IS-DEV/GOV/10005/Logreglan-Protected/okuskirteini-v5',
```
kksteini, line 121: When connecting to the dev server, it's 'okuskirteini-v5' instead of 'Okuskirteini-v5' for some reason.
This fixes it quickly though a case could be made to contact the relevant parties and ask them to change it to a consistant value.
This feels low priority though.

libs/api/domains/driving-license/src/lib/drivingLicense.service.ts
```diff
@@ -52,10 +52,12 @@ export class DrivingLicenseService {
 
   async getDrivingLicense(
     nationalId: User['nationalId'],
+    token?: string,
```
kksteini, line 55: For now, this is optional and is used in v5 fallbacks when a disqualification is present on the driving license.

albinagu (approved):

brynjarorng (approved):
LGTM(Devops)

Toti91 (approved):
Code owned LGTM

PR #11105, opened 2023-05-10
fix(j-s): Validate and send updated ruling info to server

unakb:
[Úrskurðarskjár Landsréttar - lagfæringar](https://app.asana.com/0/1199153462262248/1204561410209285)

## What

- Add validation to Ruling page for Court of Appeal 
- Send updated information to the server
- Remove auto selected radio button for the appeal ruling decision

## Why

So that the ruling decision is updated in the backend when changed and so that the user cannot continue to the next step without having filled out all required fields

## Checklist:


gudjong (approved):
Looks ok to me

PR #11113, opened 2023-05-10
feat(auth-admin): Permissions graphql and frontend connection

GunnlaugurG:
# ...

## What

Created functionality for adding and removing permissions for client
- Modified rest endpoint for scopes to return the domain name to check if scope is locked
- Graphql field resolver for available permissions for selected tenant
- Graphql mutation modification to patch addedScopes and removedScopes
- Modified Client query to fetch allowedScopes connected to client and available scopes for selected tenant
- Connected the Permission and the AddPermssion components to display correct values
- Display item as locked if they belong in another tenant
- Modified the ContentCard to not display multiEnv actions when needed
- Sync and save to all env removed from Client Url card
- Sync and save to all env removed if there is only one available env

## Why

So users can add and remove permissions 

## Screenshots / Gifs

https://github.com/island-is/island.is/assets/34029342/bc1cc9c2-5d17-4bae-9635-1b593ece4395

## Checklist:


libs/portals/admin/ids-admin/src/components/Client/Permissions.tsx
```diff
@@ -1,86 +1,78 @@
...
+  }
 
   const handleModalClose = () => {
     setIsModalVisible(false)
   }
 
-  const handleModalOpen = () => {
-    setIsModalVisible(true)
+  // If removing permissions, we want to remove it from both the permissions and addedPermissions array and add it to the removedPermissions array if it doesn't exist there
+  const handleRemovedPermissions = (
```
saevarma, line 43: ~The function is handling removal of a single permission
```suggestion
  const handleRemovedPermission = (
```

libs/portals/admin/ids-admin/src/components/forms/AddPermissions/AddPermissions.tsx
```diff
@@ -8,125 +8,138 @@ import {
...
 }
 
-type Permission = {
-  id: string
-  label: string
-  description: string
-  api: string
-  locked?: boolean
-}
+function AddPermissions({ isVisible, onClose, onAdd }: AddPermissionsProps) {
```
saevarma, line 23: When this modal starts quering the scopes available in the tenant remember to filter out allowed scopes.

libs/portals/admin/ids-admin/src/components/forms/EditApplication/EditApplication.action.ts
```diff
@@ -27,6 +27,9 @@ export enum ClientFormTypes {
 }
 
 const splitStringOnCommaOrSpaceOrNewLine = (s: string) => {
+  if (!s) {
+    return []
+  }
   return s.split(/\s*,\s*|\s+|\n+/)
```
saevarma, line 33:
```suggestion
  return s?.split(/\s*,\s*|\s+|\n+/) ?? []
```

libs/portals/admin/ids-admin/src/components/forms/EditApplication/EditApplication.action.ts
```diff
@@ -220,7 +223,12 @@ export const schema = {
         }),
     })
     .merge(defaultSchema),
-  [ClientFormTypes.permissions]: defaultSchema,
+  [ClientFormTypes.permissions]: z
+    .object({
+      addedScopes: z.string().transform(splitStringOnCommaOrSpaceOrNewLine),
+      removedScopes: z.string().transform(splitStringOnCommaOrSpaceOrNewLine),
```
saevarma, line 229: After saving the settings we need to reset these arrays to empty?

saevarma (commented):
Overall nice job 👏
We just need to change to use gql query in the modal instead of the loader.

libs/portals/core/src/utils/getTranslatedValue.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export const getTranslatedValue = (
+  translatedValue: AuthAdminTranslatedValue[],
+  locale: string,
+) => {
+  return (
+    translatedValue.find(({ locale: l }) => l === locale)?.value ||
+    translatedValue[0]?.value ||
+    ''
+  )
```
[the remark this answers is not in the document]
GunnlaugurG: My idea was that if we do not have the locale, we would fallback to a default message (which is at index 0), but you are right that could display some weird state

libs/portals/admin/ids-admin/src/components/forms/EditApplication/EditApplication.action.ts
GunnlaugurG: are you saying that we need to clear these values in the Zod object, or on the context ?

libs/portals/admin/ids-admin/src/components/Client/Permissions.tsx
```diff
@@ -138,7 +131,21 @@ function Permissions({ data = mockData }: PermissionsProps) {
           </T.Table>
         </ShadowBox>
       )}
-      <AddPermissions onClose={handleModalClose} isVisible={isModalVisible} />
+      <input
```
snaerseljan, line 134: A better pattern handling hidden input arrays is the following way.

```
          {addedScopes.map(({ name }) => (
              <input
                key={env}
                type="hidden"
                name="addedScopes"
                value={name}
              />
            ))}
```

Then in the action

const schema = {
    ...,
    addedScopes: zfd.repeatable(z.optional(z.array(z.string())),
} 
```

That way you don't have to do the transformations back and forth

Comment (GunnlaugurG) [reply libs/portals/admin/ids-admin/src/components/Client/Permissions.tsx:166]: That is awesome, thank you for that tip!

File: libs/portals/admin/ids-admin/src/components/Client/Permissions.tsx [hunk 7]
```diff
@@ -1,86 +1,110 @@
...
+      }
+    }
+  }, [actionData])
+
+  const {
+    addedScopes,
+    removedScopes,
+    setRemovedScopes,
+    setAddedScopes,
+  } = useContext(ClientContext)
```
saevarma, line 62: This component should just keep the state of added and removed scopes and the modal's `onAdd` function feeds back the added scope to update the state.

libs/portals/admin/ids-admin/src/components/Client/Permissions.tsx
```diff
@@ -1,86 +1,110 @@
...
   const [isModalVisible, setIsModalVisible] = useState(false)
+  const params = useParams()
+  const [permissions, setPermissions] = useState<AuthAdminClientAllowedScope[]>(
+    allowedScopes ?? [],
+  )
+  const actionData = useActionData() as EditApplicationResult<
+    typeof schema[typeof ClientFormTypes.permissions]
+  >
+
+  const actionDataRef = useRef(actionData?.data)
```
saevarma, line 41: This is initialied as `undefined` on the initial render as the action hasn't been posted. In the useEffect you should just be able to check if action contains data then clear state. We need to make sure that we can repeat this, i.e. we can remove or add scopes and save and then add some more and save.

libs/portals/admin/ids-admin/src/components/Client/Permissions.tsx
```diff
@@ -1,86 +1,110 @@
...
+      if (!isEqual(actionData?.data, actionDataRef?.current)) {
+        if (actionData?.data) {
+          setAddedScopes([])
+          setRemovedScopes([])
+        }
+        if (actionData?.globalError) {
+          toast.error(formatMessage(m.globalErrorMessage))
+        }
+      }
+    }
```
saevarma, line 54: If there is `globalError` we don't have data so this if statements can be simplified.
```suggestion
    if (actionData?.intent === ClientFormTypes.permissions) {
        if (actionData?.globalError) {
          toast.error(formatMessage(m.globalErrorMessage))
        }
       else if (actionData?.data) {
          setAddedScopes([])
          setRemovedScopes([])
        }
    }
```

libs/portals/admin/ids-admin/src/components/Client/Permissions.tsx
```diff
@@ -98,33 +122,34 @@ function Permissions({ data = mockData }: PermissionsProps) {
...
-                <T.HeadData>
-                  {formatMessage(m.permissionsTableLabelAPI)}
-                </T.HeadData>
                 <T.HeadData>{/* For matching column count */}</T.HeadData>
               </T.Row>
             </T.Head>
             <T.Body>
-              {data.map((item) => (
-                <T.Row key={item.id}>
+              {[...permissions].map((item) => (
```
saevarma, line 129: `permissions` is already an array
```suggestion
{permissions.map((item) => (
```

libs/portals/admin/ids-admin/src/components/Client/Permissions.tsx
```diff
@@ -1,86 +1,110 @@
...
+  const handleRemovedPermission = (
+    removedPermission: AuthAdminClientAllowedScope,
+  ) => {
+    const newPermissions = permissions.filter(
+      (item) => item.name !== removedPermission.name,
+    )
+    const newAddedScopes = addedScopes.filter(
+      (item) => item.name !== removedPermission.name,
+    )
+    setAddedScopes(newAddedScopes)
```
saevarma, line 84: Inline state edit for both added scopes and permissions.
```suggestion
    setAddedScopes((prev) => prev.filter((item) => item.name !== removedPermission.name))
```

libs/portals/admin/ids-admin/src/components/Client/Permissions.tsx
```diff
@@ -1,86 +1,110 @@
...
+    setPermissions(newPermissions)
+
+    setRemovedScopes((prevState) => {
+      if (
+        [...addedScopes, ...prevState].find(
+          (item) => item.name === removedPermission.name,
+        )
+      ) {
+        return prevState
```
saevarma, line 93: I don't understand why concat `addedScopes`?

libs/portals/admin/ids-admin/src/components/forms/AddPermissions/AddPermissions.tsx
```diff
@@ -8,125 +8,167 @@ import {
...
+    onCompleted: (data) => {
+      const scopes = data.authAdminScopes?.data
+        .map((item) => {
+          return (
+            item.environments.find(
+              (scope) => scope.environment === selectedEnvironment.environment,
+            ) ?? null
+          )
+        })
+        .filter((item) => item !== null)
```
saevarma, line 59: You can use the `isDefined` utility here
```suggestion
        .filter(isDefined)
```

libs/portals/admin/ids-admin/src/components/forms/AddPermissions/AddPermissions.tsx
```diff
@@ -8,125 +8,167 @@ import {
...
+
+  const {
+    selectedEnvironment,
+    setAddedScopes,
+    addedScopes,
+    removedScopes,
+    setRemovedScopes,
+  } = useContext(ClientContext)
+
+  const { loading } = useGetAvailableScopesQuery({
```
saevarma, line 43: You can get the data from the query and use then component so you don't need the local availableScope state
```suggestion
  const {data, loading } = useGetAvailableScopesQuery({
```

libs/portals/admin/ids-admin/src/components/forms/AddPermissions/AddPermissions.tsx
```diff
@@ -8,125 +8,167 @@ import {
...
-  },
-]
+    // Remove the scopes that were added that were also in the removedScopes array
+    setAddedScopes(
+      newAddedScopes.filter(
+        (item) => !removedScopes.some((rem) => rem.name === item.name),
+      ),
+    )
+    // Remove the scopes that were added from the removedScopes array
+    setRemovedScopes(
```
saevarma, line 86: This should just be handled in the `onAdd` function of the `Permissions` component.

saevarma (commented):
Found few more details that we should improve before merging this.

libs/portals/admin/ids-admin/src/components/Client/Permissions.tsx
GunnlaugurG: if you are talking about the [...addedScopes, ...prevState] then I am checking if it already exists in the removedScopes, or if it exists in addedScopes. Then I dont want to add it to the removedScopes state, then it just enough to remove it from the addedScopes array.

GunnlaugurG: If it was previously in the addedScopes it means that it was added from AddPermission.tsx and has not been sent to the server, so when it is removed, we dont want it in the array for removedScopes that gets sent to the server.

libs/portals/admin/ids-admin/src/components/Client/Permissions.tsx
GunnlaugurG: You are right, I don't know why I was so fixated on putting it in the scope..

libs/portals/admin/ids-admin/src/components/Client/Permissions.tsx
GunnlaugurG: you are right that this gets initialised as undefined until it has been posted. But the thing is that the submit action for this shares a function with a lot of other submit actions, so lets say I submit a change on the display name, then if I would only check if it has data, but don't check if it was a permission that was submitted or a displayName it would reset the added and removed arrays. The isEqual check may be a bit over the top, I will remove that, my idea there was to prevent a reset if the same data gets posted twice.

libs/portals/core/src/utils/getTranslatedValue.ts
```diff
@@ -0,0 +1,8 @@
+import { AuthAdminTranslatedValue } from '@island.is/api/schema'
+
+export const getTranslatedValue = (
+  translatedValue: AuthAdminTranslatedValue[],
```
snaerseljan, line 4: This is already on main copied from this PR btw :)

snaerseljan (approved):
Nicely done 👏

libs/portals/admin/ids-admin/src/shared/components/ContentCard/ContentCard.tsx
```diff
@@ -71,6 +75,12 @@ const ContentCard: FC<ContentCardProps> = ({
 
   const actionDataRef = useRef(actionData?.data)
 
+  useEffect(() => {
+    if (typeof isDirty !== 'function') {
```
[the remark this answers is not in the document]
saevarma: I disagree, it is unnecessary to introduce a dependency on lodash and `typeof` is faster.

PR #11119, opened 2023-05-11
fix(ids-admin): Permissions table sticky header

snaerseljan:
# Client permissions table fixed header

## What

In our client details page we have a permissions table section which is scrollable. This PR makes the header sticky and by doing that a much better UX.

## Why

Bad UX when scrolling

## Checklist:


saevarma (approved):
🙌

PR #11126, opened 2023-05-11
feat(search-toolkit): Remove unnecessary property

rafnarnason:
using relatedArticles property here is causing unnecessary api calls to contentful - removing those from gql query and cleaning up the query in general

## What

## Why

## Screenshots / Gifs

## Checklist:

---
  - https://app.asana.com/0/0/1204545857539832


RunarVestmann (approved):

eirikurn (approved):
I would appreciate if you could dive deeper and remove all other fields which are not needed in the search. Especially in the autocomplete.

kksteini (approved):

PR #11133, opened 2023-05-12
feat(admin-api): Api scope PATCH in GQL and rest

snaerseljan:
# Admin api scope PATCH

## What

**admin-api** scope patch endpoint and updateScope resolver method in **api**

## Why

Ids admin has to have scope patch to let clients manage their scopes

## Checklist:


libs/auth-api-lib/src/lib/resources/admin/admin-scope.service.ts
```diff
@@ -138,4 +145,127 @@ export class AdminScopeService {
       translations,
     )
   }
+
+  /**
+   * Finds all none icelandic translated values
+   */
+  private findNoneIcelandicTranslatedValueDto(
```
saevarma, line 152: ~Should this function be located in the `admin-translation.service.` instead?

saevarma (approved):
💯

PR #11154, opened 2023-05-15
feat(auth): Mutation for patching consent.

valurefugl:
https://www.notion.so/GQL-mutation-and-schema-to-update-consents-1725a15cbc1146b1a5e2a992180afa4f?pvs=4

## What

## Why

## Screenshots / Gifs

## Checklist:


libs/api/domains/auth/src/lib/dto/updateConsent.input.ts
```diff
@@ -0,0 +1,11 @@
+import { Field, InputType } from '@nestjs/graphql'
+
+@InputType('UpdateAuthConsentInput')
+export class UpdateConsentInput {
```
saevarma, line 4: As the mutation is named `patchAuthConsent` we should probably reflect that in the input name.
```suggestion
@InputType('PatchAuthConsentInput')
export class PatchConsentInput {
```

valurefugl: Fixed.

thordurhhh (approved):

PR #11155, opened 2023-05-15
fix(estate): Estate hiding options for prod

albinagu:
It was decided to begin with "Búsetuleyfi" only, so we are hiding the rest of the options on prod for now, but will continue with finishing up/testing the rest on dev.

## Checklist:

---
  - https://app.asana.com/0/0/1204590737451215


libs/application/templates/estate/src/lib/constants.ts
```diff
@@ -33,3 +33,30 @@ export const EstateTypes = {
...
+          value: EstateTypes.divisionOfEstateByHeirs,
+          label: EstateTypes.divisionOfEstateByHeirs,
+        },
+      ]
+    : [
+        {
+          value: EstateTypes.permitToPostponeEstateDivision,
+          label: EstateTypes.permitToPostponeEstateDivision,
+        },
+      ]
```
kksteini, line 62: I'm ok with this but it will slow down the process of releasing.
It may be considered bad form to hotfix these switches but I am not sure :)

kksteini (approved):

kksteini (approved):

libs/application/templates/estate/src/lib/constants.ts
kksteini: Although, I doubt a quadruple feature flag is the answer either :laughing:

albinagu: I was pointed at much better solution actually, stay tuned!

PR #11188, opened 2023-05-17
fix(system-e2e): Fixing a couple of tests

pshomov:
# ...

Missing data for payment plan and popup window that was stopping the test.
Someone changed the locale for the test user and that tripped the sessions tests up.

## What

## Why

## Screenshots / Gifs

## Checklist:

### DIFF
--- a/apps/system-e2e/src/tests/islandis/service-portal/smoke/financial-payment-application.spec.ts
+++ b/apps/system-e2e/src/tests/islandis/service-portal/smoke/financial-payment-application.spec.ts
@@ -49,30 +49,40 @@ test.describe('MS - Fjármál overview', () => {
         `role=button[name="${label(m.scheduleApplication)}"]`,
       )
 
-      const pagePromise = context.waitForEvent('page')
-      applicationButton.click()
-      const newPage = await pagePromise
-      await newPage.waitForLoadState()
+      const [delegationPopup] = await Promise.all([
+        page.waitForEvent('popup'),
+        // Click - Open Link in Popup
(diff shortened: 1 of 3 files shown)


apps/system-e2e/src/tests/islandis/service-portal/smoke/financial-payment-application.spec.ts
```diff
@@ -49,30 +49,40 @@ test.describe('MS - Fjármál overview', () => {
...
+      ])
 
+      // Choose delegation
+      await delegationPopup
+        .getByRole('button', { name: 'Gervimaður Færeyjar' })
+        .click()
       // Assert
-      expect(newPage.url()).toContain('umsoknir/greidsluaaetlun')
+      await expect(
+        delegationPopup.getByText('Viltu nota gervigögn?'),
```
thordurhhh, line 64: Should this use the translation string?

thordurhhh (approved):
One minor question. But otherwise LGTM.

apps/system-e2e/src/tests/islandis/service-portal/smoke/financial-payment-application.spec.ts
pshomov: I was not sure which text to use since there were 2 🤷🏼‍♂️. Suggestions?

thordurhhh: It's hardcoded in the application since this is not for production. So this is perfectly fine.

PR #11191, opened 2023-05-17
fix(service-portal): Add session storage hide modal on param

thordurhhh:
## What

* When `hide_onboarding_modal` query param is visible, add to session storage.

## Why

* For testing, disable once then the whole session is disabled, avoiding the popup if user is redirected.

## Checklist:


thorkellmani (approved):
lgtm

PR #11208, opened 2023-05-19
fix(delegation-api): Read domainName from client and fallback to parse from clientId.

saevarma:
## What

Read the domain name from the client model and fallback to try to parse it from the client ID if not configured.

## Why

We recently added the domainName field to the client model and this uses that to read the domain name. It can be missing so we continue to fallback to the client id.

## Screenshots / Gifs

N/A

## Checklist:


snaerseljan (approved):
Nice catch

PR #11219, opened 2023-05-22
fix(web): General petitions web view update

albinagu:
Screenshot 2023-05-22 at 11 37 14

## Checklist:


kksteini (approved):
LGTM

RunarVestmann (commented):
I'm wondering whether we'd like to do something specific if for example a user types in a random id in the url
image

Also if I click on the "Setja nafn mitt á þennan lista" I get an error if the list id isn't valid, could we do something about that :/ 
image

apps/web/screens/PetitionView/PetitionView.tsx
```diff
@@ -253,4 +258,31 @@ const PetitionView = ({ namespace }: PetitionViewProps) => {
   )
 }
 
+interface InitialProps {
```
RunarVestmann, line 281: You could type the component as a Screen instead of creating this InitialProps interface

RunarVestmann: I commited that change with your permission :)

RunarVestmann (approved):

PR #11245, opened 2023-05-23
feat(ids-admin): Various fixes in edit permissions and client

snaerseljan:
# Various fixes in edit permission and edit client

## What

- Remove unnecessary audits.
- Updates edit client screen context to use a provider and a `useClient` hook
- Move some logic to the `ClientProvider` instead of having it in the `Client.tsx` component which is called `EditClient.tsx` now.
- Change environment tags on permission card to links to set the currently selected env state in permission.
- Removed unnecessary defaultChecked on checkboxes in `PermissionAccessControl.tsx` component.
- `useEnvironment()` hook to get the current environment from the URL query string, or the first item in environments array if none query string is provided.
- Skip unavailable environments if !allowPublishing, so they don't show up in the list.
- Move `Client` and `Clients` to screens

## TODO

- [x] Fix minor regression in sync dropdown update

## Checklist:


libs/portals/admin/ids-admin/src/components/Client/ClientContext.tsx
```diff
@@ -0,0 +1,158 @@
...
+      }),
+      { preventScrollReset: true },
+    )
+  }
+
+  const checkIfInSync = (variables: string[]) => {
+    for (const v of variables) {
+      const variable = v as keyof AuthAdminClientEnvironment
+
+      for (const env of client.environments) {
```
saevarma, line 109: ~We could optimise this to skip the selected environment.

saevarma (approved):
💅

PR #11246, opened 2023-05-23
feat(license-service): Use drivers-license-v5

thorkellmani:
## What

* Feature flag and digital drivers license barcode based conditional injection of license service drivers license clients
* Use v4 drivers license client in license api
* Use v4/v5 drivers license client in license service IF user is feature flagged, or the license being scanned is v5

## Why

A migration of digital drivers license to a more stable version is underway. Migrating the most common digital license in the country at once seems risky, since we can't test it fully on dev, so the plan is to batch the migrations. The feature flag implementation is to enable batching. Alongside this the license provider client has been updated but we can't drop the old client until every license has been migrated.
The drivers license client with `old` prefix will be deleted as soon as the migration is complete.

## Screenshots / Gifs

## Checklist:


brynjarorng (approved):
LGTM

kksteini (approved):
CODEOWNER files LGTM.

I do however think that with the addition of v4 is starting to highlight that the backend is in a need of a rewrite/refactor.

Now there is v1, v2, v4 and v5.

libs/api/domains/license-service/src/lib/client/driving-license-client/drivingLicenseMappers.ts
```diff
@@ -7,22 +6,110 @@ import {
...
+    {
+      identifier: 'rettindaflokkar',
+      value: license.categories
+        ? license.categories?.reduce((acc, curr) => `${acc} ${curr.nr}`, '')
+        : '',
+    },
+    {
+      identifier: 'rettindi',
+      value: license.categories
+        ? license.categories?.reduce(
```
thordurhhh, line 71: Could we make this more readable?

eirikurn (approved):
Approving core files 👍

PR #11270, opened 2023-05-24
feat(auth-admin): Require unique clientId on create

GunnlaugurG:
# ...

## What

Feature to stop creation of new clients when the client id already exists, even though it has been archived

## Why

So there will be no clients with same primary key

## Screenshots / Gifs

## Checklist:


libs/api/domains/auth-admin/src/lib/client/clients.service.ts
```diff
@@ -208,6 +221,7 @@ export class ClientsService extends MultiEnvironmentService {
       ?.meClientsControllerFindByTenantIdAndClientId({
         tenantId: input.tenantId,
         clientId: input.clientId,
+        checkArchive: false,
```
snaerseljan, line 224: This should be an optional check? Right?

libs/auth-api-lib/src/lib/clients/admin/admin-clients.service.ts
```diff
@@ -102,12 +102,14 @@ export class AdminClientsService {
...
     tenantId: string,
     clientId: string,
+    checkArchived = false,
   ): Promise<AdminClientDto> {
     const client = await this.clientModel.findOne({
       where: {
         clientId,
         domainName: tenantId,
         enabled: true,
+        ...(checkArchived && { archived: null }),
```
snaerseljan, line 112: Shouldn't this be changed to the following instead

- When checkArchived true
```
{
  enabled: true,
}
```

- When checkArchived false:
```
{
  enabled: true,
  archived: null
}
```

libs/portals/admin/ids-admin/src/lib/messages.ts
```diff
@@ -778,4 +778,8 @@ export const m = defineMessages({
     defaultMessage:
       'Should personal representatives automatically get this scope for their clients',
   },
+  clientIdAlreadyExists: {
+    id: 'ap.ids-admin:client-id-already-exists',
+    defaultMessage: 'ClientId already exists',
```
snaerseljan, line 783: Just to match name on label

```suggestion
    defaultMessage: 'Client ID already exists',
```

libs/portals/admin/ids-admin/src/screens/Client/CreateClient/CreateClient.action.ts
```diff
@@ -69,6 +70,36 @@ export const createClientAction: WrappedActionFn = ({ client }) => async ({
   }
 
   const { data } = result
+
+  let idAvailableOnAllEnvs = false
```
snaerseljan, line 74: I think you should go at this another way. I would skip this availability check on the client and make the check within the `createClient()` in the `clients.service.ts`. That way you don't have to go back and forth with requests and you do not have to provide special logic in the graphql layer for the archived check, that would just happen within the service.

snaerseljan (commented):
Nicely done 👍  
I had some comments that I think you should consider. Especially regarding the check availability. I think you should use another approach

libs/auth-api-lib/src/lib/clients/admin/admin-clients.service.ts
GunnlaugurG: No I dont think so, the `enabled: true` was put there to fetch only those that have not been deleted, then we realised that `archived` was controlling if it had been deleted

libs/portals/admin/ids-admin/src/screens/Client/CreateClient/CreateClient.action.ts
GunnlaugurG: I had this discussion with @saevarma and we decided on this approach. I ran into some issues with the approach that you are suggesting where the client could not determine why the request was failing, if it was because of a duplicate id or because of some other issue or what environment was causing the issue. I know that there are workarounds for this, but at that time with a little time left in the sprint we decided on this approach.

snaerseljan (approved):
Great job @GunnlaugurG 👍

snaerseljan (approved):
Great job @GunnlaugurG 👍

PR #11271, opened 2023-05-24
feat(auth-admin): Default environment to highest available

GunnlaugurG:
# ...

## What

Default client will be from the highest available environment
_Hirarchy: **Development -> Staging -> Production**_
So if a client is opened that exists in Production and Staging, the default environment will be Production

## Why

So users will default to the environment they are using

## Screenshots / Gifs

## Checklist:


snaerseljan (approved):
Excellento!

PR #11262, opened 2023-05-24
chore(core): Upgrade react-router-dom

saevarma:
## What

Upgrade `react-router-dom` to `6.11.2`

## Why

To be able to use `useFetcher` without needing to specify basepath of the href.

## Screenshots / Gifs

N/A

## Checklist:

- [ ] I have performed a self-review of my own code
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] Formatting passes locally with my changes
- [ ] I have rebased against main before asking for a review

PR #11272, opened 2023-05-24
feat(ids-admin): Add permission dropdown

saevarma:
## What

[Adds tenants dropdown in add permissions modal](https://www.notion.so/Allow-super-user-to-add-cross-tenant-permission-to-application-401e727f23764714b460964295d8ba5d)

Depends on #11262 and needs to be merged after.

3 known bugs I would like a second 👀 on:
* Modal padding cuts of the bottom of the dropdown
* Double filter on the availableScopes
* Empty state if tenant doesn't have scopes.

## Why

So normal users can grant client permissions from other tenants that they own. 
So super users can grant client permissions from any tenant.

## Screenshots / Gifs

image

image

image

## Checklist:


libs/portals/admin/ids-admin/src/screens/Client/components/AddPermissions/AddPermissions.tsx
```diff
@@ -0,0 +1,243 @@
...
+  useGetAvailableScopesLazyQuery,
+} from './AddPermissions.generated'
+
+import * as styles from './AddPermissions.css'
+
+type AuthAdminScopes = GetAvailableScopesQuery['authAdminScopes']['data'][0]['environments']
+
+interface AddPermissionsProps {
+  isVisible: boolean
+  onClose: () => void
```
snaerseljan, line 36: nitnit: TS funcs can be shorten on onClose and onAdd

libs/portals/admin/ids-admin/src/screens/Client/components/AddPermissions/AddPermissions.tsx
```diff
@@ -0,0 +1,243 @@
...
+  useEffect(() => {
+    if (fetcher.state === 'idle' && !fetcher.data) {
+      fetcher.load(IDSAdminPaths.IDSAdmin)
+    }
+  }, [fetcher])
+
+  useEffect(() => {
+    if (fetcher.data) {
+      // When fetcher has loaded data we filter available tenants for the selected environment
+      const availableTenants = (fetcher.data as AuthTenants).filter((tenant) =>
```
snaerseljan, line 85: nitnit: Maybe it is a little cleaner to store fetcher.data in a var below fetcher.

```
  const fetcher = useFetcher<AuthTenants | undefined>()
  const tenants = fetcher.data
```

snaerseljan (commented):
Really nice job!

Some nits and one logic error

libs/portals/admin/ids-admin/src/screens/Client/components/AddPermissions/AddPermissions.tsx
```diff
@@ -0,0 +1,243 @@
...
+    } else {
+      selected.set(value.name, value)
+    }
+    setSelected(new Map(selected))
+  }
+
+  const handleTenantChange = (opt: Option) => {
+    setSelectedTenant(opt)
+
+    getAvailableScopesQuery({
```
snaerseljan, line 155: nit: since this query fn is called twice you could extract it into a helper fn. Something like
```
  const fetchAvailableScopes = (tenantId: string) =>
    getAvailableScopesQuery({
      variables: {
        input: {
          tenantId,
        },
      },
    })
```

libs/portals/admin/ids-admin/src/screens/Client/components/AddPermissions/AddPermissions.tsx
```diff
@@ -0,0 +1,243 @@
...
+          onChange={(opt) => handleTenantChange(opt as Option)}
+          value={selectedTenant}
+          options={availableTenants.map((tenant) =>
+            formatOption(tenant, locale),
+          )}
+        />
+      </Box>
+
+      <ShadowBox flexShrink={1} overflow="auto">
+        <T.Table>
```
snaerseljan, line 196: Regarding modal dropdown cut off, I think an empty state or at lease min height on the table would fix it or at least make it better
```
        <T.Table style={{ minHeight: 300 }}>
```

libs/portals/admin/ids-admin/src/screens/Client/components/AddPermissions/AddPermissions.tsx
```diff
@@ -0,0 +1,243 @@
...
+              ),
+          ),
+      )
+    }
+  }, [authAdminScopes?.data])
+
+  // Add the selected scopes to the addedScopes array for
+  const handleAdd = () => {
+    // Add the new scopes to the permissions array
+    onAdd([...selected.values()])
```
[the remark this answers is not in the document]
saevarma: No 🤔 if the scope is selected and added it will be filtered out from availableScopes so you won't see it again to check it.

libs/portals/admin/ids-admin/src/screens/Client/components/AddPermissions/AddPermissions.tsx
```diff
@@ -0,0 +1,243 @@
...
+  }
+
+  // Handle checkbox change
+  const onChange = (value: AuthAdminClientAllowedScope) => {
+    if (selected.has(value.name)) {
+      selected.delete(value.name)
+    } else {
+      selected.set(value.name, value)
+    }
+    setSelected(new Map(selected))
```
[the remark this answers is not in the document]
saevarma: The new map is needed to trigger a rerender. But I refactored this to use `useRef` instead of `useState` as selecting scope doesn't need to trigger a rerender.

libs/portals/admin/ids-admin/src/screens/Client/components/AddPermissions/AddPermissions.tsx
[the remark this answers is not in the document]
saevarma: But there is another error, if you change tenants then the checkbox is resetted but the selected scopes not.

snaerseljan (approved):
Fantastic job. Really like the new changes 💯

PR #11282, opened 2023-05-25
fix(j-s): Adding check where route should go in form in court of appeal

MargretFinnboga:
# Adding check where route should go in form in court of appeal

[Landsréttur á að byrja á skrefinu sem það var komið á](https://app.asana.com/0/1199153462262248/1204522261372986)

## What

- Adding check which route to go to when filling form of court of appeal
- Adding stepper to investigation case

## Why

- Both needed for better UX

## Checklist:


gudjong (approved):
Looks good to me

PR #11295, opened 2023-05-26
fix(j-s): Disabling rows in case list in court of appeal

MargretFinnboga:
# Disabling certain rows in case list in court of appeal

[Málalisti Landsréttur - Röðun í síu falin](https://app.asana.com/0/1199153462262248/1204559062187683)

## What

- disabling rows to not be sortable

## Why

- rows should not be sortable

## Checklist:

### DIFF
--- a/apps/judicial-system/web/src/components/Table/Table.tsx
+++ b/apps/judicial-system/web/src/components/Table/Table.tsx
@@ -159,7 +159,7 @@ const Table = <T extends object>(
                     'Header',
                   )} `}</Text>
                   <span>
-                    {props.sortableColumnIds?.includes(column.id) ? (
+                    {!column.disableSortBy ? (
                       column.isSorted ? (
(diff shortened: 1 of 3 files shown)


apps/judicial-system/web/src/components/Table/Table.tsx
```diff
@@ -159,7 +159,7 @@ const Table = <T extends object>(
                     'Header',
                   )} `}</Text>
                   <span>
-                    {props.sortableColumnIds?.includes(column.id) ? (
+                    {!column.disableSortBy ? (
```
gudjong, line 162: `sortableColumnIds` is used by `PassedCases` and this change overrides that use. It would be better to settle for one way to define sortable columns and either use `sortableColumnIds` or `disableSortBy`, not both.

gudjong (changes requested):

PR #11298, opened 2023-05-30
feat(j-s): Appeal in Court

gudjong:
# Appeal in Court

[Landsréttur - kært í þinghaldi - yfirlitsskjár Héraðsdóms](https://app.asana.com/0/1199153462262248/1204512772398820/f)

## What

- Sets a case's appeal state to APPEALED when completed in district court if either the prosecutor or defendant appealed in court.
- Fixes link in page header on appeals court pages.

## Why

- So the appeals court can handle cases that are appealed in court.

## Checklist:


apps/judicial-system/web/src/routes/CourtOfAppeal/components/CaseOverviewHeader/CaseOverviewHeader.strings.ts
```diff
@@ -1,11 +1,19 @@
...
     defaultMessage:
       'Kært af {appealedByProsecutor, select, true {sækjanda} other {verjanda}} {appealedDate}',
     description:
       'Notaður sem texti fyrir "Kært af" á yfirlitsskjá mála hjá Landsrétti.',
   },
+  appealedByInCourt: {
+    id:
+      'judicial.system.core:court_of_appeal.case_overview_header.appealed_by_in_court',
+    defaultMessage:
+      '{appealedByProsecutor, select, true {Sóknaraðili} other {Varnaraðili}} kærði í þinghaldi',
```
unakb, line 15: In the other string we say sækjandi and here we say sóknaraðili (and verjanda vs. varnaraðila), why do we use different titles for the same person?

apps/judicial-system/backend/src/app/modules/case/case.controller.ts
```diff
@@ -231,9 +232,33 @@ export class CaseController {
...
+      } else {
+        update.accusedPostponedAppealDate = nowFactory()
+      }
+
+      const autoStates = transitionCase(
+        CaseTransition.APPEAL,
+        states.state ?? theCase.state,
+        states.appealState ?? theCase.appealState,
+      )
+      update.state = autoStates.state
```
unakb, line 260: Do we need to do this here? transitionCase shouldn't change the case state when called with APPEAL, and we've already set the new case state in line 223.

unakb (approved):
Just the 2 questions, otherwise looks good

apps/judicial-system/backend/src/app/modules/case/case.controller.ts
gudjong: I did not want to depend on knowledge about the transition method, it may change in the future.

PR #11299, opened 2023-05-30
feat(consultation-portal): Kam 1329 fe change graphql to reflect on changes in millilagi

karlarnar:
# FE: Change graphql to reflect on changes in millilagi

https://veflausnir.atlassian.net/browse/KAM-1329
https://veflausnir.atlassian.net/browse/KAM-1326
https://veflausnir.atlassian.net/browse/KAM-1327

## What

- Change graphql queries for changes in api
- Translation strings for texts in Case.json
- Make a share-able component called BlowoutList for use for both stakeholders and relatedCases boxes and in turn removing the stakeholders component
- If advice is private, then it should not show the advice
- User can select if he wants the advice to be private, but only if the case allows advices to be private

## Why

## Screenshots / Gifs

image
image
image

## Checklist:

### DIFF
--- a/libs/application/templates/european-health-insurance-card/src/fields/ReviewScreen/ReviewScreen.tsx
+++ b/libs/application/templates/european-health-insurance-card/src/fields/ReviewScreen/ReviewScreen.tsx
@@ -1,4 +1,4 @@
-import { Address, NationalRegistry, Answer } from '../../lib/types'
+import { Address, Answer, NationalRegistry } from '../../lib/types'
 import { Box, Divider, Stack, Text } from '@island.is/island-ui/core'
 
 import { FC } from 'react'
@@ -51,7 +51,7 @@ const ReviewScreen: FC<FieldBaseProps> = ({ application }) => {
               <Box>
                 <Text variant="h5">
                   {formatText(
-                    'Umsækjendur sem vilja fá tímabundið bráðabirgðakort',
+                    e.review.sectionPersonsWhoWantPDFLabel,
                     application,
                     formatMessage,
                   )}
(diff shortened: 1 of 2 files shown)


apps/consultation-portal/screens/Case/Case.json
```diff
@@ -171,5 +177,9 @@
       "text": "Umsagnarfrestur er liðinn.",
       "textCont": "Niðurstöður hafa verið birtar og samráði lokið."
     }
+  },
+  "adviceCard": {
+    "hiddenName": "",
```
jonagnar, line 182: tómt, á kannski að vera þannig?

apps/consultation-portal/screens/Case/Case.tsx
```diff
@@ -133,9 +134,12 @@ const CaseScreen = ({ chosenCase, caseId }: Props) => {
             order={[2, 2, 2, 3, 3]}
           >
             <Stack space={3}>
-              {!isMobile && <CaseStatusBox status={chosenCase.statusName} />}
+              <CaseStatusBox status={chosenCase.statusName} />
               {chosenCase?.stakeholders?.length > 0 && (
-                <Stakeholders chosenCase={chosenCase} />
+                <BlowoutList list={chosenCase.stakeholders} isStakeholder />
+              )}
+              {chosenCase?.relatedCases?.length > 0 && (
```
jonagnar, line 141: Best practice advice: setja segðir í vel nefndar breytur og reference-a svo breytuna í JSX.
t.d.
const isStakeholdersNotEmpty = chosenCase?.stakeholders?.length > 0;
etc.

apps/consultation-portal/screens/Case/components/index.ts
jonagnar, line 1: The curse of export default 🧙‍♂️

apps/consultation-portal/types/interfaces.ts
jonagnar, line 1: afhverju alltaf interface 😿

apps/consultation-portal/screens/Case/Case.json
korsbacka: nei, meikar meiri sense að taka það út og birta ekki neitt

apps/consultation-portal/screens/Case/Case.tsx
korsbacka: Miklu meira lesanlegri þannig, útfæri það þannig :)

PR #11312, opened 2023-05-31
feat(application-system): Update create charge flow

obmagnusson:
## What

Refactor create charge flow. Prevent multiple charges per application. Better UX

## Why

Multiple charges per application where being created if communiction with fjs was shaky og something came up during create charge. 

## Checklist:


libs/application/api/payment/src/lib/payment.service.ts
```diff
@@ -162,58 +185,123 @@ export class PaymentService {
...
-      }
-    } catch (e) {
-      this.logger.error('Error creating charge', e)
-      throw new InternalServerErrorException('Error creating charge')
+  async createNewCharge(
+    paymentModel: Payment,
+    user: User,
+    extraData: ExtraData[] | undefined,
+  ): Promise<ChargeResponse> {
+    // throw new Error('Method not implemented.')
```
Thorrik58, line 276: remove?

libs/application/api/payment/src/lib/payment.service.ts
```diff
@@ -162,58 +185,123 @@ export class PaymentService {
...
+      if (chargeStatus === null) {
+        //No charge present in FJS - we need to create a new one
+        const chargeResult = await this.createNewCharge(
+          paymentModel,
+          user,
+          extraData,
+        )
+        user4 = chargeResult.user4
+      } else if (
+        chargeStatus.statusResult.status ===
```
Thorrik58, line 214: readability could be improved by having all these payment status checks in a function which would return the correct status

Thorrik58 (approved):

PR #11318, opened 2023-05-31
fix(j-s): Adding appeal state in all case list tables 

MargretFinnboga:
# Adding appeal state in all case list tables 

[Kært / Móttekið tögg á málalista eftir að mál eru kærð til Landsréttar](https://app.asana.com/0/1199153462262248/1204724403296599)

## What

- Update tag appeal ruling to state make the component handle all appeal states and ruling 
- Adding tag appeal state to all case list 

## Why

- Added more tags and for more clean code 

## Screenshots / Gifs
Screenshot 2023-05-31 at 13 37 10

## Checklist:


apps/judicial-system/web/src/components/TagAppealState/TagAppealState.tsx
```diff
@@ -0,0 +1,102 @@
...
+  appealRulingDecision?: CaseAppealRulingDecision
+}
+
+const TagAppealState: React.FC<Props> = ({
+  appealRulingDecision,
+  appealState,
+}) => {
+  const { formatMessage } = useIntl()
+
+  if (!appealRulingDecision) return null
```
gudjong, line 24: This means that a tag is only shown when an appeal ruling decision has been entered. Newly appealed and received appeals will not get a tag. The code will work if this line is removed.

gudjong (changes requested):

apps/judicial-system/web/src/components/TagAppealState/TagAppealState.tsx
MargretFinnboga: You are right I just forgot to remove this! 🙈

gudjong (approved):

PR #11324, opened 2023-06-01
fix(j-s): Appeal in Court

gudjong:
# Appeal in Court

[Landsréttur - kært í þinghaldi - yfirlitsskjár Héraðsdóms](https://app.asana.com/0/1199153462262248/1204512772398820/f)

## What

- Fixes state transition when appealing in court.
- Allows case completion after case has been appealed.
- Appeal transition need further modifications to allow the reopening of appealed cases. That will be handled separately.

## Why

- Verified bug.

## Checklist:


MargretFinnboga (approved):
Nice 🤝

PR #11325, opened 2023-06-01
feat(devx): Add 'yarn clean'

AndesKrrrrrrrrrrr:
## What
Add single command to get to a clean project.

## Why
Resetting is a semi-frequent action. Having a simple way to reliably get a clean repo should speed up troubleshooting.

## Checklist:

### DIFF
--- a/apps/consultation-portal/screens/Case/Case.json
+++ b/apps/consultation-portal/screens/Case/Case.json
@@ -67,7 +67,7 @@
       "button": "Skrá mig inn"
     },
     "setEmailCardSkeleton": {
-      "text": "Skráðu netfang hérna. Þú færð svo tölvupóst sem þú þarf að staðfesta.",
+      "text": "Skráðu netfang hérna. Þú færð svo tölvupóst sem þú þarft að staðfesta.",
       "input": {
         "label": "Netfang",
         "placeholder": "nonni@island.is"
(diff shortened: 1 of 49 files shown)


package.json
```diff
@@ -21,6 +21,7 @@
     "deploy": "nx deploy",
     "test": "nx test",
     "lint": "nx workspace-lint && nx lint",
+    "clean": "./scripts/clean-generated-files.sh && rm -rf .cache",
```
eirikurn, line 24: Should also remove:

- `node_modules`?
- `node_modules/.cache`
- `.yarn/cache`?
- `dist`?

eirikurn (approved):
Approving with ideas for additions 👍

package.json
AndesKrrrrrrrrrrr: Moved to own script to allow cleaning even if yarn is broken.

Also, added your ideas.

brynjarorng (changes requested):
All looks good except for the last error

scripts/clean.sh
```diff
@@ -0,0 +1,124 @@
...
+      $clean_job || log "Job $job failed"
+    else
+      log "Skipping $job"
+    fi
+  done
+}
+
+cli "$@"
+if (return 2>/dev/null); then
+  echo "Exiting without running (only parsed arguments) because you sourced it."
```
brynjarorng, line 120: Could this`because you sourced it.` be changed to something like `because the script was sourced.` to be as friendly as possible to the devs? :)

brynjarorng (approved):
LGTM

PR #11332, opened 2023-06-01
feat(auth-admin): Delete client

GunnlaugurG:
# ...

## What

Allow users to delete client on one or all environments

## Why

So cliets can be removed

## Screenshots / Gifs

## Checklist:

### DIFF
--- a/libs/api/domains/auth-admin/src/lib/client/clients.service.ts
+++ b/libs/api/domains/auth-admin/src/lib/client/clients.service.ts
@@ -21,6 +21,7 @@ import { ClientSecret } from './models/client-secret.model'
 import { Client } from './models/client.model'
 import { ClientAllowedScope } from './models/client-allowed-scope.model'
 import { environments } from '../shared/constants/environments'
+import { DeleteClientInput } from './dto/delete-client.input'
 
 @Injectable()
 export class ClientsService extends MultiEnvironmentService {
@@ -309,6 +310,23 @@ export class ClientsService extends MultiEnvironmentService {
       tenantId: input.tenantId,
       clientId: input.clientId,
     })
+  }
+
+  async deleteClient(user: User, input: DeleteClientInput): Promise<boolean> {
+    try {
+      await this.adminApiByEnvironmentWithAuth(
+        input.environment,
+        user,
+      )?.meClientsControllerDelete({
+        tenantId: input.tenantId,
+        clientId: input.clientId,
+      })
+    } catch (error) {
+      this.handleError(error, input.environment)
+      return false
+    }
+
+    return true
   }
 
   async revokeSecret(user: User, input: RotateSecretInput) {
--- a/libs/api/domains/auth-admin/src/lib/client/dto/delete-client.input.ts
+++ b/libs/api/domains/auth-admin/src/lib/client/dto/delete-client.input.ts
@@ -1,0 +1,14 @@
+import { Field, InputType } from '@nestjs/graphql'
+import { Environment } from '@island.is/shared/types'
+
+@InputType('AuthAdminDeleteClientInput')
+export class DeleteClientInput {
+  @Field(() => String, { nullable: false })
+  tenantId!: string
+
+  @Field(() => String, { nullable: false })
+  clientId!: string
+
+  @Field(() => Environment, { nullable: false })
+  environment!: Environment
+}
--- a/libs/auth-api-lib/src/lib/clients/admin/admin-clients.service.ts
+++ b/libs/auth-api-lib/src/lib/clients/admin/admin-clients.service.ts
@@ -2,6 +2,7 @@ import {
   BadRequestException,
   ForbiddenException,
   Injectable,
+  NotFoundException,
 } from '@nestjs/common'
 import { InjectModel } from '@nestjs/sequelize'
 import { Includeable, Op, Transaction } from 'sequelize'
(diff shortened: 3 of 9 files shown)


libs/portals/admin/ids-admin/src/lib/messages.ts
```diff
@@ -778,4 +778,28 @@ export const m = defineMessages({
     defaultMessage:
       'Should personal representatives automatically get this scope for their clients',
   },
+  successDeletingClient: {
+    id: 'ap.ids-admin:success-deleting-client',
+    defaultMessage: 'Successfully deleted client',
```
saevarma, line 783:
```suggestion
    defaultMessage: 'Successfully deleted application',
```

libs/portals/admin/ids-admin/src/lib/messages.ts
```diff
@@ -778,4 +778,28 @@ export const m = defineMessages({
...
+  },
+  deleteClient: {
+    id: 'ap.ids-admin:delete-client',
+    defaultMessage: 'Delete client in {environment} environment',
```
saevarma, line 787:
```suggestion
    defaultMessage: 'Delete application in {environment} environment',
```

libs/portals/admin/ids-admin/src/lib/messages.ts
```diff
@@ -778,4 +778,28 @@ export const m = defineMessages({
...
+    id: 'ap.ids-admin:close-delete-modal',
+    defaultMessage: 'Close',
+  },
+  deleteClientDescription: {
+    id: 'ap.ids-admin:delete-client-action-card-label',
+    defaultMessage: 'All your apps using this client will stop working.',
+  },
+  deleteClientAllEnv: {
+    id: 'ap.ids-admin:delete-client-all-env',
+    defaultMessage: 'Delete client in all environments',
```
saevarma, line 803:
```suggestion
    defaultMessage: 'Delete application in all environments',
```

libs/portals/admin/ids-admin/src/lib/messages.ts
```diff
@@ -778,4 +778,28 @@ export const m = defineMessages({
...
+    id: 'ap.ids-admin:delete',
+    defaultMessage: 'Delete',
+  },
+  closeDeleteModal: {
+    id: 'ap.ids-admin:close-delete-modal',
+    defaultMessage: 'Close',
+  },
+  deleteClientDescription: {
+    id: 'ap.ids-admin:delete-client-action-card-label',
+    defaultMessage: 'All your apps using this client will stop working.',
```
saevarma, line 799: I'm suggesting an update in Figma as we are always talking about application instead of client.
```suggestion
    defaultMessage: 'All your apps using this application ID will stop working.',
```

saevarma (commented):
Overall really good, great improvements in readability and making it cleaner 👏
Just few comments on improvements, and one UX comment on how we should handle the success/error responses.

saevarma (commented):
Daði has updated the design in Figma.
I would change the GQL mutation to always send delete request to all environment available to the GQL, as they should just respond with 204 No Content if the client doesn´t exist or is already archived.

libs/api/domains/auth-admin/src/lib/client/dto/delete-client.input.ts
```diff
@@ -0,0 +1,14 @@
...
+
+@InputType('AuthAdminDeleteClientInput')
+export class DeleteClientInput {
+  @Field(() => String)
+  tenantId!: string
+
+  @Field(() => String)
+  clientId!: string
+
+  @Field(() => [Environment], { nullable: false })
```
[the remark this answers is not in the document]
GunnlaugurG: yes, my bad :D

libs/auth-api-lib/src/lib/clients/admin/admin-clients.service.ts
```diff
@@ -203,6 +204,34 @@ export class AdminClientsService {
...
+    const client = await this.clientModel.findOne({
+      where: {
+        clientId,
+        domainName: tenantId,
+      },
+    })
+
+    if (!client || client.archived) {
+      throw new NotFoundException()
+    }
```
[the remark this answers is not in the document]
GunnlaugurG: yes, my bad :D

libs/api/domains/auth-admin/src/lib/client/clients.service.ts
```diff
@@ -313,30 +312,21 @@ export class ClientsService extends MultiEnvironmentService {
...
-      environment,
-      success: true,
-    }))
-
+  async deleteClient(user: User, input: DeleteClientInput): Promise<boolean> {
     await Promise.all(
-      response.map((resp, index) =>
-        this.adminApiByEnvironmentWithAuth(resp.environment, user)
+      Object.values(Environment).map((resp) =>
+        this.adminApiByEnvironmentWithAuth(resp, user)
```
saevarma, line 318: We have a constant array of environment which you can just use directly instead of `Object.values`
```suggestion
      environments.map((env) =>
        this.adminApiByEnvironmentWithAuth(env, user)
```

libs/api/domains/auth-admin/src/lib/client/clients.service.ts
```diff
@@ -313,30 +312,21 @@ export class ClientsService extends MultiEnvironmentService {
...
           ?.meClientsControllerDelete({
             tenantId: input.tenantId,
             clientId: input.clientId,
           })
           .catch((error) => {
-            response[index].success = false
-            this.handleError(error, resp.environment)
+            this.handleError(error, resp)
```
saevarma, line 324: The `handleError` swallows errors, so this will cause the service to always return true, even though none of the request succeeds. So the service needs to keep track of success of each request, and then check if one of it succeeds then return true otherwise return false.

saevarma (approved):
Great job 🙌

PR #11333, opened 2023-06-02
chore(assets): Increase real estate assets timeout

thordurhhh:
## What

Increase real estate assets timeout

## Why

It was recommended by the service provider to at least keep a 30sec timeout as when their service is getting a lot of traffic it can be a bit slow.

## Checklist:


brynjarorng (approved):
LGTM

Tryggvig (approved):

PR #11339, opened 2023-06-02
fix(j-s): Enables Case Correction After Appeal

gudjong:
# Enables Case Correction After Appeal

[Landsréttur - ráða við að leiðrétta mál sem eru í kæruferli](https://app.asana.com/0/1199153462262248/1204723926688693/f)

## What

- Allows the district court to modify the case after it has been appealed.

## Why

- This should be possible.

## Checklist:


MargretFinnboga (approved):
Consider removing unused import, otherwise looks good 🥇

PR #11337, opened 2023-06-02
feature(financial-statements-inao): Fix bug on error from UserProfileApi

fjandakornid:
# Fix bug on error from UserProfileApi

RES-100

## What

Fix bug on error from UserProfileApi

## Why

Users can not continue with the application if error occurs.

## Checklist:

- [x] I have performed a self-review of my own code
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] Formatting passes locally with my changes
- [x] I have rebased against main before asking for a review

PR #11351, opened 2023-06-05
fix(financial-statements-inao): Fix bug on error from UserProfileApi…

programm-ingovals:
… (#11337)

Issue when entering an application but missing UserProfileInfo


fjandakornid (approved):
LGTM

PR #11369, opened 2023-06-06
feat(j-s): Defender Case List

unakb:
# Defender Case List

[Málalisti fyrir verjendur](https://app.asana.com/0/1199153462262248/1204701548656409/f)

## What

- Add a new case list page for defenders 
- Fix sorting by defendants in the table for past cases
- Refactoring and separating table components into smaller ones

## Why

- So that defenders can see a list of the cases they have been assigned to as defenders
- So that we can sort past cases by defendants
- So that we can easily refactor other tables and reuse table data components that have the same functionalities

## Screenshots / Gifs

image

## Checklist:


apps/judicial-system/web/messages/Core/tables.ts
```diff
@@ -52,4 +52,24 @@ export const tables = defineMessages({
     description:
       'Notaður sem texti sem segir til um hvort mál sé framlenging í málalista.',
   },
+  created: {
```
unakb, line 55: I decided to keep these strings here for now for consistency, and we can then move them closer to a shared tables component when we've landed on how we want to refactor our tables

apps/judicial-system/web/src/routes/Shared/Cases/PastCases.tsx
```diff
@@ -82,6 +84,15 @@ const DurationDate = ({ date }: { date: string | null }) => {
   )
 }
 
+const sortDefendants = (rowA: Row<Case>, rowB: Row<Case>) => {
```
unakb, line 87: Added this as a quick fix for the defendant sorting, needs more refactoring later

gudjong (approved):
Really nice

PR #11377, opened 2023-06-07
fix(j-s): Change "seen by" to "opened by" defender

unakb:
# Change texts for seen by defender alert

[Laga virkni að verjandi hafi sótt kröfu ](https://app.asana.com/0/1199153462262248/1204456504011227/f)

## What

Change from "seen by defender" to "opened by defender"

## Why

It's more accurate seeing as we only really know whether a defender has opened a case, not whether he has truly seen it.

## Screenshots / Gifs

image

## Checklist:


apps/judicial-system/backend/src/app/modules/case/limitedAccessCase.service.ts
```diff
@@ -189,9 +189,9 @@ export class LimitedAccessCaseService {
       throw new NotFoundException(`Case ${caseId} does not exist`)
     }
 
-    if (!theCase.seenByDefender) {
+    if (!theCase.openedByDefender) {
```
gudjong, line 192: Please consider moving this update. The original issue that lead to the accompanying Asana ticket was that this field was set even if subsequent guard tests failed and the defender was never allowed to see the case. This is still true, but less likely because the `findById` method is no longer called during defender login.

It would probably be safer to set the `openedByDefender` field in the method `getById` in `LimitedAccessCaseController` just before the case is returned.

gudjong (approved):
One suggestion

apps/judicial-system/backend/src/app/modules/case/limitedAccessCase.service.ts
unakb: Fixed, could you take another look to make sure I understood correctly 😊?

apps/judicial-system/backend/src/app/modules/case/limitedAccessCase.service.ts
```diff
@@ -268,7 +265,9 @@ export class LimitedAccessCaseService {
       })
     }
 
-    await this.messageService.sendMessagesToQueue(messages)
+    if (messages.length > 0) {
```
gudjong, line 268: Nice.

gudjong (approved):

PR #11379, opened 2023-06-07
test(system-e2e): Ids admin e2e tests

snaerseljan:
# E2E tests

## What

Ids admin tests for tenants, permissions, permission, clients and client

## Checklist:


apps/system-e2e/src/utils/pageHelpers.ts
```diff
@@ -0,0 +1,2 @@
+export const getInputByName = (name: string) => `input[name="${name}"]`
+export const getTextareaByName = (name: string) => `textarea[name="${name}"]`
```
AndesKrrrrrrrrrrr, line 2: Should be using role selectors
```
page.getByRole('textbox', {name: 'my_name'})
```

AndesKrrrrrrrrrrr (commented):
Lovely tests :heart:

If possible, I'd suggest using `dataTestId` less, and opting for `getBy*` selectors. If it's hard to use `getBy*`, it's probably also hard for accessibility tools.

Toti91 (approved):
UI code LGTM

apps/system-e2e/src/utils/pageHelpers.ts
snaerseljan: Thanks for the review 👍 

Yes that is what I started to do initially. However the e2e would just hang and fails by doing it that way. I am not sure what the reason is.
I created a new branch with updated code to use `getByRole` method. If you know what could be causing the problem I am all ears.
[ids-admin-e2e-update](https://github.com/island-is/island.is/tree/ids-admin-e2e-update)

Otherwise, this is a minor thing and the other code works well.

AndesKrrrrrrrrrrr (approved):

PR #11415, opened 2023-06-09
feat(application-estate): Generate PDF of form data to send to DC

kksteini:
District Commissioner asking for PDFs to be generated.
Example:
[estate.pdf](https://github.com/island-is/island.is/files/11706304/estate.pdf)

Note, these pdfs are there by request as backup to the form data. They are not shown to the applicant at any point and therefore have no requirements for appearence/look.


libs/application/template-api-modules/src/lib/modules/templates/estate/estate.service.ts
```diff
@@ -1,4 +1,5 @@
 import { Injectable } from '@nestjs/common'
+import fs from 'fs'
```
kksteini, line 2:
```suggestion
```

rafnarnason (approved):
lgtm 🌮

PR #11426, opened 2023-06-12
feat(ids-api): Add endpoint for sending sms.

valurefugl:
https://www.notion.so/Add-SMS-endpoint-in-ids-api-25669a7ee61a415bb502221aecd72331?pvs=4

## What

Add endpoint for sending sms.

## Why

So we can add two factor authentication to the IDS.

## Screenshots / Gifs

## Checklist:


apps/services/auth/ids-api/src/app/notifications/notifications.controller.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  @Post('sms')
+  @Documentation({
+    description: 'Send a message via SMS.',
+    response: { status: 202 },
+  })
+  sendSms(@Body() message: SmsMessage): Promise<void> {
+    return this.auditService.auditPromise(
+      {
+        system: true,
```
eirikurn, line 36: Should use `@CurrentAuth() auth`?

charts/identity-server/values.prod.yaml
```diff
@@ -437,6 +437,9 @@ services-auth-ids-api:
     CONFIGCAT_SDK_KEY: '/k8s/configcat/CONFIGCAT_SDK_KEY'
     DB_PASS: '/k8s/services-auth/api/DB_PASSWORD'
     IDENTITY_SERVER_CLIENT_SECRET: '/k8s/services-auth/IDENTITY_SERVER_CLIENT_SECRET'
+    NOVA_PASSWORD: '/k8s/gjafakort/NOVA_PASSWORD'
+    NOVA_URL: '/k8s/service-portal-api/NOVA_URL'
+    NOVA_USERNAME: '/k8s/gjafakort/NOVA_USERNAME'
```
eirikurn, line 442: We probably need to create these in the production IDS account. Should create new names while we're at it, can devops to copy the values.

eirikurn (approved):
Approving with comments.

apps/services/auth/ids-api/src/app/notifications/notifications.controller.ts
valurefugl: Fixed.

charts/identity-server/values.prod.yaml
valurefugl: Renamed and new secrets created on staging and ids-prod.

brynjarorng (approved):
LGTM

apps/services/auth/ids-api/src/app/notifications/notifications.controller.ts
valurefugl: Actually not fixed since `@CurrentAuth` returns interface `Auth` and esbuild fails. How would we prefer to handle that?

valurefugl: ..and fixed now.

PR #11453, opened 2023-06-13
feat(summer-of-perf): Connection pooling for website requests

eirikurn:
## What

Configures connection pooling for these requests:

- web (Public Website) -> api (GraphQL API)
- api -> Contentful

The pools are configured like this:

- maxSockets: 50 - opens max 50 sockets per process per integration. This is a documented best practice to not consume excessive resources under load.
- timeout: 20 seconds - going up from the previous default of 10 second timeouts
- freeSocketTimeout: 10 seconds - keeps unused sockets open for 10 seconds, then closes them if they're not needed.

## Why

These two integrations are a source of MANY requests, which can benefit from connection pooling. By reusing connections we save resources and precious milliseconds, especially considering that each Contentful request required setting up a full SSL connection.

There is also a [small sliver of hope](https://community.prismic.io/t/fetch-error-on-rest-api-econnreset/3632/24) that connection pooling and maxSockets can reduce occurrences of ECONNRESET errors in production.

The max socket configuration is based on these references:

- https://github.com/aws/aws-sdk-js-v3/issues/1959#issuecomment-777709813
- https://learn.microsoft.com/en-us/azure/app-service/app-service-web-nodejs-best-practices-and-troubleshoot-guide#my-node-application-is-making-excessive-outbound-calls
- https://medium.com/trendyol-tech/nodejs-performance-best-practices-more-than-cliche-9baa573cbf03#42c8


libs/nest/feature-flags/src/lib/feature-flag.client.ts
```diff
@@ -1,14 +1,22 @@
...
 
-export const FeatureFlagClientProvider = {
+let clientSingleton: ReturnType<typeof createClient>
+
+export const FeatureFlagClientProvider: Provider = {
   provide: FEATURE_FLAG_CLIENT,
   inject: [FeatureFlagConfig.KEY],
   useFactory({ sdkKey }: ConfigType<typeof FeatureFlagConfig>) {
-    return createClient({ sdkKey })
+    // Configcat verbosely complains if you create the same client twice, which we should only be doing in tests.
```
snaerseljan, line 15: ahhh thank you!

snaerseljan (approved):
Really nice job @eirikurn.

Although you might think those connection handshakes take little time, then this definitely adds up and saves us as you said precious milliseconds.

Nice job 💯

siggifv (approved):

PR #11474, opened 2023-06-14
fix: Caching of hashed webpack assets

eirikurn:
## What

Cache JS and CSS bundles in our SPA projects. 

## Why

To improve performance and continue serving old assets after deploying new code.

Some files have 16 char hash while others have 20 char hashes. This is configured in NX's webpack config.


scripts/dockerfile-assets/nginx/default.conf.template
```diff
@@ -14,11 +14,11 @@ server {
         add_header Cache-Control "no-store, no-cache, must-revalidate";
     }
 
-    location ~ "^$BASEPATH(.+\.[0-9a-f]{20}\..+)$" {
+    location ~ "^$BASEPATH(.+\.[0-9a-f]{16,20}\..+)$" {
```
brynjarorng, line 17:
```suggestion
    location ~ "^$BASEPATH(.+\.([0-9a-f]{16}|[0-9a-f]{20})\..+)$" {
```

Would it not then make sense to match either 16 or 20 times instead of taking everything in between as well?

brynjarorng (changes requested):

scripts/dockerfile-assets/nginx/default.conf.template
brynjarorng: Agreed, though I think more specific is almost always better this should likely be fine.

brynjarorng (approved):
LGTM

PR #11479, opened 2023-06-14
fix(service-portal): Don't translate the names in service-portal

thordurhhh:
## What

Block translations of name display in the service portal.
The translation of names is the most important part for now, we might block translation of some additional data later.

https://www.w3.org/International/questions/qa-translate-flag

Additional change: Add translation availability to `<Text>` component

## Why

We have had a few complaints from users using Google Translate that think their name is wrong in Þjóðskrá, so we need to prevent the translations of names.

## Screenshots / Gifs

Gervimaður Færeyjar when using Google Translate (example):
Screenshot 2023-06-14 at 10 50 52

## Checklist:


snaerseljan (approved):
💯 Nice

thorkellmani (approved):
lgtm

PR #11487, opened 2023-06-15
fix(ids-admin): Fixing two bugs in permissions

snaerseljan:
# Fixes two bugs in permissions

## What

https://www.notion.so/Add-permissions-modal-doesn-t-filter-scopes-correctly-after-saving-473c9ded9edc42909d429d75e305dae1?pvs=4

Then I discovered another bug when adding permission and saving deleting it and adding it again. Then the save button was wrongly disabled

## Checklist:


eirikurn (approved):
🙌

PR #11488, opened 2023-06-15
fix(j-s): Feature hide appeal to court of appeals email

unakb:
[Skip sending appeal notifications if case was appealed in court ](https://app.asana.com/0/1199153462262248/1204823130701579/f)

## What

Skip sending appeal notifications if case was appealed in court

## Why

Because everyone involved will already know the case was appealed since they were in court when it happened.

## Checklist:


apps/judicial-system/backend/src/app/modules/case/case.service.ts
```diff
@@ -715,6 +716,14 @@ export class CaseService {
     theCase: Case,
     user: TUser,
   ): Promise<void> {
+    // If case was appealed in court we don't need to send these messages
+    if (
+      theCase.accusedAppealDecision === CaseAppealDecision.APPEAL ||
+      theCase.prosecutorAppealDecision === CaseAppealDecision.ACCEPT
+    ) {
+      return Promise.resolve()
```
gudjong, line 724: If you make the method `addMessagesForAppealedCaseToQueue` async you don't need to explicitely return a promise. You can do a simple return instead.

gudjong: However, this check is done a bit prematurely. We still need to post all the case file delivery messages to queue. The only message that should not be queued is the notification message.

gudjong (changes requested):

gudjong (approved):

PR #11495, opened 2023-06-15
fix(j-s): Updating tables and fixing filtering 

MargretFinnboga:
# Updating tables and slowly removing react-tables 

[Málalisti. Varnaðili raðast ekki í stafrófsröð í afgreiddum málum hjá öllum notendum og kærð mál/afgreidd mál hjá Landsrétti](https://app.asana.com/0/1199153462262248/1204724531400377)

## What

Wanted to start by creating table, multi-use table component and updating a few tables not all, just be sure that they work! Safety first right!?  

- Moved Table component that uses react table to ReactTable component 
- Created Table component with these helper table blocks and tables
- Added to shared cases check if the table is loading before showing error message 
- Added useSort hook that has useSortCases and useSortAppealCases 
 
> H E L P E R S 🔨 
	**TableContainer** -> Because tables have the same style and same loading state 
	**TableSkeleton** -> Used as loading state for tables 
	**TableHeaderText** -> All header text are the same, takes in title 
	**SortButton** -> Button in header that you can sort from 
	**DurationDate** -> this is used in PastCases thought it was cleaner to have it as it own
	**CourtCaseNumber** -> used in table column, makes the tbody more clean and is always the same 
	**ColumnCaseType** -> Used as column to show case type 
	**CateGoryCard** -> Used in mobile version of table 

> T A B L E S 📑 
        Reason for two different types is they take in different props for cases, its just a start, hopefully one day maybe we 
        can unit them 🤷🏼 
	**AppealCasesTable** -> stores the table and mobile table
	**PastCasesTable** -> stores spec test, table and mobile table  

## Why

- React tables wasn't filtering right 

## Screenshots / Gifs

Screenshot 2023-06-15 at 11 55 36

Screenshot 2023-06-15 at 11 27 28

## Checklist:


apps/judicial-system/web/src/components/Table/DurationDate/DurationDate.tsx
```diff
@@ -0,0 +1,48 @@
+import { useIntl } from 'react-intl'
+import parseISO from 'date-fns/parseISO'
+import { tables } from '@island.is/judicial-system-web/messages'
+import { Text } from '@island.is/island-ui/core'
+import { Case, CaseState } from '@island.is/judicial-system/types'
+
+import { formatDate } from '@island.is/judicial-system/formatters'
+
+export function getDurationDate(
+  state: Case['state'],
```
unakb, line 10: How come you decided to infer the types from Case here instead of explicitly defining them?

apps/judicial-system/web/src/routes/Shared/Cases/ActiveCases.tsx
```diff
@@ -42,6 +42,7 @@ import { displayCaseType } from './utils'
 import * as styles from './Cases.css'
 import MobileCase from './MobileCase'
 import { cases as m } from './Cases.strings'
+import ColumnCaseType from '@island.is/judicial-system-web/src/components/Table/ColumnCaseType/ColumnCaseType'
```
unakb, line 45: Consider removing unused displayCaseType import

apps/judicial-system/web/src/utils/hooks/index.ts
```diff
@@ -11,3 +11,4 @@ export { default as useOnceOn } from './useOnceOn'
 export type { CaseFileStatus, CaseFile } from './useCourtUpload'
 export { UploadState, useCourtUpload } from './useCourtUpload'
 export { default as useAppealAlertBanner } from './useAppealAlertBanner'
+export { default as useSortCases } from './useSort/useSortCases'
```
unakb, line 14: consider adding useSortAppealCases here for consistency

apps/judicial-system/web/src/components/Table/SortButton/SortButton.tsx
```diff
@@ -0,0 +1,43 @@
...
+      component="button"
+      display="flex"
+      alignItems="center"
+      className={styles.thButton}
+      onClick={onClick}
+    >
+      <Text fontWeight="regular">{title}</Text>
+      <Box
+        className={cn(styles.sortIcon, {
+          [styles.sortAccusedNameAsc]: sortAsc,
```
unakb, line 29: Nice! Consider renaming the styles for the sortIcon. Since you've extracted the component there's no longer a need for having separate style names for each one, so they can now be .sortAsc and sortDes

apps/judicial-system/web/src/components/Table/PastCasesTable/PastCasesTable.tsx
```diff
@@ -0,0 +1,161 @@
...
+          <TableHeaderText title={formatMessage(tables.caseNumber)} />
+          <th className={cn(styles.th, styles.largeColumn)}>
+            <SortButton
+              title={capitalize(formatMessage(core.defendant, { suffix: 'i' }))}
+              onClick={() => requestSort('defendant')}
+              sortAsc={getClassNamesFor('defendant') === 'ascending'}
+              sortDes={getClassNamesFor('defendant') === 'descending'}
+            />
+          </th>
+          <TableHeaderText title={formatMessage(tables.type)} />
```
unakb, line 103: In the current past cases table we can sort cases by type, after this we'll only be able to sort by defendant. I'm actually not sure which is right though as the active cases table can't be sorted by type. Should we check with Guðný if it's ok that we remove this sorting, or did you already talk to her?
It does feel like a weird sorting by type to me since we also have the dropdown filter at the top, but there might be a reason for it being sortable as well that I'm not aware of

apps/judicial-system/web/src/utils/hooks/useSort/useSortAppealCases.ts
```diff
@@ -0,0 +1,58 @@
+import { AppealedCasesQueryResponse } from '@island.is/judicial-system-web/src/routes/CourtOfAppeal/Cases/Cases'
+import { useState, useMemo } from 'react'
+
+const useSortAppealCases = (
+  defaultColumn: string,
+  defaultDirection: 'ascending' | 'descending',
+  data: AppealedCasesQueryResponse[],
```
unakb, line 7: I wonder about the abstraction of the AppealedCasesQueryResponse[] vs CaseListEntry[], they're almost exactly the same, and it would be nice to use the same sort hook for all tables. But that's another refactor and another conversation so I think we should maybe just keep that for another PR 😊

unakb (approved):
Really nice refactoring, much easier to work with the components broken down like this 👏 There's just a few nitpicks and a question that would be good to figure out before merging 🙏

PR #11498, opened 2023-06-15
fix(driving-instructor-registrations): DLP requirements changes and client update

kksteini:
# Requirements change
An instructor can allow a student to have their practice permit confirmed if they deem that student to be ready instead of requiring 10 lessons explicitly. This now simply hinges upon `Ökuskóli 1` being completed successfully.

## Video

In the video, the Gervimaður starts with Ö1 but in status 9 (canceled) and as such it does not show up.
Then we call the API such that an entry for Ö1 is created, subsequently fulfilling the requirements.
Then the driving instructors issues the permit.

[Screencast from fim 15.jún 2023 14:46:25.webm](https://github.com/island-is/island.is/assets/77672665/a606b8b4-24cd-4b41-bf4c-138b86a24d17)

## EDIT
Samgöngustofa maintains that the client will not include the status code by the time this would be hotfixed/released.
As such, I've removed the status checks for Ökuskóli 1 that were previously present in this PR

## Checklist:

### DIFF
--- a/apps/judicial-system/backend/src/app/modules/case/case.service.ts
+++ b/apps/judicial-system/backend/src/app/modules/case/case.service.ts
@@ -837,18 +837,19 @@ export class CaseService {
       }
     }
 
+    if (
+      updatedCase.prosecutorStatementDate?.getTime() !==
+      theCase.prosecutorStatementDate?.getTime()
+    ) {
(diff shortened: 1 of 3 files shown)


rafnarnason (approved):
lgtm 🌭

PR #11512, opened 2023-06-16
feat(endorsement-system): NationalRegV2 update and municipality and filtering

rafnarnason:
Upgrades to NationalRegV2 to get endorser Address Data
So that endorser NationalReg Data is stored in DB
and filtered NationalReg data based on display rules
commented out one test which is not mocked and erroring on trying to call NationalRegV2 - new test suite in next PR

## What

## Why

## Screenshots / Gifs

## Checklist:

### DIFF
--- a/apps/services/endorsements/api/src/app/modules/endorsement/endorsement.service.ts
+++ b/apps/services/endorsements/api/src/app/modules/endorsement/endorsement.service.ts
@@ -21,7 +21,7 @@ import { REQUEST } from '@nestjs/core'
 import { Request } from 'express'
 
 
-export interface EndorsementRequest extends Request {
+interface EndorsementRequest extends Request {
   auth: {
     nationalId: string
   }
@@ -74,13 +74,6 @@ export class EndorsementService {
   async findEndorsements({ listId }: FindEndorsementsInput, query: any) {
     this.logger.info(`Finding endorsements by list id "${listId}"`)
 
-
-    ////////////// 
-
-    /////////////
-
-
-
     const endorsements = await paginate({
       Model: this.endorsementModel,
       limit: query.limit || 10,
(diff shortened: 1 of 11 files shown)


apps/services/endorsements/api/src/app/modules/endorsement/endorsement.service.ts
```diff
@@ -68,6 +80,33 @@ export class EndorsementService {
       orderOption: [['counter', 'DESC']],
       where: { endorsementListId: listId },
     })
+
+    // endorsement data display rules for everyone(public,users,admins) except owner ...
+    // ... list owner sees all data
+    const user = this.request.auth as User
+    const listOwnerNationalId = await this.getListOwnerNationalId(listId)
+    if (user?.nationalId != listOwnerNationalId) {
```
alexdiljar, line 88: Should this be in the interceptor files, where we are already masking nationalIds?

apps/services/endorsements/api/src/app/modules/endorsement/endorsement.service.ts
```diff
@@ -68,6 +80,33 @@ export class EndorsementService {
...
+        if (!endorsement.meta.showName) {
+          endorsement.meta.fullName = ''
+          endorsement.meta.locality = ''
+        }
+      })
+    }
+    return endorsements
+  }
+
+  async getListOwnerNationalId(listId: string): Promise<string | null> {
```
alexdiljar, line 99: Would it be possible to user the getListOwner function in the endorsementList service that already exists}

apps/services/endorsements/api/src/app/modules/endorsement/endorsement.service.ts
rafnarnason: good call - I move it to the interceptor

PR #11521, opened 2023-06-19
feat(j-s): Get police case info

unakb:
[Búa til nýtt function í police service sem skilar upplýsingum um mál](https://app.asana.com/0/1199153462262248/1204823390153048/f)

## What

Retrieve relevant information from police cases

## Why

So that we can use information that's already been saved to the police system instead of users having to input it again

## Checklist:


apps/judicial-system/backend/src/app/modules/police/models/policeCaseInfo.model.ts
```diff
@@ -0,0 +1,10 @@
+import { ApiProperty } from '@nestjs/swagger'
+
+export class PoliceCaseInfo {
+  @ApiProperty()
+  caseNumber!: string
```
gudjong, line 5: Please rename to `policeCaseNumber` to be consistant with other models.

apps/judicial-system/backend/src/app/modules/police/models/policeCaseInfo.model.ts
```diff
@@ -0,0 +1,10 @@
...
+  @ApiProperty()
+  crimeScene?: string
+  @ApiProperty()
+  crimeDate?: Date
```
gudjong, line 9: Please see comments to the corresponding model in the api project.

apps/judicial-system/backend/src/app/modules/police/police.service.ts
```diff
@@ -247,6 +248,94 @@ export class PoliceService {
...
+    const promise = this.fetchPoliceDocumentApi(
+      `${this.xRoadPath}/V2/GetDocumentListById/${caseId}`,
+    )
+
+    return promise
+      .then(async (res: Response) => {
+        if (res.ok) {
+          const response = await res
+            .json()
+            .then((result) =>
```
gudjong, line 264: Please consider removing this conditional second JSON parse. The api has been fixed, so we no longer need to handle this here.

gudjong (approved):
Some naming considerations

apps/judicial-system/api/src/app/data-sources/backend.ts
```diff
@@ -46,6 +46,7 @@ import {
   IndictmentCount,
   UpdateIndictmentCountInput,
 } from '../modules/indictment-count'
+import { PoliceCaseInfo } from '../modules/police/models/policeCaseInfo.model'
```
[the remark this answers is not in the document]
unakb: This is the only model being imported from this area right now and there's no existing index file that makes sense to put this in, so I'm not sure I see the benefit. The other types are being imported from the types library which I decided not to do for PoliceCaseInfo as I didn't add it to types and want to use the generated schema instead.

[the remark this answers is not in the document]
gudjong: The idea in the api and backend projects is to have a single index file at the top of each module. Everything exported from the module is exported from the index file. No index files are used within the modules themselves. This was done to make it a bit clearer which types/classes/enums/etc are need outside each module.

PR #11522, opened 2023-06-19
fix(ehic): Add expiry date, exclude old card when applying for pdf if applying for new card

unnarb:
# ...

## What

## Why

## Screenshots / Gifs

## Checklist:


libs/application/templates/european-health-insurance-card/src/lib/helpers/applicantHelper.ts
```diff
@@ -56,6 +56,17 @@ export function getFullName(
   return null
 }
 
+export function getPlasticExpiryDate(resp: CardResponse): Date | undefined {
+  try {
```
valurefugl, line 60: Why do we need try/catch here?

valurefugl (approved):

libs/application/templates/european-health-insurance-card/src/lib/helpers/applicantHelper.ts
unnarb: if the Date conversion would fail.. i.e. expiryDate is some gibberish

PR #11539, opened 2023-06-20
feat(ids-admin): Publish permission

snaerseljan:
# Publish permission

## What

This PR adds Publish permission feature to our permission/scope.
Now we can easily duplicate environment with one click just like in our client logic.
This is now a modal instead of a modal route. I have updated the PublishClient also to be just a modal instead of a modal route. The reason for the modal route change is that the publish route was giving us unexpected behaviour when navigating back and forward.
 
## Why

It will easy our life easier when creating new env in permission

## TODO

- [x] Fix tests

## Screenshots / Gifs

image

And if error occurs

image

## Checklist:


apps/services/auth/admin-api/src/app/v2/scopes/test/me-scopes.spec.ts
```diff
@@ -545,19 +561,23 @@ describe('MeScopesController', () => {
...
         expect(response.status).toEqual(testCase.expected.status)
-
         if (response.status === 200) {
           // Assert response
           expect(response.body).toStrictEqual({
             ...testCase.expected.body,
+            grantToAuthenticatedUser: testCase.expected.body
+              .grantToAuthenticatedUser
+              ? testCase.expected.body.grantToAuthenticatedUser
+              : true,
```
eirikurn, line 571: This always assigns true.

libs/api/domains/auth-admin/src/lib/scope/dto/publish-scope.input.ts
```diff
@@ -0,0 +1,17 @@
+import { Environment } from '@island.is/shared/types'
+import { Field, InputType } from '@nestjs/graphql'
+
+@InputType('AuthAdminPublishScopeInput')
+export class PublishScopeInput {
+  @Field(() => Environment, { nullable: false })
```
eirikurn, line 6:
```suggestion
  @Field(() => Environment)
```

Nullable false is the default. Same for below.

libs/api/domains/auth-admin/src/lib/scope/scope.service.ts
```diff
@@ -84,6 +95,51 @@ export class ScopeService extends MultiEnvironmentService {
...
+    // Fetch the scope from source environment
+    const sourceInput = await this.adminApiByEnvironmentWithAuth(
+      input.sourceEnvironment,
+      user,
+    )
+      ?.meScopesControllerFindByTenantIdAndScopeName({
+        tenantId: input.tenantId,
+        scopeName: input.scopeName,
+      })
+      .catch((error) => this.handleError(error, input.sourceEnvironment))
```
eirikurn, line 116:
```suggestion
      ?.catch((error) => this.handleError(error, input.sourceEnvironment))
```
Will this throw if `adminApiByEnvironmentWithAuth` returns undefined?

libs/auth-api-lib/src/lib/resources/admin/admin-scope.service.ts
```diff
@@ -113,10 +113,34 @@ export class AdminScopeService {
       throw new BadRequestException(`Scope name "${input.name}" already exists`)
     }
 
+    // We need to manually validate displayName and description, since we reuse the same dto for both create and publish
+    const translatedValuesErrorMsg =
+      'Scope displayName and description are required'
+
+    if (!input.displayName || !input.description) {
+      throw new BadRequestException(translatedValuesErrorMsg)
+    }
```
eirikurn, line 122: Don't understand why this isn't validated like other fields?

libs/auth-api-lib/src/lib/resources/admin/admin-scope.service.ts
```diff
@@ -113,10 +113,34 @@ export class AdminScopeService {
       throw new BadRequestException(`Scope name "${input.name}" already exists`)
     }
 
+    // We need to manually validate displayName and description, since we reuse the same dto for both create and publish
```
eirikurn, line 116: This API doesn't have any concept of "publish". This PR is reusing the DTO for post (create) and patch (update).

apps/services/auth/admin-api/src/app/v2/scopes/test/me-scopes.spec.ts
```diff
@@ -265,21 +260,42 @@ const createTestCases: Record<string, CreateTestCase> = {
       body: expectedCreateOutput,
     },
   },
+  'should publish scope and have access as current user': {
```
eirikurn, line 263: Maybe rephrase to explain expanded create/post functionality, since the rest API has no concept of publish.

libs/portals/admin/ids-admin/src/screens/Client/PublishClient/PublishClient.tsx
```diff
@@ -1,130 +1,116 @@
...
+              targetEnvironment: data.targetEnvironment,
+              sourceEnvironment: data.sourceEnvironment,
+            },
+          },
+        })
+      } catch (e) {
+        setError(true)
+      }
+    },
+    [publishClient, tenantId],
```
eirikurn, line 64: Check these lint warnings?

eirikurn (commented):
Great job! 🎉Focused on reviewing the backend code and just skimmed the frontend. Happy to pair review specific frontend code if you want.

eirikurn (approved):
Approving since the comments were all nice to have.

libs/api/domains/auth-admin/src/lib/scope/scope.service.ts
snaerseljan: I am pretty sure it does not, at least this does not throw

```
const p = new Promise((resolve) => resolve(undefined))
await p.catch((err) => {
    throw err
})
```

or am I missing or misunderstanding something

apps/services/auth/admin-api/src/app/v2/scopes/test/me-scopes.spec.ts
snaerseljan: I will add isDefined check there

libs/auth-api-lib/src/lib/resources/admin/admin-scope.service.ts
snaerseljan: The fields are optional code wise because we extend the AdminPatchScopeDto, but required in create in the real world.

```
export class AdminCreateScopeDto extends AdminPatchScopeDto {
  @IsString()
  @IsNotEmpty()
  @ApiProperty({ example: '@island.is' })
  name!: string
}
```

libs/auth-api-lib/src/lib/resources/admin/admin-scope.service.ts
snaerseljan: Removed comment

PR #11545, opened 2023-06-20
fix(general-petition): download Pdf in admin

albinagu:
## Checklist:

### DIFF
--- a/libs/application/templates/estate/src/fields/DoneImage/assets/Grieving.tsx
+++ b/libs/application/templates/estate/src/fields/DoneImage/assets/Grieving.tsx
@@ -1,7 +1,12 @@
 import React from 'react'
 
 const Grieving: React.FC = () => (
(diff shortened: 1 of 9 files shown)


libs/portals/admin/petition/src/components/ExportList/index.tsx
```diff
@@ -22,24 +32,56 @@ interface Props {
...
     <Box className={styles.buttonWrapper}>
       <DropdownMenu
         icon="download"
+        iconType="outline"
         menuLabel={formatMessage(m.downloadPetitions)}
         items={[
+          {
+            title: formatMessage(m.asPdf),
+            render: () => (
+              <a
```
kksteini, line 72: Should we use something like `<Link>` or is that not applicable/possible in this case seeing that this is some special render for the pdf?

albinagu: sí

kksteini (approved):

PR #11552, opened 2023-06-21
chore(contentful-apps): Remove parent-slug contentful extension from monorepo

RunarVestmann:
# Remove parent-slug contentful extension from monorepo

## What

* The extension is no longer needed since we are now updating the baby article url field via the search-indexer instead of relying on an extension in Contentful to do that.

## Checklist:


eirikurn (approved):
🎉❤️

kksteini (approved):

PR #11563, opened 2023-06-21
fix(j-s): Court of appeal case goes to overview site

MargretFinnboga:
# ...

[Landsréttur - notandi í LR á að lenda á fyrsta skjá - yfirliti og laga hliðarstiku](https://app.asana.com/0/1199153462262248/1204814541638432)

## What

- Sending user always to overview site 

## Why

- Because otherwise that step is always skipped, sot his was decided

## Checklist:


unakb (approved):
This tackles one of the issues on the asana task but the other one with the side panel is still buggy, we can just create a new ticket for that

PR #11565, opened 2023-06-21
fix(j-s): Clear mocks in message handler tests

unakb:
[Refactora message handler tests ](https://app.asana.com/0/1199153462262248/1204308113516262)

## What

Clear mocks between tests in message handler service

## Why

So that tests don't return false positives when checking if fetch has been called with a certain method that has already been triggered by a previous test.

## Checklist:


MargretFinnboga (approved):
Nice 🐞

PR #11575, opened 2023-06-22
feat(Application-system): Custom delegation support.

Thorrik58:
# Custom delegation support.

## What

Adding in custom delegation support for application system.
If a user has a custom delegation and wants to access an application that allows custom delegations. The user must have the required scope for that application defined in the template in order to do so.
First test will be the veiðileyfi application.

## Why

Needed to reduce workload on procuration holders and allow 3rd parties to submit applications for the company when needed.

## Screenshots / Gifs

## Checklist:


thordurhhh (approved):
Code owned files lgtm

apps/application-system/api/src/app/modules/application/application.controller.ts
```diff
@@ -188,6 +195,7 @@ export class ApplicationController {
     @CurrentUser() user: User,
     @Query('typeId') typeId?: string,
     @Query('status') status?: string,
+    @Query('scopeCheck') scopeCheck?: boolean,
```
obmagnusson, line 198: Should we permit the client to determine whether a scope check should occur? Is it possible for a user to elevate their access privileges by modifying the query?

apps/application-system/api/src/app/modules/application/tools/applicationAccess.service.ts
```diff
@@ -122,6 +137,64 @@ export class ApplicationAccessService {
...
+
+    // Custom delegations must have required scopes on their template
+    if (!requiredScopes) {
+      return false
+    }
+
+    // If we dont have the scopes available like in service portal we need to fetch
+    // them for the user logged in
+    if (scopeCheck) {
+      const delegations = await this.actorDelegationsApi
```
obmagnusson, line 168: We should consider moving this API call to the client library, creating a separate service for better reusability.

obmagnusson (commented):
Some initial thoughts

apps/application-system/api/src/app/modules/application/tools/applicationAccess.service.ts
Thorrik58: I agree, would be happy to implement it once this has been merged to main as there is some pressure to get this in before the freeze.

apps/application-system/api/src/app/modules/application/application.controller.ts
Thorrik58: I think its ok as its just filtering out unneeded data, the client can already see his scopes if he wishes to.
But @eirikurn would probably be the best to determine if its a bigger issue.

obmagnusson (approved):

libs/clients/auth/public-api/src/lib/authPublicApiClient.module.ts
```diff
@@ -2,9 +2,18 @@ import { Module } from '@nestjs/common'
...
 
+export const AuthPublicApiClientConfigProvider = {
+  provide: AuthPublicApiClientConfig.KEY,
+  useValue: AuthPublicApiClientConfig,
+}
 @Module({
-  providers: [ApiConfiguration, ...exportedApis],
+  providers: [
+    ApiConfiguration,
+    AuthPublicApiClientConfigProvider,
```
eirikurn, line 14: Why?

Services should be able to inject it if it's been loaded correctly in the application root module.

libs/auth-api-lib/src/lib/delegations/delegations-incoming.service.ts
```diff
@@ -187,10 +189,14 @@ export class DelegationsIncomingService {
...
-    const delegations = ([] as MergedDelegationDTO[])
+    let delegations = ([] as MergedDelegationDTO[])
       .concat(...delegationSets)
       .filter((delegation) => delegation.fromNationalId !== user.nationalId)
 
+    if (otherUser) {
+      delegations = delegations.filter((d) => d.fromNationalId === otherUser)
+    }
+
     const mergedDelegationMap = delegations.reduce(
```
eirikurn, line 200: I would really like to merge scopes (and types) across delegations in a smarter way to support interesting edge/use cases, but it's not a priority for now.

eirikurn (approved):
Nice work. Mainly concerned about the performance of the findAll endpoint.

Approving to unblock. Feel free to prioritise the different performance suggestions in this or later PRs.

apps/application-system/api/src/app/modules/application/application.controller.ts
```diff
@@ -227,6 +235,7 @@ export class ApplicationController {
           application as BaseApplication,
           user,
           templates[typeId],
+          scopeCheck,
```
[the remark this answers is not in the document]
Thorrik58: Good catch, should've thought about that. Ill fix what I can.

libs/clients/auth/public-api/src/lib/authPublicApiClient.module.ts
Thorrik58: I made the change to get my schemas to build.
No idea why it behaves like it does on my end.
```
[App] Error	7/18/2023, 5:45:50 PM [ExceptionHandler] Nest can't resolve dependencies of the AuthPublicApiClientConfiguration (?). Please make sure that the argument CONFIGURATION(AuthPubliClient) at index [0] is available in the AuthPublicApiClientModule context.

       Potential solutions:
       - If CONFIGURATION(AuthPubliClient) is a provider, is it part of the current AuthPublicApiClientModule?
       - If CONFIGURATION(AuthPubliClient) is exported from a separate @Module, is that module imported within AuthPublicApiClientModule?
         @Module({
           imports: [ /* the Module containing CONFIGURATION(AuthPubliClient) */ ]
         })
```

PR #11609, opened 2023-06-28
fix(application-driving-license-duplicate): Update client and application

kksteini:
* Changed one radio field to a checkbox instead, since the endpoint does not have the granularity of choice expected
* Updated driving-license client v5 in order to get the definitions for a new parameter into the collaborative endpoint (stolenOrLost)

## Checklist:


libs/application/template-api-modules/src/lib/modules/templates/driving-license-duplicate/driving-license-duplicate.service.ts
```diff
@@ -68,13 +68,14 @@ export class DrivingLicenseDuplicateService extends BaseTemplateApiService {
...
     }
 
-    // TODO: add reason to duplicate submission?
-    // Currently we are tracking "stolen" vs "lost" in the application
-    // Does this need to be tracked in the license system?
     await this.drivingLicenseService
       .drivingLicenseDuplicateSubmission({
         districtId: parseInt(answers.district.toString(), 10),
         ssn: nationalId,
+        // Always true since submission doesn't happen until
```
kksteini, line 75:
```suggestion
        // Always true since submission doesn't happen before
```

albinagu (approved):

PR #11637, opened 2023-06-30
fix(j-s): Ruling Date

gudjong:
# Ruling Date

[Setja lokadag á ákærur](https://app.asana.com/0/1199153462262248/1204905731848700/f)

## What

- Renames the field `rulingDate` to `rulingSignatureDate` which better reflects the date stored in that field.
- Creates a new field `rulingDate` and sets it on case completion. For R cases the field is set to the value of the field `courtEndTime`. For S cases the field is set to `now`.
- Uses the new field `rulingDate` as the effective date of accepted R cases.
- Shows completed R cases that are missing a `rulingSignatureDate` in the list of active cases only for district court users.

## Why

- A ruling date was not being set for S cases and it was convenient to cleanup the logic around the effective date of R cases.

## Screenshots

image
image

## Checklist:


apps/judicial-system/web/src/routes/Shared/Cases/Cases.tsx
```diff
@@ -99,8 +101,8 @@ export const Cases: React.FC = () => {
   const { user } = useContext(UserContext)
   const [isFiltering, setIsFiltering] = useState<boolean>(false)
 
-  const isProsecutor = user?.role === UserRole.PROSECUTOR
-  const isRepresentative = user?.role === UserRole.REPRESENTATIVE
+  const isProsecutionUser = user && isProsecutionRole(user?.role)
+  const isDistrictCourtUser = user && isCourtRole(user?.role)
```
MargretFinnboga, line 105: nice 🌻

MargretFinnboga (approved):
Looks good ⚡  consider removing one unused import otherwise 👍🏼

PR #11653, opened 2023-07-03
fix(service-portal): Update my info overveiw paths

thordurhhh:
## What

* Update paths for my information and sub pages.

## Why

With the additions of **petitions** to the navigation tree, the primary parent path of **my-information** does not make much sense as when we will display it in the navigation it will become active once the user enters the petition site. That behaviour is expected because the petition site is a direct child of the my-information overview site. We can fix this by having the navigation set up as all the other service-portal pages. The subpath `/yfirlit` will be added for family + personal information, mostly national registry information.

Instead of:

`/minarsidur/min-gogn`
`/minarsidur/min-gogn/minar-upplysingar`
`/minarsidur/min-gogn/maki/minar-upplysingar`
`/minarsidur/min-gogn/listar`
`...`

We will have:

`/minarsidur/min-gogn/yfirlit` (`/minarsidur/min-gogn` will redirect to `/minarsidur/min-gogn/yfirlit`)
`/minarsidur/min-gogn/yfirlit/minar-upplysingar`
`/minarsidur/min-gogn/yfirlit/maki/minar-upplysingar`
`/minarsidur/min-gogn/listar`
`...`

## Checklist:


thorkellmani (approved):
lgtm

PR #11662, opened 2023-07-04
chore(devx): Disable schema generation in postinstall

AndesKrrrrrrrrrrr:
We're only running (slow!) schema generation on every `yarn install`. Luckily it's cached so developers aren't as frequently impacted. On a cold install we run schemas, but with the new `yarn` version there is no output, so the install seems to be hanging.
```
# Cold install (only building island.is@workspace)
➤ YN0000: ┌ Link step
➤ YN0007: │ island.is@workspace:. must be built because it never has been before or the last one failed
➤ YN0000: └ Completed in 4m 47s

# Warm install
➤ YN0000: ┌ Link step
➤ YN0008: │ island.is@workspace:. must be rebuilt because its dependency tree changed
➤ YN0000: └ Completed in 2s 164ms
```

## Checklist:


README.md
```diff
@@ -45,6 +44,15 @@ If you are running on Windows we recommend using [Docker and WSL2](https://docs.
...
 
+### Fresh start
+
+Install dependencies and generate some library files:
+
+```bash
+yarn install
+yarn schemas
+```
+
```
eirikurn, line 55: Devs often need to generate schemas when they switch between branches with different schemas or pull latest from main.

Devs might be used to `yarn install` also generating schemas, so the docs should explain that devs now need to also run `yarn schemas` manually.

eirikurn (changes requested):

eirikurn (approved):

eirikurn:
LGTM. Please announce this change on slack 👍

PR #11666, opened 2023-07-04
fix(island-ui): Remove redundant translate=yes

thordurhhh:
## What

Remove default `translate="yes"` from `<Text />`

## Why

With the [recent change](https://github.com/island-is/island.is/pull/11479) of opting out of translations for the `<Text />` component, the default of `translate="yes"` has been added to every Text component where developers are not specifically setting it. The default should be undefined. Today it will add `translate="yes"` attribute as the default option. With `undefined` it will still be translatable as default, but less clutter in the HTML.

## Checklist:


snaerseljan (approved):
LGTM

You are sure it will not break some Text components usage removing default value?

PR #11679, opened 2023-07-05
feat(service-portal): Add national registry v3 - data broker service 

thorkellmani:
## What

Add a client and a service for fetching user info

## Why

There're devs who need data from the National registry that's only provided through the new service. We're releasing the client early so they can access the data, we'll then integrate the new service in our own apps later on.

## Checklist:


lodmfjord (approved):
LGTM

thordurhhh (approved):
Nice! 🙌

apps/api/infra/api.ts
```diff
@@ -179,6 +179,29 @@ export const serviceSetup = (services: {
         staging: 'https://samradapi-test.devland.is',
         prod: 'https://samradapi.island.is',
       },
+      NATIONAL_REGISTRY_B2C_CLIENT_ID: {
+        dev: 'b464afdd-056b-406d-b650-6d41733cfeb7',
+        staging: 'ca128c23-b43c-443d-bade-ec5a146a933f',
+        prod: '2304d7ca-7ed3-4188-8b6d-e1b7e0e3df7f',
+      },
```
eirikurn, line 186: Should move these into a "xroad" config pack?

libs/api/domains/national-registry-v3/src/lib/graphql/models/nationalRegistryAddress.model.ts
```diff
@@ -0,0 +1,19 @@
...
+
+  @Field(() => String, { nullable: true })
+  city?: string | null
+
+  @Field(() => String, { nullable: true })
+  municipalityText?: string | null
+
+  @Field(() => String, { nullable: true })
+  municipalityCode?: string | null
+}
```
eirikurn, line 19: I would really like to have a single national-registry domain and normalise (share) these object types as much as possible. It may take more work now, but it will make our schema nicer and I'm sure it will save us work and pain later when we transition all our applications to this new service.

libs/api/domains/national-registry-v3/src/lib/nationalRegistryV3.resolver.ts
```diff
@@ -0,0 +1,165 @@
...
+    nullable: true,
+  })
+  @Audit()
+  async resolveResidenceHistory(
+    @Context('req') { user }: { user: User },
+    @Parent() person: Person,
+  ): Promise<Array<NationalRegistryV3Service> | null> {
+    /*return this.nationalRegistryV3Service.getNationalRegistryResidenceHistory(
+      person.nationalId,
+    )*/
```
eirikurn, line 106: Not ready?

libs/api/domains/national-registry-v3/src/lib/nationalRegistryV3.service.ts
```diff
@@ -0,0 +1,310 @@
...
+  }
+
+  async getNationalRegistryPerson(nationalId: string): Promise<Person | null> {
+    const person = await this.fetchData(nationalId)
+
+    return this.extractPerson(person)
+  }
+
+  async getAddress(nationalId: string): Promise<Address | null> {
+    const data = await this.fetchData(nationalId)
```
eirikurn, line 56: Are all of the field resolvers fetching the same person again and again?

You could include the raw data in a private optional field of the `Person` type (eg `rawData`) so you don't have to fetch the data again in the field resolvers.

libs/clients/national-registry/v3/src/lib/apiConfig.ts
```diff
@@ -0,0 +1,38 @@
...
+
+export const ApiConfig = {
+  provide: 'NationalRegistryV3ClientProviderConfiguration',
+  scope: LazyDuringDevScope,
+  useFactory: (
+    xroadConfig: ConfigType<typeof XRoadConfig>,
+    config: ConfigType<typeof NationalRegistryV3ClientConfig>,
+  ) =>
+    new Configuration({
+      fetchApi: createEnhancedFetch({
```
eirikurn, line 19: In the current national registry client we configure our own redis caching to speed up lookups and provide `stale-if-error` handling when there are problems with the service.

I wonder if we can trust this new service to be more reliable and faster?

libs/api/domains/national-registry-v3/src/lib/nationalRegistryV3.service.ts
```diff
@@ -0,0 +1,310 @@
...
+      ? parentData.forsja.born
+      : []
+
+    const childDetails: Array<Person | null> = await Promise.all(
+      children.map(async (child) => {
+        if (!child.barnKennitala || !child.barnNafn) {
+          return null
+        }
+
+        const childDetails = await this.fetchData(child.barnKennitala)
```
eirikurn, line 224: I wonder if we should use the new batch endpoint for querying children and parents?

libs/clients/national-registry/v3/src/lib/apiConfig.ts
thorkellmani: reimplement the caching perhaps?

libs/api/domains/national-registry-v3/src/lib/nationalRegistryV3.service.ts
thorkellmani: yeah this was a bit undercooked, i'll fix

libs/api/domains/national-registry-v3/src/lib/nationalRegistryV3.service.ts
thorkellmani: I'm unfamiliar, can you elaborate?

libs/api/domains/national-registry-v3/src/lib/nationalRegistryV3.resolver.ts
thorkellmani: Not available in the new service. They only provide a user's current residence, the last icelandic residence a user had, and the residence the user had as of last 1st of december.

PR #11693, opened 2023-07-06
feat(web): Add relative url mapping so CMS links can point to prod and code makes them relative

RunarVestmann:
# Add relative url mapping so CMS links can point to prod and code makes them relative

## What

For E2E tests to not leave the environment they are testing once they navigate using a link, this was deemed the best approach. Instead of relying on CMS users to create relative links we can instead handle it so if they create an absolute link to prod then we'll automatically make that link relative

## Checklist:

### DIFF
--- a/libs/cms/src/lib/models/utils.ts
+++ b/libs/cms/src/lib/models/utils.ts
@@ -1,15 +1,3 @@
-const prefix = 'https://island.is'
-
 /** Make sure that links are relative in case someone links to production directly */
-export const getRelativeUrl = (url: string) => {
-  const trimmedUrl = url.trim()
-
-  if (trimmedUrl.startsWith(prefix)) {
-    if (trimmedUrl.length === prefix.length) {
-      return '/'
-    }
-    return trimmedUrl.slice(prefix.length)
-  }
-
-  return trimmedUrl
-}
+export const getRelativeUrl = (url: string) =>
+  new URL(url.trim(), 'https://island.is').pathname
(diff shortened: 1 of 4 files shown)


libs/cms/src/lib/models/utils.ts
```diff
@@ -0,0 +1,15 @@
+const prefix = 'https://island.is'
+
+/** Make sure that links are relative in case someone links to production directly */
+export const getRelativeUrl = (url: string) => {
+  const trimmedUrl = url.trim()
+
+  if (trimmedUrl.startsWith(prefix)) {
```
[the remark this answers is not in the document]
RunarVestmann: I've pushed new code that uses the `new URL()` method instead.

libs/cms/src/lib/models/utils.ts
```diff
@@ -0,0 +1,3 @@
+/** Make sure that links are relative in case someone links to production directly */
+export const getRelativeUrl = (url: string) =>
+  new URL(url.trim(), 'https://island.is').pathname
```
AndesKrrrrrrrrrrr, line 3: Could maybe remove all hosts? This is indeed the most common, even the only host in our urls. But what if someone accidentally adds `http://localhost` or `https://my-feature.dev01.devland.is`?

RunarVestmann: Hmm, ok so if the host is similar to the check made here: 
image

Then we should remove it, to cover the cases where a link is perhaps pointing to dev? What if a link is supposed to be pointing to a feature deployed link, would we still want to overwrite it and make it relative?

RunarVestmann: I noticed an error in my logic I was basically making external links be relative (by joinking their pathname) so I fixed that

Ingimarsson (approved):

PR #11695, opened 2023-07-06
fix(service-portal): Download/copy link for list in mobile

albinagu:
## Checklist:

### DIFF
--- a/libs/service-portal/petitions/src/lib/messages.ts
+++ b/libs/service-portal/petitions/src/lib/messages.ts
@@ -110,6 +110,11 @@ export const m = defineMessages({
     defaultMessage: 'Sem CSV',
     description: '',
   },
+  linkToList: {
+    id: 'sp.petitions:linkToList',
+    defaultMessage: 'Hlekk á lista',
+    description: '',
+  },
   copyLinkToList: {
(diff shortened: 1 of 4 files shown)


libs/service-portal/petitions/src/lib/messages.ts
```diff
@@ -110,6 +110,11 @@ export const m = defineMessages({
     defaultMessage: 'Sem CSV',
     description: '',
   },
+  linkToList: {
+    id: 'sp.petitions:linkToList',
+    defaultMessage: 'Hlekk á lista',
```
kksteini, line 115:
```suggestion
    defaultMessage: 'Hlekkur á lista',
```

kksteini (approved):

libs/service-portal/petitions/src/lib/messages.ts
albinagu: þetta er í dropdown sem heitir Sækja
er það þá Hlekk eða hlekkur?

kksteini: Oh, didn't know the context. My bad, no "Hlekk" is correct.
I just assumed nominative :sweat_smile:

PR #11698, opened 2023-07-06
feat(web): Aircraft registry search on public web

RunarVestmann:
# Aircraft registry connected component on public web

## What

* Added a component that can be embedded into rich text in the CMS. This component is similar to what is seen here: https://www.samgongustofa.is/flug/loftfor/loftfaraskra/
* The data comes from a service in xroad but that service only exists on dev. Since this component is something that only appears if it is embedded inside of rich text in the CMS I deem it's fine to get this PR to main even though there is a need for a future PR that updates the xroad paths once we can call this xroad service on staging and prod

## Screenshots / Gifs

image

## Checklist:


libs/clients/aircraft-registry/src/lib/apiConfig.ts
```diff
@@ -0,0 +1,33 @@
...
+  useFactory: (
+    xroadConfig: ConfigType<typeof XRoadConfig>,
+    config: ConfigType<typeof AircraftRegistryClientConfig>,
+  ) => {
+    return new Configuration({
+      fetchApi: createEnhancedFetch({
+        name: 'clients-aircraft-registry',
+        logErrorResponseBody: true,
+        treat400ResponsesAsErrors: true,
+        timeout: 20000,
```
eirikurn, line 22: Sure 400 responses be treated as errors? Can skip `logErrorResponseBody` and `timeout` since these are the current default values.

libs/api/domains/aircraft-registry/src/lib/models/aircraft.model.ts
```diff
@@ -0,0 +1,61 @@
...
+  @Field(() => String, { nullable: true })
+  postcode?: string | null
+
+  @Field(() => Number, { nullable: true })
+  ssn?: number | null
+}
+
+@ObjectType('AircraftRegistryAircraft')
+export class Aircraft {
+  @Field(() => String, { nullable: true })
```
eirikurn, line 32: Should all of these fields be nullable?

libs/api/domains/aircraft-registry/src/lib/dto/allAircraftsResponse.ts
```diff
@@ -0,0 +1,17 @@
...
+  aircrafts?: Aircraft[] | null
+
+  @Field(() => Number, { nullable: true })
+  pageNumber?: number | null
+
+  @Field(() => Number, { nullable: true })
+  pageSize?: number | null
+
+  @Field(() => Number, { nullable: true })
+  totalCount?: number | null
```
eirikurn, line 16: Why is all of this nullable?

libs/api/domains/aircraft-registry/src/lib/aircraftRegistry.resolver.ts
```diff
@@ -0,0 +1,38 @@
...
+
+const defaultCache: CacheControlOptions = { maxAge: CACHE_CONTROL_MAX_AGE }
+
+@Resolver()
+export class AircraftRegistryResolver {
+  constructor(
+    private readonly aircraftRegistryService: AircraftRegistryService,
+  ) {}
+
+  @CacheControl(defaultCache)
```
eirikurn, line 18: Should use `@CacheField()` in the object types if you want these results to be cacheable.

eirikurn (approved):
Minor comments

libs/api/domains/aircraft-registry/src/lib/aircraftRegistry.resolver.ts
RunarVestmann: I agree, it didn't pop up as a "lint error" so I simply forgot about that decorator, thanks for the reminder

libs/api/domains/aircraft-registry/src/lib/models/aircraft.model.ts
RunarVestmann: Since the openapi schema suggests that these values can be null, I basically went overboard and typed everything as nullable 🙃

brynjarorng (approved):
LGTM

Ingimarsson (approved):

PR #11729, opened 2023-07-12
fix(j-s): Court of appeal route navigation updated 

MargretFinnboga:
# ...

[Landsréttur - laga hliðarstiku 
](https://app.asana.com/0/1199153462262248/1204873846081389)

## What

- Instead of messing with the side panel using the route to logic 

## Why

- More simple solution and more scalable 

## Checklist:


unakb (approved):
Nice ! Much more user friendly behaviour I think 🙌

PR #11736, opened 2023-07-13
fix: increased ids memory limits

brynjarorng:
Increased the memory and CPU requested by the search indexer and IDS based on current limits and requests from the production cluster


lodmfjord (approved):
LGTM

PR #11744, opened 2023-07-14
fix(core): Broken metric server response

eirikurn:
## What

- Fix metric server.
- Also fixes audit name of delegation api.

## Why

This caused this runtime error in our logs:

> TypeError [ERR_INVALID_ARG_TYPE]: The "chunk" argument must be of type string or an instance of Buffer or Uint8Array. Received an instance of Promise


valurefugl (approved):

brynjarorng (approved):
LGTM

PR #11746, opened 2023-07-14
fix(j-s): disable police case number input field if the police case number is from Löke 

MargretFinnboga:
# ...

[Læsa málsnúmers reit fyrir mál sem koma úr LÖKE](https://app.asana.com/0/1199153462262248/1205042351918768)

## What

- When police case number come from Löke, not allow editing that field

## Why

- To not have inconsistency in the two systems 

## Checklist:


unakb (approved):
Nice!

PR #11760, opened 2023-07-19
feat(datadog): CI testing

AndesKrrrrrrrrrrr:
Test result dashboard is working at https://app.datadoghq.eu/ci/test-services

Need thoughts on:
- Where to set `NODE_OPTIONS`
- Whether to configure environment variables in scripts or workflow


.github/actions/unit-test/action.yml
```diff
@@ -41,7 +43,7 @@ runs:
       shell: bash
       env:
         DOCKER_REGISTRY: ${{ inputs.docker-registry }}
-        NODE_OPTIONS: --max-old-space-size=8192
```
brynjarorng, line 44: Could this cause failures? We added this to fix large tests running out of memory

scripts/ci/30_test.sh
```diff
@@ -1,6 +1,15 @@
 #!/bin/bash
 set -euxo pipefail
 
+: "${DD_CIVISIBILITY_AGENTLESS_ENABLED:=true}"
+: "${DD_SITE:=datadoghq.eu}"
+: "${DD_ENV:=dev}"
+: "${DD_SERVICE:=unit-test-action}"
+: "${DD_API_KEY:='<set-api-key>'}"
+: "${NODE_OPTIONS:=--max-old-space-size=8192}"
```
brynjarorng, line 9: Is it not redundant to have the max old space config here since it is also defined below?

scripts/ci/30_test.sh
```diff
@@ -1,6 +1,15 @@
...
+
+# Default to big old-space, and more options for testing, but allow overriding
+NODE_OPTIONS="--max-old-space-size=8193 --unhandled-rejections=warn --require=dd-trace/ci/init ${NODE_OPTIONS:-}"
```
brynjarorng, line 12: Could this not cause errors if someone else defines this and now it is defined twice?

brynjarorng (changes requested):

scripts/ci/30_test.sh
AndesKrrrrrrrrrrr: Later flags should override earlier flags if defined multiple times (if boolean or single value), or add to a list (like with `--require`, which would add _more_ `--require`s)

scripts/ci/30_test.sh
AndesKrrrrrrrrrrr: This allows overriding/adding settings using environment variables. Below we are setting some options we think are a minimum. That way we will always have at least `--max-old-space-size=8192`, even if the user sets `NODE_OPTIONS` to e.g. `--verbose` (which would disable `--max-old-space-size`).

AndesKrrrrrrrrrrr: Though, the default in line 9, is indeed redundant.

brynjarorng (approved):
LGTM

PR #11784, opened 2023-08-01
chore: allow extending .envrc with .envrc.private

busla:
## What

Extend .envrc to allow loading custom .envrc files

## Why
.envrc should allow developers to load their own custom configurations along with the defaults set by team core.


peturgq (approved):

saevarma:
I thought that the idea with `.envrc` was to share common unsensitive environment settings recommended for the repo, while programmers can use `.env.secret` to add custom (and sensitive) environment settings that are not committed.

Have you any thoughts on if/how we should share common env settings to developers?

eirikurn (changes requested):
This is likely to change current dx workflows for all developers. E.g. with get-secrets. 

Imo should close this PR and discuss strategies for supporting alternate env workflows on Slack if it's important.

busla:
reopened the PR with a fix.

`direnv` will now load `.envrc.private` if it exists for custom config

apps/application-system/api/src/environments/environment.ts
saevarma, line 1: Is this file change supposed to be part of this PR

saevarma:
You should update the PR's title to reflect the changes, for the generated changelog.

saevarma (approved):
LGTM

PR #11791, opened 2023-08-03
fix(j-s): Fix Delete Issue

gudjong:
# Fix Delete Issue

[Böggur í málsnúmer lögreglu virkni](https://app.asana.com/0/1199153462262248/1204998926887904/f)

## What

- Fixes an issue with deleting police case numbers from indictments.
- Not the most elegant solution but seems to work consistently.

## Why

- Verified bug.

## Screenshots / Gifs

https://github.com/island-is/island.is/assets/795382/c46cdc89-4888-4163-b5dd-bc318c6c3a81

## Checklist:


unakb (approved):
Like you said, it works, but it's hard to write elegant code around the police case data structure as is. 
I'd like to discuss whether we should consider refactoring how we work with the police case info, there's a lot of puzzling together data happening in the components that we could abstract more

PR #11799, opened 2023-08-03
feat(web): Section with video

RunarVestmann:
# Section with video

## What

* Add new content type similar to "section with image" that use video instead.
I first tried reusing the Section with image content type by having a boolean and a video field along with a conditional field rendering contentful app but that wasn't so great since the fields didn't completely disappear once you toggled the bool (along the fact that it's a bit more complicated to grasp). So I just ended up with adding a Section with video since that is simpler for users in Contentful.

## Why

* This is needed for the [SHH](https://shh.is/) organization since most of their content is sign language translated via a video format

## Screenshots / Gifs

Test example using random video and text:
image

Actual Figma screenshot:
image

## Checklist:


Ingimarsson (approved):

Toti91 (approved):
LGTM

PR #11826, opened 2023-08-10
fix(core): Add @nestjs/axios in esbuild.json externalDependencies.

saevarma:
## What

Adding `@nestjs/axios` as `externalDependencies` in `esbuild.json` for
* `judicial-system-api`
* `judicial-system-backend`
* `user-notification` which applies also for the worker.

## Why

These services are using `cms-translations` library that uses the `CmsModule` which includes `HttpHealthIndocator` from `@nestjs/axios`. Needed to fix regression from the monorepo upgrade.

## Screenshots / Gifs

N/A

## Checklist:


rafnarnason (approved):
lgtm

unakb (approved):

PR #11831, opened 2023-08-14
fix(service-portal): Fixes login error in portals on localhost

snaerseljan:
# Login fix in our portals.

## What

There is a bug on localhost when logging into our portals, i.e. service-portal or admin-portal. The useEffectOnce hook used in `react-use` library does not account for React 18 strict mode. Therefore the effect runs twice which results in OIDC callback is called twice and that causes the error.

I wrote a new `useEffectOnce` hook which addresses the problem.

## Why

Error on localhost

## Checklist:


saevarma (approved):
💯

RunarVestmann (approved):

Toti91 (approved):

PR #11836, opened 2023-08-14
fix(application-system): Delegations screen stuck in loading state

Toti91:
# Delegations screen stuck in loading state

## What

React 18 `StrictMode` exposed a bug (local development only) in the applications screen where the delegations check gets stuck because we were  setting a `delegationCheck` by toggling it on/off instead of directly setting it to true. Because of the double render of StrictMode in development mode this check was always being toggled twice and therefore never changed to the correct value.

## Why

Fixes regression after big NX update 

## Screenshots / Gifs

## Checklist:


Halli1414 (approved):
LGTM

obmagnusson (approved):

PR #11838, opened 2023-08-15
chore(j-s): Unit Test Performance

gudjong:
# Unit Test Performance

[Test improvements](https://app.asana.com/0/1199153462262248/1205309042829965/f)

## What

- Various test improvements (performance, cleanup).

## Why

- Trying to make test behave better in the build pipeline.

## Checklist:


apps/judicial-system/backend/src/app/modules/notification/test/notificationController/getAllCaseNotificationsRolesRules.spec.ts
```diff
@@ -0,0 +1,29 @@
...
+  let rules: any[]
+
+  beforeEach(() => {
+    rules = Reflect.getMetadata(
+      'roles-rules',
+      NotificationController.prototype.getAllCaseNotifications,
+    )
+  })
+
+  it('should give permission to five roles', () => {
```
oddsson, line 21: Consider changing the description of this test to be a little less specific, something like "should give permission to appropriate roles". That way we don't need to update the title if roles are added or removed.

oddsson (commented):
One minor comment

apps/judicial-system/backend/src/app/modules/notification/test/notificationController/getAllCaseNotificationsRolesRules.spec.ts
gudjong: Good point. Will keep in mind for the future, but there is a huge number of test with this kind of descriptions 😅

oddsson (approved):

PR #11842, opened 2023-08-15
feat(j-s): Use new ids

unakb:
# Use new ids

[Ný innskráning island.is](https://app.asana.com/0/1199153462262248/1204723923830375/f)

## What

Use new ids login page 

## Why

So users can login via app as well as phone number

## Checklist:

---
  - https://app.asana.com/0/0/1205314680853560


apps/judicial-system/api/src/app/modules/auth/auth.controller.ts
```diff
@@ -146,12 +103,89 @@ export class AuthController {
...
+            res,
+            redirectRoute,
+            new Entropy({ bits: 128 }).string(),
+          )
+        }
+        if (!verifiedToken) {
+          this.logger.debug('Token verification unsuccessful')
+        }
+      } catch (error) {
+        this.logger.debug('Token verification failed')
```
unakb, line 185: log error

apps/judicial-system/api/src/app/modules/auth/auth.controller.ts
```diff
@@ -146,12 +103,89 @@ export class AuthController {
...
+            {
+              nationalId: token.nationalId,
+            },
+            res,
+            redirectRoute,
+            new Entropy({ bits: 128 }).string(),
+          )
+        }
+        if (!verifiedToken) {
+          this.logger.debug('Token verification unsuccessful')
```
unakb, line 182: log error

gudjong (approved):

apps/judicial-system/api/infra/judicial-system-api.ts
```diff
@@ -7,15 +7,20 @@ export const serviceSetup = (services: {
...
+      AUTH_IDS_SCOPE: 'openid profile',
+      AUTH_IDS_CLIENT_ID: '@rettarvorslugatt.island.is/web',
+      AUTH_IDS_DOMAIN: {
+        dev: 'https://identity-server.dev01.devland.is',
+        staging: 'https://identity-server.staging01.devland.is',
+        prod: 'https://innskra.island.is',
       },
-      AUTH_AUDIENCE: {
-        dev: ref((h) => `judicial-system.${h.env.domain}`),
-        staging: ref((h) => `judicial-system.${h.env.domain}`),
```
AndesKrrrrrrrrrrr, line 17: I'm not too fond of hard-coded URLs. This change will probably break feature-deployments.

apps/judicial-system/api/src/environments/environment.ts
AndesKrrrrrrrrrrr, line 1: What is this file used for? Why isn't `apps/judicial-system/api/infra/judicial-system-api.ts` enough?

unakb: This is just legacy config that we're phasing out

apps/judicial-system/api/infra/judicial-system-api.ts
unakb: Could you elaborate, do you mean rather doing something like this? :
 dev: ref(
          (h) =>
            `https://judicial-system.${h.env.domain}/api/auth/callback/identity-server`,
        ),
        
        Or do you mean they should be kept somewhere else like the AWS parameter store?

AndesKrrrrrrrrrrr: Yes, that should work. Can we try it out and make a feature deployment to be sure?

apps/judicial-system/api/infra/judicial-system-api.ts
```diff
@@ -7,15 +7,24 @@ export const serviceSetup = (services: {
...
-        prod: 'rettarvorslugatt.island.is',
+      AUTH_IDS_REDIRECT_URI: {
+        dev: ref(
+          (h) =>
+            `https://judicial-system.${h.env.domain}/api/auth/callback/identity-server`,
+        ),
+        staging: ref(
+          (h) =>
+            `https://judicial-system.${h.env.domain}/api/auth/callback/identity-server`,
+        ),
```
AndesKrrrrrrrrrrr, line 25:
```suggestion
        dev: ref(
          (ctx) =>
            `${
              ctx.featureDeploymentName ? `${ctx.featureDeploymentName}-` : ''
            }judicial-system.dev01.devland.is`,
          ),
        staging: ref(
          (h) =>
            `https://identity-server.staging01.devland.is/api/auth/callback/identity-server`,
        ),
```

AndesKrrrrrrrrrrr (commented):
Other than my suggested change to support feature-deployments, LGTM

apps/judicial-system/api/infra/judicial-system-api.ts
unakb: I've applied this change

AndesKrrrrrrrrrrr (approved):

PR #11846, opened 2023-08-16
fix(j-s): Repaint happening on input

oddsson:
# Repaint happening on input

[Asana](https://app.asana.com/0/1199153462262248/1203856233777611/f)

## What

When filling in a indictment's police case number info, the tags that show the indictment sub types would repaint (flicker), when entering the police case number. This was happening because the tag's `key` was being changed on every keystroke when entering the police case number.

## Why

This was a small UI bug causing the UI to flicker when entering the police case number.

### Before 

https://github.com/island-is/island.is/assets/3789875/50cfb46b-40f2-469b-82ad-00d25361aee3

### After

https://github.com/island-is/island.is/assets/3789875/32e0334c-0863-4a87-94db-82e4695f58f4

## Checklist:


unakb (approved):
Looks good!

PR #11867, opened 2023-08-17
fix(j-s): Update case facts and legal arguments when case is resent

unakb:
[Yfirskrifa texta í kröfu ef ákærandi uppfærir kröfu eftir að fyrirtaka hefur verið skráð. Svæði:  málsatvik og lagarök](https://app.asana.com/0/1199153462262248/1203865140443233/f)

## What

Overwrite pre-existing text in court case facts and court legal arguments

## Why

So that judges get the updated information if a prosecutor changes it after judges have received a case

## Checklist:


apps/judicial-system/backend/src/app/modules/case/case.controller.ts
```diff
@@ -229,6 +229,11 @@ export class CaseController {
       update.rulingModifiedHistory = `${history}${today} - ${user.name} ${user.title}\n\n${update.rulingModifiedHistory}`
     }
 
+    if (update.caseResentExplanation) {
+      update.courtCaseFacts = `Í greinargerð sóknaraðila er atvikum lýst svo: ${theCase.caseFacts}`
+      update.courtLegalArguments = `Í greinargerð er krafa sóknaraðila rökstudd þannig: ${theCase.legalArguments}`
+    }
+
```
unakb, line 236: I intentionally hardcoded the strings instead of putting them in Contenful, I didn't feel it was necessary but if there are any objections I can change it 🤗
It just felt like a bit of code overhead for strings that are unlikely to change, and we don't have a string formatter in the Controller so I'd have to inject it or move this logic to the service away from the other similar logic.

oddsson (approved):

PR #11868, opened 2023-08-17
feat(nest-audit): Add log to console options in audit module

snaerseljan:
# Log to Console support in AuditModule

This PR includes log to console support in our AuditModule. This is great for when developers want their logs in Datadog for example.
We also added started to use the new log to console in our `auth-admin-api` mutation methods.

Developers only need to add one property to the audit decorator or audit method. See example below .

Using decorator
```
  @Audit<SomeType>({
    **alsoLog: true,**
  })
```

Using auditService.audit method

```
return this.auditService.auditPromise<SomeType>(
      {
        namespace,
        auth: user,
        action: 'update',
        resources: (scope) => scope.name,
        **alsoLog: true,**
      },
      this.someService.updateSomething(...),
    )
```

## Checklist:


saevarma (approved):
Solid and clean 💯

PR #11901, opened 2023-08-22
fix(j-s): Revoked investigation case sms text

oddsson:
# Revoked investigation case sms text

[Asana](https://app.asana.com/0/1199153462262248/1205268614076274/f)

## What

When cases are revoked, an SMS notification is sent to the court. This PR fixes a bug where that SMS said that a restriction case was revoked when an investigation case was revoked.

## Checklist:


unakb (approved):
Looks good 😄

PR #11906, opened 2023-08-22
fix(general-petition): Removing unused feature flags

albinagu:
## Checklist:


kksteini (approved):
Nice! :+1:

libs/application/templates/driving-license-duplicate/src/lib/drivingLicenseDuplicateTemplate.ts
```diff
@@ -18,7 +18,7 @@ import { Events, States, Roles } from './constants'
 import { dataSchema } from './dataSchema'
 import { m } from './messages'
 import { ApiActions } from './constants'
-import { FeatureFlagClient, Features } from '@island.is/feature-flags'
+import { FeatureFlagClient } from '@island.is/feature-flags'
 import {
   DrivingLicenseDuplicateFeatureFlags,
   getApplicationFeatureFlags,
```
saevarma, line 24: What about this?

And also L#76-87?

libs/application/templates/marriage-conditions/src/lib/MarriageConditionsTemplate.ts
```diff
@@ -16,7 +16,6 @@ import {
   DistrictsApi,
 } from '@island.is/application/types'
 import { assign } from 'xstate'
-import { Features } from '@island.is/feature-flags'
 import { getSpouseNationalId } from './utils'
 import { FeatureFlagClient } from '@island.is/feature-flags'
```
saevarma, line 20: What about this and L#60-70 ?

libs/application/templates/operating-license/src/lib/OperatingLicenseTemplate.ts
```diff
@@ -12,7 +12,7 @@ import {
 import { dataSchema } from './dataSchema'
 import { Roles, States, Events, ApiActions } from './constants'
 import { m } from './messages'
-import { FeatureFlagClient, Features } from '@island.is/feature-flags'
+import { FeatureFlagClient } from '@island.is/feature-flags'
```
saevarma, line 15: Ditto and the formloader below?

libs/application/templates/driving-license-duplicate/src/lib/drivingLicenseDuplicateTemplate.ts
albinagu: these are being used for testing on dev with fake data

libs/application/templates/marriage-conditions/src/lib/MarriageConditionsTemplate.ts
albinagu: keeping this for now, 
for testing on dev with fake data

libs/application/templates/operating-license/src/lib/OperatingLicenseTemplate.ts
albinagu: same as above

libs/application/templates/marriage-conditions/src/lib/MarriageConditionsTemplate.ts
saevarma: Okay, I see. Fake Data sounds like something that should always be an option in the dev environment and I feel like the feature flags is not the correct tool to use for that. You should look into the parental leave application, that renders on dev a prerequisite subsection that asks the user if he wants to use fake data or not.

saevarma (approved):
LGTM and nice clean up on feature flags.

But check out my comment for possible improvements on the fake data and further clean up on feature flags.

libs/application/templates/marriage-conditions/src/lib/MarriageConditionsTemplate.ts
albinagu: @saevarma 

Okay, parental leave application, however, is using the deprecated `isRunningOnEnvironment` check. The comment above it says "Any decisions based on the environment name should be rather based on feature flags and/or configuration." 

I would also want to be able to clean up feature flags thats are used for fake data, but im rather unsure at this point what the best way is. Perhaps good food for thought, to establish what we want to continue with.

saevarma: I missed that comment. Thanks for clarifying. I will take this up with Norda about general strategy on fake data in dev environment.

PR #11922, opened 2023-08-23
fix(service-portal): Update first reg date in get veichle search model

thordurhhh:
## What

Update the date field in search model to represent it's true value.

## Why

Otherwise it will be converted to a timestamp string, and therefore be different to other dates in the model.

## Checklist:


thorkellmani (approved):
camelCase?

PR #11928, opened 2023-08-24
checkpoint

oddsson:
# ...

## What

## Why

## Screenshots / Gifs

## Checklist:

### DIFF
--- a/apps/judicial-system/backend/src/app/modules/police/police.service.ts
+++ b/apps/judicial-system/backend/src/app/modules/police/police.service.ts
@@ -2,6 +2,7 @@ import { uuid } from 'uuidv4'
 import { Base64 } from 'js-base64'
 import { Agent } from 'https'
 import fetch from 'isomorphic-fetch'
+import { z } from 'zod'
 
 import {
   BadGatewayException,
@@ -172,6 +173,17 @@ export class PoliceService {
         if (res.ok) {
           if (this.config.policeCaseApiV2Available) {
             const response = await res.json()
+            const a = z.object({
+              rvMalSkjolMals_ID: z.string(),
+              heitiSkjals: z.string(),
+              malsnumer: z.string(),
+            })
+
+            const aa = z.object({
+              skjol: z.array(a),
+            })
+
+            aa.parse(response)
 
             return (
               response.skjol?.map(


apps/judicial-system/backend/src/app/modules/police/police.service.ts
```diff
@@ -37,8 +38,45 @@ import { PoliceCaseInfo } from './models/policeCaseInfo.model'
 export class PoliceService {
   private xRoadPath: string
   private agent: Agent
-
   private throttle = Promise.resolve({} as UploadPoliceCaseFileResponse)
+  private policeCaseFileStructure = z.object({
```
oddsson, line 42: @gudjong, do you have a preference where this validation should be? Here or in a separate file or somewhere else?

oddsson, line 42: Also, do you think we should only validate the things we are using from this API or validate the whole response, like we are doing here?

apps/judicial-system/backend/src/app/modules/police/police.service.ts
```diff
@@ -309,6 +339,7 @@ export class PoliceService {
         })
       })
       .catch((reason) => {
+        console.log(reason)
```
gudjong, line 342: Please remove console.log

apps/judicial-system/backend/src/app/modules/police/police.service.ts
gudjong: I think this is the perfect place.

gudjong: I think we should only validate what we are using. I expect the response to change over time and have a lot of stuff we are not using. We should tolerate changes to unused parts of the data.

apps/judicial-system/backend/src/app/modules/police/police.service.ts
```diff
@@ -171,43 +209,31 @@ export class PoliceService {
...
-              }) => ({
-                id: file.rvMalSkjolMals_ID,
-                name: file.heitiSkjals.endsWith('.pdf')
-                  ? file.heitiSkjals
-                  : `${file.heitiSkjals}.pdf`,
-                policeCaseNumber: file.malsnumer,
-                displayDate: file.dagsStofnad,
-              }),
-            )
+            const response: z.infer<typeof this.policeCaseFileStructure>[] =
```
gudjong, line 225: I think it would make more sense to vaildate the V2 structure and not the old one. Actually, this else case is being deleted in another PR.

gudjong (changes requested):
Please only validate fields we are using so that we tolerate changes to other parts of the data.

apps/judicial-system/backend/src/app/modules/police/police.service.ts
```diff
@@ -47,8 +48,26 @@ function getChapter(category?: string): number | undefined {
...
   private xRoadPath: string
   private agent: Agent
-
   private throttle = Promise.resolve({} as UploadPoliceCaseFileResponse)
+  private policeCaseFileStructure = z.object({
+    rvMalSkjolMals_ID: z.number(),
+    heitiSkjals: z.string(),
+    malsnumer: z.string(),
+    domsSkjalsFlokkun: z.string(),
+    dagsStofnad: z.string(),
```
gudjong, line 57: Please make domsSkjalaFlokkun and dagsStofnad optional. I have seen an example where the former is not included and I suspect the latter may also be excluded in some cases.

gudjong (approved):
One request for change

PR #11942, opened 2023-08-24
feat(portals): Create a standardised error handling in my-pages

snaerseljan:
# Standarizing errors

## What

On Mínar síður we have a couple of ways of showing users error messages. It can be for example an error box, inline error or a error toast. These error implementation are a common approaches of telling the user that something went wrong somewhere in the application pipeline. 

However, the errors that occur on Mínar síður are usually not the fault of the website itself, but rather in the underling system, namely some external service that the mínarsíður is using under the hood.

When the errors are displayed to the user the errors are common and the users most likely think that the error occurred on Mínar síður and therefore blame the website for being error prone and perhaps poorly implemented, instead of blaming the external system.

We want to standardise errors in the monorepo, so that all projects implement the same design and same implementation in code, i.e. components.

## Solution

A suggestion to solve the issue is to make our middle layer in our backend services return error messages with reference to what service actually failed. By doing that we can easily display to the user the name and icon of the service that actually failed within the users error message.

### Technical API solution

We suggest extending our middleware library enhanchedFetch  to return error messages with a unique key (organizationSlug) of the client(service) that failed. The unique key is the same organizationSlug that is used in Contentful organization content type, i.e. slug. The organizationSlug is passed with the error. It is passed if with the body if there is a problem. See server error response below

image

**Technical Client solution** 

Lets break it down into sections.

## island-ui

- Create a dynamic `<ProblemTemplate />` component that can handle design in a clean an dynamic way

## libs/react-spa/shared

- Create multiple components with the `<ProblemTemplate />` component from `island-ui`, i.e. InternalServerError, NotFound, NoData etc.
- Create `<Problem />` component that is the source of truth what kind of error should be rendered, i.e. InternalServerError, NotFound, etc.

# Error Types in SPA's

## ErrorBoundary

- Use new Problem component to display standardised 500 error message

## Module/Route error
- Create a reusable error component (<Problem />) that can be used in both route and module error. This component handles displaying the correct error messages, i.e. server error, service error, not found or custom errors
-  When an error happens within a route and the error has organisationSlug, then service error will render. This can happen deep within the application or in React Router loader/action pattern. This is kind of a fallback error.
- When we want to display the service error somewhere within the route then users will manually have to add the `<Problem/>` component.

## Component error

- Developers will manually add the <Problem /> component to display the desired error type.

## PR content split into bullets

- Added support for organizationSlug in enchancedFetch library
- Added organizationSlugs for every client, the client slugs are typed, 
- API client errors will pass the `organizationSlug` property within the `problem` object within the error. This will be handled automatically.
- Created `<ProblemTemplate />` component in UI library that standardises error components. This component is meant to be the main component that combines all the most common error designs
- Created multiple components in our`react-spa/shared` library. The spa-clients should only use one component thought which is the `<Problem />` component.
- New standardise error components have been implemented in both service-portal and admin-portal. **Note that we did not update the modules, except the delegation module.** We only implemented the infrastructure in the projects. Teams will have to update their own modules with the new components.
- We implemented new components for reference in one module, i.e. delegations module in service-portal.
- Implemented new error screen in our auth library and fixed showing loading spinner when signinRedirect errors to show the new error screen. See image below
- Added support to translate messages where modules are initialized.
- Add client not found to ids-admin, see image below 

### ApplicationErrorBoundary example
image

### Rendering error within a module or error happens within a loader function, (ModuleErrorScreen)
image

### Third party service error happening when loader function throws service error
image

### Not found page within mínar síður
image

### Form 500 error
image

### Form 500 third party service error small
image

### Form 500 third party service error large
image

Here are all the versions of the Problem component, first in code then in desktop and finally in mobile

### Code
image

### Desktop
errors_desktop

### Mobile
image

Identity server error has been updated to use the new problem component
image

When errors happens within a module, i.e. delegations
image

Third party service down error within a component
image

No data, with defaults
image

No data, with custom content
image

No data, with tag
image

Ids-admin client not found implementation
image

## Checklist:


libs/portals/core/src/screens/ModuleErrorScreen.tsx
```diff
@@ -1,30 +1,19 @@
...
+import { MessageDescriptor } from 'react-intl'
+
+import { Box } from '@island.is/island-ui/core'
+import { Problem } from '@island.is/react-spa/shared'
 
 interface ModuleErrorScreenProps {
   name: string | MessageDescriptor
 }
 
 export const ModuleErrorScreen = ({ name }: ModuleErrorScreenProps) => {
```
snaerseljan, line 11: What should we do with the route.name? Ignore it or just throw it away

GunnlaugurG (approved):
Great job 🥳

rafnarnason (approved):
lgtm

libs/clients/aircraft-registry/src/lib/apiConfig.ts
```diff
@@ -17,6 +17,7 @@ export const ApiConfig = {
     return new Configuration({
       fetchApi: createEnhancedFetch({
         name: 'clients-aircraft-registry',
+        organizationSlug: 'samgongustofa',
```
RunarVestmann, line 20: In the dev meeting it was discussed that the organizationSlug would be the english slug in Contentful but perhaps we've changed our minds and decided to use the Icelandic one instead

libs/island-ui/core/src/lib/ProblemTemplate/ProblemTemplate.tsx
```diff
@@ -0,0 +1,136 @@
...
+      textAlign="center"
+      dataTestId={dataTestId}
+      {...(expand && {
+        width: 'full',
+        height: 'full',
+        justifyContent: 'center',
+      })}
+    >
+      {tag && (
+        <div>
```
RunarVestmann, line 110: Does this div do anything useful or could we skip it?

RunarVestmann (commented):
Minor comments

thordurhhh (approved):
Love it 👏

Thorrik58 (approved):

libs/clients/aircraft-registry/src/lib/apiConfig.ts
snaerseljan: Un second though we have decided to have the icelandic slugs since developers recognise the Icelandic slugs better

RunarVestmann (approved):

libs/react-spa/shared/src/lib/messages.ts
saevarma, line 1: We need to provide the corresponding english translation in Contentful.

~ My OCD picked up that the `defaultMessage` and `id` props are swapped to what I'm normally used to read... except for the last element 🤪

libs/service-portal/vehicles/src/module.tsx
```diff
@@ -36,7 +36,7 @@ export const vehiclesModule: PortalModule = {
       name: m.yourVehicles,
       path: VehiclePaths.AssetsMyVehicles,
       enabled: userInfo.scopes.includes(ApiScope.vehicles),
-      loader: translationLoader({ userInfo, client }),
+      loader: translationLoader({ userInfo, ...rest }),
```
saevarma, line 39: ~Use this pattern for other modules where you used props instead of object destructuring?

libs/island-ui/core/src/index.ts
```diff
@@ -58,6 +58,8 @@ export * from './lib/PdfViewer/PdfViewer'
 export * from './lib/PageLoader'
 export * from './lib/VisuallyHidden/VisuallyHidden'
 export * from './lib/Drawer/Drawer'
+export * from './lib/ProblemTemplate/ProblemTemplate'
+export * from './lib/ProblemTemplate/ProblemTemplate.css'
```
saevarma, line 62: We don't need to export the css? Unless I missed something 🙈

saevarma (approved):
Amazing job 🙌

libs/react-spa/shared/src/components/problems/ThirdPartyServiceError.tsx
```diff
@@ -0,0 +1,56 @@
...
+  organizationSlug,
+  size = 'large',
+  ...rest
+}: ThirdPartyServiceErrorProps & TestSupport) => {
+  const { formatMessage } = useLocale()
+  const { data, error, loading } = useGetOrganizationsQuery()
+
+  const organization = useMemo(
+    () =>
+      data?.getOrganizations.items.find((org) => org.slug === organizationSlug),
```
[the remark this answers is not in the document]
snaerseljan: Really good point @RunarVestmann. We will let the Icelandic translations due for now, but we will definitely keep this in mind for future endeavours

libs/auth/react/src/lib/auth/AuthErrorScreen.tsx
```diff
@@ -1,36 +1,38 @@
...
+      variant="error"
+      expand
+      tag="Villa"
+      title="Innskráning mistókst"
+      message={
+        <>
+          Vinsamlegast reyndu aftur síðar.{' '}
+          <Button variant="text" onClick={signIn}>
+            Aftur í innskráningu
+          </Button>
```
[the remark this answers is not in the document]
snaerseljan: Lets discuss offline in a call

libs/clients/inna/src/lib/apiConfig.ts
```diff
@@ -14,6 +10,7 @@ export const ApiConfig = {
     new Configuration({
       fetchApi: createEnhancedFetch({
         name: 'clients-inna',
+
```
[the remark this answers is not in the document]
snaerseljan: Responsible party? I could not find Inna nor Námsferlaveita. What client is this in Contentful?

libs/portals/shared-modules/delegations/src/lib/messages.ts
```diff
@@ -121,7 +121,8 @@ export const m = defineMessages({
   },
   confirmError: {
     id: 'sp.access-control-delegations:confirm-error',
-    defaultMessage: 'Ekki tókst að vista réttindi. Vinsamlegast reyndu aftur',
+    defaultMessage:
+      'Ekki tókst að vista réttindi. Vinsamlegast reyndu aftur síðar',
```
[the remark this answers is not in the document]
snaerseljan: ✅

libs/island-ui/core/src/index.ts
snaerseljan: This is specifically imported in NoData component

libs/clients/inna/src/lib/apiConfig.ts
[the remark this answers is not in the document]
snaerseljan: Follow up task has been created 
https://www.notion.so/Add-Organization-slugs-to-clients-that-are-missing-slug-bd5081b1492f4ed597b22edf8967ddf7?pvs=4

libs/auth/react/src/lib/auth/AuthErrorScreen.tsx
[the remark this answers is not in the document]
snaerseljan: Fixed

davidpxn (approved):

karlarnar (approved):

PR #11965, opened 2023-08-28
chore(core): Update prettier to v2.5

oddsson:


.yarn/releases/yarn-3.2.3.cjs
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env node
 /* eslint-disable */
 //prettier-ignore
```
snaerseljan, line 3: This file should probably be ignored
Maybe try to ignore it in `.prettierignore` or replace the prettier-ignore comment with
`/* prettier-ignore */`

snaerseljan (approved):
LGTM. Great job 👍 

You should look into `.yarn/releases/yarn-3.2.3.cjs` to be ignored.
I also propose that you add a feature flag tag and try out at least some projects like, Mínar síður, the web project and admin portal Just to verify that there aren't any runtime errors. I find it very very unlikely though.

saevarma:
Admin merging this PR as it contains only automatically touched file by prettier update.

PR #11966, opened 2023-08-28
feat(j-s): Indictment LÖKE File Chapters

gudjong:
# Indictment LÖKE File Chapters

[LÖKE - Forflokkun skjala úr LÖKE yfir Í RVG svo auðveldara sé að raða í skjalskrá](https://app.asana.com/0/1199153462262248/1204853378242937/f)

## What

- Forflokkar skjöl frá LÖKE í kafla í ákærum.
- Lagar villu í drag and drop endurröðun.
- Notar LÖKE file id til að ákvarða hvaða skjölum hefur þegar verið hlaðið úr LÖKE.

## Why

- Forflokkun sparar tíma fyrir notendur.

## Checklist:


apps/judicial-system/api/src/app/modules/police/models/policeCaseFile.model.ts
```diff
@@ -13,6 +13,9 @@ export class PoliceCaseFile implements TPoliceCaseFile {
   @Field()
   readonly policeCaseNumber!: string
 
-  @Field()
-  readonly displayDate!: string
+  @Field({ nullable: true })
+  readonly chapter?: number
+
+  @Field({ nullable: true })
```
oddsson, line 19: I thought this date field was non nullable, i.e. I thought it was always set. Is there a reason we are making it optional?

apps/judicial-system/backend/src/app/modules/police/police.service.ts
```diff
@@ -33,6 +33,29 @@ import { UploadPoliceCaseFileResponse } from './models/uploadPoliceCaseFile.resp
 import { policeModuleConfig } from './police.config'
 import { PoliceCaseInfo } from './models/policeCaseInfo.model'
 
+function getChapter(category?: string): number | undefined {
+  if (!category || category.length < 2) {
+    return undefined
+  }
+
+  switch (category.substring(0, 2)) {
```
oddsson, line 41: This is a bit too hard coded in my opinion. Consider casting the chapter number to an int and return `chapterNumber - 1`. If the type casting fails or returns a `NaN`, return `undefined`. That way we don't need to change this code if a chapter is added or removed in the future.

oddsson (approved):
Neat! Couple of things that to consider.

apps/judicial-system/api/src/app/modules/police/models/policeCaseFile.model.ts
gudjong: Prombem is, we don't know. The data from LÖKE may or may not contain these new fields and by keeping them optional we avoid potential runtime errors.

PR #11976, opened 2023-08-29
chore(j-s): Use generated CaseCustodyRestrictions type

oddsson:
# Use generated CaseCustodyRestrictions type

[Asana](https://app.asana.com/0/1199153462262248/1205359398381996/f)

## What

Use generated CaseCustodyRestrictions type

## Why

We are moving towards using generated GraphQL types in the front end

## Checklist:

### DIFF
--- a/apps/judicial-system/web/src/types/index.ts
+++ b/apps/judicial-system/web/src/types/index.ts
@@ -9,6 +9,7 @@ import {
   CaseAppealState,
   CaseAppealRulingDecision,
   Defendant,
+  CaseCustodyRestrictions,
 } from '@island.is/judicial-system-web/src/graphql/schema'
 import {
   Case,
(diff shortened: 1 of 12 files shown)


apps/judicial-system/web/src/types/index.ts
```diff
@@ -238,6 +239,7 @@ export interface TempCase
     | 'appealedByRole'
     | 'appealRulingDecision'
     | 'defendants'
+    | 'CaseCustodyRestrictions'
```
unakb, line 242: Should this not be requestedCustodyRestrictions?

apps/judicial-system/web/src/types/index.ts
```diff
@@ -252,6 +254,7 @@ export interface TempCase
   appealedByRole?: UserRole
   appealRulingDecision?: CaseAppealRulingDecision
   defendants?: Defendant[]
+  custodyRestrictions: CaseCustodyRestrictions[]
```
unakb, line 257: and this as well I think

apps/judicial-system/web/src/types/index.ts
oddsson: Yeb! Thanks, fixed

unakb (approved):
Nice 👍

PR #11982, opened 2023-08-30
feat(web): Make SectionWithVideo title optional

RunarVestmann:
# Make SectionWithVideo title optional

## What

* There are cases where we don't want to render the title field so I added a boolean prop indicating whether we want to show it or not as well as a way to turn the divider on top on or off
* I also added support for embedding forms in the content field for anchor pages (was requested by Digital Iceland)

## Checklist:


Toti91 (approved):
Codeowned file looks good

siggifv (approved):

PR #11985, opened 2023-08-30
chore(service-portal): outbound link clicks

thordurhhh:
## What

Change the `Outbound Link: Click` goal to be opt in, instead of automatically tracking every outbound click.

## Why

Now we are tracking every outbound click. Masking the url is not an option when using the `outbound-links` script directly.
So this way we are choosing where to track the external click and what props should be sent (masked paths).

## Checklist:


thorkellmani (approved):
Very cool and nice 💅

snaerseljan (approved):
LGTM! Well done 👍

PR #11994, opened 2023-08-30
fix(ids-api): Fix flaky test in ids-api

snaerseljan:
# Fix flaky test in ids-api

## What

Attempt to fix flaky test in ids-api. Most likely the reason for this flaky test is that faker will not generate unique scope names. At least that is my presumption.

This PR will ensure that those names are unique since they are used as primary keys in the scope model

## Why

Our CI has some flaky tests, i.e. https://github.com/island-is/island.is/actions/runs/5761331315/job/15619298536

## Checklist:

### DIFF
--- a/apps/services/auth/ids-api/test/setup.ts
+++ b/apps/services/auth/ids-api/test/setup.ts
@@ -1,3 +1,6 @@
+import { getModelToken } from '@nestjs/sequelize'
+import { TestingModuleBuilder } from '@nestjs/testing'
+
 import {
   ApiScope,
   Domain,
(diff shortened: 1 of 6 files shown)


apps/services/auth/ids-api/test/setup.ts
```diff
@@ -33,16 +36,32 @@ type Scopes = {
...
+
+const randomWords = createUniqueWords(2)
 
 export const defaultScopes: Scopes = {
   // Test user has access to this scope
   testUserHasAccess: {
     name: '@identityserver.api/authentication',
     domainName: defaultDomains[0].name,
   },
-  scope1: { name: faker.lorem.word(), domainName: defaultDomains[0].name },
```
saevarma, line 44: ~`faker` import becomes unused.

saevarma (approved):
This is a sneaky bug if we have a parallel race condition calling faker and getting the same name. Lets try this out. We should also look into upgrade Faker.

PR #11996, opened 2023-08-30
fix(service-portal): Optimize our SPA bundles size

snaerseljan:
# Optimize our SPA bundles

## What

Our bundles are including data from our api mocks in production. That should not be included.
The removal of those exports in minimized the main.js bundle by 230KB

**Before**
main.js 6.92 MiB

**After**
main.js **6.69** MiB

## Why

Smaller bundles = faster load time

## Checklist:


saevarma (approved):
🙌

PR #12001, opened 2023-08-31
chore(githook): Make helm chart hook less aggressive

AndesKrrrrrrrrrrr:
The check was overly aggressive, generating charts if _any_ file in `/infra` was changed.
- Change helm charting condition to only chart on app chart changes
- Do charting in the background instead of blocking commit execution

## Checklist:


saevarma (approved):
LGTM

PR #12011, opened 2023-09-01
feat(clients-middlewares): Create helper function to add header value to cache key for withCache

saevarma:
## What

Adds a new function `defaultCacheKeyWithHeader` that accepts a header name and appends it to the default calculated cacheKey based on the request url and search params.

## Why

We have put it in our API Design Guide that APIs should use a static placeholder for resourceIds and read the actual resource ID from a header value for PII reasons.

This breaks the current implementation for the `defaultCacheKey` when client libraries are using cache in `enhancedFetch`.

## Screenshots / Gifs

N/A

## Checklist:


snaerseljan (approved):
Nice catch. What a man! 🥇

PR #12073, opened 2023-09-07
fix(service-portal): Adding scope, fixing messages

jonbjarnio:
Added api scope to routes, fixed wrong message in navigation and fixed messages file

## What

Enable extract-strings script to correctly extract strings from messages file and adding scopes to routes.

## Why

To display correct messages and have correct scopes.

## Checklist:


libs/service-portal/core/src/lib/messages.ts
```diff
@@ -506,7 +506,10 @@ export const m = defineMessages({
     defaultMessage:
       'Hér er markmiðið að þú getir fundið yfirlit yfir leyfisbréf og vottorð til starfsréttinda. Unnið er að því að koma öllum leyfisbréfum og vottorðum um starfsréttindi á einn stað.',
   },
-
+  occupationalLicensesNavIntro: {
+    id: 'service.portal:occupational-licenses-description',
```
thordurhhh, line 510: Same id twice 😎

thordurhhh (approved):

PR #12120, opened 2023-09-14
feat(search-indexer): Use feature flag instead of env variable for turning nested entry resolving on or off

RunarVestmann:
# Use feature flag instead of env variable for turning nested entry resolving on or off

## What

* We are storing CMS data in Elasticsearch so we can search through it and provide a backup in case the CMS goes down.
* The CMS calls an endpoint on the search-indexer service when content changes but when something nested gets changed (like an accordion) then we need to traverse up the tree to find the page (indexable entry) that we need to re-index so the update is displayed on the web. But this process is especially slow if something nested gets updated with a lot of backlinks then (with the current implementation) it takes a long time and can even make the search-indexer run out of memory. So when something like that happens it's good to be able to turn off the nested resolution so new content actually starts appearing on the web again

## Checklist:

### DIFF
--- a/apps/services/search-indexer/infra/search-indexer-service.ts
+++ b/apps/services/search-indexer/infra/search-indexer-service.ts
@@ -1,4 +1,4 @@
-import { ref, service, ServiceBuilder } from '../../../../infra/src/dsl/dsl'
+import { service, ServiceBuilder } from '../../../../infra/src/dsl/dsl'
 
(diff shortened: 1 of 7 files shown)


peturgq (approved):

saevarma (approved):
LGTM

ingolfurs (approved):

PR #12131, opened 2023-09-15
fix(auth-admin): Fix replica issue when creating client

GunnlaugurG:
# ...

## What

Query client from primary database

## Why

The create endpoint for client often fails because it is created on main database, but the response is queried from a replica database. In most cases the replica does not yet have the instance that was created and returns a NoContentException

## Screenshots / Gifs

## Checklist:


saevarma (approved):
Nice catch 🐞

PR #12138, opened 2023-09-15
chore(driving-license-client): Refactor into only v4/v5

kksteini:
# What 
Rewriting the driving license backend/service to only use v4 and v5 clients instead of having the mix of v1,v2,v4 and ,v5.
V4 is for unauthenticated requests when needed, V5 is token based.

# Why?
The third party can stop maintaining the legacy v1/v2 and focus entirely on v4/v5 instead.

Note: there is a legacy version of a method to get the current license but no structural changes are needed a.t.m since the error behaviour of v4 stays the same when compared to v1/v2.


libs/api/domains/driving-license/src/lib/drivingLicense.service.ts
```diff
@@ -242,10 +242,13 @@ export class DrivingLicenseService {
     nationalId: string,
     type: DrivingLicenseApplicationType,
   ): Promise<ApplicationEligibility> {
-    const assessmentResult = await this.getDrivingAssessmentResult(nationalId)
+    const token = user.authorization.toLowerCase().startsWith('bearer')
+      ? user.authorization.split(' ')[1]
+      : user.authorization // removes the Bearer prefix
+    const assessmentResult = await this.getDrivingAssessmentResult(token)
```
kksteini, line 248: Note to self:
Consider trying to standardize these?
Actually it would be better to explicitly call `replace` on 'Bearer' I think

libs/clients/driving-license/src/lib/drivingLicenseApi.service.ts
```diff
@@ -318,31 +291,47 @@ export class DrivingLicenseApi {
...
-    const res = await this.v1.apiOkuskirteiniKennitalaFinishedokugerdiGet({
-      kennitala: params.nationalId,
+    const res = await this.v5.apiDrivinglicenseV5Hasfinisheddrivingschool3Get({
+      apiVersion: v5.DRIVING_LICENSE_API_VERSION_V5,
+      apiVersion2: v5.DRIVING_LICENSE_API_VERSION_V5,
+      jwttoken: params.token.replace('Bearer ', ''),
     })
 
-    return (res.hefurLokidOkugerdi ?? 0) > 0
+    return Boolean(res.hasFinishedDrivingSchool3)
```
kksteini, line 322: Return type is `boolean | undefined` which should be ok given that

```
Boolean(undefined) => false,
Boolean(true) => true,
Boolean(false) => false
```

rafnarnason (approved):

busla (approved):
LGTM

obmagnusson (approved):
LGTM

PR #12149, opened 2023-09-18
chore(national-registry): Remove fate of child from service

thordurhhh:
## What

Remove fate (`Afdrif`) from Soffia service.

## Why

It is unreliable and therefore the service provider does not want island.is to use it.

## Checklist:


birkir (approved):
Looks good. We need to schedule a new build for this.

thordurhhh:
> Looks good. We need to schedule a new build for this.

Alright, would it be OK if we merge this and release it in the next release (10th october)

PR #12152, opened 2023-09-18
fix(infra): allow pvc volume refs accross services

busla:
# Remove GEOIP db fetch on build time and use a cronjob

## Current setup 

GEOIP database is downloaded and bundled with each Docker image. Let's add a k8s cronjob to update it every 30 mins and store it in a shared PVC.

This has the following sideaffects:
- NX is unable to cache the builds
- services-sessions is the only service using Maxmind GEOIP
- The GEOIP db is not updated until the next build.

## New setup

This PR adds support in infra for a service to reference an existing PVC defined by another service with the `useExisting` attr when volumes are defined.

- fix: import configcat-node typo
- build: run geoip-lite as initContainer instead of build-time
- build: remove GEOIP_LICENSE_KEY references, no more stateful builds
- fix(infra): allow pvc volume refs accross services

# ...

## What

## Why

## Screenshots / Gifs

## Checklist:


charts/islandis/values.dev.yaml
```diff
@@ -2265,7 +2265,13 @@ services-sessions:
           paths:
             - '/'
   namespace: 'services-sessions'
-  pvcs: []
+  pvcs:
+    - accessModes: 'ReadWriteMany'
+      mountPath: '/webapp/node_modules/geoip-lite/data'
```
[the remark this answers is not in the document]
busla: I was hoping for an env var from the library but the docs didn't mention it. But when looking at the source code they actually do :tada:

https://github.com/geoip-lite/node-geoip/blob/main/scripts/updatedb.js#L36

infra/src/dsl/types/output-types.ts
```diff
@@ -4,6 +4,7 @@ import { ReferenceResolver, EnvironmentConfig } from './charts'
 // Output types
 export type ContainerRunHelm = {
   command: string[]
+  image?: string
```
peturgq, line 7: Seems to be missing as an input into the helm chart.

infra/src/dsl/types/input-types.ts
```diff
@@ -149,6 +149,7 @@ export type PersistentVolumeClaim = {
   name?: string
   size: '1Gi' | '5Gi' | '10Gi' | string
   accessModes: AccessModes
+  useExisting?: boolean
```
peturgq, line 152: We should maybe discuss if this shouldn't just be a reference to the volume object.

peturgq (commented):
Some concerns

charts/islandis/values.dev.yaml
```diff
@@ -2227,6 +2227,7 @@ services-sessions:
     DB_NAME: 'services_sessions'
     DB_REPLICAS_HOST: 'dev-vidspyrna-aurora.cluster-ro-c6cxecmrvlpq.eu-west-1.rds.amazonaws.com'
     DB_USER: 'services_sessions_read'
+    GEODATADIR: '/geoip-lite/data'
```
AndesKrrrrrrrrrrr, line 2230: Consider changing this to what it effects. If the data is `lite`, it should say so in the filename. This way you could add `hevy` data in same folder.
```suggestion
    GEODATADIR: '/geoip/data'
```

charts/islandis/values.dev.yaml
```diff
@@ -2280,9 +2287,77 @@ services-sessions:
...
+    - 'node'
+  enabled: true
+  env:
+    GEODATADIR: '/geoip-lite/data'
+    NODE_OPTIONS: '--max-old-space-size=208'
+    REDIS_URL_NODE_01: '["clustercfg.general-redis-cluster-group.5fzau3.euw1.cache.amazonaws.com:6379"]'
+    SERVERSIDE_FEATURES_ON: ''
+  grantNamespaces:
+    - ''
+  grantNamespacesEnabled: true
```
AndesKrrrrrrrrrrr, line 2308: If the list is empty, why grant?
```suggestion
  grantNamespacesEnabled: false
```

charts/islandis/values.dev.yaml
```diff
@@ -2280,9 +2287,77 @@ services-sessions:
...
+      initialDelaySeconds: 3
+      path: '/'
+      timeoutSeconds: 3
+  hpa:
+    scaling:
+      metric:
+        cpuAverageUtilization: 70
+        nginxRequestsIrate: 5
+      replicas:
+        max: 3
```
AndesKrrrrrrrrrrr, line 2324: In what cases can this possibly need more than 1?

charts/islandis/values.dev.yaml
busla: the library is called geoip-lite, not geoip.

charts/islandis/values.dev.yaml
busla: this is a readonly generated file from the DSL

infra/src/dsl/types/input-types.ts
busla: the ref() is incomplete, let's iterate on that later.

infra/src/dsl/types/output-types.ts
busla: I am not following

charts/islandis/values.dev.yaml
busla: never, let me take a look in the DSL

saevarma (approved):
Great improvements!

apps/services/sessions/infra/sessions.ts
```diff
@@ -1,8 +1,17 @@
-import { json, service, ServiceBuilder } from '../../../../infra/src/dsl/dsl'
+import { PersistentVolumeClaim } from '../../../../infra/src/dsl/types/input-types'
+import { ref, service, ServiceBuilder } from '../../../../infra/src/dsl/dsl'
```
brynjarorng, line 2: Is the ref import needed? Looks possibly unused

apps/services/sessions/infra/sessions.ts
```diff
@@ -99,3 +116,37 @@ export const workerSetup = (): ServiceBuilder<'services-sessions-worker'> =>
...
+        'license_key=$(GEOIP_LICENSE_KEY)',
+      )
+      .resources({
+        limits: {
+          cpu: '200m',
+          memory: '256Mi',
+        },
+        requests: {
+          cpu: '10m',
+          memory: '128Mi',
```
brynjarorng, line 140: Is the delta between the cpu limits not quite high? This could possibly cause issues with over provisioning when many nodes are starting up such as in deployments or large scale ups

infra/src/dsl/types/output-types.ts
busla: @peturgq bump

apps/services/sessions/infra/sessions.ts
busla: well spotted, I was attempting to use `ref()` instead of passing the service as a param.

apps/services/sessions/infra/sessions.ts
busla: you think that also affects cronjobs?

infra/src/dsl/types/output-types.ts
peturgq: Here you're only setting the initcontainer image to be the same as the service image https://github.com/island-is/island.is/blob/758a34831f77b57df9813cf760a8481266fa3529/infra/helm/libs/api-template/templates/deployment.yaml#L72

So we need something in the template that reads and sets the initcontainer image from the init container value.

apps/services/sessions/infra/sessions.ts
brynjarorng: Hmm, the limits do affect cron jobs, maybe we should just see how much this actually requires and tune it after the fact? I still think 10m cpu is quite small. Maybe put this up to 150? That allows the container to use 0.15% of a CPU core. I do however suspect it will require more but that can just be tuned later

infra/src/dsl/types/output-types.ts
busla: Ahh!

Well spotted, that I didn't see :pray:

busla: let's keep it in scope, see updated template

PR #12158, opened 2023-09-18
feat(web): Transport authority ship search on public web

RunarVestmann:
# Transport authority ship search on public web

## What

* Move https://www.samgongustofa.is/siglingar/skrar-og-utgafa/skipaskra/uppfletting to Ísland.is

## Screenshots / Gifs

image

## Checklist:


busla (approved):
LGTM

siggifv (approved):

libs/clients/ship-registry/src/lib/ship-registry.service.ts
```diff
@@ -0,0 +1,96 @@
...
+    postalCode: string
+  }
+}
+
+const getValueOrEmptyString = (value?: string | number | null) => {
+  return value ? value : ''
+}
+
+const mapRegion = (ship: Ship) => {
+  return `${getValueOrEmptyString(ship.umdaemisbokstafir)}${
```
snaerseljan, line 49: This method is not very readable. Definitely consider to split up, store in variables, etc.

libs/clients/ship-registry/src/lib/ship-registry.service.ts
```diff
@@ -0,0 +1,96 @@
...
+  }
+}
+
+@Injectable()
+export class ShipRegistryClientService {
+  constructor(private readonly shipApi: ShipApi) {}
+
+  async findShipByNameOrNumber(input: FindShipByNameOrNumberRequest) {
+    try {
+      const data = (await this.shipApi.findShipByNameOrNumber(
```
snaerseljan, line 82: Hmmm is this not typed? It should be generated

snaerseljan (approved):
Nice work! A few comments that would be good to look at.

libs/clients/ship-registry/src/lib/ship-registry.service.ts
RunarVestmann: The generated types were off due to the swagger being incorrect, I've told the devs about this and this is a workaround for now

libs/api/domains/ship-registry/src/lib/dto/ship-search.input.ts
```diff
@@ -0,0 +1,7 @@
+import { Field, InputType } from '@nestjs/graphql'
+
+@InputType('ShipRegistryShipSearchInput')
+export class ShipSearchInput {
+  @Field(() => String)
+  qs!: string
```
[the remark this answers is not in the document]
RunarVestmann: Since the xroad endpoint calls it qs I for whatever reason just used the same name here, I could've used something else but for the sake of time I'd like to keep it like this if that's alright

PR #12162, opened 2023-09-19
fix(auth-admin): Fix flaky client creation

GunnlaugurG:
# ...

## What

Creating client often fails to create secret, since there is a fail in fetching the client after creating. 

## Why

So the client secret will be created right after the client creation

## Screenshots / Gifs

## Checklist:


saevarma (approved):
LGTM 🙌

PR #12165, opened 2023-09-19
ci: ignore schemas cache on label

busla:
# skip generated files cache on PR label

Allow setting `skip-generated-cache` label on PRs to force schemas generation
## What

run code generators, e.g. `yarn schemas` regardless of cache.

## Why

There have been incidents that cache has been fetched even though generated files have been modified. Setting the `skip-generated-cache` should be considered as a debugging tool for fixing the root issue.


snaerseljan (approved):

AndesKrrrrrrrrrrr (approved):
LGTM

PR #12194, opened 2023-09-21
fix(national-registry-v3): Update spouse model marital status text

thordurhhh:
## What

`maritalStatus` can return any string or a specific string.

## Why

This is a temporary solution until we remove v1 and replace with v3. 
Then we will always return `String`

## Checklist:


jonbjarnio (approved):
LGTM

PR #12205, opened 2023-09-22
fix(j-s): Form stepper

oddsson:
# Form stepper

[Asana](https://app.asana.com/0/1199153462262248/1204638155577489/f)

## What

This PR does two things to the form stepper

- Show extension sections before appeal sections 
- Fix an issue where the extension section was active, even in completed cases. This only happened to prosecutor roles. 

## Why

It's more logical to show the extension section before the appeal sections because if a case is extended, you are appealing the extension.

## Checklist:


unakb (approved):
Makes sense

PR #12207, opened 2023-09-22
feat(core): Scan and upload feature flags

baering:
## What

Scans repo for feature flag usages and uploads references of usage to ConfigCat to be able to view in UI

## Why

Feature flags allow us to enable or disable features without code changes. Once a feature is fully developed the flag can be safely removed from code, but we have to remember to do that. This PR adds a scan job to the PR workflow to find usage of feature flags so that we can see when they are not being used from the ConfigCat UI.

## Screenshots / Gifs

image

image

## Checklist:


busla (changes requested):
Is there a way to test trigger the action before we merge? The `scanflags` step is skipped in both runs here:

https://github.com/island-is/island.is/actions/workflows/pullrequest.yml?query=branch%3Afeature%2Fscan-and-upload-feature-flags

baering:
> Is there a way to test trigger the action before we merge, maybe a `--dry-run` arg exist? The `scanflags` step is skipped in both runs here:
> 
> https://github.com/island-is/island.is/actions/workflows/pullrequest.yml?query=branch%3Afeature%2Fscan-and-upload-feature-flags

I've tested it locally using [act](https://github.com/nektos/act) where I just removed the `if` condition that checks if it is a PR merge. The reason I have that condition is to avoid running it all the time. It runs without issues using a local github action environment.

busla (approved):
LGTM

PR #12226, opened 2023-09-25
fix(j-s): Display "new" tag on list of COA cases

unakb:
[Tögg á málalista Landsréttar - bæta við nýtt og breyta virkni á móttekið](https://app.asana.com/0/1199153462262248/1205520219032813/f)

## What

Added "new" tag to COA case list

## Why

So that the court of appeals staff can easily see whether a case has been received by them or not

## Screenshots / Gifs

image

## Checklist:


apps/judicial-system/web/src/components/TagAppealState/TagAppealState.tsx
```diff
@@ -44,10 +49,19 @@ const TagAppealState: React.FC<React.PropsWithChildren<Props>> = ({
       }
     }
     if (state === CaseAppealState.RECEIVED) {
-      return {
-        color: 'darkerBlue',
-        text: formatMessage(tables.receivedTag),
-      }
+      if (
+        user?.institution?.type === InstitutionType.COURT_OF_APPEALS &&
+        !appealCaseNumber
```
oddsson, line 54: I prefer positive checks in if statements, i.e.

`user?.institution?.type === InstitutionType.COURT_OF_APPEALS && appealCaseNumber ? received : new`

What do you think?

oddsson (approved):
One minor thing

apps/judicial-system/web/src/components/TagAppealState/TagAppealState.tsx
unakb: Actually on second thought, it makes more sense to keep it this way because this is the only specific case where we want to display the purple new tag. In all other cases we want to default to the second, also cases that aren't being viewed by InstitutionType.COURT_OF_APPEALS, so to make a positive check work I'd have to complicate the code more.

PR #12271, opened 2023-09-29
fix(air-discount-scheme): default lifetime of code and added type of code

Tryggvig:
## What

Allow admins to manually create connecting flight codes

## Why

Because Umboðsmaður Alþingins had comments regarding Loftbrú

## Screenshots / Gifs

## Checklist:


libs/portals/admin/air-discount-scheme/src/screens/Overview/Overview.loader.ts
```diff
@@ -18,16 +19,18 @@ import {
...
+    return start ? startOfDay(value) : endOfDay(value)
   }
 
   if (isValidDate(new Date(val))) {
     return val
   }
-
-  return parseDate(val, 'dd.MM.yyyy', endOfDay(new Date()))
+  const value = parseDate(val, 'dd.MM.yyyy', new Date())
+  return start ? startOfDay(value) : endOfDay(value)
```
saevarma, line 33: This is exactly the same is in the first if statement. And why is L30 only returning val which is string when other paths return either startOf or endOf day which is a date object.
What is the goal of this function and I feel like it could be refactored to be more simple and readable.

Tryggvig: Was originally supposed to handle if someone was setting a wrong date manually into query string, but lets move from that, now it just returns the string back if the date is not valid.

libs/portals/admin/air-discount-scheme/src/screens/Overview/Overview.loader.ts
```diff
@@ -72,8 +75,9 @@ export type OverviewLoaderReturnType = {
 
 export const overviewLoader: WrappedLoaderFn = ({ client }) => {
   return async ({ request }): Promise<OverviewLoaderReturnType> => {
+    console.log(request)
     const input = validateSearchParams({ request, schema })
-
+    console.log(input)
```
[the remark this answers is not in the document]
Tryggvig: console.lol

saevarma (approved):
LGTM

PR #12222, opened 2023-09-25
feat(j-s): Send SMS to court assistants when a case is appealed

unakb:
[Senda sms í vaktsímann hjá dómritara þegar kært er utan þinghalds ](https://app.asana.com/0/1199153462262248/1205443167661876/f)

## What

Add an SMS notification trigger when a case is appealed

## Why

So we can notify the court assistant or judge on duty by SMS that a case has been appealed

## Screenshots / Gifs

Attach Screenshots / Gifs to help reviewers understand the scope of the pull request

## Checklist:

- [x] I have performed a self-review of my own code
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] Formatting passes locally with my changes
- [x] I have rebased against main before asking for a review

PR #12280, opened 2023-10-02
fix(j-s): Send case notification e-mail to defender 

unakb:
[Senda verjanda póst við úthlutun fyrirtökutíma jafnvel þótt boð í fyrirtöku sé ekki sent. ](https://github.com/island-is/island.is/pull/12222)

## What

Send case e-mail notification to defender when court date is set, even if calendar invite is not sent

## Why

The defender should still receive a notification about the case even though they dont get a calendar invite 

## Checklist:


apps/judicial-system/backend/src/app/modules/notification/notification.service.ts
```diff
@@ -1925,7 +1966,15 @@ export class NotificationService {
...
-            messages = []
+            // We still want to send the defender a link to the case even if
+            // the judge chooses not to send a calendar invitation
+            messages = [
+              this.getNotificationMessage(
+                MessageType.SEND_DEFENDER_ASSIGNED_NOTIFICATION,
+                user,
+                theCase,
+              ),
+            ]
```
unakb, line 1977: I'm not sure how I feel about doing this this way, but I didn't really have an idea of a cleaner way to do it without making the code messier or adding en entirely new event just to handle this specific case, which felt like overkill?
Feedback appreciated though. I'm a little wary of the code around notifications, it feels like it's getting harder to maintain and keep track of

gudjong (approved):

apps/judicial-system/backend/src/app/modules/notification/notification.service.ts
gudjong: Your implementation fits well with the current notification design. I agree that notifications are getting pretty complicated. My main concern is with the granularity of notifications and consequently our ability to retry failed notifications. At the moment, we basically say that a notification succeeds if we managed to successfully send a notification to one of the participants. A more robust approach would treat notifications to individual recipients separately. This would most likely also address some of your concerns with maintainability.

PR #12297, opened 2023-10-04
build(infra): improve --max-old-space-size default calculation

busla:
# --max-old-space-size defaults and reduce sessions-geoip worker resource limit

Default NODE_OPTIONS=--max-old-space-size in infra service serializer is statically subtracting -48 from the user defined resources.limits.memory

Let's allow 90% of the memory limit.

Also, sessions-geoip-worker is requesting too much cpu/mem, let's request only 1cpu/2Gi

---
  - https://app.asana.com/0/0/1205649089148846


infra/src/dsl/output-generators/utils.ts
```diff
@@ -0,0 +1,13 @@
...
+ *
+ * @param v A string representing a numeric value.
+ * @returns The floored result of 90% of the given number.
+ * @example
+ *
+ * getMaxOldSpaceSizeRatio("100");  // Returns 90
+ * getMaxOldSpaceSizeRatio("2048"); // Returns 1843
+ */
+export const getMaxOldSpaceSizeRatio = (v: string) => {
+  return Math.floor(parseInt(v, 10) * (9 / 10))
```
AndesKrrrrrrrrrrr, line 12: Maybe allow an optional ratio parameter?
Maybe change to `function` instead of `const`?

AndesKrrrrrrrrrrr (approved):

brynjarorng (approved):
LGTM

saevarma (approved):
LGTM

PR #12299, opened 2023-10-04
feat(j-s): Defender access changes

unakb:
[Aðgengi allra lögmanna (skráðir í lögmannaskrá) ](https://app.asana.com/0/1199153462262248/1205581325882016)

## What

Check lawyer registry when defenders without cases attempt to log in 

## Why

So the log in experience is better for defenders with no active cases

## Checklist:


apps/judicial-system/api/src/app/modules/auth/auth.controller.ts
```diff
@@ -162,6 +161,7 @@ export class AuthController {
 
       if (verifiedUserToken) {
         this.logger.debug('Token verification successful')
+        res.clearCookie(CODE_VERIFIER_COOKIE.name, CODE_VERIFIER_COOKIE.options)
```
gudjong, line 164: 🧹

apps/judicial-system/api/src/app/modules/auth/auth.service.ts
```diff
@@ -1,17 +1,20 @@
 import fetch from 'isomorphic-fetch'
 import jwt from 'jsonwebtoken'
 import jwksClient from 'jwks-rsa'
+import { uuid } from 'uuidv4'
 
 import { Inject, Injectable, UnauthorizedException } from '@nestjs/common'
 import { ConfigType } from '@nestjs/config'
 
-import type { User } from '@island.is/judicial-system/types'
+import { type User, UserRole } from '@island.is/judicial-system/types'
```
gudjong, line 9: Interesting, didn't know this was possible.

apps/judicial-system/api/src/app/modules/auth/auth.service.ts
```diff
@@ -32,18 +35,50 @@ export class AuthService {
...
+        `${this.config.backendUrl}/api/cases/limitedAccess/defender?nationalId=${nationalId}`,
+        {
+          headers: { authorization: `Bearer ${this.config.secretToken}` },
+        },
+      )
+      if (res.ok) {
+        return await res.json()
+      }
+    } catch (error) {
+      console.warn('Defender not found', error)
```
gudjong, line 49: Please check specifically for a NotFoundException and only log and pass through for that exceptions. Also, consider logging as info. An error is logged futher down the call path "Blocking login attempt from an unauthorized user". That message should probably be logged as a warning instead.

If the exception is anything else it should be rethrown as it indicates an error on the backend or a problem with the connection to the backend.

apps/judicial-system/api/src/app/modules/auth/auth.service.ts
```diff
@@ -32,18 +35,50 @@ export class AuthService {
...
+          mobileNumber: lawyerRegistryInfo.phoneNr,
+          active: true,
+          title: 'verjandi',
+          id: uuid(),
+          created: new Date().toString(),
+          modified: new Date().toString(),
+        } as User
+      }
+    } catch (error) {
+      console.warn('Defender not found in lawyer registry', error)
```
gudjong, line 78: Consider logging as info for the same reason as above. Also, consider a more accurate message as the reason for the failure could be anything.

apps/judicial-system/api/src/app/modules/defender/defender.service.ts
```diff
@@ -0,0 +1,59 @@
...
+    })
+
+    if (response.ok) {
+      const lawyers = await response.json()
+      const lawyersMapped = (lawyers || []).map(mapToLawyer)
+      return lawyersMapped
+    }
+
+    const reason = await response.text()
+    console.error('Failed to get lawyers:', reason)
```
gudjong, line 32: Please remove console.log. If you still want to log, consider injecting a logger and use it instead. Also, please do not log as error as that will trigger alerts in operation.

apps/judicial-system/api/src/app/modules/auth/auth.service.ts
gudjong: Please do not use console logging. Instead, inject a logger and use it instead.

apps/judicial-system/api/src/app/modules/defender/defender.service.ts
```diff
@@ -0,0 +1,59 @@
...
+    if (response.ok) {
+      const lawyer = await response.json()
+      const lawyerMapped = {
+        ...mapToLawyer(lawyer),
+      }
+      return lawyerMapped
+    }
+
+    const reason = await response.text()
+    console.error('Failed to get lawyer:', reason)
```
gudjong, line 56: Please do not use console logging and please do not log as error if you still want to log.

gudjong (changes requested):

brynjarorng (approved):
LGTM

gudjong (approved):
Looks great

PR #12303, opened 2023-10-04
fix(user-profile): Allow 204 response in userprofile service

thordurhhh:
## What

Update `userProfileControllerFindOneByNationalId` to allow for 204.
Update userprofile service to use raw response to allow for `200` and `204` responses at the same time.

## Why

The direct response will not allow two successful responses at the same time so we need to go for the raw response and handle that specifically.

## Checklist:


jonbjarnio (approved):
LGTM! 🤝

libs/api/domains/user-profile/src/lib/userProfile.service.ts
```diff
@@ -49,6 +52,25 @@ export class UserProfileService {
...
+        return response.value()
+      },
+      (error) => {
+        if (error instanceof FetchError && error.status === IGNORED_STATUS) {
+          return null
+        }
+        throw error
+      },
+    )
+  }
```
saevarma, line 72: This is a nice utility wrapper to handle our 204 responses. Can you move it to the `@island.is/clients/middlwares` lib as the `handle404` handler.

saevarma (approved):
Nice! 🙌

PR #12344, opened 2023-10-10
feat(service-portal): optimize user info queries

thorkellmani:
## What
* Use lazy queries for user information fetches
* Filter children information by the child's national id

## Why
The lazy queries enables us to abstain from fetching until we've received the feature flag data. The child filtering reduces the data being transferred.

## Checklist:


thordurhhh (approved):
🙌

PR #12371, opened 2023-10-11
fix(sessions): use jobs for geoip

busla:
# Sessions

Run geoip as a job instead of cronjob for more timing control.

The job will be run before starting other services in the namespace by using `pre-upgrade,pre-install` annotations.

```yaml
    'helm.sh/hook': 'pre-install,pre-upgrade',
    'helm.sh/hook-delete-policy': 'before-hook-creation,hook-succeeded'
```

We could also consider replacing `initContainers` that run migrations to use `Jobs` with the above annotation to skip running migrations every time the services autoscale.

### DIFF
--- a/apps/services/sessions/src/deleteme.ts
+++ b/apps/services/sessions/src/deleteme.ts
@@ -1,0 +1,1 @@
+console.log('deleteme')
--- a/charts/islandis/values.dev.yaml
+++ b/charts/islandis/values.dev.yaml
@@ -2306,6 +2306,7 @@ services-sessions-geoip-job:
     GEODATADIR: '/geoip-lite/data'
     GEOTMPDIR: '/geoip-lite/data/tmp'
     NODE_OPTIONS: '--max-old-space-size=921'
+    REDIS_URL_NODE_01: '["clustercfg.general-redis-cluster-group.5fzau3.euw1.cache.amazonaws.com:6379"]'
     SERVERSIDE_FEATURES_ON: ''
   grantNamespaces:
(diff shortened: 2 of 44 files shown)


apps/services/sessions/src/deleteme.ts
saevarma, line 1: Delete me?

charts/islandis/values.dev.yaml
```diff
@@ -2302,8 +2302,7 @@ services-sessions-geoip-worker:
   env:
     GEODATADIR: '/geoip-lite/data'
     GEOTMPDIR: '/geoip-lite/data/tmp'
-    NODE_OPTIONS: '--max-old-space-size=1'
-    REDIS_URL_NODE_01: '["clustercfg.general-redis-cluster-group.5fzau3.euw1.cache.amazonaws.com:6379"]'
+    NODE_OPTIONS: '--max-old-space-size=2'
```
saevarma, line 2305: I'm a bit confused, max 2mb? and how was the previous 1mb enough.
I also don't see the infra changed that would cause a change on this chart value?

busla: well spotted :pray:

busla: fixed

busla:
> Delete me? 

we wont get a feature-deployment for a services unless there are changes in source code. We have a ticket for adding `apps/**/infra` to the NX affected computation.

Will remove before setting automerge label

lodmfjord (approved):
LGTM

PR #12379, opened 2023-10-12
chore(core): Refactor schema generation

eirikurn:
This PR improves our schema generation in the following ways:

- Rename "schema*" scripts to "codegen*" which generates schemas, types and client code.
- Refactor codegen related scripts into "codegen/frontend-client", "codegen/backend-schema" and "codegen/backend-client".
- Configure NX [inputs](https://nx.dev/reference/project-configuration#inputs-namedinputs), [outputs](https://nx.dev/reference/project-configuration#outputs) and [dependsOn](https://nx.dev/reference/project-configuration#dependson) to get optimised caching and dependency resolution.

The biggest benefit are in dependency resolution and caching.

### Caching

Each codegen script is now cached based on inputs and outputs. This should speed up warm codegen by quite a bit. 

Also, when we enable distributed caching with NX, we can use it to replace the naive "generated-files" caching in our current CI.

### Dependency resolution

Before it was quite challenging to codegen for services which depend on other services which depend on yet other services (internal or external). Our workarounds broke caching, ran some scripts multiple times, were difficult to understand and sometimes resulted in race condition CI failures.

Now NX automatically figures out the correct order to generate schemas, as deep as needed. For example the following chain just works automatically:

```
GraphQL API -> Application System Client -> Application System API -> User Profile Client -> User Profile API -> Islykill Client
```

## Notes

- There is still a "dependency" wall between frontend apps and backend apps. Backend changes will generally "affect" all backend apps which depend on it, but will not "affect" any frontend apps. This was an attempt to keep feature deployments small by default. This means we still have a `scripts/codegen.ts` script which first runs "codegen/backend-schema" on all projects before running "codegen/frontend-client" on all projects. There might be benefits in removing this dependency wall, but we leave that for the future.
- There are imperfections in the input configuration which may cause false cache hits. This is due to how NX by default [ignores "gitignored" files in inputs/cache keys](https://github.com/nrwl/nx/issues/6821), and all of our codegen is git ignored. There are usually enough code changes to invalidate these caches, but if this becomes a problem we can either try using a [`calculateHash` runtime input](https://github.com/nrwl/nx/issues/6821#issuecomment-1577457968) or [dependent task outputs](https://nx.dev/reference/project-configuration#dependent-tasks-output).


busla:
Regarding the `.gitignore` issue, does the strategy in [this comment](https://github.com/nrwl/nx/issues/6821#issuecomment-1527053077) help?

saevarma (approved):
💯🚀

thorkellmani (approved):
LGTM

eirikurn:
> Regarding the `.gitignore` issue, does the strategy in [this comment](https://github.com/nrwl/nx/issues/6821#issuecomment-1527053077) help?

For some reason it did not work in my testing, and I'm pretty sure I wasn't triggering the folder limitation. But if someone can get it to work then that would be ideal.

PR #12385, opened 2023-10-12
fix(j-s): Remove .pdf ending from indictment case files

unakb:
[Ekki skeyta við tegund skjals í skjalaskrá](https://app.asana.com/0/1199153462262248/1205641161073618/f)

## What

Remove .pdf from indictment case files display names

## Why

Set the user generated filename automatically when a file is created. In the future we might want to rename the field. The reason for setting it is so it can be used to display the name of the file without .pdf at the end, making it easier for prosecutors to arrange indictment case files, as they don't need to manually remove the .pdf ending of each one.

## Screenshots / Gifs

## Checklist:


apps/judicial-system/backend/src/app/modules/file/file.service.ts
```diff
@@ -207,15 +207,18 @@ export class FileService {
...
       ...createFile,
       state: CaseFileState.STORED_IN_RVG,
       caseId: theCase.id,
-      name: createFile.key.slice(
-        isIndictmentCase(theCase.type)
-          ? INDICTMENT_NAME_BEGINS_INDEX
-          : NAME_BEGINS_INDEX,
-      ),
+      name: fileName,
+      userGeneratedFilename: fileName.replace(/\.pdf$/, ''),
```
unakb, line 221: Not an ideal solution, but I feel like it's the cleanest at this time and solves the issue at hand. In the future maybe we could rename the field since this isn't technically user generated when created. Just rename it display name or something like that.

apps/judicial-system/backend/src/app/modules/file/dto/createFile.dto.ts
```diff
@@ -46,4 +46,9 @@ export class CreateFileDto {
   @IsString()
   @ApiPropertyOptional()
   readonly policeFileId?: string
+
+  @IsOptional()
+  @IsString()
+  @ApiPropertyOptional()
+  readonly userGeneratedFilename?: string
```
gudjong, line 53: Please consider removing this field from the dto as it is never used.

gudjong (approved):

PR #12393, opened 2023-10-12
feat(web): applications overview

splicebison:
# ...

## What

## Why

## Screenshots / Gifs

## Checklist:

### DIFF
--- a/apps/web/hooks/useLinkResolver/useLinkResolver.ts
+++ b/apps/web/hooks/useLinkResolver/useLinkResolver.ts
@@ -37,6 +37,10 @@ export const routesTemplate = {
     is: '/s/stafraent-island/[slug]',
     en: '',
   },
+  applications: {
+    is: '/umsoknir',
+    en: '/en/applications',
+  },
   page: {
     is: '/stafraent-island',
     en: '',
--- a/apps/web/screens/Applications/Applications.tsx
+++ b/apps/web/screens/Applications/Applications.tsx
@@ -26,8 +26,10 @@ import {
   GridRow,
   GridColumn,
   Button,
+  Table as T,
 } from '@island.is/island-ui/core'
-import { SearchInput, Card, CardTagsProps } from '@island.is/web/components'
+import { ProcessEntryLinkButton } from '@island.is/island-ui/contentful'
+import { SearchInput, CardTagsProps } from '@island.is/web/components'
 import { useI18n } from '@island.is/web/i18n'
 import { useNamespace } from '@island.is/web/hooks'
 import { CustomNextError } from '@island.is/web/units/errors'
@@ -261,7 +263,9 @@ const Applications: Screen<CategoryProps> = ({
       link: getItemLink(item),
       categorySlug: item.category?.slug ?? item.parent?.category?.slug,
       category: item.category ?? item.parent?.category,
+      organizationTitle: item.organization?.length && item.organization[0].title,
       hasProcessEntry: checkForProcessEntries(item),
+      processEntry: item.processEntry,
       group: item.group,
       ...getItemImages(item),
       labels: getLabels(item),
@@ -476,16 +480,23 @@ const Applications: Screen<CategoryProps> = ({
               )}
             </Stack>
             <ColorSchemeContext.Provider value={{ colorScheme: 'blue' }}>
-              <Stack space={2}>
+              <T.Table>
+                <T.Head>
+                  <T.Row>
+                    <T.HeadData>Name</T.HeadData>
+                    <T.HeadData>Government Agency</T.HeadData>
+                    <T.HeadData></T.HeadData>
+                  </T.Row>
+                </T.Head>
+                <T.Body>
                 {filteredItems.map(
                   (
                     {
-                      typename,
-                      image,
-                      thumbnail,
                       labels,
-                      parentTitle,
-                      ...rest
+                      link,
+                      title,
+                      organizationTitle,
+                      processEntry,
                     },
                     index,
                   ) => {
@@ -500,19 +511,37 @@ const Applications: Screen<CategoryProps> = ({
                       })
                     })
 
+                    console.log(processEntry);
+
                     return (
-                      <Card
-                        key={index}
-                        tags={tags}
-                        dataTestId="search-result"
-                        image={thumbnail ? thumbnail : image}
-                        subTitle={parentTitle}
-                        highlightedResults={true}
-                        {...rest}
-                      />
+                      <T.Row key={index}>
+                        <T.Data>
+                          <NextLink
+                            {...link}
+                            passHref
+                            legacyBehavior
+                          >
+                            {title}
+                          </NextLink>
+                        </T.Data>
+                        <T.Data>{organizationTitle}</T.Data>
+                        <T.Data>
+                          {processEntry?.processLink && (
+                            <ProcessEntryLinkButton
+                              processTitle={processEntry.processTitle ?? title}
+                              processLink={processEntry.processLink}
+                              buttonText="Apply"
+                              size="small"
+                            />
+                          )}
+                        </T.Data>
+                      </T.Row>
                     )
                   },
-                )}{' '}
+                )}
+                </T.Body>
+              </T.Table>
+              <Stack space={2}>
                 {totalSearchResults > 0 && (
                   <Box paddingTop={6}>
                     <Pagination
--- a/apps/web/screens/Applications/Applications.tsx
+++ b/apps/web/screens/Applications/Applications.tsx
@@ -56,6 +56,7 @@ import {
   OrganizationPage,
   Link as LinkItem,
   ProjectPage,
+  SortField,
 } from '@island.is/web/graphql/schema'
 import { AnchorPageType } from '@island.is/web/utils/anchorPage'
 import { ActionType, reducer, initialState } from '../Search/Search.state'
@@ -80,6 +81,7 @@ type SearchQueryFilters = {
 interface CategoryProps {
   q: string
   page: number
+  sort: string
   searchResults: GetSearchResultsDetailedQuery['searchResults']
   countResults: GetSearchCountTagsQuery['searchResults']
   namespace: GetNamespaceQuery['getNamespace']
@@ -101,6 +103,7 @@ const stringToArray = (value: string | string[]) =>
 const Applications: Screen<CategoryProps> = ({
   q,
   page,
+  sort,
   searchResults,
   countResults,
   namespace,
@@ -496,6 +499,7 @@ const Applications: Screen<CategoryProps> = ({
                       title,
                       organizationTitle,
                       processEntry,
+                      parentTitle,
                     },
                     index,
                   ) => {
@@ -577,6 +581,7 @@ Applications.getProps = async ({ apolloClient, locale, query }) => {
   const queryString = single(query.q) || '*'
   const q = queryString !== '*' ? queryString : '';
   const page = Number(single(query.page)) || 1
+  const sort = single(query.sort) === 'title' ? SortField.Title : SortField.Popular
   const category = query.category ?? ''
   const types = stringToArray('webArticle') as SearchableContentTypes[];
   const organization = query.organization ?? ''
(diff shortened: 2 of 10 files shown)


saevarma (approved):
Core files LGTM

apps/web/components/Organization/Slice/FeaturedArticles/FeaturedArticlesSlice.tsx
```diff
@@ -63,30 +65,30 @@ export const FeaturedArticlesSlice: React.FC<
...
+                <FocusableBox
+                  key={article.slug}
+                  borderRadius="large"
+                  href={url.href}
+                >
+                  <TopicCard
+                    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+                    // @ts-ignore make web strict
+                    tag={
+                      hasProcessEntries(article as Article) &&
```
RunarVestmann, line 80: It seems that this change is impacting articles that have a stepper on them

Dev 
image

This deployed change
image

This field in the CMS sets the text and what it should be
image

Is there a way we can do this and still keep the old functionality present?

apps/web/hooks/useLinkResolver/useLinkResolver.ts
```diff
@@ -39,6 +39,10 @@ export const routesTemplate = {
     is: '/s/stafraent-island/[slug]',
     en: '',
   },
+  applications: {
+    is: '/umsokn',
+    en: '/en/applications',
```
kksteini, line 44: Just as a curiousity.
The icelandic is in singular but the english in plural.

Would it look something like this?
```typescript
is: '/umsoknir/',
en: '/en/applications
```

RunarVestmann (changes requested):
I've only looked at two files and I've already got a comment. I'll just leave it here and look at the rest later if that's ok for now :)

apps/web/hooks/useLinkResolver/useLinkResolver.ts
kksteini: Ah but of course, /umsoknir/ are the applications themselves.
I see.

apps/web/components/Organization/Slice/FeaturedArticles/FeaturedArticlesSlice.tsx
splicebison: @RunarVestmann Moving forward, all Process Entries will only be used for actual applications. Previously, the old code was only checking for the main Process Entry field on an Article however the search (correctly) looks for both the field and additionally the embeded content. This now checks everywhere the same as search so that the approach is the same throughout the site.

There's been a large internal discussion on how best to approach this topic considering Applications do not technically exist within the content system. Process Entry was meant to represent this however has grown in a way that it's currently being used for things that are not actual applications. This content strategy will change moving forward.

With that in mind, I am not sure we should be allowing a manual override of a tag in this case. An article either is or is not an application and that is based solely on the presence of Process Entry (field or embeded in rich text). This will ensure that the UI treatment is always the same.

Is the problem only a matter of labeling? If so, I can simply check if a: process entry exists and, if so, display either the dropdown value or the default global value for the tag. I would like to avoid on relying on the dropdown field being populated or not to dictate whether the tag should show. Feel free to ping me. This topic is some what complicated given the data structure.

kksteini (approved):
CODEOWNER files look good to me.

apps/web/hooks/useLinkResolver/useLinkResolver.ts
splicebison: Unfortunately `/umsoknir` is used by the application system (it's a redirect in production) so we can not use it. We asked to do so already. The product owner decided on the singular version.

apps/web/components/Organization/Slice/FeaturedArticles/FeaturedArticlesSlice.tsx
splicebison: I've reviewed this. These articles do indeed Process Entry so this is not the desired outcome as the tag should show based on the util. I'll investigate why it's not detecting the  process entry.

apps/web/hooks/useLinkResolver/useLinkResolver.ts
splicebison: I agree it reads strangely. I can open the topic again with the product owner but we proposed a few options.

apps/web/components/Organization/Slice/FeaturedArticles/FeaturedArticlesSlice.tsx
splicebison: @RunarVestmann I found the issue. It's because these articles do not contain process entries at all. They have a Stepper content type embeded in the rich text. This unfortunately adds yet another complication to the mental map of what constitutes an "application".

In the meantime, I have changed the code to conditionally render based on the button text just for this section but this should get properly addressed once we implement the new CTA Link content type as logging in to see if you need a visa (step by step) is not an actual application. The UI should also be different to reflect this. I'll maintain a separate PR later for this new content type that should include this change as well. See latest commit.

apps/web/hooks/useLinkResolver/useLinkResolver.ts
albinagu: agree also 👍

apps/web/screens/Applications/Applications.tsx
```diff
@@ -0,0 +1,331 @@
...
+            )}
+          </GridColumn>
+        </GridRow>
+      </GridContainer>
+    </>
+  )
+}
+
+Applications.getProps = async ({ apolloClient, locale, query }) => {
+  // TODO - file bugreport with next-usequerystate regarding missing default on server
```
RunarVestmann, line 230: Has this TODO been resolved?

apps/web/screens/Applications/Applications.tsx
```diff
@@ -0,0 +1,331 @@
...
+    {
+      data: { searchResults },
+    },
+    {
+      data: { searchResults: countResults },
+    },
+    namespace,
+  ] = await Promise.all([
+    apolloClient.query<GetApplicationsQuery, QuerySearchResultsArgs>({
+      fetchPolicy: 'no-cache', // overriding because at least local caching is broken
```
RunarVestmann, line 276: Did you try adding an environment variable to your .env.secret file which contains the bypass cache key, after I did that I've noticed that appending ?bypass-cache=<bypass-key> in the url locally does indeed refresh the cache

apps/web/screens/Applications/Applications.tsx
```diff
@@ -0,0 +1,331 @@
...
+        query: GET_NAMESPACE_QUERY,
+        variables: {
+          input: {
+            namespace: 'Applications',
+            lang: locale,
+          },
+        },
+      })
+      .then((variables) =>
+        JSON.parse(variables?.data?.getNamespace?.fields || '[]'),
```
RunarVestmann, line 315: Why not default to a '{}' instead?

apps/web/screens/Category/Category/Category.tsx
```diff
@@ -352,39 +354,31 @@ const Category: Screen<CategoryProps> = ({
...
+                            href={
+                              linkResolver(
+                                article.__typename?.toLowerCase() as LinkType,
+                                [article.slug],
+                              ).href
+                            }
+                            // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+                            // @ts-ignore make web strict
+                            tag={
+                              hasProcessEntries(article as Article) &&
```
RunarVestmann, line 370: This page also doesn't display the "Skref fyrir skref" tag anymore

Ísland.is
image

This PR
image

apps/web/screens/Organization/Services.tsx
```diff
@@ -267,8 +269,7 @@ const ServicesPage: Screen<ServicesPageProps> = ({
                   // eslint-disable-next-line @typescript-eslint/ban-ts-comment
                   // @ts-ignore make web strict
                   tag={
-                    (!!article.processEntry ||
-                      article.processEntryButtonText) &&
+                    hasProcessEntries(article) &&
```
RunarVestmann, line 272: Same here, "Skref fyrir skref" isn't displayed

libs/cms/src/lib/models/articleReference.ts
```diff
@@ -31,6 +32,9 @@ export class ArticleReference {
   @Field()
   intro?: string
 
+  @CacheField(() => [SliceUnion], { nullable: true })
+  body?: Array<typeof SliceUnion> = []
```
RunarVestmann, line 36: Why are we adding the body field to the article reference model? I thought the article reference was only there to prevent potential circularity between subArticle <-> article.

apps/web/screens/Applications/Applications.tsx
splicebison: This was remnants from copying over the Search screen as the template for this page. I think it can be safely removed. I'll run some local tests.

apps/web/screens/Applications/Applications.tsx
```diff
@@ -0,0 +1,331 @@
...
+          order,
+          types,
+          ...(tags.length && { tags }),
+          size: PERPAGE,
+          page,
+        },
+      },
+    }),
+    apolloClient.query<GetSearchCountTagsQuery, QuerySearchResultsArgs>({
+      fetchPolicy: 'no-cache', // overriding because at least local caching is broken
```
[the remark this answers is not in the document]
splicebison: same as above

apps/web/screens/Applications/Applications.tsx
splicebison: Same as above. Copy-pasta issue. Will change. I want to also add type safety throughout the file where it's missing. Linter should have picked this up since it's defaulting to `any`.

libs/cms/src/lib/models/articleReference.ts
splicebison: The Featured Article model contains a collection of Article References. This was meant to be able to check if these articles contain a Process Entry type embeded in the rich text in addition to the main Process Entry filed to determine if they are Applications or not. I am unaware of what the original intent was but this slice's articles are currently populated with these references.

apps/web/screens/Category/Category/Category.tsx
splicebison: In the meantime, I am checking additionally for the buttonText presence. I do think this should be address in a later PR in a more meaningful way

libs/cms/src/lib/models/articleReference.ts
RunarVestmann: I'm wondering another approach might be better, like if all we are using the body field for is simply to check if there is a process entry connected to this article, perhaps adding a boolean field instead that indicates if there is a process entry or not might be better?

splicebison: Agreed. In fact, I think this would be best in general for all Articles, not just here as I do not really like the current way the code checks for the presence of process entries. I'll remove this and address this in a separate PR as it doesn't really affect the current over view.

libs/island-ui/core/src/lib/Filter/FilterInput/FilterInput.tsx
```diff
@@ -1,27 +1,31 @@
...
   value: string
+  button?: Omit<InputButton, 'name' | 'type'>
   onChange: (value: string) => void
   onKeyDown?: (
     event: React.KeyboardEvent<HTMLInputElement | HTMLTextAreaElement>,
   ) => void
   backgroundColor?: ResponsiveProp<InputBackgroundColor>
 }
 
+const icon: InputIcon = { name: 'search', type: 'outline' }
```
RunarVestmann, line 19: I'm just wondering if this is in Figma, nice job adding a story to storybook for this 👍

RunarVestmann (approved):

libs/island-ui/core/src/lib/Filter/FilterInput/FilterInput.tsx
splicebison: Good question. I can talk with design and make sure it's added.

PR #12400, opened 2023-10-13
chore: util script for changelogs

busla:
# Changelog script

Helper for generating changelogs instead of manually checking for two most recent release branches


brynjarorng (approved):
LGTM (Geggjað)

PR #12436, opened 2023-10-17
fix(island-ui): Hide PDF pagination when not needed

thordurhhh:
## What

* Hide pagination when there is only one PDF page.
* Add option for custom error component to PDF viewer. The error will show default error string if component is not provided.

## Why

* No need to show pagination for only one page, most of the PDF display is for single pages and it looks kinda strange to have the pagination in view all the time.
* Custom errors needed for Document Inbox.

### Pagination

Current:
Screenshot 2023-10-16 at 10 53 13

Becomes:
Screenshot 2023-10-16 at 10 53 56

### Error component

Error component custom:
Screenshot 2023-10-26 at 00 01 18

Error component default:
Screenshot 2023-10-26 at 00 01 25

## Checklist:


osk (approved):
Nice!

Should we have the default error be in Icelandic?

Tryggvig (approved):

PR #12437, opened 2023-10-17
fix(j-s): Failing Playwright tests

unakb:
# Fix failing e2e regression tests

[Playwright test failures í pípunni ](https://app.asana.com/0/1199153462262248/1205581325882057)

## What
Added a timeout delay before keyboard typing for certain textboxes

## Why
The tests are failing more often than not, firstly due to playwright typing faster than the page takes to send prefill updates for other fields, so required fields get overwritten and sometimes end up empty. This can be fixed by adding a short timeout before the 2 problematic input boxes.

## Checklist:


apps/system-e2e/src/support/session.ts
```diff
@@ -157,11 +157,19 @@ export async function judicialSystemSession({ browser }: { browser: Browser }) {
   const context = await browser.newContext()
   const page = await context.newPage()
   const authUrlPrefix = urls.authUrl
-  await ensureCognitoSessionIfNeeded(
-    page,
-    JUDICIAL_SYSTEM_HOME_URL,
-    authUrlPrefix,
-  )
-  await page.close()
+  // Retry a few times in case login fails due to deployment not being stable
```
AndesKrrrrrrrrrrr, line 160: If you really need retrying, why not move it into the `ensureCognitoSessionIfNeeded` to benefit all unstable tests?

apps/system-e2e/src/support/session.ts
```diff
@@ -157,11 +157,19 @@ export async function judicialSystemSession({ browser }: { browser: Browser }) {
...
+    await ensureCognitoSessionIfNeeded(
+      page,
+      JUDICIAL_SYSTEM_HOME_URL,
+      authUrlPrefix,
+    )
+    const cookies = await context.cookies()
+    if (cookies.length > 0) {
+      break
+    }
+    await new Promise((resolve) => setTimeout(resolve, 5000))
```
AndesKrrrrrrrrrrr, line 171:
```suggestion
    await sleep(5000)  // from ../utils.ts
```

apps/system-e2e/src/tests/judicial-system/smoke/custody-prosecutor.spec.ts
```diff
@@ -64,13 +64,15 @@ test.describe('Custody Prosecutor', () => {
     await page.locator('input[id=reqValidToDate]').fill(today)
     await page.keyboard.press('Escape')
     await page.locator('input[id=reqValidToDate-time]').fill('16:00')
+    await page.waitForTimeout(1000)
```
AndesKrrrrrrrrrrr, line 67: Instead of waiting for some arbitrary time, why not wait for the thing you're waiting for? Are you waiting for a HTTP response? A UI element update?

apps/system-e2e/src/tests/judicial-system/smoke/custody-prosecutor.spec.ts
```diff
@@ -64,13 +64,15 @@ test.describe('Custody Prosecutor', () => {
...
     await page.locator('textarea[name=lawsBroken]').click()
-    await page.keyboard.type('Einhver lög voru bortin')
+    await page.keyboard.type('Einhver lög voru brotin')
```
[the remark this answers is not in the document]
unakb: I was under the impression this just makes the typing of each letter slower, I want to wait a little before starting the typing at all. Is that a misunderstanding on my part?

apps/system-e2e/src/support/session.ts
unakb: I mostly just wanted to see if this would fix the issue before introducing it to every test that's run in other projects :)

apps/system-e2e/src/support/session.ts
unakb: TIL, thanks!

apps/system-e2e/src/tests/judicial-system/smoke/custody-prosecutor.spec.ts
unakb: Honestly a lack of familiarity with Playwright on my part coupled with impatience to test out our hypothesis 😊 I'll have a look in to seeing if I can implement this in a different way.

apps/system-e2e/src/tests/judicial-system/smoke/custody-prosecutor.spec.ts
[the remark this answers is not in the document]
AndesKrrrrrrrrrrr: You are correct. I misunderstood the problem. I thought the fast typing was the issue. Maybe do both?

PR #12449, opened 2023-10-19
fix: set default service upstream to true for fully zero downtime deployments

brynjarorng:
# Upstream timeouts from the ingress controller

## What

We have been seeing timeouts for scale down events and when we roll over to a new release due to nginx sending requests to pods that have already been set to a terminating state.

## Why

This can cause the user to hang for a while and then see a 504 timeout from nginx with no obvious error message


peturgq:
Excellent, 
Since this is a fundamental change in how our ingress controller routes requests, I suggest we let this simmer on dev for all the services for a while before going on prod.

peturgq (approved):

PR #12450, opened 2023-10-19
fix(service-portal): Adding pagination to my vehicles

Tryggvig:
# ...

https://www.notion.so/B-ta-vi-pagination-yfirlit-kut-kja-6a1935186eef4699bdace43d78a7bffb?pvs=4

## What

Adding pagination to my vehicles

## Why

People with more than 50 vehicles could not see all vehicles.

## Checklist:

### DIFF
--- a/libs/api/domains/vehicles/src/lib/api-domains-vehicles.service.ts
+++ b/libs/api/domains/vehicles/src/lib/api-domains-vehicles.service.ts
@@ -7,14 +7,15 @@ import {
   BasicVehicleInformationGetRequest,
   PdfApi,
   VehicleSearchDto,
-  PersidnoLookupDto,
   PublicVehicleSearchApi,
+  PersidnoLookupResultDto,
 } from '@island.is/clients/vehicles'
 import { VehiclesDetail } from '../models/getVehicleDetail.model'
 import { ApolloError } from 'apollo-server-express'
 import { AuthMiddleware } from '@island.is/auth-nest-tools'
 import type { Auth, User } from '@island.is/auth-nest-tools'
 import { basicVehicleInformationMapper } from '../utils/basicVehicleInformationMapper'
+import { handle404 } from '@island.is/clients/middlewares'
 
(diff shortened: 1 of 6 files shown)


libs/service-portal/assets/src/screens/Overview/Overview.tsx
```diff
@@ -234,24 +218,27 @@ const VehiclesOverview = () => {
...
-                  placeholder={formatMessage(m.searchPlaceholder)}
-                />
-              </Box>
-            </GridColumn>
-          </GridRow>
-        )}
+        {!loading &&
+          !error &&
+          vehicles.length > 4 &&
+          false && ( //will be visited again on next version of service.
```
thorkellmani, line 224: u promise?

libs/api/domains/vehicles/src/lib/api-domains-vehicles.service.ts
```diff
@@ -81,8 +82,8 @@ export class VehiclesService {
       const { data } = res
       if (!data) return null
       return data[0]
-    } catch (e) {
-      return this.handle4xx(e, 'Failed to get vehicle search')
+    } catch (error) {
+      return this.handleError(error, 'Failed to get vehicle search')
```
thorkellmani, line 86: HandleError contains a logger.error function, but I don't think this error message, or the error messages below, are adding any useful logging messages that'll assist in debugging that the client fetch error log won't already have provided. I would either like a more descriptive error message for the logs, or just skipping the logging if not possible.

libs/service-portal/assets/src/screens/Overview/Overview.tsx
fostbraedur: promise

Tryggvig: promise

thorkellmani (approved):
LGTM 🦭

PR #12456, opened 2023-10-19
test: added hmr-benchmark script

busla:
# hmr-benchmark script

Run duration benchmark on `nx serve` targets to measure changes in configuration before deciding on different bundlers or configuration changes.

Example:

```bash
yarn hmr-benchmark --app services-sessions --message "Some NX log output that determines that the app has reloaded"
```


lodmfjord (approved):
lgtm

saevarma (approved):
LGTM

PR #12461, opened 2023-10-19
fix(passport): Fixing legal residence check

joningi98:
## What

Currently, our passport application system restricts access to users without legal residence, preventing them from applying for passports on behalf of their children. 

## Checklist:

---
  - https://app.asana.com/0/0/1205738665206036


kksteini (approved):
LGTM

libs/application/template-api-modules/src/lib/modules/shared/api/national-registry/national-registry.service.ts
```diff
@@ -31,11 +31,34 @@ export class NationalRegistryService extends BaseTemplateApiService {
     auth,
     params,
   }: TemplateApiModuleActionProps<NationalRegistryParameters>): Promise<NationalRegistryIndividual | null> {
```
obmagnusson, line 33: This functionality complicates this function somewhat. So I would suggest breaking this up like so. 
``` 
const result = await this.getIndividual(auth.nationalId);

if (!result) {
        throw new TemplateApiError(
            {
                title: coreErrorMessages.nationalIdNotFoundInNationalRegistryTitle,
                summary: coreErrorMessages.nationalIdNotFoundInNationalRegistrySummary,
            },
            400
        );
    }
    
    this.validateIndividual(result, params);
    
    if (params?.includeChildrenInValidation) {
        const children = await this.nationalRegistryApi.getCustodyChildren(auth);
        await this.validateChildren(children, params);
    }

    return result;

``` 
Where validateChildren loops through each child uses the validateIndividual and  then a separate function for each validation.

libs/application/template-api-modules/src/lib/modules/shared/api/national-registry/national-registry.service.ts
```diff
@@ -31,11 +31,34 @@ export class NationalRegistryService extends BaseTemplateApiService {
...
+    const individuals: Array<NationalRegistryIndividual | null> = []
     const result = await this.getIndividual(auth.nationalId)
+    individuals.push(result)
+
+    if (params?.allowDomicilePassOnChild) {
```
obmagnusson, line 38: Are we not checking for more than just domicile?  `includeChildrenInValidation`  as suggestion ?

obmagnusson (changes requested):
Sry for the late reply. Couple of comments.

libs/application/template-api-modules/src/lib/modules/shared/api/national-registry/national-registry.service.ts
joningi98: I've broken up the function, makes the validation part easier to read

libs/application/template-api-modules/src/lib/modules/shared/api/national-registry/national-registry.service.ts
```diff
@@ -31,39 +30,100 @@ export class NationalRegistryService extends BaseTemplateApiService {
...
+    return individual
+  }
+
+  private validateIndividual(
+    individual: NationalRegistryIndividual,
+    isChild: boolean,
+    params?: NationalRegistryParameters,
+  ) {
+    if (params?.legalDomicileIceland && !params?.allowPassOnChild) {
+      this.validateDomicleInIceland(individual, params.allowPassOnChild)
```
kksteini, line 65: Won't `params.allowPassOnChild` always be `false` or `undefined` here?
Is that expected?

libs/application/template-api-modules/src/lib/modules/shared/api/national-registry/national-registry.service.ts
```diff
@@ -31,39 +30,100 @@ export class NationalRegistryService extends BaseTemplateApiService {
...
+
+    if (params?.ageToValidate && !isChild) {
+      this.validateAge(params, individual)
+    }
+  }
+
+  private async validateChildren(params: NationalRegistryParameters, childrenId: string[]) {
+    for (const id of childrenId) {
+      const individual = await this.getIndividual(id)
+      console.log('validateChildren', individual)
```
kksteini, line 83: Shuold this console log be here?

libs/application/template-api-modules/src/lib/modules/shared/api/national-registry/national-registry.service.ts
joningi98: your right!

libs/application/template-api-modules/src/lib/modules/shared/api/national-registry/national-registry.service.ts
```diff
@@ -31,39 +30,91 @@ export class NationalRegistryService extends BaseTemplateApiService {
...
+        {
+          title: coreErrorMessages.nationalRegistryCitizenshipNotIcelandic,
+          summary: coreErrorMessages.nationalRegistryCitizenshipNotIcelandic,
+        },
+        400,
+      )
+    }
+  }
+
+  private validateAge(params: any, individual: NationalRegistryIndividual) {
```
obmagnusson, line 116: params: NationalRegistryParameters

libs/application/template-api-modules/src/lib/modules/shared/api/national-registry/national-registry.service.ts
```diff
@@ -31,39 +30,91 @@ export class NationalRegistryService extends BaseTemplateApiService {
...
+  ) {
+    for (const id of childrenId) {
+      const individual = await this.getIndividual(id)
+      if (individual) {
+        this.validateIndividual(individual, true, params)
       }
     }
+  }
+
+  private validateDomicleInIceland(individual: NationalRegistryIndividual) {
```
obmagnusson, line 89: Typo : validateDomicleInIceland -> validateDomicileInIceland

obmagnusson (changes requested):
Nice ! Minor comments. Has this been tested on other applications?

obmagnusson (approved):

PR #12463, opened 2023-10-20
feat(service-portal): Use publishdate ahead of doc date for inbox item

thordurhhh:
### Minar sidur inbox documents
Use publish date if available ahead of the document date. The document date will be used as a fallback if the publish date is not present.

## Why

The document date can be a few days old once the document is published so it can confuse the user when a new document can be a few days old. This will ensure that the date on the document is the same date as the day it will be shown to the user.

## Checklist:


thorkellmani (approved):
LGTM

PR #12473, opened 2023-10-23
chore: move system test utils to new library

lodmfjord:
# Create new library for system tests

## What

Foundation for moving system tests to their own project folders.

## Why

Currently @island-is/qa is the code owner of all system tests, which is not ideal.


AndesKrrrrrrrrrrr (approved):

libs/testing-system/src/index.ts
saevarma, line 1: Please export here directly and remove the `testing-system.ts` file.

saevarma (changes requested):
I assume that this PR is moving files from `apps/system-e2e/src/support|utils` to this new library. This PR should also remove those file from `apps/system-e2e` and import from this new library.

andes-it:

andes-it:

PR #12481, opened 2023-10-23
fix(application-estate): Advocate expansion introduced to estate member

kksteini:
District commissioners are requesting that data always be expanded. This is the second iteration on that task.
Turns out there is an advocate property on estate members that wasn't added to the upload data type and this PR fixes that.

Several other fixes in that regards as well.

## Checklist:


joningi98 (approved):
LGTM

PR #12488, opened 2023-10-24
fix(national-registry): Add fate to model

thordurhhh:
## What

Add fate to model

## Why

It is still being relied on by some services. Will remove completely before next release

## Checklist:


jonbjarnio (approved):
LGTM 🤝

PR #12494, opened 2023-10-24
fix(service-portal): Fixing bugs in health

jonbjarnio:
## What

Fixing text and adding feature flag checks

## Why

Bug fixing and better UX

## Checklist:


libs/service-portal/health/src/screens/HealthOverview/HealthOverview.tsx
```diff
@@ -34,6 +36,23 @@ export const HealthOverview = () => {
...
+      const ffEnabled = await featureFlagClient.getValue(
+        `isLicenseServiceDrivingLicenceClientV2Enabled`,
+        false,
+      )
+      if (ffEnabled) {
+        setEnabledPaymentPage(ffEnabled as boolean)
+      }
+    }
+    isFlagEnabled()
+  }, [])
```
osk, line 54: `featureFlagClient` as dependency?

libs/service-portal/health/src/screens/HealthOverview/HealthOverview.tsx
```diff
@@ -34,6 +36,23 @@ export const HealthOverview = () => {
...
+  const [enabledPaymentPage, setEnabledPaymentPage] = useState<boolean>(false)
+
+  useEffect(() => {
+    const isFlagEnabled = async () => {
+      const ffEnabled = await featureFlagClient.getValue(
+        `isLicenseServiceDrivingLicenceClientV2Enabled`,
+        false,
+      )
+      if (ffEnabled) {
+        setEnabledPaymentPage(ffEnabled as boolean)
```
osk, line 50:
```suggestion
        setEnabledPaymentPage(true)
```

since we now it's true, or just pass it without `as`.

osk (approved):
Approve with comments.

PR #12495, opened 2023-10-24
feat(service-portal): add ehic card

thorkellmani:
## What

* Add EHIC to license service

## Checklist:


libs/api/domains/license-service/src/lib/mappers/ehicCardMapper.ts
```diff
@@ -0,0 +1,101 @@
...
+          typedPayload.hasTempCard && typedPayload.tempCardPdf
+            ? {
+                label: getLabel('downloadCard', locale, label),
+                value: getDocument(typedPayload.tempCardPdf, 'pdf'),
+                type: GenericUserLicenseMetaLinksType.Download,
+              }
+            : undefined,
+          {
+            label: getLabel('applyForNewCard', locale, label),
+            value: '/umsoknir/evropska-sjukratryggingakortid',
```
thordurhhh, line 95: Is this for this value?:
https://github.com/island-is/island.is/pull/12495/files#diff-531d0a7b989a8548e3a5d5d0076e87ee38c796c382fe1302b343678ea889b53cR105

thordurhhh (approved):
LGTM. Just one minor question 😺

libs/api/domains/license-service/src/lib/mappers/ehicCardMapper.ts
thorkellmani: different things, but this shouldn't be hardcoded, will fix l8r

libs/clients/p-card/src/clientConfig.json
saevarma, line 1: Looks like there is some codeowner misconfiguration as Core should probably not be a codeowner here.

saevarma (approved):
Core files LGTM

libs/clients/p-card/src/clientConfig.json
thorkellmani: will fix!

PR #12502, opened 2023-10-24
feat(application-ctao): Gopro client connections

Halli1414:
# Gopro client connections

https://featapplication-complaints-to-althingi-om-beta.dev01.devland.is/umsoknir/kvortun-til-umbodsmanns-althingis/

010-2129 has some issues with dataproviders.
010-2719 Works fine to test and others should be okay too.

## What

New shared AttachmentS3Service.
Set up and configure client for connection to Ombudsman gopro service.
Finish mapping data.
PDF generation.
Minor variable naming consistency fixes.

## Why

Reduce code duplication by moving AttachmentS3Service from data-protection-complaint to shared lib.
Establish data communication pipeline from backend to Althingi Ombudsman.

## Screenshots / Gifs

## Checklist:


libs/application/template-api-modules/src/lib/modules/shared/services/attachment-s3.service.ts
```diff
@@ -0,0 +1,95 @@
...
+      ) as Array<{
+        key: string
+        name: string
+      }>
+      if (!answers) continue
+      const list = await this.toDocumentDataList(
+        answers,
+        attachmentAnswerKeys[i],
+        application,
+      )
```
Toti91, line 41: Can be simplified to a more functional approach

```ts
attachmentAnswerKeys.forEach((key) => {
...
})
```

libs/application/template-api-modules/src/lib/modules/templates/complaints-to-althingi-ombudsman/complaints-to-althingi-ombudsman.service.ts
```diff
@@ -1,26 +1,72 @@
...
+        application,
+      )
+    const buffer = await generateComplaintPdf(application)
+    const pdf = {
+      content: buffer.toString('base64'),
+      fileName: 'kvörtun',
+      type: 'Kvörtun',
+      subject: 'Kvörtun',
+    } as DocumentInfo
+
```
Toti91, line 49: No reason to type cast here, can just specify type on initialization

```ts
const pdf: DocumentInfo = { ... }
```

lodmfjord (approved):
LGTM

Toti91 (commented):
Just some syntax nitpicks otherwise looks good but a backend developer should review this

libs/clients/althingi-ombudsman/README.md
saevarma, line 1: Was going to comment on 'ombudsman' if you had accidentally swapped 'o' and 'u' until I googled it and found out that this is the english term for this role 🤣

libs/clients/althingi-ombudsman/src/lib/apiConfiguration.ts
```diff
@@ -0,0 +1,43 @@
...
+  provide: 'AlthingiOmdudsmanClientApiConfiguration',
+  useFactory: (
+    xRoadConfig: ConfigType<typeof XRoadConfig>,
+    config: ConfigType<typeof AlthingiOmbudsmanClientConfig>,
+    idsClientConfig: ConfigType<typeof IdsClientConfig>,
+  ) =>
+    new Configuration({
+      fetchApi: createEnhancedFetch({
+        name: 'clients-althingi-ombudsman',
+        timeout: 30000, // 30 sec timeout
```
saevarma, line 20: This timeout scares me... no user is going to wait for half a minute 🙈

libs/clients/althingi-ombudsman/src/lib/clients-althingi-ombudsman.module.ts
```diff
@@ -0,0 +1,24 @@
...
+
+@Module({
+  exports: [...exportedApis, TokenMiddleware],
+  providers: [
+    ApiConfiguration,
+    {
+      provide: TokenMiddleware,
+      useFactory: (
+        config: ConfigType<typeof AlthingiOmbudsmanClientConfig>,
+        exportedApis,
```
saevarma, line 17:
```suggestion
        securityApi,
```

libs/clients/althingi-ombudsman/src/lib/clients-althingi-ombudsman.config.ts
```diff
@@ -0,0 +1,26 @@
...
+})
+
+export const AlthingiOmbudsmanClientConfig = defineConfig<
+  z.infer<typeof schema>
+>({
+  name: 'AlthingiOmbudsmanClient',
+  schema,
+  load(env) {
+    return {
+      scope: env.optionalJSON('') ?? [''],
```
saevarma, line 17: Skip this?

libs/clients/althingi-ombudsman/src/lib/client-althingi-ombudsman.middleware.ts
```diff
@@ -0,0 +1,37 @@
...
+
+@Injectable()
+export class TokenMiddleware implements Middleware {
+  constructor(
+    private readonly password: string,
+    private readonly username: string,
+    private readonly securityApi: SecurityApi,
+  ) {}
+
+  async pre(context: RequestContext): Promise<FetchParams | void> {
```
saevarma, line 17: So this will be triggered before each request.
What is the lifetime of the token returned by the securityApi? Can it be reused while it is still valid?

libs/clients/althingi-ombudsman/src/lib/apiConfiguration.ts
obmagnusson: The timeout is significant due to the API entering a dormant state after inactivity. Upon reactivation, a 'wake-up' delay occurs, affecting the first user post-inactivity. We've extended the wait time instead of returning an error on application submissions, to avoid user confusion. Plus this API is only used for application submissions.

obmagnusson (commented):
Looks good and is inline with what we talked about. Minor addtional comments from me.

libs/clients/althingi-ombudsman/src/lib/client-althingi-ombudsman.middleware.ts
Halli1414: The token should be valid for 7 days however I think that is a bit long for caching. I'll cache it for 1 day to reduce unnecessary authentication calls.

libs/application/template-api-modules/src/lib/modules/shared/services/attachment-s3.service.ts
```diff
@@ -25,22 +25,16 @@ export class AttachmentS3Service {
...
         name: string
       }>
-      if (!answers) continue
-      const list = await this.toDocumentDataList(
-        answers,
-        attachmentAnswerKeys[i],
-        application,
-      )
+      if (!answers) return
+      const list = await this.toDocumentDataList(answers, key, application)
```
Toti91: Actually, on second look I see that you are awaiting a promise inside the loop, did you test this after changing it? I'm 90% this won't work correctly in a forEach, instead you should use `for (const key of attachmentAnswerKeys)` so that it correctly awaits response in each loop

https://stackoverflow.com/questions/37576685/using-async-await-with-a-foreach-loop

Halli1414: I can only test attachments on feature deploy so I haven't done it yet, but will take one round of testing once the build is done.

Toti91 (approved):

obmagnusson (approved):
LGTM

libs/clients/althingi-ombudsman/src/lib/const.ts
```diff
@@ -0,0 +1,2 @@
+export const SECURITY_TOKEN_KEY = 'securityToken'
```
saevarma, line 1: ~My concern here would be that this key is to generic and should be something like `althingiOmbudsmanSecurityToken`

libs/clients/althingi-ombudsman/src/lib/apiConfiguration.ts
```diff
@@ -25,7 +25,7 @@ export const ApiConfiguration = {
               issuer: idsClientConfig.issuer,
               clientId: idsClientConfig.clientId,
               clientSecret: idsClientConfig.clientSecret,
-              scope: config.scope,
+              scope: [],
```
saevarma, line 28: Not sure why this client library needs `autoAuth` as it is using `securityApi` to get Althingis custom token.

saevarma (approved):
LGTM

PR #12503, opened 2023-10-24
fix(service-portal): UserInfoDetail double query

thorkellmani:
## What

Use lazy query to load NationalRegistryPerson when the feature flag has loaded, never before.

## Why

Don't want any unecessary queries to Soffía

## Checklist:


thordurhhh (approved):
Nice 💯

PR #12509, opened 2023-10-25
fix(application-system-api): Remove nested SequelizeModule.forRootAsync

saevarma:
## What

* Remove nested module calls to `SequelizeModule.forRootAsync` in Application System API.
* Adding optional env variable db config for development to override config to use db proxy.

## Why

`forRootAsync` should only be called on the root module then subsequent calls should use `forFeature` to load specific models into scope.

As the multiplate `forRootAsync` calls are all connecting to the same database and are done without using the `name` property they [override each other](https://docs.nestjs.com/techniques/database#multiple-databases-1). So this setup isn't creating 3 different DB connection pools for more connections.

I discovered this when trying to start the api locally and connecting to the db proxy but constantly saw failed to connect using `dev_db` user.

## Screenshots / Gifs

N/A

## Checklist:


apps/application-system/api/sequelize.config.js
```diff
@@ -1,9 +1,9 @@
 /* eslint-env node */
 module.exports = {
   development: {
-    username: 'dev_db',
-    password: 'dev_db',
-    database: 'dev_db',
+    username: process.env.DB_USER || 'dev_db',
```
Thorrik58, line 4: Could you replace this with a more specific env?

Thorrik58 (changes requested):

Thorrik58 (approved):

PR #12529, opened 2023-10-26
fix: encoding from execsync

lodmfjord:
# ...

When execSync fail it throws with buffer message if no enocding is defined (default is buffer)

This simply defines encoding 

```{encoding: "utf-8"}```

## What

More readable error messages when the commands fail :)

## Why

It makes my job easier. 

## Screenshots / Gifs

## Checklist:

### CODE REVIEW

**Comment by busla** [comment]
Perhaps create a shared lib and wrap `execSync` with proper error handling and encoding?

**Comment by lodmfjord** [comment]
> Perhaps create a shared lib and wrap `execSync` with proper error handling and encoding?

Done,

File: libs/base-utils/project.json [hunk 1]
```diff
@@ -0,0 +1,30 @@
+{
+  "name": "base-utils",
+  "$schema": "../../node_modules/nx/schemas/project-schema.json",
+  "sourceRoot": "libs/base-utils/src",
```
AndesKrrrrrrrrrrr, line 4: Since this is shared, I prefer `libs/utils`, and maybe making this the `base` export.
```suggestion
  "sourceRoot": "libs/utils/src",
```

libs/base-utils/src/index.js
AndesKrrrrrrrrrrr, line 1: This file seems useless (as with other `.js` files). Aren't we transpiling all code we need?

libs/base-utils/src/lib/execsync.js
```diff
@@ -0,0 +1,47 @@
+// @ts-check
+/**
+ * This is written in javascript so it does not need to be transpiled
+ * and can be used in scripts
+ */
```
AndesKrrrrrrrrrrr, line 5: What about `ts-node` for this specific purpose?

scripts/_run-aws-eks-commands.js
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env node
 
 const yargs = require('yargs')
-const { execSync } = require('child_process')
+const { execSync } = require('../libs/base-utils/src')
```
AndesKrrrrrrrrrrr, line 4: OK, so we _do_ use it in non-transpiled code. Why isn't this `ts`? Should it be?

AndesKrrrrrrrrrrr (commented):
How about skipping the wrapper, and write an error linting rule & fix where relevant? That would fix the core issue, and prevent future regressions.

lodmfjord:
I created eslint plugin to do this. And it was useless since most of the files are outside of the scope being linted in CI.

Creating a wrapper lib was also useless since this is rarely used, and the likelyhood of people just using the function directly was most most likely (since the gain is none).

So I wen back to short simple solution. Fix the code.

AndesKrrrrrrrrrrr (approved):

andes-it:

andes-it:

PR #12531, opened 2023-10-26
fix(service-portal): Move sp core strings to core file

thordurhhh:
## What

Move sp core strings to core file

## Why

when you `extract-strings` on a module with a string that includes a namcepace used in another module ... it will cause the current strings to become the only available strings in the CMS for said namespace. 🛑 

## Checklist:


libs/service-portal/core/src/lib/messages.ts
```diff
@@ -1175,4 +1175,16 @@ export const m = defineMessages({
     id: 'service.portal:close-active-document',
     defaultMessage: 'Loka skjali',
   },
+  month: {
+    id: 'service.portal:month',
+    defaultMessage: 'Mánuður',
+  },
+  dateOfOnvoiceShort: {
```
thorkellmani, line 1182:
```suggestion
  dateOfInvoiceShort: {
```

thorkellmani (approved):
woopsy

PR #12550, opened 2023-10-27
feat(user-profile): Create islyklar user profile if it does not exist

GunnlaugurG:
# ...

## What

If user profile does not exist in islyklar db we will need to create it. 

## Why

So the update will not fail if the record does not exist on islyklar service

## Screenshots / Gifs

## Checklist:


apps/services/user-profile/src/app/v2/islykill.service.ts
```diff
@@ -1,11 +1,49 @@
...
+        mobile: userData.mobile,
+        bankInfo: userData.bankInfo,
+        lastLogin: userData.lastLogin,
+        nextLastLogin: userData.nextLastLogin,
+        lastPassChange: userData.lastPassChange,
+        canNudge: userData.canNudge,
+        onlyCert: userData.onlyCert,
+        nudgeLastAsked: userData.nudgeLastAsked,
+      }
+    } catch ({ status, message }) {
```
snaerseljan, line 32: Is this correctly typed? Isn't this unknown?

You could do 
```
if (err instanceof Error) {
    // Now TS correctly types the error
  }
```

snaerseljan: You will probably also have to type the status field

apps/services/user-profile/src/app/v2/islykill.service.ts
```diff
@@ -1,11 +1,49 @@
...
+      if (status === 404) {
+        return {
+          ssn: nationalId,
+          noUserFound: true,
```
snaerseljan, line 36: nit: userNotFound sounds better 😅

apps/services/user-profile/src/app/v2/islykill.service.ts
```diff
@@ -15,20 +53,64 @@ export class IslykillService {
...
-      .islyklarPut({
-        user: {
-          ssn: nationalId,
-          ...(email && { email }),
-          ...(phoneNumber && { mobile: phoneNumber }),
-        },
+    try {
+      const user = {
+        ssn: nationalId,
+        ...(email && { email }),
```
snaerseljan, line 59: What happens if both are undefined?

apps/services/user-profile/src/app/v2/islykill.service.ts
```diff
@@ -15,20 +53,64 @@ export class IslykillService {
...
+        ...(phoneNumber && { mobile: phoneNumber }),
+      }
+
+      return await this.islyklarApi.islyklarPut({ user })
+    } catch ({ message }) {
```
snaerseljan, line 64: same as above

apps/services/user-profile/src/app/v2/islykill.service.ts
```diff
@@ -15,20 +53,64 @@ export class IslykillService {
...
+      throw new InternalServerErrorException(
+        'Unable to update islykill settings for user',
+        message,
+      )
+    }
+  }
+
+  async createIslykillSettings(
+    nationalId: string,
+    { email, mobile }: { email?: string; mobile?: string },
```
snaerseljan, line 74: same as above what happens if both are undefined

apps/services/user-profile/src/app/v2/islykill.service.ts
```diff
@@ -15,20 +53,64 @@ export class IslykillService {
...
+    phoneNumber,
+  }: {
+    nationalId: string
+    email?: string
+    phoneNumber?: string
+  }): Promise<PublicUser> {
+    const { noUserFound } = await this.getIslykillSettings(nationalId)
+
+    if (noUserFound) {
+      return await this.createIslykillSettings(nationalId, {
```
snaerseljan, line 104: nit: Maybe have the same inputs for create and update methods, then you can reuse their input types

apps/services/user-profile/src/app/v2/islykill.service.ts
```diff
@@ -15,20 +53,64 @@ export class IslykillService {
...
-          ...(phoneNumber && { mobile: phoneNumber }),
-        },
+    try {
+      const user = {
+        ssn: nationalId,
+        ...(email && { email }),
+        ...(phoneNumber && { mobile: phoneNumber }),
+      }
+
+      return await this.islyklarApi.islyklarPut({ user })
```
snaerseljan, line 63: You have redundant return await in some places. You should be good by just returning

snaerseljan (approved):
Great job. Just some nits and questions

apps/services/user-profile/src/app/v2/islykill.service.ts
GunnlaugurG: Can I not just do `err: Error` should that not result in the same instance ?

PR #12560, opened 2023-10-30
feat(j-s): Case header info

oddsson:
# Case header info

[Asana](https://app.asana.com/0/1199153462262248/1205752688071902/f)

## What

This PR does three things 

1. Today, we display info about when the district court received an appeal in the alert banner for the district court. When COA has made a ruling, we no longer show that info. This PR adds that info in the header of the case overview
Screenshot 2023-10-30 at 12 58 42

2. We add info about who appealed and when in the header of the case overview. We were displaying that info for COA users but now we display it for all users.

3. Always show the info discussed in pt1 for COA users.

## Why

This gives more users more overview of what's happened in the case so far.

## Checklist:


gudjong (approved):
If I understand the changes correctly, then the new appeal info is shown for all users except defenders. Was that the intention?

oddsson:
You are right, I forgot about the defenders. I've made some changes to that.

PR #12565, opened 2023-10-30
fix(application-estate): Property mapping fixes

kksteini:
Aggressive regex was filtering out legitimate properties, coming in as 6 or 7 digits (missing `F` prefix).
This is causing properties present in the District Commissioner's database to be lost when transitioting to our application system.

Note, that it may look strange to test the data coming from DC but since we're injecting them into the application's answers (for convenience) we validate data the same way we would in the application frontend with our dataschema.

## Checklist:

### DIFF
--- a/libs/clients/syslumenn/src/lib/syslumennClient.utils.ts
+++ b/libs/clients/syslumenn/src/lib/syslumennClient.utils.ts
@@ -321,7 +321,7 @@ export const assetMapper = (assetRaw: EignirDanarbus): EstateAsset => {
   return {
     description: assetRaw.lysing ?? '',
     assetNumber: assetRaw.fastanumer ?? '',
-    share: assetRaw.eignarhlutfall ?? 1,
+    share: assetRaw.eignarhlutfall / 100.0 ?? 1,
   }
 }
 
@@ -339,7 +339,7 @@ export const mapEstateRegistrant = (
             (a) =>
               a.tegundAngalgs === TegundAndlags.NUMBER_0 &&
               a?.fastanumer &&
-              /^[fF]{0,1}\d{7}$/.test(a.fastanumer),
+              /^[Ff]{0,1}\d{7}$|^[Ll]{0,1}\d{6}$/.test(a.fastanumer),
           )
           .map(assetMapper)
       : [],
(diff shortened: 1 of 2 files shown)


libs/clients/syslumenn/src/lib/syslumennClient.utils.ts
```diff
@@ -321,7 +321,7 @@ export const assetMapper = (assetRaw: EignirDanarbus): EstateAsset => {
   return {
     description: assetRaw.lysing ?? '',
     assetNumber: assetRaw.fastanumer ?? '',
-    share: assetRaw.eignarhlutfall ?? 1,
+    share: assetRaw.eignarhlutfall / 100.0 ?? 1,
```
kksteini, line 324: FYI: SyslData is now sending numbers from 0-100 but we're expecting a float between 0.0-1.0

libs/clients/syslumenn/src/lib/syslumennClient.utils.ts
```diff
@@ -392,9 +392,9 @@ export const mapEstateInfo = (syslaData: DanarbuUppl): EstateInfo => {
       ? syslaData.eignir
           .filter(
             (a) =>
+              a?.tegundAngalgs !== undefined &&
```
kksteini, line 395: FYI: Wanted to reverse the boolean checking order but for some reason Typescript complains if we don't explicitly ask for !== undefined. It's serving the same purpose here.

albinagu (approved):

libs/clients/syslumenn/src/lib/syslumennClient.utils.ts
RunarVestmann: What if the assetRaw.eignarhlutfall is null or undefined?

`undefined / 100.0 ?? 1` is equal to NaN when I open up node in a terminal and test this out. Wouldn't it be better if we want to keep the same functionality to just divide if we are sure that we've got a value and otherwise just return 1?

RunarVestmann (changes requested):

libs/clients/syslumenn/src/lib/syslumennClient.utils.ts
kksteini: That's a good point, thanks. Will fix

kksteini: Just as an interesting aside on the differences between undefined and null:
`null / 100.0 ?? 1` evaluates to 0.
The property is asserted to be there (lacking the optional marking `?`).

However, the type being `number` and not `number | null` is in question when inspecting some answers from the client giving us `null` in places so it's wise to guard against `undefined` as well.

kksteini: Ok.
Thinking this over.
Let's guard against `undefined` but not against null.

Since the API is freely belting out `null` values on a type defined as `number` I'm going to let it stand for 0.
I'll raise this issue with the District Commissioner client maintainer

kksteini: Ok, we guard against `undefined` but not `null`.
If we see a `null` it stands for 0.

RunarVestmann (approved):

PR #12572, opened 2023-10-31
fix(service-portal): Update breadcrumbs

thordurhhh:
## What

* Only display go back with arrow on mobile
* Use all ancestors to determine if an item in breadcrumbs has a parameter name

## Why

* Go back could be displayed in the middle of crumbs if middle item has breadCrumbsHidden=true
* Ancestors with param name will display the string, example 
`Yfirlit | id | Nánar` instead of `Yfirlit | 123 | Nánar`.

## Screenshots / Gifs

Was 😞 
Screenshot 2023-10-31 at 10 37 03

Becomes 😃 
Screenshot 2023-10-31 at 10 36 35

## Checklist:


thorkellmani (approved):
LGTM

PR #12576, opened 2023-10-31
feat(user-profile): Max retries for email and phone

GunnlaugurG:
# ...

## What

Counter for failed verification attempts for both email and phone number

## Why

To prevent brute force on verifying email and phone number

## Screenshots / Gifs

## Checklist:

### DIFF
--- a/apps/services/user-profile/src/app/user-profile/verification.service.ts
+++ b/apps/services/user-profile/src/app/user-profile/verification.service.ts
@@ -70,7 +70,7 @@ export class VerificationService {
     const emailCode = this.generateVerificationCode(codeLength)
 
     const [record] = await this.emailVerificationModel.upsert(
-      { nationalId, email, hash: emailCode, created: new Date() },
+      { nationalId, email, hash: emailCode, created: new Date(), tries: 0 },
       {
         returning: true,
       },
--- a/apps/services/user-profile/src/app/user-profile/verification.service.ts
+++ b/apps/services/user-profile/src/app/user-profile/verification.service.ts
@@ -22,6 +22,7 @@ const LOG_CATEGORY = 'verification-service'
 
 export const SMS_VERIFICATION_MAX_AGE = 5 * 60 * 1000
 export const SMS_VERIFICATION_MAX_TRIES = 5
+export const EMAIL_VERIFICATION_MAX_TRIES = 5
 
 /**
   *- email verification procedure
@@ -107,8 +108,11 @@ export class VerificationService {
   async confirmEmail(
     confirmEmailDto: ConfirmEmailDto,
     nationalId: string,
-    transaction?: Transaction,
+    options: { transaction?: Transaction; maxRetries?: number } = {
+      maxRetries: EMAIL_VERIFICATION_MAX_TRIES,
+    },
   ): Promise<ConfirmationDtoResponse> {
+    const { transaction, maxRetries = EMAIL_VERIFICATION_MAX_TRIES } = options
     const verification = await this.emailVerificationModel.findOne({
       ...(transaction && { transaction }),
       where: { nationalId, email: confirmEmailDto.email },
@@ -133,11 +137,21 @@ export class VerificationService {
       }
     }
 
-    if (confirmEmailDto.hash !== verification.hash) {
-      // TODO: Add tries?
+    if (verification.tries >= maxRetries) {
       return {
-        message: `Email verification with hash ${confirmEmailDto.hash} does not exist`,
+        message:
+          'Too many failed email verifications. Please restart verification.',
         confirmed: false,
+      }
+    }
+
+    if (confirmEmailDto.hash !== verification.hash) {
+      await verification.increment({ tries: 1 })
+      const remaining = maxRetries - verification.tries
+      return {
+        message: `Email code is not a match. ${remaining} tries remaining.`,
+        confirmed: false,
+        remainingAttempts: remaining,
       }
     }
 
@@ -178,8 +192,12 @@ export class VerificationService {
   async confirmSms(
     confirmSmsDto: ConfirmSmsDto,
     nationalId: string,
-    transaction?: Transaction,
+    options: { transaction?: Transaction; maxRetries?: number } = {
+      maxRetries: SMS_VERIFICATION_MAX_TRIES,
+    },
   ): Promise<ConfirmationDtoResponse> {
+    const { transaction, maxRetries = SMS_VERIFICATION_MAX_TRIES } = options
+
     const phoneNumber = parsePhoneNumber(confirmSmsDto.mobilePhoneNumber, 'IS')
     const mobileNumber =
       phoneNumber.country === 'IS'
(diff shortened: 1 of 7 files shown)


apps/services/user-profile/src/app/user-profile/verification.service.ts
```diff
@@ -133,11 +137,21 @@ export class VerificationService {
...
+      return {
+        message:
+          'Too many failed email verifications. Please restart verification.',
+        confirmed: false,
+      }
+    }
+
     if (confirmEmailDto.hash !== verification.hash) {
-      // TODO: Add tries?
+      await verification.increment({ tries: 1 })
```
saevarma, line 149: I didn't now about the `increment` method, it should solve our concurrency issue 🙌 
According to the [docs](https://sequelize.org/api/v6/class/src/model.js~model#instance-method-increment) you could do `await verification.increment('tries')`. And it documents that it returns the updated model so don't we need to do

```suggestion
      verification = await verification.increment('tries')
```

apps/services/user-profile/src/app/user-profile/verification.service.ts
```diff
@@ -178,8 +192,12 @@ export class VerificationService {
   async confirmSms(
     confirmSmsDto: ConfirmSmsDto,
     nationalId: string,
-    transaction?: Transaction,
+    options: { transaction?: Transaction; maxRetries?: number } = {
+      maxRetries: SMS_VERIFICATION_MAX_TRIES,
+    },
   ): Promise<ConfirmationDtoResponse> {
+    const { transaction, maxRetries = SMS_VERIFICATION_MAX_TRIES } = options
```
saevarma, line 199: Same simplification of desctructuring here.

apps/services/user-profile/src/app/v2/test/me-user-profile.spec.ts
```diff
@@ -331,7 +331,7 @@ describe('MeUserProfile', () => {
       expect(res.status).toEqual(400)
       expect(res.body).toMatchObject({
         statusCode: 400,
-        message: 'Email verification with hash 000 does not exist',
+        message: 'Email code is not a match. 3 tries remaining.',
```
saevarma, line 334: Weren't we going to allow maximum of 3 tries, so 2 tries should be remaining?
We should add a test case to test that we return the error about max tries reached.

apps/services/user-profile/src/app/v2/user-profile.service.ts
```diff
@@ -114,9 +114,13 @@ export class UserProfileService {
...
       )
 
-      promises.map(({ confirmed, message }) => {
+      promises.map(({ confirmed, message, remainingAttempts }) => {
         if (confirmed === false) {
-          throw new BadRequestException(message)
+          throw new BadRequestException({
+            message,
+            remainingAttempts,
+            statusCode: 400,
```
saevarma, line 122: `BadRequestException` should automatically be a `400` response.

FYI this is a bit tricky to handle. I currently doesn't have any better suggestion, as this won't be used by anyone else then our profile collection UI, were we do confirmation of the email and phone number separately.

apps/services/user-profile/src/app/v2/user-profile.service.ts
```diff
@@ -90,7 +90,7 @@ export class UserProfileService {
       this.formatPhoneNumber(userProfile.mobilePhoneNumber)
 
     return await this.sequelize.transaction(async (transaction) => {
-      const commonArgs = [nationalId, transaction] as const
+      const commonArgs = [nationalId, { transaction, maxRetries: 3 }] as const
```
saevarma, line 93: Maybe we should name the property `maxRetries` --> `maxTries` to as it is clearer about total tries, rather than total tries = maxRetries + 1. You can think about it.

apps/services/user-profile/src/app/user-profile/verification.service.ts
```diff
@@ -107,8 +108,11 @@ export class VerificationService {
   async confirmEmail(
     confirmEmailDto: ConfirmEmailDto,
     nationalId: string,
-    transaction?: Transaction,
+    options: { transaction?: Transaction; maxRetries?: number } = {
+      maxRetries: EMAIL_VERIFICATION_MAX_TRIES,
+    },
   ): Promise<ConfirmationDtoResponse> {
+    const { transaction, maxRetries = EMAIL_VERIFICATION_MAX_TRIES } = options
```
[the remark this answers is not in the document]
GunnlaugurG: I tried that first but had issues with that because the whole object can be nuill that why I need the default object on the argument input level, and it can also only have transaction and not the maxRetries, thats why I need to object destruction. 
If I skip the object destruction in line 115 and send in `{transaction: transaction}` it will not initialize `maxRetries` with the initial value.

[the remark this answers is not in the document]
GunnlaugurG: I haven't tried skipping the default values on the argument level though and only use the object destruction. That would probably work, and is much cleaner

apps/services/user-profile/src/app/user-profile/verification.service.ts
GunnlaugurG: Yes, you are right, I didn't even think about that. It should cover the concurrency issue.

apps/services/user-profile/src/app/v2/user-profile.service.ts
valurefugl: Is this perhaps preferable?
`          throw new BadRequestException({
            message,
            statusCode: 400,
            cause: { remainingAttempts },
          })`

apps/services/user-profile/src/app/v2/test/me-user-profile.spec.ts
GunnlaugurG: fixed with comment above where I should set `verification = await verification.increment('tries')`
And I will add test :D

apps/services/user-profile/src/app/v2/user-profile.service.ts
saevarma: @valurefugl any specific reason for wrapping the `remainingAttempts` in an object?

valurefugl: Just this, https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause

saevarma: As discussed offline we are not going to use the `cause` property for this.

apps/services/user-profile/src/app/user-profile/verification.service.ts
```diff
@@ -138,11 +142,21 @@ export class VerificationService {
       }
     }
 
+    if (verification.tries >= maxTries) {
+      return {
+        message:
+          'Too many failed email verifications. Please restart verification.',
+        confirmed: false,
```
saevarma, line 149: To make it easier for the client to handle both this and 0 remaining attempts. We should provide here the field `remainingAttempts: 0`

saevarma (approved):
LGTM

PR #12581, opened 2023-10-31
feat(web): link card on articles

splicebison:
# Link Card

## What

Allowing Link Card on Articles to be used in place of Process Entries for a cta link experience.

## Why

Moving forward, Process Entries will be reserved only for actual application links. This change replaces the need to use the Process Entry type for regular CTA links that content managers would still like to call out in a similar fashion.

## Checklist:


apps/web/screens/Article/Article.tsx
```diff
@@ -492,7 +492,7 @@ const ArticleScreen: Screen<ArticleProps> = ({
         printHidden
       >
         {/**
```
albinagu, line 494: nitpick: maybe remove the {/** and */} 
since lines below are using //
;D

albinagu (approved):
lgtm

apps/web/screens/Article/Article.tsx
splicebison: This was there previously. I haven't altered this (my ide removed extra spaces around it). But I can remove it

PR #12587, opened 2023-11-01
fix(estate): Adding relation with applicant

albinagu:
## Checklist:


libs/application/templates/estate/src/fields/EstateMembersRepeater/AdditionalEstateMember.tsx
```diff
@@ -143,6 +148,33 @@ export const AdditionalEstateMember = ({
...
+          <InputController
+            id={emailField}
+            name={emailField}
+            label={m.email.defaultMessage}
+            defaultValue={field.email || ''}
+            backgroundColor="blue"
+            error={error?.email}
+            required
+          />
+        </GridColumn>
```
kksteini, line 177: Context: This is necessary because the zod validation on the base level interrupts the type in the application template.

libs/application/templates/estate/src/fields/EstateMembersRepeater/index.tsx
```diff
@@ -7,15 +7,20 @@ import {
   Button,
   GridColumn,
   GridRow,
+  Select,
```
kksteini, line 10: Unused?

libs/application/templates/estate/src/lib/constants.ts
```diff
@@ -36,3 +36,12 @@ export const EstateTypes = {
...
 }
+
+export const relationWithApplicant = [
+  'Barn (sameiginlegt barn hjóna)',
+  'Stjúpbarn (barn látins maka)',
+  'Barnabarn (sameiginlegt barnabarn hjóna)',
+  'Stjúpbarnabarn (barnabarn látins maka)',
+  'Barnabarnabarn (sameiginlegt barnabarnabarn hjóna)',
+  'Stjúpbarnabarnabarn (barnabarnabarn látins maka)',
+]
```
kksteini, line 47: Can we add a comment about why the list is here?

```suggestion
// The District Commissioner's relation endpoint was not suitable for this list.
// Another endpoint won't be created for this static list.
// This is here because it is specific for this application and is not subject to change.
export const relationWithApplicant = [
  'Barn (sameiginlegt barn hjóna)',
  'Stjúpbarn (barn látins maka)',
  'Barnabarn (sameiginlegt barnabarn hjóna)',
  'Stjúpbarnabarn (barnabarn látins maka)',
  'Barnabarnabarn (sameiginlegt barnabarnabarn hjóna)',
  'Stjúpbarnabarnabarn (barnabarnabarn látins maka)',
]
```

libs/application/templates/estate/src/lib/dataSchema.ts
```diff
@@ -132,6 +133,14 @@ export const estateSchema = z.object({
...
         },
       )
+      .refine(
+        ({ enabled, relationWithApplicant }) => {
+          return enabled ? relationWithApplicant !== '' : true
+        },
+        {
+          path: ['relationWithApplicant'],
+        },
+      )
```
kksteini, line 143: Will this work for all four paths of the application?
If the `relationWithApplicant` field isn't visible, the mapper will still default relationWithApplicant to `''`, right?
Is it possible this will make validation fail while this field is invisible?

albinagu: oh, i forgot to remove it, thanks MAN

libs/application/templates/estate/src/lib/constants.ts
albinagu: i think ill double check with the list again

kksteini (approved):

RunarVestmann (approved):
Code owned files look good to me

PR #12588, opened 2023-11-01
chore(j-s): Change background color on files that can't be opened

oddsson:
# Change background color on files that can't be opened

[Anasa](https://app.asana.com/0/1199153462262248/1205710564242572/f)

## What

When we list case files, there is an option to say whether files can be opened or not. When files can't be opened, we now display them gray instead of white to make it clearer that it's disabled

Screenshot 2023-11-01 at 10 07 53

## Checklist:


unakb:
I think there was a misunderstanding about the task, continued convo in Asana

unakb (changes requested):
I think there was a misunderstanding about the task, we wanted to handle file links that display as if they're clickable but they aren't

unakb (approved):
Nice :)!

PR #12596, opened 2023-11-01
feat(j-s): Expose heightened security option for police system cases

unakb:
[Gæti LÖKE fyrirskipað hækkað öryggisstig?](https://app.asana.com/0/1199153462262248/1205570629152969)

## What

Expose heightened security option when police creates cases through API

## Why

So that they can mark their cases as requiring heightened security if needed

## Checklist:


gudjong (changes requested):
This change correctly modifies the internal case controller. However, the xrd api project also needs to be modified to expose the new argument to xroad.

gudjong (approved):

gudjong (approved):

PR #12599, opened 2023-11-01
chore(service-portal): Add tab navigation as button navigation

thordurhhh:
## What

Add tab navigation as button navigation
Add subnav support for tabnav

## Why

There is no support in the <Tabs /> component to support nested navigation.
And this approach is more natural for a nav + screen rendering instead of using the <Tabs />

## Screenshot

Screenshot 2023-11-01 at 15 36 56

## Checklist:


thorkellmani (approved):
LGTM

PR #12601, opened 2023-11-02
chore(j-s): Add tail to all sms and emails with a link to RVG

oddsson:
# Add tail to all sms and emails with a link to RVG

Asana

## What

Add tail to all sms and emails with a link to RVG.

## Checklist:


unakb (approved):

unakb:
Some tests might break, but looks good !

apps/judicial-system/backend/src/app/modules/notification/notification.service.ts
```diff
@@ -853,6 +869,8 @@ export class NotificationService {
           }),
       defenderName ?? '',
       defenderEmail ?? '',
+      [],
```
unakb, line 872: Is there a reason this Attachment argument is a empty array while the others are undefined ?

apps/judicial-system/backend/src/app/modules/notification/notification.service.ts
```diff
@@ -1492,6 +1517,8 @@ export class NotificationService {
       html,
       theCase.registrar.name,
       theCase.registrar.email,
+      [],
```
unakb, line 1520: Same here

unakb (approved):
Just one question otherwise nice

apps/judicial-system/backend/src/app/modules/notification/notification.service.ts
oddsson: Nope, just a mistake..

PR #12607, opened 2023-11-03
feat(user-profile): System lookup endpoint

GunnlaugurG:
## What

A endpoint to fetch user profile by provided national Id for external

## Why

So other services can fetch the user profile from our database if they have sufficient access

## Screenshots / Gifs

## Checklist:


apps/services/user-profile/src/app/v2/test/user-profile.spec.ts
```diff
@@ -0,0 +1,132 @@
...
+import { SequelizeConfigService } from '../../sequelizeConfig.service'
+import { FixtureFactory } from './fixtureFactory'
+
+const testNationalId = '0101302399'
+
+const testUserProfile = {
+  nationalId: testNationalId,
+  email: 'test@test.is',
+  mobilePhoneNumber: '1234567',
+}
```
saevarma, line 17: Use the `createNationalId` and `createPhoneNumber` from `@island.is/testing/fixtures`. It is also nice to use the Faker library to create fake data, i.e. the email.
```suggestion
const testUserProfile = {
  nationalId: createNationalId(),
  email: faker.internet.email(),
  mobilePhoneNumber: createPhoneNumber(),
}
```

apps/services/user-profile/src/app/v2/test/user-profile.spec.ts
```diff
@@ -0,0 +1,132 @@
...
+        AppModule: AppModule,
+        SequelizeConfigService: SequelizeConfigService,
+      })
+
+      const server = request(app.getHttpServer())
+
+      // Act
+      const res = await server
+        .get('/v2/users/.national-id')
+        .set('X-Param-National-Id', '1234567890')
```
saevarma, line 33: We should just use the `testUserProfile.nationalId` as we are testing that the access token is not provided for authentication. Same in the 403 test case below.
```suggestion
        .set('X-Param-National-Id', testUserProfile.nationalId)
```

apps/services/user-profile/src/app/v2/test/user-profile.spec.ts
```diff
@@ -0,0 +1,132 @@
...
+          scope: [UserProfileScope.system],
+        }),
+      })
+
+      server = request(app.getHttpServer())
+    })
+
+    afterEach(async () => {
+      await app.cleanUp()
+    })
```
saevarma, line 90: We usually try to set this up in a `beforeAll`/`afterAll` if it can be reused between tests, so the app is only initialised once. Then we can use `beforeEach`/`afterEach` to restore some state if neede, like truncating a db table or resetting a mock function calls.

apps/services/user-profile/src/app/v2/user-profile.controller.ts
```diff
@@ -0,0 +1,52 @@
...
+    description: 'Get user profile for given nationalId.',
+    response: { status: 200, type: UserProfileDto },
+  })
+  @Audit<UserProfileDto>({
+    resources: (profile) => profile.nationalId,
+  })
+  findUserProfile(
+    @Headers('X-Param-National-Id') nationalId: string,
+  ): Promise<UserProfileDto> {
+    if (!nationalId || nationalId.length !== 10) {
```
saevarma, line 47: We should use the `kennitala` package to verify the nationalId.

saevarma (changes requested):

apps/services/user-profile/src/app/v2/test/user-profile.spec.ts
```diff
@@ -109,17 +114,16 @@ describe('UserProfileController', () => {
 
     it('GET /v2/user/.national-id should return 200 with the UserProfileDto when the User Profile exists in db', async () => {
       // Arrange
-      const fixtureFactory = new FixtureFactory(app)
-      const userProfile = fixtureFactory.createUserProfile(testUserProfile)
+      await new FixtureFactory(app).createUserProfile(testUserProfile)
```
saevarma, line 117: As we now have `beforeAll` the in memory db is shared for all tests so this profile will exist in other test cases. It would also make sense to make the `FixtureFactory` available in the `describe` scope like the `server`. 
I think we should use the `beforeEach` to clear the db before each test case so we prevent any data leakage between test cases.

apps/services/user-profile/src/app/v2/user-profile.controller.ts
```diff
@@ -1,21 +1,22 @@
...
 import { Documentation } from '@island.is/nest/swagger'
 import { Audit, AuditService } from '@island.is/nest/audit'
 import { UserProfileScope } from '@island.is/auth/scopes'
 import { IdsAuthGuard, Scopes, ScopesGuard } from '@island.is/auth-nest-tools'
 
-import { UserProfileDto } from './dto/user-profileDto'
+import { UserProfileDto } from './dto/user-profile.dto'
 import { UserProfileService } from './user-profile.service'
 
+import * as kennitala from 'kennitala'
```
saevarma, line 18: ~ Import sort 😉

saevarma (approved):
🚀

PR #12616, opened 2023-11-03
feat(user-profile-api): Needs nudge refactor. Refactor removing values. Add tests.

saevarma:
## What

This PR implements the logic when the user needs nudge or should be prompted with onboarding (needsNudge=null)
* needsNudge logic with test cases.
* Change to use empty string instead of null to remove email or phone.
* Don't call islyklar API if not changing email and phone.

## Why

So user profile UI can correctly navigate the user to appropriate screens.

## Screenshots / Gifs

N/A

## Checklist:


apps/services/user-profile/src/app/v2/user-profile.service.ts
```diff
@@ -82,12 +86,11 @@ export class UserProfileService {
       )
     }
 
-    const { nationalNumber, parsedPhoneNumber } =
+    const formattedPhoneNumber =
```
GunnlaugurG, line 89: If the phone number is not defined then `formattedPhoneNumber` will be `false` and when islyklar service is called with `false` as phone number it will result in `500 Internal Server Error`.. This should by `const formattedPhoneNumber = isMobilePhoneNumberDefined ? formatPhoneNumber(userProfile.mobilePhoneNumber) : null`

apps/services/user-profile/src/app/v2/user-profile.service.ts
```diff
@@ -1,7 +1,8 @@
 import { BadRequestException, Inject, Injectable } from '@nestjs/common'
 import { InjectModel } from '@nestjs/sequelize'
-import { parsePhoneNumber } from 'libphonenumber-js'
 import { isEmail } from 'class-validator'
+import subMonths from 'date-fns/subMonths'
+import { CountryCallingCode, parsePhoneNumber } from 'libphonenumber-js'
 import pick from 'lodash/pick'
```
GunnlaugurG, line 6: ~Unused imports

apps/services/user-profile/src/app/app.module.ts
```diff
@@ -18,6 +19,7 @@ import { UserProfileModule as UserProfileV2Module } from './v2/user-profile.modu
     }),
     UserProfileModule,
     UserProfileV2Module,
+    ProblemModule,
```
GunnlaugurG, line 22: Question: Is this the reason that we now return title and type in the BadRequestExceptions ?

libs/testing/fixtures/src/lib/phoneNumber.ts
```diff
@@ -0,0 +1,6 @@
+import faker from 'faker'
+
+/**
+ * Creates a random 7 digit phone number
```
GunnlaugurG, line 4: ~ Maybe add to the description that all phone numbers start with the digit 7.

apps/services/user-profile/src/app/v2/test/me-create-verification.spec.ts
```diff
@@ -5,7 +5,7 @@ import { createCurrentUser } from '@island.is/testing/fixtures'
 import { UserProfileScope } from '@island.is/auth/scopes'
 import { setupApp } from '@island.is/testing/nest'
 
-import { FixtureFactory } from './fixtureFactory'
+import { FixtureFactory } from '../../../../test/fixture-factory'
```
GunnlaugurG, line 8: ~SmsVerification: unused import

GunnlaugurG (changes requested):
Nice cleanup and improvements. Just one bug with the phonenumber and minor nits.

GunnlaugurG (approved):
LGTM!

snaerseljan (approved):

PR #12648, opened 2023-11-07
feat(service-portal): hms loans client

Valur:
https://www.notion.so/L-na-og-grei-sluyfirlit-HMS-516aac3051c2425aaa433b3864fe1049

## What

Add client and api domain for hms loans service

## Why

To be able to fetch loan information in service portal

## Checklist:


libs/api/domains/hms-loans/src/lib/models/loanOverview.model.ts
```diff
@@ -0,0 +1,142 @@
+import { Field, ObjectType } from '@nestjs/graphql'
```
thordurhhh, line 1: We should probably translate these models.

Valur: I did consider translating it but I think we would need some finance specialist to do it 🙈

libs/api/domains/hms-loans/src/lib/api-domains-hms-loans.resolver.ts
```diff
@@ -0,0 +1,36 @@
...
+  @Audit()
+  async getHmsLoansLoanOverview(@CurrentUser() user: User) {
+    return await this.hmsLoansService.getHmsLoansLoanOverview(user)
+  }
+
+  @Query(() => [PaymentOverview], { nullable: true })
+  @Audit()
+  async getHmsLoansPaymentOverview(@CurrentUser() user: User) {
+    return this.hmsLoansService.getHmsLoansPaymentOverview(user)
+  }
```
saevarma, line 35: The [naming convention](https://github.com/island-is/island.is/pull/12730) of queries would be `hmsLoansLoanOverview` and `hmsLoansPaymentOverview` where they start with they domain name `hmsLoans` then a noun for the object/collection being queried. But I would also consider shortening them to `hmsLoans` for the list of loans (we don't need to repeat `Loans` when the object name is in the domain name) and `hmsLoansPayments`

```suggestion
  @Query(() => [LoanOverview], { name: 'hmsLoans', nullable: true })
  @Audit()
  async getHmsLoans(@CurrentUser() user: User) {
    return await this.hmsLoansService.getHmsLoansLoanOverview(user)
  }

  @Query(() => [PaymentOverview], { name: 'hmsLoansPayments', nullable: true })
  @Audit()
  async getHmsLoansPayments(@CurrentUser() user: User) {
    return this.hmsLoansService.getHmsLoansPaymentOverview(user)
  }
```

libs/api/domains/hms-loans/src/lib/models/loanOverview.model.ts
```diff
@@ -0,0 +1,142 @@
...
+export class Medgreidendur {
+  @Field(() => String, { nullable: true })
+  nafnMedgreidandi?: string | null
+
+  @Field(() => String, { nullable: true })
+  medGreidandi?: string | null
+}
+
+@ObjectType('HmsLoansLoanOverview')
+export class LoanOverview {
```
saevarma, line 31: I would argue that this object should just be named `Loan`. It is implied to being an overview if you have list of `Loan` objects.

```suggestion
@ObjectType('HmsLoansLoan')
export class Loan {
```

libs/api/domains/hms-loans/src/lib/models/loanOverview.model.ts
saevarma: I would like to push for this to be translated to english 🤮

libs/api/domains/hms-loans/src/lib/models/paymentOverview.model.ts
```diff
@@ -0,0 +1,34 @@
+import { Field, ObjectType } from '@nestjs/graphql'
+
+@ObjectType('HmsLoansPaymentOverview')
+export class PaymentOverview {
```
saevarma, line 4: Same here for payments.
```suggestion
@ObjectType('HmsLoansPayment')
export class Payment {
```

libs/clients/hms-loans/.babelrc
saevarma, line 1: clients libraries shouldn't need `.babelrc` file. This is probably just from the generator you used to create the library.
If you use nest generator this file is not created.
```
yarn nx g @nx/nest:lib library-name --dry-run
```

saevarma (approved):
Few recommendations about naming otherwise LGTM

libs/clients/hms-loans/.babelrc
Valur: Removed

libs/api/domains/hms-loans/src/lib/models/paymentOverview.model.ts
Valur: Fixed

libs/api/domains/hms-loans/src/lib/models/loanOverview.model.ts
Valur: Fixed

libs/api/domains/hms-loans/src/lib/models/loanOverview.model.ts
Valur: I created a separate ticket for translating the schema
https://www.notion.so/Translate-loan-graphql-schema-to-english-f93a385f6c6544418289d5f0ce56f051

thorkellmani (approved):
LGTM (that icelandic model tho 🫠 )

peturgq:
Infra part looks good, however I can see not only are some of the models in Icelandic but also most of the classes and fields.

Please review this coding standards document
https://docs.devland.is/technical-overview/code-standards#language

Valur:
This PR is on hold while the api is being translated, should happen in the next couple of days

libs/api/domains/hms-loans/src/lib/models/loanOverview.model.ts
Valur: Api has been translated

PR #12653, opened 2023-11-08
fix(estate): tweak for rateofexchange calculations

albinagu:
Allow user to input both dot and commas

Screenshot 2023-11-08 at 09 48 17

## Checklist:


libs/application/templates/estate/src/lib/utils.ts
```diff
@@ -34,3 +34,6 @@ export function isEstateInfo(
 
 export const isValidString = (string: string | undefined) =>
   string && /\S/.test(string)
+
+export const isNumericalString = (string: string | undefined) =>
+  string && /^[0-9]/.test(string)
```
kksteini, line 39: This is testing whether something begins with a number. Is that sufficient?
F.X. 
```typescript
/^[0-9]/.test("0abc") ==> true
```

Do we want something like "all characters are numbers and at least one number"?

Then maybe
```typescript
/^[0-9]+$/.test("0abc") ==> false
/^[0-9]+$/.test("1234") ==> true
```

kksteini: Furthermore, will punctuation enter into this?
If we want to support something like `0.1234` and `1,000,000` then perhaps `/^[0-9]+$|^[0-9][0-9,.]+[0-9]$/` would do the trick?

Playing around with it yields:
```typescript
> b
/^[0-9]+$|^[0-9][0-9,.]+[0-9]$/

> b.test("0.123")
true

> b.test("1")
true

> b.test(",0,2")
false

> b.test(".")
false

> b.test("1,000,000")
true

> b.test("0abc")
false

> b.test("abc0")
false

> b.test("20.3")
true
```

albinagu: totally, i needed help with that! thanks <3

albinagu: will 0,123 pass?

kksteini: Yes, that bad?

kksteini (approved):

PR #12668, opened 2023-11-08
fix(service-portal): Vehicles display values and unit display

thordurhhh:
## What

* Add missing units
* Remove unused fields (co should not be displayed in lookup)

Refactor gql 🪄 

## Why

* Missing units
* co should not be displayed in lookup

## Checklist:


thorkellmani (approved):
LGTM

PR #12687, opened 2023-11-09
test(j-s): Fix e2e test

unakb:
## What

Add in checking a radio button

## Why

To fix tests

## Checklist:

### DIFF
--- a/apps/system-e2e/src/tests/judicial-system/smoke/custody-prosecutor.spec.ts
+++ b/apps/system-e2e/src/tests/judicial-system/smoke/custody-prosecutor.spec.ts
@@ -43,7 +43,7 @@ test.describe('Custody Prosecutor', () => {
     await page
       .locator('input[name=defenderEmail]')
       .fill('jl-auto-defender@kolibri.is')
-    await page.getByRole('checkbox').last().check()
+    await page.locator('input[name=defender-access-no]').click()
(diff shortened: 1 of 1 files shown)


apps/system-e2e/src/tests/judicial-system/smoke/custody-prosecutor.spec.ts
```diff
@@ -43,7 +43,7 @@ test.describe('Custody Prosecutor', () => {
     await page
       .locator('input[name=defenderEmail]')
       .fill('jl-auto-defender@kolibri.is')
-    await page.getByRole('checkbox').last().check()
+    await page.locator('input[name=defender-access-no]').click()
```
AndesKrrrrrrrrrrr, line 46:
```suggestion
    await page.getByRole('input', { name: 'defender-access-no' }).click()
```

AndesKrrrrrrrrrrr (approved):
Should preferably use the accessible selectors, but if you can't get it to work, then so be it. Passing tests are better than failing tests.

apps/system-e2e/src/tests/judicial-system/smoke/custody-prosecutor.spec.ts
unakb: I just tried with getByRole and I can't get it to work :(

PR #12693, opened 2023-11-09
fix(user-profile): Reverting using formatted phone number in sms verification

saevarma:
## What

Revert using formatted phone number in sms verification as Nova SMS service throws and error on the formatted number.

## Why

To fix error in sending sms with Nova.

## Screenshots / Gifs

N/A

## Checklist:

- [ ] I have performed a self-review of my own code
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] Formatting passes locally with my changes
- [ ] I have rebased against main before asking for a review

PR #12695, opened 2023-11-09
fix(user-profile): Parse phone number to find match in confirmSMS

saevarma:
## What

Continue fix in #12693 that the phonenumber  needs to be parsed in confirmSMS

## Why

So confirmSMS finds the match and succeeds to confirm the sms code.

## Screenshots / Gifs

N/A

## Checklist:


GunnlaugurG (approved):
lgtm

PR #12706, opened 2023-11-10
fix(portals-admin-icelandic-names-registry): Fix name update mutation

saevarma:
## What

As the query results from the formState contains the internal GQL `__typename` property it breaks the mutation as the input type of the mutation doesn't expect this field.

## Why

So users can update the name after it has been initally created.

## Screenshots / Gifs

N/A

## Checklist:


kksteini (approved):

lodmfjord (approved):
LGTM

baering (approved):

PR #12729, opened 2023-11-13
chore(delegation-api): Increase memory of delgation api

saevarma:
## What

Increase memory of delgationapi to request 256m and limit 512m to make it more stable while we debug.

## Why

The delegation api is being restarted multiple times a day, so trying to increase the memory to make it more stable.
The usage of the delegation system has increased but there might also be some expensive methods in the application code that can be improved.

This hopefully makes the api more stable while we work on fixes.

## Screenshots / Gifs

N/A

## Checklist:


lodmfjord (approved):
LGTM

PR #12747, opened 2023-11-14
fix(web): Use <a> tags for links that are pointing to a project path (like /minarsidur)

RunarVestmann:
# Use `<a>` tags for links that are pointing to a project path (like /minarsidur)

## What

* When users click on a link to /samradsgatt on the web it'll show them an error page instead of redirecting them to the consultation portal since we're at the moment using NextLinks

There are two downsides to this approach
1. We need to maintain the `ProjectBasePath` enum in the code and make sure we update it when new projects are added
2. This might mean that sometimes we'll create `<a>` tags when they technically aren't necessary (this can be the case if a project has a base path that uses the Link component

## Why

* We want links to work correctly in next.js when they point to another hosted project

## Checklist:


libs/shared/utils/src/lib/shouldLinkBeAnAnchorTag.ts
```diff
@@ -0,0 +1,11 @@
...
+
+/** Some links point users to another project (like if a user clicks on a /minarsidur link in the web project),
+ * then the link should be an anchor tag instead of a nextlink for example
+ * */
+export const shouldLinkBeAnAnchorTag = (path: string) => {
+  for (const basePath of Object.values(ProjectBasePath)) {
+    if (path.includes(basePath)) return true
+  }
+  return false
+}
```
saevarma, line 11: ?
```suggestion
export const shouldLinkBeAnAnchorTag = (path: string) => Object.values(ProjectBasePath).includes(path)
```

RunarVestmann: Let's say the path is "/minarsidur/eitthvad"

If that gets passed in it'll return true with the previous version of the function but if we use your version of the function it'll return false.

saevarma: Yup agreed. You version handles nested routes which we need to do.

libs/shared/constants/src/lib/projectBasePath.ts
```diff
@@ -0,0 +1,7 @@
+export enum ProjectBasePath {
+  ServicePortal = '/minarsidur',
+  ConsultationPortal = '/samradsgatt',
+  ApplicationSystem = '/umsoknir',
+  AirDiscountScheme = '/loftbru',
```
saevarma, line 5: AirDiscountScheme has been moved so `/loftbru` is actually a page on the web and the application link is now `/minarsidur/loftbru`. So we should drop this so we don't render `<a>` tag for a internal link.

libs/shared/constants/src/lib/projectBasePath.ts
```diff
@@ -0,0 +1,7 @@
...
+  AdminPortal = '/stjornbord',
+}
```
saevarma, line 7: We should add `CarRecycling = '/app/skilavottord`` for skilavottord-web

saevarma (approved):
Just minor changes on the paths otherwise looks good to me.

libs/shared/constants/src/lib/projectBasePath.ts
RunarVestmann: Done

thorkellmani (approved):
LGTM

PR #12750, opened 2023-11-14
feat(web): Charts

baering:
### Be able to fetch and make use of remote statistical data
A way to fetch remote statistical data from an external source and make it available through a micro-service that can be queried from `api`.

The micro-service uses `EnhancedFetchAPI` for data fetching, results are cached using redis.

### New Contentful Content-Types in `Contentful` and components in `web` to display them
New Content-Types that content editors can make use of to display data from external statistical or other data sources. New components in `web` are responsible for displaying content from the new Content-Types. The data is displayed in a way that is accessible to anyone.

#### Chart
A wrapper to render any graphical representation of data using `ChartComponent`s.

At the most basic level a `Chart` has at least the following fields:
- `title`
- `description`
- one or more `ChartComponent`
- data range and / or number of data points to display

A `Chart` is rendered as a slice and can be embedded to any richText.

#### ChartComponent
A `ChartComponent` is anything that presents data within a `Chart`

At the most basic level a `ChartComponent` has at least the following fields:
- `type`: for example a `line` or a `bar`
- `sourceDataKey`: data handed to the chart is a list of objects, this is the id of the field within each entry you wish to render using this component

##### Example of `sourceDataKey`
The chart is given an array of data

```typescript
type Item = {
  apples: number
  oranges: number
}

type Data = Item[]
```

To render how the number of apples changes over time in a line we create a new `Chart`, give it a title and description and then create a new `ChartComponent`. The component has type of `line` and `sourceDataKey` of `apples`.

#### ChartNumberBox
A standalone component to display a value for a data key and optionally how it has evolved over a period of time.

At the most basic level `ChartNumberBox` has the following fields:
- `title`
- `description`: to give more information on tooltip hover and for screen readers
- `sourceDataKey`: to control the value being displayed as explained earlier
- `date`: optional and if left out will use the most recent value for the `sourceDataKey`

## Why

To be able to tell the story of our journey through statistics.

### As seen by Contentful editors
image

### As displayed in `web`
image

## Checklist:


peturgq (approved):

apps/web/components/Charts/v2/constants.ts
```diff
@@ -0,0 +1,65 @@
...
+/**
+ * By default this will be the order of colors given to chart components
+ * Unless they specifically opt to overwrite them
+ */
+export const DEFAULT_COLORS = [
+  theme.color.blue300,
+  theme.color.blueberry300,
+  theme.color.purple300,
+  theme.color.roseTinted300,
+  // '#0061FF',
```
RunarVestmann, line 47: Are these colors here commented out for a reason, I'm mostly wondering if we can delete the comments if we won't use the hex colors here

apps/web/components/Charts/v2/hooks/accessibility.ts
```diff
@@ -0,0 +1,14 @@
+import { useMemo } from 'react'
+
+import { Chart } from '@island.is/web/graphql/schema'
+
+export const useGetChartTableSettings = ({ components, xAxisKey }: Chart) =>
+  useMemo(() => {
+    const tableHead = [...components.map((c) => c.sourceDataKey)]
```
RunarVestmann, line 7: I think spreading the result of a .map() inside an array isn't necessary

apps/web/components/Charts/v2/renderers/ChartNumberBox/ChartNumberBox.css.ts
```diff
@@ -0,0 +1,48 @@
...
+export const wrapper = style({
+  width: '100%',
+  display: 'flex',
+  flexDirection: 'column',
+  gap: theme.spacing[2],
+  maxWidth: '100%',
+  flexWrap: 'wrap',
+
+  '@media': {
+    [`screen and (min-width: ${theme.breakpoints.xl}px)`]: {
```
RunarVestmann, line 14: Is it possible to use themeUtils.responsiveStyles here and get the same result?

apps/web/components/Charts/v2/hooks/data.ts
```diff
@@ -0,0 +1,62 @@
...
+  chartType,
+}: UseGetChartDataProps) => {
+  const queryResult = useQuery<QueryResponse>(GET_MULTIPLE_STATISTICS, {
+    variables: {
+      input: {
+        sourceDataKeys: components.map((c) => c.sourceDataKey),
+        dateFrom: dateFrom ?? undefined,
+        dateTo: dateTo ?? undefined,
+        numberOfDataPoints: numberOfDataPoints ?? undefined,
+        interval: undefined, // TODO
```
RunarVestmann, line 41: What does this TODO here mean 🙃

apps/web/components/Charts/v2/hooks/data.ts
```diff
@@ -0,0 +1,62 @@
...
+}
+
+export const useGetChartData = ({
+  dateFrom,
+  dateTo,
+  numberOfDataPoints,
+  components,
+  chartType,
+}: UseGetChartDataProps) => {
+  const queryResult = useQuery<QueryResponse>(GET_MULTIPLE_STATISTICS, {
```
RunarVestmann, line 34: Have you considered using graphql field resolvers instead of making a separate query on the client to get the graph data?

RunarVestmann (commented):
I skimmed over the PR and have a few comments, feel free to poke me on Slack if you want me to look at the PR again 🙃

apps/web/components/Charts/v2/hooks/data.ts
baering: I tried to make `Chart` a flexible drop-in embed for any rich text, and that's why they fetch their own data like this. If we switch to a field resolver, do we need to query the data separately for each page that uses them?

RunarVestmann: If you look at the resolvers in the `libs/cms/src/lib/cms.resolver.ts` file you'll see that we can create a chart resolver so that everytime a chart is being resolved we can fetch it's relevant data. And that can be cached so if the same chart is being resolved within the cache time then the answer will just be there in the cache instead of us having to fetch it every time. However you are achieving this caching also when calling via the client but the user experience is a bit different.  

1. If we have a field resolver then the initial page will be more "complete" since the data will be there (but initial load time might increase if we get a cache miss)
2. If we fetch via the client then the user gets the initial page quicker but the charts will have to wait to render until there is an answer

I just wanted to mention that option 1 is available if you deem that better but honestly I think that option 2 is fine as well

apps/web/components/Charts/v2/renderers/AccessibilityTableRenderer.tsx
```diff
@@ -0,0 +1,73 @@
...
+              return (
+                <tr>
+                  <th scope="row">
+                    {xAxisValueType === 'date'
+                      ? formatDate(xAxisValue)
+                      : xAxisValue}
+                  </th>
+                  {tableSettings.tableHead.map((key) => {
+                    // eslint-disable-next-line
+                    // @ts-ignore
```
RunarVestmann, line 60: Can we not add a type hint to the key instead of ts-ignoring the line?

apps/web/components/Charts/v2/renderers/ChartComponentRenderer.tsx
```diff
@@ -0,0 +1,164 @@
...
+}
+
+type CustomLabelProps = {
+  cx: number
+  cy: number
+  midAngle: number
+  outerRadius: number
+  innerRadius: number
+  percent: number
+  payload?: {
```
RunarVestmann, line 64: Why not have the value property defined here instead of ts-ignoring below?

apps/web/components/Charts/v2/renderers/ChartNumberBox/ChartNumberBox.tsx
```diff
@@ -0,0 +1,126 @@
...
+  if (!queryResult.data || queryResult.data.length === 0) {
+    return <p>{messages[activeLocale].noDataForChart}</p>
+  }
+
+  return (
+    <div className={styles.wrapper}>
+      {boxData.map((data, index) => {
+        const value =
+          // eslint-disable-next-line
+          // @ts-ignore
```
RunarVestmann, line 88: Can we somehow make this more typesafe?

apps/web/components/Charts/v2/renderers/ChartRenderer.tsx
```diff
@@ -0,0 +1,162 @@
...
+        id={skipId}
+        chart={slice}
+        componentsWithAddedProps={componentsWithAddedProps}
+        data={queryResult.data}
+      />
+    </Box>
+  )
+}
+
+const CARTASIAN_GRID_HELPERS = [
```
RunarVestmann, line 125: It should be spelled "CARTESIAN" right?

apps/web/components/Charts/v2/renderers/LegendRenderer.tsx
```diff
@@ -0,0 +1,138 @@
...
+        flexWrap: 'wrap',
+        columnGap: '15px',
+        rowGap: '15px',
+        marginTop: '30px',
+        justifyContent: 'center',
+      }}
+    >
+      {payload.map((entry, index) => {
+        // eslint-disable-next-line
+        // @ts-ignore yes it exists
```
RunarVestmann, line 53: Can we somehow make this more typesafe?

apps/web/components/Charts/v2/renderers/LegendRenderer.tsx
```diff
@@ -0,0 +1,138 @@
...
+  data,
+}: LegendRendererProps) => {
+  return (
+    <Legend
+      aria-hidden="true"
+      content={(props) => (
+        <CustomLegendRenderer
+          {...props}
+          // eslint-disable-next-line
+          // @ts-ignore
```
RunarVestmann, line 130: Can we somehow make this more typesafe?

apps/web/components/Charts/v2/renderers/TooltipRenderer.tsx
```diff
@@ -0,0 +1,59 @@
...
+  const isActive = active && payload && payload.length
+
+  if (!isActive) {
+    return null
+  }
+
+  return (
+    <ul
+      style={{
+        backgroundColor: 'white',
```
RunarVestmann, line 18: I'm wondering if this could be done via a vanilla extract className instead of being inlined 🤷

apps/web/components/Charts/v2/utils/chart.ts
```diff
@@ -0,0 +1,43 @@
...
+  if (types.length === 0) {
+    // No known types found in chart component
+    // So we cant decide chart base
+    return null
+  } else if (types.length === 1) {
+    // Only have one known type, return it
+    return types[0] as ChartType
+  } else if (types.includes('pie')) {
+    // We have multiple types, and pie is one of them
+    console.error(`Pie compnent can not be used with other component types`)
```
RunarVestmann, line 38: There's a typo here

RunarVestmann (approved):
Stefna files look good to me, there are still a few minor things I commented on but I'll approve.

apps/web/components/Charts/v2/renderers/ChartRenderer.tsx
```diff
@@ -0,0 +1,162 @@
...
+    style={{
+      fontSize: '1rem',
+      fontFamily: 'IBM Plex Sans',
+    }}
+    tickFormatter={(v) => {
+      if (typeof v === 'number') {
+        if (v < 1000) {
+          return v
+        }
+        return `${Math.round(v / 1000)} þ`
```
splicebison, line 157: The feature deploy is broken so I haven't checked locally yet but what about internationalization here?

baering: Well spotted. I'll fix this.

Unfortunately the feature deployment does not work because this change is dependant on `search-indexer`.

splicebison (approved):
Given the magnitude of this PR, I'd prefer seeing some tests along with it. Other than that and my one comment about internationalization, this is approved.

apps/web/components/Charts/v2/renderers/ChartRenderer.tsx
splicebison: No worries, just wasn't sure if I was missing something. Thanks for addressing.

baering:
Waiting on #12839 to fix build error

saevarma (approved):
Really clean PR. Great Job! 🚀

PR #12769, opened 2023-11-15
feat(signature-collection): Portals and api

alexdiljar:
Creating the initial signature collection domain. Mocking responses to enable the development of the project's frontend in applications, service portal and admin portal. The client will be done and connected in a separate PR. 
Endpoints mock the endpoints planned out here https://www.notion.so/Me-m-lakerfi-2-0-2f4d1e51c8994594a84c56f5f8d88da0?pvs=4 

## Checklist:


Thorrik58 (approved):
LGTM on the one file owned by application system team

package.json
```diff
@@ -330,6 +330,7 @@
     "@nx/eslint-plugin": "16.9.0",
     "@nx/express": "16.9.0",
     "@nx/jest": "16.9.0",
+    "@nx/js": "16.9.0",
```
saevarma, line 333: Nice to conform this to othe Nx packages. We should then remove `@nrwl/js` because this package is a rename of that package.

package.json
```diff
@@ -283,7 +283,7 @@
     "testcontainers": "7.23.2",
     "to-querystring": "1.1.1",
     "triple-beam": "1.3.0",
-    "tslib": "1.14.1",
+    "tslib": "2.3.0",
```
saevarma, line 286: Looking at the change log we should go to the latest available version, 2.6.2

package.json
```diff
@@ -263,7 +263,7 @@
     "regenerator-runtime": "0.13.7",
     "remark-parse": "9.0.0",
     "rosetta": "1.0.0",
-    "rxjs": "7.0.0",
+    "rxjs": "^7.8.0",
```
saevarma, line 266: We always pin our versions and looking at the changelog we should go to the latest available verison, 7.8.1

package.json
saevarma, line 1: Not sure why those package updates are part of this PR but I like it :)

libs/api/domains/signature-collection/project.json
saevarma, line 1: Is this `SignatureCollection` not same as `EndorsementSystem` ?

libs/api/domains/signature-collection/src/lib/dto/area.input.ts
```diff
@@ -0,0 +1,9 @@
+import { IsString } from 'class-validator'
+import { Field, InputType } from '@nestjs/graphql'
+
+@InputType()
+export class AreaInput {
```
saevarma, line 5: Mutations, Queries, Types and Models needs to be prefixed with the api domain name.

https://docs.devland.is/technical-overview/api-design-guide/graphql-naming-conventions#integrating-naming-conventions-into-shared-api

You can either name the class `SignatureCollectionAreaInput` or use the `@InputType` decorator to control the GQL schema name and keep the class name short as it is.

```
@InputType('SignatureCollectionAreaInput')
export class AreaInput {
```

libs/api/domains/signature-collection/src/lib/signatureCollection.module.ts
```diff
@@ -0,0 +1,10 @@
+import { Module } from '@nestjs/common'
+import { SignatureCollectionResolver } from './signatureCollection.resolver'
+import { SignatureCollectionService } from './signatureCollection.service'
+
+@Module({
+  controllers: [],
+  providers: [SignatureCollectionService, SignatureCollectionResolver],
+  exports: [],
```
saevarma, line 8: ~ `controllers` and `exports` can be removed.

libs/api/domains/signature-collection/src/lib/signatureCollection.resolver.ts
```diff
@@ -0,0 +1,199 @@
...
+import { SignatureCollectionIdInput } from './dto/id.input'
+import { SignatureCollectionSignature } from './models/signature.model'
+import { SignatureCollectionNationalIdsInput } from './dto/signatureListNationalIds.input'
+import { SignatureCollectionBulk } from './models/bulk.model'
+import { SignatureCollectionSignee } from './models/signee.model'
+import { SignatureCollectionListInput } from './dto/singatureList.input'
+import { SignatureCollectionFindSignatureInput } from './dto/findSignature.input'
+import { SignatureCollectionAreaInput } from './dto/area.input'
+
+@UseGuards(IdsAuthGuard, IdsUserGuard)
```
saevarma, line 22: You don't need both `IdsAuthGuard` and `IdsUserGuard`. `IdsUserGuard` is more strict as it extends the `IdsAuthGuard` but requires that a user `nationalId` is present in the token. I assume from the context that there is always authenticated users there so we should prefer the `IdsUserGuard`.

Then when you add logic to the service layer you have to figure out if we should apply scope check here, or if there is a microservice behind this that will do scope check for user authorization.

libs/api/domains/signature-collection/src/lib/signatureCollection.service.ts
```diff
@@ -0,0 +1,291 @@
...
+
+  async test(): Promise<SignatureCollectionSuccess> {
+    return new Promise((resolve) => {
+      setTimeout(() => {
+        resolve({ success: true })
+      }, 300)
+    })
+  }
+
+  //   CanCreate
```
saevarma, line 31: ~ Redundant comments as the method names are self-explanatory

package.json
```diff
@@ -263,7 +263,7 @@
     "regenerator-runtime": "0.13.7",
     "remark-parse": "9.0.0",
     "rosetta": "1.0.0",
-    "rxjs": "7.0.0",
+    "rxjs": "^7.8.1",
```
saevarma, line 266: We want to pin our versions

```suggestion
    "rxjs": "7.8.1",
```

saevarma (approved):
Approved with minor comments that should be addressed before re-applying the `automerge` label.

PR #12775, opened 2023-11-15
fix(service-portal): Don't render button element inside anchors

Valur:
## What

Render <Button> component as span when inside links

## Why

<button> inside link is not valid

## Checklist:

### DIFF
--- a/libs/service-portal/information/src/components/PersonalInformation/Forms/ProfileForm/components/FormButton.tsx
+++ b/libs/service-portal/information/src/components/PersonalInformation/Forms/ProfileForm/components/FormButton.tsx
@@ -27,11 +27,15 @@ export const FormButton: FC<React.PropsWithChildren<Props>> = ({
         {children}
       </Button>
     ) : (
-      <button type="submit" disabled={disabled}>
-        <Button variant="text" size="small" disabled={disabled}>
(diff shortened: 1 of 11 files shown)


libs/service-portal/information/src/components/PersonalInformation/Forms/ProfileForm/components/FormButton.tsx
```diff
@@ -27,11 +27,9 @@ export const FormButton: FC<React.PropsWithChildren<Props>> = ({
         {children}
       </Button>
     ) : (
-      <button type="submit" disabled={disabled}>
-        <Button variant="text" size="small" disabled={disabled}>
-          {children}
-        </Button>
-      </button>
+      <Button type="submit" variant="text" size="small" disabled={disabled}>
```
thordurhhh, line 30: I remember this code!
The text variant will render the button as span I believe. So you probably need to specifically set the render as button.
```suggestion
      <Button type="submit" as="button" variant="text" size="small" disabled={disabled}>
```

thordurhhh:
I think we need to add `unfocusable={true}` to all of the buttons rendered as span to drop the double tabbing.

osk (approved):

PR #12779, opened 2023-11-16
fix(service-portal): avoid double tabbing on button links

Valur:
## What

Button component always has tabindex=0 for some reason, need to take  that off when rendered as span inside link

## Why

To avoid doubple tabbing inside links

## Checklist:


thordurhhh (approved):
niice

PR #12780, opened 2023-11-16
fix(service-portal): Document rsk link handler

thordurhhh:
## What

[This old edge case](https://github.com/island-is/island.is/blob/main/libs/api/domains/documents/src/lib/documentBuilder.ts#L20) where we need to specifically add the url for niðurstaða álagningar should be in the front end as well.

## Why

After the changes in Postbox, the detail document case is not handled, so we need both. I have been in contact with the service providers and they asked us to put this in for now. Hopefully not for too long.

## Checklist:


thorkellmani (approved):
LGTM

PR #12810, opened 2023-11-20
feat(service-portal): Add simple path masking

thordurhhh:
## What

Add simple path masking. Not really to be used for top security, but fine for low level.

[Feature deployment available](https://service-portalnat-id-path-masking-beta.dev01.devland.is/minarsidur) ✨ 

## Why

So we won't have any national Id's in paths. A simple base64 string that's not reversible.

## Screenshots / Gifs

Screenshot 2023-11-21 at 14 24 27

This should not slow anything down. Here's a screenshot of the performance test

Once:
Screenshot 2023-11-21 at 14 23 04

1000 iterations:
Screenshot 2023-11-21 at 14 20 55

## Checklist:


libs/shared/utils/src/lib/simpleEncryption.ts
```diff
@@ -0,0 +1,44 @@
...
+
+const ALGORITHM = 'aes-256-cbc'
+
+/**
+ *
+ * @param text The string you wish to hide
+ * @param key You secret key
+ * @returns URL safe Base64
+ */
+export const encrypt = (text: string, key: string): string => {
```
osk, line 12: Maybe we should call this something other than `encrypt`? `maskString`, `maskData`?

libs/shared/utils/src/lib/simpleEncryption.ts
```diff
@@ -0,0 +1,44 @@
...
+  const derivedKey = hashKey(key)
+
+  const cipher = createCipheriv(ALGORITHM, derivedKey, Buffer.alloc(16))
+  const encrypted =
+    cipher.update(text, 'utf-8', 'base64') + cipher.final('base64')
+
+  return Base64.encodeURI(encrypted)
```
osk, line 19: Could this throw? Should we handle here and return `string | null`?

libs/shared/utils/src/lib/simpleEncryption.ts
```diff
@@ -0,0 +1,44 @@
...
+    const encryptedData = Base64.decode(encryptedText)
+    const derivedKey = hashKey(key)
+    const decipher = createDecipheriv(ALGORITHM, derivedKey, Buffer.alloc(16))
+
+    return (
+      decipher.update(encryptedData, 'base64', 'utf-8') +
+      decipher.final('utf-8')
+    )
+  } catch (e) {
+    return encryptedText
```
osk, line 38: Log error?
Also returning the `encryptedText` could cause weird behaviour, I'd rather return `null`

libs/service-portal/information/src/screens/Child/Child.tsx
```diff
@@ -29,21 +29,22 @@ import { spmm } from '../../lib/messages'
...
   const { formatMessage } = useLocale()
   const userInfo = useUserInfo()
-  const { nationalId } = useParams() as UseParams
+  const { baseId } = useParams() as UseParams
 
   const { data, loading, error } = useNationalRegistryChildCustodyQuery({
     variables: {
       api: 'v3',
-      childNationalId: nationalId,
+      childNationalId: decrypt(baseId, userInfo.profile.nationalId),
```
osk, line 47: Need to handle if `decrypt` fails

libs/service-portal/information/src/screens/Child/Child.tsx
```diff
@@ -72,7 +74,7 @@ const Child = () => {
   )
 
   const isChildOrChildOfUser = isChild || isChildOfUser
-  if (!nationalId || (!loading && !isChildOrChildOfUser && !error))
+  if (!baseId || (!loading && !isChildOrChildOfUser && !error))
```
osk, line 77: This makes my head spin a bit (but that is for another time to fix)

libs/service-portal/information/src/components/FamilyMemberCard/FamilyMemberCard.tsx
```diff
@@ -8,6 +8,7 @@ import { InformationPaths } from '../../lib/paths'
 interface Props {
   title: string
   nationalId: string
+  baseId?: string
```
osk, line 11: What does it mean if `baseId` is `undefined`?

thordurhhh: Not all every familymember has a baseId. Like for example the logged in user and spouse.

thorkellmani (approved):
LGTM 🔎

PR #12846, opened 2023-11-22
chore(estate): only validate nationalid for stocks if non empty

HelgiHH:
# ...

[Attach a link to issue if relevant](https://app.asana.com/0/1204431115744096/1205939903295167)

## What

user should be able to move forward without entering national Id for stocks

## Checklist:


libs/application/templates/estate/src/lib/dataSchema.ts
```diff
@@ -274,15 +274,16 @@ export const estateSchema = z.object({
   stocks: z
     .object({
       organization: z.string(),
-      nationalId: z.string(),
+      nationalId: z.string().optional(),
```
splicebison, line 277: Won't  this produce an equivalent type `string | undefined`? If so, the below code would fail on undefined as  `kennitala.isValid()` accepts only `string | number`.

splicebison: Oh, ignore that. I see in the code below the additional check for undefined `nationalId && kennitala.isValid(nationalId)`

splicebison (approved):

PR #12848, opened 2023-11-23
fix(web): Use <a> tags for links that are pointing to a project path (like /minarsidur) (#12747)

RunarVestmann:
# Use `<a>` tags for links that are pointing to a project path (like /minarsidur) (#12747)


libs/shared/constants/src/lib/projectBasePath.ts
```diff
@@ -0,0 +1,7 @@
+export enum ProjectBasePath {
+  ServicePortal = '/minarsidur',
+  ConsultationPortal = '/samradsgatt',
+  ApplicationSystem = '/umsoknir',
+  AdminPortal = '/stjornbord',
+  CarRecycling = '/app/skilavottord',
+}
```
baering, line 7: Should these replace hardcoded paths for the projects?

Example:
image
image

baering (approved):
Codeowner files LGTM

libs/shared/constants/src/lib/projectBasePath.ts
RunarVestmann: Honestly yes, these paths should be used where possible but since this is a hotfix PR I'll make another PR for main to do exactly that

thorkellmani (approved):
LGTM

Toti91 (approved):

PR #12861, opened 2023-11-23
chore(j-s): Strip identifiable strings from error logs

oddsson:
# Strip identifiable strings from error logs

[Asana](https://app.asana.com/0/1199153462262248/1205990120996971/f)

## What

Strip identifiable strings from error logs

## Why

A third party service we use is throwing errors with identifiable strings, like defendant's name and national id. We need to mask these strings before displaying them in our logs.

## Checklist:


apps/judicial-system/backend/src/app/modules/defendant/defendant.service.ts
```diff
@@ -305,7 +305,17 @@ export class DefendantService {
         return { delivered: true }
       })
       .catch((reason) => {
-        this.logger.error('Failed to update case with defendant', { reason })
+        const sanitizedReason = JSON.stringify(reason)
+          .replace(
+            /Participant with id: \d{10}/g,
+            'Participant with id: **********',
+          )
```
gudjong, line 312: Stripping national ids from error logs is not necessary as it is already being done elsewhere.

apps/judicial-system/backend/src/app/modules/defendant/defendant.service.ts
```diff
@@ -305,7 +305,17 @@ export class DefendantService {
...
+          .replace(/\) gegn(.*?)'/g, ') gegn **********')
+
+        this.logger.error(
+          'Failed to update case with defendant',
+          JSON.parse(sanitizedReason),
+        )
```
gudjong, line 318: The error message has already been sanitized in the court service, so there is not need to do it here also.

gudjong (approved):
Please consider removing sanitation from the defendant service as the error has already been sanitised in the court service.

PR #12863, opened 2023-11-23
chore(authmodule): removed mentions of audience from AuthModule Readme

jonarnarbriem:
# ...

## What

Removing mentions of Audience from readme and docs

## Why

We're no longer using audience for API authorization

## Screenshots / Gifs

## Checklist:


libs/auth-nest-tools/README.md
```diff
@@ -20,12 +20,11 @@ Import and configure the AuthModule, example:
...
   imports: [
     AuthModule.register({
-      audience: 'protected_resource',
-      issuer: 'https://localhost:6001',
+      issuer: 'https://localhost:6001'
     }),
 ```
 
-where `audience` is the name your resource was registered under in IdS and `issuer` the IdS url.
+where `issuer` is the IdS url.
```
saevarma, line 27: We should also add a section that describes that the `ScopesGuard` is now required to verify at least one scope, so the API is verifying that the token is intended for this service. Otherwise it would be authorising all valid IDS tokens, intended other services.

libs/auth-nest-tools/README.md
```diff
@@ -10,7 +10,7 @@ There are a couple of guards available.
 - `IdsAuthGuard`: same as `IdsUserGuard` but does not verify the `nationalId` claim. Information from the JWT can be accessed using the CurrentAuth parameter decorator.
 - `ScopesGuard`: checks if the access token has required scopes. These can be configured using the Scopes decorator.
 
-You should generally add `IdsUserGuard` and `ScopesGuard` to endpoints that return user resources for the authenticated user. You can use `IdsAuthGuard` for endpoints that need to be available for clients authenticating with client credentials.
+You should generally add `IdsUserGuard` to endpoints that return user resources for the authenticated user and `IdsAuthGuard` for endpoints that need to be available for clients authenticating with client credentials. You should always use `ScopesGuard` to protect endpoints that contain sensitive information, without `ScopesGuard` the endpoint will be available to all authenticated users.
```
saevarma, line 13:
```suggestion
You should generally add `IdsUserGuard` to endpoints that return user resources for the authenticated user and `IdsAuthGuard` for endpoints that need to be available for clients authenticating with client credentials. You should always use `ScopesGuard` and the `@Scopes` decorator to protect endpoints, without it the API would authorise all valid access tokens issued by IAS.
```

saevarma (approved):
One little change otherwise looks good

PR #12869, opened 2023-11-24
Fix formatting of phone number in user profile settings

snaerseljan:
# Fix formatting of phone number in user profile settings read only field

From this

To this
image

## Checklist:


thordurhhh (approved):
Nice 💪

libs/shared/utils/src/lib/formatPhoneNumber.ts
saevarma, line 1: Looks like this is not used and can be removed.

saevarma (approved):
Nice job!

libs/shared/utils/src/lib/formatPhoneNumber.ts
snaerseljan: Ahhh nice spot

Toti91 (approved):
Nice

PR #12881, opened 2023-11-27
fix(service-portal): vehicle graphql layer needs to continue support for app 

Tryggvig:
# Graphql revert to allow app to use old service

## What

The app continues to use the old service for vehicles for a couple of months, we need to make sure it can still use it

## Why

The app needs to work

## Checklist:


thordurhhh (approved):
😍

PR #12889, opened 2023-11-27
chore(search-indexer): Increase memory limit

saevarma:
## What

Search indexer is running out of memory in dev (which is using Contentful's preview API). This causes content not to be synced to Elastic and hence not able to test new components in dev.

Current memory increase: 
- request: 3gb
- limit: 4gb

## Why

To unblock syncing from Contentful to Elasticsearch.

**Note** This has been done couple of times due to increasing content in Contentful. We need to refactor how search indexer is requesting content from Contentful as current implementation is not scaling with growing content. But we don't want to do that in a rush, hence increasing memory to unblock testing while we properly refactor the indexer.

## Screenshots / Gifs

n/a

## Checklist:


brynjarorng (approved):
LGTM

baering (approved):

PR #12903, opened 2023-11-28
fix(service-portal): Mileage title loading

thordurhhh:
## What

Set loading for the title while fetching first info

## Why

The state of the title can change. Default is loading, then the state is shown once we know it from the API.

## Checklist:


thorkellmani (approved):
LGTM

PR #12915, opened 2023-11-29
fix(service-portal): hotfix - Update default provider for intro header

thordurhhh:
https://github.com/island-is/island.is/pull/12900


thorkellmani (approved):
LGTM

PR #12923, opened 2023-11-29
feat(admin-portal): Signature Collection

albinagu:
Feature branch: 
https://adminportalsignaturecollection-beta.dev01.devland.is/stjornbord/medmaelasofnun

Stjórnborð fyrir meðmælasöfnun.
- Setting up a new module in Admin Portal for ÞÍ to review lists during signature collections / elections.

## Checklist:


libs/portals/admin/signature-collection/assets/img.jpg
AndesKrrrrrrrrrrr, line 1: SVG?

AndesKrrrrrrrrrrr (approved):
CODOWNERS LGTM

thorkellmani (approved):
LGTM

libs/portals/core/src/components/IntroHeader.tsx
```diff
@@ -49,8 +53,11 @@ export const IntroHeader = ({
       </GridColumn>
       {img && (
         <GridColumn span={['8/8', '3/8']}>
-          <Hidden print={hideImgPrint} below="lg">
-            <Box textAlign="center" padding={[6, 0]}>
+          <Hidden print={hideImgPrint} below={imgHiddenBelow}>
+            <Box
+              textAlign={imgPosition === 'right' ? 'right' : 'center'}
```
thorkellmani, line 58: The conditional seems unnecessary, is there a point to the check? Why not use the `imgPosition` argument directly?

thorkellmani (commented):
One minor comment that i forgot in the earlier approval

libs/portals/core/src/components/IntroHeader.tsx
albinagu: oh, right! :D thanks!

kksteini (approved):

libs/portals/admin/signature-collection/src/screens/List/index.tsx
```diff
@@ -0,0 +1,106 @@
...
+          <PortalNavigation
+            navigation={signatureCollectionNavigation}
+            title={formatMessage(m.signatureListsTitle)}
+          />
+        </GridColumn>
+        <GridColumn
+          paddingTop={[5, 5, 5, 2]}
+          offset={['0', '0', '0', '1/12']}
+          span={['12/12', '12/12', '12/12', '8/12']}
+        >
```
saevarma, line 48: Looks like same layout as in `AllLists` and `List` so you could use[ similar pattern](https://github.com/island-is/island.is/blob/main/libs/portals/admin/ids-admin/src/module.tsx#L75) to what we did in IDS Admin for tenant. We created a empty parent route which is the shared layout and uses the `<Outlet>` component from react router. Then nested routes are rendered in the outlet.

libs/portals/admin/signature-collection/src/module.tsx
```diff
@@ -0,0 +1,33 @@
...
+import { m } from './lib/messages'
+import { SignatureCollectionPaths } from './lib/paths'
+import { listsLoader } from './screens/AllLists/AllLists.loader'
+import { AdminPortalScope } from '@island.is/auth/scopes'
+import { listLoader } from './screens/List/List.loader'
+
+const AllLists = lazy(() => import('./screens/AllLists'))
+const List = lazy(() => import('./screens/List'))
+
+const allowedScopes: string[] = [AdminPortalScope.petitionsAdmin]
```
saevarma, line 12: As this is a standalone admin module we should use a new separate scope for access control. And looks like you still have to look into scopes on the gql domain.

libs/portals/admin/signature-collection/src/lib/utils.ts
```diff
@@ -0,0 +1,19 @@
...
+  { value: 'Austfirðingafjórðungur', label: 'Austfirðingafjórðungur' },
+]
+
+export type Filters = {
+  area: Array<string>
+  input: string
+}
+
+export const resultsForComparison = {
+  nationalId: '010130-2989',
```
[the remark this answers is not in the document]
albinagu: i know, using while wip

saevarma (approved):
LGTM

PR #12879, opened 2023-11-27
feat(car-recycling): new application

birkirkristmunds:
Car recycling application

## Why
for applicants to apply for car recycling at Recycling fund (Úrvinnslusjóður)

/umsoknir/skilavottord/

## Checklist:

- [x] I have performed a self-review of my own code
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] Formatting passes locally with my changes
- [x] I have rebased against main before asking for a review

PR #12932, opened 2023-11-29
fix(core): Fix schema-files cache busting

eirikurn:
## What

- Make codegen/backend-clients rebuild when `clientConfig.{yml,yaml,json} files change.
- Make codegen/backend-clients rebuild when `openapi.{yml,yaml,json}` files generated by dependent targets change.
- Make codegen/frontend-clients rebuild when api.graphql changes.

## Why

So that changes in those files cache-bust later targets.


scripts/check-tags.ts
```diff
@@ -11,6 +11,7 @@ interface Project {
 
 const checkTags = async (filePath: string) => {
   const projectText = await readFile(filePath, 'utf8')
+  console.log(filePath)
```
saevarma, line 14: console.log?

saevarma (approved):
🚀

nx.json
```diff
@@ -114,7 +114,7 @@
     },
     "codegen/backend-client": {
       "dependsOn": ["^codegen/backend-client", "^codegen/backend-schema"],
-      "inputs": ["schema-files", "^schema-files"]
+      "inputs": ["schema-files", "^schema-outputs"]
```
saevarma, line 117: I was reviewing Deloittes car recycling PR #12879 and there we are creating backend client for GQL api, not rest. Don't we need to add GQL files here as inputs for backend-client to resolve Nx cache issues that could happen?

eirikurn: Good point. Added 👍

brynjarorng (approved):
LGTM

PR #12933, opened 2023-11-29
feat(user-profile): Set DataStatus for email and phone on user profile update

GunnlaugurG:
## What

Update both emailStatus and phoneStatus to corresponding values when updating either email or phone number. 

## Why

So these values will be backwards compatible with the UserProfile V1.

## Checklist:


apps/services/user-profile/src/app/v2/dto/user-profile.dto.ts
```diff
@@ -53,4 +54,14 @@ export class UserProfileDto {
...
+
+  @ApiPropertyOptional()
+  @IsOptional()
+  @IsString()
+  readonly emailStatus?: DataStatus
+
+  @ApiPropertyOptional()
+  @IsOptional()
+  @IsString()
+  readonly mobileStatus?: DataStatus
```
saevarma, line 66: We should not put those fields in the API response.

saevarma (approved):

apps/services/user-profile/src/app/v2/dto/user-profile.dto.ts
GunnlaugurG: Nice catch, forgot to remove it :D

PR #12938, opened 2023-11-30
fix(vehicles): Add feature flag for vehicle mileage resolver (#12937)

thordurhhh:
hotfix: https://github.com/island-is/island.is/pull/12937


thorkellmani (approved):
lgtmaagj

PR #12950, opened 2023-12-01
fix(service-portal): LinkResolver Onclick to undefined

thordurhhh:
## What

Link resolver onClick should default to undefined so no click event is registered.

## Why

Now its registering an empty onclick.

## Checklist:


libs/service-portal/core/src/components/LinkButton/LinkButton.tsx
```diff
@@ -4,18 +4,28 @@ import { isExternalLink } from '../../utils/isExternalLink'
...
+      /**
+       * default variant is "text"
+       */
+      variant?: 'text'
+      icon?: never
+    }
+
+type LinkButtonProps = SharedProps & Props
+
+export const LinkButton: FC<React.PropsWithChildren<LinkButtonProps>> = ({
```
thorkellmani, line 28: Unnecessary `FC<React.PropsWithChildren>` since the children aren't used.

thorkellmani (approved):
One comment

PR #12988, opened 2023-12-05
fix(core): Fix codegen

saevarma:
## What

Changing `nx.json` `namedInputs.schema-outputs` to list files directly instead of using `dependentTasksFiles` and rather use the named input using `^` character instead to indicate files of dependent tasks.

## Why

This caused bug in Nx resolving invalidating cache making the script run for very long.

## Screenshots / Gifs

N/A

## Checklist:


snaerseljan (approved):
LGTM

PR #12989, opened 2023-12-05
test(j-s): Court custody flow test

unakb:
[Sjálfvirk regression test - taka á móti gæslu hjá dómstól og úrskurða.](https://app.asana.com/0/1199153462262248/1205848941408753)

## What

Add a test for the court receiving a case and submitting results

## Why

So we can continue our case flow e2e tests

## Checklist:


apps/system-e2e/src/support/session.ts
```diff
@@ -153,13 +153,16 @@ export async function session({
   return context
 }
 
-export async function judicialSystemSession({ browser }: { browser: Browser }) {
+export async function judicialSystemSession(
+  { browser }: { browser: Browser },
+  homeUrl?: string,
```
AndesKrrrrrrrrrrr, line 158:
```suggestion
  { browser, homeUrl }: { browser: Browser, homeUrl?: string },
```

apps/system-e2e/src/tests/judicial-system/smoke/custody-tests.spec.ts
```diff
@@ -0,0 +1,8 @@
+import { test } from '@playwright/test'
+import * as prosecutor from './custody-prosecutor'
+import * as court from './custody-court'
+
+test.describe.serial('Custody tests', () => {
+  prosecutor.addTests()
+  court.addTests()
+})
```
AndesKrrrrrrrrrrr, line 8: If the tests _need_ to be run after each other, maybe put them in the same file and section? If not, why test them serially?

AndesKrrrrrrrrrrr (approved):
Approved with minor comments. Fix at your own leisure

apps/system-e2e/src/tests/judicial-system/smoke/custody-tests.spec.ts
unakb: I did it like this because there are still around 4 more tests (with 2 new user roles) that need to be added to complete the chain, and I was worried a single file would become monstrous and really hard to work with 🥲

PR #12855, opened 2023-11-23
feat(j-s): Ask for help button

oddsson:
# Ask for help button

[Asana](https://app.asana.com/0/1199153462262248/1205724728618381/f)

## What

The "Ask for help" button is now on the login screen, but only for Icelandic users, to guard against bots.

## Why

Sometimes users need help when logging in.

## Screenshots / Gifs

Screenshot 2023-11-23 at 13 33 07

## Checklist:

- [x] I have performed a self-review of my own code
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] Formatting passes locally with my changes
- [x] I have rebased against main before asking for a review

PR #12995, opened 2023-12-05
fix(j-s): Use request-ip to get client ip address

oddsson:
# Use request-ip to get client ip address

[Asana](https://app.asana.com/0/1199153462262248/1205724728618381/f)

## What

In #12855, we implemented a feature where we detect what country a user is and only render a button to a feedback form if the user is in Iceland. This was done to prevent bots spamming our form (it's a simple form that is not able to detect or stop bots from spamming it).

To get the client ip in the previous PR we use `https://api.ipify.org` and that worked fine locally, but did not work on dev. This is probably because we were calling ipify on the client, and not the server.  

In this PR we use [request-ip](https://www.npmjs.com/package/request-ip) instead of ipify to do the same thing. I think this way is a bit cleaner, although we are adding a dependency, I think it's ultimately worth it.

## Checklist:


unakb (approved):

apps/judicial-system/web/pages/api/geoLocation/getCountryCode.ts
saevarma, line 1: Did you look into getting the IP address from request header?

oddsson: Yea, that's pretty much what `request-ip` does, it looks for the ip address in the request header. But apparently the IP can be set in various places and so I opted for using request-ip..

Screenshot 2023-12-06 at 13 42 05

Do you know how to get the ip from the request header, in as reliable way, without using request-ip? I'm all for droping this dependency, if you can help me get it another way :)

PR #13001, opened 2023-12-06
feat(vehicles): Update vehicle auth in vehicle API

thordurhhh:
## What

Update vehicle auth in vehicle API

## Why

Add improved auth for mileage api. By default it will allow anyone to get. Which is fine, but in our case we just want the logged in user to do it.

## Checklist:


libs/api/domains/vehicles/src/lib/api-domains-vehicles.service.ts
```diff
@@ -162,6 +190,11 @@ export class VehiclesService {
       return null
     }
 
+    const hasAuth = this.hasVehicleServiceAuth(auth, input.permno)
+    if (!hasAuth) {
+      return null
+    }
```
thorkellmani, line 196: This feels iffy. Returning null here can either mean no cars found (i think) or the user is unauthorized which are are 2 very different things.

libs/api/domains/vehicles/src/lib/api-domains-vehicles.service.ts
```diff
@@ -181,6 +214,11 @@ export class VehiclesService {
   ): Promise<PostMileageReadingModel | null> {
     if (!input) return null
 
+    const hasAuth = this.hasVehicleServiceAuth(auth, input.permno)
+    if (!hasAuth) {
+      return null
+    }
```
thorkellmani, line 220: Same as above

libs/api/domains/vehicles/src/lib/api-domains-vehicles.service.ts
```diff
@@ -194,6 +232,11 @@ export class VehiclesService {
   ): Promise<PutMileageReadingModel | null> {
     if (!input) return null
 
+    const hasAuth = this.hasVehicleServiceAuth(auth, input.permno)
+    if (!hasAuth) {
+      return null
+    }
+
```
thorkellmani, line 239: Same as above

thorkellmani (changes requested):
Few comments

libs/api/domains/vehicles/src/lib/api-domains-vehicles.service.ts
thordurhhh: Yeah this should probably not return anything at all. A call to the auth service would be enough because it will throw an error on no auth anyway.

libs/api/domains/vehicles/src/lib/api-domains-vehicles.service.ts
```diff
@@ -136,6 +144,26 @@ export class VehiclesService {
...
+      const res = await this.getVehicleDetail(auth, {
+        clientPersidno: auth.nationalId,
+        permno,
+      })
+      if (!res) return false
+      return !!res
+    } catch (e) {
+      this.logger.error(UNAUTHORIZED_LOG, {
+        category: LOG_CATEGORY,
+        ...e,
```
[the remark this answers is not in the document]
thordurhhh: but yeah, changed now.

PR #13013, opened 2023-12-06
feat(user-notification): Send email in notification worker

snaerseljan:
# Feature to add email notification to notification worker

## Before merging

- [x] ✅  Waiting on [This PR](https://github.com/island-is/island.is/pull/13073) to be merged to add additional check to send email

## What

We are updating our user profile to support property called `emailNotification`. When the user has this flagged enabled want to send him an email at the same time as he is supposed to receive push notification.
This PR adds sending email to the user, notifying him that he has received an email in is mailbox on Ísland.is.

 ### Summary of what I did.

- Added sending email support to our notification worker.
- Refactored a little bit.
- Added support for attributes in the featureFlagService.

### Example of Request to API endpoint

```
POST http://localhost:3333/v1/notifications

PAYLOAD: 
{
    "recipient": "0101302479",
    "args": [
        {
            "key": "organization",
            "value": "Ísland.is"
        },
    ],
    "templateId": "HNIPP.POSTHOLF.NEW_DOCUMENT"
}
```

## Why

So the user will receive email that he has new mail in his inbox on island.is

## Screenshots / Gifs

Icelandic email
image

English email
image

## Checklist:


thordurhhh (approved):
Nice ✅

rafnarnason (approved):
lgtm

unakb (approved):

apps/api/src/app/app.module.ts
```diff
@@ -333,7 +333,6 @@ const environment = getConfig
         NationalRegistryClientConfig,
         SyslumennClientConfig,
         ElectronicRegistrationsClientConfig,
-        FeatureFlagConfig,
```
saevarma, line 336: We should just leave the gql api alone. And Gulli is also going to add feature flag to the user profile domain, so if it was not using this config it will be used soon.

apps/services/user-notification/src/app/modules/notifications/notifications.service.ts
```diff
@@ -32,30 +48,29 @@ export class NotificationsService {
...
-    return 'is-IS'
+  async mapLocale(locale?: string | null): Promise<string> {
+    return locale === 'en' ? locale : 'is-IS'
   }
+
   async getTemplates(
     locale?: string | null | undefined,
   ): Promise<HnippTemplate[]> {
-    locale = await this.mapLocale(locale)
+    const mappedLocale = await this.mapLocale(locale)
```
saevarma, line 58: Why renaming instead of just overwriting as it was?

apps/services/user-notification/src/app/modules/notifications/notifications.service.ts
```diff
@@ -108,47 +125,81 @@ export class NotificationsService {
...
-        'notificationBody',
-        'notificationDataCopy',
-        'clickAction',
-      ]
-      // find {{arg.key}} in string and replace with arg.value
-      const regex = new RegExp(/{{[^{}]*}}/)
+  /**
+   * Checks if the arguments provided in the request body are valid for the template
+   */
+  validateArgs(args: ArgumentDto[], template: HnippTemplate) {
```
saevarma, line 161: I would turn this check around and check that at least all the args required by the template are provided. We should just ignore additional args.

apps/services/user-notification/src/app/modules/notifications/notificationsWorker.service.ts
```diff
@@ -29,6 +51,185 @@ export class NotificationsWorkerService implements OnApplicationBootstrap {
...
+        name: 'Ísland.is',
+        address: 'no-reply@island.is',
+      },
+      to: {
+        name: fullName,
+        address: profile.email,
+      },
+      subject: template.notificationTitle,
+      template: {
+        title: 'Testinga',
```
saevarma, line 109: Find appropriate title?

apps/services/user-notification/src/app/modules/notifications/notificationsWorker.service.ts
```diff
@@ -29,6 +51,185 @@ export class NotificationsWorkerService implements OnApplicationBootstrap {
...
+      return
+    }
+
+    const [template, individual] = await Promise.all([
+      this.notificationsService.getTemplate(message.templateId, profile.locale),
+      this.nationalRegistryService.getName(profile.nationalId),
+    ])
+
+    const fullName = individual?.fulltNafn ?? ''
+    const isEnglish = profile.locale === 'en'
```
saevarma, line 205: We have a `Locale` enum in shared to use instead of hardcoding strings.

libs/clients/national-registry/v3/src/lib/apiConfig.ts
```diff
@@ -1,16 +1,11 @@
...
-  XRoadConfig,
-} from '@island.is/nest/config'
+import { ConfigType, XRoadConfig } from '@island.is/nest/config'
 
 import { Configuration } from '../../gen/fetch'
 import { NationalRegistryV3ClientConfig } from './nationalRegistryV3.config'
 
 export const ApiConfig = {
   provide: 'NationalRegistryV3ClientProviderConfiguration',
-  scope: LazyDuringDevScope,
```
saevarma, line 13: We need this so it can run codegen in CI without environment variables present.

apps/services/user-notification/src/app/modules/notifications/notificationsWorker.service.ts
```diff
@@ -29,6 +51,185 @@ export class NotificationsWorkerService implements OnApplicationBootstrap {
...
+        address: 'no-reply@island.is',
+      },
+      to: {
+        name: fullName,
+        address: profile.email,
+      },
+      subject: template.notificationTitle,
+      template: {
+        title: 'Testinga',
+        body: [
```
saevarma, line 110: We need to sync offline about the email body on monday as we need to support all the active push notifications.

apps/services/user-notification/src/app/modules/notifications/dto/hnippTemplate.response.ts
```diff
@@ -3,16 +3,25 @@ import { ApiProperty } from '@nestjs/swagger'
...
   notificationBody!: string
+
   @ApiProperty({ example: 'Some extra text ...' })
   notificationDataCopy?: string
+
   @ApiProperty({ example: '//inbox/{{documentId}}' })
   clickAction?: string
+
+  @ApiProperty({ example: 'https://island.is/minarsidur/postholf' })
+  clickActionWeb?: string
```
saevarma, line 20: @osk was going to map out actions in a notion table so we can sync this for mobile and web. We should sync about this offline on monday.

saevarma (commented):
Great job, clean and tidy as always.

Minor things that needs to fixed and then the two things that we need to sync with the other notifications teams before we merge this.

apps/api/src/app/app.module.ts
snaerseljan: This is being used already and it was double imported. It cant hurt to fix this :)
image

apps/services/user-notification/src/app/modules/notifications/notifications.service.ts
snaerseljan: This is purely matter of style. I like not to override input parameters especially if they are to be mutated. It can lead to wrong code. In this case it does not matter but more of a principle with me

apps/services/user-notification/src/app/modules/notifications/notifications.service.ts
snaerseljan: Hmm don´t know if I understand you. It is checking if the arg length and temp length are the same before
image

And also this is just old code moved to the service and refactored a little bit

apps/services/user-notification/src/app/modules/notifications/notificationsWorker.service.ts
snaerseljan: 🤦

apps/services/user-notification/src/app/modules/notifications/notificationsWorker.service.ts
snaerseljan: It is not in shared,it is only an union type and an array. Leaving as is because I would have to create and Locales enum in shared and TS would complain about the enums not being the same since profile.locale is of UserProfileLocaleEnum, which is generated btw. If I would import from `/api/domains/user-profile/types/locale.enum` it would still complain since the type is generated.

So leaving as is!

libs/clients/national-registry/v3/src/lib/apiConfig.ts
snaerseljan: Oh snap. Then we have to look into this. The notficationWorker.service does not run when this is defined, at least not in development. When Scope.REQUEST is present it wont run

andes-it:

saevarma (approved):
Great joint effort!

PR #13033, opened 2023-12-07
fix(web): Add optional locale prop to FaqList component

RunarVestmann:
# Add optional locale prop to FaqList component

## Why

* So that EmbeddedVideo disclaimer has correct translations

### Before (this should be in english)
image

### After
image

## Checklist:


Toti91 (approved):
nice

mannipje (approved):

PR #13049, opened 2023-12-11
fix(island-ui): DatePicker with time

albinagu:
Within the Admin portal, we now have a case where user needs to be able to edit time. Minor changes to Datepicker adding a couple of props and styling for time input:

Screenshot 2023-12-11 at 10 34 02

## Checklist:

### DIFF
--- a/libs/island-ui/core/src/lib/DatePicker/DatePicker.tsx
+++ b/libs/island-ui/core/src/lib/DatePicker/DatePicker.tsx
@@ -9,7 +9,7 @@ import {
 import getYear from 'date-fns/getYear'
 import is from 'date-fns/locale/is'
 import en from 'date-fns/locale/en-US'
-import { dateFormat } from '@island.is/shared/constants'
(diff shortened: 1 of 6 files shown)


kksteini (approved):
Codeowner files LGTM

stjanilofts (approved):

stjanilofts (approved):

libs/island-ui/core/src/lib/DatePicker/DatePicker.tsx
```diff
@@ -143,6 +151,9 @@ export const DatePicker: React.FC<React.PropsWithChildren<DatePickerProps>> = ({
               size={size}
             />
           }
+          timeFormat="HH:mm"
```
saevarma, line 154: Are we using the same time format for IS and EN ? Shouldn't this be the `timeFormat` from the shared lib `date.ts` ?

saevarma (commented):
Has this changed been added to the design system in Figma?

PR #13053, opened 2023-12-11
feat(application-estate): Settlement Entry Requirements

kksteini:
* Rejection messages added based on validation
* Client Config updated
* Client Config generation updated

image

Note: District Commissioner's have stopped updating the `/dev` path and the newer client configs for development shall be retrieved from `/dev`
## Checklist:


RunarVestmann (approved):

libs/application/templates/estate/src/forms/OverviewSections/assetsAndDebts.ts
```diff
@@ -51,11 +51,7 @@ export const overviewAssetsAndDebts = [
               (asset.marketValue ? formatCurrency(asset.marketValue) : '0 kr.'),
             m.propertyShare.defaultMessage +
               ': ' +
-              (asset.share
-                ? asset.share > 1
-                  ? asset.share + '%'
-                  : asset.share * 100 + '%'
-                : '0%'),
+              (asset.share ? asset.share + '%' : ''),
```
HelgiHH, line 54: some of these might cause conflicts with https://github.com/island-is/island.is/pull/13055

libs/application/template-api-modules/src/lib/modules/templates/estate/estate.service.ts
```diff
@@ -69,6 +69,36 @@ export class EstateTemplateService extends BaseTemplateApiService {
       )
     }
 
+    const { availableSettlements } = applicationData
+    console.log('Application answers', JSON.stringify(applicationAnswers))
```
HelgiHH, line 73: this console is intentional right?

HelgiHH (approved):
lgtm, might want to check on 2 comments/suggestions

libs/application/template-api-modules/src/lib/modules/templates/estate/estate.service.ts
kksteini: Haha nope

kksteini: Fixed, thank you

libs/application/templates/estate/src/forms/OverviewSections/assetsAndDebts.ts
kksteini: The mappers have been rewritten to give the original value in percentages.
Is there something I'm missing here?

The value of share will now always be as a percentage, no [0,1] ratios

obmagnusson (approved):

PR #13099, opened 2023-12-13
fix(signature-collection): Admin - add filter by candidate

albinagu:
## Checklist:

### DIFF
--- a/libs/portals/admin/signature-collection/src/screens/AllLists/index.tsx
+++ b/libs/portals/admin/signature-collection/src/screens/AllLists/index.tsx
@@ -34,36 +34,53 @@ const Lists = () => {
   const [page, setPage] = useState(1)
   const [filters, setFilters] = useState<Filters>({
     area: [],
+    candidate: [],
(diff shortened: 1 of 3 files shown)


libs/portals/admin/signature-collection/src/screens/AllLists/index.tsx
```diff
@@ -34,37 +34,54 @@ const Lists = () => {
...
-            list.owner.name
-              .toLowerCase()
-              .includes(filters.input.toLowerCase()) ||
-            list.area.name.toLowerCase().includes(filters.input.toLowerCase())
-          )
-        },
+    filteredList = filteredList.filter((list) => {
+      return (
+        // Filter by area
+        (filters.area.length === 0 || filters.area.includes(list.area.name)) &&
```
HelgiHH, line 53: Might want to consider creating constant for these like const FilterArea = filters.area etc...

HelgiHH (approved):
lgtm

PR #13107, opened 2023-12-14
fix(j-s): E2E, Verify update requests completion before continuing

unakb:
## What

Waiting for update responses before continuing in our court e2e test

## Why

If there's a delay in response time then some steps prematurely fire and create a race condition, often breaking the test when run on dev

## Checklist:

### DIFF
--- a/apps/system-e2e/src/support/api-tools.ts
+++ b/apps/system-e2e/src/support/api-tools.ts
@@ -38,3 +38,20 @@ export async function mockApi(page: Page, url: string, response: any) {
     })
   })
 }
+
+export async function verifyRequestCompletion(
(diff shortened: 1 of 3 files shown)


apps/system-e2e/src/support/api-tools.ts
```diff
@@ -38,3 +38,20 @@ export async function mockApi(page: Page, url: string, response: any) {
     })
   })
 }
+
```
unakb, line 41: The code above doesn't seem to be in use, but I didn't want to remove it without asking first since it's not our code.

AndesKrrrrrrrrrrr (approved):

PR #13108, opened 2023-12-14
chore(user-profile): Migration worker

baering:
## What

- Create a table to import user profiles from Advania
- Update existing phone number format to match
- Migrate advania user profiles into user profiles following a set of rules

Migration rules:
- does the migrated profile exist in `user_profile`?
  - no
    - create a new profile in `user_profile` with the new data and set `last_nudge` to `null`*
  - yes
    - does the email and mobile phone number match?
      - yes
        - keep `user_profile` data as is, set `last_nudge` = `nudge_last_asked` from migrated profile
          - this will be as if the user used the new onboarding flow and won't need to go through it again
      - no
        - was the exported profile exported more recently than the profile in `user_profile`
          - yes
            - override `user_profile` with migrated profile, set `last_nudge` = `null`*
          - no
            - is the email or mobile phone number in `user_profile` verified?
              - yes
                - set `user_profile.last_nudge` = `user_profile.modified`
              - no
                - set `user_profile.last_nudge` = `null`*
                
_* setting `last_nudge` = `null` will trigger the onboarding flow for the next login, with prefilled email/phone if available_

This logic is reflected in unit tests:
image

## Why

To be able to deprecate the old user profile system and start collecting user profiles into the new one.

## Checklist:


apps/services/user-profile/infra/service-portal-api.ts
```diff
@@ -23,6 +23,11 @@ export const serviceSetup = (): ServiceBuilder<'service-portal-api'> =>
         staging: 'false',
         prod: 'false',
       },
+      USER_PROFILE_WORKER_PAGE_SIZE: {
+        dev: '3000',
+        staging: '3000',
+        prod: '3000',
+      },
```
baering, line 30: Should I just leave this out?

apps/services/user-profile/src/app/worker/test/worker.spec.ts
```diff
@@ -0,0 +1,569 @@
...
+        exported: new Date(2023, 10, 1),
+        nudgeLastAsked: new Date(2023, 10, 5),
+      }))
+
+      const existingProfiles = advaniaProfiles.map((p) => ({
+        nationalId: p.ssn,
+        email: p.email,
+        mobilePhoneNumber: p.mobilePhoneNumber,
+        modified: new Date(2023, 0, 1),
+        lastNudge: new Date(), // not null
```
saevarma, line 98: ~ Existing profiles in our DB would have lastNudge as null as v1 is not storing it. But pretty sure this doesn't matter for this test case.

apps/services/user-profile/src/app/worker/test/worker.spec.ts
```diff
@@ -0,0 +1,569 @@
...
+      const fieldsToCompare = ['nationalId', 'email', 'mobilePhoneNumber']
+
+      expect(userProfilesBeforeMigration.length).toBe(
+        userProfilesAfterMigration.length,
+      )
+
+      for (let i = 0; i < userProfilesBeforeMigration.length; i += 1) {
+        const profileBefore = userProfilesBeforeMigration[i]
+        const profileAfter = userProfilesAfterMigration[i]
+        const advaniaProfile = advaniaProfiles[i]
```
saevarma, line 121: This depends on the findAll method to return the same order, wondering if that is guaranteed or if this could be flaky and we should set explicit order in findAll? And it needs to be in the same order as the `advaniaProfiles` array.

apps/services/user-profile/src/app/worker/test/worker.spec.ts
saevarma, line 1: 👏

apps/services/user-profile/src/app/worker/userProfileAdvania.model.ts
```diff
@@ -0,0 +1,84 @@
...
+  timestamps: true,
+  indexes: [
+    {
+      fields: ['ssn'],
+    },
+  ],
+})
+export class UserProfileAdvania extends Model {
+  @CreatedAt
+  @ApiProperty()
```
saevarma, line 24: ~ `@ApiProperty` is not needed for this model as we are not returning it in a rest endpoint. But it doesn't harm anything so up to you to remove it or leave it.

apps/services/user-profile/infra/service-portal-api.ts
saevarma: No I like the idea of processing this in pages.

saevarma (approved):
10/10 🚀

peturgq (approved):

thordurhhh (approved):
Nice one 🙌 
LGTM

apps/services/user-profile/migrations/20231105100001-format-phone-numbers.js
```diff
@@ -0,0 +1,20 @@
...
+    BEGIN;
+      UPDATE user_profile
+      SET mobile_phone_number = CONCAT('+354-', mobile_phone_number)
+      WHERE mobile_phone_number NOT LIKE '+354-%' AND mobile_phone_number IS NOT NULL;
+    COMMIT;
+  `)
+  },
+
+  down: () => {
+    // We could create a column `mobile_phone_number_backup` to store the
```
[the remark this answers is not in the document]
kksteini: Should it then be removed?

kksteini (approved):
CODEOWNER files LGTM

baering:
Will address comments in another PR to get this into main. Thanks for your reviews.

PR #13109, opened 2023-12-14
chore(j-s): Disable multiple click on the same case

oddsson:
# Disable multiple click on the same case

[Asana](https://app.asana.com/0/1199153462262248/1205924191397151/f)

## What

When a user clicks a case in the case list, a request is sent to the server to get the case and open it. We were not disabling the line the user clicked and that means that if the request is slow and the user clicked the same case again, another request is sent to the server. This can lead to performance issues. This PR implements a new hook, `useCaseList`, that can be used to detect what case was clicked and whether or not we should show a loading indicator on the case that is being opened. The loading indicator should be visible two seconds after the first click. Once a case is clicked, it cannot be clicked again.

## Screenshots / Gifs

https://github.com/island-is/island.is/assets/3789875/c18499ea-c733-49ea-8867-883f80a9178b

## Checklist:


unakb (approved):
Looks good

PR #13127, opened 2023-12-14
chore(j-s): Near Operation 1

gudjong:
# Near Operation

[Near operation fyrir users, institutions, court upload](https://app.asana.com/0/1199153462262248/1206173927604693/f)

## What

- Uses near operation GraphQL generation.

## Why

- Recommended

## Checklist:


apps/judicial-system/web/src/utils/mutations.ts
```diff
@@ -87,7 +87,7 @@ export const CreateUserMutation = gql`
 `
 
 export const UsersQuery = gql`
-  query Users {
+  query SomeUsers {
```
oddsson, line 90: Is this the correct query name?

oddsson (approved):

apps/judicial-system/web/src/utils/mutations.ts
gudjong: Yes, for now. This is just a temporary name to prevent naming conflicts. This particular query will be removed in later PRs.

PR #13136, opened 2023-12-15
fix(parental-leave): alert message and period length fix

karenbjorg:
## What

Add alert message for the time being because applicant can't register employee if he is receiving unemployment benefits.
Two different period length calculation that was causing validation error.

## Checklist:


ylfahfa (approved):
LGTM

PR #13180, opened 2023-12-18
feat(license-service): Refactor license service

thorkellmani:
## What

* Make license service support multiple licenses of the same type
* Add license id for unique id for each license

## Why

Passports should go through the license service, and for that to be done the license service must  be able to support multiple licenses of the same type

## Checklist:


thordurhhh (approved):
👏 👏

PR #13184, opened 2023-12-18
feat(user-notifications): Connect notifications to user-profile v2

GunnlaugurG:
## What

Updated user-notifacations to connect to user-profile v2, and use emailNotifications to decide if email should be sent or not.

## Why

So settings from user-profile v2 controls email notifications to users.

## Screenshots / Gifs

## Checklist:


apps/services/user-notification/src/app/modules/notifications/notifications.module.ts
```diff
@@ -84,26 +89,35 @@ import {
...
+      provide: api,
       useFactory: () =>
-        new userProfile.UserProfileApi(
-          new userProfile.Configuration({
+        new api(
+          new Configuration({
             basePath: environment.userProfileServiceBasePath,
             fetchApi: createEnhancedFetch({
-              name: 'services-user-notification',
+              name: name,
```
saevarma, line 108: When this is refactored to the client library the convention is to use the same name for all apis provided. In this case it should be `clients-user-profile`

apps/services/user-profile/src/app/v2/userToken.controller.ts
```diff
@@ -0,0 +1,47 @@
...
+import { UserDeviceTokenDto } from '../user-profile/dto/userDeviceToken.dto'
+
+const namespace = '@island.is/user-profile/v2/userTokens'
+
+@UseGuards(IdsAuthGuard, ScopesGuard)
+@Scopes(UserProfileScope.admin)
+@ApiTags('v2/userTokens')
+@ApiSecurity('oauth2', [UserProfileScope.admin])
+@Controller({
+  path: 'userTokens',
```
saevarma, line 19: The device tokens are sub-resource of the users so we should let the path be `users/.national-id/device-tokens`

```suggestion
  path: 'users/.national-id/device-tokens',
```

apps/services/user-profile/src/app/v2/userToken.controller.ts
```diff
@@ -0,0 +1,47 @@
...
+  version: ['2'],
+})
+@Audit({ namespace })
+export class UserTokenController {
+  constructor(private readonly userTokenService: UserTokenService) {}
+
+  @Get('/.nationalId')
```
saevarma, line 26: Then this is handling the base route of the controller
```suggestion
  @Get()
```

apps/services/user-profile/src/app/v2/userToken.controller.ts
```diff
@@ -0,0 +1,47 @@
...
+          description: 'National id of the owner of the device tokens',
+        },
+      },
+    },
+    response: { status: 200, type: [UserDeviceTokenDto] },
+  })
+  @Audit<UserDeviceTokenDto[]>({
+    resources: (profile) => profile.map((p) => p.id),
+  })
+  findUserDeviceToken(
```
saevarma, line 42: Let the function name reference that this is handling the collection endpoint, it is getting list of tokens.
```suggestion
  findAllUserDeviceTokens(
```

apps/services/user-profile/test/fixture-factory.ts
```diff
@@ -64,4 +65,13 @@ export class FixtureFactory {
       tries: tries,
     })
   }
+
+  async createUserDeviceToken({ nationalId, deviceToken }) {
+    const userDeviceTokenModel = this.get(UserDeviceTokens)
+
+    return await userDeviceTokenModel.create<UserDeviceTokens>({
```
saevarma, line 72:
```suggestion
    return userDeviceTokenModel.create<UserDeviceTokens>({
```

apps/services/user-profile/src/app/v2/userToken.controller.ts
```diff
@@ -0,0 +1,47 @@
...
+import { Documentation } from '@island.is/nest/swagger'
+
+import { UserTokenService } from './userToken.service'
+import { UserDeviceTokenDto } from '../user-profile/dto/userDeviceToken.dto'
+
+const namespace = '@island.is/user-profile/v2/userTokens'
+
+@UseGuards(IdsAuthGuard, ScopesGuard)
+@Scopes(UserProfileScope.admin)
+@ApiTags('v2/userTokens')
```
saevarma, line 16: I would use the same tag `v2/users` so the codegen will generate this in the same api class.

saevarma (approved):
LGTM

PR #13199, opened 2023-12-20
chore(vehicles): Update who can register mileage

thordurhhh:
## What

* Add resolver to vehicles to check if authenticated user is allowed to post reading to mileage.

## Why

* Updated requirements

## Checklist:


libs/api/domains/vehicles/src/lib/api-domains-vehicles.service.ts
```diff
@@ -163,6 +163,48 @@ export class VehiclesService {
...
+    auth: User,
+    permno: string,
+  ): Promise<boolean> {
+    const res = await this.getVehicleDetail(auth, {
+      clientPersidno: auth.nationalId,
+      permno,
+    })
+
+    // String of owners where owner can delegate registration.
+    const allowedCoOwners = process.env.VEHICLES_ALLOW_CO_OWNERS?.split(', ')
```
osk, line 176: Safer to split on `,` and trim.

```suggestion
    const allowedCoOwners = process.env.VEHICLES_ALLOW_CO_OWNERS?.split(',').map(i => i.trim())
```

libs/api/domains/vehicles/src/lib/api-domains-vehicles-mileage.resolver.ts
```diff
@@ -101,4 +101,16 @@ export class VehiclesMileageResolver {
...
     })
   }
+
+  @ResolveField('canUserRegisterVehicleMileage', () => Boolean, {
+    nullable: true,
+  })
+  resolveCanUserRegisterMileage(
+    @Context('req') { user }: { user: User },
+    @Parent() overview: VehicleMileageOverview,
+  ): Promise<boolean> {
```
osk, line 111: Should this be a complex type that has `canRegister` as property? Then if we need to add more info about the mileage (e.g. if somebody should not see it? dunno, but something might come up!) we could add property instead of having to create a new thing

osk (approved):
lookin' good!

libs/api/domains/vehicles/src/lib/api-domains-vehicles-mileage.resolver.ts
osk: nm, reading it wrong

saevarma (approved):
Core files LGTM

libs/api/domains/vehicles/src/models/getVehicleDetail.model.ts
```diff
@@ -302,6 +302,9 @@ export class VehiclesOperator {
   @Field(() => Number, { nullable: true })
   serial?: number | null
 
+  @Field(() => Boolean, { nullable: true })
+  mainOperator?: boolean | null
+
   @Field(() => Boolean, { nullable: true })
   mainoperator?: boolean | null
 }
```
thorkellmani, line 310: ?

libs/service-portal/assets/src/screens/VehicleMileage/VehicleMileage.tsx
```diff
@@ -125,6 +125,8 @@ const VehicleMileage = () => {
     data?.vehicleMileageDetails?.requiresMileageRegistration
 
   const actionLoading = putActionLoading || postActionLoading
+  const hasUserPostAccess =
+    data?.vehicleMileageDetails?.canUserRegisterVehicleMileage === true
```
thorkellmani, line 129: Seems unnecessary

thorkellmani (approved):
LGTM, minor comments tho

lodmfjord (approved):
lgtm

PR #13206, opened 2023-12-21
fix(auth-admin): Publish super admin fields

GunnlaugurG:
## What

Fix issue where non super admins can publish a client with super admin fields. Super admin fields should have default value instead.

## Why

Non super admins should not have access to set those fields even when they are publishing from another branch

## Screenshots / Gifs

## Checklist:


libs/auth-api-lib/src/lib/clients/admin/admin-clients.service.ts
```diff
@@ -768,4 +779,8 @@ export class AdminClientsService {
       ),
     )
   }
+
+  private isUserSuperAdmin = (user: User) => {
```
saevarma, line 783: ~ I would rename the function to better match the `AdminPortalScope.idsAdminSuperUser`

```suggestion
  private isSuperUser = (user: User) => {
```

saevarma (approved):
LGTM

PR #13208, opened 2023-12-21
fix(old-age-pension): send in application type for old-age pension eligible

karenbjorg:
## What

send in application type for old-age pension eligible

## Why

Different age check for different type of old-age pension

## Checklist:


veronikasif (approved):
LGTM

PR #13273, opened 2024-01-04
feat(service-portal): Add vehicle list v3

thordurhhh:
## What

Add new query resolver for vehicle list, let's call it v3.

## Why

The first one could not return all the info we needed.
The second one was too slow to respond to large queries, so it would eventually time out.
This one solves all our current problems.

## Checklist:


libs/api/domains/vehicles/src/models/usersVehicles.model.ts
```diff
@@ -196,3 +199,60 @@ export class VehiclesList {
   })
   createdTimestamp?: string
 }
+
+@ObjectType()
+export class VehicleListed {
```
thorkellmani, line 204: Weird name, why not just Vehicle?

libs/api/domains/vehicles/src/models/usersVehicles.model.ts
```diff
@@ -196,3 +199,60 @@ export class VehiclesList {
...
+  colorName?: string
+
+  @Field({ nullable: true })
+  modelYear?: string
+
+  @Field(() => Boolean, { nullable: true })
+  requiresMileageRegistration?: boolean | null
+
+  @Field(() => Boolean, { nullable: true })
+  canRegisterMilage?: boolean | null
```
thorkellmani, line 233: Why can this be null? Isn't either true/false or unknown (undefined)?

libs/api/domains/vehicles/src/models/usersVehicles.model.ts
```diff
@@ -196,3 +199,60 @@ export class VehiclesList {
...
+  nextMainInspection?: Date
+}
+
+@ObjectType()
+export class VehiclesListV2 {
+  @Field(() => [VehicleListed], { nullable: true })
+  vehicleList?: VehicleListed[]
+
+  @Field({ nullable: true })
+  paging?: VehiclePaging
```
thorkellmani, line 254: PaginatedResponse?

libs/api/domains/vehicles/src/lib/api-domains-vehicles.service.ts
```diff
@@ -70,6 +72,22 @@ export class VehiclesService {
...
+  ): Promise<CurrentVehiclesWithMilageAndNextInspDtoListPagedResponse> {
+    return await this.getVehiclesWithAuth(
+      auth,
+    ).currentvehicleswithmileageandinspGet({
+      ...input,
+      permno: input.permno
+        ? input.permno.length < 5
+          ? `${input.permno}*`
+          : `${input.permno}`
+        : undefined,
```
thorkellmani, line 87: spooky

thorkellmani (approved):
LGTM 🦓 
Some comments for posterity

libs/api/domains/vehicles/src/lib/api-domains-vehicles.service.ts
thordurhhh: its for wildcard search. 5 and over will be used for direct search. 1-5 wildcard. otherwise undefined resulting in no 
permno query sent.

libs/api/domains/vehicles/src/models/usersVehicles.model.ts
thordurhhh: I kinda agree, but there are three different vehicle responses.

1. Detail vehicle old service
2. Overview vehicle old service
3. Overview vehicle new service.

This is nr 3. A type of vehicle only shown in the list. So that's where the name comes from 🤷 
`Vehicle` could be a bit confusing because of that.

libs/api/domains/vehicles/src/models/usersVehicles.model.ts
thordurhhh: yeah makes sense 👍

PR #13280, opened 2024-01-04
fix(service-portal): hotfix - Vehicle mileage editing mode

thordurhhh:
hotfix: https://github.com/island-is/island.is/pull/13277


thorkellmani (approved):
LGTM

PR #13287, opened 2024-01-05
fix(service-portal): Update validation rule for mileage input

thordurhhh:
## What

Update validation rule for mileage input

## Why

If we're in editing mode, we want to find the highest confirmed registered number, ignoring all Island.is registrations from today.

API was updated to add new transactions instead of updating the current one, so we need to make changes in the UI to reflect that.

## Checklist:


thorkellmani (approved):
LGTM, might optimize the utility function tho

PR #13289, opened 2024-01-05
fix(service-portal): Hotfix - Read date validation mileage input

thordurhhh:
https://github.com/island-is/island.is/pull/13287


thorkellmani (approved):
Lgtm

PR #13302, opened 2024-01-08
test(j-s): Use pdf buffer for appeal files

unakb:
## What

Use a generated pdf buffer for appeal case files instead of a static pre-made pdf

## Why

It's cleaner and we can change the file names and text as needed to easily decipher between different types of case files in the UI

## Checklist:


AndesKrrrrrrrrrrr (approved):
Nice and clean :heart:

PR #13317, opened 2024-01-09
fix(service-portal): Update position for header nav in desktop

thordurhhh:
## What

Update how we calculate the sidebar position for desktop navigation menu.

## Why

When zoomed out the menu goes wayyy off, and in some cases off screen.

* Note the `75%` number would only kick in when the rare case of `getBoundingClientRect()` would not return a number. The browser support for `getBoundingClientRect()` is more than enough for what we want to cover.

#### Currently zoomed out 67% Chrome:

Screenshot 2024-01-09 at 13 26 04

#### After changes zoomed out 67% Chrome:

Screenshot 2024-01-09 at 13 25 24

## Checklist:


thorkellmani (approved):
LGTM

PR #13330, opened 2024-01-10
feat(user-notification): Old data cleanup cronjob

rafnarnason:
This pr adds a cronjob feature which in accordance with the decision to only keep notifications until six months old deletes all table rows older than 6 months - using kubernetes cronjob ... plus updated readme

## What

## Why

## Screenshots / Gifs

## Checklist:


apps/services/user-notification/src/app/modules/notifications/notifications.service.ts
```diff
@@ -285,4 +286,30 @@ export class NotificationsService {
       )
     }
   }
+
+  @Cron('0 * * * *') // run once every hour
```
kksteini, line 290: Is it necessary to do this every hour?
If the resolution is for six months is it not sufficient f.x. to do it once a day with f.x: `@Cron('0 0 * * *')`

kksteini (approved):

apps/services/user-notification/src/app/modules/notifications/notifications.service.ts
rafnarnason: true - I just want to see this working on dev hourly to begin with, and then followup PR changing the cron a once a day

AndesKrrrrrrrrrrr: Hey, just noticed this while browsing.

How about using '@hourly' or '@weekly'?

rafnarnason: splendid idea

apps/services/user-notification/src/app/app.module.ts
```diff
@@ -5,6 +5,7 @@ import { SequelizeModule } from '@nestjs/sequelize'
 import { SequelizeConfigService } from './sequelizeConfig.service'
 import { AuthConfig, AuthModule } from '@island.is/auth-nest-tools'
 import { environment } from '../environments/environment'
+import { ScheduleModule } from '@nestjs/schedule'
```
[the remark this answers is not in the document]
rafnarnason: Kubernetes Cron implemented, please review

package.json
```diff
@@ -326,6 +326,7 @@
     "@graphql-codegen/typescript-react-apollo": "3.3.3",
     "@graphql-codegen/typescript-resolvers": "2.7.3",
     "@jeremybarbet/contentful-extension-scripts": "0.22.7",
+    "@nestjs/schedule": "4.0.0",
```
baering, line 329: Remove this dependency?

baering (changes requested):

package.json
rafnarnason: of course - well spotted

baering (approved):
Nice LGTM 👍

PR #13341, opened 2024-01-10
feat(portals-admin): Institution application lookup

Halli1414:
# Institution application lookup

## What

Admin portal for Institutions to look up their own applications from users.
This screen is the same as the current admin application portal with restrictions on which appliactions they can query.
New endpoints for instituition application lookup.
New mapper for Institutions nationalIds, some are missing

## Why

Island.is admin are getting many requests for application look up from institutions, this feature should reduce suchs requests.

## Screenshots / Gifs

image

## Checklist:


thordurhhh (approved):
service-portal LGTM.

libs/application/api/core/src/lib/application/application.service.ts
```diff
@@ -111,6 +115,58 @@ export class ApplicationService {
...
+    page: number,
+    count: number,
+    status?: string,
+    applicantNationalId?: string,
+    from?: string,
+    to?: string,
+  ): Promise<ApplicationPaginatedResponse> {
+    const statuses = status?.split(',')
+    const typeIds = this.getTypeIdsForInstitution(nationalId)
+    const toDate = to ? endOfDay(new Date(to)) : undefined
```
obmagnusson, line 129: No option to filter by a specific Time?

apps/application-system/api/src/app/modules/application/tools/applicationAdminPaginated.serializer.ts
```diff
@@ -0,0 +1,175 @@
+import {
```
obmagnusson, line 1: Should just use the admin one.

obmagnusson (changes requested):
Great awesome well job!! 💯 🎉 . Just some minor nuances and questions really.

libs/application/api/core/src/lib/application/application.service.ts
Halli1414: I was trying to mimic the admin portal where to date is inclusive but you're right, should be done client side and leave the posibility of a specific time filter from the api

libs/auth-api-lib/seeders/data/scope-application-system-admin.ts
```diff
@@ -1,6 +1,18 @@
 import { compose, createScope } from './helpers'
 
 export const up = compose(
+  createScope({
```
saevarma, line 4: Please contact and keep @jonarnarbriem in the loop to verify the configuration in each deployed environment as the seed script might be buggy, it has not been maintained for a while.

saevarma (approved):
LGTM

PR #13346, opened 2024-01-11
chore(shared): Change max-age to 15 instead of 30 seconds

RunarVestmann:
# Change max-age to 15 instead of 30 seconds

## What

* To reduce the time it takes for new content to appear on the prod web we are opting for halving the amount of time that needs to pass until pages are considered stale

## Why

* Content changes take a while to reach production, this is an effort (among other things) to reduce that time
* This was a decision from a meeting with representatives of Digital Iceland and the Core team

## Checklist:


saevarma (approved):
LGTM

PR #13358, opened 2024-01-12
feat(j-s): Add prosecutors, judges and registrar's emails to info cards

oddsson:
# Add prosecutors, judges and registrar's emails to info cards

[Asana](https://app.asana.com/0/1199153462262248/1206292245391388/f)

## What

Add prosecutors, judges and registrar's emails to info cards

## Why

If prosecutors, f.x. want to communicate with judges in a specific case, there is no way for them to contact them... Until now :) 

## Screenshots / Gifs

Screenshot 2024-01-12 at 10 59 01

## Checklist:


apps/judicial-system/web/src/components/FormProvider/limitedAccessCase.graphql
```diff
@@ -54,6 +54,7 @@ query LimitedAccessCase($input: CaseQueryInput!) {
       id
       name
       title
+      email
```
unakb, line 57: Isn't the defender also allowed to see the email addresses of the judge and registrar? If so please add that to the field selectors below

apps/judicial-system/web/src/components/InfoCard/InfoCard.tsx
```diff
@@ -47,15 +59,18 @@ const UniqueDefenders: React.FC<
       </Text>
       {uniqueDefenders.map((defender, index) =>
         defender?.name ? (
-          <Box display="flex" key={defender.name}>
-            <Text>
-              {`${defender.name}${defender.email ? `, ${defender.email}` : ''}${
-                defender.phoneNumber ? `, s. ${defender.phoneNumber}` : ''
-              }`}
+          <Box display="flex" key={defender.name} role="paragraph">
```
unakb, line 62: Should we hide this if the user is a defender so we're not displaying his own email to him for no reason?

apps/judicial-system/web/src/components/FormProvider/limitedAccessCase.graphql
oddsson: Jess! Thanks 🙏

apps/judicial-system/web/src/components/InfoCard/InfoCard.tsx
oddsson: I think we should leave it as is. You could make the case that it's good for defenders to see that the correct email and phone number is set in RVG..

unakb (approved):
Nice :)

PR #13364, opened 2024-01-12
fix(service-portal): Use overview url as fallback for detailurl

thordurhhh:
## What

Use overview url as fallback for detail-url

## Why

We have seen instances of documents not containing a url in detail, when there is in fact a url in overview.

## Checklist:


Tryggvig (approved):
LGTM 🫡

PR #13365, opened 2024-01-12
fix(service-portal): Overview fallback url Hotfix

thordurhhh:
hotfix: https://github.com/island-is/island.is/pull/13364 🔥


thorkellmani (approved):
lGTM

PR #13394, opened 2024-01-16
fix(ids-api): Making the user profile scope optional to not break service startup

saevarma:
## What

Refactoring the `UserProfileClientConfig` to allow the scope to be optional.

## Why

So service startup doesn't break if scope is missing.

## Screenshots / Gifs

N/A

## Checklist:


snaerseljan (approved):
👌 nice one

lodmfjord (approved):
LGTM

PR #13414, opened 2024-01-16
fix(user-notifications): Add IDS client config env var for service infra

saevarma:
## What

Add IDS client config for the service infra. Technically its only required by the worker, but we need to refactor the module setup of the notifications service a bit, and I don´t want to do that in a rush.

## Why

To fix the service to be able to startup.

## Screenshots / Gifs

N/A

## Checklist:


lodmfjord (approved):
LGTM

PR #13441, opened 2024-01-18
test(j-s): Fix buggy test and some flakiness

unakb:
## What

Our tests have been running very smoothly locally but they are very flaky when run on dev. We think its due to network requests running slower in that environment (coupled with the speed of playwright) and so we need to add in more waiting for requests to finish before continuing.
Also added an option to not update the entire workingCase when posting updates, instead we just change the state value that we just updated. 

## Screenshots / Gifs

https://github.com/island-is/island.is/assets/4820859/63342166-9d34-47e8-a327-74e01d0208b7

## Checklist:


apps/judicial-system/web/src/routes/CourtOfAppeal/Ruling/Ruling.tsx
```diff
@@ -114,6 +114,7 @@ const CourtOfAppealRuling: React.FC<React.PropsWithChildren<unknown>> = () => {
       ],
       workingCase,
       setWorkingCase,
+      false,
```
oddsson, line 117: I'm confused.. `setAndSendCaseToServer` only takes three arguments (updates, workingCase and setWorkingCase). Am I missing something?

oddsson (approved):

PR #13459, opened 2024-01-19
feat(user-notification): Add user notification client

thordurhhh:
## What

Add user-notification client

## Why

New client for requests to user-notification service

## Checklist:


libs/clients/user-notification/src/lib/apiConfiguration.ts
```diff
@@ -0,0 +1,38 @@
...
+            ? {
+                issuer: idsConfig.issuer,
+                clientId: idsConfig.clientId,
+                clientSecret: idsConfig.clientSecret,
+                scope: config.scope,
+                mode: 'auto',
+              }
+            : undefined,
+        }),
+        basePath: `${config.basePath}/v1`,
```
baering, line 34: Was the `/v1` basePath postfix intentional?

baering (approved):
LGTM

libs/clients/user-notification/src/lib/apiConfiguration.ts
thordurhhh: Yes, it was. Perhaps it should be generated by the service as with some other services. But it is not auto generated in this case.

robertaandersen (approved):
LGTM

PR #13473, opened 2024-01-21
chore(vehicles): Add mileage option as number

thordurhhh:
## What

* Update mileage service to use number instead of string. 
* Keeping in old string value `mileage` for backwards compatibility.

## Why

* Service change requires us to use `mileage` as integer. 
* Backwards compatibility has to be in place as the app can not update their code right away. So until they can use the integers we will need to keep `mileage` as a string value.

## Checklist:


libs/api/domains/vehicles/src/lib/api-domains-vehicles-mileage.resolver.ts
```diff
@@ -59,23 +60,37 @@ export class VehiclesMileageResolver {
...
   @Audit()
-  postVehicleMileageReading(
+  async postVehicleMileageReading(
     @Args('input') input: PostVehicleMileageInput,
     @CurrentUser() user: User,
   ) {
-    return this.vehiclesService.postMileageReading(user, input)
+    const res = await this.vehiclesService.postMileageReading(user, {
+      ...input,
+      mileage: Number(input.mileage ?? input.mileageNumber),
```
osk, line 69: I'd move out of the call and extend the logic a bit, something like

```
let mileage

if (input.mileageNumber) {
  mileage = input.mileageNumber
} else if (input.mileage) {
  mileage = Number.parseInt(input.mileage)
}

if (!mileage || Number.isNaN(mileage)) {
  // error
}
  
```

thordurhhh: Yeah, my initial though was to handle this differently.
But after thinking about this, why not let the service itself handle if there is a faulty input?

With the setup in the PR it will return an error if the input from `mileage` is anything other than a 'string as number' (eg. "ABC").
And if it is a correct string "123" it will pass.

Same with if there's no input for `input.mileage` it will get the input from `input.mileageNumber` and if that's incorrect it will return an error as well. Same as it would in the initial setup where we would just take the input from the mutation as is and let the service handle it.

osk (approved):

PR #13491, opened 2024-01-22
chore(j-s): Prosecutors Office

gudjong:
# Prosecutors Office

[Geyma embætti stofnanda sérstaklega í RVG](https://app.asana.com/0/1199153462262248/1206153194238639/f)

## What

- Stores the prosecutors office id directly in the case.

## Why

- To make sure the case stays with the prosecutoras office even when the creating prosecutor moves to a different office.

## Checklist:


apps/judicial-system/backend/src/app/modules/case/internalCase.service.ts
```diff
@@ -463,6 +463,7 @@ export class InternalCaseService {
             prosecutorId:
               creator.role === UserRole.PROSECUTOR ? creator.id : undefined,
             courtId: creator.institution?.defaultCourtId,
+            prosecutorsOfficeId: creator.institution?.id,
```
oddsson, line 466: Please remove Institution and User imports as they are no longer being used.

apps/judicial-system/backend/src/app/modules/case/internalCase.service.ts
```diff
@@ -587,6 +575,15 @@ export class InternalCaseService {
             this.config.archiveEncryptionKey,
             { iv: CryptoJS.enc.Hex.parse(uuidFactory()) },
           ).toString(),
+          // To decrypt:
```
oddsson, line 578: Is this meant to be a part of this PR?

apps/judicial-system/backend/src/app/modules/case/test/internalCaseController/create.spec.ts
```diff
@@ -97,11 +97,13 @@ describe('InternalCaseController - Create', () => {
 
   describe('creating case', () => {
     const userId = uuid()
+    const prosecutorsOfficeId = uuid()
```
oddsson, line 100: Please remove Institution import as it's no longer being used.

oddsson (approved):
Nice! There are just a couple of unused imports..

apps/judicial-system/backend/src/app/modules/case/internalCase.service.ts
gudjong: Yes, every time we need to decrypt the arvhived information we have had a hard time figuring out how.

PR #13495, opened 2024-01-23
fix(application-estate): Estate unconditional marshalling

kksteini:
Need to expand representative data for json with empty fields, if not present in the answers data.
Otherwise the backend of Sýslumenn breaks on the other end due to "JsonSerializationError"s.

## Screenshots / Gifs

## Checklist:


libs/application/template-api-modules/src/lib/modules/templates/estate/utils/processUploadData.ts
```diff
@@ -105,16 +105,12 @@ export const generateRawUploadData = (
...
-          },
-        }
-      : { representative: undefined }),
+    representative: {
+      email: answers.representative?.email ?? '',
+      name: answers.representative?.name ?? '',
+      phoneNumber: answers.representative?.phone ?? '',
+      ssn: answers.representative?.nationalId ?? '',
+    },
     ...(answers.deceasedWithUndividedEstate?.spouse?.nationalId
```
kksteini, line 114: This looks to be conditional in the same way but both branches are populated unlike the one fixed here.

albinagu (approved):

PR #13496, opened 2024-01-23
feat(application-estate): Making estates selectable in the beginning of the application

kksteini:
## What

A selection based on estates from the response.
Note: This has breaking changes but there are protections in place (albeit not perfect) that will have to be removed after some time has passed. This is to ensure that legacy applications in the estate system will not be invalidated.

## Why

Previously we established that dealing with Estates, one at a time, would be sufficient and they would simply "pop" off the stack. However, older cases are not going away in some cases and therefore we need the users to be able to choose.

## Screenshots / Gifs

[Screencast from 2024-01-25 13-39-54.webm](https://github.com/island-is/island.is/assets/77672665/6d7b5548-600a-4321-8cc8-9f048f2affb9)

## Checklist:


libs/application/templates/estate/src/forms/Sections/estateInfoSelection.ts
```diff
@@ -0,0 +1,49 @@
...
+export const estateInfoSelection = buildMultiField({
+  id: 'estateSelect',
+  title: m.chooseEstate,
+  description: m.chooseEstateDescription,
+  children: [
+    buildDescriptionField({
+      id: 'spaceSubmission',
+      title: '',
+      space: 'containerGutter',
+    }),
```
albinagu, line 20: sleppa þessu space-i?

HelgiHH (approved):
lgtm

PR #13521, opened 2024-01-23
chore(j-s): Refactor updateCase functionality

oddsson:
# Refactor updateCase functionality

[Asana](https://app.asana.com/0/1199153462262248/1201255961086607/f)

## What

Refactor updateCase functionality. We used to call the server with updates and overwrite the case state with a new case we got from the server. This causes issues on slow internet connections and in our e2e tests. 

We now call the server, but don't listen to it's response. 

## Checklist:


unakb (changes requested):
Tested it a bit locally and I found one place where the change is breaking, its in the AppealCase component where you choose the 3 appeal judges, the dropdowns don't update anymore to remove the chosen judge, so you can choose the same one 3 times.

AndesKrrrrrrrrrrr (approved):
system-e2e LGTM

unakb (approved):

PR #13522, opened 2024-01-23
feat(user-profile): Only migrate email/phone when migrated profile has values

baering:
## What

When this rule applies:
- Migrated profile exists in the database
- Existing and migrated profiles do not have matching email/phone
- Migrated profile was exported more recently than the existing profile's modified date

The next step is to overwrite email/phone in the existing profile with the values from the migrated profile, because when an existing user profile is updated it _should_ also be updated through an API call to advania where the migrated profiles come from. But in case it wasn't, and there is null values in the fields of the migrated profile but not the existing profile, we do not want to overwrite those values.

So, given:

```typescript
const SampleMigratedAdvaniaProfileBothNull = {
  // ... some fields
  email: null,
  mobilePhoneNumber: null,
  exported: '2024-01-23',
}

const SampleMigratedAdvaniaProfileBothDefined = {
  // ... some fields
  email: 'valid.email@example.dev',
  mobilePhoneNumber: '8333337',
  exported: '2024-01-23',
}

const SampleExistingUserProfileEmailAndPhone = {
  // ... some fields
  email: 'some.valid.email@example.dev',
  emailVerified: true,
  emailStatus: 'VERIFIED',
  mobilePhoneNumber: '1333337',
  mobilePhoneNumberVerified: true,
  mobileStatus: 'VERIFIED',
  modified: '2023-12-24',
}

const SampleExistingUserProfileOnlyEmail = {
  // ... some fields
  email: 'some.valid.email@example.dev',
  emailVerified: true,
  emailStatus: 'VERIFIED',
  mobilePhoneNumber: null,
  mobilePhoneNumberVerified: false,
  mobileStatus: 'NOT_VERIFIED',
  modified: '2023-12-24',
}

const SampleExistingUserProfileOnlyPhone = {
  // ... some fields
  email: null,
  emailVerified: false,
  emailStatus: 'NOT_VERIFIED',
  mobilePhoneNumber: '1333337',
  mobilePhoneNumberVerified: true,
  mobileStatus: 'VERIFIED',
  modified: '2023-12-24',
}
```

- For `SampleExistingUserProfileEmailAndPhone` no overwrite would occur for either migrated profile
- For `SampleExistingUserProfileOnlyEmail` the phone would be overridden given `SampleMigratedAdvaniaProfileBothDefined`
- For `SampleExistingUserProfileOnlyPhone` the email would be overridden given `SampleMigratedAdvaniaProfileBothDefined`

_When either field in an existing user profile is overridden by the same field in a migrated profile the `status` is set to `NOT_VERIFIED` and `verified` is set to `false`._

For more info see the unit tests.

## Why

To have belti and axlabönd.

## Screenshots / Gifs

## Checklist:


saevarma (approved):
🛡️

PR #13531, opened 2024-01-24
feat(web): Allow search-indexer to call web

RunarVestmann:
# Allow search-indexer to call web

search-indexer service needs to be able to call web so we can invalidate the redis cache for web graphql queries


lodmfjord (approved):
LGTM

PR #13532, opened 2024-01-24
fix(consultation-portal): Kam 2072 throwing 500 errors when it should not

karlarnar:
# Throwing 500 errors when it should not

https://veflausnir.atlassian.net/browse/KAM-2072

## What

- Moving error screens to components
- Removing 500 page because if going straight to 500 showed as 500 error
- Fix loading state in /minaraskriftir not showing
- Going to /minaraskriftir goes straight to login instead of flashing the site
- Email box does not flash email input if user is verified
- if there are 500 errors when fetching server side, user is shown a 500  page instead of being redirected
- if there are 404 errors in server side, then the user is shown a 404 page instead of  flashing the content and then shown a 404 page
- Going to /500 shows a 404 page
- Removing some unused props
- fixing lint errors
- Ran playwright tests and all passed

## Why

## Screenshots / Gifs

## Checklist:


diljath (approved):
LGTM

PR #13537, opened 2024-01-24
fix(j-s): Display create case in mobile view

unakb:
[Leyfa notendum að stofna mál á mobile](https://app.asana.com/0/1199153462262248/1206404825734473/f)

## What

Show create case button in mobile view

## Checklist:


apps/judicial-system/web/src/routes/Shared/Cases/Cases.tsx
```diff
@@ -86,7 +86,7 @@ const CreateCaseButton: React.FC<
   }, [formatMessage, user?.role])
 
   return (
-    <Box display={['none', 'none', 'block']}>
+    <Box>
```
oddsson, line 89: I think we can just remove the Box all together..

oddsson (approved):

PR #13567, opened 2024-01-26
fix(user-profile): Move nudge error message

GunnlaugurG:
## What

The error message for nudge was inline with the checkbox (default checkbox behaviour). I moved the error message out of the checkbox component to be below the checkbox and save button  

## Why

Because it looked a bit strange inline.

## Screenshots / Gifs

image

## Checklist:


thorkellmani (approved):
LGTM

PR #13583, opened 2024-01-29
feat(web): Chart improvements and fixes

baering:
## What

Chart Number Box:
- Fixing floating point rendering issue (28.9999997% vs 29%)
- Reducing font size for main value on small devices
- Fixed error in YoY / MoM calculation

Pie chart:
- Increasing chart size
- Improving labels

General:
- Improving formatting for number and percentages displayed in charts
- Made it possible to use the `sourceData` field to use custom data (JSON blob) rather than the statistic data source
- Connected the interval field to result filtering so it is now possible to skip every n-th entry in the results
  - _For example if you want to get 36 months of data but only show every 12th month to show difference between years_

Left for later:
- Refactor the `interval` property that is now located on `Chart Component` contentful model up and into the `Chart` model to control all components

## Why

To improve UX and DX.

## Screenshots / Gifs
Some examples

#### Before
image
_This floating point issue has been fixed_

#### After
image
_Percentages differ from an incorrect division that has now been fixed_

#### Before
image

#### After
image

#### Before
image

#### After
image
_Reduced font size on mobile and an example of formatting change, not showing such precision unless we have a low number_

#### Before
image
_Using the interval field (=12) in Contentful with 24 months of data but nothing happens_

#### After
image
_Interval fields (=12) now skips everything but the `i % 12 === 0` entries_

## Checklist:

### DIFF
--- a/libs/clients/statistics/src/lib/statistics.utils.ts
+++ b/libs/clients/statistics/src/lib/statistics.utils.ts
@@ -14,6 +14,7 @@ import {
   MONTH_NAMES,
 } from './statistics.constants'
 import { EnhancedFetchAPI } from '@island.is/clients/middlewares'
+import type { Unwrap } from '@island.is/shared/types'
 
 export const _tryToGetDate = (value: string | null) => {
   if (!value) {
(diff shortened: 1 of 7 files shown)


libs/clients/statistics/src/lib/statistics.utils.ts
```diff
@@ -14,6 +14,7 @@ import {
   MONTH_NAMES,
 } from './statistics.constants'
 import { EnhancedFetchAPI } from '@island.is/clients/middlewares'
+import type { Unwrap } from '@island.is/shared/types'
```
RunarVestmann, line 17: Unused import

RunarVestmann (approved):

libs/clients/statistics/src/lib/statistics.utils.ts
baering: Well spotted, thanks

PR #13584, opened 2024-01-29
chore(j-s): Police Case Number Prefix

gudjong:
# Police Case Number Prefix

[Get policeCaseNumberPrefix from workingCase.prosecutorsOffice instead of user.institution](https://app.asana.com/0/1199153462262248/1206400159541687/f)

## What

- Uses the police case number prefix from the case's prosecutors office in already existing cases.
- Continues to use the police case number prefix from the user's prosecutors office in new cases.

## Why

- More accurate.

## Checklist:


unakb (approved):
Tested locally, works

PR #13605, opened 2024-01-29
feat(web): Pension calculator

RunarVestmann:
# Pension calculator

## What

* Bringing https://www.tr.is/reiknivel/ to Ísland.is
* "Custom Page" content type added (so the page can fetch translations, configurations and display an alert banner)
* Custom Page content type gets it's translations from the Translation Namespace content type so I added support for that similar to how other apps in the monorepo do it
* GraphQL endpoint added for the pension calculation (I also added stefna as a code owner to the api domain and the client lib since now we own code there)
* There were multiple copies of a `formatCurrency` function throughout the web project so I reused a single one instead
* `size` prop was added to SelectController component so we can appropriately change the size

## Screenshots / Gifs

https://github.com/island-is/island.is/assets/43557895/a0a3a67e-075f-4a4a-b48c-23be7e8d2579

## Checklist:


AndesKrrrrrrrrrrr (approved):
CODEOWNERS LGTM

Toti91 (approved):
Code owned lgtm

libs/api/domains/social-insurance/src/lib/dtos/pensionCalculation.input.ts
```diff
@@ -0,0 +1,112 @@
...
+  livingCondition?: LivingCondition
+
+  @Field(() => Boolean, { nullable: true })
+  mobilityImpairment?: boolean
+
+  @Field(() => String, { nullable: true })
+  startDate?: string
+
+  @Field(() => Number, { nullable: true })
+  childCount?: number
```
thorkellmani, line 60: FYI you don't need to explicitly declare the Boolean or the String. However, you need to assign the number (`childCount`) to either Int or Float, according to the[ NestJS docs](https://docs.nestjs.com/graphql/resolvers)

`The type function is required when there's the potential for ambiguity between the TypeScript type system and the GraphQL type system. Specifically: it is not required for string and boolean types; it is required for number (which must be mapped to either a GraphQL Int or Float).`

libs/api/domains/social-insurance/src/lib/models/pensionCalculation.model.ts
```diff
@@ -0,0 +1,32 @@
+import { Field, ObjectType } from '@nestjs/graphql'
+import { CacheField } from '@island.is/nest/graphql'
+
+@ObjectType('SocialInsurancePensionCalculationResponseItem')
+class PensionCalculationItem {
+  @Field(() => String, { nullable: true })
+  name?: string | null
```
thorkellmani, line 7: Isn't it unnecessary to have it be both `null` and `nullable`?

libs/api/domains/social-insurance/src/lib/utils.ts
```diff
@@ -0,0 +1,183 @@
...
+    privatePensionPayments: input.privatePensionPayments,
+    taxCard: input.taxCard,
+    ageOfFirst75DisabilityAssessment: input.ageOfFirst75DisabilityAssessment,
+    installmentClaims: input.installmentClaims,
+    livingConditionRatio: input.livingConditionRatio,
+
+    // Fields that are calculated or mapped
+    hurryPension,
+    delayPension,
+    start: startPension, // TODO: perhaps this is wrong?
```
thorkellmani, line 99: Todo? :eyes:

libs/api/domains/social-insurance/src/lib/utils.ts
```diff
@@ -0,0 +1,183 @@
...
+  }
+
+  let hurryPension = 0
+  let delayPension = 0
+  let startPension: 1 | 2 = 2
+
+  if (input.birthdate) {
+    const birthdate = new Date(input.birthdate)
+
+    const defaultPensionAge = 67 // TODO: hardcoding 67 is not ideal
```
thorkellmani, line 53: Todo? 👀

libs/api/domains/social-insurance/src/lib/models/pensionCalculation.model.ts
RunarVestmann: I want to keep it though 
image

mannipje (approved):

thorkellmani (approved):
LGTM 🦆

PR #13614, opened 2024-01-30
fix(auth-react): Promise needs to be awaited to handle error and trigger the error screen

saevarma:
## What

Adding `await` to correcly handle the error and trigger the error screen when there are issues starting the authentication flow.

## Why

We want to show the general error screen when apps cannot load the authentication due to IDS is down as it is better UX than the three loading dots.

## Screenshots / Gifs

Current behaviour:
image

Fixed:
image

## Checklist:


baering (approved):
LGTM.

Should the call to `userManager.signinRedirect` be extracted into a function to avoid having to write the same code multiple times if it fails?

Something like:

```typescript
type TODO = any
const signinRedirect = (args: TODO) => {
  return userManager.signinRedirect(args).catch((e) => {
    console.log(e)
    setError(e)
  })
}
```

Then `signinRedirect(args)` will handle setting the error if it fails.

Or maybe it's just fine like is 🤷

PR #13615, opened 2024-01-30
fix(car-recycling): Fix E2E mileage not been set

birkirkristmunds:
## What

- Added mileage when deregistering vehicles

## Why

- Fix broken E2E test

## Checklist:


apps/system-e2e/src/tests/islandis/application-system/acceptance/car-recycling.spec.ts
```diff
@@ -75,6 +75,16 @@ applicationTest.describe('Car recycling', () => {
...
         ).toBeVisible()
+
+        const mileage = page.getByRole('textbox', {
+          name: label(carRecyclingMessages.cars.mileage),
+        })
+
+        if (mileage) {
+          await mileage.selectText()
+          await mileage.type('12345')
+        }
```
AndesKrrrrrrrrrrr, line 86: Why is this conditional? We should be using fake users, which we _know_ have mileage.

birkirkristmunds: You are right, my mistake. 
I have removed the if statement

AndesKrrrrrrrrrrr (approved):

PR #13621, opened 2024-01-30
fix(signature-collection): shared service and update config

alexdiljar:
# ...

## What

## Why

## Screenshots / Gifs

## Checklist:

### DIFF
--- a/libs/api/domains/signature-collection/src/lib/guards/needsOwner.guard.ts
+++ b/libs/api/domains/signature-collection/src/lib/guards/needsOwner.guard.ts
@@ -1,15 +1,17 @@
 import { Injectable, CanActivate, ExecutionContext } from '@nestjs/common'
-import { Observable } from 'rxjs'
 import { Reflector } from '@nestjs/core'
 import { BYPASS_AUTH_KEY, getRequest } from '@island.is/auth-nest-tools'
 import { OwnerAccess } from '../decorators/needsOwner.decorator'
+import { SignatureCollectionSignee } from '../models/signee.model'
+import { SignatureCollectionService } from '../signatureCollection.service'
 
 @Injectable()
 export class NeedsOwnerGuard implements CanActivate {
-  constructor(private reflector: Reflector) {}
-  canActivate(
-    context: ExecutionContext,
-  ): boolean | Promise<boolean> | Observable<boolean> {
+  constructor(
+    private reflector: Reflector,
+    private readonly signatureCollectionService: SignatureCollectionService,
+  ) {}
+  async canActivate(context: ExecutionContext): Promise<boolean> {
     const bypassAuth = this.reflector.getAllAndOverride<boolean>(
       BYPASS_AUTH_KEY,
       [context.getHandler(), context.getClass()],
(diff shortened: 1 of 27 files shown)


libs/api/domains/signature-collection/src/lib/guards/needsOwner.guard.ts
```diff
@@ -19,27 +21,35 @@ export class NeedsOwnerGuard implements CanActivate {
     if (bypassAuth) {
       return true
     }
-    const isOwner = this.reflector.get<OwnerAccess>(
+    const ownerRestirction = this.reflector.get<OwnerAccess>(
```
kksteini, line 24:
```suggestion
    const ownerRestriction = this.reflector.get<OwnerAccess>(
```

libs/api/domains/signature-collection/src/lib/guards/needsOwner.guard.ts
```diff
@@ -19,27 +21,35 @@ export class NeedsOwnerGuard implements CanActivate {
...
       'is-owner',
       context.getHandler(),
     )
     // IsOwner decorator not used
-    if (!isOwner) {
+    if (!ownerRestirction) {
```
kksteini, line 29:
```suggestion
    if (!ownerRestriction) {
```

libs/api/domains/signature-collection/src/lib/guards/needsOwner.guard.ts
```diff
@@ -19,27 +21,35 @@ export class NeedsOwnerGuard implements CanActivate {
...
     const user = request.user
     const isDelegatedUser = !!user?.actor?.nationalId
-    if (signee.isOwner) {
-      if (isDelegatedUser) {
-        return isOwner === OwnerAccess.AllowActor ? true : false
+    const { candidate } = signee
+
+    if (signee.isOwner && candidate) {
+      // Check if user is an actor for owner and if so check if registered collector, if not actor will be added as collector
+      if (isDelegatedUser && ownerRestirction === OwnerAccess.AllowActor) {
```
kksteini, line 42:
```suggestion
      if (isDelegatedUser && ownerRestriction === OwnerAccess.AllowActor) {
```

libs/application/template-api-modules/src/lib/modules/templates/signature-collection/signature-list-signing/signature-list-signing.service.ts
```diff
@@ -90,6 +94,10 @@ export class SignatureListSigningService extends BaseTemplateApiService {
         throw new TemplateApiError(errorMessages.maxReached, 405)
       }
     }
+    if (lists.length === 0) {
+      throw new TemplateApiError(errorMessages.active, 405)
```
kksteini, line 98: If lists.length === 0, does that mean there is nothing found or does that reperesent that a user is not allowed to see them?

I don't see 405s that often, so mostly just curious :)

libs/clients/signature-collection/src/lib/signature-collection.types.ts
```diff
@@ -1,5 +1,5 @@
 import { EinstaklingurMaFrambodInfo, MedmaeliDTO } from '../../gen/fetch'
-import { List } from './types/list.dto'
+import { List, ListBase } from './types/list.dto'
```
kksteini, line 2: List and User unused imports?

libs/clients/signature-collection/src/lib/types/collection.dto.ts
```diff
@@ -1,67 +1,32 @@
...
     sofnunEnd: endTime,
     svaedi: areas,
     frambodList: candidates,
     kosning,
   } = collection
-  if (id == null || startTime == null || endTime == null || areas == null) {
+  if (!id || startTime == null || endTime == null || areas == null) {
     logger.warn(
       'Received partial collection information from the national registry.',
       collection,
```
kksteini, line 32: What happens if `id` is set and all other values, `startTime`, `endTime` and `areas` are `undefined`?

Should these be tested against with `!startTime || !endTime || ...etc`?

kksteini: Ah, double equals. Nevermind.
However, should we avoid double equals?

kksteini: Because, doesn't this feel yucky? :laughing: 
```ts
> startTime = undefined
undefined
> startTime == null
true
> startTime === null
false
```

kksteini (approved):

libs/application/template-api-modules/src/lib/modules/templates/signature-collection/signature-list-signing/signature-list-signing.service.ts
alexdiljar: This should be a 404, but it does not mean that nothing exists hust thtat there is nothing currently open for the user.

PR #13625, opened 2024-01-31
chore(j-s): Refactor Cases list

oddsson:
# Refactor Cases list

Asana

## What

Refactor active cases table. This refactor mainly does the following.

- Swap out the x button to delete cases with a context menu.
- Allow users to open cases in a new tab via the context menu. Note that this PR does not implement command clicking a case to open it in a new tab.
- Stop scrolling to the top when entering a page.

## Screenshots / Gifs

https://github.com/island-is/island.is/assets/3789875/a2c85e93-dec2-42e2-9bf8-e7a9c586001f

## Checklist:


unakb (commented):
1 failing test

apps/judicial-system/web/src/routes/Shared/Cases/ActiveCases.tsx
```diff
@@ -360,59 +345,41 @@ const ActiveCases: React.FC<React.PropsWithChildren<Props>> = (props) => {
...
+                          menuLabel={`Valmynd fyrir mál ${c.courtCaseNumber}`}
+                          items={[
+                            {
+                              title: formatMessage(m.contextMenu.openCase),
+                              onClick: () => handleOpenCase(c.id, true),
+                              icon: 'open',
+                            },
+                            {
+                              title: formatMessage(m.contextMenu.deleteCase),
+                              onClick: async () => {
```
unakb, line 358: Just wondering, should we be worried about the possibility of someone accidentally deleting a case when trying to open in a new tab since the 2 options are right next to each other? Is there a need for some kind of confirmation before the deletion happens?

unakb (approved):
Just some thoughts but it all works so I approved :)

apps/judicial-system/web/src/routes/Shared/Cases/ActiveCases.tsx
oddsson: Awesome point! Thanks.. Resolved IRL..

PR #13631, opened 2024-01-31
fix(consultation-portal): KAM-2095: await ctx

karlarnar:
# slug is sometimes null in serversideprops

https://veflausnir.atlassian.net/browse/KAM-2095

## What

- Sometimes when serversideprops is fetching data, the ctx.query.slug is null, so we will try to await the ctx.query.slug
- We will also console.error the id just in case

## Why

## Screenshots / Gifs

## Checklist:


diljath (approved):
LGTM

PR #13635, opened 2024-02-01
feat(j-s): Click send files to actually send files

oddsson:
# Click "send files" to actually send files

[Asana](https://app.asana.com/0/1199153462262248/1206418836382689/f)

## What

We are changing how file upload works on the screens "Appeal to court of appeals", "Statement" and "Appeal case files". From now on, users must click a "Send files" button to actually upload the files, rather than the upload happening when users select files.

## Why

This is clearer for the users.

### When everything works

https://github.com/island-is/island.is/assets/3789875/73655b9c-4a7b-4b60-84c1-516bb27643c1

 

### When a file fails to upload

https://github.com/island-is/island.is/assets/3789875/a7950c31-60d0-4f6e-90b5-47c13faf863a

## Checklist:


RunarVestmann (approved):

gudjong (changes requested):
I tested this localli and noticed two things:

- When uploading many files, the files that are wating for an upload have a "retry" icon.
- When uploading many files and some succeed and others fail, then the successfully uploaded file have a "remove" iceon. Pressing the remove icen removes the files from the upload list, but does not actually remove the file. It would probably be best to not show a "remove" icon for these files.

oddsson:
Thanks 🙏 

> I tested this localli and noticed two things:
> 
>     * When uploading many files, the files that are wating for an upload have a "retry" icon.

>     * When uploading many files and some succeed and others fail, then the successfully uploaded file have a "remove" iceon. Pressing the remove icen removes the files from the upload list, but does not actually remove the file. It would probably be best to not show a "remove" icon for these files.


Regarding point one: You are right, it's confusing that the upload files have a "Retry" icon while waiting to be uploaded. However, the upload component does have a "uploading" icon, which is the retry icon but it spins. So now when files are being uploaded, they get that icon, which I think is clearer for the users.

Regarding point two: Correct again :) Instead of hiding it, files now always have the "Remove" icon (unless the uploading icon is visible) but if the file has been uploaded, we remove the file, otherwise we remove the file from state.

PR #13645, opened 2024-02-01
fix(consultation-portal): KAM-2095: Check if id is a number before qu…

karlarnar:
…erying the API (#13644)

# Check if id is a number before querying the API

https://veflausnir.atlassian.net/browse/KAM-2095
https://github.com/island-is/island.is/pull/13644

## What

Check if the id from the slug is a number before querying the api

## Why

Google bots were trying to access pages here: https://www.google.com/search?q=samradsgatt.island.is+filetype%3Apdf and tried to use the id for mal => https://island.is/samradsgatt/c13fbf57-f69e-ec11-9baf-005056bcce7e which would result in querying the api with the id c13fbf57-f69e-ec11-9baf-005056bcce7e which should only be an integer

## Screenshots / Gifs

## Checklist:

### DIFF
--- a/apps/consultation-portal/pages/mal/[slug].tsx
+++ b/apps/consultation-portal/pages/mal/[slug].tsx
@@ -13,6 +13,8 @@ interface CaseProps {
   is500: boolean
 }
 
+const isNumber = (str: string) => /^\d+$/.test(str)
+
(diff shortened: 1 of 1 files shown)


diljath (approved):
LGTM

PR #13659, opened 2024-02-01
feat: added PodDisruptionBudget support to DSL

busla:
# Prevent voluntary service disruption

PodDisruptionBudget is a Kubernetes resource that ensures that a minimum number of pods are available at any given time. It is used to prevent Kubernetes from killing all pods of a service at once. Mapped to a [PodDisruptionBudget](https://kubernetes.io/docs/tasks/run-application/configure-pdb/).

## Example

image


apps/air-discount-scheme/api/src/index.ts
```diff
@@ -1 +1,2 @@
 export { Discount } from './app/modules/discount/discount.model'
+// deploy-feature
```
saevarma, line 2: Remove comment before merge

saevarma (approved):
LGTM
Is this something we want to enable for the auth APIs + IDS as well?

peturgq (approved):

apps/air-discount-scheme/api/src/index.ts
busla: we really need to include changes in `apps/*/infra/*` in NX affected resolution 😮‍💨

PR #13660, opened 2024-02-02
fix(application-ctao): Xroad path

Halli1414:
# Xroad path for Complaint to Althingi Ombudsman

## What

Move the service path from parameter store to code

## Why

Originally this value was set up as a secret in parameter store but does not need to be there.
Now this value should be properly set up in code.

## Checklist:


libs/clients/althingi-ombudsman/src/lib/clients-althingi-ombudsman.config.ts
saevarma, line 1: You should update the codeowners and set your team as the codeowner of this library.

saevarma (approved):
LGTM

peturgq (approved):

PR #13667, opened 2024-02-02
fix(clients-statistics): Update stale-while-revalidate lifetime to 14 days so we can use cache.

saevarma:
## What

Make stale-while-revalidate valid for 2 weeks so if the site receives a visit once every 2 weeks we should have data in cache.
Remove enhanced fetch debug log env variable.

## Why

To configure caching for prod.
Debugging is done so turnoff to remove log noice.

## Checklist:


busla (changes requested):
`infra/*` LGTM

busla (approved):
LGTM

PR #13724, opened 2024-02-08
fix(j-s): Remove defendant plea alert box from all case types but indictments

oddsson:
# Remove defendant plea alert box from all case types but indictments

[Asana](https://app.asana.com/0/1199153462262248/1206362125834834/f)

## What

Remove defendant plea alert box from all case types but indictments

## Why

This only applies to indictments

## Checklist:


unakb (approved):
Lgtm!

PR #13728, opened 2024-02-08
fix(intellectual-properties): Fix design image list error

thorkellmani:
## What

* Fix json mapping error

## Why

* Its failing

## Checklist:


libs/api/domains/intellectual-properties/src/lib/intellectualProperties.service.ts
```diff
@@ -524,15 +536,27 @@ export class IntellectualPropertiesService {
...
   }
 
   async getDesignImages(
     user: User,
     designId: string,
   ): Promise<Array<Image> | null> {
-    const response = await this.ipService.getDesignImages(user, designId)
+    const response = await this.ipService
+      .getDesignImages(user, designId)
+      .catch((e) => {
```
thordurhhh, line 547: Could also wrap with handle204 ?

thordurhhh (approved):
🚀

PR #13763, opened 2024-02-13
feat(service-portal): Implement standard error/empty states in ui

thorkellmani:
## What

Standardized error/empty state for (almost) every screen in the service portal
Licenses and occupational licenses will be updated later on since they're a bit different

Also updating the `husnaedis-og-mannvirkjastofnun` -> `hms` slug since it's out of sync with the contentful slug

## Why

So each module in the service portal doesn't reinvent the error state wheel every time.

## Screenshots / Gifs

## Checklist:


libs/service-portal/education/src/screens/SecondarySchoolGraduationDetail/SecondarySchoolGraduationDetail.tsx
```diff
@@ -22,6 +22,7 @@ import { useGetInnaDiplomasQuery } from '../SecondarySchoolCareer/Diplomas.gener
 import { tagSelector } from '../../utils/tagSelector'
 import { defineMessage } from 'react-intl'
```
thordurhhh, line 23: unused imports here

thordurhhh (approved):
Nice one ⛷️

libs/react-spa/shared/src/components/problems/Problem.tsx
```diff
@@ -114,7 +114,7 @@ export const Problem = ({
           return (
             <ThirdPartyServiceError
               organizationSlug={organizationSlug}
-              size={size}
+              size={size ?? 'small'}
```
saevarma, line 117: This is changing the default sizing from `large` to `small` for everyone using the shared component. I´m not sure that is what we want. What is the goal here? When do you need it small?

thorkellmani: According to the designs when the error is from a specific service provider, we want to display a small warning. 
image
However, when the error doesn't come from a specific source, we want the large box.
image

Somehow overlooked how far-reaching this change would be.

saevarma (approved):
The Core file LGTM

PR #13779, opened 2024-02-14
fix(fa): Removing guard for fetching api keys so employees can get municipality info 

MargretFinnboga:
# Fixing error 

## Checklist:


raldur (approved):
Death to decorators! 🔥

If it works, it works :)

PR #13791, opened 2024-02-15
chore(system-e2e): service portal assets smoke test

thordurhhh:
## What

Add e2e smoketest for asset screen.
Use mountebank for mocking.

## Why

There was no test for the asset page as the service is unavailable on dev. Now we use mountebank to mock the responses from the service.

## Video

https://github.com/island-is/island.is/assets/24840451/c53b2fc3-8869-4d70-9a73-e910000f6398

## Checklist:


apps/system-e2e/src/tests/islandis/service-portal/setup-xroad.mocks.ts
```diff
@@ -0,0 +1,111 @@
...
+          lodamat: 75000000,
+          notkunBirting: 'Íbúðarhúsalóð',
+          flatarmal: '300000',
+          flatarmalEining: 'm²',
+        },
+        thinglystirEigendur: {
+          thinglystirEigendur: [
+            {
+              nafn: 'Jón Jónsson',
+              kennitala: '1234567890',
```
AndesKrrrrrrrrrrr, line 80: Please change to a repeated number, e.g. `222222-2222` or something starting with `010130-` like most of our gervimenn

apps/system-e2e/src/tests/islandis/service-portal/smoke/assets.spec.ts
AndesKrrrrrrrrrrr, line 1: I think this needs to be in an `acceptance/` folder to use mocks. Smoke tests are run directly against `beta.dev01.devland.is` with no mocks.

thordurhhh: Even if it works with the current setup locally?

AndesKrrrrrrrrrrr: Yes, because in smoke tests (run after merge to `main`), we don't spawn the required `mountebank` container.

PR #13798, opened 2024-02-15
fix(documents): Add url document details

thordurhhh:
## What

Add special case for url document in postholf

## Why

There is no way to get the type in the overview list call. So we need to get by title of doc, as with previous RSK document. 

NOTE: This is a temporary solution to get the url doc to work in the app. Already works on web. We have a new way to approach this, this is only for app compatibility for now. We will remove this soon when we will all be using the same way to display documents.

## Checklist:


thorkellmani (approved):
LGTM

PR #13811, opened 2024-02-16
fix(services-sessions): Fix sessions column too large value

saevarma:
## What

DataDog is reporting errors about a field being to long for database column.
So we migrate the user agent column to be TEXT so it allows Postgres unlimited varchar column.
We then take the first 1024 chars from the user agent string.

## Why

To fix error and lost session tracking due to these failures.

## Screenshots / Gifs

N/A

## Checklist:

### DIFF
--- a/apps/services/sessions/src/app/sessions/sessions.service.ts
+++ b/apps/services/sessions/src/app/sessions/sessions.service.ts
@@ -12,6 +12,7 @@ import { CreateSessionDto } from './create-session.dto'
 import { Session } from './session.model'
 import { SessionsQueryDto } from './sessions-query.dto'
 import { SessionsResultDto } from './sessions-result.dto'
+import { USER_AGENT_MAX_LENGTH } from './constants'
 
 @Injectable()
 export class SessionsService {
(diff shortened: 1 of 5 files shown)


GunnlaugurG (approved):
LGTM !

apps/services/sessions/src/app/sessions/sessions.service.ts
```diff
@@ -97,8 +98,9 @@ export class SessionsService {
 
     return this.sessionModel.create({
       ...rest,
+      userAgent: userAgent.substring(0, USER_AGENT_MAX_LENGTH),
       sessionId: sid,
-      device: this.formatUserAgent(session.userAgent),
+      device: this.formatUserAgent(userAgent),
```
baering, line 103: The `device` column is also of type `string`. Do we need to make sure `formatUserAgent` never exceeds the max length (default length 255)?

saevarma: No I don't think we need to worry about that
the format function is 
```
private formatUserAgent(userAgent: string): string | undefined {
    const ua = uaParser(userAgent)
    const browser = ua.browser.name || ''
    const os = ua.os.name || ''

    return browser || os
      ? `${browser}${browser && os ? ` (${os})` : os}`
      : undefined
  }
```
And the combination of browers name and os name are from ua-parser-js
https://www.npmjs.com/package/ua-parser-js

PR #13812, opened 2024-02-16
fix(service-portal): hotfix - Update recycle car url

thordurhhh:
hotfix https://github.com/island-is/island.is/pull/13810 🔥


thorkellmani (approved):
LGTM

PR #13821, opened 2024-02-16
fix(auth): Speed up delegation queries

eirikurn:
## What

Speed up complicated delegation SQL queries which are causing timeouts in production for specific users.

## Why

These queries are super slow in production because they have a LEFT OUTER JOIN against all delegations in the system containing the respective api scopes.

The original goal is to filter api scopes based on custom delegations which the authenticated user has against the current subject. So this PR changes removes the offending joins and instead conditionally queries the list of custom delegation scopes before the main query, which can then match against them with a simple IN filter.

## Checklist:

### DIFF
--- a/libs/auth-api-lib/src/lib/resources/delegation-resources.service.ts
+++ b/libs/auth-api-lib/src/lib/resources/delegation-resources.service.ts
@@ -1,5 +1,6 @@
 import { ForbiddenException, Inject, Injectable } from '@nestjs/common'
 import { InjectModel } from '@nestjs/sequelize'
+import startOfDay from 'date-fns/startOfDay'
 import { isCompany } from 'kennitala'
 import { and, Op, or } from 'sequelize'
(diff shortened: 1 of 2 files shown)


libs/auth-api-lib/src/lib/resources/delegation-resources.service.ts
```diff
@@ -388,4 +361,28 @@ export class DelegationResourcesService {
     }
     return false
   }
+
+  private async findCustomDelegationScopes(user: User): Promise<string[]> {
```
saevarma, line 365: ~ Rename to indicate that this only looks up for actor, otherwise returns empty `[]`
```suggestion
  private async findActorCustomDelegationScopes(user: User): Promise<string[]> {
```

saevarma (approved):
Really important bug discovery 🔍 
and a simple fix 💅

PR #13833, opened 2024-02-19
fix(j-s): Wait for Accept Test

gudjong:
# Wait for Accept Test

[Laga E2E test](Laga E2E test)

## What

- Waits for case acceptance in judge flow to make sure the request leaves the server before the test quits.
- Waits for file selection in file picker.

## Why

- Sometimes the test quits before the request to accept the case leaves the server.
- We tried removing the file picker dealy, but that caused file picking to fail in some cases. Further investigation needed.

## Screenshots / Gifs

The following screen shot indicates that the "prosecutor should appeal case" test failed but the real failure is in "court should submit decision in case":

image

## Checklist:

### DIFF
--- a/apps/system-e2e/src/tests/judicial-system/smoke/custody-tests.spec.ts
+++ b/apps/system-e2e/src/tests/judicial-system/smoke/custody-tests.spec.ts
@@ -220,7 +220,10 @@ test.describe.serial('Custody tests', () => {
 
     // Confirmation
     await expect(page).toHaveURL(`/domur/stadfesta/${caseId}`)
-    await page.getByTestId('continueButton').click()
+    await Promise.all([
+      await page.getByTestId('continueButton').click(),
+      verifyRequestCompletion(page, '/api/graphql', 'TransitionCase'),
+    ])
   })
 
   test('prosecutor should appeal case', async ({ prosecutorPage }) => {
--- a/apps/system-e2e/src/tests/judicial-system/smoke/custody-tests.spec.ts
+++ b/apps/system-e2e/src/tests/judicial-system/smoke/custody-tests.spec.ts
@@ -221,7 +221,7 @@ test.describe.serial('Custody tests', () => {
     // Confirmation
     await expect(page).toHaveURL(`/domur/stadfesta/${caseId}`)
     await Promise.all([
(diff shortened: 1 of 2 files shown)


apps/system-e2e/src/tests/judicial-system/smoke/custody-tests.spec.ts
```diff
@@ -220,7 +220,10 @@ test.describe.serial('Custody tests', () => {
 
     // Confirmation
     await expect(page).toHaveURL(`/domur/stadfesta/${caseId}`)
-    await page.getByTestId('continueButton').click()
+    await Promise.all([
+      await page.getByTestId('continueButton').click(),
```
AndesKrrrrrrrrrrr, line 224: No need for await if you're awaiting the collector
```suggestion
      page.getByTestId('continueButton').click(),
```

apps/system-e2e/src/tests/judicial-system/utils/helpers.ts
```diff
@@ -30,19 +30,20 @@ async function createFakePdf(title: string) {
...
   clickButton: () => Promise<void>,
   fileName: string,
-) {
+) => {
   const fileChooserPromise = page.waitForEvent('filechooser')
   await clickButton()
 
   const fileChooser = await fileChooserPromise
-  await fileChooser.setFiles(await createFakePdf(fileName))
+  await page.waitForTimeout(1000)
```
AndesKrrrrrrrrrrr, line 42: Why wait some time (which slows tests down), when you could be waiting for some _thing_

gudjong: The reason for this is explained in the PR description. This wait was in the initial file selection test, but later removed because we thought we had found the problem. That, however, turned out to be wrong so we are adding it back. We will continue to improve, but we want and need the tests to pass to be useful. So, we will lieve with this for now.

AndesKrrrrrrrrrrr: An unfortunate solution, but we must go on and get stuff done :muscle:

AndesKrrrrrrrrrrr (approved):

PR #13844, opened 2024-02-20
feat(user-notification): Seen and read and senderId

rafnarnason:
Changing data structure
dropping status in favour of seen and read booleans
adding senderId property
new endpoints for read seen counts and mark-all-as-seen

## What

## Why

## Screenshots / Gifs

## Checklist:


apps/services/user-notification/src/app/modules/notifications/dto/notification.dto.ts
```diff
@@ -81,6 +84,10 @@ export class RenderedNotificationDto {
   @IsUUID()
   messageId!: string
 
+  @ApiProperty({ example: 'Þjóðskrá' })
+  @IsString()
+  senderId!: string | null | undefined
```
kksteini, line 89: `!`
`undefined`

It's definitely a string (or null or undefined) :laughing: 

---

Isn't this a bit strange?

apps/services/user-notification/src/app/modules/notifications/notifications.service.ts
```diff
@@ -341,4 +341,48 @@ export class NotificationsService {
...
+      this.logger.error('Error getting unseen notifications count:', error)
+      throw new InternalServerErrorException('Error getting unseen notifications count')
+    }
+  }
+
+  async markAllAsSeen(user: User): Promise<void> {
+    try {
+      await this.notificationModel.update(
+        { seen: true },
+        { where: { recipient: user.nationalId, seen: false } }
```
kksteini, line 379: I'm not 100% certain how `CreateOptional` works but as `seen` is declared that way, is it then possible that `null` entries for `seen` exist?

Should this be something like `{ where: { recipient: user.nationalId, seen: { $not: true } }`? (I don't remember the ORM too well)

kksteini (approved):

apps/services/user-notification/src/app/modules/notifications/dto/notification.dto.ts
rafnarnason: fixed

thordurhhh (approved):
🙌

PR #13867, opened 2024-02-22
fix(auth-api-lib): add primary key to delegation index meta table

Herdismaria:
Use national_id column as primary key in delegation_index_meta table

## Checklist:


snaerseljan (approved):
Excellent!

PR #13877, opened 2024-02-22
feat(documents-v2): New documents client

thorkellmani:
## What

* New documents client
* Add new resolver to documents api domain for possible usage

## Why

Old client is way out of date and is proving a nuisance for the logs. 

## Screenshots / Gifs

## Checklist:


robertaandersen (approved):
LGTM

libs/clients/documents-v2/src/lib/documentsClientV2.service.ts
```diff
@@ -0,0 +1,179 @@
...
+        ? CustomersListDocumentsOrderEnum[input.order]
+        : undefined,
+      sortBy: input.sortBy
+        ? CustomersListDocumentsSortByEnum[input.sortBy]
+        : undefined,
+    })
+
+    if (!documents.totalCount) {
+      //throw maybe?
+      return null
```
thordurhhh, line 36: I think returning null would be ideal instead of throwing here.

thorkellmani: ok so leave it as is

thordurhhh (approved):
Love it 😻

libs/clients/documents-v2/src/lib/documentsClientV2.service.ts
saevarma: ~ Then remove the comment?

saevarma (approved):
Core files LGTM

libs/clients/documents-v2/src/lib/documentsClientV2.service.ts
thorkellmani: will do

PR #13883, opened 2024-02-22
fix(admin-portal): Fix environment changing on save

baering:
## What

[Issue](https://www.notion.so/e10da093139346a2ba34481d6c6c8df6?v=6da1b8b6dcb344e98178f4fba1480ae9&p=a8abfb1a5292489fab59246a4a7c8046&pm=s)

When updating a client in the ids-admin-portal. If you enter the edit screen with one environment selected (f.x. `production` initially), but make a change to the environment (to `staging`), and then save. The environment will fall back to the original environment on save.

## Why

To improve UX and not have to change the environment again.

## Checklist:


GunnlaugurG (approved):
LGTM!

PR #13912, opened 2024-02-26
fix(j-s): Only show appeal option for prosecutors and defenders

oddsson:
# Only show appeal option for prosecutors and defenders

[Asana](https://app.asana.com/0/1199153462262248/1205942829380443/f)

## What

Only show appeal option for prosecutors and defenders

## Why

Only these roles can withdraw an appeal

## Checklist:

### DIFF
--- a/apps/judicial-system/web/src/components/ContextMenu/ContextMenu.tsx
+++ b/apps/judicial-system/web/src/components/ContextMenu/ContextMenu.tsx
@@ -12,7 +12,10 @@ import {
   useBoxStyles,
 } from '@island.is/island-ui/core'
 import { TestSupport } from '@island.is/island-ui/utils'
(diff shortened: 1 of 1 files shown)


apps/judicial-system/web/src/components/ContextMenu/ContextMenu.tsx
```diff
@@ -31,6 +34,10 @@ export const useContextMenu = () => {
 
   const shouldDisplayWithdrawAppealOption = useCallback(
     (caseEntry: CaseListEntry) => {
+      if (!isProsecutionUser(user) && !isDefenceUser(user)) {
```
unakb, line 37: Consider just calling isProsecutionUser once and storing the result for reuse in line 44

PR #13922, opened 2024-02-27
fix(skilavottord): Applicants getting errors when deregistering vehicles

birkirkristmunds:
[TRS-8](https://dit-iceland.atlassian.net/browse/TRS-8)

## What

- Removed calling Samgongustofa SOAP & therefore no checks for locks, other owners etc.
- Added more parameters being sent from the application to the backend instead of the ones we got from the SOAP call

## Why
Some applicants are getting error that they don’t own the vehicle when trying to deregistering their vehicle.
The reason for that is the vehicle has an other owner.

## Checklist:


essgje (approved):
lftm

PR #13936, opened 2024-02-28
feat(ci): Add test-everything label

AndesKrrrrrrrrrrr:
Add a label to forcefully test everything

## Checklist:


lodmfjord (approved):
LGTM 

One comment. Now we have two ways to do the same thing. 

The magic env and this label.

I think this is nicer personally since it is more self explaining.

PR #13939, opened 2024-02-28
feat(j-s): Flag for users who can confirm appeals

oddsson:
# Flag for users who can confirm appeals

[Asana](https://app.asana.com/0/1199153462262248/1206633724454052/f)

## What

Implement a flag for users who can confirm appeals

## Why

This is needed for an feature we are building where certain users can electronically confirm appeals.

## Screenshots / Gifs

Screenshot 2024-02-28 at 10 04 41

image

## Checklist:


apps/judicial-system/web/src/routes/Shared/Cases/prisonCases.graphql
```diff
@@ -55,6 +56,7 @@ query PrisonCases {
       email
       role
       active
+      canConfirmAppeal
```
unakb, line 59: Does the prison need this info ?

apps/judicial-system/web/src/routes/CourtOfAppeal/AppealCase/appealCaseUsers.graphql
```diff
@@ -18,6 +18,7 @@ query AppealCaseUsers {
       active
     }
     active
+    canConfirmAppeal
```
unakb, line 21: Do we use this information here? Consider only adding it to the graphql queries where the field is used

apps/judicial-system/web/src/routes/Shared/Cases/prisonCases.graphql
unakb: Never mind, saw its a required field so ignore me :D

unakb (approved):
Nice, works locally :D

unakb (approved):
Looks good, tried locally and works

PR #13954, opened 2024-02-29
fix(auth): Hotfix delegation group by query

eirikurn:
## What

Hotfix access control system when user has custom delegation.

## Why

Regression


GunnlaugurG (approved):
LGTM

PR #13955, opened 2024-02-29
feat(service-portal): Display inbox document by id

thordurhhh:
## What

When visiting the postholf overview. You can pass in the id of the document in the url, it will open and display on the correct page.

## Why

New feature ✨ 

## Checklist:


libs/service-portal/documents/src/screens/Documents.loader.ts
```diff
@@ -0,0 +1,35 @@
...
+              pageSize: 10,
+              messageId: idParam,
+            },
+          },
+        })
+        return data?.getDocumentPageNumber.documentPage ?? 1
+      }
+      return 1
+    } catch {
+      return null
```
thorkellmani, line 33: return number?

libs/service-portal/documents/src/screens/Overview/Overview.tsx
```diff
@@ -51,14 +51,16 @@ import {
...
+import { useGetDocumentInboxLineQueryDetailsLazyQuery } from './DocumentDetail.generated'
 
 const pageSize = 10
 
 export const ServicePortalDocuments = () => {
   useNamespaces('sp.documents')
   const userInfo = useUserInfo()
   const { formatMessage } = useLocale()
-  const [page, setPage] = useState(1)
+  const loaderNumber = useLoaderData() as number
```
thorkellmani, line 62: scary

thorkellmani (changes requested):
Nitpicks

libs/service-portal/documents/src/screens/Overview/Overview.tsx
thordurhhh: Yeah, that's the go to solution for the useLoaderData with typescript. In this case not so scary as we only return a number always :D

thorkellmani (commented):
LGTM 🍼

PR #13962, opened 2024-02-29
feat(web): Chart improvements

baering:
## What

- Handle non-date x axis date (like a list of names)
- Be able to flip X and Y axis
- Offer `Custom Style Config` in Contentful to allow for more flexibility when displaying data

## Why

To allow for more uses cases and improve UX/DX.

## Screenshots / Gifs

Example of a non-date X axis and flipping X and Y axis with multiple segments in a bar:
image

### Flip axis
Normal bar chart:
image

Flipped:
image

### Custom styling
Example chart with default styling
image

But we get the request to:
- Show all the ticks on the X-axis
- Have the Y axis start at min data value and end at max data value

We can achieve that by putting the following values into the `Custom Style Config` field:

```typescript
{
  chart: {
    margin: {
      right: 35,
      top: 10,
    },
  },
  yAxis: {
    width: 50,
    domain: ['dataMin', 'dataMax'],
    tick: {
      dx: -5,
    },
  },
  xAxis: {
    height: 55,
    angle: 35,
    interval: 0,
    tick: {
      dy: 15,
      dx: 18,
    },
  },
}
```

image

## Checklist:


RunarVestmann (approved):

apps/web/components/Charts/v2/types/index.ts
```diff
@@ -52,3 +61,33 @@ export interface ComponentStyle {
   renderIndex: number
   renderIndexForType: number
 }
+
+export interface CustomStyleConfig {
```
saevarma, line 65: How is the documentation of this object in Contentful for developers/Content editors of the supported structure? I see the field in Contentful is a free form JSON?

saevarma (approved):
LGTM

apps/web/components/Charts/v2/types/index.ts
baering: Yeah I think I added (or meant to add) a "Please consult with a developer" hint below.

For now that's how it is. But this uses the same terms as the [Recharts API](https://recharts.org/en-US/api).

PR #13966, opened 2024-02-29
feat(delegation api): update delegation index when custom or personal representative delegations are updated

Herdismaria:
[Delegation index](https://www.notion.so/Delegation-Index-5aa41671d64d4031b276d45423b11741?pvs=4)
## What

Reindex custom delegations when user updates/creates/deletes them.
Reindex personal representative delegations when user updates/creates/deletes

## Checklist:


apps/services/auth/personal-representative/src/openApi.ts
```diff
@@ -3,7 +3,7 @@ import { DocumentBuilder } from '@nestjs/swagger'
 export const openApi = new DocumentBuilder()
   .setTitle('Personal Representative Service API')
   .setDescription(
-    'Service API for personal representativve. Only intended for the minstry of welfare.\nThe swagger document can be downloaded by appending `-json` to the last path segment.',
+    'Service API for personal representative. Only intended for the ministry of welfare.\nThe swagger document can be downloaded by appending `-json` to the last path segment.',
```
baering, line 6: 🕵️‍♀️

libs/auth-api-lib/src/lib/delegations/delegations-index.service.ts
```diff
@@ -99,27 +116,148 @@ export class DelegationsIndexService {
...
+            type: d.type,
           },
         }),
       ),
+      this.delegationIndexModel.destroy({
+        where: {
+          fromNationalId: deleted.map((d) => d.fromNationalId),
+          toNationalId: deleted.map((d) => d.toNationalId),
+          provider: deleted.map((d) => d.provider),
+          type: deleted.map((d) => d.type),
```
baering, line 152: Should this not be like above for updated? 🤔 

Will this work as intended? Trying to figure it out in my head.

Let's say deleted contains:

```typescript
const deleted = [
  { fromNationalId: '1001', toNationalId: '1002', /* ... */ },
  { fromNationalId: '1005', toNationalId: '1007', /* ... */ },
  { fromNationalId: '1010', toNationalId: '1012', /* ... */ },
  { fromNationalId: '1020', toNationalId: '1030', /* ... */ },
]
```

Will this destroy query all delegationIndices where any of this condition applies:
* fromNationalId is in ['1001', '1005', '1010', '1020']
* toNationalId is in ['1002', '1007', '1012', '1030']

So if there are indicies with these same ids that are not supposed to get deleted, won't there be a risk that they will? Maybe I'm wrong, but it feels weird.

baering (approved):

baering (approved):

PR #13977, opened 2024-03-01
fix(system-e2e): Fix tests to handle new login system

GunnlaugurG:
## What

Changed test locators and urls so tests will run against the new login system that will be out soon. The changes are backwards compatible so tests against the current login system will still work.

## Why

To be able to test against the new login system

## Screenshots / Gifs

## Checklist:


apps/system-e2e/src/tests/islandis/admin-portal/smoke/access-control.spec.ts
```diff
@@ -12,8 +12,10 @@ test.describe('Admin portal access control', () => {
     contextGranter = await session({
       browser,
       storageState: 'service-portal-faereyjar.json',
+      delegation: '65° ARTIC ehf.',
```
AndesKrrrrrrrrrrr, line 15: Note that this will break on `staging`, since the name is different. If you omit this parameter, we choose the first delegation available (which incidentally is always the `65...` delegation)

apps/system-e2e/src/support/login.ts
```diff
@@ -51,14 +53,17 @@ export async function idsLogin(
...
+    {
+      waitUntil: 'domcontentloaded',
+    },
+  )
 
   // Handle delegation on login
   if (page.url().startsWith(urls.authUrl)) {
     debug('Still on auth site')
-    const delegations = page.locator('button[name="SelectedNationalId"]')
+    const delegations = page.locator('.identity-card--name')
```
AndesKrrrrrrrrrrr, line 66: Could you change this to an accessible selector?

apps/system-e2e/src/tests/islandis/service-portal/smoke/auth.spec.ts
```diff
@@ -93,7 +94,7 @@ test('Service portal can use english sign-in', async ({ browser }) => {
   await page.goto(icelandicAndNoPopupUrl('/minarsidur'))
 
   // Act
-  await page.getByRole('link', { name: 'English' }).click()
+  await page.getByText('English').click()
```
AndesKrrrrrrrrrrr, line 97: Could you change this to an accessible selector?

AndesKrrrrrrrrrrr (approved):
Please use accessible selectors whenever possible. This ensures that our apps continue to be accessible for those who need it.

Great job on updating the tests :heart: 

Approving so you can fix at your own leisure.

PR #13987, opened 2024-03-04
feat(delegation-api): add custom delegation scope field to delegation index

Herdismaria:
[Delegation index pitch](https://www.notion.so/Delegation-Index-5aa41671d64d4031b276d45423b11741?pvs=4)

## What
Add scopes field to delegation index to keep track of api scopes for custom delegations.

## Why
To facilitate delegation look up.

## Checklist:


apps/services/auth/ids-api/src/app/delegations/test/delegation-index.spec.ts
```diff
@@ -306,6 +306,98 @@ describe('DelegationsIndexService', () => {
...
+        const fromNationalId = testcase.customDelegations[0].fromNationalId
+        const testDelegation = await delegationModel.findOne({
+          where: {
+            fromNationalId,
+            toNationalId: user.nationalId,
+          },
+        })
+
+        // Add remove scope from delegation
+        if (testDelegation) {
```
baering, line 323: Wondering why there is an if statement here and in line 366. Do you expect this to not exist at some point?

libs/auth-api-lib/src/lib/delegations/delegations-index.service.ts
```diff
@@ -24,6 +29,18 @@ type SortedDelegations = {
...
+const hasAllSameScopes = (a: string[] | undefined, b: string[] | undefined) => {
+  // Only custom delegations have scopes and they are never undefined
+  if (!a && !b) {
+    return true
+  }
+
+  if (a?.length !== b?.length) {
+    return false
+  }
+  return a?.every((s) => b?.includes(s))
```
baering, line 41: `=== true` so it doesn't return undefined in any case?

baering (approved):
LGTM ✅

apps/services/auth/ids-api/src/app/delegations/test/delegation-index.spec.ts
Herdismaria: Just to make typescript happy, `delegationModel.findOne` can return null 🤷‍♀️

libs/auth-api-lib/src/lib/delegations/delegations-index.service.ts
Herdismaria: Will fix 👍

PR #13997, opened 2024-03-04
fix(skilavottord): Check if registered owner is the current user on de-registration

birkirkristmunds:
[TRS-12](https://dit-iceland.atlassian.net/browse/TRS-12)

## What
Getting latest information about the current owner from Samgöngustofa and update database if needed.

## Why

Check if registered owner is the current user on de-registration.
If previous vehicle owner has sent in a car-recycling application and then sold the car to another person before taking the vehicle to de-registration, then before this fix he is the current owner in the database.

## Checklist:


essgje (approved):
lftm

PR #14000, opened 2024-03-04
feat(j-s): Deliver Indictment Case Indictment to Police

gudjong:
# Deliver Indictment Case Indictment to Police

[Skila ákæru  til LÖKE](https://app.asana.com/0/1199153462262248/1206483875888908/f)

## What

- Updates a police indictment case in LÖKE with the indictment in RVG when the case is received by a district court.
- If the case has uploaded indictment documents in RVG, then these documents are sent to LÖKE.
- Otherwise, an indictment document is generated and sent to LÖKE.

## Why

- Less work for the prosecution.

## Checklist:


unakb (approved):
Looks good to me

apps/judicial-system/backend/src/app/modules/notification/notification.service.ts
```diff
@@ -2410,15 +2401,6 @@ export class NotificationService {
             ]
           }
           break
-        case NotificationType.REVOKED:
```
unakb, line 2413: 🙏 Nice cleanup

PR #14017, opened 2024-03-05
fix(delegation-index): stop using master db when indexing all delegations

Herdismaria:
## Why

Indexing all delegations doesn't happen right after db write, so there is no need to use master database when indexing. 

## Checklist:


saevarma (approved):
💅

PR #14023, opened 2024-03-06
fix: show 404 instead of 500 erorr when regulation is not found

Valur:
https://island.is/reglugerdir/nr/0250-2025

When regulations are not found or the urls are invalid, show 404 error instead of 500

## Checklist:


thordurhhh (approved):
nice

PR #14030, opened 2024-03-06
fix(application-estate): Share-parsing function for icelandic commas

kksteini:
Invalid data was reaching the pre-filled state of the application due to `50,02` being cast into string in the DC's respsonse instead of passing `50.02` as float.

## Checklist:

---
  - https://app.asana.com/0/0/1206670473640411


RunarVestmann (approved):
Looks good to me

PR #14039, opened 2024-03-07
feat(delegation-api): new delegation type for personal representative

Herdismaria:
## What

Use new delegation type enum when indexing delegation. For personal representative delegations we include the personal representative `right` in the type name to facilitate lookup by scopes when retrieving delegations for nationalId.  This is the start of delegation refactor which is easy to introduce into the delegation index at this time.

## Checklist:


libs/auth-api-lib/src/lib/delegations/types/delegationIndexItem.ts
saevarma, line 1: This enum and type isn´t used in this PR?

libs/auth-api-lib/src/lib/delegations/delegations-index.service.ts
```diff
@@ -250,13 +262,38 @@ export class DelegationsIndexService {
...
+        const delegationsWithRights = d.reduce(
+          (acc: DelegationDTOWithStringType[], delegation) => {
+            if (!delegation.rights) {
+              return acc
+            }
+
+            const delegations = delegation.rights.map((right) => ({
+              ...delegation,
+              type: `${delegation.type}:${right.code}`,
+              rights: [right],
```
saevarma, line 276: Shouldn´t this be mapping to `DelegatoinDTOWithStringType`?
That is based on the `DelegationDto` which doesn't include the `rights` property. So maybe use `as` to force the type and remove the `rights`.

libs/auth-api-lib/src/lib/delegations/types/delegationIndexItem.ts
Herdismaria: Only the enum, but I just fixed that.

apps/services/auth/delegation-api/src/app/delegations/test/delegation-index/delegation-index.spec.ts
```diff
@@ -479,7 +467,56 @@ describe('DelegationsIndexService', () => {
...
+      // Add new delegation right to existing personal representation delegation
+      if (personalRepresentative) {
+        const rightType = await factory.createPersonalRepresentativeRightType({
+          code: prRight2,
+        })
+        await factory.createPersonalRepresentativeRight({
+          rightTypeCode: rightType.code,
+          personalRepresentativeId: personalRepresentative.id,
+        })
+      }
```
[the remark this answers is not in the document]
Herdismaria: Fixed the cleanup after test and remove the test for `prRight2`

saevarma (approved):
Great job!

PR #14066, opened 2024-03-08
feat(native-app): Ehic license

bjarnif:
Adding Ehic license support in the app wallet. The license is under a feature flag called "isEhicEnabled" under the mobile config in Configcat.

I used Gervimaður Ameríka to see data.

Simulator Screenshot - iPhone 15 Pro - 2024-03-08 at 10 06 28
Simulator Screenshot - iPhone 15 Pro - 2024-03-08 at 10 06 22

## Checklist:


snaerseljan (approved):
Looking good! Nice work

apps/native/app/src/screens/wallet/wallet.tsx
```diff
@@ -110,6 +111,7 @@ export const WalletScreen: NavigationFunctionComponent = ({ componentId }) => {
           GenericLicenseType.FirearmLicense,
           showDisability ? GenericLicenseType.DisabilityLicense : null,
           showPCard ? GenericLicenseType.PCard : null,
+          showEhic ? GenericLicenseType.Ehic : null,
         ].filter(Boolean) as GenericLicenseType[],
```
snaerseljan, line 115: ~I wrote a is Defined function here back in the day

https://github.com/island-is/island.is/blob/main/libs/shared/utils/src/lib/isDefined.ts

it is especially handy when you are filtering null or undefined values from list. That way you skip the TS casting
Usage is basically 

```
list.filter(isDefined)
```

PR #14081, opened 2024-03-08
fix(user-notifications): In memory cache fix

kksteini:
The template was being rewritten according to arguments and therefore the in-memory representation of the object would have its template string overwritten. This would/could result in the template being written to cache with argument replacements already in place.

## Checklist:

### DIFF
--- a/apps/services/user-notification/src/app/modules/notifications/notificationsWorker.service.ts
+++ b/apps/services/user-notification/src/app/modules/notifications/notificationsWorker.service.ts
@@ -212,7 +212,11 @@ export class NotificationsWorkerService implements OnApplicationBootstrap {
 
     const formattedTemplate = this.notificationsService.formatArguments(
       message.args,
-      template,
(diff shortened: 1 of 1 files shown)


saevarma (approved):
Nice comment addition as this is a really hidden gotcha 😅

stjanilofts (approved):

apps/services/user-notification/src/app/modules/notifications/notificationsWorker.service.ts
```diff
@@ -212,7 +212,11 @@ export class NotificationsWorkerService implements OnApplicationBootstrap {
 
     const formattedTemplate = this.notificationsService.formatArguments(
```
eirikurn, line 213: Imo should shallow copy inside this function instead. It already returns the formatted template and it's an antipattern to mutate function arguments. 🤷‍♂️

PR #14084, opened 2024-03-08
chore(logging): Delete and log more

AndesKrrrrrrrrrrr:
We're logging user/pass in some of these logs. Removing, when possible, and minor refactor for clearer logging (ensuring no `Object [object]` logs.

## Checklist:


apps/skilavottord/ws/src/app/modules/recyclingRequest/recyclingRequest.service.ts
```diff
@@ -89,11 +89,9 @@ export class RecyclingRequestService {
         )
       }
     } catch (err) {
-      throw new Error(
-        `Failed on deregistered vehicle ${vehiclePermno.slice(
-          -3,
-        )} because: ${err}`,
-      )
+      delete err.data
```
birkirkristmunds, line 92: What data are we deleting here? 
There is no data object in the error object, but there is a data in error.config, error.config.data

image

AndesKrrrrrrrrrrr: In one of these error logs, we are logging user/pass as the HTTP `--data` section. Updated the PR description

busla: I think there's also a generic error handler wrapper somewhere in `libs/shared`

birkirkristmunds (approved):
LGTM

PR #14113, opened 2024-03-12
feat(delegation-api): webhooks for delegation index

Herdismaria:
[Create webhook endpoint to reindex delegations](https://www.notion.so/Create-webhook-endpoint-to-reindex-delegations-071d08ff980f48ceae4ac6c9642be824?pvs=4)

## What

Create web hooks for third party services, that have delegations, to `create`/`update`/`delete` delegations from delegation index.

Side task: deprecate `DelegationType` and `DelegationProvider` enums from `delegation-api` and start using `AuthDelegationType` and `AuthDelegationProvider` from shared types.

## Why

To keep delegation index up to date when delegations are changed for e.g. National registry and company registry.

## Checklist:


libs/api/domains/auth/src/lib/resolvers/delegation.resolver.ts
```diff
@@ -22,12 +22,12 @@ import {
   PatchDelegationInput,
   UpdateDelegationInput,
 } from '../dto'
-import { Delegation } from '../models'
+import { Delegation, MergedDelegation } from '../models'
```
saevarma, line 25: We should fixed these import in the otherway, meaning we discourage the use of `index.ts` when importing inside a app/library and should import directly from the file. Importing from barrel files can cause weird hard to debug bugs.

libs/api/domains/auth/src/lib/resolvers/delegation.resolver.ts
```diff
@@ -22,12 +22,12 @@ import {
...
 import { ActorDelegationsService } from '../services/actorDelegations.service'
 import { ActorDelegationInput } from '../dto/actorDelegation.input'
 import { MeDelegationsService } from '../services/meDelegations.service'
-import type { DelegationDTO, MergedDelegationDTO } from '../services/types'
-import { MergedDelegation } from '../models/delegation.model'
+import type { DelegationDTO } from '@island.is/clients/auth/delegation-api'
+import type { MergedDelegationDTO } from '@island.is/clients/auth/public-api'
```
saevarma, line 30: ~ We prefer to have the import sorted into groups as

```
npm packages

monorepo module imports (@island.is/.... )

app/library relative imports

css imports.
```
But we currently need to manually do it, we need to add automatic ESLint/Prettier rules and formatting for this.

libs/auth-api-lib/src/lib/delegations/delegations-index.service.ts
```diff
@@ -170,6 +175,45 @@ export class DelegationsIndexService {
...
+  async createOrUpdateDelegationRecord(delegation: DelegationRecordInputDTO) {
+    const valid = validateDelegationTypeAndProvider(
+      delegation.type,
+      delegation.provider,
+    )
+
+    if (!valid) {
+      throw new Error('Invalid delegation type and provider combination')
+    }
+
```
saevarma, line 188: We should also verify that the nationalIds are valid and not the same value and that the `validTo` field, if provided, needs to be in the future.

apps/services/auth/delegation-api/src/app/delegations/delegation-index.controller.ts
```diff
@@ -0,0 +1,148 @@
...
+  @Documentation({
+    description: 'Create or update a delegation index record.',
+    response: { status: 200, type: DelegationRecordDTO },
+    request: requestDocumentation,
+  })
+  async createOrUpdateDelegationRecord(
+    @CurrentAuth() auth: Auth,
+    @Headers('X-Param-Id') delegationInfo: string,
+    @Body() body: CreateDelegationRecordInputDTO,
+  ) {
```
saevarma, line 82: Add explicit return type for readability 
```suggestion
  ): Promise<DelegationRecordDTO> {
```

apps/services/auth/delegation-api/src/app/delegations/delegation-index.controller.ts
```diff
@@ -0,0 +1,148 @@
...
+
+  @Put('.id')
+  @Documentation({
+    description: 'Create or update a delegation index record.',
+    response: { status: 200, type: DelegationRecordDTO },
+    request: requestDocumentation,
+  })
+  async createOrUpdateDelegationRecord(
+    @CurrentAuth() auth: Auth,
+    @Headers('X-Param-Id') delegationInfo: string,
```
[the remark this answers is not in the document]
Herdismaria: Nestjs Pipes don't have access to headers.

libs/api/domains/auth/src/lib/resolvers/mergedDelegation.resolver.ts
```diff
@@ -1,13 +1,13 @@
...
-import { DelegationType } from '@island.is/clients/auth/delegation-api'
 import { IdentityClientService } from '@island.is/clients/identity'
 import { Identity } from '@island.is/api/domains/identity'
 import { IdsUserGuard } from '@island.is/auth-nest-tools'
 
-import { MergedDelegation } from '../models'
-import type { MergedDelegationDTO } from '../services/types'
+import { MergedDelegation } from '../models/delegation.model'
+import { AuthDelegationType } from '@island.is/clients/auth/delegation-api'
+import type { MergedDelegationDTO } from '@island.is/clients/auth/public-api'
```
saevarma, line 10: ~ Import sort order

saevarma (approved):
🚀

PR #14150, opened 2024-03-15
feat(services-license): Add license type restriction to barcode

snaerseljan:
# License type restrictions for creating barcodes

We only want to allow specific types  of license to create barcode for them

## Checklist:


libs/api/domains/license-service/src/lib/licenseService.service.ts
```diff
@@ -525,8 +526,15 @@ export class LicenseServiceService {
   }
 
   async createBarcode(user: User, genericUserLicense: GenericUserLicense) {
+    const genericUserLicenseType = genericUserLicense.license.type
+
+    // Return null if the license type is not supported
+    if (!barcodeLicenseTypes.includes(genericUserLicenseType)) {
+      return null
+    }
```
thorkellmani, line 534: Don't we want to log something maybe?

thorkellmani (approved):
Cool, one question tho

libs/api/domains/license-service/src/lib/licenceService.type.ts
```diff
@@ -13,6 +13,18 @@ export enum GenericLicenseType {
...
+ * List of license types that can be scanned with a barcode scanner.
+ */
+export const barcodeLicenseTypes: GenericLicenseType[] = [
+  GenericLicenseType.DriversLicense,
+  GenericLicenseType.HuntingLicense,
+  GenericLicenseType.AdrLicense,
+  GenericLicenseType.MachineLicense,
+  GenericLicenseType.FirearmLicense,
+  GenericLicenseType.DisabilityLicense,
+]
```
eirikurn, line 26: ~The license clients already have similar logic in `clientSupportsPkPass` and `licenseIsValidForPkPass`. Would be nice to refactor to be consistent and not define this in multiple places.

snaerseljan: Good idea. I think I will use `clientSupportsPkPass` since that reflects the restrictions in barcodeLicenseTypes

brynjarorng (approved):
LGTM

eirikurn (approved):

PR #14151, opened 2024-03-15
fix(gbh): Extra residence check

obmagnusson:
## What

 - Added a  getCurrentResidence in the  nationalRegistry client lib.
 - Update the residency check dataprovider   
> - First it checks the current residence and then looks up the residence history.

## Why

This is due to the fact that the residency registrations in National Registry that where made before 1986 dont appear in the `/buseta` api endpoint. But it does so in the `/logheimili` endpoint.  

## Checklist:


saevarma (approved):
LGTM

Toti91 (approved):

PR #14161, opened 2024-03-15
feat(university-gateway): Parsing raw html to our design system

Ballioli:
# ...

## What

## Why

## Screenshots / Gifs
image

## Checklist:


libs/api/domains/university-gateway/src/lib/utils.ts
```diff
@@ -0,0 +1,18 @@
+import sanitizeHtml from 'sanitize-html'
+import { NodeHtmlMarkdown } from 'node-html-markdown'
+import { richTextFromMarkdown } from '@contentful/rich-text-from-markdown'
+
+export const convertHtmlToContentfulRichText = async (
+  html: string,
+  id?: string,
+) => {
+  const sanitizedHtml = sanitizeHtml(html)
```
RunarVestmann, line 9: I wonder id it'd make sense to add this function to a more shared place instead of copying it. But let's do that if this ends up being used more, rule of three 🙃

RunarVestmann (approved):

libs/api/domains/university-gateway/src/lib/utils.ts
Ballioli: Agree

PR #14179, opened 2024-03-19
fix(service-portal): Update single doc display

thordurhhh:
## What

Update single doc display

## Why

Bug in previous version


thorkellmani (approved):
cool

PR #14181, opened 2024-03-19
fix(scripts): Get secrets only "working" when debugging

AndesKrrrrrrrrrrr:
`yarn get-secrets my-project` was broken. Here we fix it.

The logger updates made all `console.log` use `logger.debug`. Now the logger uses `console.error` for all logs to send them all stderr, and the secrets rendering uses `console.log` (stdout) so we can redirect to the secrets file.

## Checklist:


brynjarorng (approved):
LGTM

PR #14185, opened 2024-03-19
feat(university-gateway): English titles for organizations, Using english titles in frontend of háskólanám

Ballioli:
Adding to the graphql the ability to fetch the english titles of organizations, updating frontend of haskolanam to use english tiles of universities

# ...

## What

## Why

## Screenshots / Gifs
image
image

## Checklist:


jonarnarbriem (approved):

Toti91 (approved):
Codeowned LGTM

apps/web/screens/UniversitySearch/LandingPage.tsx
```diff
@@ -59,13 +59,15 @@ const LandingPage: Screen<LandingPageProps> = ({
...
     const newArray = [...universities]
     newArray.sort((x, y) => {
-      const titleX = x.contentfulTitle || ''
-      const titleY = y.contentfulTitle || ''
-      return titleX.localeCompare(titleY)
+      const titleX =
+        locale === 'is' ? x.contentfulTitle || '' : x.contentfulTitleEn || ''
+      const titleY =
+        locale === 'is' ? y.contentfulTitle || '' : y.contentfulTitleEn || ''
+      return titleX.localeCompare(titleY, locale)
```
RunarVestmann, line 66: Perhaps using the sortAlpha function instead would help out with having "á" after "a" and so forth (not sure though if that even is a problem but just wanted to mention it

libs/cms/src/lib/cms.contentful.service.ts
```diff
@@ -216,6 +216,7 @@ export class CmsContentfulService {
 
   async getOrganizationTitles(
     organizationKeys: string[],
+    locale?: 'en' | 'is',
```
RunarVestmann, line 219: Isn't there a better type that could be used here instead of 'en' | 'is' ?

RunarVestmann (approved):

libs/cms/src/lib/cms.contentful.service.ts
Ballioli: Já mætti alveg vera type Locale = 'is' | 'en'

apps/web/screens/UniversitySearch/LandingPage.tsx
Ballioli: Has not been a problem for this case,  but I didn't know we had a sortAlpha in our utils, good to know!

PR #14191, opened 2024-03-19
feat(j-s): Implement confirmation on uploaded indictment pdfs

oddsson:
# Implement confirmation on uploaded indictment pdfs

[Asana](https://app.asana.com/0/1199153462262248/1206862815304548/f)

## What

In a [previous PR](https://github.com/island-is/island.is/pull/14130), we implemented a confirmation stamp on indictment PDF files that RVG creates, in traffic violation cases.

This PR does the same thing, but on indictment PDF files that are uploaded manually to RVG, in non-traffic violation cases. 

A major downside to this implementation is that we use two different PDF libraries, one for creating traffic violation pdf files and another one for stamping uploaded indictment pdfs. This means that we had to create the stamp twice, one for each lib. [This issue has been documented in Asana.](https://app.asana.com/0/1199153462262248/1206876804335379/f)

## Why

All indictments need to be confirmed by a "supervisor" user (Lögreglustjórinn á Höfuðboragarsvæðinu, f.x.). After confirmation we need a stamp with the credentials of the person who confirmed it.

## Screenshots / Gifs

Screenshot 2024-03-19 at 13 53 53

## Checklist:


libs/judicial-system/types/src/lib/case.ts
```diff
@@ -247,6 +249,24 @@ export const isCompletedCase = (state?: CaseState | null): boolean => {
   return Boolean(state && completedCaseStates.includes(state))
 }
 
+export const isTrafficViolationCase = (
```
unakb, line 252: Consider calling this function from the stepHelper instead of having the logic in 2 places, and then just add the extra CaseFileCategory.Indictment check in to the stepHelper function. Also consider renaming either one so it's clear what the difference is

unakb (approved):
Just one suggestion to consider, otherwise nice :D

libs/judicial-system/types/src/lib/case.ts
oddsson: Do you mean moving this function to the stepHelper and calling it from there? The stepHelper file is in under `web` so I can't call a function in there from the backend. Am I misunderstanding?

unakb: No I mean calling this function from the stepHelper and removing that logic from the isTrafficViolationCase there (and renaming it as well), so the stepHelper function would be something like 

`export const isTrafficViolationCaseCheck = (workingCase: Case): boolean => {
  const isTrafficViolation = isTrafficViolationCase(
    workingCase.indictmentSubtypes,
    workingCase.type as CaseType,
  )
  return Boolean(
    isTrafficViolation &&
      !(
        workingCase.caseFiles &&
        workingCase.caseFiles.find(
          (file) => file.category === CaseFileCategory.INDICTMENT,
        )
      ),
  )
}`
(except maybe with a better name)
My thinking is it might get confusing down the line to have 2 different functions that have the same name and do almost the same thing, instead of reusing the shared logic from the libs function and then adding the missing piece to the stepHelper one.

oddsson: Ah, gotcha ;) 

Onit..

PR #14195, opened 2024-03-19
fix: create-db.sh feature deploy script now supports readonly users

busla:
# ...

## What

## Why

## Screenshots / Gifs

## Checklist:

### DIFF
--- a/infra/scripts/container-scripts/create-db.sh
+++ b/infra/scripts/container-scripts/create-db.sh
@@ -1,39 +1,101 @@
+# This script creates or updates database users with appropriate permissions and passwords in PostgreSQL.
+# It also handles read-only and write users based on the naming convention and enables specified Postgres extensions.
+
 #!/bin/bash
 set -euo pipefail
 shopt -s inherit_errexit
 
+function create_or_update_user {
+  local user=$1
+  local password_key=$2
+  local db_name=$3
+  local is_read_only=$4
+
+  echo "Configuring user: $user"
+
+  # Attempt to fetch existing password
+  set +e # Disable script stopping for this command
+  local password=$(node secrets get "$password_key")
+  local fetch_status=$?
+  set -e # Re-enable script stopping on errors
+
+  local password_generated="false"
+  if [[ $fetch_status -ne 0 || -z "$password" ]]; then
+    echo "Generating new password for $user."
+    # Generate a new password if it doesn't exist in SSM
+    set +e
+    password=$(tr -dc A-Za-z0-9 </dev/urandom | head -c 27)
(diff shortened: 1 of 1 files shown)


infra/scripts/container-scripts/create-db.sh
```diff
@@ -1,39 +1,101 @@
...
+  local fetch_status=$?
+  set -e # Re-enable script stopping on errors
 
-psql -c "create user $DB_USER with encrypted password '$DB_PASSWORD'" || true
+  local password_generated="false"
+  if [[ $fetch_status -ne 0 || -z "$password" ]]; then
+    echo "Generating new password for $user."
+    # Generate a new password if it doesn't exist in SSM
+    set +e
+    password=$(tr -dc A-Za-z0-9 </dev/urandom | head -c 27)
```
brynjarorng, line 27: We should consider using `/dev/rand` since it is more secure and these are passwords. We might have to also consider the case that there is not enough randomness but I doubt it since this runs on the EKS nodes which should basically always have plenty of things to generate randomness from.

infra/scripts/container-scripts/create-db.sh
```diff
@@ -1,39 +1,101 @@
...
+   IF NOT EXISTS (
+      SELECT FROM pg_catalog.pg_roles
+      WHERE  rolname = '$user') THEN
 
-if node secrets store "$DB_PASSWORD_KEY" "$DB_PASSWORD"; then
-    psql -c "alter user $DB_USER with password '$DB_PASSWORD'"
-fi
+      CREATE ROLE $user LOGIN ENCRYPTED PASSWORD '$password';
+   ELSE
+      ALTER ROLE $user WITH ENCRYPTED PASSWORD '$password';
```
brynjarorng, line 45: This is risky to do, especially in a prod environment since it allows for a SQL injection. Would still like to discuss this or take a look at alternatives if possible

It should be "fine" on dev since devs already have access to secrets to do this. Still don't know if we can actually protect against this without some drastic changes in how the pipeline works.

brynjarorng (approved):
Two comments but otherwise LGTM

infra/scripts/container-scripts/create-db.sh
busla: never heard of  `/dev/rand` 🤷🏼 do you mean `/dev/random` ?

infra/scripts/container-scripts/create-db.sh
busla: we could do _some_ mitigation, use env vars instead of shell vars 🤷🏼

PR #14196, opened 2024-03-19
feat(user-profile): UserProfile migration script, add nextNudge

GunnlaugurG:
## What

Add next nudge to the user profile migration script.
Rules:
When `lastNudge` is set then `nextNudge` should be `6 months + lastNudge` 

## Why

Because of changes in the user-profile lookup where we now use nextNudge to determine if user should be prompted with a nudge.

## Screenshots / Gifs

## Checklist:


saevarma (approved):
🚀

PR #14208, opened 2024-03-20
chore(j-s): Table refactoring

oddsson:
# Table refactoring

[Asana](https://app.asana.com/0/1199153462262248/1206831212252825/f)

## What

In this PR, we are refactoring how we write tables in RVG. Before this PR we defined tables in separate components. We f.x. had one component for the table for active case, one for past cases, another for prison cases, etc.

This was problematic because, updating table styles or functionality was extremely inefficient, since you had to change multiple tables.

This PR implements a new Table component. This component takes in table data as props and can therefor be used everywhere. 

It also removes an unused table component from the project 👻

## Checklist:

## Summary by CodeRabbit

- **New Features**
	- Enhanced the Context Menu in the judicial system application with additional functionality and internationalization support.


apps/judicial-system/web/src/utils/hooks/useCaseList/index.tsx
```diff
@@ -11,6 +11,7 @@ import {
   DEFENDER_ROUTE,
 } from '@island.is/judicial-system/consts'
 import {
+  CaseTransition,
```
unakb, line 14: Please remove unused imports

apps/judicial-system/web/src/components/Table/Table.tsx
```diff
@@ -0,0 +1,228 @@
...
+      isSortable: boolean
+      key: sortableTableColumn
+    }
+  }[]
+  data: CaseListEntry[]
+  columns: { cell: (row: CaseListEntry) => ReactNode }[]
+  contextMenu?: {
+    menuItems: MenuItems
+  }
+  onDeleteCase?: (theCase: CaseListEntry) => void
```
unakb, line 43: Discussed offline as well, but I think we should try to keep the Table component as flexible as possible, so I think we need to avoid specific things like this. I had an idea of an alternative way where we can pass a function that generates the items in as a prop rather than the actual menu items. Alternatively, we could pass the entire ContextMenu in as a render function, I think just passing in a generation of menu items is enough though for now. Here's my thinking: https://github.com/island-is/island.is/commit/339c5491712e45b740a2711581d858fd6c710f1e

unakb:
Also we're missing the confirmation modal that should come before deleting a case

apps/judicial-system/web/src/components/ContextMenu/ContextMenu.tsx
```diff
@@ -36,6 +44,35 @@ interface ContextMenuProps {
...
+
+  const openCaseInNewTabMenuItem = (id: string): ContextMenuItem => {
+    return {
+      title: formatMessage(strings.openInNewTab),
+      onClick: () => handleOpenCase(id, true),
+      icon: 'open',
+    }
+  }
+
+  const deleteCaseMenuItem = (
```
unakb, line 59: Consider removing this as it's not used anymore

unakb (approved):
Just one comment, otherwise looks good! Lets get it to main so we can continue the refactoring when we have time 🙌

PR #14211, opened 2024-03-20
feat(native-app): Add loftbrú to app

thoreyjona:
# ...

## What

Add air discount to native app. Uses feature flag 'isAirDiscountEnabled' in ConfigCat.
I had no user with data for the Airfare usage table or with connection codes so I mocked that data when testing.

## Why

## Screenshots / Gifs
<div>

</div>

## Checklist:


apps/native/app/src/screens/air-discount/air-discount.tsx
```diff
@@ -0,0 +1,278 @@
+import {
+  Alert,
+  EmptyListSmall,
+  Heading,
+  Link,
+  LinkText,
+  Skeleton,
+  Typography,
+} from '@ui'
+import React from 'react'
```
snaerseljan, line 10: unused import

apps/native/app/src/screens/air-discount/air-discount.tsx
```diff
@@ -0,0 +1,278 @@
...
+      title: {
+        text: intl.formatMessage({ id: 'airDiscount.screenTitle' }),
+      },
+    },
+  }))
+
+const SkeletonItem = () => {
+  const theme = useTheme()
+  return (
+    <View>
```
snaerseljan, line 51: Is the View necessary?
I have not seen the convention in the project to have multiple components in one file. The SkeletonItem is only used once in this component so you could also just have it in the main component.

apps/native/app/src/screens/air-discount/air-discount.tsx
```diff
@@ -0,0 +1,278 @@
...
+        }}
+      />
+    </View>
+  )
+}
+
+const Empty = () => {
+  const theme = useTheme()
+  return (
+    <View style={{ marginBottom: theme.spacing[4] }}>
```
snaerseljan, line 76: ~same as above.

apps/native/app/src/screens/air-discount/air-discount.tsx
```diff
@@ -0,0 +1,278 @@
...
+              <FormattedMessage
+                id="airDiscount.tosLinkText"
+                defaultMessage="Notendaskilmálar"
+              />
+            </LinkText>
+          </Link>
+        </TOSLink>
+
+        {(loading || flightLegsLoading) && !error && <SkeletonItem />}
+        {!loading && !error && noRights && <Empty />}
```
snaerseljan, line 177: Shouldn't there be a error state? 
That way the rendering of the state could be easier, e.g. something like

```
{error ? <ErrorComp /> : loadingState ? <SkeletonItem/> : noRights ? <Empty /> : null}
```

apps/native/app/src/screens/air-discount/air-discount.tsx
```diff
@@ -0,0 +1,278 @@
...
+        {data && !noRights && (
+          <View>
+            <AlertWrapper>
+              <Alert
+                type="warning"
+                title={intl.formatMessage({ id: 'airDiscount.alertTitle' })}
+                message={intl.formatMessage({
+                  id: 'airDiscount.alertDescription',
+                })}
+                hasBorder={true}
```
snaerseljan, line 188: no need to pass true. hasBorder will evaluate to true

apps/native/app/src/screens/air-discount/air-discount.tsx
```diff
@@ -0,0 +1,278 @@
...
+            {data?.airDiscountSchemeDiscounts
+              ?.filter(
+                (discount) =>
+                  !(
+                    discount.user.fund?.used === 0 &&
+                    discount.user.fund.credit === 0
+                  ),
+              )
+              .map((discount) => {
+                return (
```
snaerseljan, line 207: ~nit: Redundant return in map. You can return without return statement.

```
 .map((discount) => (
    <AirDis....
)
```

apps/native/app/src/screens/air-discount/air-discount.tsx
```diff
@@ -0,0 +1,278 @@
...
+
+            <Typography weight="600" style={{ marginBottom: theme.spacing[1] }}>
+              <FormattedMessage
+                id="airDiscount.myRights"
+                defaultMessage="Mín réttindi"
+              />
+            </Typography>
+            {data?.airDiscountSchemeDiscounts
+              ?.filter(
+                (discount) =>
```
snaerseljan, line 200: ~nit: You can extract user from discount for shorter code or better yet to move filter into a separate function

apps/native/app/src/screens/air-discount/air-discount.tsx
```diff
@@ -0,0 +1,278 @@
...
+                  !(
+                    discount.user.fund?.used === 0 &&
+                    discount.user.fund.credit === 0
+                  ),
+              )
+              .map((discount) => {
+                return (
+                  <AirDiscountCard
+                    key={`loftbru-item-${discount.discountCode}`}
```
snaerseljan, line 209: ~nit: Same regarding destructuring of properties

apps/native/app/src/screens/air-discount/air-discount.tsx
```diff
@@ -0,0 +1,278 @@
...
+        {connectionCodes && connectionCodes.length > 0 && (
+          <View>
+            <Typography weight="600" style={{ marginBottom: theme.spacing[1] }}>
+              <FormattedMessage
+                id="airDiscount.activeConnectionCodes"
+                defaultMessage="Virkir afsláttarkóðar fyrir áframhaldandi flug"
+              />
+            </Typography>
+            {connectionCodes.map((discount) => {
+              return discount.connectionDiscountCodes.map((connectionCode) => {
```
snaerseljan, line 234: ~nit same regarding returns

apps/native/app/src/screens/air-discount/airfares-usage-table.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+import { Typography, dynamicColor } from '@ui'
+import { GetAirDiscountFlightLegsQuery } from '../../graphql/types/schema'
+
+const Host = styled.View`
+  margin-bottom: ${({ theme }) => theme.spacing[3]}px;
+`
+
+const FlightLeg = styled.View`
+  padding-top: ${({ theme }) => theme.spacing[2]}px;
```
snaerseljan, line 13: ~nit: You can also just do `padding` and then `padding-bottom` to override

apps/native/app/src/screens/air-discount/airfares-usage-table.tsx
```diff
@@ -0,0 +1,68 @@
...
+  flex-direction: row;
+  justify-content: space-between;
+  margin-bottom: ${({ theme }) => theme.spacing[1]}px;
+`
+
+interface AirDiscountProps {
+  data: GetAirDiscountFlightLegsQuery['airDiscountSchemeUserAndRelationsFlights']
+}
+
+export function AirfaresUsageTable({ data }: AirDiscountProps) {
```
snaerseljan, line 37: ~nit: In this monorepo it is encouraged to use function expressions instead

apps/native/app/src/screens/air-discount/airfares-usage-table.tsx
```diff
@@ -0,0 +1,68 @@
...
+  return (
+    <Host>
+      {data &&
+        data.map((flightLeg, index) => {
+          return (
```
snaerseljan, line 42: ~nit: Can skip the return

apps/native/app/src/screens/air-discount/airfares-usage-table.tsx
```diff
@@ -0,0 +1,68 @@
...
+                      value={flightLeg.flight.bookingDate}
+                      day="numeric"
+                      month="long"
+                    />
+                    {' - '}
+                    <FormattedTime value={flightLeg.flight.bookingDate} />
+                  </Typography>
+                )}
+              </Info>
+              <Typography variant="heading5" style={{ marginTop: 4 }}>
```
snaerseljan, line 60: ~nit: Can use theme.spacing.smallGutter instead of 4

apps/native/app/src/screens/more/more.tsx
```diff
@@ -130,6 +133,19 @@ export const MoreScreen: NavigationFunctionComponent = ({ componentId }) => {
               }
             />
           )}
+          {showAirDiscount && (
+            <ListButton
+              title={intl.formatMessage({ id: 'profile.airDiscount' })}
+              onPress={() => navigateTo(`/air-discount`)}
+              icon={
+                <Image
+                  source={airplaneIcon as any}
```
snaerseljan, line 142: I am pretty sure that any is not needed here. If you can skip then please remove the others as well :)

apps/native/app/src/ui/lib/card/air-discount-card.tsx
```diff
@@ -0,0 +1,154 @@
...
+  )};
+`
+
+const CopyCodeButton = styled.TouchableOpacity`
+  align-items: center;
+  justify-content: center;
+  padding: ${({ theme }) => theme.spacing[2]}px;
+`
+
+const ValidUntil = styled.Text`
```
snaerseljan, line 57: Can you use Typography component here?

apps/native/app/src/ui/lib/card/air-discount-card.tsx
```diff
@@ -0,0 +1,154 @@
...
+              marginBottom: theme.spacing[2],
+            }}
+          >
+            {discountCode}
+          </Typography>
+        </Right>
+      </Content>
+      {credit !== 0 && (
+        <CopyCodeContainer>
+          <CopyCodeButton onPress={() => Clipboard.setString(code ?? '')}>
```
snaerseljan, line 142: Should button be visible if code does not exist?

apps/native/app/src/ui/lib/empty-state/empty-list.tsx
```diff
@@ -48,3 +69,13 @@ export function EmptyList({ title, description, image }: HeadingProps) {
     </Host>
   )
 }
+
+export function EmptyListSmall({ title, description, image }: HeadingProps) {
```
snaerseljan, line 73: I would rather use size property here to the EmptyList and that component would determine if the list should be small or not

snaerseljan (commented):
Nice work 👏. Some comments and most of them nits. nits are usually things marked with tilda and are looked as an optional thing to do for the developer :)

apps/native/app/src/screens/air-discount/air-discount.tsx
thoreyjona: I was following the pattern I found [here](https://github.com/island-is/island.is/blob/main/apps/native/app/src/screens/vehicles/vehicles.tsx#L61)

apps/native/app/src/ui/lib/card/air-discount-card.tsx
thoreyjona: No, or at least did not find a way to make it work. The thing is that this `ValidUntil` text has different colors for light/dark theme than the regular texts. The Typography component can receive a color prop but does not support values on the format of dark/light mode.  So we would need to do a change in the Typography component for that to work and I decided that was out of scope for this, since it was only this one place.

apps/native/app/src/ui/lib/link/link-text.tsx
```diff
@@ -1,20 +1,24 @@
...
-  children: string
+  children: string | React.ReactNode
 }
 
 export function LinkText({ children }: LinkTextProps) {
-  return <Host>{children}</Host>
+  const theme = useTheme()
+  return (
+    <Host>
+      <Typography variant="heading5" color={theme.color.blue400}>
```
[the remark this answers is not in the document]
thoreyjona: Daði wanted to add this to all usages for this component in the app. (Only used in two other places than the one Loftbrú)

snaerseljan (approved):
Approving since my comments are nits  and should not be in the way :)

apps/native/app/src/screens/air-discount/air-discount.tsx
thoreyjona: And no the view is not necessary, I had some style on it that I then removed - good catch :)

apps/native/app/src/ui/lib/card/air-discount-card.tsx
thoreyjona: I was following the implementation from `AirDiscountOverview` and the only condition for it to be hidden is if credit is 0, see [here](https://github.com/island-is/island.is/blob/main/libs/service-portal/air-discount/src/screens/AirDiscountOverview/AirDiscountOverview.tsx#L202). 
But I thought about it though, can add if we want 😄

apps/native/app/src/screens/more/more.tsx
thoreyjona: Removed it and my editor was fine and looks the same when running the app, so removing it everywhere :)

PR #14212, opened 2024-03-20
feat(license-service): Adding expiration date to the resolver and updating logic

snaerseljan:
# Updates to the barcode logic

## What

- Added expiration date to the barcode field resolver. This is so the clients won't have to decode the tokens to get the expire date.
- Added error type to verify license error enum. Now we are validating on the server if the old JSON string format has expiration date or not.
- Added barcode type enum to the verify license resolver. Then the client can determine what is being done on the server, e.g. V2 or PKPass or unknown

## Why

This is all done for the clients to make their live easier.

## Checklist:


thorkellmani (approved):
LGTM

libs/api/domains/license-service/src/lib/licenseService.service.ts
```diff
@@ -567,7 +573,11 @@ export class LicenseServiceService {
       }),
     ])
 
-    return token
+    return {
+      token,
+      // Make sure to convert the expiration time to milliseconds
+      exp: new Date(exp * 1000),
```
eirikurn, line 579: Is the expiration value in nanoseconds? 🤔

libs/api/domains/license-service/src/lib/licenseService.service.ts
```diff
@@ -624,21 +634,51 @@ export class LicenseServiceService {
...
+
+    const parsedExpireTime = expires || date
+
+    // If the expiration date is in the past, the barcode is expired
+    if (parsedExpireTime && !isAfter(new Date(parsedExpireTime), new Date())) {
+      return {
+        barcodeType: VerifyLicenseBarcodeType.PK_PASS,
+        valid: false,
+        error: VerifyLicenseBarcodeError.EXPIRED,
+      }
```
eirikurn, line 681: Isn't the expiration already being checked elsewhere on the server? Weird to check twice if the pkpass data is expired.

If not, we need to add `licenseType` also to this response to support current app design.

libs/api/domains/license-service/src/lib/licenseService.service.ts
snaerseljan: Short answer, no the exp is in seconds because of the jwt lib.

Longer answer 😅 
Unfortunately the `jsonwebtoken` package does not support setting `exp` in milliseconds. I tried it without luck.
Here is a sample from their docs

```
// Signing a token with 1 hour of expiration:
jwt.sign({
  exp: Math.floor(Date.now() / 1000) + (60 * 60),
  data: 'foobar'
}, 'secret');
```

and this is how I am doing it to make sure the expiration date is exactly the same, e.g. what is used in the token and returned from the field resolver

```
  async createToken(data: LicenseTokenData): Promise<{
    token: string
    exp: number
  }> {
    const exp = Math.floor(Date.now() / 1000) + BARCODE_EXPIRE_TIME_IN_SEC

    return new Promise((resolve, reject) =>
      sign(
        {
          data,
          exp,
        },
        this.config.barcodeSecretKey,
        {},
        (err, encoded) => {
          if (err || !encoded) {
            return reject(err)
          }

          return resolve({
            token: encoded,
            exp,
          })
        },
      ),
    )
  }
```

However when seeing this I should rather multiply by 1000 within the func

libs/api/domains/license-service/src/lib/dto/VerifyLicenseBarcodeResult.dto.ts
```diff
@@ -8,7 +8,10 @@ import { GenericLicenseType } from '../licenceService.type'
 import { DriverLicenseData } from './licenses/DriverLicenseData.dto'
 
 export enum VerifyLicenseBarcodeError {
+  // When the token is expired
   TOKEN_EXPIRED = 'TOKEN_EXPIRED',
+  // When the license barcode is expired, e.g. if the JSON string data includes an expired expiration date
+  EXPIRED = 'EXPIRED',
```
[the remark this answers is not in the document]
snaerseljan: Good point!

libs/api/domains/license-service/src/lib/licenseService.service.ts
snaerseljan: No I checked the smartApi just throws the error if date is expired. 
image

snaerseljan: And here with the error check
image

snaerseljan: And now with the licenseType :)
image

libs/api/domains/license-service/src/lib/licenseService.service.ts
```diff
@@ -624,21 +634,51 @@ export class LicenseServiceService {
     data: string,
   ): Promise<VerifyLicenseBarcodeResult> {
     if (isJWT(data)) {
-      return this.getDataFromToken(data)
+      // Verify the barcode data as a token, e.g. new barcode format
+      const tokenData = await this.getDataFromToken(data)
```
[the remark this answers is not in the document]
snaerseljan: Yes good point, solved it with re verifying the token by ignoring the expiration date.

libs/api/domains/license-service/src/lib/licenseService.service.ts
eirikurn: Sure. I just turned this around in my head. 👍

eirikurn (approved):
Approved with minor comments

libs/api/domains/license-service/src/lib/licenseService.service.ts
```diff
@@ -590,9 +590,11 @@ export class LicenseServiceService {
       const payload = await this.barcodeService.verifyToken(token)
       code = payload.c
     } catch (error) {
-      this.logError(error)
+      this.logger.warn(error, {
```
eirikurn, line 593: ~Should review the other logError statements with same in mind.

eirikurn (approved):

PR #14225, opened 2024-03-21
fix(service-portal): document display single param

thordurhhh:
continue: https://github.com/island-is/island.is/pull/14223


thorkellmani (approved):
LGTM

PR #14227, opened 2024-03-21
fix(service-portal-documents): hotfix - single doc path

thordurhhh:
hotfix: https://github.com/island-is/island.is/pull/14225


thorkellmani (approved):
LGTM

PR #14284, opened 2024-03-25
fix: added error handling and removed logger print when fetching secrets

brynjarorng:
# ...

## What

Better logs for feature deployment create db and stop printing the secret into console

## Why

## Screenshots / Gifs

## Checklist:


infra/src/secrets.ts
```diff
@@ -77,7 +100,11 @@ yargs(hideBin(process.argv))
 
       const { Parameter } = await ssm.getParameter(parameterInput).promise()
       if (Parameter) {
-        return logger.info(Parameter.Value)
+        if (Parameter.Value && Parameter.Value.length > 0) {
+          console.log(Parameter.Value)
```
lodmfjord, line 104: Do we want to console log what comes out of parameter store?

busla (approved):

infra/src/secrets.ts
brynjarorng: In this case yes, If we use the logger, everything is always put into STD out which means that this export `a=$(node secrets get "$PGPASSWORD_KEY")` will ALWAYS print the output. Using console.log changes this behavior and will not print the secret

busla (changes requested):
.

infra/src/secrets.ts
busla: I dont follow. We need the value for the `create-db.sh` script

lodmfjord: thought this was something new was just checking the code

brynjarorng: Yes, the values is set in env in both cases. But using `console.log()` it is NOT printed in the console. If we use the logger, the env value is set AND it is always printed in plaintext to the console

lodmfjord: cool thanks for pointing out. I thought this was just debug log.

lodmfjord (approved):
LGTM

busla (approved):
LGTM

PR #14290, opened 2024-03-25
fix(j-s): Case ruling autofill

unakb:
[Böggur á þingbókarskjá](https://app.asana.com/0/1199153462262248/1206628179872951)

## What

Fix case ruling autofill logic

## Why

Weird things were happening when the autofill didn't trigger at the correct time, causing users to be unable to continue on to complete cases without jumping between steps first 

## Checklist:

## Summary by CodeRabbit

- **Refactor**
	- Simplified the `Overview` component in both Investigation and Restriction cases by removing unused logic and improving state management.
- **New Features**
	- Enhanced ruling logic in Investigation and Restriction cases to conditionally assign values based on specific criteria, improving decision-making processes.


gudjong (commented):
This PR moves the ruling autofill for extended cases to a new location, but it deletes the autofill for non-extended cases. Is that intentional?

busla:
@coderabbitai review

unakb:
> This PR moves the ruling autofill for extended cases to a new location, but it deletes the autofill for non-extended cases. Is that intentional?

Yes, because the same ruling autofill logic is executed during the initialization of the [Ruling](https://github.com/island-is/island.is/blob/86df324d2839cf27d7cec44b17e9d504e1648453/apps/judicial-system/web/src/routes/Court/InvestigationCase/Ruling/Ruling.tsx#L85) components, so I didn't see a reason for it to be performed here too. If there's some reason I didn't see from reading the code though definitely let me know 😊 I was just trying to lessen the repetition because the logic was in 4 components before, with this change it's only in 2.

gudjong (approved):

PR #14310, opened 2024-03-26
feat(user-profile): Migration Date verification

GunnlaugurG:
## What

Don't return email and phone number to third party clients when querying for user-profile that has last_nudge older then USER-PROFILE-MIGRATION-DATE. 
Show the email and phone number to first party clients that meet the same conditions but `isRestricted` property will be `true` so first party clients can decide by them selves what to do with the data.

## Why

So personal information will not go to third party clients when they have not seen the consent on the user-profile collection screen.

## Screenshots / Gifs

## Checklist:

## Summary by CodeRabbit

- **New Features**
    - Introduced the concept of client types (`FIRST_PARTY` and `THIRD_PARTY`) to better categorize users.
    - Added support for a new `isRestricted` property in user profiles to denote access restrictions.
    - Implemented dynamic client type filtering in user profile retrieval, enhancing customization based on client type.
- **Refactor**
    - Enhanced user profile service with improved configuration management, leveraging the `ConfigModule` for dynamic settings.
    - Updated email sending and worker service logic to use configurations, increasing flexibility and maintainability.
- **Tests**
    - Expanded testing to cover new scenarios involving different client types and the `isRestricted` user profile property.
- **Chores**
    - Adjusted mock data to include the `isRestricted` field, ensuring consistency in testing environments.


apps/services/user-profile/src/app/v2/user-profile.service.ts
```diff
@@ -16,10 +16,15 @@ import { UserProfileDto } from './dto/user-profile.dto'
...
 import { DataStatus } from '../user-profile/types/dataStatusTypes'
 import { NudgeType } from '../types/nudge-type'
+import { ClientType } from '../types/ClientType'
 
 export const NUDGE_INTERVAL = 6
 export const SKIP_INTERVAL = 1
 
+export const MIGRATION_DATE = new Date(
+  process.env.USER_PROFILE_MIGRATION_DATE ?? '2022-01-01',
+)
```
saevarma, line 26: To be consistent we should add `migrationDate` to `environment.ts` where we initialise it from the env variable and have dev fallback value but make the app startup fail in production if it is missing. So better yet would be to refactor the environment.ts setup to a Config and use `env.required` with dev fallback value.

apps/services/user-profile/src/app/v2/user-profile.service.ts
```diff
@@ -34,37 +39,48 @@ export class UserProfileService {
...
-        email: null,
-        mobilePhoneNumber: null,
-        locale: null,
-        mobilePhoneNumberVerified: false,
-        emailVerified: false,
-        documentNotifications: true,
-        needsNudge: null,
-        emailNotifications: true,
-      }
+      return this.filterByClientTypeAndRestrictionDate(
```
saevarma, line 50: Here we should just assign `userProfile` to the default profile object.

apps/services/user-profile/src/app/v2/user-profile.service.ts
```diff
@@ -34,37 +39,48 @@ export class UserProfileService {
...
-      email: userProfile.email,
-      mobilePhoneNumber: userProfile.mobilePhoneNumber,
-      locale: userProfile.locale,
-      mobilePhoneNumberVerified: userProfile.mobilePhoneNumberVerified,
-      emailVerified: userProfile.emailVerified,
-      documentNotifications: userProfile.documentNotifications,
-      needsNudge: this.checkNeedsNudge(userProfile),
-      emailNotifications: userProfile.emailNotifications,
-    }
+    return this.filterByClientTypeAndRestrictionDate(
```
saevarma, line 68: We should just pass the user profile object to the function and let it map to the dto object along with the property filtering.

saevarma (approved):
🚀

PR #14323, opened 2024-03-27
chore(island-ui): Input styling

oddsson:
# Input styling

[Asana](https://app.asana.com/0/1199153462262248/1206655083158870/f)

## What

Firefox applies a 0.54 opacity on placeholder texts in input fields. In this PR that opacity is set to 1 for consistency between browsers. [Source](https://stackoverflow.com/a/18416140)

## Why

This enforces styling consistency between browsers

## Checklist:

## Summary by CodeRabbit

- **Style**
	- Improved visibility of placeholder text in input fields on Firefox.


libs/island-ui/core/src/lib/Input/Input.mixins.ts
```diff
@@ -200,6 +200,9 @@ export const optionDescriptionSizes = {
 export const inputPlaceholder = {
   color: theme.color.dark300,
   fontWeight: theme.typography.light,
+
+  // Firefox applies 0.54 opacity to placeholder text. This disables that.
+  opacity: 1,
 }
 
 export const placeholder = {
```
RunarVestmann, line 208: I wonder if it'd also be beneficial to add the same styling on the placeholder class since it looks like it's being used for the Select component. Could you check if that is also having the firefox issue regarding opacity not being 1?

oddsson: Dropdown placeholders do not have this issue on Firefox. The LÖKE input at the top is a dropdown..

Justice League Silence of the lambs Lota 28 Screenshot Feb 23

PR #14327, opened 2024-03-27
fix(application-estate): Advocate Mapper

kksteini:
If there is no advocate recorded for a child (under 18) then we must attach an empty advocate object to the child in order for the frontend to pick up on that this should be rendered on first display.
The React frontend was populating the advocate object for an invisible member (on initial members) but those changes were not being shown until the user would update the array of members (by removing or appending).

## Checklist:

## Summary by CodeRabbit

- **New Features**
	- Enhanced estate templates to include conditional logic for members under the age of 18, improving data accuracy and user experience.

### DIFF
--- a/libs/application/template-api-modules/src/lib/modules/templates/estate/utils/mappers.ts
+++ b/libs/application/template-api-modules/src/lib/modules/templates/estate/utils/mappers.ts
@@ -8,6 +8,7 @@ import { estateSchema } from '@island.is/application/templates/estate'
 import { infer as zinfer } from 'zod'
 import { UploadData } from '../types'
 import { filterEmptyObjects } from './filters'
+import { info } from 'kennitala'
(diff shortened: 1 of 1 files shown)


stjanilofts (approved):

libs/application/template-api-modules/src/lib/modules/templates/estate/utils/mappers.ts
```diff
@@ -35,6 +36,8 @@ const estateMemberMapper = (element: EstateMember) => {
           phone: '',
           email: '',
         }
+      : info(element?.nationalId).age < 18
```
kksteini, line 39: For reference on the kennitala package in its current state.
`info(undefined).age < 18` is evaluated to `undefined < 18` → `false`

PR #14348, opened 2024-04-03
feat(j-s): Allow appeal when appeal decision in court is "not applicable"

unakb:
[Sækjandi/verjandi getur kært á kærufresti þó hann sé ekki viðstaddur í þinghaldi. ](https://app.asana.com/0/1199153462262248/1206876997460769)

## What

Allow defender and prosecutor to appeal a case if the appeal decision in court was "not applicable"

## Why

They should be able to send an appeal in these circumstances. 

## Checklist:

## Summary by CodeRabbit

- **New Features**
    - Enhanced the case overview functionality by adding the ability to determine the appealability of decisions for prosecutors and defenders.


libs/judicial-system/types/src/lib/case.ts
```diff
@@ -273,6 +273,16 @@ export const getStatementDeadline = (appealReceived: Date): string => {
   ).toISOString()
 }
 
+export const isAppealableDecision = (decision?: CaseAppealDecision | null) => {
```
gudjong, line 276: This implementation is fine as it is, but it would also be possible to add properties to the case in the api transformer. A property called `canBeAppealed` is already added there and instead of exposing this function, two new properties could be added `prosecutorCanAppeal` and `defenderCanAppeal`.

Both implementations are valid, but I feel adding new properties to the case would be more consistent with the existing implemention.

gudjong (approved):
There are some unused imports that should be removed.

I have one comment about the new exported function `isAppealableDecision`.

libs/judicial-system/types/src/lib/case.ts
unakb: True, I like that idea, I'm going to change to doing that for consistency

apps/judicial-system/api/src/app/modules/case/interceptors/case.transformer.ts
```diff
@@ -34,13 +34,14 @@ const getAppealInfo = (theCase: Case): AppealInfo => {
...
-        prosecutorAppealDecision === CaseAppealDecision.POSTPONE),
-  )
+  const hasBeenAppealed = Boolean(appealState)
 
-  appealInfo.hasBeenAppealed = Boolean(appealState)
+  appealInfo.canBeAppealed =
+    !hasBeenAppealed &&
+    (isAppealableDecision(accusedAppealDecision) ||
+      isAppealableDecision(prosecutorAppealDecision))
+
```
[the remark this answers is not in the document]
gudjong: Can you assist?

PR #14357, opened 2024-04-04
fix(service-portal): hotfix - mileage and type

thordurhhh:
hotfix:

https://github.com/island-is/island.is/pull/14356
https://github.com/island-is/island.is/pull/14354


Tryggvig (approved):
🏎️

PR #14376, opened 2024-04-05
feat(delegation-index): Add subject id to delegation index records

Herdismaria:
[Add subject_id column in the delegation index table](https://www.notion.so/Add-subject_id-column-in-the-delegation-index-table-7f1140d74f0b40df996cdd0a96a772e8?pvs=4)
[Update delegation index query endpoint to add subjectId property](https://www.notion.so/Update-delegation-index-query-endpoint-to-add-subjectId-property-91bc267c23e34cc88e4f9ffeeb656e73?pvs=4)

## What

Add subjectId column to delegation index. Find or create subjectId when indexing delegations.
Use correct name in emails to delegation holders

## Why

When sending email to delegation holders we want to include third party login link in the email so users can be logged in to the correct account.

## Checklist:

## Summary by CodeRabbit

- **New Features**
	- Enhanced delegation management with the ability to handle user identities and subject IDs more effectively.

- **Bug Fixes**
	- Improved sorting and indexing methods in delegation services to handle various delegation types and providers efficiently.

- **Chores**
	- Updated code ownership for the user-notification service to ensure proper team responsibilities.


libs/auth-api-lib/src/lib/delegations/dto/delegation-index.dto.ts
```diff
@@ -11,6 +11,11 @@ export class DelegationRecordDTO {
   @IsString()
   @ApiProperty()
   toNationalId!: string
+
+  @IsOptional()
+  @IsString()
+  @ApiProperty({ type: String, nullable: true })
+  subjectId?: string | null
```
saevarma, line 18: Is this ever optional on the delegation index record?
We are always going to make sure that this is also inserted with the record, by creating a new subject ID if is does not exists in the user identities.

libs/auth-api-lib/src/lib/delegations/models/delegation-index.model.ts
```diff
@@ -61,6 +61,12 @@ export class DelegationIndex extends Model<
   })
   customDelegationScopes?: string[]
 
+  @Column({
+    type: DataType.STRING,
+    allowNull: true,
+  })
+  subjectId?: string | null
```
saevarma, line 68: Should be required?

libs/auth-api-lib/migrations/20240327103605-add-subject-id-to-delegation-index.js
```diff
@@ -0,0 +1,25 @@
...
+
+module.exports = {
+  async up(queryInterface, Sequelize) {
+    return queryInterface.sequelize.transaction(async (transaction) => {
+      await queryInterface.addColumn(
+        'delegation_index',
+        'subject_id',
+        {
+          type: Sequelize.STRING,
+          allowNull: true,
```
saevarma, line 11: Should be required field.

We need to either truncate the current index table, or create the column with allow null, to some one time migration of populating the subject ID, then set the column to allowNull false.

saevarma (approved):
🚀

PR #14395, opened 2024-04-08
fix(native-app): more screen, fix Loftbrú out of view on small screens and not scrollable in view

thoreyjona:
## What
On smaller screens the Loftbú item in the More screen was sometimes out of view and not possible to scroll all of it into view

## Why

Added margin to the list items so it is possible to scroll all the item into view now. Did not want to resize everything so it would fit inside the viewport since we can have even smaller screens where it would still be an issue, instead of making it scrollable in view is better solution IMHO.

## Screenshots / Gifs
Before it was stuck like this (not scrollable fully into view):

Now it is possible to scroll it into view to be like this:

## Checklist:

## Summary by CodeRabbit

- **Style**
	- Updated the `More` screen layout with improved spacing for a cleaner user interface.


snaerth (approved):
Nice job

snaerseljan (approved):

PR #14416, opened 2024-04-09
feat(social-insurance-administration): call user profile in external data, if no email found send to IDS user profile page

karenbjorg:
## What

[JIRA 349](https://dit-iceland.atlassian.net/jira/software/c/projects/TS/boards/82?selectedIssue=TS-349)

Call user profile for email and phone number instead of getting that information from TR.
If no email found we show a warning alert message with ids link, applicant can register email there and is redirected back to the application.
Implemented for old-age pension, pension supplement, household supplement and additional support for the elderly. 

Use applicantInformationMultiField for displaying applicant information for oap, ps and asfte.

## Why

In the first implementation we were retrieving email and phone number from TR, and if the applicant didn't have email registed at TR we sent them to their My pages to register one because the email needed to be verified.

Now with this IDS user profile page, we can let applicant register email address to the user profile and send them back to the application.

## Screenshots / Gifs

Screenshot 2024-04-09 at 14 49 20

## Checklist:


libs/application/template-api-modules/src/lib/modules/shared/api/user-profile/user-profile.service.ts
```diff
@@ -47,6 +53,24 @@ export class UserProfileService extends BaseTemplateApiService {
...
           )
         }
+
+        if (params?.validateEmail && !results?.email) {
+          throw new TemplateApiError(
+            {
+              title: coreErrorMessages.noEmailFound,
+              summary: {
+                ...coreErrorMessages.noEmailFoundDescription,
+                defaultMessage:
```
obmagnusson, line 63: You are just working with the default message. If summary has a value in contentful this replace wont happen. You will want to pass it in with the  `values`  property. 
```
          summary: {
            ...coreErrorMessages.noEmailFoundDescription,,
            values: { link: this.getIDSLink(application) },
          },
```

libs/application/template-api-modules/src/lib/modules/shared/api/user-profile/user-profile.service.ts
```diff
@@ -67,4 +91,8 @@ export class UserProfileService extends BaseTemplateApiService {
         throw error
       })
   }
+
+  private getIDSLink(application: ApplicationWithAttachments) {
+    return `${this.idsClientConfig.issuer}/app/user-profile/email?state=update&returnUrl=${this.idsClientConfig.redirectUri}/ellilifeyrir/${application.id}`
```
obmagnusson, line 96: Should get the slug from the application type. This is a shared provider so it should work for every application.

libs/nest/config/src/lib/configurations/IdsClientConfig.ts
```diff
@@ -12,5 +12,9 @@ export const IdsClientConfig = defineConfig({
       '@island.is/clients/dev',
     ),
     clientSecret: env.required('IDENTITY_SERVER_CLIENT_SECRET'),
+    redirectUri: env.required(
```
saevarma, line 15: This should not be part of the `IdsClientConfig` object. You should add it to the `BaseTemplateAPIModuleConfig` as a 

```
userProfile: {
  redirectUri: string
}
```

then add it to `apps/application-system/api/src/environments/environment.ts`

saevarma (changes requested):

ylfahfa (approved):
lgtm 🙌

libs/application/template-api-modules/src/lib/modules/shared/api/user-profile/user-profile.service.ts
karenbjorg: ofc 🙈 will fix this

obmagnusson (approved):

saevarma (approved):
🚀

PR #14503, opened 2024-04-16
fix(estate): modify property number in needed before looking up

stjanilofts:
# Modify property number in needed before looking up

It seems that sometimes property numbers that come from the API are missing the L or F prefix.

This checks the length of the number, and
- if 6 numbers, prefixes with L
- if 7 numbers, prefixes with F

Edit: The real issue was the additional real estate repeater component was displaying a hidden version of the initial assets and when F/L was missing in the property number it would fail. It does not need to be there so this just skips rendering a hidden version of it.

## Checklist:

## Summary by CodeRabbit

- **Bug Fixes**
    - Updated property number handling in the estate module to ensure proper formatting by prefixing numbers with 'F' or 'L'.
    - Updated `assetNumber` values and descriptions for two assets in fake data generation.
- **Refactor**
    - Refactored the mapping of fields to include a conditional check for field.initial.
    - Added a conditional check within the fields.map function to handle field.initial before rendering the AdditionalRealEstate component.


HelgiHH (approved):

libs/application/templates/estate/src/fields/RealEstateRepeater/AdditionalRealEstate.tsx
```diff
@@ -65,8 +65,21 @@ export const AdditionalRealEstate = ({
...
+    if (propertyNumber && !isNaN(propertyNumber.charAt(0))) {
+      const str = String(propertyNumber).trim()
+      if (str.length === 7) {
+        propertyNumber = `F${str}`
+      }
+
+      if (str.length === 6) {
+        propertyNumber = `L${str}`
+      }
+    }
```
kksteini, line 81: I'm wondering, would it be simpler to just allow lookup with pure numbers in front instead of requiring `L` or `F`?

Modifying this here to have Ll and fF as optional?

https://github.com/island-is/island.is/blob/c0f8107bdf785d81c3b0635bab4c5fd0e9746d83/libs/application/templates/inheritance-report/src/lib/utils/helpers.ts#L82

libs/application/templates/estate/src/fields/RealEstateRepeater/index.tsx
```diff
@@ -111,17 +111,23 @@ export const RealEstateRepeater: FC<
...
-            remove={handleRemoveProperty}
-            index={index}
-            error={error && error[index] ? error[index] : null}
-          />
-        </Box>
-      ))}
+      {fields.map((field: AssetFormField, index) => {
+        if (field.initial) {
+          return null
+        }
```
kksteini, line 117: Do we need the "hidden" field below in that case?

kksteini (approved):

PR #14533, opened 2024-04-18
fix(api): Feature flag to disable national registry v1

eirikurn:
## What:
- Update all deprecated resolvers to go through api switcher.
  - Implemented queries used by the island.is app using v3 api.
  - Others either return null or throw an error if `api === 'v3'`
- Added feature flag to disable v1 api.

## Why:
- API v1 has being deprecated and has caused issues in production.
- When API v3 was implemented, it used an upgraded national registry graphql schema which supports both v1 and v3 apis.
- The older schema is still being used by the island.is app but only supports v1. This was a blocker to removing v1 completely.
- A separate PR will update the app to use the newer schema.
- We plan to keep the old schema running for some time to support older versions of the app. - By implementing v3 in the subset of the old schema used by the app, we can remove the v1 implementation from our backend without waiting for the app to be fully deployed.

## Summary by CodeRabbit

- **New Features**
  - Introduced feature flags for better control over API versioning.
  - Enhanced user and family member data retrieval with updated services.
  - Added legal residence handling and improved null value management.

- **Refactor**
  - Switched multiple resolvers to use `NationalRegistryService` for data fetching instead of `SoffiaService`.
  - Updated methods to handle new API versioning and feature flags.

- **Bug Fixes**
  - Addressed potential null values in user and child data retrieval methods.

- **Documentation**
  - Updated model fields to reflect optional status and compatibility with new API versions.


saevarma (approved):
🚀

PR #14590, opened 2024-04-23
fix(api): Public vehicle search - stop logging 404 as error

RunarVestmann:
# Public vehicle search - stop logging 404 as error

## What

* 404's today are being logged as errors, now they'll no longer be logged since a 404 result is quite common since it's a specific vehicle search after all

## Why

* To not raise awareness in logs for no reason

## Checklist:

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved error handling in the vehicle search feature to enhance reliability and user experience by updating error handling in the `getPublicVehicleSearch` method.


libs/api/domains/vehicles/src/lib/api-domains-vehicles.service.ts
```diff
@@ -142,10 +143,14 @@ export class VehiclesService {
...
-    })
-    return data
+    try {
+      const data = await this.publicVehiclesApi.publicVehicleSearchGet({
+        search,
+      })
+      return data
+    } catch (error) {
+      handle404(error)
+    }
```
thordurhhh, line 153: Nice 🙌  should probably just change to this though:

```suggestion
    return await this.publicVehiclesApi
      .publicVehicleSearchGet({
        search,
      })
      .catch(handle404)
```

So it will return other errors

thordurhhh (changes requested):
Just one minor thing

libs/api/domains/vehicles/src/lib/api-domains-vehicles.service.ts
RunarVestmann: Done

thordurhhh (approved):

PR #14620, opened 2024-04-26
feat(j-s): Postpone indictment hearing

gudjong:
# Postpone Indictment Hearing

[Margar fyrirtökur - næsta á dagskrá og fyrri fyrirtökur](https://app.asana.com/0/1199153462262248/1206610654815672/f)

## What

- Allows the court to postpone an indictment, either setting a new court date or postponing indefinitely with an explanation.
- Notifications are not included in this PR and are part of a different task.
- Info card for cases marked postponed indefinitely is not complete. They are implemented the same as for scheduled cases, but without a date. Instead there is a text indicating that the case has been postponed. A separate task will be created for implementing this when the design is ready.

## Why

- Part of new court workflow.

## Screenshots / Gifs

image

image

image

## Checklist:

- [x] I have performed a self-review of my own code
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] Formatting passes locally with my changes
- [x] I have rebased against main before asking for a review

## Summary by CodeRabbit

- **New Features**
  - Enhanced internationalization support for court case conclusion messages.
  - Introduced a new `Conclusion` component for managing final steps in court cases.

- **Enhancements**
  - Improved navigation and route names to align with judicial process stages.
  - Enhanced data handling for improved user information display.

- **Bug Fixes**
  - Cleaned up code by removing unnecessary console log statements.

- **Refactor**
  - Updated message IDs and default messages for better alignment with court procedures.

- **Style**
  - Improved code maintainability by adjusting import paths and component properties.

- **Documentation**
  - Updated component documentation to reflect new fields and functionalities related to case postponements.

- **Tests**
  - Adjusted test setups to accommodate changes in data models and routes.

PR #14648, opened 2024-04-30
chore(j-s): Get arraignmentDate for defenders

oddsson:
# Get arraignmentDate for defenders

[Asana](https://app.asana.com/0/1199153462262248/1207063877745563/f)

## What

Add `arraignmentDate` to `defendersCases` graphql. 

## Why

We use arraignmentDate to show either "RECEIVED" or "SCHEDULED" state in the list of cases

## Checklist:

## Summary by CodeRabbit

- **New Features**
	- Added court date information to the case listings for defenders.
	- Renamed the `courtDate` parameter to `scheduledDate` for improved case state mapping.
	- Removed unnecessary import statement for `CaseListEntry` in the ContextMenu component.


apps/judicial-system/web/src/components/TagCaseState/TagCaseState.tsx
```diff
@@ -27,7 +27,7 @@ export const mapCaseStateToTagVariant = (
   state?: CaseState | null,
   caseType?: CaseType | null,
   isValidToDateInThePast?: boolean | null,
-  courtDate?: string | null,
+  arraignmentDate?: string | null,
```
gudjong, line 30: This tag is also used when courtDate is set, so renaiming it is probably meaning less unless you name it neither arraignmentDate nor courtDate.

apps/judicial-system/web/src/routes/Defender/Cases/defenderCases.graphql
```diff
@@ -8,6 +8,9 @@ query DefenderCases($input: CaseListQueryInput) {
     courtCaseNumber
     validToDate
     policeCaseNumbers
+    arraignmentDate {
```
gudjong, line 11: The cases query only returns courtDate. It is calculated for each case as either the arraignmentDate or the courtDate. If there is a courtDate on the case, then that date is used, otherwise arraignmentDate is used.

Note that in addition, postponedIndefinitelyExplanation is returned with the cases query. This field is need when the case has been postponed indefinitely. The case is still scheduled (á dagskrá), but no date is displayed, just the text "Frestað".

gudjong (changes requested):
I think it would be best to wait until https://github.com/island-is/island.is/pull/14620 has been merged to main as that PR has some important addition related to this PR.

apps/judicial-system/web/src/components/TagCaseState/TagCaseState.tsx
oddsson: Ah, thanks. I've renamed this to scheduledDate, what do you think about that?

gudjong: Nice

gudjong (approved):

PR #14649, opened 2024-04-30
chore: Don't log `4xx` as `error`

AndesKrrrrrrrrrrr:
Why is this even a feature?

## Checklist:

## Summary by CodeRabbit

- **Configuration Updates**
  - Improved error handling by removing the `treat400ResponsesAsErrors` option from various API configurations. This change standardizes how 400 responses are processed across multiple modules, enhancing consistency and reliability in error management.


libs/clients/middlewares/src/lib/createEnhancedFetch.spec.ts
```diff
@@ -209,7 +209,7 @@ describe('EnhancedFetch', () => {
 
   it('can be configured to open circuit for 400 errors', async () => {
     // Arrange
-    env = setupTestEnv({ treat400ResponsesAsErrors: true })
```
saevarma, line 212: This test case needs to be removed completely as this will no longer be valid when we remove this feature

saevarma (changes requested):

libs/clients/middlewares/src/lib/createEnhancedFetch.spec.ts
AndesKrrrrrrrrrrr: Removed :heavy_check_mark:

saevarma (approved):
🚀
Merge this?

mannipje (approved):

libs/clients/icelandic-health-insurance/health-insurance/src/lib/clients-health-insurance-v2.module.ts
```diff
@@ -13,7 +13,6 @@ export class HealthInsuranceV2Client {
       fetchApi: createEnhancedFetch({
         name: 'clients-health-insurance',
         organizationSlug: 'sjukratryggingar',
-        treat400ResponsesAsErrors: true,
```
obmagnusson, line 16: This one is a little out of the ordinary as the health insurance api has some internal validation in the submit endpoint for the AccindentNotification application that returns a 412 on validation errors. So If we remove this the application can be moved to a success state without the actual application being submitted to Sjukratryggingar.

saevarma: Could that be handled by the client library integrating with the insurance api?

libs/clients/housing-benefit-calculator/src/lib/housing-benefit-calculator.provider.ts
```diff
@@ -18,7 +18,6 @@ export const ApiConfig = {
       fetchApi: createEnhancedFetch({
         name: 'clients-housing-benefit-calculator',
         logErrorResponseBody: true,
-        treat400ResponsesAsErrors: true,
```
RunarVestmann, line 21: I wonder if this'll break the try catch logic we have regarding receiving a 401 response

image

RunarVestmann: After glancing the code a bit and taking a look at how the public vehicle search works (that has never had the treat400ResponsesAsErrors set to true) I denote that this won't affect how we are handling different status codes via try catch blocks. I've also tagged Sævar who asked for our review to confirm it in a Slack thread.

image

RunarVestmann (approved):

libs/clients/icelandic-health-insurance/health-insurance/src/lib/clients-health-insurance-v2.module.ts
saevarma: @obmagnusson  4xx responses are still thrown even though this property is removed. The client generated code always throws for response code >= 400. This property was controlling log level and affecting the circuit breaker.

obmagnusson (approved):

PR #14736, opened 2024-05-08
fix(j-s): Set correct active step in received indictments

oddsson:
# Set correct active step in received indictments

[Asana](https://app.asana.com/0/1199153462262248/1206783638024031/f)

## What

Prosecutors should see "Málsmeðferð héraðsdóms" step in the form stepper as active when an indictment is in the state `RECEIVED`. 

## Why

When the indictment is received, there is no way for prosecutors to re-send it to the court. The only thing left to do is wait for a result from the court and therefore, this step should be active.

## Screenshots / Gifs

Screenshot 2024-05-08 at 11 23 08

## Checklist:

## Summary by CodeRabbit

- **New Features**
	- Enhanced the conditions under which sections are active for prosecution users, improving the relevance of displayed information based on case state.
  
- **Tests**
	- Implemented new tests to ensure accurate section visibility for indictment cases in specific states.


apps/judicial-system/web/src/utils/hooks/useSections/index.ts
```diff
@@ -408,6 +408,7 @@ const useSections = (
       isActive:
         isProsecutionUser(user) &&
         isIndictmentCase(type) &&
+        state !== CaseState.RECEIVED &&
```
gudjong, line 411: Consider also checking that `state !== CaseState.MAIN_HEARING` as that state can also occur during the court proceedings.

apps/judicial-system/web/src/utils/hooks/useSections/index.ts
```diff
@@ -899,8 +900,9 @@ const useSections = (
     return {
       name: formatMessage(sections.indictmentsCourtSection.title),
       isActive:
-        (isDistrictCourtUser(user) || isDefenceUser(user)) &&
-        !isCompletedCase(state),
+        (isProsecutionUser(user) && state === CaseState.RECEIVED) ||
```
gudjong, line 903: Consider renaiming this method to something other than `getIndictmentsCourtSections` as it is also used for prosecutors. Or is this a mistake?

Also consider checking that state === CaseState.MAIN_HEARING as that state can also occur during the court proceedings.

gudjong (approved):

apps/judicial-system/web/src/utils/hooks/useSections/index.ts
oddsson: It's not really used by prosecutors, per se. It's the court section, i.e. "Héraðsdómur RVG" f.x. Everyone sees this section but in this case we are deciding whether or not it's active or not.

PR #14757, opened 2024-05-10
feat(personal-representative): PR Delegation Type

GunnlaugurG:
## What

Added PersonalRepresentativeDelegationType to replace the PersonalRepresentativeRight table.

## Why

To support the incoming changes for delegation types where we allow dynamic adds on PersonalRepresentativesRightType or DelegationType (name of the new table)

## Screenshots / Gifs

## Checklist:

## Summary by CodeRabbit

- **New Features**
  - Introduced delegation types and providers for personal representatives.
  - Added support for creating and managing delegation types and providers within the app.

- **Bug Fixes**
  - Improved test cases to handle delegation types and providers more effectively.

- **Documentation**
  - Updated data transfer object (DTO) classes with new properties and validation decorators.

- **Refactor**
  - Changed types and added methods in models to support new delegation functionalities.
  - Reorganized and enhanced property validation in DTO classes.

- **Chores**
  - Added migration scripts to create and populate the `personal_representative_delegation_type` table.


libs/auth-api-lib/src/lib/personal-representative/models/personal-representative-delegation-type.model.ts
```diff
@@ -0,0 +1,74 @@
...
+  indexes: [
+    {
+      fields: ['personal_representative_id'],
+    },
+    {
+      fields: ['delegation_type_id'],
+    },
+  ],
+})
+export class PersonalRepresentativeDelegationTypeModel extends Model {
```
saevarma, line 27: Add type-safety definition to the model definition using the `InferAttribtues` and `InferCreationAttributes`.

libs/auth-api-lib/src/lib/personal-representative/services/personalRepresentative.service.ts
```diff
@@ -225,7 +258,9 @@ export class PersonalRepresentativeService {
       return await this.sequelize.transaction(async (t) => {
         const newPr = await this.personalRepresentativeModel.create(
           { ...personalRepresentative },
-          { transaction: t },
+          {
+            transaction: t,
```
saevarma, line 262: ~ You can also name the `t` param as `transaction` then simply do

```suggestion
            transaction,
```

libs/auth-api-lib/src/lib/personal-representative/services/personalRepresentative.service.ts
```diff
@@ -237,17 +272,53 @@ export class PersonalRepresentativeService {
...
+                  },
+                  { transaction: t },
+                )
+
+              return created.id
+            } catch (error) {
+              throw new BadRequestException(
+                `Error creating personal representative delegation type: ${error}`,
+              )
+            }
```
saevarma, line 300: This can be a misleading error classification and devops prefer structured error log object, meaning not inline string it in the message just add it as as a object param.

If the `rightTypeCode` does not exists in the `delegation_type` table, or the PR does not exists then this is correctly a users bad request. But if we have some random db failure (connection exhaustion, timeouts etc.) this should be Internal Server error, causing 500.

This catches everything and re-throws it as a 400 bad request. 

You could try checking if the error contains some foreign key reference error and then throw bad request other wise throw the same error.

```suggestion
            } catch (error) {
              if (error somehow contains foreign key ref error) {
                throw new BadRequestException(
                  'Error creating personal representative delegation type',
                  error
                )
              }
              else {
                 throw error
              }
            }
```

libs/services/auth/testing/src/fixtures/fixture-factory.ts
```diff
@@ -397,6 +430,22 @@ export class FixtureFactory {
...
+  async createPersonalRepresentativeDelegationType({
+    delegationTypeCode,
+    personalRepresentativeId,
+    id,
+  }: {
+    delegationTypeCode: string
+    personalRepresentativeId: string
+    id?: string | null
+  }) {
+    return await this.get(PersonalRepresentativeDelegationTypeModel).create({
```
saevarma, line 442: returning with await is redundant unless inside a try/catch. Then you need to await the result to catch if it would throw.

```suggestion
    return this.get(PersonalRepresentativeDelegationTypeModel).create({
```

saevarma (changes requested):

libs/auth-api-lib/src/lib/personal-representative/services/personalRepresentative.service.ts
```diff
@@ -237,17 +272,53 @@ export class PersonalRepresentativeService {
...
         const rightTypes =
           await this.personalRepresentativeRightTypeModel.findAll({
             where: { code: rightCodes.map((rc) => rc.rightTypeCode) },
           })
         result.rights = rightTypes.map((rt) => rt.toDTO())
+
+        const delegationTypes = await this.delegationTypeModel.findAll({
+          where: {
+            id: rightCodes.map(
+              (rc) => `PersonalRepresentative:${rc.rightTypeCode}`,
```
GunnlaugurG, line 316: use getPErsonalRepresentativeDelegationType here instead

libs/services/auth/testing/src/fixtures/fixture-factory.ts
```diff
@@ -399,6 +400,36 @@ export class FixtureFactory {
...
+      where: { id: 'talsmannagrunnur' },
+      defaults: {
+        id: 'talsmannagrunnur',
+        name: 'Talsmannagrunnur',
+        description: 'Talsmannagrunnur',
+        delegationTypes: [],
+      },
+    })
+
+    const [delegationType] = await this.get(DelegationTypeModel).findCreateFind(
```
Herdismaria, line 417: You could use the `createDelegationType` method below in this file

Herdismaria (approved):
Just some nit picks

saevarma (approved):
Nice job 🚀

PR #14763, opened 2024-05-10
fix(skilavottord): Check current owner on deregistration

birkirkristmunds:
# ...

[TRS-12](https://dit-iceland.atlassian.net/browse/TRS-12)

## What

Check if current owner is correctly registered in the Skilavottord DB when deregistering the vehicle

## Why

## Checklist:

### DIFF
--- a/apps/skilavottord/ws/src/app/services/icelandicTransportAuthority.services.ts
+++ b/apps/skilavottord/ws/src/app/services/icelandicTransportAuthority.services.ts
@@ -60,7 +60,10 @@ export class IcelandicTransportAuthorityServices {
     }
   }
 
-  async doGet(restURL: string, queryParams: { [key: string]: string }) {
+  async doGet(
+    restURL: string,
+    queryParams: { [key: string]: string } | undefined,
+  ) {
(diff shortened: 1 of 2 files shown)


apps/skilavottord/ws/src/app/services/icelandicTransportAuthority.services.ts
```diff
@@ -70,16 +73,20 @@ export class IcelandicTransportAuthorityServices {
...
+    let fullUrl = restURL
+
+    if (queryParams) {
+      // Convert the query parameters to a query string
+      const queryString = Object.entries(queryParams)
+        .map(
+          ([key, value]) =>
+            `${encodeURIComponent(key)}=${encodeURIComponent(value)}`,
+        )
+        .join('&')
```
helgifr, line 85: Can't this be done by doing `const queryString = new URLSearchParams(queryParams)` and then `queryString.toString()`?

birkirkristmunds: Fixed

helgifr (approved):

PR #14777, opened 2024-05-13
fix(native-app): don't show photo for driving license unless feature flag is on

thoreyjona:
## What

Don't show driving license photo unless feature flag `isBarcodeEnabled` is turned on

## Checklist:

## Summary by CodeRabbit

- **New Features**
	- Enhanced the Wallet Pass screen to selectively display barcode options based on the type of license.


apps/native/app/src/screens/wallet-pass/wallet-pass.tsx
```diff
@@ -349,6 +349,7 @@ export const WalletPassScreen: NavigationFunctionComponent<{
           nativeID={`license-${licenseType}_destination`}
           type={licenseType}
           logo={
+            isBarcodeEnabled &&
```
thoreyjona, line 352: @eirikurn @snaerseljan

1. I could not find any Gervimaður that has a photo for his driving license so I could not verify 100% that this works.

2. There is also a variable called allowLicenseBarcode that is the following:
```
const allowLicenseBarcode = isBarcodeEnabled && (data?.license?.pkpass && data?.license?.pkpassStatus === GenericUserLicensePkPassStatus.Available) && !data?.payload?.metadata?.expired
```

Just wanted to make sure that we don't want to use that, I only used the feature flag since that is was noted in the Notion task 😄

snaerseljan: Yes this is sufficient :)

snaerseljan (approved):
Nice one 👍

PR #14793, opened 2024-05-14
fix(island-ui): refactor Input stories

albinagu:
Current: 
https://ui.devland.is/?path=/docs/form-input--docs

Updated:
Screenshot 2024-05-14 at 15 09 16
Screenshot 2024-05-14 at 13 19 00
Screenshot 2024-05-14 at 13 18 13

## Checklist:

## Summary by CodeRabbit

- **New Features**
  - Enhanced Input component with various new configurations and examples, including different sizes, background colors, and states (disabled, read-only, with error, etc.).

- **Style**
  - Updated Input component styles to feature transparent borders and 100% height.
  - Improved error state styling by adding a red background color.

- **Documentation**
  - Enriched storybook documentation for the Input component with new examples and configurations.


libs/island-ui/core/src/lib/Input/Input.stories.tsx
```diff
@@ -2,148 +2,149 @@ import React from 'react'
 
 import { withFigma } from '../../utils/withFigma'
 import { Input } from './Input'
+import { size } from 'lodash'
+import { ro } from 'date-fns/locale'
```
oddsson, line 6: I think this is an unused imports, right?

libs/island-ui/core/src/lib/Select/Select.css.ts
```diff
@@ -258,13 +258,13 @@ export const optionDescriptionTruncated = style({
 export const optionSizes = styleVariants({
   xs: wrapMedia(inputMixins.inputSizes.xs, `${wrapper} &`),
   sm: wrapMedia(inputMixins.inputSizes.sm, `${wrapper} &`),
-  md: wrapMedia(inputMixins.inputSizes.md, `${wrapper} &`),
```
oddsson, line 261: Could you please explain this change? Seems kind of weird that `sm` and `md` are the same.

oddsson (changes requested):
Nice, just a couple of questions.

libs/island-ui/core/src/lib/Select/Select.css.ts
albinagu: right, the design' been changed for select dropdown options to be 18px instead of 24px as it looks pretty big. I've changed it here to test it out. Not sure if i should keep this change or make a different approach?

oddsson: Ah, I get it now. I think this approach is fine then..

oddsson (approved):

PR #14798, opened 2024-05-14
fix(native-app): Fix token refresh to logout

snaerseljan:
# Fix app startup and token check to not fail to login screen

## Checklist:

## Summary by CodeRabbit

- **Refactor**
  - Renamed the `refresh` function to `authRefresh` for clarity.
  - Updated error handling in several functions, introducing new constants for specific error cases.

- **Bug Fixes**
  - Improved token expiration and refresh logic in the `fetchUserInfo` function.
  - Enhanced error handling and conditional checks in the `checkIsAuthenticated` function.

- **New Features**
  - Added logic to handle logout based on specific error conditions.


apps/native/app/src/stores/auth-store.ts
```diff
@@ -207,12 +220,18 @@ export async function checkIsAuthenticated() {
...
       return false
     }
   }
 
-  fetchUserInfo().catch(async (err) => {
+  try {
+    await fetchUserInfo()
+
+    return true
+  } catch (e) {
```
eirikurn, line 234: This should mainly check if `fetchUserInfo()` fails due to `refresh()` failing with a 401 response. Anything else should be assumed as `authenticated === true`.

To test 401 refresh errors, try overriding the refreshToken with some garbage.

apps/native/app/src/stores/auth-store.ts
```diff
@@ -62,54 +62,67 @@ export const authStore = create<AuthStore>((set, get) => ({
...
   cognitoAuthUrl: undefined,
   cookies: '',
-  async fetchUserInfo(_refresh = false) {
+  async fetchUserInfo(shouldRefresh = false) {
     const appAuthConfig = getAppAuthConfig()
     // Detect expired token
     const expiresAt = get().authorizeResult?.accessTokenExpirationDate ?? 0
+
     if (new Date(expiresAt) < new Date()) {
       await get().refresh()
```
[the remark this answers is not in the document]
snaerseljan: Yes hold your horses the PR was not ready yesterday. I will remember to set to draft next time

apps/native/app/src/stores/auth-store.ts
```diff
@@ -62,54 +62,67 @@ export const authStore = create<AuthStore>((set, get) => ({
...
-        return userInfo
+    )
+
+    if (res.status === 401) {
+      // Attempt to refresh the access token
+      const successfulRefresh = await get().refresh()
+
+      if (!shouldRefresh && successfulRefresh) {
+        // Retry the userInfo call
+        return get().fetchUserInfo(true)
```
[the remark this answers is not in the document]
snaerseljan: What do you mean my change. I only changed the parameter name. The logic did not change at all?
Rename the variable definitely.
What extra refresh are you talking about?

apps/native/app/src/stores/auth-store.ts
snaerseljan: Yes good point I added offline check as well

snaerseljan:
@eirikurn Now the PR should be ready . There where a few gotchas

Here are a few points.

- config environment was flickering from prod to dev because of logic error in environmentStore. This led to that the idsissuer in dev was sometime the prod url, but most of the time the dev url. That led to failed userinfo call.
- Single session client was also a gotcha since the refreshToken errors invalid_grant which is a global error for wrong refreshtoken. Me and  Þórey where using the same user so that introduced the error a lot.

apps/native/app/src/stores/auth-store.ts
```diff
@@ -62,54 +65,73 @@ export const authStore = create<AuthStore>((set, get) => ({
...
-        const userInfo = await res.json()
-        set({ userInfo })
-        return userInfo
+    )
+
+    if (res.status === 401) {
+      // Attempt to refresh the access token
+      const successfulRefresh = await get().refresh()
+
+      if (!skipRefresh && successfulRefresh) {
```
eirikurn, line 90: The difference from before has to do with [short circuit behaviour](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_AND#short-circuit_evaluation) in `&&`.

Before it only called refresh if `_refresh === false`. Now it calls and waits for a refresh even though `skipRefresh === false` and the actual refresh result is ignored.

apps/native/app/src/stores/auth-store.ts
```diff
@@ -166,27 +188,25 @@ export const authStore = create<AuthStore>((set, get) => ({
...
 export const useAuthStore = createUse(authStore)
 
-export async function readAuthorizeResult(): Promise<AuthorizeResult | null> {
+export async function readAuthorizeResult(): Promise<void> {
   const { authorizeResult } = authStore.getState()
 
-  if (authorizeResult) {
-    return authorizeResult as AuthorizeResult
-  }
+  if (authorizeResult) return
```
eirikurn, line 194: ~Not a fan of skipping braces as it has historically caused hard to spot bugs. Though code standard/eslint/prettier should cover this.

apps/native/app/src/stores/auth-store.ts
```diff
@@ -207,23 +227,37 @@ export async function checkIsAuthenticated() {
...
+    await fetchUserInfo()
+
+    return true
+  } catch (e) {
+    const err = e as Error
 
-  return true
+    if (!offlineStore.getState().isConnected) {
+      return true
+    } else if (REFRESH_TOKEN_ERROR === err?.message) {
```
eirikurn, line 246: ~Instead of using offline store, which can have false positives or false negatives, I would prefer to actually handle different refresh responses propertly:

- actual "expired/invalid refresh token" errors should log the user out (401, invalid_grant, whatever).
- other refresh errors can be ignored (no internet, server down, 500, whatever).

eirikurn (commented):
Let me know if you want to merge as is.

apps/native/app/src/stores/auth-store.ts
snaerseljan: Yeah good point! I will update

apps/native/app/src/stores/auth-store.ts
snaerseljan: Yeah man that just like your opinion man :)
https://www.youtube.com/watch?v=XkADiJCYS2k

apps/native/app/src/stores/environment-store.ts
```diff
@@ -48,7 +48,7 @@ export interface EnvironmentStore extends State {
 export const environmentStore = create<EnvironmentStore>(
   persist(
     (set, get) => ({
-      environment: environments.prod,
+      environment: config.isTestingApp ? environments.dev : environments.prod,
```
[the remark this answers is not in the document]
snaerseljan: This is doing that. 🤔 I am confused.

The reason for the change as explained above was when the app restarted or started up again, the config was for brief time was production and then dev environments. That resulted in that the execution of `checkIsAuthenticated` which fetches userinfo with token from dev but idsIssuer url for production, which led to a logout.

Or maybe I am misunderstanding you completely 🤷‍♂️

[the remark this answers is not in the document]
eirikurn: Sorry, I mixed up the sides of the diff and though you were changing it to be always prod. Good change 👍

eirikurn (approved):

andes-it:

PR #14834, opened 2024-05-16
fix(island-ui): Select stories with TypeScript

albinagu:
## Checklist:


libs/island-ui/core/src/lib/Select/Select.tsx
```diff
@@ -4,6 +4,7 @@ import cn from 'classnames'
 // @ts-ignore make web strict
 import ReactSelect, { createFilter, GroupBase } from 'react-select'
 import CreatableReactSelect from 'react-select/creatable'
+import { Props } from 'react-select'
```
oddsson, line 7:
```suggestion
```

I believe this is not used anywhere

oddsson (approved):

PR #14839, opened 2024-05-16
feat(j-s): JWT verification for digital-mailbox-api

oddsson:
# JWT verification for digital-mailbox-api

[Asana](https://app.asana.com/0/1199153462262248/1207323461913719/f)

## What

Verify JWT sent with requests to digital-mailbox-api. 

## Why

So that only people authenticated on island.is can call this api.

## Checklist:

## Summary by CodeRabbit

- **New Features**
  - Enhanced authentication configuration for better security and flexibility.
  - Added token verification for user authentication.

- **Improvements**
  - Updated authentication settings including scope, client ID, issuer URL, and redirect URIs.
  - Improved connection testing with token verification.

- **Bug Fixes**
  - Fixed issues related to authentication bypass and backend URL configuration.


apps/judicial-system/digital-mailbox-api/src/app/app.config.ts
```diff
@@ -1,15 +1,28 @@
 import { defineConfig } from '@island.is/nest/config'
 
-export default defineConfig({
-  name: 'AppModule',
+export const authModuleConfig = defineConfig({
+  name: 'AuthModule',
```
unakb, line 4: I think we should name this something other than AuthModule (and AppModule to be fair wasn't the best name either). Maybe just DigitalMailboxModule or something like that

apps/judicial-system/digital-mailbox-api/src/app/app.service.ts
```diff
@@ -21,8 +24,54 @@ export class AppService {
...
+    const verifiedUserToken = await this.verifyIdsToken(token)
+
+    if (!verifiedUserToken.sub) {
+      throw new UnauthorizedException('Invalid token')
+    }
+
     return this.test()
   }
+
+  async verifyIdsToken(token: string) {
```
unakb, line 38: I think we should move this to a guard so it's easier to use for all the endpoints

unakb: We could start with something along the lines of this 
https://github.com/island-is/island.is/commit/bfb45f65bd6f01f957cc222108066b8f2f6e9c79

I don't know whether we want to have this in our lib or in this project, but could be a start to just have it here

unakb: I also realized we can use the passport library to handle the verification, a draft of the idea :
https://github.com/island-is/island.is/pull/14882

Should we start with something like this? It also makes it easier to extract the national id to verify that the user can access cases only connected to their national id

oddsson:
Closed in favour of https://github.com/island-is/island.is/pull/14882

PR #14842, opened 2024-05-16
feat(j-s): Store MD5 Sum for Confirmed Indictments

gudjong:
# Store MD5 Sum for Confirmed Indictments

[Geyma MD5 check sum fyrir ákærur](https://app.asana.com/0/1199153462262248/1207234344573824/f)
Skila ákæru pdf í umferðalagabrotum til Auðar](https://app.asana.com/0/1199153462262248/1206746352017705/f)

## What

- Moves INDICTMENT_CONFIRMED event log to the case service to guarantee that it succeeds if and only if the SUBMIT transition succeeds.
- Lazily confirms indictments on first access after indictment confirmation.
- Rewrites AWS S3 key handling, deligating key prefix logic to the AWS S3 Service.
- Stores the MD5 hash for confirmed indictments.

## Why

- The main reason for this PR is the MD5 hash, which can be used for later verification of indictment confirmation.

## Checklist:


apps/judicial-system/backend/src/app/modules/aws-s3/awsS3.service.ts
```diff
@@ -2,29 +2,63 @@ import { S3 } from 'aws-sdk'
...
+  `${completedIndictmentPrefix}${key}`
+
+const formatS3Key = (caseType: CaseType, caseState: CaseState, key: string) =>
+  isIndictmentCase(caseType)
+    ? isCompletedCase(caseState)
+      ? formatS3CompletedIndictmentKey(key)
+      : formatS3IndictmentKey(key)
+    : formatS3RequestKey(key)
+
+const formatConfirmedKey = (key: string) =>
```
oddsson, line 34: Super nitpick -- Consider moving this function to line nr 26, so that all the `format` functions are in the same place.

oddsson (approved):

PR #14868, opened 2024-05-21
chore: make sure to clear apolloMMKVStorage on all logouts

thoreyjona:
Make sure to clear apolloMMKVStorage on all logouts, both intentional (when user logs out himself) and unintentional (e.g. when the user logs into the same user in another device we log the old one out automatically).

## Checklist:

## Summary by CodeRabbit

- **Refactor**
  - Improved logout process to clear all storages for better data management and security.
  - Removed unnecessary function call from `mmkv` store.
  - Added new import statements for improved functionality in `auth-store.ts`.
  - Replaced `console.log` with `console.error` for error logging in `notifications-store.ts`.


apps/native/app/src/stores/auth-store.ts
```diff
@@ -136,6 +137,7 @@ export const authStore = create<AuthStore>((set, get) => ({
     return false
   },
   async logout() {
+    apolloMKKVStorage.clearStore()
     const appAuthConfig = getAppAuthConfig()
     // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
     const tokenToRevoke = get().authorizeResult!.accessToken!
```
thoreyjona, line 143: @eirikurn the ticket in notion talked about making sure that Zustand state is reset. The zustand state for authStore is reset here a few lines below (see [here](https://github.com/island-is/island.is/blob/5d48548fa8d670eb53dca1df6923a9e86d68e8bf/apps/native/app/src/stores/auth-store.ts#L157) where userInfo and authorizeResult are set to undefined. Is the notion ticket talking about offline-store maybe? I don't see that that is currently reset anywhere in the code on logout - is that what you were talking about? Sorry don't have a lot of context here 😄

eirikurn: Good question. Clearing mmkv + apollo cache should handle the offline store. @snaerseljan was there anything else? 🤔

eirikurn (approved):
LGTM

snaerth:
@eirikurn No just when we start using MMKV in other stores, we need to make sure to define the reset function on those stores on logout, since the Zustand state will not clear its values.

PR #14877, opened 2024-05-22
chore: update all ioredis packages

lodmfjord:
Updated ioredis and all packages that depend on ioredis.

Removed patch for cache-manager-ioredis-yet

## Summary by CodeRabbit

- **Chores**
  - Updated dependencies to ensure better performance and security:
    - `@keyv/redis` to version `2.8.5`
    - `bull` to version `4.12.8`
    - `cache-manager-ioredis-yet` to version `2.0.4`
    - `ioredis` to version `5.4.1`


package.json
```diff
@@ -139,9 +139,9 @@
     "aws4": "1.11.0",
     "axios": "1.6.2",
     "bottleneck": "2.19.5",
-    "bull": "4.10.4",
+    "bull": "4.12.8",
```
saevarma, line 142: I wonder how tightly coupled this is to @nestjs/bull, looks like they only update their bull dependency since 4.10.4 without needing to update nestjs/bull code. So this should be safe.

We will then do nestjs updates of nestjs/bull and nestjs/cache-manager during the annual monorepo upgrade.

saevarma (approved):
🚀

PR #14887, opened 2024-05-22
feat(auth-api-lib): add support for delegation types to api-scopes

Herdismaria:
[Update admin api client patch endpoint to be able to accept delegation type](https://www.notion.so/Update-admin-api-client-patch-endpoint-to-be-able-to-accept-delegation-type-1f1d485ea3634a3ebdfd1b29c59e430c?pvs=4)

What
Start using the new delegation types db table in auth admin api.

When creating or updating supported delegation types for a api scope we now both update the old boolean fields in the api_scope table and create/remove rows from the new api_scope_delegation_types table.

Why
We want to move to using a new supportedDelegationType array field when retrieving supported delegation types for an api_scope.

## Checklist:

## Summary by CodeRabbit

- **New Features**
  - Enhanced delegation type management within API scopes.
  - Added functionality to include and exclude delegation types in API scope operations.

- **Bug Fixes**
  - Improved validation and processing of delegation types for API scopes.

- **Documentation**
  - Updated API documentation to cover new delegation type features.

- **Refactor**
  - Renamed properties in delegation type models for consistency.
  - Optimized methods for handling delegation types and their relationships.


libs/auth-api-lib/src/lib/delegations/models/delegation-type.model.ts
```diff
@@ -62,6 +64,9 @@ export class DelegationTypeModel extends Model<
   @BelongsToMany(() => Client, () => ClientDelegationType)
   clients!: Client[]
 
+  @BelongsToMany(() => ApiScope, () => ApiScopeDelegationType)
+  apiScopes!: ApiScope[]
```
GunnlaugurG, line 68: Here you want to make this field creation optional so we will not have to add this field to the creation model.
```suggestion
  apiScopes!: CreationOptional<ApiScope[]>
```

libs/services/auth/testing/src/fixtures/fixture-factory.ts
```diff
@@ -600,6 +600,7 @@ export class FixtureFactory {
           providerId: delegationProvider.id,
           clients: [],
           provider: delegationProvider,
+          apiScopes: [],
```
GunnlaugurG, line 603: After you make this field CreationOptional in the `deletaion-type.model.ts` file you should be able to remove this.

GunnlaugurG (commented):
Two minor comments, otherwise LGTM!

saevarma (approved):
🚀

PR #14932, opened 2024-05-26
chore(app): Make codemagic cancel builds if there are no app changes

eirikurn:
## What

Configure codemagic to check that there are app changes since last build.

## Why

Before, the Github Check would show as "Failed" when there were new app changes. Now it shows as "Skipped". This helps devops monitoring.

## Summary by CodeRabbit

- **Chores**
  - Implemented conditional workflow triggers based on changes in the `apps/native/app` directory.


thoreyjona (approved):
Nice 😄

PR #14933, opened 2024-05-26
fix(clients-statistics): Process csv grouped digit strings

baering:
## What

Sometimes CSV lines contain commas that are only there for presentational purposes. We cannot split columns on those commas so we need to make sure that does not happen.

## Why

So that we do not end up with incorrect column values.

## Checklist:

## Summary by CodeRabbit

- **New Features**
  - Improved CSV parsing by adding advanced handling for commas within quotes.
  
- **Bug Fixes**
  - Enhanced processing of percentages and numbers with commas in data sources.


saevarma (approved):
🚀

PR #14963, opened 2024-05-28
feat(university-gateway): Removed DegreeType.Other

jonarnarbriem:
We expect all programs sent to us from the universities to be of an accepted type and thus type OTHER as a default option makes no sense.

We still want to throw an error as we want to know if data sent from the universities doesn't match our definition, this is now done by the mapStringToEnum() function.

# ...

## What

## Why

## Screenshots / Gifs

## Checklist:

## Summary by CodeRabbit

- **Refactor**
  - Simplified degree type mapping in the Reykjavik University application client.
  - Adjusted mode of delivery mapping without explicit handling of empty values.

- **Breaking Changes**
  - Removed the `OTHER` value from the `DegreeType` enum.


davidkef (approved):
lgtm

PR #14998, opened 2024-05-29
fix(native-app): notifications minor fixes

thoreyjona:
## What

* Fix icon on mark all as read button
* Add spinner to notification sheet when fetching
* Update cache correctly when clicking mark all as read

## Checklist:

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved cache logic when marking all notifications as read.
  - Updated loading indicators in the notifications modal header.
  - Adjusted icon styles for better visual consistency.

- **Style**
  - Modified button icon spacing for improved layout.


apps/native/app/src/screens/notifications/notifications.tsx
```diff
@@ -251,6 +233,7 @@ export const NotificationsScreen: NavigationFunctionComponent = ({
         title={intl.formatMessage({ id: 'notifications.screenTitle' })}
         onClosePress={() => Navigation.dismissModal(componentId)}
         style={{ marginHorizontal: 16 }}
+        showLoading={loading}
```
eirikurn, line 236: There's a pattern to do something like this:

```
showLoading={res.loading && !!res.data}
```

I.e. if there's no data, we have skeleton loading UI below so we don't show the spinner in that case. But we do show it when we have are showing data but are loading/refreshing new data.

eirikurn (approved):
Nice. Just one comment.

apps/native/app/src/screens/notifications/notifications.tsx
thoreyjona: great point! Thanks, will fix 👏

PR #15015, opened 2024-05-30
fix: build

lodmfjord:
# ...

Sometimes findBestGoodRefPR and findBestGoodRefBranch get stuck in promise, even though value has been found. 

This fix makes the function return a promise that is resolved (async function - DO NOT DO THIS) and then on return the process is quit with process.exit(0) so we don't have to handle promises not resolved (VERY LAZY).

This is aggressive, doesn't follow any agreed upon pattern, and the author should rightfully be mocked and scolded for this not so great solution but because this code will soon be dropped this is a quick-hack to make it work to open ways for node-upgrade pr. 

## Summary by CodeRabbit

- **Improvements**
  - Enhanced asynchronous operations for better performance and reliability.

- **Fixes**
  - Improved error handling and process flow to ensure smooth execution.

### DIFF
--- a/.github/actions/main.ts
+++ b/.github/actions/main.ts
@@ -37,9 +37,10 @@ const FULL_REBUILD_NEEDED = 'full_rebuild_needed'
 
   if (rev === 'rebuild') {
     console.log(FULL_REBUILD_NEEDED)
-    return
+    process.exit(0);
   }
   rev.branch = rev.branch.replace(/'/g, '')
   rev.ref = rev.ref.replace(/'/g, '')
   console.log(JSON.stringify(rev))
+  process.exit(0);
 })()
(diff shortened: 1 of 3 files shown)


.github/actions/main.ts
```diff
@@ -37,9 +37,10 @@ const FULL_REBUILD_NEEDED = 'full_rebuild_needed'
 
   if (rev === 'rebuild') {
     console.log(FULL_REBUILD_NEEDED)
-    return
+    process.exit(0)
```
AndesKrrrrrrrrrrr, line 40: Looks kinda funky

AndesKrrrrrrrrrrr (approved):
Pretty good shim to _make stuff work_

PR #15027, opened 2024-05-30
fix(documents): Document total count as number

thordurhhh:
## What

Document total count should always return a number.

## Why

We were a bit too unforgiving in our approach for accepting `totalCount`. Now we return 0 when getting a falsy from the service.

## Checklist:

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of document counts by setting `totalCount` to 0 when undefined or null, preventing errors and ensuring smoother user experience.


thorkellmani (approved):
Lgtm

PR #15038, opened 2024-05-31
feat(admins-portal): applications statistics for portals admin

jonnigs:
# ...

## What

Be able to fetch statistics for all application types. How many of each type have been created in a selected period and how many of each status.

## Why

This is run every month and c/p into a large excel file, this should eliminate the need for the excel file and the need to have a programmer perform and sql query at the start of each month. 

## Screens

Screenshot 2024-05-31 at 12 30 21
hots / Gifs

## Checklist:

## Summary by CodeRabbit

- **New Features**
  - Introduced a `StatisticsForm` component for filtering and displaying application statistics with date pickers and filters.
  - Added a `Statistics` function component for displaying application statistics, including date filtering and error handling.
  - Implemented `StatisticsTable` for visualizing statistics data.
  - Integrated a new query hook for fetching application statistics data.


libs/api/domains/application/src/lib/application-admin/application-admin.resolvers.ts
```diff
@@ -47,4 +49,23 @@ export class ApplicationAdminResolver {
...
+  @Scopes(AdminPortalScope.applicationSystemAdmin)
+  async applicationApplicationsAdminStatistics(
+    @CurrentUser() user: User,
+    @Args('locale', { type: () => String, nullable: true })
+    locale: Locale = 'is',
+    @Args('input')
+    input: ApplicationApplicationsAdminStatisticsInput,
+  ) {
+    const test =
+      await this.applicationService.getApplicationCountByTypeIdAndStatus(
```
obmagnusson, line 63: Just return this directly

libs/api/domains/application/src/lib/application.model.ts
```diff
@@ -131,6 +131,30 @@ export class Application {
   status!: ApplicationResponseDtoStatusEnum
 }
 
+@ObjectType()
+export class ApplicationStatistics {
+  @Field(() => String)
+  typeid!: string
```
obmagnusson, line 137: Should we use the friendly display name here ?

apps/application-system/api/src/app/modules/application/admin.controller.ts
```diff
@@ -42,6 +43,42 @@ export class AdminController {
...
+        },
+      },
+    },
+  })
+  async getCountByTypeIdAndStatus(
+    @Query('startDate') startDate: string,
+    @Query('endDate') endDate: string,
+  ) {
+    const test =
+      await this.applicationService.getApplicationCountByTypeIdAndStatus(
```
obmagnusson, line 75: return directly

libs/portals/admin/application-system/src/components/StatisticsForm/StatisticsForm.tsx
```diff
@@ -0,0 +1,69 @@
...
+  Select,
+} from '@island.is/island-ui/core'
+import { theme } from '@island.is/island-ui/theme'
+import { useLocale } from '@island.is/localization'
+import { useEffect, useState } from 'react'
+import { useWindowSize } from 'react-use'
+import { ApplicationFilters, MultiChoiceFilter } from '../../types/filters'
+import { m } from '../../lib/messages'
+import { statusMapper } from '../../shared/utils'
+import { log } from 'console'
```
jonnigs, line 16:
```suggestion
```

libs/portals/admin/application-system/src/components/StatisticsForm/StatisticsForm.tsx
```diff
@@ -0,0 +1,68 @@
...
+export const StatisticsForm = ({ onDateChange }: Props) => {
+  const { formatMessage } = useLocale()
+  const { width } = useWindowSize()
+
+  const filters: ApplicationFilters = {
+    nationalId: undefined,
+    period: { from: undefined, to: undefined },
+    institution: undefined,
+    status: undefined,
+  }
```
Toti91, line 31: Aren't these values always undefined?

libs/portals/admin/application-system/src/components/StatisticsForm/StatisticsForm.tsx
```diff
@@ -0,0 +1,68 @@
...
+            size="xs"
+            locale="is"
+          />
+        </Box>
+        <DatePicker
+          id="periodTo"
+          label=""
+          backgroundColor="blue"
+          minDate={filters.period.from}
+          selected={filters.period.to}
```
Toti91, line 59: We should pass the `dateInterval` state from the parent as a prop and use it as "selected" on the datepickers. that way we can also have default selected dates, for example from `start-of-current-month` to `now`

libs/application/api/core/src/lib/application/application.service.ts
```diff
@@ -55,6 +55,29 @@ export class ApplicationService {
...
+        COUNT(*) FILTER (WHERE status = 'approved') AS approved 
+      FROM public.application 
+      WHERE created BETWEEN '${startDate}' AND '${endDate}' 
+      GROUP BY typeid;`
+
+    const [results] = await this.sequelize.query(query, {
+      replacements: { startDate, endDate },
+    })
+
+    return JSON.stringify(results)
```
obmagnusson, line 78: Should return a type here. Could use sth like:
```
    return this.sequelize.query<ApplicationStatistics>(query, {
        replacements: { startDate, endDate },
        type: QueryTypes.SELECT,
      })

```

libs/portals/admin/application-system/src/components/StatisticsForm/StatisticsForm.tsx
obmagnusson: remove?

libs/portals/admin/application-system/src/module.tsx
```diff
@@ -23,14 +28,31 @@ const getScreen = ({ userInfo }: PortalModuleRoutesProps): React.ReactNode => {
 export const applicationSystemAdminModule: PortalModule = {
   name: m.applicationSystem,
   layout: 'full',
-  enabled: ({ userInfo }) => {
-    return userInfo.scopes.some((scope) => allowedScopes.includes(scope))
-  },
+  enabled: ({ userInfo }) =>
+    userInfo.scopes.some((scope) => allowedScopes.includes(scope)),
```
Toti91, line 32: The statistics should only be available to users with the `applicationSystemAdmin` scope

We have to find a way to exclude Statistics from the view if the user does not have `applicationSystemAdmin`

Toti91 (changes requested):
Looks very nice
Couple of comments though

libs/portals/admin/application-system/src/components/StatisticsForm/StatisticsForm.tsx
jonnigs, line 31:
```suggestion
```

libs/portals/admin/application-system/src/module.tsx
Toti91: This should work

```ts
routes: (props) => [
  // .....
  {
    name: m.statistics,
    path: ApplicationSystemPaths.Statistics,
    element: <Statistics />,
    enabled: props.userInfo.scope?.includes(AdminPortalScope.applicationSystemAdmin),
  },
]
```

obmagnusson (changes requested):
Nice Work. Couple of comments on the backend of things.

libs/portals/admin/application-system/src/components/StatisticsForm/StatisticsForm.tsx
jonnigs: Default dates are a nice touch! Just pushed an update for this.

libs/api/domains/application/src/lib/application.model.ts
jonnigs: do you mean typeId instead of typeid?

obmagnusson: No just meant to use the display name from the Template for displaying purposes. Like the template typeId `CriminalRecord`  would have the display name `Sakavottorð` but for this first iteration I think we should leave this as is.

jonnigs: Ah, gotcha. Makes sense.

obmagnusson (approved):

libs/portals/admin/application-system/src/components/StatisticsForm/StatisticsForm.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+  return (
+    <Box
+      display="flex"
+      alignItems="center"
+      justifyContent="spaceBetween"
+      flexDirection={['column', 'column', 'column', 'row']}
+    >
+      <Box display="flex" flexDirection={['column', 'column', 'row']}>
+        <Box marginX={[0, 0, 2]} marginY={[2, 2, 0]}>
```
Toti91, line 22: Looking at the screenshot, maybe change to `marginRight`, so that the filters line up with rest of content

Toti91 (changes requested):
Just 2 nitpicks otherwise veeery nice!

Toti91 (approved):

PR #15090, opened 2024-06-04
fix(service-portal): Filter for unread. Loading state for total page count

thordurhhh:
## What

Filter for unread. Loading state for total page count

## Why

* unread messages were caught in the sanitation filter. Need to set an exception for special case for `opened` as we have three states:
    * `undefined` / `null` => Show all documents
    * `true` => Show only opened
    * `false` => Show only unopened
* pagination disappeared so there was no way to set 0 count on pagination.

## Checklist:


thorkellmani (approved):
lgtm

PR #15095, opened 2024-06-04
fix(parental-leave): review actionCard in template

ylfahfa:
# Review actionCard

[TS-362](https://dit-iceland.atlassian.net/browse/TS-362)

## What

Review action card texts and make sure they are accurate and describing the correct state transitions.

## Why

So applicants have a better overview of their application history.

## Screenshots / Gifs

N/A

## Checklist:


apps/application-system/form/src/routes/AssignApplication.tsx
```diff
@@ -41,7 +42,7 @@ export const AssignApplication = () => {
     },
   )
 
-  useEffect(() => {
+  useEffectOnce(() => {
```
Toti91, line 45: What is the reason for this change?

ylfahfa: When we were reviewing the state changes and history logs we noticed that when the employer opens the email notifying them of the employer's parental leave the state transition gets triggered twice (because of useEffect - strict in React 18). This set the whole history log chain out of sync. We are aware that this isn't the case on prod but wanted to be able to see it correctly on dev.

Toti91 (approved):

birkirkristmunds (approved):
LGTM

PR #15104, opened 2024-06-05
feat(j-s): Update subpoena compensation claim decision 

unakb:
## What

- Added compensation claim acceptance decision to Defender
- Added acceptance decision to the subpoena PATCH request
- Only return indictment cases to the digital mailbox API if they have received an Arraignment date

## Checklist:

## Summary by CodeRabbit

- **New Features**
  - Added `getLawyer` method to retrieve defender information by national ID.

- **Enhancements**
  - Updated the `getAllCases` method to return an array of cases.
  - Added optional language parameter to `getSubpoena` and `updateSubpoena` methods for multilingual support.


apps/judicial-system/digital-mailbox-api/src/app/modules/cases/case.controller.ts
```diff
@@ -61,10 +61,11 @@ export class CaseController {
   async getSubpoena(
     @Param('caseId') caseId: string,
     @CurrentUser() user: User,
+    @Query() query?: { lang: string },
```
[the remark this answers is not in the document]
unakb: It is used, query?.lang instead of query.lang though.

apps/judicial-system/backend/src/app/modules/defendant/models/defendant.model.ts
```diff
@@ -125,4 +125,8 @@ export class Defendant extends Model {
   })
   @ApiProperty()
   verdictViewDate?: string
+
+  @Column({ type: DataType.BOOLEAN, allowNull: true })
+  @ApiPropertyOptional({ type: Boolean })
+  acceptCompensationClaim?: boolean
```
gudjong, line 131: Should this field not be added to the judicial-system-api defendant model as well?

gudjong (approved):
Approved with comments

apps/judicial-system/backend/src/app/modules/defendant/models/defendant.model.ts
unakb: True, I added it in! Although we're not technically using it anywhere on the web yet I think we probably will need to use it in the front end at some point in the near future.

PR #15112, opened 2024-06-05
fix(service-portal): Update get for documents

thordurhhh:
## What

Update document list get

## Why

No need for sending id

## Checklist:

- [x] I have performed a self-review of my own code
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] Formatting passes locally with my changes
- [x] I have rebased against main before asking for a review

## Summary by CodeRabbit

- **Refactor**
  - Updated the document retrieval process to remove the `nationalId` parameter from various components and services. This change streamlines the handling of document lists and enhances data processing efficiency.

PR #15114, opened 2024-06-05
fix(service-portal): Update get for documents hotfix

thordurhhh:
hotfix: https://github.com/island-is/island.is/pull/15112


jonbjarnio (approved):
🚀

PR #15135, opened 2024-06-06
chore(application-system): remove padding for component that shouldn't render

jonnigs:
## What

Remove padding for components that shouldn't render

## Why

System was adding unexpected spacing in some cases, where multiField has a `space` attribute and one of its children has a condition that equates to `false`

## Screenshots / Gifs

Before:
Screenshot 2024-06-06 at 15 36 52

After: 
Screenshot 2024-06-06 at 15 37 18

## Checklist:

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected import path for the `overview` module in the passport annulment form.

- **Refactor**
  - Removed unused import of `Application` and eslint-disable comment in the reference template form.
  - Removed unused import of `cn` library in the form shell component.

- **New Features**
  - Enhanced `FormMultiField` component to conditionally display fields based on user authentication and specific conditions.
  - Introduced logic to hide fields in `FormMultiField` based on the `isNavigable` property.


libs/application/ui-shell/src/components/FormMultiField.tsx
```diff
@@ -82,11 +84,26 @@ const FormMultiField: FC<
...
+          application.externalData,
+          user,
+        )
+
+        const paddingBottom = !evaluatedCondition
+          ? 0
+          : index === children.length - 1
+          ? 0
+          : space
+
```
Toti91, line 101: This makes sense, small improvement:

The condition is already being evaluated in the `convertMultiFieldToScreen` reducer function,
it's being passed down to every child field as `isNavigable`

So in theory we could skip this and just do something like

```ts
const paddingBottom = !field.isNavigable ? 0 : index === children.length - 1 ? 0 : space
```

libs/application/ui-shell/src/components/FormMultiField.tsx
```diff
@@ -45,6 +46,7 @@ const FormMultiField: FC<
 }) => {
   const { description, children, space = 0 } = multiField
   const { formatMessage } = useLocale()
+  const { userInfo: user } = useAuth()
```
Toti91, line 49: Not needed anymore!

Toti91, line 49:
```suggestion
```

libs/application/ui-shell/src/components/FormMultiField.tsx
```diff
@@ -17,6 +17,7 @@ import { useLocale } from '@island.is/localization'
 import ConditionHandler from './ConditionHandler'
 import FormField from './FormField'
 import { FieldDef, MultiFieldScreen } from '../types'
+import { useAuth } from '@island.is/auth/react'
```
Toti91, line 20:
```suggestion
```

Toti91 (approved):

PR #15140, opened 2024-06-07
feat(j-s): Actor title in indictment confirmation

oddsson:
# Actor title in indictment confirmation

[Asana](https://app.asana.com/0/1199153462262248/1207512584142109/f)

## What

Add the actor's title after their name in the confirmation stamp on confirmed indictments. To do this, we needed to convert the sizes in this stamp from pixels to points (see reasoning 👇).

1. Resolution Independence
Points are a unit of measurement that is independent of screen resolution. One point is defined as 1/72 of an inch. This means that content defined in points will appear the same size regardless of the display's resolution (DPI or PPI). In contrast, pixels are tied to the resolution of the display device, which can vary widely.

2. Consistency Across Devices and Media
Using points ensures that the document's content will look consistent across different devices (monitors, tablets, smartphones) and media (print and digital). When you use pixels, the appearance of the document can change depending on the device's resolution. For example, a 100-pixel wide line will look different on a 72 DPI screen compared to a 300 DPI screen.

3. Print Fidelity
PDFs are often used for documents that need to be printed. Printers operate at much higher resolutions than screens, typically 300 DPI or more. Using points allows for precise control over print dimensions and ensures that printed content maintains its intended size and layout. Pixels, on the other hand, do not translate directly to print dimensions and can lead to inconsistencies.

4. Scalability
Designing with points allows PDFs to be easily scalable. Text and graphics defined in points can be scaled up or down without losing quality. When using pixels, scaling up can lead to pixelation and loss of quality.

5. Standardization
The PDF format was designed to be device-independent and uses points as its standard unit of measurement. This standardization simplifies document creation and ensures that PDFs meet the expectations for size and layout consistency.

6. Better Typography
Points are also the standard unit in typography, which is crucial for maintaining the integrity of fonts and text layout in a document. This ensures that the text looks professional and is easy to read across different media. 

## Why

It needs to be clear who exactly is the person responsible for the indictment's confirmation. People can have the same name and then there is no way of telling who made the confirmation. Two persons with the same name and title is highly unlikely, however.

### Before

Screenshot 2024-06-07 at 10 36 42

### Absurdly long name
Screenshot 2024-06-07 at 10 11 45

### Normal name length 

Screenshot 2024-06-07 at 10 12 29

## Checklist:

## Summary by CodeRabbit

- **New Features**
  - Added actor's title to indictment confirmation in PDF documents.

- **Refactor**
  - Streamlined import statements for better organization and maintainability.
  - Updated PDF layout and text rendering functions to improve consistency.

- **Bug Fixes**
  - Ensured consistent line width in PDF documents for better visual quality.


apps/judicial-system/backend/src/app/formatters/pdfHelpers.ts
```diff
@@ -111,123 +113,147 @@ export const addPoliceStar = (doc: PDFKit.PDFDocument) => {
 
 export const addIndictmentConfirmation = (
   doc: PDFKit.PDFDocument,
-  confirmedBy: string,
-  institutionName: string,
-  createdDate: Date,
+  confirmation: IndictmentConfirmation,
```
gudjong, line 116: Note that indictmentPdf is importing from pdfHelpers and vice versa.

Please consider either of these two approaches:
- Move the definition of IndictmentConfirmation to pdfHelpers, or
- call addIndictmentConfirmation with individual arguments rather than a confirmation object.

gudjong (approved):
One suggestion

PR #15150, opened 2024-06-07
feat(native-app): Use new document endpoints for inbox

thoreyjona:
## What

Switch out document endpoints in app to use new endpoints:

```
listDocumentsV2 -> documentsV2
getDocument -> documentV2
postMailAction -> postMailActionV2
```

Also start using `unreadCount` variable from the server. (There is a small bug with this that when filtering the documents the `unreadCount` changes based on documents filtered but Advania said that they will fix it soon, I can remove that change if we want to go with this live first).

## Why

Old endpoints will be deprecated. Mínar síður already uses the new ones on dev.

One of the biggest changes is that we no longer get fileType ("pdf", "html" or "url") with the listDocuments query. (Previous implementation was not working well, was even broken for some documents on prod. The value was not coming from the service (Advania) itself (the service does not know the type of the file until it fetches the document detail) but there was instead some 3 year old function 'deciding' on the fileType based on the sender etc. 

So now we read the document from the cache if we have it and then fetch only the content to add. 

## Checklist:

## Summary by CodeRabbit

- **New Features**
  - Updated document handling to use `DocumentV2`, improving data structure and clarity.
  - Enhanced document detail view with better data fetching and error handling.

- **Bug Fixes**
  - Fixed issues with document queries and mutations to ensure compatibility with new document versioning.

- **Refactor**
  - Renamed and replaced various document-related queries and fragments to align with `DocumentV2`.
  - Improved data handling and conditional rendering based on updated document properties.


apps/native/app/src/graphql/fragments/document.fragment.graphql
```diff
@@ -1,15 +1,20 @@
...
   categoryId
   bookmarked
+  sender {
+    id
+    name
+  }
+  content {
+    type
+    value
+  }
```
eirikurn, line 18: Should this be here?

apps/native/app/src/graphql/queries/inbox.graphql
```diff
@@ -6,28 +6,33 @@ query ListOrganizations {
...
+  documentV2(input: $input) {
+    ...ListDocument
   }
 }
 
-mutation PostMailActionMutation($input: PostMailActionResolverInput!) {
-  postMailAction(input: $input) {
+query GetDocumentContent($input: DocumentInput!) {
+  documentV2(input: $input) {
+    content {
```
eirikurn, line 27: Might want to add `id` field for apollo cache merging logic.

apps/native/app/src/screens/document-detail/document-detail.tsx
```diff
@@ -197,29 +205,35 @@ export const DocumentDetailScreen: NavigationFunctionComponent<{
...
-            page: 1,
-            pageSize: 50,
-          },
-        },
-      })
+      // If we don't have the document in the cache, fetch it
+      void getDocument()
+    } else if (doc.data?.content === null) {
+      // Content is missing, fetch it
+      void getDocumentContent()
```
eirikurn, line 219: This looks like it has waterfall'y performance.

Maybe we should just combine the two in one lazy query which includes the document content and run it `if (doc.missing || doc.data?.content === null)`.

apps/native/app/src/screens/document-detail/document-detail.tsx
```diff
@@ -197,29 +205,35 @@ export const DocumentDetailScreen: NavigationFunctionComponent<{
...
   const [visible, setVisible] = useState(false)
   const [loaded, setLoaded] = useState(false)
   const [pdfUrl, setPdfUrl] = useState('')
   const [touched, setTouched] = useState(false)
-  const hasPdf = Document.fileType?.toLocaleLowerCase() === 'pdf'
-  const isHtml = typeof Document.html === 'string' && Document.html !== ''
+
+  const loading = docRes.loading || !accessToken
+  const fileTypeLoaded =
+    !docContent.loading && !docContent.error && !docRes.loading && !docRes.error
```
eirikurn, line 230: This is probably `true` during initial render. Simpler to check if `Document?.content?.type` has a value?

apps/native/app/src/screens/inbox/inbox.tsx
```diff
@@ -316,7 +302,7 @@ export const InboxScreen: NavigationFunctionComponent<{
     res.refetch()
   }, [refresh])
 
-  const items = res.data?.data ?? []
+  const items = useMemo(() => res.data?.data ?? [], [res.data])
```
eirikurn, line 305: Nice catch!

eirikurn (commented):
Good stuff!

apps/native/app/src/graphql/fragments/document.fragment.graphql
thoreyjona: Yes, we are using this when fragment both when fetching the list and also when fetching document details that should include this, also hopefully one day we will get the type in the list as well 😊

apps/native/app/src/screens/document-detail/document-detail.tsx
thoreyjona: Good point. Should we then fetch all of the data at all times? Because in most cases when we come from the inbox screen we have all the info we need except for `content.type` and `content.value`, so I was trying to prevent fetching all the other info when we already have it. I think in most cases we already have all the other data, the other call is just a 'belti and axlabönd' kind of thing 😄

apps/native/app/src/graphql/fragments/document.fragment.graphql
thoreyjona: But I could exclude this from the list call and just add it for fetching the whole document.

eirikurn (approved):

PR #15152, opened 2024-06-07
feat(ui-core): isMulti option for Select

Toti91:
# isMulti option for Select

## What

Add support for `isMulti` for our Select component to be able to select multiple options.

Typing could be better, but trying to make this change as frictionless as possible is hard..

The main problem is in the application-system `SelectFormField`, which can't be properly typed with a [Discriminated union](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#discriminated-unions) type because the field type is already part of [another union](https://github.com/island-is/island.is/blob/main/libs/application/types/src/lib/Fields.ts#L587), at least I think that is the problem. Which means we can't get the correct type for `onSelect` based on if `isMulti` is true or false.. It still works fine, but you would have to assert the type on the `onSelect` if you use it with `isMulti`, e.g.

```ts
buildSelectField({
  // ...
  isMulti: true,
  onSelect: (value) => {
    // value can be either a single value or an array
    // even though "isMulti" is true
    if (Array.isArray(value)) {
      // do something
    }
  }
})
```

If anyone has any ideas how to fix this then it would be appreciated.
The ui-core Select component and the SelectController are correctly typed.

## Why

Increased need for this option, f.x. in the application-system

## Screenshots / Gifs

image

## Checklist:

## Summary by CodeRabbit

- **New Features**
  - Added multi-select functionality to `SelectField` and `AsyncSelectField` components, allowing users to select multiple options.
  - Introduced new components `MultiValue` and `MultiValueLabel` to enhance multi-value selection in the `Select` component.

- **Enhancements**
  - Updated styling to support and visually represent multi-value selections.
  - Added a new story in the component library to demonstrate the multi-select feature.

- **Bug Fixes**
  - Adjusted the `SelectController` to handle multi-select scenarios correctly.


albinagu (approved):

jonnigs (approved):
LGTM

PR #15171, opened 2024-06-10
feat(core): hide upload button in some cases

jonnigs:
## What

Make the UX for the file upload a little bit better. File upload elements can take in one file or many. This hides the upload button when one file for a single file upload has been filled in.

## Why

When a file upload for a single file, all ready had a file, there was nothing to indicate that another one couldn't be uploaded and there was no difference between the two

## Screenshots / Gifs

Single and multiple without files: 
Screenshot 2024-06-10 at 14 23 24

Single and multiple with files: 
Screenshot 2024-06-10 at 14 23 08

## Checklist:

## Summary by CodeRabbit

- **Refactor**
  - Simplified component definitions and prop destructuring in `FileUploadController` and `FileUploadFormField`.
  - Updated import statements and rendering logic in `InputFileUpload` for improved performance and readability.


disaerna (approved):
UI files changes look good 👍🏼 

Would suggest to check the text though.
Single: Mismatch in title and description "Upload file - Upload **a** file" and then the button says "Veljið **skjöl** ..."
Multiple: Description says "Upload **a** files".

Also the button is icelandic and the title and desc in english.

libs/island-ui/core/src/lib/InputFileUpload/InputFileUpload.tsx
```diff
@@ -322,9 +322,11 @@ export const InputFileUpload = ({
...
       <Text>{description}</Text>
       <Box marginY={4}>
-        <Button variant="ghost" icon="attach" disabled={disabled}>
-          {buttonLabel}
-        </Button>
+        {(multiple || (!multiple && fileList.length === 0)) && (
+          <Button variant="ghost" icon="attach" disabled={disabled}>
+            {buttonLabel}
+          </Button>
+        )}
```
Toti91, line 329: Maybe we should let the condition wrap the parent `<Box>` so that unnecessary margin is avoided

jonnigs: I want the space, this is a bit cramped without it. I thought this was a bit more clean than adding a space on this condition as well

PR #15175, opened 2024-06-10
Fix(scripts): `yarn get-secrets` fails silently when not logged in

AndesKrrrrrrrrrrr:
If my session has expired or there's some issue with my AWS config, the script fails silently. Here we fix that

- Fix silent failure
- Make failure clearly visible


lodmfjord (approved):
LGTM

PR #15211, opened 2024-06-12
fix(web): Remove autoAuth for PensionCalculatorApi due to 500 error when calculating pension

RunarVestmann:
# Remove autoAuth for PensionCalculatorApi due to 500 error when calculating pension

## What

*A PR: https://github.com/island-is/island.is/pull/14812 caused an the calculation page to only display a 500 error due to a token exchange in the backend not being successful

This PR will address that issue and we'll not to hotfix it to production as well since the change in the PR mentioned above has reached prod

### Before
image

### After
image

## Checklist:

## Summary by CodeRabbit

- **New Features**
  - Introduced an `autoAuth` property to enhance API configuration flexibility. This allows for conditional authentication based on the new `autoAuth` flag.


thorkellmani (approved):
🚤

PR #15212, opened 2024-06-12
fix(web): Hotfix - Remove autoAuth for PensionCalculatorApi due to 500 error when calculating pension #15211

RunarVestmann:
# Hotfix - Remove autoAuth for PensionCalculatorApi due to 500 error when calculating pension #15211


thorkellmani (approved):
🚤

PR #15246, opened 2024-06-14
fix(native-app): offline mode more fixes

thoreyjona:
## What

* Fix offline mode when you go offline with app open
* Persist userInfo in app so we show it correctly when offline. 
I tried to only persist userInfo.name and userInfo.nationalId but did not find a proper way to do it (zustand whitelist only accepts keys of the store). 

This is the info that is stored: 
Is something of this too delicate to persist? E.g. the sub? @eirikurn 
```  
{
  "auth_01": {
    "state": {
      "userInfo": {
        "sub": "E5D3E5E081EC4BB765D25485030499577E0F5FC10A89599DBDEB5CD6231DB9BD",
        "nationalId": "0101303019",
        "subjectType": "person",
        "name": "Gervimaður Afríka",
        "gender": "male",
        "birthdate": "1930-01-01",
        "locale": "en"
      }
    },
    "version": 0
  }
}

```
## Checklist:

## Summary by CodeRabbit

- **New Features**
  - Enhanced authentication store with token persistency, auto-refresh, login, and logout capabilities.
  - Improved app lock screen functionality with more precise overlay dismissal using lock screen component ID.

- **Bug Fixes**
  - Corrected the handling of the `lockScreenActivatedAt` parameter, ensuring it is set to `null` properly.

- **Refactor**
  - Added a condition to `setupEventHandlers()` to improve app lock screen behavior based on `lockScreenActivatedAt` status.

### DIFF
--- a/apps/native/app/src/stores/auth-store.ts
+++ b/apps/native/app/src/stores/auth-store.ts
@@ -1,3 +1,4 @@
+import AsyncStorage from '@react-native-community/async-storage'
 import { Alert } from 'react-native'
 import {
   authorize,
@@ -9,6 +10,7 @@ import {
 import Keychain from 'react-native-keychain'
 import createUse from 'zustand'
 import create, { State } from 'zustand/vanilla'
+import { persist } from 'zustand/middleware'
 import { bundleId, getConfig } from '../config'
 import { getIntl } from '../contexts/i18n-provider'
 import { getApolloClientAsync } from '../graphql/client'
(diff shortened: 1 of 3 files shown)


apps/native/app/src/stores/auth-store.ts
```diff
@@ -59,134 +61,143 @@ const getAppAuthConfig = () => {
...
+          true,
+        )
+        return true
+      },
+    }),
+    {
+      name: 'auth_01',
+      getStorage: () => AsyncStorage,
+      whitelist: ['userInfo'],
+    },
```
thoreyjona, line 198: Only thing I changed is to add `persist(` at the top and then this object with three lines at the bottom 😊

apps/native/app/src/utils/app-lock.ts
```diff
@@ -55,13 +55,13 @@ export function showAppLockOverlay({
   })
 }
 
-export function hideAppLockOverlay() {
-  // Dismiss all overlays
-  Navigation.dismissAllOverlays()
+export function hideAppLockOverlay(lockScreenComponentId: string) {
+  // Dismiss the lock screen
+  Navigation.dismissOverlay(lockScreenComponentId)
```
heita-is, line 60: Ahh nice catch

apps/native/app/src/utils/lifecycle/setup-event-handlers.ts
```diff
@@ -126,9 +126,10 @@ export function setupEventHandlers() {
         if (lockScreenComponentId) {
           if (
             lockScreenActivatedAt !== undefined &&
+            lockScreenActivatedAt !== null &&
```
heita-is, line 129: ~nit: You could use `!isDefined(lockScreenActivatedAt)` utility func instead of those two lines

heita-is (approved):
Awesome nice catch on the dismiss overlay 😅

snaerseljan (approved):

PR #15284, opened 2024-06-19
fix(inheritance-report): keep "deleted" asset in view as disabled

albinagu:
Screenshot 2024-06-19 at 14 27 28

## Checklist:

## Summary by CodeRabbit

- **New Features**
  - Introduced `RealEstateLookup` and `VehicleLookup` components for property and vehicle lookup functionalities.

- **Improvements**
  - Enhanced `EstateAndVehiclesRepeater` with dynamic field management and total calculations.
  - Added `disabled` prop to `DeceasedShare` component for improved control.

- **Renaming**
  - Renamed `HeirsAndPartitionRepeater` to `HeirsRepeater`.
  - Renamed `HeirsOverview` to `OverviewHeirs`.

- **Bug Fixes**
  - Corrected button class name in `OtherAssetsRepeater`.

- **Style**
  - Updated import path for styles in `PrintScreen` component.

- **Data Filtering**
  - Added filtering to include only enabled `realEstate` and `bankAccounts` entries.


libs/application/templates/inheritance-report/src/fields/EstateAndVehiclesRepeater/FieldComponent.tsx
```diff
@@ -0,0 +1,103 @@
...
+    required: readOnly ? false : field.required,
+    loading: fieldName === loadingFieldName,
+    suffix: field.suffix,
+    onChange: () => onAfterChange?.(),
+    error: error,
+    readOnly: readOnly,
+    disabled: disabled,
+  }
+
+  let content = null
```
kksteini, line 47: Does content have to be null?
What about just declaring the variable since we have an override in the default case of the switch? 
```suggestion
  let content
```

kksteini (approved):
Maybe consider disabling the any rule or getting assistance on how to use the `useFieldArray<T>` for some type.

I tried looking into it but got into very strange issues with the typing system and useFieldArray. I can see why people have not figured out how to do it, there are no instances of this not being `any` in the codebase so far.

Otherwise, disable any eslint just to reduce warnings and explicitly let other programmers know that this is known (on purpose)

PR #15294, opened 2024-06-20
chore(core): Remove tsconfig.base.json paths wildcard

saevarma:
## What

Remove `*` from `tsconfig.base.json`.

## Why

It is set to the `shared/types` project but we should never import from it using only the filename. We should always import as `@island.is/shared/types`

## Screenshots / Gifs

N/A

## Checklist:


saevarma:
Done as part of the Nx upgrade

PR #15361, opened 2024-06-26
fix(native-app): Fix app lock biometrics race condition

snaerseljan:
# Fix app lock biometrics race condition

This is an attempt to fix multiple stacking of the app lock screen because of frequent app state changes.

## What

Based on the logs below, the issue seems to be that the app state changes between inactive, background, and active states in quick succession, causing multiple overlays to be shown and hidden rapidly. This leads to multiple instances of the App Lock overlay being displayed.

Solution track overlay state without using the navigation library component did appear event since that is too late to fire.
Rather use a flag to track whether the overlay is currently shown, preventing duplicate overlays.

## Why

Here is what happens:

Status can be inactive, active and background

inactive
show app lock screen
active
hide app lock screen
status inactive
status background
show app lock screen
show app lock screen
status active
hide app lock screen

and the stack count then is
shown 3
hidden 2

Next time the user presses the biometrics button this happens

status inactive
show app lock screen
call unlockApp
hide app lock screen
status active

Now the stack count is

shown 1
hidden 1

But the whole stack count of app lock screen is still 1 in favour of shown

## Checklist:

## Summary by CodeRabbit

- **New Features**
  - Improved overlay handling to provide a more seamless user experience with application events.


thoreyjona (approved):
🚀

PR #15365, opened 2024-06-26
fix(ui-fields): TableRepeaterField with NationalIdWithName component

albinagu:
Screenshot 2024-06-26 at 16 47 16
Screenshot 2024-06-26 at 16 47 28
Screenshot 2024-06-26 at 16 48 15

## Checklist:

## Summary by CodeRabbit

- **New Features**
  - Introduced support for custom value mapping in table repeater fields, specifically for components like `nationalIdWithName`.

- **Improvements**
  - Enhanced table repeater functionality to handle specific component types more effectively.
  - Improved user data presentation by enabling the mapping and formatting of national IDs within the table.


libs/application/ui-fields/src/lib/TableRepeaterFormField/TableRepeaterFormField.tsx
```diff
@@ -76,15 +79,29 @@ export const TableRepeaterFormField: FC<Props> = ({
...
+      //let table account for name thats being lookup up
+      insert(tableRows, item.id, 'name')
+      insert(tableHeader, item.label, 'Nafn')
+
+      //nationalIdWithName returns an object that we extract entries from and add to values
+      if (values) {
+        values = values.map((value: any) => flattenValueObject(value, item.id))
+      }
+    }
+  })
```
Halli1414, line 103: Might be a good idea to extract this to a seperate function. We should try to make it clear that this is a solution for components that don't fit nicely into the table repeater. We should also keep an eye on this and possibly update the tableRepeater field to better accommodate components if this gets out of hands.

libs/application/ui-components/src/components/NationalIdWithName/NationalIdWithName.tsx
```diff
@@ -81,13 +81,13 @@ export const NationalIdWithName: FC<
...
   } else if (!errorMessage) {
-    nationalIdFieldErrors = getErrorViaPath(errors, nationaIdField)
+    nationalIdFieldErrors = getErrorViaPath(errors, nationalIdField)
   }
 
   // get default values
   const defaultNationalId = nationalIdDefaultValue
     ? nationalIdDefaultValue
-    : getValueViaPath(application.answers, nationaIdField, '')
+    : getValueViaPath(application.answers, nationalIdField, '')
```
jonnigs, line 90:
```suggestion
    : getValueViaPath(application.answers, nationalIdField)
```
Last input is optional and can be skipped

libs/application/ui-components/src/components/NationalIdWithName/NationalIdWithName.tsx
```diff
@@ -81,13 +81,13 @@ export const NationalIdWithName: FC<
...
   const defaultName = nameDefaultValue
     ? nameDefaultValue
     : getValueViaPath(application.answers, nameField, '')
```
jonnigs, line 93:
```suggestion
    : getValueViaPath(application.answers, nameField)
```

libs/application/ui-fields/src/lib/TableRepeaterFormField/utils.ts
```diff
@@ -0,0 +1,29 @@
...
+  // need to extract entries from and add to values
+  if (values) {
+    values.forEach((value: any, index: string) => {
+      if (typeof value[item.id] === 'object') {
+        const { [item.id]: nestedObject, ...rest } = value
+        values[index] = { ...nestedObject, ...rest }
+      }
+    })
+  }
+}
```
jonnigs, line 29: Can this be done with a type or a generic instead of any?
This isn't returning anything, wouldn't it be better to return a new object instead of changing the input object

libs/application/ui-fields/src/lib/TableRepeaterFormField/utils.ts
```diff
@@ -0,0 +1,29 @@
...
+  // nationalIdWithName returns an object that we
+  // need to extract entries from and add to values
+  if (values) {
+    values.forEach((value: any, index: string) => {
+      if (typeof value[item.id] === 'object') {
+        const { [item.id]: nestedObject, ...rest } = value
+        values[index] = { ...nestedObject, ...rest }
+      }
+    })
+  }
```
jonnigs, line 28: Early return to avoid nesting ´if(!values)return...´

libs/application/ui-fields/src/lib/TableRepeaterFormField/TableRepeaterFormField.tsx
```diff
@@ -85,6 +88,8 @@ export const TableRepeaterFormField: FC<Props> = ({
   const staticData = getStaticTableData?.(application)
   const canAddItem = maxRows ? savedFields.length < maxRows : true
 
+  checkForCustomMappedComponents(tableItems, values)
```
jonnigs, line 91: Shouldn't this be caught in a new constant?

jonnigs (approved):

PR #15366, opened 2024-06-26
fix(core): Support Fish shell in the clean.sh script

saevarma:
## What

The `^^` operator in bash for uppercasing isn´t supported in fish, making the clean.sh script fail. So using the tr instead.

## Why

So the same script can work on both bash and fish.

## Screenshots / Gifs

## Checklist:

## Summary by CodeRabbit

- **Refactor**
  - Improved the script for cleaning tasks to dynamically convert job names to uppercase for better consistency and readability.

### DIFF
--- a/scripts/clean.sh
+++ b/scripts/clean.sh
@@ -155,7 +155,8 @@ clean_all() {
   local jobs=("generated" "caches" "yarn" "node_modules")
 
   for job in "${jobs[@]}"; do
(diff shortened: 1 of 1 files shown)


AndesKrrrrrrrrrrr (approved):
We're not specifically supporting other shells than Bash. We can, however, do it on a best-effort basis with basic fixes like these. Thank you :heart:

PR #15371, opened 2024-06-27
fix(fa): application finanical aid adding array check 

MargretFinnboga:
# hotfix

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved safety when accessing the `childrenFiles` array to prevent potential errors.


libs/application/templates/financial-aid/src/fields/Summary/SummaryForm.tsx
```diff
@@ -62,7 +62,7 @@ const SummaryForm = ({ application, goToScreen }: FAFieldBaseProps) => {
     childrenFiles: UploadFile[],
     income: ApproveOptions,
   ) => {
-    if (childrenFiles.length > 0) {
+    if (childrenFiles && childrenFiles.length > 0) {
```
Rikkiabb, line 65: Could also just do:

childrenFiles?.length > 0 

Or even:

childrenFiles?.length

Just saying 🤷  but approved

apps/financial-aid/open-api/src/app/models/application.model.ts
```diff
@@ -12,6 +12,7 @@ import {
   DirectTaxPaymentModel,
   StaffModel,
 } from './index'
+import { ChildrenModel } from './children.model'
```
Rikkiabb, line 15: I think the rabbit is correct 🐰

Rikkiabb (approved):

PR #15429, opened 2024-07-03
fix(j-s): Case file drag fixes

oddsson:
# Case file drag fixes

[Asana](https://app.asana.com/0/1199153462262248/1207648320806232/f)

## What

There are two issues with the case file generation that could result in the case file UI and PDF to not match. 

1. It's currently possible to drag chapters.
2. If a file is dragged outside of the CaseFileAccordionItem component, we don't update the database. This happens because we use `onPointerUp` on the file being dragged to update the db but if the cursor is not on the file, the handler is not called.

This PR disables dragging of chapters as well as allowing drag in the X axis. This fixes pt. 2 because now the cursor is always on the file being dragged.

### Before

https://github.com/island-is/island.is/assets/3789875/e01c70df-4ddf-43d0-a90c-d85c982afe68

#### Notice the network calls being made even though the file is being dragged outside the component

https://github.com/island-is/island.is/assets/3789875/555d4380-6e55-47a2-abba-859ad7ddc61c

## Checklist:

## Summary by CodeRabbit

- **New Features**
  - Improved cursor handling for drag-and-drop actions within the Indictments Case Files section for better user experience.


unakb (approved):
lgtm!

PR #15435, opened 2024-07-03
fix(j-s): Handle empty string

unakb:
[Höndla tóma strengi í defender reitum á defendant fyrir info card ](https://app.asana.com/0/1199153462262248/1207705265556030)

## What

Display empty defender string if defender name is an empty string 

## Checklist:

## Summary by CodeRabbit

- **Bug Fixes**
  - Updated the logic for displaying the defender's name to ensure the correct message is shown when no defender is assigned.


apps/judicial-system/web/src/components/InfoCard/DefendantInfo/DefendantInfo.tsx
```diff
@@ -88,8 +88,9 @@ export const DefendantInfo: FC<DefendantInfoProps> = ({
         {displayDefenderInfo && (
           <Box display="flex" key={defendant.defenderName} role="paragraph">
             <Text as="span">{`${formatMessage(strings.defender)}: ${
-              defendant.defenderName ??
-              formatMessage(strings.noDefenderAssigned)
+              defendant.defenderName
+                ? defendant.defenderName
+                : formatMessage(strings.noDefenderAssigned)
```
gudjong, line 93: Ah, I should have spotted this in my earlier review. You could also do
```
defendant.defenderName || formatMessage(strings.noDefenderAssigned)
```
if you like that better.

Its sometimes a bit tricky to decide if you want to use `??` or `||`, the former is appropriate if you only want to replace `undefined` or `null` and the latter if you want to replace a falsy value. Most of the time I use `??` but that was obviously not correct in this situation.

gudjong (approved):

PR #15444, opened 2024-07-04
feat(form-system) Application

Hyosam:
# ...

- An initial shell for the application part of the form-system project within apps/portals/form-system
- Two libraries to be shared by both the new app and the form building part of the form-system project:
- graphql library 
- Ui library

## Checklist:

- [x ] Formatting passes locally with my changes


apps/portals/form-system/src/index.src.html
```diff
@@ -0,0 +1,44 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8" />
+    <title>FormTODO - Ísland.is</title>
```
baering, line 5: Remember to fill out the `FormTODO`'s here

baering (changes requested):
A good start to get the shell up on main.

I added some comments for you to review.

Some thoughts about introducing vite/swc. We prefer our projects to have as similar of a setup as we can to make maintaining them simpler and would prefer to upgrade them all at once if we decide to make changes like this. So let's revert these changes for now, this is something we can discuss during dev discipline meetings later and then make the change for more projects.

PR #15478, opened 2024-07-09
feat(native-app): update nav bar right buttons

thoreyjona:
## What

Updating nav bar buttons. Only show notification bell on home screen and settings icon on more screen. Move scan icon on wallet screen to be on the right side instead of left.

## Why

Simplify nav bar buttons and create space for screen specific buttons when needed. 

## Screenshots / Gifs

https://github.com/island-is/island.is/assets/19665778/abc3acf9-81d8-425a-880b-b2f0286568e5

## Checklist:

### DIFF
--- a/apps/native/app/src/screens/applications/applications.tsx
+++ b/apps/native/app/src/screens/applications/applications.tsx
@@ -20,11 +20,9 @@ import {
   useListSearchQuery,
 } from '../../graphql/types/schema'
 import { createNavigationOptionHooks } from '../../hooks/create-navigation-option-hooks'
-import { useActiveTabItemPress } from '../../hooks/use-active-tab-item-press'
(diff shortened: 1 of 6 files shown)


apps/native/app/src/screens/applications/applications.tsx
```diff
@@ -139,13 +135,6 @@ export const ApplicationsScreen: NavigationFunctionComponent = ({
     [],
   )
 
-  useActiveTabItemPress(3, () => {
-    flatListRef.current?.scrollToOffset({
-      offset: -150,
-      animated: true,
```
thoreyjona, line 145: This was causing the screen to scroll a bit down. Removing at least for now, until we decide if we want this and then what pressing the active tab should do (was not doing anything I think).

eirikurn (approved):
LGTM

PR #15483, opened 2024-07-09
fix(native-app): open correct browser when authenticating using passkeys

thoreyjona:
## What

When authenticating with passkeys, make sure to open up the in app browser instead of opening the real browser. 
Was broken in one place since I forgot to send in componentId.

## Checklist:


eirikurn (approved):
LGTM

PR #15493, opened 2024-07-10
chore: Update AWS dependencies & consolidate AWS usage

AndesKrrrrrrrrrrr:
Merged pull request from #15400 and #15470

In this PR:
- Updated `@aws-sdk/*` dependencies
- Removed `aws-sdk` (v2) import where possible
- De-duplicated S3-related logic (mostly uploading/downloading files in application forms)

## Checklist:

## Summary by CodeRabbit

- **New Features**
	- Updated the GitHub Actions workflow to improve repository checkout and Node.js environment setup.
	- Enhanced configuration structure for better type safety and flexibility in environment settings.
	- Integrated AWS services into various modules to enhance functionality and streamline file operations.
	- Introduced a lazy loading mechanism for the Elasticsearch client, improving service robustness.
- **Bug Fixes**
	- Resolved issues related to improved error handling and code clarity.
- **Documentation**
	- Improved explanations and clarity regarding changes to the environment configurations.


svanaeinars (approved):

apps/judicial-system/backend/src/app/modules/aws-s3/awsS3.service.ts
```diff
@@ -37,72 +45,66 @@ const formatS3Key = (caseType: CaseType, key: string) => {
...
-            resolve(data)
-          }
-        },
-      )
+  ): Promise<{ url: string; fields: Record<string, string> }> {
+    const { url, fields } = await createPresignedPost(this.s3Client, {
+      Bucket: this.config.bucket,
+      Key: formatS3Key(caseType, key),
+      Conditions: [
+        ['content-length-range', 0, 10485760], // 10 MB
```
gudjong, line 67: Is there a hard limit on the file size? We are dealing with court documents and they may get bigger than this.

gudjong (commented):
I tested the code locally and it woks fine.
I have one question about file size range in `createPresignedPost`.

apps/application-system/api/src/environments/environment.ts
```diff
@@ -34,84 +48,98 @@ const devConfig = {
...
+  | 'clientLocationOrigin'
+  | 'email'
+  | 'emailOptions'
+  | 'generalPetition'
+  | 'islykill'
+  | 'jwtSecret'
+  | 'presignBucket'
+  | 'userProfile'
+  | 'xRoadBasePathWithEnv'
+  | 'smsOptions'
```
jonnigs, line 87: this is used in at least two places, should this be it's own type?

libs/application/template-api-modules/src/lib/modules/shared/shared.service.ts
```diff
@@ -153,10 +150,13 @@ export class SharedTemplateApiService {
...
     templateGenerator: AttachmentEmailTemplateGenerator,
     application: Application,
-    fileContent: string,
+    fileContent: Parameters<AttachmentEmailTemplateGenerator>[1],
     recipientEmail: string,
     locale = 'is',
   ) {
+    this.logger.debug('Sending email with attachment', {
+      applicationId: application.id,
+    })
```
jonnigs, line 159: Should this be here?

libs/application/template-api-modules/src/lib/modules/templates/children-residence-change/children-residence-change.service.ts
```diff
@@ -25,39 +24,33 @@ import { SmsService } from '@island.is/nova-sms'
...
-
-export const PRESIGNED_BUCKET = 'PRESIGNED_BUCKET'
+import { AwsService } from '@island.is/nest/aws'
 
 type props = Override<
   TemplateApiModuleActionProps,
   { application: CRCApplication }
 >
 
+export const PRESIGNED_BUCKET = 'PRESIGNED_BUCKET'
```
jonnigs, line 34: Is there a constants.ts file that this belongs in?

apps/application-system/api/src/environments/environment.ts
```diff
@@ -18,10 +31,11 @@ const devConfig = {
       process.env.WEB_FRONTEND_PORT ?? '4242'
     }/umsoknir`,
     emailOptions: {
-      useTestAccount: !(process.env.USE_SES === 'true'),
-      useNodemailerApp: process.env.USE_NODEMAILER_APP === 'true' ?? false,
+      useTestAccount: !(process.env.USE_SES === 'true') ?? '',
+      useNodemailerApp:
+        process.env.USE_NODEMAILER_APP === 'true' ?? false ?? '',
```
jonnigs, line 36: Guðjón is also changing those lines on PR https://github.com/island-is/island.is/pull/15514 maybe the two of you should make sure the two changes won't clash an break things

Valur (approved):

apps/services/search-indexer/src/migrate/lib/aws.ts
```diff
@@ -43,66 +69,64 @@ interface PackageStatuses {
...
+  const domainPackage = domainPackageList.DomainPackageDetailsList?.find(
     (listItem) => listItem.PackageID === packageId,
   )
 
   return {
-    packageStatus: packages.PackageDetailsList[0].PackageStatus,
-    domainStatus: domainPackage?.DomainPackageStatus ?? 'DISSOCIATED',
+    packageStatus: packages.PackageDetailsList?.[0]
+      .PackageStatus as PackageStatus,
+    domainStatus: domainPackage?.DomainPackageStatus ?? 'DISSOCIATING',
```
kksteini, line 94: I assume the `DISSOCIATED` → `DISSOCIATING` was caught because of the es client switchout/update?
So I'm further assuming that the type system will bark at us if `AVAILABLE` and `DISSOCIATING` are misspelt?

If not, I'd like to see these enumerated somewhere :>

apps/services/search-indexer/src/migrate/lib/aws.ts
```diff
@@ -190,31 +216,64 @@ const uploadFileToS3 = (options: Omit<PutObjectRequest, 'Bucket'>) => {
...
+      if (chunks.every((chunk) => typeof chunk === 'string')) {
+        // If all chunks are strings, concatenate them and convert to Buffer
+        resolve(Buffer.from(chunks.join('')))
+      } else if (
+        chunks.every(
+          (chunk) => Buffer.isBuffer(chunk) || chunk instanceof Uint8Array,
+        )
+      ) {
+        // If all chunks are Buffer or Uint8Array, use Buffer.concat
+        resolve(Buffer.concat(chunks))
```
kksteini, line 267: I'm not sure how big the buffers are going to be but do you think it is worth considering iterating through the chunks once, and then deciding on a resolve strategy?

Dirty pseudocode:
```pseudocode
declare x;
for every chunk {
  x["string"] += 1 if chunk is string
  x["isBuffer"] += 1 if chunk.isBuffer() or chunk is instanceof u8[]
};

..do the if else if else dance \O/
```

kksteini: Actually, nevermind. Thinking more on this. If we're frequently dealing with mixed types then exhausting the chunks once would likely run slower than the current solution.

libs/content-search-toolkit/src/services/elastic.service.ts
```diff
@@ -36,8 +34,12 @@ import { tagAggregationQuery } from '../queries/tagAggregation'
 import { typeAggregationQuery } from '../queries/typeAggregation'
 import { rankEvaluationQuery } from '../queries/rankEvaluation'
 import { filterDoc, getValidBulkRequestChunk } from './utils'
+import {
+  createAWSConnection,
+  awsGetCredentials,
+} from '@acuris/aws-es-connection'
 
-type RequestBodyType<T = Record<string, any>> = T | string | Buffer
+type RequestBodyType<T = Record<string, unknown>> = T | string | Buffer
```
kksteini, line 42: Appreciate this `any` to `unknown` :relaxed:

kksteini (approved):
LGTM

apps/judicial-system/backend/src/app/modules/aws-s3/awsS3.service.ts
svanaeinars: Yes, this is a hard limit. What size do you suggest?

gudjong: The files we upload to S3 are later uploaded to the district courts. The API we are working against there currently has a limit of 100 MB. Our limit needs to reflect that. Note that the bucket we are writing to in S3 is not the same as any buckets used by island.is. Until now, we have not been restricted to any limit as far as I know.

AWS S3 itself has a maximum object size limit:
	•	Single PUT operation: The maximum size for a single PUT operation (which includes uploads via pre-signed URLs) is 5 GB.
	•	Multipart upload: For larger files, you can use the multipart upload API to upload objects up to 5 TB in size.

apps/application-system/api/src/environments/environment.ts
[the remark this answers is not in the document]
gudjong: Actually, we will never reach the last part becuase false is never undefined. Note that ?? is not the same as ||.

apps/application-system/api/src/environments/environment.ts
```diff
@@ -18,10 +31,11 @@ const devConfig = {
...
       options: {
-        region: process.env.EMAIL_REGION ?? 'eu-west-1',
+        region: process.env.EMAIL_REGION ?? 'eu-west-1' ?? '',
```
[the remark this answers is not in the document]
gudjong: Agree that the last empty string is not reachable, but the reason is because 'eu-west-1' is always defined, not that it is always true.

apps/application-system/api/src/environments/environment.ts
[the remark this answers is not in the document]
svanaeinars: This line has been modified to only include the first check, since it will always yield a boolean

apps/application-system/api/src/environments/environment.ts
svanaeinars: This is used in two placed, both in this file, a new type has been created

libs/application/template-api-modules/src/lib/modules/shared/shared.service.ts
svanaeinars: I think for debugging purposes this is good to have, so I think it was included on purpose

apps/application-system/api/src/environments/environment.ts
svanaeinars: Noted, will keep an eye on this :)

apps/services/search-indexer/src/migrate/lib/aws.ts
svanaeinars: Well spotted, these types are already defined in PackageStatus and DomainPackageStatus so I modified the code to use those.

apps/services/search-indexer/src/migrate/lib/aws.ts
svanaeinars: Marking this as resolved

apps/judicial-system/backend/src/app/modules/aws-s3/awsS3.service.ts
gudjong: Thank you.

gudjong (approved):

libs/api-catalogue/elastic/src/lib/queries/search.model.ts
```diff
@@ -82,6 +82,7 @@ export const searchQuery = ({
       { 'title.keyword': { order: 'asc' } },
     ],
     size: limit + 1, // if we have a page number add it as offset for pagination
+    search_after: [] as typeof searchAfter,
```
eirikurn, line 85: ~Please make sure that the ES client does not handle this value differently from the property being skipped previously. I've been 🔥  too many times by similar bugs.

package.json
```diff
@@ -140,9 +153,9 @@
     "apollo-server-express": "3.10.2",
     "apollo-server-plugin-response-cache": "3.8.2",
     "archiver": "6.0.1",
-    "aws-elasticsearch-connector": "9.0.0",
-    "aws-sdk": "2.814.0",
-    "aws4": "1.11.0",
+    "aws-elasticsearch-connector": "9.2.0",
+    "aws-sdk": "2.1651.0",
+    "aws4": "1.13.0",
```
eirikurn, line 158: ~Was hoping this PR managed to refactor away some of these alternate packages. 🙃

libs/clients/cms/src/lib/cms.config.ts
```diff
@@ -18,14 +18,7 @@ export const CmsConfig = defineConfig({
...
     redis: {
-      nodes: env.requiredJSON('REDIS_URL_NODE_01', [
-        'localhost:7000',
-        'localhost:7001',
-        'localhost:7002',
-        'localhost:7003',
-        'localhost:7004',
-        'localhost:7005',
-      ]),
+      nodes: env.requiredJSON('REDIS_URL_NODE_01', []),
```
eirikurn, line 21: Why clear the devFallback values? This might break current developer workflows (run docker compose, start app normally).

libs/api/domains/file-upload/src/lib/file-upload.resolver.ts
```diff
@@ -8,9 +7,13 @@ export class FileUploadResolver {
...
+  async createUploadUrl(
     @Args('filename') filename: string,
-  ): Promise<S3.PresignedPost> {
-    return this.fileStorageService.generatePresignedPost(filename)
+  ): Promise<PresignedPost> {
+    const result = await this.fileStorageService.generatePresignedPost(filename)
+    return {
+      url: result.url,
+      fields: result.fields,
+    }
```
eirikurn, line 17: ~Can the result be returned directly?

libs/file-storage/src/lib/file-storage.service.ts
```diff
@@ -37,31 +47,25 @@ export class FileStorageService {
...
-          reject(err)
-        } else {
-          resolve(data)
-        }
-      })
+      Key: key,
+      Conditions: [
+        ['content-length-range', 0, 10000000], // Max 10MB
+      ],
+      Expires: PRESIGNED_POST_EXPIRES / 1000, // createPresignedPost expects seconds
```
eirikurn, line 56: ~Can just update the constant further up in this file.

libs/nest/aws/src/lib/aws.service.ts
```diff
@@ -31,44 +90,54 @@ export class AwsService {
...
       Key: fileName,
       Body: content,
       ...uploadParameters,
-    }
+    })
 
-    const { Location: url } = await this.s3.upload(uploadParams).promise()
-    return url
+    await this.s3Client.send(command)
+    return `https://${bucket}.s3.amazonaws.com/${fileName}`
```
eirikurn, line 101: ~Should we hardcode this here? Do S3 URLs depend on region and more?

libs/nest/aws/src/lib/aws.service.ts
```diff
@@ -1,24 +1,83 @@
...
+    return out.Body?.transformToString(encoding)
+  }
 
-    return await this.s3.getObject(downloadParams).promise()
+  async getFileB64(fileName: string): Promise<string | undefined>
+  async getFileB64(
+    bucket: string,
+    fileName: string,
+  ): Promise<string | undefined>
+  async getFileB64(
```
eirikurn, line 75: ~No need to minify B64
```suggestion
  async getFileBase64(fileName: string): Promise<string | undefined>
  async getFileBase64(
    bucket: string,
    fileName: string,
  ): Promise<string | undefined>
  async getFileBase64(
```

libs/nest/aws/src/lib/aws.service.ts
```diff
@@ -1,24 +1,83 @@
...
+  ): Promise<string | undefined>
+  async getFileEncoded(
+    bucket: string,
+    fileName?: string,
+    encoding: Encoding = 'base64',
+  ): Promise<string | undefined> {
+    const out = fileName
+      ? await this.getFile(bucket, fileName)
+      : await this.getFile(bucket)
+    return out.Body?.transformToString(encoding)
```
eirikurn, line 67: Using optional positional arguments can be a bit dangerous.

Afaics, if I call `awsService.getFileB64('https://mybucket.s3...')` it will end up looking a file with "base64" as `fileName` and the S3 URL as a bucket name.

eirikurn (changes requested):

eirikurn (commented):
P.s. only reviewed core files. 🙏

libs/nest/aws/src/lib/aws.service.ts
svanaeinars: I changed the implementation so these functions arre now using `interface` instead of positional arguments. Also added a new argument `s3Uri`, for the cases where we only have a uri.

libs/nest/aws/src/lib/aws.service.ts
svanaeinars: Changed to getFileBase64

libs/file-storage/src/lib/file-storage.service.ts
svanaeinars: True, I changed the constant value and removed the division by 1000.

libs/api/domains/file-upload/src/lib/file-upload.resolver.ts
svanaeinars: Yes turns out it can --> done.

libs/api-catalogue/elastic/src/lib/queries/search.model.ts
svanaeinars: I reverted back to the original approach where `search_after` is only added to the query if it contains elements, and added explicit typing for the result so we don't get the linter error: `Unexpected any. Specify a different type.`.

libs/api-catalogue/elastic/src/lib/elastic.service.ts
```diff
@@ -46,18 +45,15 @@ export class ElasticService {
...
-          },
-        })
-        bulk.push(service)
-      })
+      const bulk = services.map((service) => ({
+        index: {
+          _index: this.indexName,
+          _id: service.id,
+        },
+      }))
```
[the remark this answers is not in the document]
svanaeinars: This is true, this has been updated so the service is also included.

libs/api-catalogue/elastic/src/lib/elastic.service.ts
```diff
@@ -142,14 +141,19 @@ export class ElasticService {
...
+    const result = await client.ping().catch((error) => {
       logger.error('Error in ping', error)
     })
     logger.info('Got elasticsearch ping response')
     return result
   }
+  private async getClient(): Promise<Client> {
+    if (this.client) return this.client
+    return await this.createEsClient()
+  }
```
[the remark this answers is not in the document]
svanaeinars: This has been modified, so now `this.client` is also updated in the function.

apps/services/auth/delegation-api/src/app/delegations/test/delegation-index/delegation-index.service.spec.ts
```diff
@@ -464,6 +465,7 @@ describe('DelegationsIndexService', () => {
 
       // Act
       await delegationIndexService.indexRepresentativeDelegations(nationalId)
+      await sleep(200) // Small sleep for less flakiness
```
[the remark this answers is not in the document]
svanaeinars: Thanks, I have modified the setup function to use `flatMap` instead of `map`

libs/nest/aws/src/lib/aws.service.ts
svanaeinars: I asked around and it seems the bucket url is correct since s3 bucket names are unique and do not depend on the region.

apps/services/user-notification/src/app/modules/notifications/tests/notifications.service.spec.ts
```diff
@@ -74,7 +74,7 @@ describe('NotificationsService', () => {
     service = module.get<NotificationsService>(NotificationsService)
   })
 
-  it('should be defined', () => {
```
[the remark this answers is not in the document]
svanaeinars: These shouldn't be async, I reverted the code back to previous version

libs/clients/cms/src/lib/cms.config.ts
svanaeinars: Fair point, I reverted this change

libs/application/template-api-modules/src/lib/modules/shared/shared.service.ts
AndesKrrrrrrrrrrr: In our live environment, debug logs are only logged when enabled specifically for an app. I.e. this has "no impact" unless we need the debug logs, in which case we can turn it on.

This was indeed intentional.

apps/services/user-notification/src/app/modules/notifications/tests/notifications.service.spec.ts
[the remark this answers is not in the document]
AndesKrrrrrrrrrrr: I had was having trouble with unawaited promises, and thought this might be a solution. Good roll-back :heart:

libs/clients/cms/src/lib/cms.config.ts
AndesKrrrrrrrrrrr: Yeah, this was leftovers from testing locally, and I couldn't get localstack to behave :sweat_smile:

package.json
AndesKrrrrrrrrrrr: - `'aws-sdk'` isn't imported anywhere anymore; could be removed
- `'aws4`' is still used in `kibana.service.ts` for signing; could be refactored, but the project owner would have to do that (too much application context needed)

libs/api/domains/file-upload/src/lib/file-upload.resolver.ts
AndesKrrrrrrrrrrr: This might leak unnecessary data, or not conform to the return type :thinking:

libs/nest/aws/src/lib/aws.service.ts
AndesKrrrrrrrrrrr: Also, IMO, we shouldn't hard-code _config values_, which the bucket name indeed is.

eirikurn (approved):
Approving core files. Nice job 👍🏻

package.json
AndesKrrrrrrrrrrr: Tried removing `aws-sdk` package, but it's being used by `winston-cloudwatch`. We need to replace it anyway. Strange, though, that `winston-cloudwatch` doesn't pull in `aws-sdk` itself as a dependency :confused:

apps/application-system/api/src/environments/environment.ts
```diff
@@ -34,84 +48,85 @@ const devConfig = {
...
 
   contentful: {
-    accessToken: process.env.CONTENTFUL_ACCESS_TOKEN,
+    accessToken: process.env.CONTENTFUL_ACCESS_TOKEN ?? '',
   },
-} as Environment
+}
 
-const prodConfig = {
+// TODO: Remove empty-string defaulting
```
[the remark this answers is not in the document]
AndesKrrrrrrrrrrr: The relevant team will be asked to fix this upon merge.

AndesKrrrrrrrrrrr (commented):
Self review, infra files LGTM

AndesKrrrrrrrrrrr (commented):
Self review of all files LGTM

Reverted some unnecessary changes and accidental removals of comments

Toti91 (approved):
Application system LGTM!

unnarb (approved):

MargretFinnboga (approved):

FridrikMProgramm (approved):
Lgtm

PR #15508, opened 2024-07-11
feat(j-s): Get connected Indictment cases to merge with

unakb:
[Bæta inn dropdown til að birta mál til að sameina við](https://app.asana.com/0/1199153462262248/1207762586950365)

## What
- Added dropdown to display connected cases
- Added alert box if no connected cases exist but user chose the merge option
- Added query to retrieve connected cases and return to web
- Added new component connected case selection
- Case that is merged with is shown on InfoCard
- Summary displays Merge tag

## Why
To handle the merging of one indictment case in to another

## Screenshots / Gifs

image
image

## Checklist:

## Summary by CodeRabbit

- **New Features**
  - Introduced functionality to fetch and display connected cases based on a given case ID.
  - Added a new component for selecting connected cases during the indictment process.
  - Enhanced the indictment summary to include merged cases.

- **Improvements**
  - Updated GraphQL queries and mutations to support connected and merged cases.
  - Added new fields and validation for merging cases.

- **Bug Fixes**
  - Added validation to prevent merging cases that are not in a received state.

- **Tests**
  - Added comprehensive tests for guard configurations and role permissions related to connected cases.

- **Documentation**
  - Updated and added new localized messages for better user guidance.


apps/judicial-system/backend/src/app/modules/case/case.service.ts
```diff
@@ -1319,6 +1320,49 @@ export class CaseService {
...
 
+  getConnectedIndictmentCases(
+    caseId: string,
+    defendant: Defendant,
+  ): Promise<Case[]> {
+    let whereClause: WhereOptions = {
+      type: CaseType.INDICTMENT,
+      id: { [Op.ne]: caseId },
+      state: [CaseState.RECEIVED],
+    }
```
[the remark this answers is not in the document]
unakb: I have to initialize it and reassign because I'm adding one of 2 possible extra options to it depending on how the national id is formatted (nationalId vs birth date)

apps/judicial-system/backend/src/app/modules/case/case.service.ts
```diff
@@ -1319,6 +1320,49 @@ export class CaseService {
...
+
+    if (defendant.noNationalId) {
+      whereClause = {
+        ...whereClause,
+        [Op.and]: [
+          { '$defendants.national_id$': defendant.nationalId },
+          { '$defendants.name$': defendant.name },
+        ],
+      }
+    } else {
```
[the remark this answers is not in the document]
unakb: Im confused as to what you mean, your code snippet looks the same as mine

apps/judicial-system/api/src/app/modules/case/dto/updateCase.input.ts
```diff
@@ -501,4 +501,9 @@ export class UpdateCaseInput {
   @IsOptional()
   @Field(() => CourtSessionType, { nullable: true })
   readonly courtSessionType?: CourtSessionType
+
+  @Allow()
+  @IsOptional()
+  @Field(() => ID, { nullable: true })
+  readonly mergeCaseId?: string
```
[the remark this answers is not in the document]
unakb: We do this check in the backend API so that would be redundant in this case

apps/judicial-system/api/src/app/modules/case/case.resolver.ts
```diff
@@ -57,6 +57,24 @@ export class CaseResolver {
...
+    input: CaseQueryInput,
+    @CurrentGraphQlUser()
+    user: User,
+    @Context('dataSources') { backendApi }: { backendApi: BackendApi },
+  ): Promise<Case[]> {
+    this.logger.debug('Getting connected cases')
+
+    return this.auditTrailService.audit(
+      user.id,
+      AuditedAction.GET_CASES,
```
gudjong, line 72: Is this the appropriate audited action here?

apps/judicial-system/backend/src/app/modules/case/case.service.ts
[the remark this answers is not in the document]
unakb: No problem 😌

apps/judicial-system/backend/src/app/modules/case/case.controller.ts
```diff
@@ -460,6 +460,33 @@ export class CaseController {
...
+    @CurrentCase() theCase: Case,
+  ): Promise<Case[]> {
+    this.logger.debug(`Getting connected cases for case ${caseId}`)
+
+    if (!theCase.defendants || theCase.defendants.length === 0) {
+      return []
+    }
+
+    const connectedCases = await Promise.all(
+      theCase.defendants.map((defendant) =>
```
gudjong, line 483: Wasn't the idea to only support cases with a single defendant?

apps/judicial-system/backend/src/app/modules/case/case.service.ts
```diff
@@ -1319,6 +1320,49 @@ export class CaseService {
...
+      whereClause = {
+        ...whereClause,
+        [Op.or]: [
+          { '$defendants.national_id$': defendant.nationalId },
+          { '$defendants.national_id$': formattedNationalId },
+        ],
+      }
+    }
+
+    return this.caseModel.findAll({
```
gudjong, line 1352: Please consider the previous comment concerning the number of defendants in a case.

gudjong: Is the idea to support merging cases from different district courts?

apps/judicial-system/web/src/routes/Court/Indictments/Conclusion/Conclusion.tsx
```diff
@@ -232,14 +238,21 @@ const Conclusion: FC = () => {
...
                 file.category === CaseFileCategory.COURT_RECORD &&
                 file.status === 'done',
             )
+          case CaseIndictmentRulingDecision.MERGE:
+            return (
+              uploadFiles.some(
+                (file) =>
+                  file.category === CaseFileCategory.COURT_RECORD &&
+                  file.status === 'done',
+              ) && workingCase.mergeCase?.id !== undefined
```
gudjong, line 254: Please consider using `!workingCase.mergeCase?.id` here as it also handles `null`.

apps/judicial-system/api/src/app/modules/case/dto/updateCase.input.ts
gudjong, line 508: Please note that we should guard against updating this field being updated if the case is not in state `RECEIVED`. This can be done in the case controller in the method `update` by throwing a `BadRequestException` it the state is not valid.

Also note that when implementing the transition that completes the case, that can be done in the case service in the `update` method. This is already being done when receiving cases.

gudjong (commented):
Some comments and questions.

apps/judicial-system/backend/src/app/modules/case/case.controller.ts
unakb: Yes you're right. My thinking here was to make this more generic than that specific use case as it felt a bit obscure to me to have a function called "getConnectedCases" that throws an error if there's more than one defendant, rather than returning all relevant information and handling the filtering of it higher up where we know what we want out of it. Maybe I'm overthinking it though for future use cases we aren't handling yet 😅 What do you think?

apps/judicial-system/backend/src/app/modules/case/case.service.ts
unakb: Do you mean filtering out here whether the court matches before returning? But yes my thinking was I'd handle the filtering higher up so that this would be a more generic function that truly returns all connected indictment cases, if I change it I'd consider renaming it to be more specific, like getConnectedIndictmentCasesByCourt or something like that

apps/judicial-system/web/src/routes/Court/Indictments/Conclusion/Conclusion.tsx
unakb: Good point! Will make the change 😊

apps/judicial-system/api/src/app/modules/case/dto/updateCase.input.ts
unakb: True, I'll make the change

apps/judicial-system/backend/src/app/modules/case/case.controller.ts
gudjong: I think you are right.

gudjong (approved):

apps/judicial-system/api/src/app/modules/case/case.resolver.ts
unakb: Good point, I've created a new one GET_CONNECTED_CASES

apps/judicial-system/web/src/routes/Court/Indictments/Conclusion/SelectConnectedCase.tsx
```diff
@@ -0,0 +1,99 @@
...
+      placeholder={formatMessage(strings.connectedCasePlaceholder)}
+      onChange={(selectedOption) => {
+        if (!selectedOption) {
+          return
+        }
+        setConnectedCase(selectedOption.value as string)
+      }}
+      isDisabled={connectedCasesLoading}
+    />
+  )
```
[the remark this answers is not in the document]
unakb: Your change is the same as my current code, is it not?

PR #15521, opened 2024-07-15
feat(j-s): Display merged case information on court Overview page 

unakb:
[Birta upplýsingar úr sameinuðu máli á "mömmumáli" á yfirliti dómara](https://app.asana.com/0/1199153462262248/1207762586950368)

## What

Display information about merged cases on court Overview page 

## Screenshots / Gifs

image

## Checklist:

## Summary by CodeRabbit

- **New Features**
  - Enhanced `InfoCardActiveIndictment` component to display detailed information about merged cases, including police case numbers, court case numbers, prosecutor, judge, court, and offense details.
  - Updated GraphQL query to include new fields related to court, judge, prosecutor, prosecutor's office, case files, police case numbers, and indictment subtypes for merged cases.
  - Added new message string `mergedFromTitle` for the "Merged From" section in indictment overviews.


apps/judicial-system/backend/src/app/modules/case/case.service.ts
```diff
@@ -302,7 +302,25 @@ export const include: Includeable[] = [
...
+  {
+    model: Case,
+    as: 'mergedCases',
+    where: { state: CaseState.COMPLETED },
+    include: [
+      {
+        model: CaseFile,
+        as: 'caseFiles',
+        required: false,
+        where: { state: { [Op.not]: CaseFileState.DELETED }, category: null },
```
gudjong, line 314: Don't we need to include other categories here? The category `CASE_FILE` is not needed as those files are included in the generated case files records, but I would think other categories relevant to indictments should be included.

gudjong (approved):
One question/comment.

apps/judicial-system/backend/src/app/modules/case/case.service.ts
unakb: Yes you're right, I removed the category filter in my other PR but this comment made me realize I need to filter out CASE_FILE so I'll add that in there

PR #15608, opened 2024-08-08
fix(ci): Utilize exit code of grep

AndesKrrrrrrrrrrr:
When using subshells, to get output, the exit code of the assignment is the exit code of the expression. If the expression failed and `set -e` is in effect, the script will exit.

## Summary by CodeRabbit

- **Chores**
	- Improved the readability of the workflow configuration by simplifying the conditional check for missing secrets.


robertaandersen (approved):
LGTM

PR #15614, opened 2024-08-09
feat(j-s): Upload LOKE case files with categories

oddsson:
# Upload LOKE case files with categories

[Asana](https://app.asana.com/0/322488613152836/1207400397034088/f)

## What

Police case files can have a category set on them. When police case files are uploaded to RVG and they have the code `RVSK` or `REIKN`, we want them to appear in `Sakarkostnaður` and `Onnur gogn` on the "Dómskjöl" screen respectively. These files are treated as if they were uploaded to RVG directly and are shown in the UI as if they were.

## Checklist:

## Summary by CodeRabbit

- **New Features**
	- Introduced a new case file category, `INVOICE`, to enhance categorization within the system.
	- Expanded functionality to allow `INVOICE` category case files to be processed alongside existing categories.
	- Added a utility function to accurately categorize police case files based on specific case codes.
	- Updated the response structure to include the `category` property for police case files.

- **Bug Fixes**
	- Adjusted filtering logic to ensure correct processing of invoice case files in various components.

- **Tests**
	- Enhanced test coverage to include scenarios for handling `INVOICE` case files, ensuring robustness in functionality.


apps/judicial-system/backend/src/app/modules/case/case.service.ts
```diff
@@ -713,7 +713,8 @@ export class CaseService {
             caseFile.key &&
             ((caseFile.category &&
               caseFilesCategories.includes(caseFile.category)) ||
-              (caseFile.category === CaseFileCategory.CASE_FILE &&
+              ((caseFile.category === CaseFileCategory.CASE_FILE ||
+                caseFile.category === CaseFileCategory.INVOICE) &&
```
gudjong, line 717: As invoices always comes from LÖKE, they will all have a police case number. Therefore this logic will not capture any invoices.

The new category `INVOICE` is better included in `caseFilesCategories` above where there is no dependence on police case number existence.

Note that the only reason the category `CASE_FILE` is handled separately is that it may or may not have a police case number and consequently either be a part of a case files record (skjalaskrá) or be a standalone file listed under "other case files" in the client..

apps/judicial-system/backend/src/app/modules/case/pdf.service.ts
```diff
@@ -78,7 +78,8 @@ export class PdfService {
       ?.filter(
         (caseFile) =>
           caseFile.policeCaseNumber === policeCaseNumber &&
-          caseFile.category === CaseFileCategory.CASE_FILE &&
+          (caseFile.category === CaseFileCategory.CASE_FILE ||
+            caseFile.category === CaseFileCategory.INVOICE) &&
```
gudjong, line 82: Invoices should not be included in case files records. They are standalone case files listed under "other case files" in the client.

apps/judicial-system/web/src/routes/Prosecutor/Indictments/CaseFile/CaseFile.tsx
```diff
@@ -65,7 +65,8 @@ const CaseFile = () => {
                     workingCase.caseFiles?.filter(
                       (caseFile) =>
                         caseFile.policeCaseNumber === policeCaseNumber &&
-                        caseFile.category === CaseFileCategory.CASE_FILE,
+                        (caseFile.category === CaseFileCategory.CASE_FILE ||
+                          caseFile.category === CaseFileCategory.INVOICE),
```
gudjong, line 69: Invoices should not be included in case files records. They are standalone case files listed under "other case files" in the client.

gudjong (changes requested):
I have some requested changes as the implementation results in invoices being included both in case files records (skjalaskrá) and as standalone files under "other case files". Otherwise, this implementation works.

I am a bit conserned about the user experience because all files from LÖKE are uploaded on the "Málsgögn" page, but only some of the uploaded files show up on the "Skjalaskrá" page. Files that are cost breakdowns or invoices show up on the "Dómskjöl" page. This however, is more of a UX design issue than an implementation issue.

Also note that there is something wrong with the stepper as you cannot select the "Dómskjöl" step when you are on the "Skjalaskrá" step.

I would probably have implemented this somewhat different. Rather than addiing a `category` field to `PolideCaseFileModel`, I would have let the code from LÖKE pass unchanged to the client and let the client decide the category when uplaoding the file. That way, I would have avoided creating a new category `INVOICE`, but rather would have used the category `CASE_FILE` instead and not stored the police case number with the invoice files. Also, I would not have stored the police case file number with the uplaoded cost breakdown files. I think that implementation would have been somewhat simpler. However, the current implementation also works once the suggested changes have been implemented.

oddsson:
Closing in favour of https://github.com/island-is/island.is/pull/15680

PR #15646, opened 2024-08-13
feat(form-system): Application view and components migration

Hyosam:
# ...

## What

- Functionality within the form-system that allows users to view an application within the form builder
- Migrated ui components to the ui-components library for form-system

## Checklist:

- [ x] My changes generate no new warnings


libs/portals/admin/form-system/src/components/TableRow/TableRow.tsx
```diff
@@ -19,12 +19,14 @@ import {
 } from '../../lib/utils/interfaces'
 import { useIntl } from 'react-intl'
 import { m } from '../../lib/messages'
+// eslint-disable-next-line no-restricted-imports
+import { format, parseISO } from 'date-fns'
```
magnearun, line 23: We should use direct imports for 'date-fns'.

libs/portals/admin/form-system/src/hooks/previewControlReducer.ts
```diff
@@ -0,0 +1,123 @@
+import { FormSystemStep, FormSystemGroup } from '@island.is/api/schema'
```
magnearun, line 1: I feel like this would belong in a folder named reducers or something else, not in the hooks folder

PR #15668, opened 2024-08-14
fix(native-app): Fixes Android post notifications for current Android…

snaerseljan:
# Fixes Android post notifications for current Android users

## What

Former Android users that have updated to Android version 13 and above are not receiving any push notifications. This is because of new Permission that was added. The permission was added in another pull request. 
However we need to specifically request POST_NOTIFICATION message on Android 13 and above.
This pull request check older users version and if they previously had notification enabled and if so then we request a new permission on Android.

## Why

Older users on version 13 and above have not received any push notifications.

## Checklist:

## Summary by CodeRabbit

- **New Features**
	- Introduced a custom hook for managing notification permissions on Android devices, improving the user experience.
	- Streamlined the onboarding process for requesting notification permissions specifically for Android.

- **Bug Fixes**
	- Enhanced the handling of notification permission requests, contributing to overall app reliability.

- **Documentation**
	- Updated documentation to clarify the new notification permission handling processes.

### DIFF
--- a/apps/financial-aid/backend/src/app/modules/municipality/models/municipality.model.ts
+++ b/apps/financial-aid/backend/src/app/modules/municipality/models/municipality.model.ts
@@ -137,6 +137,7 @@ export class MunicipalityModel extends Model<Municipality> {
     type: DataType.ENUM,
     allowNull: false,
     values: Object.values(ChildrenAid),
+    defaultValue: ChildrenAid.NOTDEFINED,
   })
   @ApiProperty({ enum: ChildrenAid })
   childrenAid: ChildrenAid
--- a/apps/native/app/src/screens/air-discount/air-discount.tsx
+++ b/apps/native/app/src/screens/air-discount/air-discount.tsx
@@ -1,14 +1,14 @@
 import {
   Alert,
-  EmptyList,
   Heading,
   Link,
   LinkText,
+  Problem,
   Skeleton,
   Typography,
 } from '@ui'
 import React from 'react'
-import { SafeAreaView, ScrollView, View, Image } from 'react-native'
+import { SafeAreaView, ScrollView, View } from 'react-native'
 import { NavigationFunctionComponent } from 'react-native-navigation'
 import { FormattedMessage, useIntl } from 'react-intl'
 import styled, { useTheme } from 'styled-components/native'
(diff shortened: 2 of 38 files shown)


apps/native/app/src/utils/permissions.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export const requestAndroidPostNotificationsPermission = async () => {
+  const granted = await PermissionsAndroid.request(
+    PermissionsAndroid.PERMISSIONS.POST_NOTIFICATIONS,
+  )
+
+  return granted === PermissionsAndroid.RESULTS.GRANTED
+}
+
+export const requestPostNotificationsPermission = async () => {
```
thoreyjona, line 14: ~why PostNotifications? instead of requestNotificationsPermission

apps/native/app/src/screens/home/home.tsx
```diff
@@ -93,6 +94,7 @@ export const MainHomeScreen: NavigationFunctionComponent = ({
 }) => {
   useNavigationOptions(componentId)
 
+  useAndroidNotificationPermission()
```
thoreyjona, line 97: ~More clear naming would IMHO be something that involves _get_ or _check_ in since we are checking for and getting the permission. If other functions regarding getting permissions have this naming convention then feel free to ignore this, just a bit unclear seeing it here what it does

apps/native/app/src/utils/permissions.ts
```diff
@@ -0,0 +1,30 @@
...
+  if (androidIsVersion33OrAbove()) {
+    // Allowing notifications toast on Android triggers the lock screen, so we need to prevent the lock screen from showing
```
thoreyjona, line 16: ~Isn't it a modal, not a toast?

thoreyjona (approved):
Nice, good job 😄 just a few nit picks about naming

apps/native/app/src/screens/home/home.tsx
snaerseljan: yes that is true I though about it but felt like useCheckAndRequestAndroidNot.... would be to long. Besides hovering over the hook the developer can see a clear description which is commented above the hook

apps/native/app/src/utils/permissions.ts
snaerseljan: Good point was referring to the actual permission which is POST_NOTIFICATIONS

apps/native/app/src/utils/permissions.ts
snaerseljan: Yes nice catch

PR #15707, opened 2024-08-19
fix(j-s): Don't update if input is invalid

oddsson:
# Don't update if input is invalid

[Asana](https://app.asana.com/0/1199153462262248/1208035569957773/f)

## What

We use a function to validate some inputs and render a error message if the input is invalid. This function had a flaw in that the value would be sent to the server even though the value was invalid. This PR fixes that.

## Why

It's a bug

### Before

https://github.com/user-attachments/assets/eccf5314-9dcb-4cc8-90cf-236bb2300465

### After
https://github.com/user-attachments/assets/476d430e-4806-4210-9a8c-458709943e83

## Checklist:

## Summary by CodeRabbit

- **New Features**
    - Enhanced type safety for the `AppealCase` and `CaseNumberInput` components with TypeScript annotations.
    - Improved error handling and validation logic in the form utility functions.

- **Bug Fixes**
    - Updated validation logic to ensure errors are correctly reported and handled before server updates.

- **Tests**
    - Added new test cases for the `validateAndSendToServer` function to ensure proper execution based on input validity.


apps/judicial-system/web/src/utils/formHelper.ts
```diff
@@ -50,10 +50,15 @@ export const validateAndSetErrorMessage = (
...
+  if (errorMessageSetter) {
     errorMessageSetter(validation.errorMessage)
-    return
   }
+
+  if (!validation.isValid) {
+    return false
+  }
+
+  return true
```
gudjong, line 61: Consider simplifying:

```suggestion
  return validation.isValid
```

gudjong (approved):
On suggestion

PR #15708, opened 2024-08-19
fix(native-app): force light mode for splash screens as well

thoreyjona:
## What

Force light mode on splash screens as well. The splash screens were still being rendered in dark mode if the user has their system setting set to dark mode.

Tested both on my iPhone and our android device.

## Checklist:

## Summary by CodeRabbit

- **New Features**
	- Updated the Android app's theme to a light-only mode for a brighter visual experience.
	- Added a new configuration to the iOS app to enforce a light mode interface, ensuring consistency across devices.


snaerseljan (approved):
Nice touch!

While you are at it, in the theme-provider.ts could you update the status bar from default to 'light-content' when apperanceMode is automatic. I seam to have forgot when I made the other dark mode fix

PR #15710, opened 2024-08-19
fix: move GestureHandlerRootView to only wrap license scanner

thoreyjona:
## What

There is a bug in [react-native-gesture-handler](https://github.com/software-mansion/react-native-gesture-handler/issues/514) that if we are wrapping `<GestureHandlerRootView>` and then showing an overlay with `react-native-navigation` then no touches outside the overlay are being registered until the overlay is closed.
We need this wrapper for `license-scanner` component to work on android so I moved it to only wrap that component instead of every component in the app. 

## Checklist:

## Summary by CodeRabbit

- **New Features**
  - Enhanced gesture handling capabilities in the License Scanner, improving user interaction.

- **Refactor**
  - Simplified component structure by removing unnecessary wrappers, streamlining the rendering process.


snaerseljan (approved):
Nice one!

PR #15748, opened 2024-08-22
Revert "feat(endorsement-lists): Revert endorsementCounter.

kksteini:
```sql
select
	"EndorsementList"."id",
	count("EndorsementList"."id") as "count"
from
	"endorsement_list" as "EndorsementList"
left outer join "endorsement" as "endorsements" on
	"EndorsementList"."id" = "endorsements"."endorsement_list_id"
where
	(("EndorsementList"."endorsementCounter" < 0
		or ("EndorsementList"."endorsementCounter" = 0
			and "EndorsementList"."counter" < 25))
	and ("EndorsementList"."tags" = array['generalPetition']::VARCHAR(255)[]
		and "EndorsementList"."opened_date" < '2024-08-21 17:00:13.913 +00:00'
		and "EndorsementList"."closed_date" > '2024-08-21 17:00:13.913 +00:00'
		and "EndorsementList"."admin_lock" = false))
group by
	"EndorsementList"."id";
```

An example of what is generated by the ORM.
We can see that this is problematic since the pagination library is written to assume that the endorsementCounter is already a property on the table.

This will instead have to be refactored on a later date with less generalization.
For example, another issue is that the `fn.count` on the sql statement returns a column of the type `BigInt` which is cast into a string.

## Checklist:

### DIFF
--- a/apps/services/endorsements/api/src/app/modules/endorsementList/endorsementList.model.ts
+++ b/apps/services/endorsements/api/src/app/modules/endorsementList/endorsementList.model.ts
@@ -97,9 +97,6 @@ export class EndorsementList extends Model {
   endorsements?: Endorsement[]
 
   @ApiProperty()
-  endorsementCounter?: number
-
-  @ApiProperty()
   @Column({
     type: DataType.JSONB,
     allowNull: false,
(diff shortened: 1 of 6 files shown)


rafnarnason (approved):
lgtm

eirikurn (approved):

PR #15763, opened 2024-08-23
feat(native-app): Update nav bar buttons

thoreyjona:
## What

Stop showing notification bell and settings icon on all screens. 
Now we have notification bell only on the home screen and settings icon only on the more screen. Also moved the license scan icon to be on the right on the wallet screen.

## Screenshots / Gifs
<div>
</div>
## Checklist:

## Summary by CodeRabbit

- **New Features**
	- Enhanced navigation options with context-specific right buttons across multiple screens (Home, Wallet, More).
	- Improved clarity in button rendering logic by incorporating relevant icons.

- **Bug Fixes**
	- Streamlined button configuration, removing unnecessary references and improving navigation consistency.

- **Chores**
	- Removed unused imports and optimized navigation setup for better performance.

### DIFF
--- a/apps/native/app/src/screens/document-detail/document-detail.tsx
+++ b/apps/native/app/src/screens/document-detail/document-detail.tsx
@@ -45,7 +45,6 @@ const PdfWrapper = styled.View`
   flex: 1;
   background-color: ${dynamicColor('background')};
 `
-
 const regexForBr = /<br \/>*\\?>/g
 
 // Styles for html documents
@@ -84,7 +83,7 @@ const useHtmlStyles = () => {
     <meta name="viewport" content="width=device-width">`
 }
 
-function getRightButtons({
+function getRightButtonsForDocumentDetail({
   archived,
   bookmarked,
 }: {
@@ -143,7 +142,7 @@ const { useNavigationOptions, getNavigationOptions } =
       },
       topBar: {
         noBorder: true,
-        rightButtons: getRightButtons(),
+        rightButtons: getRightButtonsForDocumentDetail(),
       },
     },
   )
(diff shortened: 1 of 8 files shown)


apps/native/app/src/utils/get-main-root.ts
```diff
@@ -10,41 +10,61 @@ import {
...
 type RightButtonProps = {
   unseenCount?: number
   theme?: ReturnType<typeof getThemeWithPreferences>
+  screen?: Screen
 }
 
 export const getRightButtons = ({
   unseenCount = notificationsStore.getState().unseenCount,
   theme = getThemeWithPreferences(preferencesStore.getState()),
+  screen,
```
snaerseljan, line 24: I think it would be more clear to the scope of the function to have rather icons instead of screens to choose from since the screens are all calling this function. Meaning screen can tell the function what kind of icon it wants without to provide it with its name.  aka notifications, settings, licesnseScan
And to be future proof to future screens it would be nice to allow array of icons, i.e. `['notifications', 'settings']` etc.
If there would be more than one icon per screen

apps/native/app/src/utils/get-main-root.ts
```diff
@@ -10,41 +10,61 @@ import {
...
 }: RightButtonProps = {}): OptionsTopBarButton[] => {
   const iconBackground = {
```
snaerseljan, line 26: ~return early with empty array if no screens or icons depending if you change it

apps/native/app/src/screens/document-detail/document-detail.tsx
```diff
@@ -46,7 +46,7 @@ const PdfWrapper = styled.View`
   background-color: ${dynamicColor('background')};
 `
 
-function getRightButtons({
+function getRightButtonsForDocumentDetail({
   archived,
   bookmarked,
 }: {
```
snaerseljan, line 52: ~Maybe set the properties to default to false instead, i.e.

```
function getRightButtonsForDocumentDetail({
  archived = false,
  bookmarked = false,
}: {
  archived: boolean
  bookmarked: boolean
}): OptionsTopBarButton[] {
```

snaerseljan (approved):
Nice work 👍 , cleans a lot up. Approving but two comments, nothing breaking :)

PR #15799, opened 2024-08-28
fix(j-s): Disable ServiceRequirement rather then hiding it when sent to public prosecutor

oddsson:
# Disable ServiceRequirement rather then hiding it when sent to public prosecutor

[Asana](https://app.asana.com/0/1199153462262248/1208084487748778/f)

## What

Today, when a case is sent to public prosecutor, we hide the ServiceRequirement section from court users. We should rather disable the section so that court users can see what ServiceRequirement was selected.

## Checklist:

## Summary by CodeRabbit

- **New Features**
	- Improved UI logic for handling defendants, enhancing clarity and user experience.
	- Added a disabled state to radio buttons based on the public prosecutor status, preventing modifications to service requirements after submission.
  
- **Bug Fixes**
	- Consolidated conditional checks for rendering components related to defendants for more efficient UI behavior.

### DIFF
--- a/apps/judicial-system/web/src/routes/Court/Indictments/Completed/Completed.tsx
+++ b/apps/judicial-system/web/src/routes/Court/Indictments/Completed/Completed.tsx
@@ -147,139 +147,139 @@ const Completed: FC = () => {
         <Box marginBottom={5} component="section">
           <IndictmentCaseFilesList workingCase={workingCase} />
         </Box>
-        {!sentToPublicProsecutor && (
-          <>
-            {isRulingOrFine && (
-              <Box marginBottom={5} component="section">
-                <SectionHeading
-                  title={formatMessage(strings.criminalRecordUpdateTitle)}
-                />
-                <InputFileUpload
-                  fileList={uploadFiles.filter(
-                    (file) =>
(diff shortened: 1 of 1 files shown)


apps/judicial-system/web/src/routes/Court/Indictments/Completed/Completed.tsx
```diff
@@ -147,139 +147,139 @@ const Completed: FC = () => {
...
-                          )
-                        }}
-                        large
-                        backgroundColor="white"
-                        label={formatMessage(
-                          strings.serviceRequirementNotRequired,
-                        )}
-                      />
-                    </BlueBox>
+        {!sentToPublicProsecutor && isRulingOrFine && (
```
gudjong, line 150: If I understand this correctly, this will hide the radio buttons when the case has been sent to the public prosecutor. Shouldn't `sentToPublicProsecutor` be removed from the condition?

gudjong (changes requested):
I think the buttons are still hidden in this implementation.

oddsson:
https://github.com/island-is/island.is/pull/15850

PR #15807, opened 2024-08-28
feat(income-plan): new income plan "application"

karenbjorg:
## Why

for applicants to send income plan to Social Insurance Administration (TR)

/umsoknir/tekjuaaetlun

Test user: Rögnvaldur ÞÍ Gunnarsson - 3111439

## Checklist:


ylfahfa (approved):

libs/api/domains/social-insurance/src/lib/dtos/temporaryCalculation.input.ts
```diff
@@ -0,0 +1,70 @@
...
+  @Field(() => Int, { nullable: true })
+  incomeCategoryNumber?: number
+
+  @Field()
+  incomeCategoryCode?: string
+
+  @Field()
+  incomeCategoryName?: string
+
+  @Field(() => Number, { nullable: true })
```
thorkellmani, line 26: NestJS would rather we either use Int og Float for the number decorator
image

libs/api/domains/social-insurance/src/lib/dtos/temporaryCalculation.input.ts
```diff
@@ -0,0 +1,70 @@
+import { InputType, Field, Int } from '@nestjs/graphql'
+
+@InputType('SocialInsuranceIncomeType')
+class IncomeType {
+  @Field(() => Int, { nullable: true })
+  incomeTypeNumber?: number
+
+  @Field()
+  incomeTypeCode?: string
```
thorkellmani, line 9: Is this nullable or not? The decorator says required, but the property says nullable.
Might want to review the others as well

libs/api/domains/social-insurance/src/lib/models/temporaryCalculation.model.ts
```diff
@@ -0,0 +1,55 @@
+import { Field, Int, ObjectType } from '@nestjs/graphql'
+
+@ObjectType('SocialInsuranceTemporaryCalculationRow')
+class TemporaryCalculationRow {
+  @Field()
+  name?: string
```
thorkellmani, line 6: See point about nullability above

libs/api/domains/social-insurance/src/lib/models/temporaryCalculation.model.ts
```diff
@@ -0,0 +1,55 @@
...
+  months?: Array<TemporaryCalculationMonth>
+}
+
+@ObjectType('SocialInsuranceTemporaryCalculationMonth')
+class TemporaryCalculationMonth {
+  @Field(() => Int)
+  month!: number
+
+  @Field(() => Number, { nullable: true })
+  amount?: number
```
thorkellmani, line 21: See point about graphql number above

libs/clients/social-insurance-administration/src/lib/socialInsuranceAdministrationClient.service.ts
```diff
@@ -132,4 +137,42 @@ export class SocialInsuranceAdministrationClientService {
...
+    ).apiProtectedV1IncomePlanWithholdingTaxGet(year)
+  }
+
+  // async getLatestIncomePlan(
+  //   user: User,
+  // ): Promise<TrWebCommonsExternalPortalsApiModelsIncomePlanIncomePlanDto> {
+  //   return this.incomePlanApiWithAuth(
+  //     user,
+  //   ).apiProtectedV1IncomePlanLatestIncomePlanGet()
+  // }
```
thorkellmani, line 166: Remove?

thorkellmani (commented):
Some nitpicks

thorkellmani (approved):
LGTM (Hugsmiðjan)

libs/application/ui-fields/src/lib/TableRepeaterFormField/TableRepeaterFormField.tsx
```diff
@@ -372,3 +284,190 @@ export const TableRepeaterFormField: FC<Props> = ({
...
+
+const Item: FC<ItemFieldProps> = ({
+  application,
+  error,
+  item,
+  dataId,
+  activeIndex,
+  values,
+}) => {
+  const { formatMessage } = useLocale()
```
Toti91, line 305:
```suggestion
const Item = ({
  application,
  error,
  item,
  dataId,
  activeIndex,
  values,
}: ItemFieldProps) => {
  const { formatMessage } = useLocale()
```

Nitpick:

We should me moving away from using `FC` to define props, especially if the component renders no `children`, since FC implicitly adds the `children` type to props, meaning this would be legal code, even though no children are rendered:

```ts
<Item>hello world</Item>
```

libs/application/ui-fields/src/lib/TableRepeaterFormField/TableRepeaterFormField.tsx
```diff
@@ -372,3 +284,190 @@ export const TableRepeaterFormField: FC<Props> = ({
...
+            clearErrors(id)
+          }
+        }}
+        application={application}
+        defaultValue={DefaultValue}
+        {...props}
+      />
+    </GridColumn>
+  )
+}
```
Toti91, line 473: I like these changes to the `TableRepeater`, but we should think about moving the `Item` component to a separate file, since this one is almost 500 lines.

Also we need to make sure that every other application that uses this field still works as expected, since this is a pretty big change

Toti91 (changes requested):

libs/application/ui-fields/src/lib/TableRepeaterFormField/TableRepeaterFormField.tsx
karenbjorg: we are the code owners of most of the applications that uses the table repeater and it works as expected :) there are two applications that we haven't tested that we think you (Norda) are the code owners of. Can you see if it works there?

- [ ] GrindavikHousingBuyoutForm
- [ ] HomeSupportForm
- [x] relativesSubSection
- [x] siblingsSubSection
- [x] EditOrAddEmployersAndPeriods
- [x] ParentalLeaveForm
- [x] IncomePlanForm
- [-] Signature collection (table repeater commented out)

eirikurn (approved):
Approving core files

Toti91 (approved):
application-system files LGTM

RunarVestmann (approved):
Looks good to me, I only did a textual review.

PR #15844, opened 2024-09-02
chore(j-s): Prison users access to appeal case files

oddsson:
# Prison users access to appeal case files

[Asana](https://app.asana.com/0/1199153462262248/1208009480047497/f)

## What

This PR implements access for prison users to appeal case files. The following rules apply

1. Prison admins (FMST) should only see the appeal ruling and only if the appeal case is completed
2. Prison staff should not see any appeal case files

## Checklist:

## Summary by CodeRabbit

- **New Features**
	- Introduced enhanced access control mechanisms for case retrieval based on user roles.
	- Added a new interceptor for managing case file responses according to user roles and appeal states.
	- Implemented functions to identify prison staff and admin users for improved role management.
	- Expanded case file categorization for prison administration cases.

- **Bug Fixes**
	- Improved access permissions logic for prison system users in case file guards.


apps/judicial-system/backend/src/app/modules/case/interceptors/caseFile.interceptor.ts
```diff
@@ -0,0 +1,48 @@
...
+export class CaseFileInterceptor implements NestInterceptor {
+  intercept(context: ExecutionContext, next: CallHandler): Observable<Case> {
+    const request = context.switchToHttp().getRequest()
+    const user: User = request.user
+
+    return next.handle().pipe(
+      map((data: Case) => {
+        if (
+          isPrisonStaffUser(user) ||
+          data.appealState !== CaseAppealState.COMPLETED
```
gudjong, line 31: This is too agressive as it will remove all files from defenders in indictment cases.

apps/judicial-system/backend/src/app/modules/file/guards/limitedAccessViewCaseFile.guard.ts
```diff
@@ -65,14 +66,13 @@ export class LimitedAccessViewCaseFileGuard implements CanActivate {
...
-        isCompletedCase(theCase.state) &&
-        caseFile.category &&
-        caseFile.category === CaseFileCategory.APPEAL_RULING
-      ) {
-        return true
-      }
+    if (
+      caseFile.category &&
+      isCompletedCase(theCase.state) &&
+      isPrisonSystemUser(user) &&
```
gudjong, line 72: This would give prison staff permission to open files. I suggest creating a new methos `isPrisonAdminUser` in the type lib and use it here instead of this method which includes all prison system users. Also, I suggest renaming the new category list used below to reflect that it is a list of files intended for prison admin users only.

gudjong (changes requested):
Two suggested changes.

apps/judicial-system/backend/src/app/modules/case/interceptors/caseFile.interceptor.ts
oddsson: Ah, good catch! I added a 

```
if (isDefenceUser(user)) {
    return data
}
```

i.e. nothing changes for defence users.

apps/judicial-system/backend/src/app/modules/file/guards/limitedAccessViewCaseFile.guard.ts
oddsson: Thaaaaanks! Good catch!

gudjong (approved):
Nice

PR #15852, opened 2024-09-02
chore(j-s): Allow public prosecutor user to open case files

unakb:
[Ríksak - skrifstofustarfsmenn geta ekki opnað nein skjöl á yfirliti máls. ](https://app.asana.com/0/1199153462262248/1208009480047488)

## What

Allow public prosecutor staff to open case files

## Why

Because they should be allowed to view the case files 

## Checklist:

## Summary by CodeRabbit

- **New Features**
	- Enhanced role-based access control by adding the `publicProsecutorStaffRule`, allowing public prosecutor staff to access case management functionalities.
	- Updated access permissions in the `ViewCaseFileGuard` to permit public prosecutors to view completed cases.

- **Bug Fixes**
	- Improved test coverage for role permissions related to accessing case files and generating signed URLs, ensuring correct validation of the new role.

- **Tests**
	- Expanded test cases for the `CaseController` and `FileController` to include checks for the new `publicProsecutorStaffRule`.


apps/judicial-system/backend/src/app/modules/case/case.controller.ts
```diff
@@ -548,6 +548,7 @@ export class CaseController {
   @RolesRules(
     prosecutorRule,
     prosecutorRepresentativeRule,
+    publicProsecutorStaffRule,
```
gudjong, line 551: Shouldn't the public prosecutors staff have access to other generated files for indictments?

gudjong (approved):
This looks good, but consider adding access to other generated files for indictments.

apps/judicial-system/backend/src/app/modules/case/case.controller.ts
unakb: Yes they should, good catch thank you! Will fix

PR #15888, opened 2024-09-04
feat(native-app): app should use locale on user profile

thoreyjona:
## What

Aligning locale on user with locale used in app. Fetching locale from server and setting it as the selected locale in the app. Also when changing locale in the app, send a mutation to the server to change it there as well.

## Why

One source of truth for user locale. 

## Checklist:

## Summary by CodeRabbit

- **New Features**
	- Introduced a new GraphQL query to retrieve user locale information.
	- Enhanced locale management in the Main Home Screen and Settings Screen to improve user experience.
	- Added a method to dynamically update the application's locale based on user preferences.

- **Bug Fixes**
	- Improved data consistency and retrieval efficiency for user-related queries.

- **Refactor**
	- Updated function declarations in the Settings Screen for better readability and consistency.


apps/native/app/src/stores/preferences-store.ts
```diff
@@ -61,6 +68,24 @@ export const preferencesStore = create<PreferencesStore>(
...
       ...(defaultPreferences as PreferencesStore),
+      async getAndSetLocale() {
+        const client = await getApolloClientAsync()
+
+        try {
+          const res = await client.query<
+            GetUserLocaleQuery,
+            GetUserLocaleQueryVariables
+          >({
+            query: GetUserLocaleDocument,
```
thoreyjona, line 79: Would we want this to have fetchPolicy: 'network-only' ? 🤔

apps/native/app/src/graphql/queries/profile.graphql
```diff
@@ -13,6 +13,12 @@ query GetProfile {
   }
 }
 
+query GetUserLocale {
+  getUserProfile {
```
snaerseljan, line 17: I wonder if we are already making the round trip to user profile if we should fetch it here and then load from the cache in settings

apps/native/app/src/screens/settings/settings.tsx
```diff
@@ -230,6 +231,30 @@ export const SettingsScreen: NavigationFunctionComponent = ({
       })
   }
 
+  const updateLocale = (value: string) => {
+    client
```
snaerseljan, line 235: useHooks here? or maybe easer when updating cache

snaerseljan (approved):
Good job 👍

apps/native/app/src/screens/settings/settings.tsx
thoreyjona: Good point. The other update functions for notifications were like this so I just added this the same way. I think I can use `onCompleted` from the hook for updating the cache, not sure though if the implementation that is used now is faster and will not wait for the mutation to finish or not, will have to look into it. Will keep as is for now, might consider changing all the functions in a separate PR :)

apps/native/app/src/graphql/queries/profile.graphql
thoreyjona: yes could be a good idea, will update :)

PR #15889, opened 2024-09-04
Feat/update tabs

thorkellmani:
# ...

## What

## Why

## Screenshots / Gifs

## Checklist:

### DIFF
--- a/libs/island-ui/core/src/lib/Tabs/Tabs.tsx
+++ b/libs/island-ui/core/src/lib/Tabs/Tabs.tsx
@@ -1,4 +1,4 @@
-import React, { FC, ReactNode, useEffect, useState } from 'react'
+import React, { FC, ReactNode, useEffect, useRef, useState } from 'react'
 import cn from 'classnames'
 import isNumber from 'lodash/isNumber'
 import { useWindowSize } from 'react-use'
@@ -29,6 +29,7 @@ interface TabInterface {
   contentBackground?: Colors
   size?: 'xs' | 'sm' | 'md'
   onChange?(id: string): void
+  //DEPRECATED
   onlyRenderSelectedTab?: boolean
 }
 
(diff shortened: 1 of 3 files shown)


disaerna (changes requested):
Few UI comments: 
* Inactive first tab has too low border-radius
* When first tab is open, you can see the light blue container behind (will fix automatically when border radius is fixed) 
* Hover state color for the horizontal line should be darker blue 
* The vertical line between tabs is too long
* When in mobile and only two tabs, the tabs should be visible and not transition into select box

libs/island-ui/core/src/lib/Tabs/Tabs.tsx
```diff
@@ -80,58 +80,106 @@ export const Tabs: FC<React.PropsWithChildren<TabInterface>> = ({
...
+    Tab value can be either an id or an index.
+  */
+  const getSelectedTabIndex = (): number => {
+    const id = tab.selectedId
+
+    if (!id) {
+      return -1
+    }
+
+    //Assuming that the id is a string of some king
```
disaerna, line 93: 👑

libs/island-ui/core/src/lib/Tabs/Tabs.tsx
```diff
@@ -66,7 +66,7 @@ export const Tabs: FC<React.PropsWithChildren<TabInterface>> = ({
   const [isMobile, setIsMobile] = useState(false)
 
   useEffect(() => {
-    if (width < theme.breakpoints.md) {
+    if (width < theme.breakpoints.lg) {
```
disaerna, line 69: Maybe we should only do this if the tabs are many? Like 5,6 or so? 🤔  and keep me MD breakpoint otherwise? What do you think?

thorkellmani: Lets test it out

disaerna:
Screenshot 2024-09-12 at 22 26 07

🫨

disaerna (approved):

PR #15890, opened 2024-09-04
fix(native-app): update inbox filter headers

thoreyjona:
Updating headers for accordions in inbox filters. Also after updating to correct Regular font then 400 is too thick, should be 300 instead.

## Checklist:

## Summary by CodeRabbit

- **Visual Enhancements**
	- Updated the title typography in the Accordion component for improved prominence.
	- Adjusted the font weight of the Checkbox label when checked, resulting in a lighter appearance.


snaerseljan (approved):
LGTM. Nice job

PR #15911, opened 2024-09-06
fix(web): Youtube id extraction from url

RunarVestmann:
# Youtube id extraction from url

## What

Id from url couldn't be extracted from the following url: https://www.youtube.com/watch?v=lHtAy38Ay8Y&feature=youtu.be

To handle that case I added a fallback that checks for the v parameter of the url

## Checklist:


baering (approved):
LGTM

PR #15918, opened 2024-09-09
feat(j-s): Subpoena delivery and endpoint for updates

unakb:
[Asana](https://app.asana.com/0/1199153462262248/1207538469149119)
[Asana](https://app.asana.com/0/1199153462262248/1208028598246434)

## What
- Send subpoena to police
- Endpoint for updates on subpoena status

## Checklist:

## Release Notes

- **New Features**
  - Introduced a new `Subpoena` module for managing subpoenas within the judicial system.
  - Added functionality for creating, updating, and delivering subpoenas to the police.
  - New API endpoints for retrieving and updating subpoenas.

- **Enhancements**
  - Improved notification system to ensure defendants receive specific notifications regarding police subpoenas.
  - Enhanced error handling when retrieving subpoenas and validating associated entities.

- **Documentation**
  - Updated API documentation to reflect new endpoints and data transfer objects for subpoenas.


apps/judicial-system/backend/src/app/modules/subpoena/internalSubpoena.controller.ts
```diff
@@ -0,0 +1,124 @@
...
+    @Param('defendantId') defendantId: string,
+    @CurrentCase() theCase: Case,
+    @CurrentDefendant() defendant: Defendant,
+    @Body() deliverDto: DeliverDto,
+  ): Promise<DeliverResponse> {
+    this.logger.debug(
+      `Delivering subpoena ${caseId} to police for defendant ${defendantId}`,
+    )
+
+    const courtDate = theCase.dateLogs?.find(
```
unakb, line 101: Use courtdate function

apps/judicial-system/backend/src/app/modules/subpoena/internalSubpoena.controller.ts
```diff
@@ -0,0 +1,124 @@
...
+      (dateLog) => dateLog.dateType === DateType.COURT_DATE,
+    )
+    try {
+      const pdf = await this.pdfService.getSubpoenaPdf(
```
unakb, line 105: Have last 3 optional but otherwise use from the case

apps/judicial-system/backend/src/app/modules/subpoena/internalSubpoena.controller.ts
```diff
@@ -0,0 +1,124 @@
...
+    @CurrentCase() theCase: Case,
+    @CurrentDefendant() defendant: Defendant,
+    @Body() deliverDto: DeliverDto,
+  ): Promise<DeliverResponse> {
+    this.logger.debug(
+      `Delivering subpoena ${caseId} to police for defendant ${defendantId}`,
+    )
+
+    const courtDate = theCase.dateLogs?.find(
+      (dateLog) => dateLog.dateType === DateType.COURT_DATE,
```
unakb, line 102: Should be arraignment date

apps/judicial-system/backend/src/app/modules/subpoena/subpoena.service.ts
```diff
@@ -0,0 +1,134 @@
...
+      const createdSubpoena = await this.policeService.createSubpoena(
+        theCase,
+        defendant,
+        subpoenaFile,
+        user,
+      )
+
+      if (!createdSubpoena) {
+        this.logger.error('Failed to create subpoena file for police')
+        return Promise.resolve({ delivered: false })
```
unakb, line 120: Remove promise resolve

apps/judicial-system/backend/migrations/20240827132504-create-subpoena.js
```diff
@@ -0,0 +1,68 @@
...
+          acknowledged: {
+            type: Sequelize.BOOLEAN,
+            allowNull: true,
+          },
+          acknowledged_date: {
+            type: 'TIMESTAMP WITH TIME ZONE',
+            allowNull: true,
+          },
+          comment: {
+            type: Sequelize.STRING,
```
gudjong, line 54: Consider changing the type to TEXT. String can only contain 256 characters.

apps/judicial-system/backend/src/app/modules/case/case.service.ts
```diff
@@ -597,7 +597,7 @@ export class CaseService {
     ]
   }
 
-  private addMessagesForSubmittedIndicitmentCaseToQueue(
+  private addMessagesForSubmittedIndictmentCaseToQueue(
```
gudjong, line 600: 🙏

apps/judicial-system/backend/src/app/modules/defendant/defendant.service.ts
```diff
@@ -203,6 +203,22 @@ export class DefendantService {
...
+        caseId,
+        [Op.or]: [
+          { nationalId: formattedNationalId },
+          { nationalId: defendantNationalId },
+        ],
+      },
+    })
+
+    if (!defendant) {
+      throw new InternalServerErrorException(
```
gudjong, line 217: This is ok, but note that this case is already handled below in the call to `getUpdatedDefendant`. Therefore, consider removing the lookup and exception here.

One thing to be aware of is that national ids are not strictly unique within a case, so in practice there may be multiple defendants with the same national id. However, we do not expect users to do such stupid things on purpose. If multiple defendants have the same national id, then all will be updated below and an error logged. Execution then proceeds as normal.

apps/judicial-system/backend/src/app/modules/notification/notification.service.ts
```diff
@@ -69,6 +69,14 @@ export class NotificationService {
           ]
         } else {
           messages = [this.getNotificationMessage(type, user, theCase)]
+          theCase.defendants?.forEach((defendant) => {
+            messages.push({
+              type: MessageType.DELIVERY_TO_POLICE_SUBPOENA,
```
gudjong, line 74: As discussed this belongs else where. Perhaps it would make sense to add a TODO comment here to remind us.

apps/judicial-system/backend/src/app/modules/subpoena/internalSubpoena.controller.ts
```diff
@@ -0,0 +1,115 @@
...
+    try {
+      const pdf = await this.pdfService.getSubpoenaPdf(theCase, defendant)
+
+      return await this.subpoenaService.deliverSubpoenaToPolice(
+        theCase,
+        defendant,
+        Base64.btoa(pdf.toString('binary')),
+        deliverDto.user,
+      )
+    } catch (error) {
```
gudjong, line 110: I wonder it this try-catch-retrow is reduntant. The Nestjs framwork will log an error if the controller thows an exception so we will end up with two error messages in the logs.

apps/judicial-system/backend/src/app/modules/subpoena/models/subpoena.model.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  @BelongsTo(() => Case, 'caseId')
+  @ApiPropertyOptional({ type: Case })
+  case?: Case
+
+  @Column({ type: DataType.BOOLEAN, allowNull: true })
+  @ApiPropertyOptional({ type: Boolean })
+  acknowledged?: string
+
+  @Column({ type: DataType.STRING, allowNull: true })
```
gudjong, line 64: As mentioned above, consider using the db type TEXT here to allow for longer strings.

apps/judicial-system/backend/src/app/modules/subpoena/subpoena.service.ts
```diff
@@ -0,0 +1,140 @@
...
+    const {
+      defenderChoice,
+      defenderNationalId,
+      defenderEmail,
+      defenderPhoneNumber,
+      defenderName,
+    } = update
+
+    const [numberOfAffectedRows] = await this.subpoenaModel.update(update, {
+      where: { subpoenaId: subpoena.subpoenaId },
```
gudjong, line 55: Consider adding `caseId` to the where clause to prefent fraudulent subpoena updates. It may not be necessary if this is only used for internal updates.

gudjong (approved):
No showstoppers, but some suggestions.

apps/judicial-system/backend/migrations/20240827132504-create-subpoena.js
unakb: Good point, I'll change it so we don't need to add in another migration for that later

apps/judicial-system/backend/src/app/modules/subpoena/subpoena.service.ts
unakb: The update request that ends up is initialized from RLS, and they don't have our caseId logged for use on their side, so the only way we could use the caseId here would by by looking it up from the subpoenaId which would make it redundant or by having them retrieve or store it on their side. I'm definitely up for discussion about it if it's something we're worried about.

PR #15921, opened 2024-09-09
feat(j-s): Confirmed court records

oddsson:
# Confirmed court records

[Asana](https://app.asana.com/0/1199153462262248/1207702780019295/f)

## What

Create a confirmation stamp on court records in S-cases

## Screenshots / Gifs

Screenshot 2024-09-09 at 13 14 11

## Checklist:

## Summary by CodeRabbit

- **New Features**
  - Introduced functionality for generating PDF documents for legal case confirmations, including indictments, rulings, and court records.
  - Enhanced PDF generation logic to include specific checks for case file categories and conditions.
  - Added a new method for confirming court record documents, improving the visual elements in the generated PDFs.

- **Improvements**
  - Streamlined logic for document confirmation and retrieval, ensuring only appropriate case files are confirmed based on case status and type.


apps/judicial-system/backend/src/app/modules/file/file.service.ts
```diff
@@ -209,7 +210,9 @@ export class FileService {
               date: theCase.rulingDate,
             },
             pdf,
-            CaseFileCategory.RULING,
+            file.category === CaseFileCategory.RULING
+              ? CaseFileCategory.RULING
+              : CaseFileCategory.COURT_RECORD,
```
gudjong, line 215: This can be simplified to be just `file.category`, no need for the condition.

apps/judicial-system/backend/src/app/modules/file/file.service.ts
```diff
@@ -236,33 +239,41 @@ export class FileService {
...
+    // Rulings and court records are only confirmed when a case is completed
+    if (
+      (file.category === CaseFileCategory.RULING ||
+        file.category === CaseFileCategory.COURT_RECORD) &&
+      !isCompletedCase(theCase.state)
+    ) {
+      return false
+    }
+
+    return true
```
gudjong, line 265: This will result in the confirmation of all other case file categories in indictments. I suggest modifying the two case file category conditionals above so that you can explicitely return `true` there and the return `false` here at the end of the functionl

gudjong (changes requested):

apps/judicial-system/backend/src/app/modules/file/file.service.ts
oddsson: Thanks 🙏

apps/judicial-system/backend/src/app/modules/file/file.service.ts
oddsson: Ah! Good catch thanks 🙏

gudjong (approved):
Nice

PR #15932, opened 2024-09-10
feat(auth-admin): Delegation reference id

GunnlaugurG:
## What

Add `referenceId` to the `delegation` table 

## Why

So we can reference a ZenDesk Id when a paper delegation is registered

## Screenshots / Gifs

## Checklist:

## Summary by CodeRabbit

- **New Features**
	- Introduced an optional `referenceId` property to delegation records, enhancing tracking and linking capabilities.
	- Updated the database schema to include the `reference_id` column in the `delegation` table, allowing for improved data management.

- **Bug Fixes**
	- Ensured that the migration process for adding and removing the `reference_id` column maintains database integrity.


libs/auth-api-lib/src/lib/delegations/models/delegation.model.ts
```diff
@@ -80,6 +80,12 @@ export class Delegation extends Model<
   @ForeignKey(() => Domain)
   domainName!: CreationOptional<string>
 
+  @Column({
+    type: DataType.STRING,
+    allowNull: true,
+  })
+  referenceId?: string
```
saevarma, line 87: We should add comment how this field is used. And as this is the source DB model, I think it suffice doing it here.

saevarma (approved):
🚀

PR #15972, opened 2024-09-12
fix(native-app): decrease size of vehicle card on home page

thoreyjona:
## What

Vehicle card in widget on home screen was set to have minHeight 176px to accommodate vehicles with long names that would wrap into two lines. (We want all vehicle cards in the widget to be of the same size, that is the reason we set minHeight to the size of the biggest possible card). This was causing really big cards so we decided to truncate the name on the home screen. They are not truncated on the vehicles screen, then they are dynamically sized based on content.

## Screenshots / Gifs
Before:
Screenshot 2024-09-12 at 09 45 47
After:
Screenshot 2024-09-12 at 09 45 12

## Checklist:


snaerseljan (approved):
Awesome much better looking

PR #15977, opened 2024-09-12
fix(native-app): view pager now swipes correct size of elements in home widgets

thoreyjona:
## What

Sending itemWidth into ViewPager to scroll whole elements correctly.

In some rare case when I was logging in and out of different users I got an error that crashed the app `Invariant Violation: inputRange must be monotonically non-decreasing 0,345,301`. After debugging I figured it is somehow that the ViewPager does not have all values updated yet or something similar so the middle number is too big and that crashes the app. **Note:** This does not happen every time and is rather hard to reproduce (and is always fixed on a reload if it happens) so after trying some fixes I decided to sort the inputs in case we have this situation just to make sure the app will not crash. (Did not see any change in the UI after this, this is more like to have "belti og axlabönd" just if case that this happens)

## Checklist:

## Summary by CodeRabbit

- **New Features**
	- Introduced a consistent width calculation for items in the ViewPager across multiple modules, improving layout consistency and responsiveness.
	- Added `itemWidth` prop to the ViewPager component for enhanced flexibility.

- **Bug Fixes**
	- Implemented sorting for the `inputRange` array in the ViewPager to enhance stability and prevent potential crashes.


snaerseljan (approved):
Nice work.

PR #15984, opened 2024-09-12
feat(web): FAQ List and Accordion - Set single expand to false

RunarVestmann:
# FAQ List and Accordion - Set single expand to false

## What

* Prevent an item from closing when another gets opened

## Why

* When you have an item open and open another often the scroll takes you away from what you just opened
* We shouldn't assume that the user wants to close other items when opening a new one

### Before

https://github.com/user-attachments/assets/273210ec-ffa1-49e4-a0c7-4db6d979362d

### After

https://github.com/user-attachments/assets/bed5bb3a-ec6a-409d-958a-ae2766280f1a

## Summary by CodeRabbit

- **New Features**
	- The Accordion component now supports multiple items being expanded simultaneously with the new `singleExpand` property set to `false`.


mannipje (approved):

disaerna (approved):
Love this change 👏🏼  I thought about it when viewing the "Tengjum ríkið" page as well 💯

PR #15996, opened 2024-09-13
feat(j-s): Enable setting civil claimants

oddsson:
# Enable setting civil claimants

[Asana](https://app.asana.com/0/1199153462262248/1208167750326140/f)

## What

- Enable setting civil claimants and their legal advocates.
- Refactor the component responsible for setting defender info. We now talk about _advocates_. This is more flexible some legal advocates in our system are not defending anyone, but rather act as a spokesperson or legal rights protectors. 

## Screenshots / Gifs

Screenshot 2024-09-24 at 11 04 05

## Checklist:

## Release Notes

- **New Features**
	- Introduced `InputAdvocate` component to replace the previous `DefenderInput` for enhanced advocate information handling.
	- Added support for multiple advocate types, including 'defender', 'spokesperson', and 'legal rights protector'.
	- Implemented new GraphQL mutations for creating, updating, and deleting civil claimants.

- **Enhancements**
	- Improved civil claimant management with new state variables and rendering logic in the processing workflow.
	- Expanded input fields and validation for civil claimants, ensuring better data handling.
	- Added new error messages for civil claimant operations to enhance user feedback.

- **Bug Fixes**
	- Simplified logic in various components for better performance and maintainability.

- **Documentation**
	- Updated message strings to support new civil claimant functionalities.


apps/judicial-system/web/src/components/Inputs/InputAdvocate.tsx
```diff
@@ -32,12 +32,15 @@ import {
 } from '@island.is/judicial-system-web/src/utils/hooks'
 import { Validation } from '@island.is/judicial-system-web/src/utils/validate'
 
-import { defenderInput as m } from './DefenderInput.strings'
+import useCivilClaimants from '../../utils/hooks/useCivilClaimants'
```
unakb, line 35: Please import from index like other hooks in line 32

gudjong (approved):
I like where you are going with this PR, creating a shared component to handle lawyer lookups. However, I feel the shared component knows too much about the components that use it. A better approach would be to make the component oblivious of the using components and inject all relevant update methods.

I think the component works as is and the changes suggested above chould be done later as part of a cooldown week.

apps/judicial-system/web/src/components/Inputs/InputAdvocate.tsx
```diff
@@ -80,34 +118,64 @@ const DefenderInput: FC<Props> = ({
...
+      clientId?: string | null,
+    ) => {
       let updatedLawyer = {
         defenderName: '',
         defenderNationalId: '',
         defenderEmail: '',
         defenderPhoneNumber: '',
       }
 
+      let updatedSpokesperson = {
```
unakb, line 133: I understand what you were thinking with this component, but it feels very complicated and confusing to read because of all the different handlings of civilClaimants vs. defendants and the 3 different types of advocates. 
I can see why you wanted to refactor and create a separate table for advocate info instead of storing it in 3 different places, all this logic is giving me a slight headache 😅
I don't think we should necessarily change it at this point though, if it works let's just get it out and keep in mind we'd like to simplify advocate handling overall from top to bottom when we have time.

gudjong: My suggestion is not to combine all the different types of advocates in a new table, but rather to inject relevant methods into the lawyer lookup component. Creating a new table would in my view combine unrelated entities just because the all come from the lawyer api. Too me, that sounds similar to creating a person table to store all human actors in the judicial system, i.e. defendants, prosecutors, judges, defenders, spokespersons, etc, just because they all have a national id.

unakb: Yes, good point. I think we should still reconsider how we handle defender information as we have fields for that both in the case table and defendants table, which seems redundant. The civilClaimants spokesperson info is unrelated enough so that could be kept as is.

gudjong: Absolutely, we have been talking about that for some time. Recently, we added the possibility to have defenders on all R-Cases and a logical next step would be to handle defenders the same way in both R- and S-Cases.

unakb (approved):

PR #16005, opened 2024-09-13
feat(auth-admin): General mandate delegation type

GunnlaugurG:
## What

Added General mandate delegation type with 1 to 1 connection to custom delegations. 

## Why

So we can use this for 'Allsherjarumboð' instead of the Custom delegation type

## Screenshots / Gifs

## Checklist:

## Summary by CodeRabbit

- **New Features**
	- Introduced a new delegation type, "GeneralMandate," enhancing authorization delegation capabilities.
	- Improved handling of delegation types to ensure consistency between custom and general mandate types during creation and removal.

- **Bug Fixes**
	- Updated logic to exclude "General Mandate" from user-facing results, ensuring a cleaner UI experience.

- **Documentation**
	- Expanded the `AuthDelegationType` enum to include the new "GeneralMandate" type for better categorization.


libs/api/domains/auth/src/lib/models/delegation.model.ts
```diff
@@ -34,6 +34,8 @@ const exhaustiveCheck = (param: never) => {
         return CustomDelegation
       case AuthDelegationType.LegalRepresentative:
         return LegalRepresentativeDelegation
+      case AuthDelegationType.GeneralMandate:
+        return CustomDelegation
```
saevarma, line 38: You can join the `case` when returning the same value

```
case AuthDelegationType.Custom:
case AuthDelegationType.GeneralMandate:
  return CustomDelegation
```

libs/auth-api-lib/src/lib/clients/clients.service.ts
```diff
@@ -620,17 +618,32 @@ export class ClientsService {
       )
     }
 
-    return Promise.all(
-      delegationTypes.map((delegationType) =>
-        this.clientDelegationType.upsert(
+    const upsertPromises = delegationTypes.map((delegationType) => {
```
saevarma, line 621: Instead of complicating this map you could do a check if `delegationTypes` includes `AuthDelegationType.Custom` just overwrite the `delegationTypes` with new array including the `GeneralMandate` (shallow copying the array just in case, so its not modifing the source array in place if it would cause issues in the source context)

```
if (delegationTypes.includes(AuthDelegationType.Custom)) {
  delegationTypes = [
    ...delegationTypes,
   AuthDelegationType.GeneralMandate
  ]
}

// Rest should be the same
```

libs/auth-api-lib/src/lib/resources/admin/admin-scope.service.ts
```diff
@@ -459,17 +459,33 @@ export class AdminScopeService {
...
-        this.apiScopeDelegationType.upsert(
-          {
-            apiScopeName,
-            delegationType,
-          },
-          { transaction },
-        ),
-      ),
-    )
+    delegationTypes.map((delegationType) => {
```
saevarma, line 462: You should do the same here just check if the array includes custom and construct the new shallow copied array with GeneralMandate added.

libs/auth-api-lib/src/lib/clients/clients.service.ts
```diff
@@ -687,16 +700,31 @@ export class ClientsService {
       )
     }
 
-    return Promise.all(
-      delegationTypes.map((delegationType) =>
-        this.clientDelegationType.destroy({
+    const destroyPromises = delegationTypes.map((delegationType) => {
```
saevarma, line 703: You should also be able to do it here.

libs/auth-api-lib/src/lib/resources/admin/admin-scope.service.ts
```diff
@@ -532,6 +548,33 @@ export class AdminScopeService {
       : undefined
 
     // remove delegation type rows
+    delegationTypes.map((delegationType) => {
```
saevarma, line 551: And here

saevarma (dismissed):

baering (approved):

PR #16009, opened 2024-09-13
fix(native-app): fix android crash due to borderRadius set to string value

thoreyjona:
Sibbi experienced that the app crashed for him with the new home screen widgets. After looking at datadog I found an error that was introduced [today](https://app.datadoghq.eu/rum/error-tracking?query=@application.id:2736367a-a841-492d-adef-6f5a509d6ec2%20&issueId=dc1d14d0-70f6-11ef-878f-da7ad0900005&source=all&from_ts=1725630312857&to_ts=1726235112857&live=false). And after reading about it I found out it is not allowed to set borderRadius to string on android devices when using inline styling (Seems to be okay when using styled components template strings). Since the `border.radius` in theme is a string it is not allowed to use that value when using inline styles. 

## Checklist:

## Summary by CodeRabbit

- **New Features**
	- Introduced a new styled component, `Cell`, for improved styling management in the Vehicle Item display.
- **Bug Fixes**
	- Updated the `borderRadius` in the General Card Skeleton for a more consistent appearance across various contexts.


snaerseljan (approved):
Jebus Android. Nice one

PR #16028, opened 2024-09-17
feat(web): Add default header for HMS organization

mannipje:
# Add default header for HMS organization

## What

Make it possible to use default header for HMS organization via config.

## Why

A design that was approved by Digital Iceland

### Before
image

### After
image

## Checklist:

## Summary by CodeRabbit

- **New Features**
	- Enhanced theme configuration logic to support additional header conditions.
	- Improved rendering flexibility for the OrganizationHeader based on header type.


apps/web/components/Organization/Wrapper/OrganizationWrapper.tsx
```diff
@@ -183,16 +183,15 @@ export const getThemeConfig = (
       ? 'organization'
       : 'default'
 
-  if (blueberryThemes.includes(theme ?? ''))
+  if (blueberryThemes.includes(theme ?? '') && !usingDefaultHeader)
```
RunarVestmann, line 186: Could we maybe do an early return since in all cases if the default header is chosen then we'd like to use the light mode header variant, right?

RunarVestmann (approved):

PR #16053, opened 2024-09-18
fix(j-s): Merge Case Access

gudjong:
# Merge Case Access

[Sameina mál - Verjandi/sækjandi á að sjá gögn úr sameinuðum málum](https://app.asana.com/0/1199153462262248/1208304796003469/f)

## What

- Allows prosecutors to access uploaded case files and generated pdf files in merged indictment cases even if the merged case was created by a different prosecutor's office.
- Allows defenders to access uploaded case files and generated pdf files in merged indictment cases even if they are not assigend as defenders in the merged case.
- Adds missing case file categories fo merged cases so they can be viewed by all in the client.
- Limites the indictments accessible to the public prosecutors office to completed indictment cases that ended with a fine or a ruling and have been sent to the public prosecutors office for review.
- Adds laws broken components to a few screen to follow design in Figma.
- This PR does not handle merge chains, i.e. if a case is merged with another case which later is merged with another case and so on. This will be handled in a later PR.

## Why

- Verified bugs.

## Screenshots / Gifs

https://github.com/user-attachments/assets/aa76c6c5-57d4-4563-b0cd-9eee7c80c5d5

## Checklist:

## Release Notes

- **New Features**
	- Enhanced API to support retrieval of signed URLs for merged cases.
	- Added new fields in GraphQL queries for more detailed case file information, including `modified`, `type`, and `submittedBy`.
	- Introduced `MergedCaseExistsGuard` for improved access control on merged cases.

- **Improvements**
	- Expanded functionality of components to manage and display connected case files more effectively.
	- Updated UI logic to conditionally display information based on case attributes like laws broken and merged cases.
	- Enhanced data retrieval processes for cases, including support for indictment counts and filtering by case states.

- **Bug Fixes**
	- Removed unnecessary props from components to streamline data handling.


apps/judicial-system/api/src/app/modules/file/file.controller.ts
```diff
@@ -54,22 +54,30 @@ export class FileController {
     )
   }
 
-  @Get('caseFilesRecord/:policeCaseNumber')
+  @Get([
```
unakb, line 57: I did not realize this was possible 😮 interesting

gudjong: I think we should use this more as it could eliminate some code duplications.

apps/judicial-system/backend/src/app/modules/case/guards/mergedCaseExists.guard.ts
```diff
@@ -0,0 +1,44 @@
...
+    const theCase: Case = request.case
+
+    if (!theCase) {
+      throw new InternalServerErrorException('Missing case')
+    }
+
+    const mergedCaseId = request.params.mergedCaseId
+
+    if (!mergedCaseId) {
+      return true
```
oddsson, line 28: Should this not be `return false`?

apps/judicial-system/web/src/components/PdfButton/PdfButton.tsx
```diff
@@ -7,7 +7,8 @@ import { UserContext } from '../UserProvider/UserProvider'
 import * as styles from './PdfButton.css'
 
 interface Props {
-  caseId: string
+  caseId?: string
+  connectedCaseParentId?: string
```
oddsson, line 11: Could you please add a comment explaining this prop and why you'd use it? I think it's a bit unclear and you'd have to know the implementation details of this component to know what this prop does.

oddsson (changes requested):
Nice, just a few wonderings ;)

apps/judicial-system/backend/src/app/modules/case/guards/mergedCaseExists.guard.ts
gudjong: No, this guard only applies if there is a `mergeCaseId`. The idea here is that if we provide a `mergeCaseId` then we replace the case with the merged case to allow us to access it in the relevant endpoint. Basically, `caseId` is used for access control to determine if the user should really be allowed to access the case. `mergeCaseId` on the other hand refers to the case we are actually accessing.

Example:
Case B is merged into case A. Case B is now closed and case A has a reference to it.
If I am the prosecutor assigned to case A, but case B came from a different prosecutors office, then I still need to be able to access case B. This guard gives me that access.
If I am a defender or spokesperson assigned to case A, but not case B, then I still need to be able to access case B. This guard gives me that access.

I hope this explaines.

apps/judicial-system/web/src/components/PdfButton/PdfButton.tsx
gudjong: Will do. Actually, using the term "connected case" in the client and "merged case" in the backend may cause some confusion, but I decided to follow the term already used in the client.

gudjong:
> Nice, just a few wonderings ;)

Good comments, I have addressed them all.

apps/judicial-system/backend/src/app/modules/case/guards/mergedCaseExists.guard.ts
oddsson: Ah, got it! Thanks

oddsson (approved):

PR #16068, opened 2024-09-18
feat(ids-admin): Delegation-Delegation-Type

GunnlaugurG:
## What

Add `delegation_delegation_type` to db and support General Mandate delegations (Allsherjarumboð) 

## Why

So the new delegations-admin module can create GeneralMandate delegation

## Screenshots / Gifs

## Checklist:

## Summary by CodeRabbit

- **New Features**
	- Introduced support for "General Mandate" delegation type across various components and services.
	- Enhanced delegation handling, allowing for improved retrieval and management of delegation data.
	- Added new methods to fetch valid and available general mandates based on national IDs.

- **Bug Fixes**
	- Improved test coverage for delegation functionalities, ensuring accurate behavior of API endpoints related to general mandates.


jonarnarbriem (approved):

libs/auth-api-lib/src/lib/delegations/delegation-scope.service.ts
```diff
@@ -161,6 +167,47 @@ export class DelegationScopeService {
     })
   }
 
+  private findValidGeneralMandateTo(
+    toNationalId: string,
+    fromNationalId: string,
+  ): Promise<Delegation[]> {
```
saevarma, line 173: We should let this be focused on finding scopes

```suggestion
  private findValidGeneralMandateScopesTo(
    toNationalId: string,
    fromNationalId: string,
  ): Promise<String[]> {
```

libs/auth-api-lib/src/lib/delegations/delegation-scope.service.ts
```diff
@@ -161,6 +167,47 @@ export class DelegationScopeService {
...
+                  where: {
+                    enabled: true,
+                  },
+                },
+              ],
+            },
+          ],
+        },
+      ],
+    })
```
saevarma, line 208: I think its better to split this into:

1. Query `DelegationModel` to check if a valid GeneralMandate delegation is found. If not found return empty array
2. If valid then find all api scopes that has GeneralMandate relation.

libs/auth-api-lib/src/lib/delegations/delegations-incoming.service.ts
```diff
@@ -182,6 +188,17 @@ export class DelegationsIncomingService {
       )
     }
 
+    // This has a provider of Custom but needs to be handled separately
```
saevarma, line 191: Custom is also provider of GeneralMandate, we should explicitly check the custom delegation type in the if above.

libs/auth-api-lib/src/lib/delegations/models/delegation-delegation-type.model.ts
```diff
@@ -0,0 +1,61 @@
...
+  DataType,
+  CreatedAt,
+  UpdatedAt,
+  BelongsTo,
+} from 'sequelize-typescript'
+
+import { Delegation } from './delegation.model'
+import { DelegationTypeModel } from './delegation-type.model'
+
+@Table({
```
saevarma, line 20: You can and should add the delegation delegation_type unique constraint in the `@Table` decorator, so tests run with the same constraints as prod.

libs/portals/shared-modules/delegations/src/lib/messages.ts
saevarma, line 1: Remember to extract-strings and add the translation in Contentful.

libs/auth-api-lib/src/lib/delegations/delegations-incoming-custom.service.ts
```diff
@@ -84,6 +83,33 @@ export class DelegationsIncomingCustomService {
...
+          validity: DelegationValidity.INCLUDE_FUTURE,
+        },
+        useMaster,
+      )
+
+    const delegationDTOs = delegations.map((d) =>
+      d.toDTO(AuthDelegationType.GeneralMandate),
+    )
+
+    return delegationDTOs.map((d) => {
```
saevarma, line 103: Here we are looping 2x through the same array, we should join these maps to safe the extra iteration.

libs/auth-api-lib/src/lib/delegations/delegations-incoming-custom.service.ts
```diff
@@ -151,6 +178,117 @@ export class DelegationsIncomingCustomService {
...
+      nationalId,
+      validity,
+    }: FindAllValidIncomingOptions & {
+      validity: DelegationValidity
+    },
+    useMaster = false,
+  ): Promise<{ delegations: Delegation[]; fromNameInfo: FromNameInfo[] }> {
+    const whereOptions = getScopeValidityWhereClause(validity)
+
+    const delegations = await this.delegationModel.findAll({
```
saevarma, line 230: We don't need to the join with ApiScope

libs/auth-api-lib/src/lib/delegations/delegations-incoming-custom.service.ts
```diff
@@ -151,6 +178,117 @@ export class DelegationsIncomingCustomService {
...
+  private async findAllIncomingGeneralMandates(
+    {
+      nationalId,
+      validity,
+    }: FindAllValidIncomingOptions & {
+      validity: DelegationValidity
+    },
+    useMaster = false,
+  ): Promise<{ delegations: Delegation[]; fromNameInfo: FromNameInfo[] }> {
+    const whereOptions = getScopeValidityWhereClause(validity)
```
saevarma, line 228: We are not using the `validFrom` fields for GeneralMandates so this becomes a bit weird. As `findAllValidGeneralMandates` uses `INCLUDE_FUTURE` which sounds strange as it should return only valid `NOW`. But you are doing that to work   around the `validFrom` check introduced by this helper.

So I would suggest to not use this helper here, and just arrange the `whereOptions` here to handle `ALL` and `NOW`

saevarma (approved):
🚀

PR #16074, opened 2024-09-19
feat(auth-admin): Create paper delegation zendesk integration

magnearun:
## What 
Validate that the Zendesk ticket referenced in delegation is valid, status should be Solved, tags should include the 'umsokn_um_umboð_a_mínum_síðum' tag (we might change the tag later to not include the ´minum_siður´ part.
Also the nationalIds in the ticket should match the ones on the delegation

Also makes small changes needed to get a valid openApi.yaml to use with Postman

## Why

Validate papar delegations info from Zendesk

## Checklist:

### DIFF
--- a/apps/services/auth/admin-api/src/app/v2/clients/me-clients.controller.ts
+++ b/apps/services/auth/admin-api/src/app/v2/clients/me-clients.controller.ts
@@ -7,6 +7,7 @@ import {
   Post,
   UseGuards,
   Delete,
+  Query,
 } from '@nestjs/common'
 import { ApiSecurity, ApiTags } from '@nestjs/swagger'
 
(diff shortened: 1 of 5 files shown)


apps/services/auth/admin-api/src/app/v2/clients/me-clients.controller.ts
```diff
@@ -74,7 +75,7 @@ export class MeClientsController {
     @CurrentUser() user: User,
     @Param('tenantId') tenantId: string,
     @Param('clientId') clientId: string,
-    @Param('includeArchived') includeArchived?: boolean,
+    @Query('includeArchived') includeArchived?: boolean,
```
GunnlaugurG, line 78: Was this intentional ?

GunnlaugurG (commented):
LGTM

apps/services/auth/admin-api/src/app/v2/clients/me-clients.controller.ts
magnearun: yes, it should be a query param and to a param, the open api yaml file had errors, so it would not work with Postman, me and Sævar made these changes to fix them.

PR #16103, opened 2024-09-20
feat(core): Integrate openapi-ts generator

eirikurn:
## What

This PR provides a proof-of-concept integrations of [@hey-api/openapi-ts](https://github.com/hey-api/openapi-ts) and @hey-api/client-fetch. As an example, it upgrades the hms-loans client.

## Why

We're considering this openapi client generator as an alternative to openapi-generator. The main cons with openapi-generator are:

- It is island.is's only dependency on Java, which complicates onboarding / CI.
- It is a massive code generation system which supports [dozens of languages and frameworks](https://openapi-generator.tech/docs/generators). TypeScript support may suffer.
- It has issues with multiple OpenAPI definitions. Our monorepo has many instances of modifying OpenAPI documents, either manually or with jq, to circumwent its limitations.
- These issues are partly due to openapi-generator's design of generating more elaborate request/response handling code.

Compared to this, `@hey-api/openapi-ts`:

- Written in TypeScript and runs natively on NodeJS.
- Has a more generic fetch-based request/response handling and focuses on providing correct TypeScript types for input/output.
- Can provide more correct types for different kinds of responses.
- Is around 35% faster to generates types.
- Is faster at runtime since it does not process requests/responses at all.
- Creates standalone api functions by default which may give us smaller JS bundles with tree shaking.

The main cons with openapi-ts compared to openapi-generator are:

- Since openapi-ts does not process requests/responses it does not support parsing or serialising Date instances. They are passed around as strings. Generally this is fine, as that's how we receive dates in JSON from the OpenAPI server, and also how we send them in JSON to the GraphQL client. 

- We can't configure and inject api client classes as easily with the NestJS custom provider system. I.e. we can't just configure, provide and export the generated "FasteignirApi" instance for any NestJS module to use. This PR handles this by configuring the default client in `onModuleInit` and wrapping all API call logic in a client service class. There are alternative options using custom clients, classes (with static functions) and custom provider keys.


eirikurn:
Merged as #20279

mrlubos:
@eirikurn did you try [transformers](https://heyapi.dev/openapi-ts/plugins/transformers) for your dates issue? Depending on how complex your schemas are, it might be good enough

I don't fully understand the other disadvantage you list, but happy to discuss if it's something that could be improved

PR #16155, opened 2024-09-25
chore(j-s): Rollback subpoena entry if we get an error while creating it

unakb:
## What

Handle police creation errors by rolling back a transaction so we don't flood the db with failed subpoenas

## Checklist:

## Summary by CodeRabbit

- **New Features**
	- Enhanced subpoena creation process to support database transactions, improving data integrity during operations.
  
- **Bug Fixes**
	- Improved atomicity of subpoena creation and updates, ensuring all operations succeed or fail together.

### DIFF
--- a/apps/judicial-system/backend/src/app/modules/subpoena/subpoena.service.ts
+++ b/apps/judicial-system/backend/src/app/modules/subpoena/subpoena.service.ts
@@ -128,11 +128,6 @@ export class SubpoenaService {
           user,
         )
 
-        if (!createdSubpoena) {
-          this.logger.error('Failed to create subpoena file for police')
-          return { delivered: false }
-        }
-
         await this.subpoenaModel.update(
           { subpoenaId: createdSubpoena.subpoenaId },
           { where: { id: subpoena.id }, transaction },
@@ -141,7 +136,10 @@ export class SubpoenaService {
 
       return { delivered: true }
     } catch (error) {
-      this.logger.error('Error delivering subpoena to police', error)
+      this.logger.error(
+        'Error while attempting to deliver subpoena to police',
+        error,
+      )
       return { delivered: false }
     }
   }
--- a/apps/judicial-system/backend/src/app/modules/subpoena/subpoena.service.ts
+++ b/apps/judicial-system/backend/src/app/modules/subpoena/subpoena.service.ts
@@ -35,7 +35,7 @@ export class SubpoenaService {
     defendant: Defendant,
     transaction: Transaction,
   ): Promise<Subpoena> {
-    return await this.subpoenaModel.create(
+    return this.subpoenaModel.create(
       {
         defendantId: defendant.id,
         caseId: defendant.caseId,
--- a/apps/judicial-system/backend/src/app/modules/subpoena/subpoena.service.ts
+++ b/apps/judicial-system/backend/src/app/modules/subpoena/subpoena.service.ts
@@ -31,11 +31,17 @@ export class SubpoenaService {
     @Inject(LOGGER_PROVIDER) private readonly logger: Logger,
   ) {}
 
-  async createSubpoena(defendant: Defendant): Promise<Subpoena> {
-    return await this.subpoenaModel.create({
-      defendantId: defendant.id,
-      caseId: defendant.caseId,
-    })
+  async createSubpoena(
+    defendant: Defendant,
+    transaction: Transaction,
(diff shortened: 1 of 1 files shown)


apps/judicial-system/backend/src/app/modules/subpoena/subpoena.service.ts
```diff
@@ -112,24 +118,26 @@ export class SubpoenaService {
...
+          theCase,
+          defendant,
+          subpoenaFile,
+          user,
+        )
+
+        if (!createdSubpoena) {
+          this.logger.error('Failed to create subpoena file for police')
+          return { delivered: false }
+        }
```
gudjong, line 134: I am not absolutely sure, but I don't think this will rollback the transaction. I think you need to throw an exception for that.

gudjong: I checked the documentation and my understanding is correct. You need to throw to trigger a rollback. The catch clause below returns the right delivered status so that should be fine.

gudjong (changes requested):
A minor fix needed to trigger rollback in case of failure.

apps/judicial-system/backend/src/app/modules/subpoena/subpoena.service.ts
unakb: Ok weird because I tried it locally and it rolled back! I'll try again tomorrow to confirm but I can always add in an exception just in case

gudjong: Hér er skjölunin fyrir þetta:

   * Start a transaction. When using transactions, you should pass the transaction in the options argument
   * in order for the query to happen under that transaction
   *
   * ```js
   *   try {
   *     const transaction = await sequelize.transaction();
   *     const user = await User.findOne(..., { transaction });
   *     await user.update(..., { transaction });
   *     await transaction.commit();
   *   } catch(err) {
   *     await transaction.rollback();
   *   }
   * })
   * ```
   *
   * A syntax for automatically committing or rolling back based on the promise chain resolution is also
   * supported:
   *
   * ```js
   * try {
   *   await sequelize.transaction(transaction => { // Note that we pass a callback rather than awaiting the call with no arguments
   *     const user = await User.findOne(..., {transaction});
   *     await user.update(..., {transaction});
   *   });
   *   // Committed
   * } catch(err) {
   *   // Rolled back
   *   console.error(err);
   * }
   * ```

Þú getur annað hvort búið til transaction og séð sjálf um commit og rollback, eða sent callback inn í transaction fallið og látið transaction fallið sjá um commit og rollback. Til að callback fallið geti rollbackað er það væntanlega með try/catch utan um callbackið og þess vegna þarf throw.

unakb: Aa ok ég veit af hverju þetta virkaði þegar ég prófaði locally - af því þá var createsubpoena kallið að kasta exception ekki skila villu eða tómu svari þannig ég datt ekkert inn í if(!created)

unakb: Í rauninni gæti ég bara tekið þessa if setningu alveg út, er að kasta exceptions ef eitthvað klikkar í lögreglukallinu þannig þetta er eiginlega unreachable kóði

apps/judicial-system/backend/src/app/modules/subpoena/subpoena.service.ts
```diff
@@ -31,11 +31,17 @@ export class SubpoenaService {
...
-  async createSubpoena(defendant: Defendant): Promise<Subpoena> {
-    return await this.subpoenaModel.create({
-      defendantId: defendant.id,
-      caseId: defendant.caseId,
-    })
+  async createSubpoena(
+    defendant: Defendant,
+    transaction: Transaction,
+  ): Promise<Subpoena> {
+    return await this.subpoenaModel.create(
```
gudjong, line 38: You don't have to await here, so you may consider removing the await.

gudjong (approved):
Nice, one suggestion.

apps/judicial-system/backend/src/app/modules/subpoena/subpoena.service.ts
unakb: Done ✔️

PR #16194, opened 2024-09-27
chore(driving-license): add delivery price if pickup = post selected

stjanilofts:
image

## Summary by CodeRabbit

- **New Features**
	- Enhanced payment charge calculation for driving license applications, improving clarity and handling of pickup scenarios.
	- Introduced constants for better management of charge item codes, including a new `DELIVERY_FEE`.

- **Bug Fixes**
	- Adjusted logic to prevent potential errors from undefined values in total price calculations.

- **Refactor**
	- Simplified method signatures for improved data access and readability.


libs/application/templates/driving-license/src/forms/draft/subSectionSummary.ts
```diff
@@ -159,23 +159,39 @@ export const subSectionSummary = buildSubSection({
...
             const targetCode =
-              answers.applicationFor === B_TEMP
+              application.answers.applicationFor === B_TEMP
                 ? 'AY114'
-                : answers.applicationFor === BE
+                : application.answers.applicationFor === BE
                 ? 'AY115'
+                : application.answers.applicationFor === B_FULL_RENEWAL_65
+                ? 'AY113'
                 : 'AY110'
```
kksteini, line 175: Do we want to create a function for this decider code so that we can maintain it more clearly elsewhere? Like in form/utils or something?

Also, can this not be simplified with something like:
```ts
const codeLookup: Record<string, string> = {
   [B_TEMP]: 'AY114',
   [BE]: 'AY115',
   [B_FULL_RENEWAL_65]: 'AY113'
}
 ```
 
 So that it can be accessed with something like
 ```ts
 const targetCode = codeLookup[application.answers.applicationFor] ?? 'AY110'
 ```
 
 Just wondering if that makes sense and if that creates a more maintainable base :)

kksteini (approved):

PR #16207, opened 2024-09-30
chore(documents-v2): Catch error and log warning

thordurhhh:
## What

Catch error `/page` and log warning.

## Why

Too many errors.
An error here is expected in some cases. It indicates a user error. Return the default of `1` which in most cases would be the number from the service.

## Checklist:

## Summary by CodeRabbit

- **New Features**
	- Enhanced error handling for retrieving page numbers, ensuring a default value of `1` is returned in case of an error.
- **Bug Fixes**
	- Improved robustness of the `getPageNumber` method to prevent failures during document retrieval.


libs/api/domains/documents/src/lib/documentV2.service.ts
```diff
@@ -201,14 +201,26 @@ export class DocumentServiceV2 {
...
-    return {
-      pageNumber: res ?? 1,
+      return {
+        pageNumber: res ?? defaultRes,
+      }
+    } catch (exception) {
+      this.logger.warn('Document page number error', {
+        category: LOG_CATEGORY,
+        documentId,
+        exception,
```
thorkellmani, line 219:
```suggestion
        error: exception,
```

thorkellmani (commented):
One thing

thorkellmani (approved):
Cool

PR #16216, opened 2024-09-30
fix(download-service): Update usage of accesstoken in jwt strategy and download service controllers

snaerseljan:
# Fix download service token usage

## What

Instead of including the accessToken in the POST request body for the download-service controllers, we now extract it via the JWT strategy. The strategy pulls the token from the request headers and provides it through the authorization field in the user context.

## Why

Preparation for BFF since no tokens will be exposed in the client.

## Checklist:

## Summary by CodeRabbit

- **New Features**
	- Simplified authorization process across multiple document retrieval methods by removing the need for access token parameters.

- **Bug Fixes**
	- Improved response header formatting for clarity and consistency in document retrieval responses.

- **Chores**
	- Removed several DTO classes related to access token validation for various document types, streamlining the overall structure.

These changes enhance user experience by simplifying access to documents while maintaining functionality.


saevarma (approved):
Nice! 🚀
But a bit scary change so we need to make sure that the front end code downloading the files keeps working. We need to test this well on dev.

snaerseljan:
> Nice! 🚀 But a bit scary change so we need to make sure that the front end code downloading the files keeps working. We need to test this well on dev.

Yes I have done the test now and this should be solid :)

thordurhhh (approved):
Lovely! Just tested, download on service-portal still good after this change.
🚀🚀

Issue #11858, opened 2023-08-17
Openapi Generator Fails on allOf


kksteini:
# What?

When updating to the latest driving-license-book client we get curious messages when running the external-openapi-generator.
The generation does not succeed fully and a lot of types/models are replaced by `unknownBaseType` which then results in our code not compiling due to typescript errors.

## To reproduce

Have XROAD proxy running.

* yarn nx run clients-driving-license-book:update-openapi-document
* yarn nx run clients-driving-license-book:schemas/external-openapi-generator

> The clientConfig.json generated in the first step is a valid openapi format when passed through a validator.

## Possibly related issues
I've found these issues: 
* https://github.com/OpenAPITools/openapi-generator/issues/2236
* https://github.com/OpenAPITools/openapi-generator/issues/2892
They closely mimic the symptoms of this current issue. Curiously I do not find much on this issue for the javascript client.

## Solutions
* In a recent PR, https://github.com/island-is/island.is/pull/11857, I had to manually edit the allOf blocks in the clientConfig since every one of them only had a single reference entry and they could be omitted. Not a great solution but possible in very certain situations. Advantages: not costly in terms of work. Disadvantages are that this doesn't fix anything, really.

* Update the openapi generator. Haven't looked deeply into this, but if we go from a major version of 1 to 2, then things start breaking pretty quickly. I tried these following versions as an update but they still had the errors present: [1.0.18-5.0.0-beta2 and 1.0.18-4.3.1]

andes-it:
This issue is stale because it has been open for 5 days with no activity. Remove stale label or this will be closed in 1 day.

andes-it:
This issue was closed because no activity was received 1 day after it was marked as stale.

andes-it:
This issue's been twiddling its thumbs for 5 days. Daydreaming, perhaps? Best do something or it's curtains in 1 day. Nudge off the stale label to give it another go.

andes-it:
This issue's shut up shop. Gave it a day post-stale, then decided, "Enough's enough!"

krillijuni:
Sometimes from Syslumenn they will include a property that is only nullable but has no type. You can simply define it as type string to circumvent any type json errors.

PR #16318, opened 2024-10-08
chore(clients-syslumenn): Update client config

kksteini:
# Warning
Please be advised that our openapi generation client is probably in need of updating. We can avoid some problems if we transform the client config heavily, as can be seen in the undesiredly long command for the update-openapi-document directive.

I yet again refer to the issue raised here: https://github.com/island-is/island.is/issues/11858 although I think an update is not being planned.

I would greatly appreciate if anyone knows if there is a simpler solution than transforming incoming client configs to this extent just to avoid config errors.

In this case we are avoiding "AnyType" models not being generated since responses have now been wrapped in "application/json" blocks, triggering the JSON serializer/deserializer which has an issue with items being empty.

## Checklist:

### DIFF
--- a/libs/clients/syslumenn/src/lib/syslumennClient.utils.ts
+++ b/libs/clients/syslumenn/src/lib/syslumennClient.utils.ts
@@ -111,16 +111,16 @@ export const mapVehicleResponse = (response: Okutaeki): VehicleDetail => {
     manufacturer: response.framleidandi ?? '',
     manufacturerType: response.framleidandaGerd ?? '',
     color: response.litur ?? '',
-    dateOfRegistration: response.skraningardagur,
+    dateOfRegistration: response.skraningardagur ?? new Date(),
   }
 }
 
 export const mapShipResponse = (response: Skip): ShipDetail => {
   return {
-    shipRegistrationNumber: response.shipRegistrationNumber.toString(),
+    shipRegistrationNumber: response.shipRegistrationNumber?.toString() ?? '',
     usageType: response.usageType ?? '',
     name: response.name ?? '',
-    initialRegistrationDate: response.initialRegistrationDate,
+    initialRegistrationDate: response.initialRegistrationDate ?? new Date(),
     mainMeasurements: {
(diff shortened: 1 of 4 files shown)


libs/clients/syslumenn/src/lib/syslumennClient.utils.ts
```diff
@@ -479,7 +480,7 @@ export const mapEstateRegistrant = (
     marriageSettlement: syslaData.kaupmaili ?? false,
     office: syslaData.embaetti ?? '',
     caseNumber: syslaData.malsnumer ?? '',
-    dateOfDeath: syslaData.danardagur ?? '',
+    dateOfDeath: syslaData.danardagur ?? new Date(), // Todo: suitable backup for date
```
RunarVestmann, line 483: Is this todo something you'd like to consider addressing?

libs/clients/syslumenn/src/lib/syslumennClient.utils.ts
```diff
@@ -589,9 +590,9 @@ export const mapInheritanceTax = (
   inheritance: ErfdafjarskatturSvar,
 ): InheritanceTax => {
   return {
-    inheritanceTax: inheritance.erfdafjarskattur,
-    taxExemptionLimit: inheritance.skattfrelsismorkUpphaed,
-    validFrom: inheritance.gildirFra,
+    inheritanceTax: inheritance.erfdafjarskattur ?? 0,
+    taxExemptionLimit: inheritance.skattfrelsismorkUpphaed ?? 0,
+    validFrom: inheritance.gildirFra ?? new Date(), // TODO: Figure out a suitable fallback for date
```
RunarVestmann, line 595: Is this todo something you'd like to consider addressing?

libs/clients/syslumenn/src/lib/syslumennClient.utils.ts
kksteini: Decided to remove and keep this as the fallback.

rafnarnason (approved):

RunarVestmann (approved):

PR #16329, opened 2024-10-09
chore(j-s): Send notifications when subpoena changes

oddsson:
# Send notifications when subpoena changes 

[Asana](https://app.asana.com/0/1199153462262248/1208157919477467/f)

## What

Send email notifications to judges and registrars when 

- Service is successful
- Service failed
- Defendant selected defender

## Why

This is useful info for court users.

## Checklist:

## Summary by CodeRabbit

- **New Features**
	- Introduced a new notification system with enhanced capabilities for handling subpoenas.
	- Added support for different notification types and improved handling of prosecutor office identifiers.
	- Added a new `SubpoenaNotificationService` for managing subpoena notifications.
	- New classes for notification bodies and subpoena notifications have been implemented.
	- Updated the `NotificationModule` to include the `SubpoenaNotificationService`.
	- Enhanced notification methods to manage service status checks effectively.

- **Bug Fixes**
	- Updated method signatures to allow for optional parameters, enhancing flexibility in notification handling.

- **Documentation**
	- Updated API tags for clarity regarding limited access subpoenas.


apps/judicial-system/backend/src/app/modules/subpoena/subpoena.service.ts
```diff
@@ -41,13 +44,29 @@ export class SubpoenaService {
...
     private readonly pdfService: PdfService,
+    private readonly messageService: MessageService,
     @Inject(forwardRef(() => PoliceService))
     private readonly policeService: PoliceService,
     @Inject(forwardRef(() => FileService))
     private readonly fileService: FileService,
     @Inject(LOGGER_PROVIDER) private readonly logger: Logger,
   ) {}
 
+  private isSuccessfulServiceStatus(serviceStatus?: ServiceStatus) {
```
unakb, line 55: There is a helper function that already does this in the lib in the defendant.ts file, did you decide not to use that on purpose? If so, why? 😊

apps/judicial-system/backend/src/app/modules/subpoena/subpoena.service.ts
```diff
@@ -41,13 +44,29 @@ export class SubpoenaService {
...
+      serviceStatus &&
+      [
+        ServiceStatus.ELECTRONICALLY,
+        ServiceStatus.DEFENDER,
+        ServiceStatus.IN_PERSON,
+      ].includes(serviceStatus)
+    )
+  }
+
+  private isFailedServiceStatus(serviceStatus?: ServiceStatus) {
```
unakb, line 66: This could be added to the lib as well

apps/judicial-system/backend/src/app/modules/notification/internalNotification.controller.ts
```diff
@@ -65,14 +66,19 @@ export class InternalNotificationController {
...
-    )
+    return notificationDto.prosecutorsOfficeId
+      ? this.institutionNotificationService.sendNotification(
+          notificationDto.type,
+          notificationDto.prosecutorsOfficeId,
+        )
+      : this.subpoenaNotificationService.sendNotification(
+          notificationDto.type,
+          notificationDto.subpoenaId,
+        )
```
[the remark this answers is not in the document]
unakb: This is a good point

apps/judicial-system/backend/src/app/modules/subpoena/subpoena.service.ts
```diff
@@ -128,6 +148,31 @@ export class SubpoenaService {
       )
     }
 
+    if (
```
unakb, line 151: You could simplify this code by creating a const thats either a NotificationType or undefined depending on if/which of these 3 cases are true. Then use that to decide whether to send a message of said type to the queue. That way you dont need to perform the same checks twice and you know both whether to send and which type to send with the same check.

apps/judicial-system/backend/src/app/modules/subpoena/subpoena.service.ts
```diff
@@ -128,6 +148,31 @@ export class SubpoenaService {
...
+                : NotificationType.DEFENDANT_SELECTED_DEFENDER,
+              subpoenaId: subpoena.subpoenaId,
+            },
+          },
+        ])
+        .catch((reason) =>
+          // Tolerate failure, but log
+          this.logger.error('Failed to dispatch notifications', { reason }),
+        )
+    }
```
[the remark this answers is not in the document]
unakb: Oh ok, I hadn't seen this yet haha, same as my comment above.

apps/judicial-system/backend/src/app/modules/notification/subpoenaNotification.service.ts
```diff
@@ -0,0 +1,222 @@
...
+      notificationModel,
+      emailService,
+      intlService,
+      config,
+      eventService,
+      logger,
+    )
+  }
+
+  private async sendServiceSuccessfulNotification(
```
unakb, line 49: We might need to keep in mind that we should only send these notifications once, are we handling that anywhere? We should only have to send these the first time the status is updated. 
While it's true that technically we shouldn't ever get here twice, we should keep in mind that since the initial change is coming from an external system it may behave weirdly and send us info we've already received.

gudjong: There are helper functions that can help with this. In `InternalNotificationService` you will find `hasSentNotification` and `hasReceivedNotification`. These could be moved to the `BaseNotificationService` and used here.

apps/judicial-system/backend/src/app/modules/notification/subpoenaNotification.service.ts
```diff
@@ -0,0 +1,222 @@
...
+    }
+
+    if (!theCase.courtCaseNumber || !theCase.id) {
+      throw new InternalServerErrorException(
+        `Unable to find courtCaseNumber for case ${theCase.id}`,
+      )
+    }
+
+    return theCase
+  }
```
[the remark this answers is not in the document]
oddsson: findBySubpoenaId either finds a subpoena or throws an error. I don't think I need to check again here.

gudjong (changes requested):
Some changes suggested

apps/judicial-system/backend/src/app/modules/notification/subpoenaNotification.service.ts
```diff
@@ -0,0 +1,222 @@
...
+
+  private async sendServiceSuccessfulNotification(
+    subpoenaId?: string,
+  ): Promise<unknown> {
+    if (!subpoenaId) {
+      return
+    }
+
+    const theCase = await this.getCase(subpoenaId)
```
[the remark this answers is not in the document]
oddsson: I'd really like some assistance in reacting to this comment. I've done some changes and now the call stack is as follows

- `/internal/subpoena/:subpoenaID` is called. There is a `SubpoenaExists` guard on the handler in the controller that handles this request.
- The handler places SUBPOENA_NOTIFICATION on the queue and sends the subpoena (from `CurrentSubpoena`) with it. 
- The subpoena flows from there to `subpoenaNotification.service` where it's used to get the caseId to get the case in the getCase function. 

By doing this, I was able to remove the `subpoenaService` dependency from `subpoenaNotification.service` and still use the `SubpoenaExists` guard.

I'd like to use the `caseExists` guard in the same way to eliminate the `caseService` dependency in `subpoenaNotification.service` but that would require me to put the `caseId` in the endpoint url (current endpoint is `/internal/subpoena/:subpoenaId` but would need to be `internal/case/:caseId/subpoena/:subpoenaId`)

Would you agree or am I overcomplicating this?

apps/judicial-system/backend/src/app/modules/notification/subpoenaNotification.service.ts
```diff
@@ -0,0 +1,193 @@
...
+      theCase.id,
+      notificationType,
+      formattedSubject,
+      formattedBody,
+      theCase.judge?.name,
+      theCase.judge?.email,
+      theCase.registrar?.name,
+      theCase.registrar?.email,
+    )
+  }
```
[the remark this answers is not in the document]
oddsson: No thanks. That would be a future iteration.

apps/judicial-system/backend/src/app/modules/notification/subpoenaNotification.service.ts
```diff
@@ -0,0 +1,193 @@
...
+          undefined,
+          true,
+        ),
+      )
+    }
+
+    const recipients = await Promise.all(promises)
+
+    return this.recordNotification(caseId, notificationType, recipients)
+  }
```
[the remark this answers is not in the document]
oddsson: Not doing this since we currently use `Promise.all` for multiple recipients. This would be a future iteration.

apps/judicial-system/backend/src/app/modules/notification/subpoenaNotification.service.ts
```diff
@@ -0,0 +1,196 @@
...
+
+import { EventService } from '../event'
+import { Subpoena } from '../subpoena'
+import { DeliverResponse } from './models/deliver.response'
+import { Notification, Recipient } from './models/notification.model'
+import { BaseNotificationService } from './baseNotification.service'
+import { notificationModuleConfig } from './notification.config'
+import { strings } from './subpoenaNotification.strings'
+
+type subpoenaNotificationType =
```
gudjong, line 26: Good idea. Please consider moving this declaration to the file containing `notification.ts` in the type library and using the type in the dto, controller and service instead of `NotificationType`. That way you avoid the cast from `NotificationType` to `subpoenaNotificationType`. Also, that allows you to use the type in the `SubpoenaService` where the notification messages are sent to the queue.

Please also consider renaming the type to `SubpoenaNotificationType` as that is the naming convention used for types.

apps/judicial-system/backend/src/app/modules/notification/subpoenaNotification.service.ts
```diff
@@ -0,0 +1,196 @@
...
+    return this.recordNotification(caseId, notificationType, recipients)
+  }
+
+  async sendNotification(
+    type: NotificationType,
+    subpoena: Subpoena,
+  ): Promise<DeliverResponse> {
+    try {
+      await this.sendSubpoenaNotification(
+        type as subpoenaNotificationType,
```
gudjong, line 185: This casting can be avoided by making the changes suggested above.

gudjong (approved):
One suggestion

PR #16344, opened 2024-10-09
chore(config): Add shellcheck config file

AndesKrrrrrrrrrrr:
Add `shellcheck` config file, so that developers' own global config isn't used instead.

The config is empty, because we're sticking with plain defaults for now. This file prevents searching for config files higher up the filesystem tree.

## Summary by CodeRabbit

- **Chores**
	- Introduced an empty configuration file for ShellCheck to enforce default settings and enhance security by preventing potential directory traversal issues.


.shellcheckrc
eirikurn, line 1: Let's have a comment then explaining what this file does and why it is empty :)

eirikurn (commented):
Approving with comment (-automerge)

eirikurn (approved):

PR #16385, opened 2024-10-14
feat(native-app): add more info on passkey screen

thoreyjona:
## What

Add link to island.is with more info about passkeys and a small clause at the bottom that this can always be changed in settings.
Also adjusted the screen more for smaller screen sizes. If height of the screen is less than 800px then I make the image smaller and have less margin between elements to be sure that everything is in view. If this is an extra extra small screen (height < 650 px) then we skip the image.

## Screenshots / Gifs
On iPhone 16 (393x852) and iPhone SE (375x667):
<div>
</div>

## Checklist:

### DIFF
--- a/apps/native/app/src/screens/passkey/passkey.tsx
+++ b/apps/native/app/src/screens/passkey/passkey.tsx
@@ -1,4 +1,4 @@
-import { Button, Typography, NavigationBarSheet } from '@ui'
+import { Button, Typography, NavigationBarSheet, LinkText } from '@ui'
 import React, { useEffect, useState } from 'react'
 import { useIntl, FormattedMessage } from 'react-intl'
 import {
@@ -7,6 +7,8 @@ import {
   SafeAreaView,
   ActivityIndicator,
   Alert,
+  TouchableOpacity,
+  useWindowDimensions,
 } from 'react-native'
 import styled, { useTheme } from 'styled-components/native'
 import {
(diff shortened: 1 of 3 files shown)


apps/native/app/src/screens/passkey/passkey.tsx
```diff
@@ -79,8 +96,9 @@ export const PasskeyScreen: NavigationFunctionComponent<{
   const intl = useIntl()
   const theme = useTheme()
   const { openBrowser } = useBrowser()
+  const { height } = useWindowDimensions()
+  const isSmallDevice = height < 800
```
thoreyjona, line 100: If I need to use this in some other place I might consider moving this to a util and maybe implement a few breakpoints but will keep as is for now since this is the only place it is needed for now.

snaerseljan, line 100: Just speculating wondering if this is done anywhere else for small or large devices, e.g some hook or util fn

snaerseljan (approved):
Awesome. It will be nice to test out the Passkeys :)

apps/native/app/src/screens/passkey/passkey.tsx
thoreyjona: Nope, not that I could find at least. I decided when I need this next time I will add it to a util/hook 😄

PR #16446, opened 2024-10-17
feat(native-app): fix sharing for android

thoreyjona:
## What

Sharing files from the inbox was broken for android. Updated the package and it is fixed now. 
Tested both on android and ios.

## Checklist:

## Summary by CodeRabbit

- **New Features**
	- Enhanced file sharing functionality with improved error handling.
- **Bug Fixes**
	- Updated dependency for `react-native-share` to ensure better performance and compatibility.


apps/native/app/src/screens/document-detail/utils/share-file.tsx
```diff
@@ -18,11 +18,15 @@ export const shareFile = ({ document, pdfUrl }: ShareFileProps) => {
...
+  try {
+    await Share.open({
+      title: document.subject,
+      subject: document.subject,
+      message: `${document.sender.name} \n ${document.subject}`,
+      type: pdfUrl ? 'application/pdf' : undefined,
+      url: pdfUrl ? `file://${pdfUrl}` : document.downloadUrl!,
+    })
+  } catch (error) {
+    // noop
```
thoreyjona, line 30: No sure what we want to do here? If the user decides to not share the document, i.e. clicks share and then x on the share modal, then it is an error from the package `[Error: User did not share]`. Should we swallow that error but display others? Not sure 😅 We don't have any good workflow for errors in the app yet 🫠

eirikurn (approved):

PR #16472, opened 2024-10-18
feat: Param decorator to parse our app user agent strings

eirikurn:
## Summary by CodeRabbit

- **New Features**
	- Introduced a new decorator for parsing user agent strings in HTTP requests, enhancing request handling in the application.
	- Added functionality to automatically inject parsed user agent information into route handlers.

- **Tests**
	- Implemented a suite of unit tests to validate the behavior of the new user agent parsing functionality.


libs/nest/core/src/lib/decorators/ParsedUserAgent.decorator.ts
```diff
@@ -0,0 +1,92 @@
...
+
+  if (oldAndroidMatch) {
+    userAgent.app.name = 'IslandIsApp'
+    userAgent.os.name = 'Android'
+    return userAgent
+  }
+
+  // Default return, if no patterns matched
+  return userAgent
+}
```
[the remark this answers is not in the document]
eirikurn: Skipping adding a new dependency and currently focused on parsing app user agents which ua-parser-js does not support.

libs/nest/core/src/lib/decorators/ParsedUserAgent.decorator.ts
```diff
@@ -0,0 +1,92 @@
...
+  }
+}
+
+/**
+ * Parses the user agent string and returns an object with the user agent, os and app information.
+ *
+ * Only supports parsing IslandIsApp user agents to start with but the interface is
+ * future compatible if we want to parse more user agents with ua-parser-js.
+ */
+export const parseUserAgent = (userAgentString: string): UserAgent => {
```
baering, line 34: A micro-optimization could be to move the regex definitions outside the function to avoid having to recompile it each time, since this can get called quite often?

baering (approved):
LGTM

libs/nest/core/src/lib/decorators/ParsedUserAgent.decorator.ts
eirikurn: Inline regex patterns are compiled once when loading the script according to spec. The only thing which could be saved is object allocation and GC, which is negligent imo.

PR #16498, opened 2024-10-21
chore: No plain `status` logs

AndesKrrrrrrrrrrr:


AndesKrrrrrrrrrrr:
Accomplished with DataDog log filters instead

PR #16515, opened 2024-10-22
fix(ui-core): Render Button string children in span 

Toti91:
# Render Button string children in span 

## What

Render direct string children of `<Button>` in a `span`.

I've tested this and it does not affect/change any usage of button that I could find

## Why

This is a workaround for [a known bug](https://github.com/facebook/react/issues/11538) in which React throws an error when re-rendering if the DOM has changed by some 3rd party (for example Chrome dynamic translation function, or some extensions):

```bash
React Router caught the following error during render NotFoundError: 
Failed to execute 'removeChild' on 'Node'
```

Sparked by [this issue](https://islandis.slack.com/archives/C01R33Z2D3N/p1729262680925209), in which users that were most likely translating the page into f.x polish, could not continue to the next screen in the application because of this.

I narrowed the bug down to the Button component when in `loading` state:

The loading state renders children in a `<span>`, but normally the children are just rendered directly. I'm guessing this difference plus the 3rd party manipulation of the DOM causes the error, at least this change fixes it.

## Screenshots / Gifs

## Checklist:

## Summary by CodeRabbit

- **New Features**
	- Improved rendering of string children in the Button component for better semantic structure and accessibility.
  
- **Bug Fixes**
	- Enhanced styling capabilities by encapsulating string children in a `<span>` element when the button variant is not set to 'text'.


RunarVestmann (approved):
Let's get this on dev and notify people of this change so we have more eyes on potential regressions

Toti91:
> Let's get this on dev and notify people of this change so we have more eyes on potential regressions

Sounds good

PR #16542, opened 2024-10-23
feat(envrc): Timing & modularity

AndesKrrrrrrrrrrr:
- Add timing to notify of slow config
- Fallback for weird shells

Re-implementation of https://github.com/island-is/island.is/pull/16458


AndesKrrrrrrrrrrr:
MacOS only has second-level granularity. This adds unneeded extra complexity

PR #16553, opened 2024-10-24
fix(infra-next-server): Next.js crashes if getting requests before it's ready

AndesKrrrrrrrrrrr:
Some applications, e.g. `web` fail on startup if getting requests too early. This fixes it, but logs errors instead of crashing.

To reproduce the error:
```bash
# In one window
watch -n 0.1 'curl -s http://localhost:4200'

# In another window
yarn nx run web:serve
```

Returning a json in `/liveness` is nicer than _only_ getting a `200` empty response when checking.

Initial idea for the root cause came from https://github.com/vercel/next.js/issues/49536.

## Summary by CodeRabbit

- **New Features**
	- Enhanced server middleware to support asynchronous request processing for improved error handling.
  
- **Bug Fixes**
	- Improved error management to prevent application crashes and provide meaningful error responses to clients.

- **Documentation**
	- Added comments to clarify the purpose of cache control headers for web fonts.


libs/infra-next-server/src/lib/health.ts
```diff
@@ -64,6 +64,6 @@ export const setupHealthchecks = (
 
   app.use('/liveness', (_, res) => {
     res.statusCode = 200
-    res.end('')
+    res.json({ ok: true })
```
AndesKrrrrrrrrrrr, line 67: Should make minimal changes
```suggestion
    res.end('')
```

libs/infra-next-server/src/lib/bootstrap.ts
```diff
@@ -85,13 +85,22 @@ export const bootstrap = async (options: BootstrapOptions) => {
     options.externalEndpointDependencies,
   )
 
-  expressApp.use((req, res) => {
-    // Configure long caching for web fonts (often in public folder).
+  expressApp.use(async (req, res) => {
     if (req.url.match('.woff2?$')) {
+      // Configure long caching for web fonts (often in public folder).
```
AndesKrrrrrrrrrrr, line 90:
```suggestion
  expressApp.use(async (req, res) => {
    // Configure long caching for web fonts (often in public folder).
    if (req.url.match('.woff2?$')) {
```

baering (approved):

PR #16600, opened 2024-10-28
fix(financial-aid): use bankNumber and ledger

jonnigs:
## What

Use bankNumber and ledger when sending bank info

## Why

bank info was returning as accountNumber-acountNumber-acountNumber but should be bankNumber-ledger-accountNumer

## Checklist:


MargretFinnboga (approved):
Nice!

PR #16627, opened 2024-10-29
feat(auth-admin): only update/delete delegation index of the types that are being indexed.

magnearun:
## What

When we save to index we send an array of delegations, some times of all types, sometimes of just one type, when the latter happens we are deleting all pre existing rows that are not of that type, since the function checks if the row is in the incoming array, if it was not in the index it is created, if it was in the index but also in the incoming array it is updated, if it is in the index but not in the incoming it is deleted. So when indexing only custom delegations we would be deleting all rows of other types since they will not be in the oncoming array. 

## Why

Have correct data in the index table

## Screenshots / Gifs

## Checklist:

## Summary by CodeRabbit

- **New Features**
	- Introduced a new test case for indexing custom delegations with personal representatives.
	- Added a test suite to verify indexing of custom delegations without affecting existing ones.
  
- **Bug Fixes**
	- Enhanced filtering of delegation types during indexing for improved efficiency.


GunnlaugurG (approved):
LGTM 🚀

PR #16741, opened 2024-11-06
fix(shared): Markdown links always open in a new tab

HjorturJ:
# All markdown links should open in a new tab by default

https://app.asana.com/0/1207402031871459/1208255253643787/f

## What

Making all markdown links open in a new tab by default

## Why

Because it makes the most sense and was requested. Still possible to overwrite and have links open in the same tab

## Checklist:

## Summary by CodeRabbit

- **New Features**
  - Enhanced the Markdown component to support optional rendering options, including the ability to open links in a new tab.
  
- **Bug Fixes**
  - Improved the flexibility of Markdown rendering with updated options handling.

- **Documentation**
  - Updated interface and method signatures to reflect new options for the Markdown component.


libs/shared/components/src/Markdown/Markdown.tsx
```diff
@@ -1,11 +1,24 @@
 import { compiler } from 'markdown-to-jsx'
-import { markdownOptions } from './markdownOptions'
+import { getMarkdownOptions } from './markdownOptions'
 
 interface Props {
   children: string
+  conditionals?: OptionalConditionalOverrides
```
Toti91, line 6: maybe rename to "options", conditionals doesn't really describe what we are doing here I feel like

libs/shared/components/src/Markdown/Markdown.tsx
```diff
@@ -1,11 +1,24 @@
...
 }
 
-export const Markdown = ({ children }: Props) =>
-  compiler(children, markdownOptions)
+export interface OptionalConditionalOverrides {
+  openLinksInNewTab?: boolean
+  hehe?: number
```
Toti91, line 11: hehe

Toti91 (changes requested):
Now I don't know if this Markdown component is used outside of the application-system. But maybe we should have it default to false for opening links in new tab, to keep the current behavior, and then just update our usage  of this component in the application-system to override the default

libs/shared/components/src/Markdown/Markdown.tsx
HjorturJ: image

libs/shared/components/src/Markdown/Markdown.tsx
HjorturJ: fair, fixed

Toti91 (approved):
LGTM

magnearun (approved):

PR #16744, opened 2024-11-06
fix(native-app): Multiple license fixes

thoreyjona:
## What

Multiple license fixes:
* Make sure add to wallet buttons are always visible at the bottom of the screen when barcodes are visible - tested in ios simulator and android device.
* Fix bugs in fire arm licenses: make sure we have margin between fields and remove empty header for listing firearms. Máni at Hugsmiðjan will fix a bug I found from the backend as well (I was getting each firearm twice, but luckily only supporting one of the types in the app so only the header was shown twice.)
* Adding offline placeholder for barcodes
* Add license errors to translate files

Additional fixes:
* Don't show bottom tab bar on vehicle details screen
* Make sure that svgs and images in html documents are not bigger than 100% in width. 

## Screenshots / Gifs
Removed most of the information since these are production screenshots.

**Image fix for html documents:**
Before vs after:
<div>
</div>

**Fixing of margin of firearm fields:**
Before and after:
<div>
</div>

**Offline placeholder for barcodes:**
Screenshot 2024-11-06 at 10 03 16

## Checklist:

## Release Notes

- **New Features**
	- Enhanced error messages for wallet functionality, providing clearer user feedback when adding passes fails.
	- Offline messages added to the LicenseCard component to inform users about barcode accessibility based on connectivity.
  
- **Bug Fixes**
	- Improved error handling in the WalletPassScreen for better localization support.

- **Improvements**
	- Document viewer updated for better HTML and image rendering.
	- Vehicle detail screen navigation options adjusted for improved layout.
	- Conditional rendering logic in FieldCard refined for clarity.

- **Styling**
	- Adjustments made to various components for improved layout and responsiveness.


apps/native/app/src/screens/wallet-pass/wallet-pass.tsx
```diff
@@ -167,6 +172,21 @@ export const WalletPassScreen: NavigationFunctionComponent<{
...
+    (!isConnected && isBarcodeEnabled)
+      ? barcodeHeight + LICENSE_CARD_ROW_GAP
+      : 0
+
+  const [key, setKey] = useState(0)
+  useEffect(() => {
+    // Used to rerender ScrollView to have correct ContentInset based on barcode/no barcode
+    // Remove once barcodes are live
+    setKey((prev) => prev + 1)
+  }, [isBarcodeEnabled])
```
thoreyjona, line 188: The feature flag is super fast to go from false to true initially so this should be super fast.

apps/native/app/src/ui/lib/field/field-card.tsx
```diff
@@ -151,7 +151,9 @@ export function FieldCard(props: FieldCardProps) {
 
   return (
     <Host>
-      {props.title !== null || (props.title !== undefined && props.code) ? (
+      {(props.title !== null &&
+        !(props.title === undefined && props.code === undefined)) ||
```
thoreyjona, line 155: I added this addition: `!(props.title === undefined && props.code === undefined)`. The reason for it was to skip rendering an empty header like this for firearms where both title and code are `undefined`: 
Screenshot 2024-11-06 at 13 25 03.

Since the licenses are all **VERY** dynamic and I don't have access to them all on dev I did not want to change the logic that was before hand and e.g. change it to just be `!props.title ||  (props.title !== undefined && props.code)` since it theoretically could break something I don't now about ☺️

snaerseljan, line 155: nit: You could do

```
(!!props.title || !!props.code)
```

instead of

```
!(props.title === undefined && props.code === undefined))
```

snaerseljan (approved):
Awesome. Nice job. I know this must have been frustrating  🙃

snaerseljan:
@thoreyjona maybe add max width on text so the second line won't only have one word
image

PR #16762, opened 2024-11-07
feat(ojoi-web): Categories page

jonbjarnio:
## What

- Updated the public official journal api document to the latest version
- Updated the landing page to display up to 20 main categories
- Updated the categories page to server side rendering
- Optimised filtering, removed use state soup and unnecessary client side calls to the api
- Fixed `CategoryCard` component so that if you pass `tags` property with `href` then the `tags` links wont be nested inside the `CategoryCard` link causing a hydration error / invalid html

## Why

So that users can easily search for cases with certain categories.

## Checklist:

## Release Notes

- **New Features**
  - Introduced utility functions for improved handling of query strings.
  - Enhanced search and filtering functionalities for categories.
  - Streamlined advert management in the home component.
  - Added error messages for category fetching failures.
  - Updated API to support nested category structures.

- **Bug Fixes**
  - Improved data fetching and state management for categories and adverts.

- **Documentation**
  - Updated API specifications to reflect new query parameters and schema changes.

These updates enhance user experience by providing better search capabilities and more informative error handling.


apps/web/screens/OfficialJournalOfIceland/OJOICategories.tsx
```diff
@@ -174,41 +90,162 @@ const OJOICategoriesPage: CustomScreen<OJOICategoriesProps> = ({
...
+        (c) => c.title.charAt(0) === letter,
+      )
+      acc.push({
+        letter,
+        categories: sortCategories(
+          letterCategories.map((c) => ({ label: c.title, value: c.slug })),
+        ),
+      })
+      return acc
+    }, [] as MalaflokkarType)
```
disaerna, line 148: Is it safe to use "as" here?

apps/web/screens/OfficialJournalOfIceland/OJOICategories.tsx
```diff
@@ -174,41 +90,162 @@ const OJOICategoriesPage: CustomScreen<OJOICategoriesProps> = ({
...
+  ])
+
+  const onDepartmentChange = (value?: string) => {
+    setQueryDepartment(value)
+    router.replace(
+      {
+        pathname: categoriesUrl,
+        query: removeEmptyFromObject({
+          ...router.query,
+          deild: value,
```
disaerna, line 174: Icelandic mixed with english 🙈

apps/web/screens/OfficialJournalOfIceland/OJOICategories.tsx
```diff
@@ -294,59 +333,62 @@ const OJOICategoriesPage: CustomScreen<OJOICategoriesProps> = ({
...
+              (cat) => cat.letter === letter,
+            )
+            return (
+              <Tag
+                key={letter}
+                active={isActive}
+                onClick={() => {
+                  toggleLetter(letter)
+                }}
+                // disabled={isDisabled}
```
disaerna, line 348: remove?

apps/web/screens/OfficialJournalOfIceland/OJOICategories.tsx
disaerna: Oh I see, it is part of the query params 🤭

disaerna (commented):
Code looks good 👍🏼 I think I need more detailed explanation about the Category Card before accepting though

disaerna (approved):
👍🏼

libs/island-ui/core/src/lib/CategoryCard/CategoryCard.tsx
```diff
@@ -13,6 +13,7 @@ import { Hyphen } from '../Hyphen/Hyphen'
 import { Text, TextProps } from '../Text/Text'
 
 import * as styles from './CategoryCard.css'
+import { Link } from '../Link/Link'
```
RunarVestmann, line 16: Move this import above the styles import?

libs/island-ui/core/src/lib/CategoryCard/CategoryCard.tsx
```diff
@@ -36,6 +37,7 @@ export type CategoryCardProps = {
   tags?: Tag[]
   tagOptions?: Pick<TagProps, 'hyphenate' | 'truncate' | 'textLeft'>
   href?: string
+  nestedHref?: boolean
```
RunarVestmann, line 40: What is a nested href? Is that something we've got in figma as well?

RunarVestmann: Couldn't we rather detect if there is a nested href instead of explicitly telling the component?

RunarVestmann: Also your solution only makes the title clickable instead of the traditional "entire card can be clicked"

apps/web/components/OfficialJournalOfIceland/OJOIUtils.ts
```diff
@@ -91,3 +91,31 @@ export const formatDate = (date?: string, df = 'dd.MM.yyyy') => {
     throw new Error(`Could not format date: ${date}`)
   }
 }
+
+export const getStringArrayFromQueryString = (
```
RunarVestmann, line 95: You could also look into using next-usequerystate to parseServerside query params

parseAsArrayOf might be something to check out

libs/island-ui/core/src/lib/CategoryCard/CategoryCard.tsx
jonbjarnio: We could always check if there is a tag passed to the component that has `href` prop. But if we want to have the whole box clickable we need to use router or some other api to allow routing instead of the anchor tag. But how the component is programmed right now allows for nested anchor tags.

apps/web/components/OfficialJournalOfIceland/OJOIUtils.ts
jonbjarnio: Yes, after looking at the docs this is exactly what I could use.

libs/island-ui/core/src/lib/CategoryCard/CategoryCard.tsx
RunarVestmann: Just read some things, you are right that <a> inside an <a> is invalid html. 

Perhaps the tags themselves should have an onClick or the card itself. However this issue is in more places, for example the "Lífsviðburðir" section on https://island.is/ also has nested <a> tags.

rafnarnason (approved):

RunarVestmann (approved):

PR #16776, opened 2024-11-08
chore(sqs): Better logs

AndesKrrrrrrrrrrr:
Currently logging raw args as list of key/value objects.

## Summary by CodeRabbit

- **New Features**
	- Enhanced logging for notifications, providing clearer context with structured argument representation and additional queue details.
	- Added a new property `queueName` to improve queue management in the service.

- **Bug Fixes**
	- Improved error handling and assertions for queue operations.

- **Documentation**
	- No changes made to existing documentation annotations.


apps/services/user-notification/src/app/modules/notifications/notifications.controller.ts
```diff
@@ -105,6 +105,8 @@ export class NotificationsController {
       messageId: id,
       ...flattenedArgs,
       ...body,
+      args: {}, // Remove args, since they're in a better format in `flattenedArgs`
+      queue: { url: this.queue.url, name: this.queue.queueName },
```
rafnarnason, line 109: not sure i understand the value of this, wont this simply log the same queue url and name per environment for millions of messages ?

AndesKrrrrrrrrrrr: The intent is to log the queue _name_, which is semi-related to the URL. Also, the URL is only available post-startup, while the name should be available on class instantiation.

This make it easier to filter by the human-readable queue name, which should be the same on any environment. The queue URL (I think), changes per environment, acting as an ID.

AndesKrrrrrrrrrrr: But yes, this does increase the log volume, but comparatively, it's minor. I'd prefer better searching/filtering over a few more cents going to the Clouds.

PR #16801, opened 2024-11-11
feat(j-s): Defender Updated Notification

gudjong:
# Defender Updated Notification

[Senda tilkynningar til dómstóls þegar verjandi er valinn á island.is](https://app.asana.com/0/1199153462262248/1208619201899978/f)

## What

- Sends notification to court when defendant chooses a defender from island.is.
- Stops updating defendants, defenders and prosecutors in court via the court API. The robot now handles defendants and prosecutors and soon defenders as well.
- Moves the initiation of defender selected notifications to the defendant service.
- Moves the sending of defender selected notifications to the defendant notification service.

## Why

- The court needs to know.

## Checklist:

## Release Notes

- **New Features**
	- Introduced a new guard for validating the existence of a defendant's national ID.

- **Improvements**
	- Enhanced error handling and logging in various services and controllers.
	- Simplified email notification processes to accommodate multiple recipients.
	- Refined message handling and update logic for subpoenas and defendants.
	- Updated the notification system to better categorize notifications related to defendants and subpoenas.

- **Bug Fixes**
	- Resolved inconsistencies in case update handling and message queue interactions.

- **Refactor**
	- Consolidated case management logic and improved the structure of notification types.
	- Updated method signatures and streamlined logic within several services.
	- Removed unnecessary imports and simplified control flow in various modules.


apps/judicial-system/backend/src/app/modules/notification/defendantNotification.service.ts
```diff
@@ -152,13 +198,13 @@ export class DefendantNotificationService extends BaseNotificationService {
 
   async sendDefendantNotification(
     type: DefendantNotificationType,
-    defendant: Defendant,
     theCase: Case,
+    defendant: Defendant,
```
unakb, line 202: Just curious, why do you prefer to have the case as the first parameter instead of the defendant? I was thinking about which made more sense when I wrote the service and I felt like since we were in the defendant notification service then it makes sense to focus on the defendant, but maybe it's better to just always have theCase first everywhere to keep things more cohesive

apps/judicial-system/backend/src/app/modules/notification/defendantNotification.service.ts
```diff
@@ -50,46 +51,71 @@ export class DefendantNotificationService extends BaseNotificationService {
...
-        this.sendEmail(
-          formattedSubject,
-          formattedBody,
-          defendant.defenderName,
-          defendant.defenderEmail,
-          undefined,
-          true,
-        ),
-      )
+    for (const recipient of to) {
```
unakb, line 62: nice 👌

apps/judicial-system/backend/src/app/modules/notification/defendantNotification.service.ts
gudjong: Good question. I guess I always think about this as a hierarchy, case includes defendant, includes subpoena, etc. It really does not matter.

PR #16808, opened 2024-11-12
chore(j-s): Logging more info for system status

unakb:
[Logga betur successful köll í xrd-api þegar birting uppfærist.](https://app.asana.com/0/1199153462262248/1208626990151899)

## What

Added logs to get a better feel for what´s going on in the system
## Checklist:

## Release Notes

- **New Features**
	- Enhanced logging capabilities for case events in the EventService, including specific event tracking.
	- Added logging for successful subpoena deliveries in the SubpoenaService.

- **Bug Fixes**
	- Improved error handling logging in the SubpoenaService during delivery processes.

- **Refactor**
	- Simplified logging processes in the CaseService and LimitedAccessCaseService by removing unnecessary logging statements.
	- Updated logging level from debug to info in the AppController for subpoena updates.
	- Clarified API documentation for indictment case retrieval in the InternalCaseController to specify digital mailbox context.
	- Refined filtering logic in the InternalCaseService to ensure only relevant indictment cases are retrieved.


apps/judicial-system/backend/src/app/modules/case/internalCase.service.ts
```diff
@@ -1223,6 +1223,12 @@ export class InternalCaseService {
       attributes: ['id', 'courtCaseNumber', 'type', 'state'],
       where: {
         type: CaseType.INDICTMENT,
+        // Make sure we don't send cases that are in deleted or other inaccessible states
```
oddsson, line 1226: Do we want to be extra safe and do this in `getIndictmentCase` as well?

unakb: I did think about that, but I'm not sure what would happen in the cases where an indictment has been sent to people's digital mailboxes, if they had a link there already to an existing case that was later deleted - whether they should still be able to link from that to the case page ? Since we can't delete things from the inbox

oddsson (approved):

PR #16860, opened 2024-11-14
feat(native-app): add third party error template and use in health overview

thoreyjona:
## What

Modify Problem component in app to support Third Party Error, so we show the name of the organization that should provide the information that is erroring.

Using that new error component in health-overview section
## Why

Error states for inputs are currently loading state which is not good - this is better. Will rewrite errors for inputs soon.

## Screenshots / Gifs

Simulator Screenshot - iPhone 16 - 2024-11-14 at 10 23 35

## Checklist:

### DIFF
--- a/apps/native/app/src/messages/en.ts
+++ b/apps/native/app/src/messages/en.ts
@@ -594,6 +594,9 @@ export const en: TranslatedMessages = {
   'problem.offline.title': 'No internet connection',
   'problem.offline.message':
     'An error occurred while communicating with the service provider',
+  'problem.thirdParty.title': 'No connection',
+  'problem.thirdParty.message':
+    'An error occurred while communicating with the service provider',
 
   // passkeys
   'passkeys.headingTitle': 'Sign in with Island.is app',
--- a/apps/native/app/src/stores/organizations-store.ts
+++ b/apps/native/app/src/stores/organizations-store.ts
@@ -30,6 +30,7 @@ interface Organization {
 interface OrganizationsStore extends State {
   organizations: Organization[]
   getOrganizationLogoUrl(forName: string, size?: number): ImageSourcePropType
+  getOrganizationNameBySlug(slug: string): string
   actions: any
 }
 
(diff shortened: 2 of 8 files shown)


apps/native/app/src/messages/en.ts
```diff
@@ -594,6 +594,9 @@ export const en: TranslatedMessages = {
   'problem.offline.title': 'No internet connection',
   'problem.offline.message':
     'An error occurred while communicating with the service provider',
+  'problem.thirdParty.title': 'No connection',
```
snaerseljan, line 597: I feel like no connection means no internet. Isn't there a better sentence?

Failed to connect to service
Unable to reach service
Service unreachable

🤷‍♂️

apps/native/app/src/stores/organizations-store.ts
```diff
@@ -72,6 +73,10 @@ export const organizationsStore = create<OrganizationsStore>(
         const uri = `${url}?w=${size}&h=${size}&fit=pad&fm=png`
         return { uri }
       },
+      getOrganizationNameBySlug(slug: string) {
+        const org = get().organizations.find((o) => o.slug === slug)
+        return org?.title ?? ''
```
snaerseljan, line 78: Some services have not set the organisations property when service errors. We should default to some message instead of empty string

apps/native/app/src/utils/get-organization-slug-from-error.ts
```diff
@@ -0,0 +1,36 @@
...
+  problem?: PartialProblem
+  exception?: {
+    problem?: PartialProblem
+  }
+}
+
+/**
+ * Extracts the organization slug from the Apollo error, if it exists.
+ */
+export function getOrganizationSlugFromError(error: ApolloError | unknown) {
```
snaerseljan, line 18: Did you try to import this from here, since this is the same?

`libs/react-spa/shared/src/utils/getOrganizationSlugFromError.ts`

snaerseljan (approved):
Nice work.

This organisation feature is a nice addon 👏

apps/native/app/src/messages/en.ts
thoreyjona: Yeah I agree, not a big fan of this, just used what the service portal is using 🤷‍♀️ Will discuss on sync today

apps/native/app/src/stores/organizations-store.ts
thoreyjona: If I don't get a name I default to showing an icon and no title 😊 I think that is what we want to do
Screenshot 2024-11-15 at 10 51 10

apps/native/app/src/utils/get-organization-slug-from-error.ts
thoreyjona: There is nothing in the app imported from these folders so I don't think it will work. I will try though

PR #16908, opened 2024-11-15
feat(native-app): support type: 'Table' in field render

thoreyjona:
## What

Support type 'Table' in Field render. 
## Why
The service portal is using this for firearm licenses. To keep everything in sync the app supports that as well now. Máni from Hugsmiðjan is creating a rule on the backend that if the user-agent is version that is older than 1.4.8 then he will send the type: 'Group' for firearm licenses to be backwards compatible, otherwise he just sends type: 'Table'

## Screenshots / Gifs

## Checklist:

### DIFF
--- a/apps/native/app/src/screens/wallet-pass/components/field-render.tsx
+++ b/apps/native/app/src/screens/wallet-pass/components/field-render.tsx
@@ -67,6 +67,31 @@ export const FieldRender = ({
                 </FieldGroup>
               )
 
+            case 'Table':
+              if (label) {
+                return (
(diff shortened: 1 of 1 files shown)


apps/native/app/src/screens/wallet-pass/components/field-render.tsx
```diff
@@ -67,6 +67,31 @@ export const FieldRender = ({
                 </FieldGroup>
               )
 
+            case 'Table':
+              if (label) {
+                return (
+                  <View key={key} style={{ marginTop: 24, paddingBottom: 4 }}>
```
snaerth, line 73: nit: Theme spacing

PR #16921, opened 2024-11-18
feat(island-ui): design tokens

albinagu:
## Checklist:

### DIFF
--- a/libs/island-ui/theme/src/lib/theme.ts
+++ b/libs/island-ui/theme/src/lib/theme.ts
@@ -3,6 +3,21 @@ import type { StyleRule } from '@vanilla-extract/css'
 import omit from 'lodash/omit'
 import * as color from './colors'
 
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const designTokens = require('./designTokens.json')
+console.log(designTokens)
+
+const colorTokensLight = designTokens[1]['01 Colors Tokens'].modes['Light Mode']
+
+const backgroundColors = colorTokensLight.background
+const borderColors = colorTokensLight.border
+const feedbackColors = colorTokensLight.feedback
+const foregroundColors = colorTokensLight.foreground
+const interactiveColors = colorTokensLight.interactive
+const overlayColors = colorTokensLight.overlay
+
+console.log(backgroundColors, borderColors, feedbackColors, foregroundColors, interactiveColors, overlayColors)
+
 export const UNIT = 8
 
 export const spacing = {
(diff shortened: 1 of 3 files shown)


libs/island-ui/theme/src/lib/colors/colors.ts
```diff
@@ -0,0 +1,202 @@
...
+export const dark200 = dark[200] ?? '#ccccd8'
+export const dark100 = dark[100] ?? '#f2f2f5'
+
+export const red600 = red[600] ?? '#b30038'
+export const red400 = red[400] ?? '#ff0050'
+export const red300 = red[300] ?? '#ff99b9'
+export const red200 = red[200] ?? '#ffccdc'
+export const red100 = red[100] ?? '#fff2f6'
+
+export const white = white400[400] ?? '#ffffff'
```
oddsson, line 28: There is only one white, can this be this instead, like `black`?

```suggestion
export const white = '#ffffff'
```

oddsson (approved):
Nice 💪

libs/island-ui/theme/src/lib/colors/colors.ts
albinagu: yea, totally 💯

libs/island-ui/theme/src/lib/theme.ts
```diff
@@ -98,10 +97,13 @@ export const theme = {
...
-      large: '8px',
-      xl: '24px',
-      circle: '50%',
+      standard: `${radius.xs.$value ?? 4}px`, // TODO: check with designers to keep as standard
+      large: `${radius.default.$value ?? 8}px`, // TODO: check with designers to keep as large
+      md: `${radius.md.$value ?? 12}px`,
+      lg: `${radius.lg.$value ?? 16}px`,
+      xl: '24px', // TODO: should be added to figma?
+      circle: '50%', // TODO: should be added to figma?
+      full: `${radius.full.$value ?? 9999}px`,
```
Toti91, line 106: `circle` and `full` achieve the same thing, should we omit either one?

Toti91 (approved):
Very nice, one small question

libs/island-ui/theme/src/lib/theme.ts
albinagu: good spot! for sure, any preference on which to omit? perhaps circle as it has not been introduced in figma

albinagu: will be addressed in the followup PR

Toti91: > perhaps circle as it has not been introduced in figma

Sounds good!

PR #16928, opened 2024-11-18
feat(payment charge): Enable the usage of quantity field when creating charges

johannaagma:
## What

Right now, we are always sending in an array of items with quantity=1 when creating charges in FJS. If we want to select the same item more than once, we add the same item multiple times in the array. 

E.g. `charges = ["SA110", "SA110"]`

FJS API is already supporting the quantity field, so we can use that.

But need to update the PaymentStateBuilder (and more) to accept an array of objects (with code+quantity fields), instead of just strings (code only)

## Why

For most applications it works fine just repeating the same item x times. But we did get in trouble with Veðbókarvottorð which allows users to buy up to 10 certificates at once -> that bug was fixed by increasing the size for the description field in our Payment model. But soon, we will be adding the Námskeið application,  which allows you to pay for 100s of people to register to course.

## Screenshots / Gifs

Sample of how it will look like in PaymentOverview
image
(Noted: faked the quantity for this screenshot, but no application would currently display multiples of the same item. But updated so Mortgage Certificate sends the FJS items with quantity, but it is displayed on our side separately - to show property number per)

## Checklist:

## Summary by CodeRabbit

- **New Features**
	- Introduced `BasicChargeItem` interface for enhanced charge item structure.
	- Updated charge item handling across various modules to include quantities.
	- Enhanced `getChargeItems` function to return structured charge item objects.

- **Bug Fixes**
	- Improved error handling and validation in payment-related functions.

- **Documentation**
	- Updated documentation to reflect changes in charge item data structure and handling.

- **Refactor**
	- Renamed and modified several utility functions for consistency in charge item retrieval.
	- Streamlined payment state configurations to utilize new charge item structures.

- **Tests**
	- Updated test cases to reflect changes in charge item handling and structure.


sigruntg (approved):

jonarnarbriem (changes requested):

jonarnarbriem (approved):

jonarnarbriem:
I still think variable names in the payment service in general are not great but that's outside of the scope of this PR and I think after the name change you made we're at least not making things worse.

apps/application-system/api/src/app/modules/application/e2e/payment/payment.service.spec.ts
```diff
@@ -177,9 +199,31 @@ describe('Payment Service', () => {
     expect(result.fulfilled).toBe(true)
   })
 
+  it('should find charge items using quantity', async () => {
```
norda-gunni, line 202: Consider adding a check for whether each of the charge items has the correct quantity in addition to checking the sum of charge items.

norda-gunni (commented):
Just one thing

norda-gunni (approved):
LGTM

kksteini (approved):

jonarnarbriem (approved):
approving a second time to see if it pleases the github gods

sigruntg (approved):

PR #17078, opened 2024-11-28
chore(core): Make Docker & CI affect everything

AndesKrrrrrrrrrrr:
Recent `main` outages have been caused by regressions in CI and Dockerfiles. These core components currently don't "affect" the applications, hence, the applications aren't built, tested, docker-built, etc. when they should be.

Hoping here to catch future regressions.

## Summary by CodeRabbit

- **New Features**
	- Enhanced build and testing configurations with the addition of new inputs for "DevOps," "CI," and "Dockerfiles."
	- Improved integration with GitHub actions and workflows.

- **Bug Fixes**
	- Updated input paths to ensure proper targeting of relevant files and directories.


busla (changes requested):
scripts/* should also be included

nx.json
```diff
@@ -152,6 +152,17 @@
         "runtime": "node --version"
       }
     ],
+    "DevOps": ["CI", "Dockerfiles"],
+    "CI": [
+      "{workspaceRoot}/**/docker-compose*.yml",
```
eirikurn, line 157: Which docker files are these? Can we make this more focused? 

We have plenty of docker-compose files in some apps which are only (?) used for development. These are already matched by `production/default` with `"{projectRoot}/**/*"`. We don't want application system docker-compose to affect all applications.

Maybe you want to mainly target stuff in the infra or script folders?

eirikurn (approved):
Approving with comment.

nx.json
AndesKrrrrrrrrrrr: Removed docker-compose, makes sense not to include development docker-compose.

I could list all the Dockerfiles I know of, and that we have currently, but no app should use Dockerfiles, and any Dockerfile created should be part of our bulid/CI. I'd like to opt for a simpler wider include rather than specifying the Dockerfiles we need; since currently, it's _all_ the Dockerfiles.

busla (approved):
LGTM

PR #17105, opened 2024-12-02
fix: add missing ns grant to api for bff

brynjarorng:
# Fix bff api
API needs a grant to allow bff communications for the service portal


svanaeinars (approved):
LGTM

PR #17118, opened 2024-12-03
fix(service-portal): Fixes for third-party paths to old login

snaerth:
# Bff thirdparty login path fixes and Bff provider updates

## What

- Updates third party paths to user `bff/login` instead.
- Updates BffProvider to handle following:
   - Support to handle authenticated and non-authenticated old login paths, aka `/minarsidur/login`.
   - Support to pull existing relevant login params and use them in the sign in process.

## Why

After the new BFF pattern was merged we want to make sure that old login path works properly.

## Checklist:

## Summary by CodeRabbit

- **New Features**
	- Enhanced passkey handling with improved URL checks for registration and authentication.
	- Added new environment variables for user notification services to streamline login processes.
	- Updated login URL structure for a more streamlined user experience.
	- Introduced new health check paths and HPA configurations for user-notification services.

- **Bug Fixes**
	- Improved login path redirection logic to ensure users are redirected correctly based on their current URL.

- **Chores**
	- Updated command execution methods in project configurations for better dependency management.


apps/web/components/Header/LoginButton.tsx
```diff
@@ -13,7 +13,7 @@ import { useI18n } from '@island.is/web/i18n'
 import { LayoutProps } from '@island.is/web/layouts/main'
 
 const minarsidurLink = '/minarsidur/'
-const minarsidurDelegationsLink = '/minarsidur/login?prompt=select_account'
+const minarsidurDelegationsLink = '/bff/login?prompt=select_account'
```
snaerth, line 16: this will redirect to /minarsidur on successful login

disaerna (approved):
Code owned file 👍🏼

libs/react-spa/bff/src/lib/BffProvider.tsx
```diff
@@ -150,13 +196,13 @@ export const BffProvider = ({
...
       window.location.href = bffUrlGenerator('/login', {
-        target_link_uri: window.location.href,
+        target_link_uri: loginQueryParams['target_link_uri'],
         ...(nationalId
           ? { login_hint: nationalId }
           : { prompt: 'select_account' }),
       })
     },
-    [bffUrlGenerator],
+    [bffUrlGenerator, loginQueryParams],
```
eirikurn, line 205: This may not work as expected. Since `loginQueryParams` is memoized. I might have browsed around before calling switchUser, and end up on some old page.

Though I'm curious if this matches previous logic as I feel like at some point we always sent you to the homepage since you might not have access to the current page. Personally, I prefer this UX where you end up on the same page, even though sometimes you just see access denied.

libs/react-spa/bff/src/lib/BffProvider.tsx
```diff
@@ -1,5 +1,12 @@
 import { useEffectOnce } from '@island.is/react-spa/shared'
-import { ReactNode, useCallback, useEffect, useReducer, useState } from 'react'
+import {
```
eirikurn, line 2: Let's also update the delegation email notification URL. 👍

eirikurn (commented):
Sweet!

libs/react-spa/bff/src/lib/BffProvider.tsx
snaerth: Then I will need to disable  the test if you are referring to this 
https://github.com/island-is/island.is/blob/55aacf7935422425da140c2e254d3d25fdeb2660/apps/services/user-notification/src/app/modules/notifications/notificationsWorker/notificationsWorker.service.ts#L479

See test here
https://github.com/island-is/island.is/blob/55aacf7935422425da140c2e254d3d25fdeb2660/apps/services/user-notification/src/app/modules/notifications/notificationsWorker/notificationsWorker.service.spec.ts#L306

RunarVestmann (approved):

libs/react-spa/bff/src/lib/BffProvider.tsx
snaerth: Ahh good point. I will update

kksteini (approved):
Codeowner files lgtm

libs/react-spa/bff/src/lib/BffProvider.tsx
```diff
@@ -100,6 +134,13 @@ export const BffProvider = ({
         return
       }
 
+      // If user is logged in and on the old login path, then start the sign-in process
```
snaerth, line 137: I now this seams wrong, but the target link uri, prompt or login hint can have changed.

svanaeinars (approved):
LGTM

eirikurn (approved):

PR #17146, opened 2024-12-05
fix(ojoi): Updated controllers and bugfixes

jonbjarnio:
## What

- Updated some of the OJOI controllers to use the input components directly instead of the built in application system controllers. 
- Fixed bug where submit button was disabled when it was not supposed to
- Fixed bug where input fields / selects did not update when user selected an older advert
- Improved and simplified the advert modal screen 

## Checklist:


thorkellmani (approved):
LGTM

PR #17042, opened 2024-11-27
fix(island-ui): Removing radius "circle" to use "full" instead

albinagu:
Followup PR for design tokens

## Checklist:

- [ ] I have performed a self-review of my own code
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] Formatting passes locally with my changes
- [ ] I have rebased against main before asking for a review

PR #17155, opened 2024-12-06
fix(native-app): change border radius circle to be full

thoreyjona:
## What

Changes in #17042 broke the app since the app has its own theme (not using the same theme as other projects in the monorepo).
I changed the `borderRadius.circle` to be `borderRadius.full`. 

## Checklist:

## Summary by CodeRabbit

- **New Features**
	- Updated the border radius terminology for circular elements from `circle` to `full` in the theme settings. 

This change enhances clarity in design specifications while maintaining all other theme properties.


eirikurn (approved):
LGTM

PR #17196, opened 2024-12-11
chore(j-s): Send signed court record to police for investigation cases

unakb:
[Við erum að senda úrskurði og þingbækur of snemma í LÖKE](https://app.asana.com/0/1199153462262248/1208270698959838)

## What

Send signed court record to police for investigation cases

## Why

Because prosecutors sometimes need to use it when continuing the investigation

## Checklist:

## Summary by CodeRabbit

- **New Features**
	- Added functionality to deliver signed court records to the police.
	- Introduced messaging capabilities for signed court records in the case management system.
	- Added a new message type for delivering signed court records to police.

- **Bug Fixes**
	- Corrected error message for case creation.

- **Tests**
	- Added unit tests for delivering signed court records to the police.
	- Implemented tests for security guards associated with the delivery method.


apps/judicial-system/backend/src/app/modules/case/internalCase.service.ts
```diff
@@ -1099,6 +1106,32 @@ export class InternalCaseService {
...
 
+  async deliverSignedCourtRecordToPolice(
+    theCase: Case,
+    user: TUser,
+  ): Promise<DeliverResponse> {
+    const delivered = await this.getSignedCourtRecordPdf(theCase)
+      .then((pdf) =>
+        this.deliverCaseToPoliceWithFiles(theCase, user, [
+          {
+            type: PoliceDocumentType.RVTB,
```
unakb, line 1117: I haven't removed this type from the original case delivery to police, so investigation cases will end up with 2 different court records - one signed and one unsigned.. We feel like that should be alright, but if anyone knows of a reason we shouldn't do that then let me know. 
The reason is that in cases where there is a delay in signing it at least the prosecutors get the unsigned copy straight away..

PR #17198, opened 2024-12-11
feat(j-s): Add punishment type to indictment overview

thorhildurt:
## References
[Fmst - Stöður hjá fangelsismálastofnun, málalisti og yfirlit máls
](https://app.asana.com/0/1199153462262248/1208652035345719)

## What
- Add the option to select punishment type to indictment overview as a FMST user
    - Migration to add the new optional field `punishment_type` 
    - New limited access controller for defendant which to follow the same pattern as we have for `case`, `subpoena`
    - Added role rules to only allow a given role to update a specific field
    - Added new radio buttons in indictment overview to add a relevant punishment type

## Why
- Enables FMST secretaries to sort completed cases in the table according to the punishment type

## Screen
https://github.com/user-attachments/assets/6f4e8ac8-8eba-4b82-b730-aa65907a1732

## Checklist:

## Summary by CodeRabbit

- **New Features**
	- Introduced functionality for updating defendant information under limited access scenarios.
	- Added new GraphQL fields for punishment types in case queries.
	- Enhanced the Indictment Overview component with a selection interface for punishment types.
	- Added a new enum for punishment types to represent various legal punishments.
	- Implemented a new controller and resolver for handling limited access updates.
	- Introduced a new mutation for limited access updates to defendant records.
	- Added new message definitions for punishment types in localization support.

- **Bug Fixes**
	- Improved error handling for updates related to limited access defendants.

- **Documentation**
	- Expanded localization support for punishment types in the Indictment Overview.

- **Chores**
	- Added new hooks and GraphQL mutations for limited access updates.


apps/judicial-system/api/src/app/modules/defendant/limitedAccessDefendant.resolver.ts
```diff
@@ -0,0 +1,54 @@
...
+    { backendService }: { backendService: BackendService },
+  ): Promise<Defendant> {
+    const { caseId, defendantId, ...updateDefendant } = input
+    this.logger.debug(
+      `Updating limitedAccess defendant ${defendantId} for case ${caseId}`,
+    )
+
+    return this.auditTrailService.audit(
+      user.id,
+      AuditedAction.UPDATE_DEFENDANT,
```
thorhildurt, line 45: Using the same value for the general update endpoint

apps/judicial-system/backend/src/app/modules/defendant/limitedAccessDefendant.controller.ts
```diff
@@ -0,0 +1,66 @@
...
+@ApiTags('limited access defendant')
+@UseGuards(JwtAuthGuard, RolesGuard)
+export class LimitedAccessDefendantController {
+  constructor(
+    private readonly defendantService: DefendantService,
+    @Inject(LOGGER_PROVIDER) private readonly logger: Logger,
+  ) {}
+
+  @UseGuards(CaseExistsGuard, DefendantExistsGuard)
+  @RolesRules(prisonSystemStaffUpdateRule)
```
thorhildurt, line 41: This will ensure that prison staff can only pass in `punishmentType` and no other fields of the `UpdateDefendantDto`

apps/judicial-system/backend/src/app/modules/defendant/limitedAccessDefendant.controller.ts
```diff
@@ -0,0 +1,66 @@
...
+    type: Defendant,
+    description: 'Updates a defendant',
+  })
+  updateDefendant(
+    @Param('caseId') caseId: string,
+    @Param('defendantId') defendantId: string,
+    @CurrentHttpUser() user: User,
+    @CurrentCase() theCase: Case,
+    @CurrentDefendant() defendant: Defendant,
+    @Body() updateDto: LimitedAccessUpdateDefendant,
```
thorhildurt, line 53: Added this type for readability, but its not really restricting anything

apps/judicial-system/web/src/routes/Prison/IndictmentOverview/IndictmentOverview.tsx
```diff
@@ -15,19 +16,46 @@ import {
...
+    defendant &&
+      limitedAccessUpdateDefendant({
+        caseId: workingCase.id,
+        defendantId: defendant.id,
+        punishmentType: updatedPunishmentType,
+      })
+    setPunishmentType(updatedPunishmentType)
+  }
+
+  const hasSetPunishmentType = (punishmentType: PunishmentType) =>
```
thorhildurt, line 56: When a `punishmentType` value is already stored in the db, by default the radio button selection is empty

apps/judicial-system/web/src/utils/hooks/useDefendants/index.ts
```diff
@@ -78,6 +84,26 @@ const useDefendants = () => {
     [formatMessage, updateDefendantMutation],
   )
 
+  // TODO: This is maybe not ideal?
+  const limitedAccessUpdateDefendant = useCallback(
```
thorhildurt, line 88: This is a bit redundant of `updateDefendant` above. In cases we handle these the mutations calls within the same function by checking the user...but maybe this is ok for now?

thorhildurt:
There are no tests yet for the new controller, but feel free to reject if that is a must with this PR otherwise I plan to add in one of the follow-ups 👍

apps/judicial-system/backend/src/app/modules/defendant/guards/rolesRules.ts
```diff
@@ -0,0 +1,12 @@
...
+import { UpdateDefendantDto } from '../dto/updateDefendant.dto'
+
+const limitedAccessFields: (keyof UpdateDefendantDto)[] = ['punishmentType']
+
+// Allows prison staff to update a specific set of fields for defendant
+export const prisonSystemStaffUpdateRule: RolesRule = {
+  role: UserRole.PRISON_SYSTEM_STAFF,
+  type: RulesType.FIELD,
+  dtoFields: limitedAccessFields,
+}
```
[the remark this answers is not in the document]
thorhildurt: Will add tests to resolve this 😇

apps/judicial-system/web/src/routes/Prison/IndictmentOverview/IndictmentOverview.tsx
```diff
@@ -15,19 +16,46 @@ import {
...
+
+  const onChange = (updatedPunishmentType: PunishmentType) => {
+    defendant &&
+      limitedAccessUpdateDefendant({
+        caseId: workingCase.id,
+        defendantId: defendant.id,
+        punishmentType: updatedPunishmentType,
+      })
+    setPunishmentType(updatedPunishmentType)
+  }
```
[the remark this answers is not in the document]
thorhildurt: Always want to update the type on change, so moving the setter before the api call

apps/judicial-system/web/src/routes/Prison/IndictmentOverview/IndictmentOverview.tsx
```diff
@@ -69,6 +95,97 @@ const IndictmentOverview = () => {
...
+                onChange={() => {
+                  onChange(PunishmentType.SIGNED_FINE_INVITATION)
+                }}
+                large
+                backgroundColor="white"
+                label={formatMessage(strings.signedFineInvitation)}
+              />
+            </Box>
+          </BlueBox>
+        </Box>
```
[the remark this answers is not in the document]
thorhildurt: Currently not using this

apps/judicial-system/backend/src/app/modules/defendant/limitedAccessDefendant.controller.ts
thorhildurt: I only added tests that check the Roles rules config but not a test that enforces the logic it applies.

apps/judicial-system/backend/migrations/20241211095654-update-defendant.js
```diff
@@ -0,0 +1,26 @@
+'use strict'
+
+module.exports = {
+  async up(queryInterface, Sequelize) {
+    return queryInterface.sequelize.transaction((t) =>
+      Promise.all([
```
unakb, line 6: You don't technically need a Promise.all here because you're doing just one thing, so you could leave that out and just have queryInterface.addColumn(... straight away.

apps/judicial-system/backend/src/app/modules/defendant/models/defendant.model.ts
```diff
@@ -201,6 +202,14 @@ export class Defendant extends Model {
   @ApiPropertyOptional({ type: Boolean })
   isSentToPrisonAdmin?: boolean
 
+  @Column({
```
unakb, line 205: Nice! I see you picked up on our tendancy to store these types of fields as strings in the database but deal with them as ENUM's in the code. We used to store them as enums in the db as well but the constant migrations when adding or removing fields became tedious so we decided on this route instead 👌

apps/judicial-system/backend/src/app/modules/defendant/test/limitedAccessDefendantController/updateDefendant.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+      courtCaseNumber?: string,
+    ) => {
+      const then = {} as Then
+
+      await limitedAccessDefendantController
+        .updateDefendant(
+          caseId,
+          defendantId,
+          user,
+          { id: caseId, courtCaseNumber, type } as Case,
```
unakb, line 64: Tiny nitpick, since you don't have any tests right now that have a courtCaseNumber set you could just leave that out of the arguments.

apps/judicial-system/web/src/routes/Prison/IndictmentOverview/IndictmentOverview.strings.ts
```diff
@@ -26,4 +26,36 @@ export const strings = defineMessages({
     defaultMessage: 'Dómur',
     description: 'Titill á Dómur hluta',
   },
+  punishmentTypeTitle: {
+    id: 'judicial.system.core:indictment_overview.punishmentType_title',
+    defaultMessage: 'Refsitegund',
+    description: 'Titill á Refsitegund',
+  },
+  imprisonment: {
+    id: 'judicial.system.core:indictment_overview.punishmentType.imprisonment',
```
unakb, line 35: And same here, consider changing these strings in to this format for consistency. Also feel free to challenge it though if you think we should consider doing it differently ☺️

```
  punishmentTypeImprisonment: {
    id: 'judicial.system.core:indictment_overview.punishment_type_imprisonment',
```

apps/judicial-system/backend/src/app/modules/defendant/models/defendant.model.ts
thorhildurt: 💡 Ah okay I see, thanks for the context 🙏

apps/judicial-system/web/src/utils/hooks/useDefendants/index.ts
unakb: Yeah I feel like for this specific case it's fine for it to be like this now. The reason for the handling in the case mutations was because we have some shared components between prosecutors and defenders so instead of needing to always have an if else inside each component to decide which mutation to use we made the single hook that checks for us. 
Since right now this specific case is still really limited (haha) to this single use I'd say its fine to have them separate and we can then just combine them later if needed 🙌

apps/judicial-system/web/src/routes/Prison/IndictmentOverview/IndictmentOverview.tsx
```diff
@@ -16,18 +18,40 @@ import {
...
+  const { limitedAccessUpdateDefendant } = useDefendants()
 
   const { onOpen } = useFileList({
     caseId: workingCase.id,
   })
 
+  const { defendants } = workingCase
+  const defendant =
+    defendants && defendants?.length > 0 ? defendants[0] : undefined
+  const [selectedPunishmentType, setPunishmentType] = useState<PunishmentType>()
```
unakb, line 40: Consider updating the workingCase state with the new punishment type choice, that way you don't need to have the punishment type as a state here. This works perfectly well though and within this context you don't need to change it, but the reason we usually update the workingCase state when doing updates is that sometimes we have other things within a component that rely on the workingCase to make decisions, and we've run in to issues where the workingCase state is incorrect because we didn't update it accordingly after a related update and causes weird behaviour.

It would be something like this if you changed it 

```
const IndictmentOverview = () => {
  const { workingCase, setWorkingCase } = useContext(FormContext)
  const { limitedAccessUpdateDefendant, updateDefendantState } = useDefendants()
  
  const { defendants } = workingCase
  const defendant =
    defendants && defendants?.length > 0 ? defendants[0] : undefined

  const onChange = (updatedPunishmentType: PunishmentType) => {
    if (!defendant) return

    const defendantUpdate = {
      defendantId: defendant.id,
      caseId: workingCase.id,
      punishmentType: updatedPunishmentType,
    }
    updateDefendantState(defendantUpdate, setWorkingCase)
    limitedAccessUpdateDefendant(defendantUpdate)
  }
```

and then you could also replace the checked logic with 

`        checked={
                  defendant?.punishmentType ===
                  PunishmentType.INDICTMENT_RULING_DECISION_FINE
                }`

if you wanted to remove the hasSetPunishmentType function

unakb: This is just a suggestion though so it's up to you which way you go, both work 😊 Just wanted to point it out because it's a pattern you'll notice a lot that the workingCase gets updated either directly before or after the update, or as part of a "setAndSendToServer" hook which performs both at once

unakb (approved):
Really nice work! 👏 Going to Approve because all comments were just considerations but feel free to ask for review again if you have any other thoughts or changes you'd like feedback on

apps/judicial-system/web/src/routes/Prison/IndictmentOverview/IndictmentOverview.tsx
thorhildurt: Aha makes sense, I will address this since I also like to simplify the `boolean` checked property 👌

apps/judicial-system/web/src/routes/Prison/IndictmentOverview/IndictmentOverview.strings.ts
```diff
@@ -26,4 +26,36 @@ export const strings = defineMessages({
...
+    id: 'judicial.system.core:indictment_overview.punishment_type_probation',
+    defaultMessage: 'Skilorðsbundið',
+    description:
+      'Notaður sem texti í skilorðsbundið valmöguleika í Refsitegund',
+  },
+  punishmentTypeFine: {
+    id: 'judicial.system.core:court.indictment_overview.punishment_type_fine',
+    defaultMessage: 'Sekt',
+    description: 'Notaður sem texti í sekt valmöguleika í Refsitegund',
+  },
```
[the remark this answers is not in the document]
thorhildurt: Good catch on the "court." string 😇

PR #17199, opened 2024-12-11
feat(user-notification): Firebase Credentials Setup for Local Development

rafnarnason:
Fixes these types of strings for local development env.secret file so the app runs locally without manual edits to the env value

## Before: 
export FIREBASE_CREDENTIALS='{\n  \"type\": \"service_account\",\n  \"project_id\":
## After: 
export FIREBASE_CREDENTIALS='{
  ...
  "type": "service_account",
  "project_id": "mobile-app",
  ...
 }'

## What

## Why

## Screenshots / Gifs

## Checklist:

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved the escaping of secret values in console logs for better display of environment variables.


kksteini (approved):

apps/services/user-notification/README.md
```diff
@@ -40,6 +40,8 @@ Get secrets:
 yarn get-secrets user-notification
 ```
 
+> **Important**: For local development, the `FIREBASE_CREDENTIALS` environment variable needs to be cleaned of newlines. After running `get-secrets`, make sure to remove any newline characters from the `FIREBASE_CREDENTIALS` value in your environment.
```
baering, line 43: Thinking "out loud": What about keeping the dev-init as is and doing this inside [UserNotificationsConfig](https://github.com/island-is/island.is/blob/main/apps/services/user-notification/src/config.ts#L27) for local development?

Something similar to:

```typescript
let firebaseCredentials = env.required('FIREBASE_CREDENTIALS', '')

if (isRunningOnEnvironment('local')) {
  firebaseCredentials = firebaseCredentials.replace(/\n/g, '')
}

return {
  // ...
  firebaseCredentials,
  // ...
}
```

baering: Yes, it is something that was implemented 4 years ago when there was no standard way to check what environment you were running on. This was before we had a feature flag service. After ConfigCat this was phased out as it was not meant to be used as a way to "feature flag" code paths by environments.

For this use case _I think it should be fine_. But if you have a better way feel free, it was the first thing that came to my mind to showcase the idea.

eirikurn: How come it works in prod and not locally? Isn’t this just a formatting bug in our get-secrets script which we can fix?

rafnarnason: yes - that would sound like the right approach - developers should never have to touch these values - I have updated get-secrets and it works for this case ... just wonder if there could be any other unintended consequences

robertaandersen (approved):
LGTM

kksteini (approved):

eirikurn (approved):

PR #17236, opened 2024-12-13
feat(secondary school): New application (Umsókn um framhaldsskóla)

johannaagma:
## What

New application to apply for secondary school (framhaldsskóli)

## Checklist:


jonnigs (changes requested):

jonnigs:
On this screen I would suggest using `buildNationalIdWithNameField`, built in form that fetches name by nationalId.
Also there is a new component called `buildFieldsRepeaterField` it looks to me like it can replace the custom component.

Screenshot 2024-12-18 at 11 02 16

jonnigs:
This whole screen is custom, would it work to use `buildAsyncSelectField` and `buildFieldsRepeaterField`?

Screenshot 2024-12-18 at 11 09 58

jonnigs:
Should Icelandic be in the list if the title says "ef annað en Íslanska"
Screenshot 2024-12-18 at 11 13 36

jonnigs:
This is a custom component, `buildFormConclusionSection` does the same thing and looks identical
Screenshot 2024-12-18 at 11 18 00

libs/application/templates/secondary-school/src/lib/messages/conclusion.ts
```diff
@@ -0,0 +1,81 @@
...
+        'Umsókn þín í framhaldsskóla hefur verið móttekin! Hægt er að senda inn nýja umsókn til aa. júní 2025',
+      description: 'Conclusion general alert message for general application',
+    },
+    accordionTitle: {
+      id: 'ss.application:conclusion.general.accordionTitle',
+      defaultMessage: 'Hvað gerist næst?',
+      description: 'Conclusion accordion title',
+    },
+    accordionTextFreshman: {
+      id: 'ss.application:conclusion.general.accordionTextFreshman',
```
Ballioli, line 38: Vantar líklega #markdown hér til að styðja markdown á contentful.
'ss.application:conclusion.general.accordionTextFreshman#markdown'

libs/application/templates/secondary-school/src/lib/messages/conclusion.ts
```diff
@@ -0,0 +1,81 @@
...
+      defaultMessage:
+        `* **xx. júní 2025**` +
+        `\n\n Umsóknartímabili lýkur. Ekki verður farið að vinna úr umsóknum fyrr en eftir þessa dagsetningu. Ef þú af einhverjum ástæðum vilt gera breytingar þarftu að eyða núverandi umsókn og gera nýja áður en tímabilinu lýkur.\n` +
+        `* **yy. - zz. júní 2025**` +
+        `\n\n Unnið er úr umsóknum á þessu tímabili og verður tilkynning send á alla umsækjendur. Hnipp verður sent í gegnum island.is og ættir þú að fá tilkynningu í appið og tölvupóst um niðurstöðuna.`,
+      description: 'Conclusion accordion text freshman',
+    },
+    accordionTextGeneral: {
+      id: 'ss.application:conclusion.general.accordionTextGeneral',
```
Ballioli, line 47: Vantar líklega #markdown hér til að styðja markdown á contentful.
'ss.application:conclusion.general. accordionTextGeneral#markdown'

libs/application/templates/secondary-school/src/fields/SchoolSelection/selectionItem.tsx
```diff
@@ -0,0 +1,413 @@
...
+  const { application, setFieldLoadingState } = props
+  const { setValue } = useFormContext()
+  const [isLoadingPrograms, setIsLoadingPrograms] = useState<boolean>(false)
+
+  const isFreshman =
+    getValueViaPath<ApplicationType>(application.answers, 'applicationType') ===
+    ApplicationType.FRESHMAN
+
+  // options for dropdowns
+  const schoolOptions = application.externalData.schools
```
Ballioli, line 32: Losna við 'as' hérna eins og annarstaðar ?
getValueViaPath<SecondarySchool[]>(application.externalData, 'schools.data')

libs/application/template-api-modules/src/lib/modules/templates/secondary-school/secondary-school.service.ts
```diff
@@ -0,0 +1,224 @@
...
+          async (attachment) => {
+            const fileContent = await this.getAttachmentAsBase64(
+              application,
+              attachment,
+            )
+            return { fileContent }
+          },
+        ),
+      ),
+    })
```
Ballioli, line 157: Spurning um að bæta við catch hér og TemplateApiError sem segir eitthvað descriptive eins og "Ekki tókst að senda inn umsókn, vinsamlegast reyndu síðar...". Losnum við default "Villa kom upp" og jafnvel logga villuna ? (warning frekar en error líklega)

libs/application/template-api-modules/src/lib/modules/templates/secondary-school/secondary-school.service.ts
```diff
@@ -0,0 +1,224 @@
...
+      name: applicant?.name || '',
+      phone: applicant?.phoneNumber || '',
+      email: applicant?.email || '',
+      address: applicant?.address || '',
+      postalCode: applicant?.postalCode || '',
+      city: applicant?.city || '',
+      isFreshman: applicationType === SecondarySchoolApplicationType.FRESHMAN,
+      contacts: contacts,
+      schools: schoolSelection,
+      nativeLanguageCode: extraInformation?.nativeLanguageCode,
```
Ballioli, line 144: Defaulta á 'is' (eða hvernig sem formattið er). Submitt failar ef notandi velur ekkert tungumál.

Ballioli (approved):

jonnigs:
> > This is a custom component, `buildFormConclusionSection` does the same thing and looks identical Screenshot 2024-12-18 at 11 18 00
> 
> I need to be able to display the application overview when clicking on "Skoða yfirlit" Or some way, so that you see this formConclusion screen right after submitting the application, but if you open the application from e.g. minar-sidur, you are able to view this application overview.
> 
> So I created this custom component to "fake it" into looking like two separate screens. Is there any other way to do that?

Is it possible for you to have a `buildFormConclusionSection` with a condition to determine which case you are in?

Another thing you can have the last screen in the main form the conclusion screen and the submission on the second to last screen. You will still go to the last screen in the form even if the application has changed states and then if you go again to the same application or refresh you will load and see the form for the new state
 
When the application has moved into the `States.COMPLETED` state and you show the overview screen, it still has the edit buttons. As I understood your requirements last Friday, this is intentional but doesn't work at the moment?

jonnigs:
> > This whole screen is custom, would it work to use `buildAsyncSelectField` and `buildFieldsRepeaterField`?
> > Screenshot 2024-12-18 at 11 09 58
> 
> I need to graphql and get programs+thirdLanguages+nordicLanguages+requireDormitory based on school selection, and then use those values to populate the 4 dropdowns and control the checkbox visibility. Would I then be calling the same enpoint 5x? Also is there some buildAsyncCheckboxField?
> 
> Also had to use Controller+Select (instead of SelectController) to be able to clear dropdowns (firstProgram, secondProgram, thirdLanguage, nordicLanguage) when selecting a value in the school dropdown. Is that possible withhout custom component?

You should be able to fetch the other values based on the first selection, if not we should update the `buildAsyncSelectField` so that it can handle it.
Same goes for the clear action, if it can't be done as it is implemented, we should update it

libs/application/template-api-modules/src/lib/modules/templates/secondary-school/emailGenerators/applicationSubmittedEmail.ts
```diff
@@ -0,0 +1,80 @@
...
+
+export const generateApplicationSubmittedEmail: ApplicationSubmittedEmail = (
+  props,
+  recipient,
+): Message => {
+  const {
+    application,
+    options: { email, clientLocationOrigin },
+  } = props
+  const answers = application.answers as SecondarySchoolAnswers
```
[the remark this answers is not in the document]
jonnigs: Yes, excellent point. My main issue is the use of the "as" keyword since it is tricking the linter and the answers are not handed type safe to the application-template-loaders

libs/application/templates/secondary-school/src/lib/constants.ts
```diff
@@ -0,0 +1,26 @@
...
+  submitApplication = 'submitApplication',
+  deleteApplication = 'deleteApplication',
+}
+
+export enum Routes {
+  PERSONAL = 'personalMultiField',
+  CUSTODIAN = 'custodianMultiField',
+  EXTRA_INFORMATION = 'extraInformationMultiField',
+  SCHOOL = 'schoolMultiField',
+}
```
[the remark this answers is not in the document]
jonnigs: Well spotted, I'll update this!
I think types, interfaces and enums are so closely related so having them all in `types.ts` makes sense for me but I can also see it making sense having `types.ts`, `interfaces.ts` and `enums.ts`, especially for larger applications or projects.

libs/application/template-api-modules/src/lib/modules/templates/secondary-school/emailGenerators/applicationRejectedEmail.ts
```diff
@@ -0,0 +1,84 @@
...
+  const applicantNationalId = getValueViaPath<string>(
+    application.answers,
+    'applicant.nationalId',
+  )
+  const applicantName = getValueViaPath<string>(
+    application.answers,
+    'applicant.name',
+  )
+
+  const subject = 'Umsókn um framhaldsskóla - Umsókn eydd'
```
jonnigs, line 32:
```suggestion
  const subject = 'Umsókn um framhaldsskóla - Umsókn eytt'
```

libs/application/templates/secondary-school/src/graphql/queries.ts
```diff
@@ -0,0 +1,10 @@
+export const GET_PROGRAMS_BY_SCHOOL_ID = `
```
jonnigs, line 1: Nitpic about the naming: This is only a string, it's not getting anything. Consider renaming to `PROGRAMS_BY_SCHOOLS_ID_QUERY` or something similar.

jonnigs:
> > > > This is a custom component, `buildFormConclusionSection` does the same thing and looks identical Screenshot 2024-12-18 at 11 18 00
> > > 
> > > 
> > > I need to be able to display the application overview when clicking on "Skoða yfirlit" Or some way, so that you see this formConclusion screen right after submitting the application, but if you open the application from e.g. minar-sidur, you are able to view this application overview.
> > > So I created this custom component to "fake it" into looking like two separate screens. Is there any other way to do that?
> > 
> > 
> > Is it possible for you to have a `buildFormConclusionSection` with a condition to determine which case you are in?
> > Another thing you can have the last screen in the main form the conclusion screen and the submission on the second to last screen. You will still go to the last screen in the form even if the application has changed states and then if you go again to the same application or refresh you will load and see the form for the new state
> > When the application has moved into the `States.COMPLETED` state and you show the overview screen, it still has the edit buttons. As I understood your requirements last Friday, this is intentional but doesn't work at the moment?
> 
> Updated to remove the Conclusion custom component, but not sure which edit buttons you are referring to, they were never visible for me in the COMPLETED state
> 
> But the page looks like this after refresh in COMPLETED state:
> 
> image
> 
> But do you know of a way to create a back button "Aftur í yfirlit" to go back to minar-sidur (see design), or should this maybe instead use the buildMessageWithLinkButtonField?

Ahh ok, it looked differently for me but I couldn't submit the application, so I had to move things around to be able to view the last screen. No worries then.

Yes I think `buildMessageWithLinkButtonField` is going to be the best way. The built in back buttons are only taking you back in the form it self

libs/api/domains/secondary-school/src/lib/graphql/main.resolver.ts
```diff
@@ -0,0 +1,32 @@
...
+  constructor(private readonly secondarySchoolApi: SecondarySchoolApi) {}
+
+  @Scopes(ApiScope.menntamalastofnun)
+  @Query(() => [SecondarySchoolProgram])
+  async secondarySchoolProgramsBySchoolId(
+    @Args('schoolId', { type: () => String }) schoolId: string,
+    @Args('isFreshman', { type: () => Boolean }) isFreshman: boolean,
+    @CurrentUser() user: User,
+  ) {
+    return await this.secondarySchoolApi.getProgramsBySchoolId(
```
magnearun, line 26: Unnecessary async/await here ?

magnearun (approved):
Core files lgtm

jonarnarbriem (approved):

brynjarorng (approved):
LGTM

libs/application/template-api-modules/src/lib/modules/templates/secondary-school/utils.ts
```diff
@@ -0,0 +1,180 @@
...
+
+export const getCleanContacts = (
+  application: Application,
+): ApplicationContact[] => {
+  const result: ApplicationContact[] = []
+
+  // Parents
+  const parents = (
+    (application.externalData.nationalRegistryParents
+      .data as NationalRegistryParent[]) || []
```
jonnigs, line 89: getValueViaPath<NationalRegistryParent>

jonnigs (approved):

jonarnarbriem (approved):

PR #17247, opened 2024-12-16
fix(services-bff): Add missing exp check on the user endpoint

snaerth:
# Fix for missing expiration check on the user endpoint

## What

Add a utility function to check if the access token has expired  before attempting to refresh it. This change improves the  efficiency of the token refresh process by ensuring that  refresh operations are only triggered when necessary.

## Checklist:


eirikurn (approved):
LGTM

PR #17285, opened 2024-12-18
feat(j-s): Add punishment type tag column in prison cases overview table

thorhildurt:
## What
- [Asana task](https://app.asana.com/0/1199153462262248/1208902010654181)
- Render punishment type tag in the prison case table overview representing the added punishment type per case
- Add a sorting functionality for the punishment type
    - Note: Due to the current defendants data structure, we by default sort by the punishment type of the first defendant

## Why
- To give a better overview of the case list by enabling prison admin to group prison cases according to punishment type 

## Screenshots / Gifs
https://github.com/user-attachments/assets/43b67956-ebf2-4ba4-b231-69a8f2362171

## Checklist:

### DIFF
--- a/apps/judicial-system/api/src/app/modules/case-list/models/caseList.model.ts
+++ b/apps/judicial-system/api/src/app/modules/case-list/models/caseList.model.ts
@@ -11,6 +11,7 @@ import {
   CourtSessionType,
   IndictmentCaseReviewDecision,
   IndictmentDecision,
+  PunishmentType,
 } from '@island.is/judicial-system/types'
 
 import { Defendant } from '../../defendant'
@@ -142,4 +143,10 @@ export class CaseListEntry {
 
   @Field(() => String, { nullable: true })
   readonly indictmentCompletedDate?: string
+
+  // TEMP: Use with caution! This key will never be populated. 
+  // It was added to bypass table component type checks for a required custom sort key
+  // until we have a resolution on how to handle multiple defendants in the case list
+  @Field(() => PunishmentType, { nullable: true })
+  readonly defendantsPunishmentType?: PunishmentType
 }
--- a/apps/judicial-system/web/src/components/Table/Table.tsx
+++ b/apps/judicial-system/web/src/components/Table/Table.tsx
@@ -180,6 +180,8 @@ const Table: FC<TableProps> = (props) => {
     switch (column) {
       case 'defendants':
         return entry.defendants?.[0]?.name ?? ''
+      case 'defendantsPunishmentType':
+        return entry.defendants?.[0]?.punishmentType ?? ''
       case 'courtCaseNumber':
         return courtAbbreviation
           ? `${courtAbbreviation}: ${entry.courtCaseNumber}`
--- a/apps/judicial-system/web/src/components/Tags/utils.ts
+++ b/apps/judicial-system/web/src/components/Tags/utils.ts
@@ -2,6 +2,7 @@ import { TagVariant } from '@island.is/island-ui/core'
 import {
   isDistrictCourtUser,
   isPublicProsecutorUser,
+  PunishmentType,
 } from '@island.is/judicial-system/types'
 
 import {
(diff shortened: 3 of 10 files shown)


apps/judicial-system/web/src/components/Table/Table.tsx
```diff
@@ -180,6 +180,8 @@ const Table: FC<TableProps> = (props) => {
     switch (column) {
       case 'defendants':
         return entry.defendants?.[0]?.name ?? ''
+      case 'defendantsPunishmentType':
```
thorhildurt, line 183: This refers to the fields added [here](https://github.com/island-is/island.is/pull/17285/files#r1890553967)

apps/judicial-system/web/src/utils/arrayHelpers.ts
thorhildurt, line 1: This is just a proposal since I see we have a quite repetitive pattern to check, for example: `case.defendants && case.defendants.length > 0`, to access the first element of a given array. 

There are some utility array helpers in `node_modules`, which we could also use but decided to go with internal implementation to begin with since then its easy to navigate to the function scope within the repo, with the external module I was testing we could just navigate to the definition. But of-course open for feedback if you have other thoughts.

apps/judicial-system/api/src/app/modules/case-list/models/caseList.model.ts
```diff
@@ -142,4 +143,10 @@ export class CaseListEntry {
 
   @Field(() => String, { nullable: true })
   readonly indictmentCompletedDate?: string
+
+  // TEMP: Use with caution! This key will never be populated.
+  // It was added to bypass table component type checks for a required custom sort key
+  // until we have a resolution on how to handle multiple defendants in the case list
+  @Field(() => PunishmentType, { nullable: true })
+  readonly defendantsPunishmentType?: PunishmentType
```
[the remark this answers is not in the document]
thorhildurt: Will ask for team feedback on this

apps/judicial-system/web/src/components/Tags/utils.ts
```diff
@@ -119,3 +120,34 @@ export const getIndictmentRulingDecisionTag = (
...
+        return strings.punishmentTypeIndictmentRulingDecisionFine
+      case PunishmentType.SIGNED_FINE_INVITATION:
+        return strings.punishmentTypeSignedFineInvitation
+      default:
+        return strings.unknown
+    }
+  }
+
+  return {
+    color: 'red' as TagVariant,
```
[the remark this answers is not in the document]
thorhildurt: This is not what we want 😅 all punishment types should have the same color

apps/judicial-system/web/src/routes/Shared/Cases/PrisonCases.tsx
```diff
@@ -211,6 +218,20 @@ export const PrisonCases: FC = () => {
             {
               cell: (row) => <ColumnCaseType type={row.type} />,
             },
+            {
+              cell: (row) => {
+                const punishmentType = isNonEmptyArray(row.defendants)
+                  ? row.defendants[0].punishmentType
+                  : undefined
```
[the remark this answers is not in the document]
thorhildurt: We will default to the first defendant since the system only allows adding one. Supporting multiple will come in the future 👍

apps/judicial-system/web/src/components/Table/Table.tsx
```diff
@@ -180,6 +180,8 @@ const Table: FC<TableProps> = (props) => {
...
+        return entry.defendants?.[0]?.punishmentType ?? ''
```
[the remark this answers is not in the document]
thorhildurt: Will consider team feedback on this

gudjong, line 184: I would normally like to do this in a transformer in the api layer, but this is temporary code so it does not really matter in this case.

apps/judicial-system/web/src/routes/Shared/Cases/PrisonCases.tsx
```diff
@@ -211,6 +218,20 @@ export const PrisonCases: FC = () => {
             {
               cell: (row) => <ColumnCaseType type={row.type} />,
             },
+            {
+              cell: (row) => {
+                const punishmentType = isNonEmptyArray(row.defendants)
```
gudjong, line 223: Same as above, but as this is temporary code it does not really matter.

gudjong (approved):

apps/judicial-system/web/src/components/Table/Table.tsx
thorhildurt: As discussed async, we will remove this either when we:
- Represents multiple defendants per case differently in the client components
- Improve the table component, especially the sorting function, to make it more generic so we don't have to have these specific conditions in the table component itself.

apps/judicial-system/web/src/routes/Shared/Cases/PrisonCases.tsx
thorhildurt: Same as [above](https://github.com/island-is/island.is/pull/17285#discussion_r1892161450)

PR #17337, opened 2024-12-20
fix(j-s): Undefined recipient email

oddsson:
# Undefined recipient email

[Asana](https://app.asana.com/0/1199153462262248/1209019333325416/f)

## What

A few days ago we implemented notifications to judges and registrars in indictment cases when they are assigned to a case. There was a bug in that change where we do not null check values and accidentally try to send emails to `undefined`. This PR addresses this issue.

We now null check the user we are about to send this notification to and their email before placing the message on the queue. To be extra safe, we also null check the email right before we send the actual message.

## Why

This is a bug. The user should not be aware of the issue but this is filling our logs with error messages.

## Checklist:

## Summary by CodeRabbit

- **New Features**
	- Enhanced logic for case updates to ensure only relevant changes trigger updates.
	- Improved error handling in the notification system to prevent sending emails to invalid addresses.

- **Bug Fixes**
	- None.

- **Style**
	- Minor adjustments for better code readability and organization.


apps/judicial-system/backend/src/app/modules/notification/services/caseNotification/caseNotification.service.ts
```diff
@@ -698,6 +698,10 @@ export class CaseNotificationService extends BaseNotificationService {
     const official =
       role === UserRole.DISTRICT_COURT_JUDGE ? theCase.judge : theCase.registrar
 
+    if (!official?.email) {
+      return Promise.resolve({ success: false })
```
unakb, line 702: Consider returning success: true here instead, as returning success: false will trigger a retry (which will also fail) and spam our logs

PR #17347, opened 2024-12-23
feat(my-pages): WIP - Health - References

disaerna:
# My pages - Health - References - NEW ✨ 

## What
* References added - WIP - only hardcoded strings

## Checklist:

### DIFF
--- a/libs/portals/my-pages/health/src/lib/navigation.ts
+++ b/libs/portals/my-pages/health/src/lib/navigation.ts
@@ -140,6 +140,17 @@ export const healthNavigation: PortalNavigationItem = {
         },
       ],
     },
+    {
+      name: messages.references,
+      path: HealthPaths.HealthReferences,
+      children: [
+        {
+          name: messages.references,
+          path: HealthPaths.HealthReferencesDetail,
+          navHide: true,
+        },
+      ],
+    },
   ],
   description: m.healthDescription,
 }
--- a/libs/portals/my-pages/health/src/module.tsx
+++ b/libs/portals/my-pages/health/src/module.tsx
@@ -6,6 +6,7 @@ import { PortalModule } from '@island.is/portals/core'
 import { messages as hm } from './lib/messages'
 import { m } from '@island.is/portals/my-pages/core'
 import { Navigate } from 'react-router-dom'
+import ReferencesDetail from './screens/References/ReferencesDetail'
 
 const HealthOverview = lazy(() =>
   import('./screens/HealthOverview/HealthOverview'),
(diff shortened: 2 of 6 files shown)


libs/portals/my-pages/health/src/screens/Referrals/Referrals.graphql
```diff
@@ -0,0 +1,34 @@
...
+        name
+        profession
+        department
+      }
+      toContactInfo {
+        name
+        profession
+        department
+      }
+    }
```
thorkellmani, line 32: Maybe add a new endpoint, `healthDirectorateReferral` which accepts an `id`. Then the query only returns 1 result instead of a list which needs to be filtered.

libs/portals/my-pages/health/src/screens/Waitlists/Waitlists.graphql
```diff
@@ -0,0 +1,12 @@
...
+    waitlists {
+      id
+      name
+      waitBegan
+      lastUpdated
+      organization
+      status
+    }
+  }
+}
```
thorkellmani, line 12: Maybe add a new endpoint, `healthDirectorateWaitlist` which accepts an `id`. Then the query only returns 1 result instead of a list which needs to be filtered.

libs/portals/my-pages/health/src/module.tsx
```diff
@@ -255,5 +265,33 @@ export const healthModule: PortalModule = {
       enabled: userInfo.scopes.includes(ApiScope.healthVaccinations),
       element: <Vaccinations />,
     },
+    {
+      name: hm.referrals,
+      path: HealthPaths.HealthReferrals,
+      key: 'Referrals',
+      enabled: userInfo.scopes.includes(ApiScope.internal),
```
thorkellmani, line 272: Is this the correct scope for each new path?

libs/portals/my-pages/health/src/lib/navigation.ts
```diff
@@ -174,5 +174,27 @@ export const healthNavigation: PortalNavigationItem = {
...
+          name: 'id',
+          path: HealthPaths.HealthReferralsDetail,
+          navHide: true,
+        },
+      ],
+    },
+    {
+      name: messages.waitlists,
+      path: HealthPaths.HealthWaitlists,
+      children: [
```
thorkellmani, line 191: Any search terms to add?

thorkellmani (commented):
Few comments

libs/portals/my-pages/health/src/lib/navigation.ts
disaerna: AH forgot about them 🫡

magnearun (approved):
Core files lgtm

thorkellmani (approved):
LGTM

PR #17348, opened 2024-12-23
fix(j-s): Add missing court case number to info card

oddsson:
# Add missing court case number to info card

[Asana](https://app.asana.com/0/1199153462262248/1208824627882916/f)

## What

The InfoCard component displayed to prosecutor users in indictments that have either been sent to the court or ruled on should show the court case number. This PR implements that.

## Why

Prosecutors need to see the court case number to effectively work the case.

## Screenshots / Gifs

Screenshot 2024-12-23 at 10 01 57

## Checklist:

## Summary by CodeRabbit

- **New Features**
	- Enhanced case information display by adding the court case number to the InfoCard component.


thorhildurt (approved):
lgtm 👌  One thing I noticed in the figma screenshot vs the PR screen shot is that the info fields in the form are not in the same order. But if we need to fix that we can prob do that in a follow up task.

oddsson:
> lgtm 👌 One thing I noticed in the figma screenshot vs the PR screen shot is that the info fields in the form are not in the same order. But if we need to fix that we can prob do that in a follow up task.

The order of items in the info cards are all over the place. There is a task in Asana to implement a consistent order of info card items so it will be fixed there.

PR #17385, opened 2025-01-02
chore(j-s): add police case numbers to revoke indictment notification email

thorhildurt:
## What
- [Asana task](https://app.asana.com/0/1199153462262248/1208009480047499)
- Instead of sending "...ákæru í máli sem ekki hefur enn fengið málsnúmer" we want to add the `policeCaseNumbers` as stated in the task description 

## Why
- To add more context to what case is being revoked

## Checklist:

## Release Notes

- **New Features**
  - Enhanced case notification system with improved message formatting.
  - Added support for including police case numbers in court notifications.

- **Bug Fixes**
  - Corrected a typo in method name for revoke notifications.
  - Improved handling of case number references in notifications.

- **Improvements**
  - More detailed and precise messaging for indictment cancellation notices.
  - Simplified message structure for revoked indictment notifications.


apps/judicial-system/backend/src/app/modules/case/case.service.ts
```diff
@@ -860,6 +860,7 @@ export class CaseService {
         type: MessageType.DELIVERY_TO_COURT_INDICTMENT_CANCELLATION_NOTICE,
         user,
         caseId: theCase.id,
+        // Q: Why do we have to pass this flag? Why can't we rely on the court case numbers not being defined on the case object?
```
thorhildurt, line 863: Question targeting the flag `withCourtCaseNumber`, which is set to `false` when `CaseState.WAITING_FOR_CANCELLATION` is true 👀

unakb: Yeah this is a little confusing, I'll try to explain 😄

Basically the requirement here was to send an exact copy of the e-mail that gets sent to the court when a prosecutor asks for cancellation. If a prosecutor could only ask for cancellation _after_ the court sets a court case number, then we wouldn't need this. But we have cases where : 

1. Prosecutor creates case and sends to court
2. Prosecutor wants to cancel case, so he asks for cancellation
3. Email of cancellation is sent to the courts (with the text not including the court case number as it hasn't been set by the court yet)
3. When the court opens said case, they get a modal which tells them the case has been cancelled, and in this modal they set the court case number and close the case (to store the info in Auður)
4. Once they have set a court case number, a number of deliveries get pushed to the queue, including the cancellation notice 
5. Cancellation notice is picked up from the queue. At this point, the case actually _does_ have a court case number, but we want to send the text that was without the court case number to the robot (so that it matches the email sent to the court)

We can discuss further offline if you want, it's quite a specific piece of logic that isn't easy to infer from just the code, maybe we should add a comment or something

apps/judicial-system/backend/src/app/modules/case/internalCase.service.ts
```diff
@@ -672,23 +672,26 @@ export class InternalCaseService {
   ): Promise<DeliverResponse> {
     await this.refreshFormatMessage()
 
+    const courtCaseNumber =
```
unakb, line 675: Consider simplifying this so that the message just takes in a single variable called caseNumber or something along those lines, and then you can just send it straight in since there are no other changes to the string except which case number(s) are getting sent in. 

The string in notifications.ts could then simply be 

'{prosecutorsOffice} hefur afturkallað ákæru í máli {caseNumber}.',

where the caseNumber is decided with a const here depending on whether or not withCourtCaseNumber is true/false and the courtCaseNumber is set.

thorhildurt:
Addressed feedback and tested again both states when (1) court emails are sent, and (2) robot email are sent and both consistently contain the police case numbers in the body 👍 
Screenshot 2025-01-07 at 14 00 25
Screenshot 2025-01-07 at 14 00 30
Screenshot 2025-01-07 at 14 00 34

unakb (approved):
Nice ! Looks good, there's just one comment from the rabbit that needs to be addressed, there's one place that needs to be changed after the notification string change because it's still sending in policecaseNumbers/courtCaseNumber instead of caseNumber

PR #17420, opened 2025-01-07
chore(j-s): Move DefenderCasesTable to new table component

oddsson:
# Move DefenderCasesTable to new table component

[Asana](https://app.asana.com/0/322488613152836/1208997552579950/f)

## What

A few months ago we created a Table component. Since then we've been gradually moving all tables to that component and this PR implements that for the DefenderCasesTable.

## Why

Having one Table component for all tables is cleaner and less fragile than the alternative.

## Checklist:

## Summary by CodeRabbit

- **New Features**
	- Introduced `TagContainer` component for improved tag layout.
  
- **Refactor**
	- Simplified type declarations in the Defender Cases component.
	- Replaced custom table implementation with a more reusable Table component.
	- Removed dedicated CSS file for Defender Cases Table.

- **Code Cleanup**
	- Removed unnecessary imports and styling-related code.
	- Streamlined context menu and table rendering logic.

- **Performance**
	- Improved component structure for better maintainability.


apps/judicial-system/web/src/routes/Defender/Cases/components/DefenderCasesTable.tsx
```diff
@@ -82,188 +42,117 @@ export const DefenderCasesTable: FC<Props> = ({
...
+                type={row.type}
+                decision={row.decision}
+                parentCaseId={row.parentCaseId}
+              />
+            ),
+          },
+          { cell: (row) => <CreatedDate created={row.created} /> },
+          {
+            cell: (row) => (
+              <Box display="flex" columnGap={1} rowGap={1} flexWrap="wrap">
```
oddsson, line 106: This CSS makes it so that the tags are placed side by side if there is enough space and wraps to two lines if not.

thorhildurt: Is it only relevant to include it for this table? or would it make sense to have it in the `TagCaseState` component to make it shared styling?

thorhildurt (commented):
Niice! Looks good to me, only had one non-blocking question and then I noticed this when testing the checkbox selection: 
Screenshot 2025-01-08 at 11 20 40
When there are cases assigned to a defender, for a given selection combo, we only show the header. Not sure if we want that or if we rather want to show the standard info alert when no cases exist for a given selection?

apps/judicial-system/web/src/routes/Defender/Cases/components/DefenderCasesTable.tsx
oddsson: You're right, this applies to other places as well. The `TagCaseState` component renders a single tag, so this doesn't apply there but I've created a new container component to be able to reuse this style elsewhere. Could you please take another look? 🙏

thorhildurt (approved):
Looks good, just one formatting issue raised from the bunny

PR #17422, opened 2025-01-07
feat(web): Add field on generic list content type for display search in list

mannipje:
# Add field on generic list content type for display search in list

## What

- Add new boolean field on generic list content type to show search or not. 

## Why

- So its possible to hide search on generic list

### Before
image

### After
image

## Checklist:

## Summary by CodeRabbit

- **New Features**
	- Added an optional `showSearchInput` property to control the visibility of search inputs in Generic List components.
	- Introduced flexibility for conditionally displaying search inputs across different parts of the application.

- **Improvements**
	- Enhanced filtering and search functionality in Generic List components.
	- Updated data models and GraphQL fragments to support the new search input configuration.


libs/cms/src/lib/models/genericList.model.ts
```diff
@@ -30,6 +30,9 @@ export class GenericList {
 
   @CacheField(() => GetGenericListItemsInputOrderBy, { nullable: true })
   defaultOrder?: GetGenericListItemsInputOrderBy
+
+  @CacheField({ nullable: true })
```
RunarVestmann, line 34: 1. This doesn't have to be a cache field
2. Shouldn't we add the GraphQL type to this field? () => Boolean, {nullable: true}

RunarVestmann (approved):

PR #17423, opened 2025-01-07
fix(web): Fix hydration error

thorkellmani:
## What

* Fix nested `<a>` tags

## Why

* Nested `<a> `tags make `Next.js` report hydration errors

## Checklist:

## Summary by CodeRabbit

- **Style**
  - Updated button properties in the `OrganizationExternalLinks` component
  - Modified button focus behavior for external links


apps/web/components/Organization/Wrapper/OrganizationWrapper.tsx
```diff
@@ -679,7 +679,6 @@ export const OrganizationExternalLinks: React.FC<
                 pureChildren={true}
               >
                 <Button
-                  as="a"
```
RunarVestmann, line 682: Try tabbing through the page, when we skip the `as="a"` then need to tab twice to proceed to the next element (because we have an <a> tag and a nested <button> tag.

Try adding unfocusable={true] and see if that makes it right

thorkellmani: Seems to work fine with the `unfocusable` addition

RunarVestmann (approved):

PR #17438, opened 2025-01-08
fix(j-s): slow down the execution of js playwright tests

thorhildurt:
[Debug and research flakiness in our E2E tests
](https://app.asana.com/0/1199153462262248/1208944787802988)

## What
- Run the judicial system E2E tests slower and experiment if it will reduce flakiness 
    - To begin with, we add 1s between every action (click, type, etc.) using the `slowMo` option. Locally the tests now take 515s or ~8.5m to execute. 
    - Research trace available [here](https://www.notion.so/hellokolibri/E2E-research-17323303f8e6805cab28d4ba7a11f06b)

## Why
- There is some flakiness in our E2E tests when we are running them during deployment.
- Due to flakiness, we have to run the E2E 5-6 times until most tests have past during some of our executions

## Checklist:

## Summary by CodeRabbit

- **Tests**
	- Added a slow motion configuration for end-to-end tests in the judicial system to improve test visibility.
	- Introduced a comment regarding a feature flag affecting test implementation.

- **Chores**
	- Added a new constant to conditionally handle routing based on case type.


apps/judicial-system/web/src/routes/Prosecutor/Indictments/Processing/Processing.tsx
```diff
@@ -76,6 +76,8 @@ const Processing: FC = () => {
     deleteCivilClaimant,
   } = useCivilClaimants()
   const router = useRouter()
+
+  // TODO: currently for e2e tests, this is failing locally due to this feature flag
```
thorhildurt, line 80: Sidenote: Discovered this when debugging the test locally, but should be resolved in follow-up PRs but added this if others see a failing tests locally 👍

apps/system-e2e/src/playwright.config.ts
```diff
@@ -70,6 +70,12 @@ const config: PlaywrightTestConfig = {
     {
       name: 'judicial-system',
       testMatch: 'tests/judicial-system/**/*.spec.[tj]s',
+      use: {
```
thorhildurt, line 73: This should only impact the JS e2e tests 👍

unakb (approved):
Nice, let's see what this tells us, will get us moving in the right direction hopefully to eliminating the flakiness 🙏

apps/system-e2e/src/playwright.config.ts
AndesKrrrrrrrrrrr: Ah, JS=Judicial System, not JavaScript :facepalm:

AndesKrrrrrrrrrrr (approved):
e2e config LGTM :+1:

PR #17470, opened 2025-01-10
chore(j-s): extract formatter to apply dative case to court names

thorhildurt:
# [Fallbeyging dómstóls í tölvupósti](https://app.asana.com/0/1199153462262248/1209016117933812)

## What
- Fixing dative case to court name in notification type where it was not applied
- Create a new formatting function to centralise the case for all court names
- Only using new function to fix mentioned bug and apply it in two other related places. Then we have a follow up to clean this logic up in other files [here](https://app.asana.com/0/1199153462262248/1209139378163316) as well 👌 

## Why
- We apply case in emails for court names and it needs to be consistent
- Instead of applying case directly for every court name, we extract the current logic and use that function to apply the case

## Screenshots / Gifs

## Checklist:

## Release Notes

- **New Features**
  - Enhanced court name formatting for notifications.
  - Added support for converting court names to dative case in Icelandic.

- **Improvements**
  - Improved error handling in notification services.
  - More robust court name processing in email notifications.

- **Technical Updates**
  - Introduced a new utility function for court name transformations.
  - Added comprehensive test cases for court name formatting.


apps/judicial-system/backend/src/app/modules/notification/services/caseNotification/caseNotification.service.ts
```diff
@@ -875,7 +876,9 @@ export class CaseNotificationService extends BaseNotificationService {
       isIndictmentCase(theCase.type)
         ? this.formatMessage(notifications.caseCompleted.prosecutorBody, {
             courtCaseNumber: theCase.courtCaseNumber,
-            courtName: theCase.court?.name?.replace('dómur', 'dómi'),
+            courtName: theCase.court?.name
+              ? applyDativeCaseToCourtName(theCase.court?.name)
```
thorhildurt, line 880: This would be the pattern for all relevant places, since `courtName` can be `undefined`. We could also do: 
`applyDativeCaseToCourtName(theCase.court?.name || ")` rather then using the ternary op 🤔 But it could also be better to just assign `courtName` with a default name on fetch to avoid needing to handling it inline everywhere...which we are currently doing in many places, or not handling it at all

oddsson: My vote is on `applyDativeCaseToCourtName(theCase.court?.name || ")` or better yet, if applyDativeCaseToCourtName could handle an empty name then `applyDativeCaseToCourtName(theCase.court?.name)`

Applying this function globally before returning the case would mean that if we were ever to use the court name without the formatting, we'd need to make a change.

thorhildurt: Let's go with `applyDativeCaseToCourtName(theCase.court?.name || ")` 👍

oddsson (approved):

oddsson (approved):

PR #17471, opened 2025-01-10
chore(j-s): Rewrites Input Advocate

gudjong:
# Rewrites Input Advocate

[Rewrite advocate input](https://app.asana.com/0/1199153462262248/1209108787511068/f)
[label - lögmaður eða réttargæslumaður kemur ekki rétt út](https://app.asana.com/0/1199153462262248/1208429646655476/f)

## What
- Rewrites input advocate component.
- Fixes labels and placeholders when advocate is lawyer.

## Why

- Cleaner code and verified bug.

## Checklist:

## Summary by CodeRabbit

- **New Features**
  - Enhanced advocate information management with more granular input controls
  - Improved internationalization support for legal professional labels and placeholders

- **Bug Fixes**
  - Streamlined update processes for defender and civil claimant information
  - Simplified input validation and state management for advocate details

- **Refactor**
  - Centralized logic for sending updates to the server
  - Removed deprecated utility functions and simplified code structure
  - Updated type definitions for legal professionals

- **Chores**
  - Removed unused internationalization strings and associated files

The changes focus on improving the user experience for managing legal professional information in the judicial system application.


apps/judicial-system/web/src/components/DefenderInfo/DefenderInfo.tsx
```diff
@@ -28,25 +28,35 @@ interface Props {
...
-        return defenderInfo.investigationCases.sections.defender.heading
+        message = defenderInfo.investigationCases.sections.defender.heading
       } else {
-        return defenderInfo.investigationCases.sections.defender.title
+        message = defenderInfo.investigationCases.sections.defender.title
       }
     }
+
+    return formatMessage(message, {
+      defenderType:
```
thorhildurt, line 54: Nit 🤏
- This property is only required for the case when we have `!isRestrictionCase(workingCase.type) && !isProsecutionUser(user)` user, right? 
- Thus for minor code readability improvement, we could also set the formatting props dynamically to know when the property is needed

```
// idea 1
const hasDefenderType = !isRestrictionCase(workingCase.type) && !isProsecutionUser(user) 
const getFormattedDefenderType = (sessionArrangement: SessionArrangements) => sessionArrangements ===
      SessionArrangements.ALL_PRESENT_SPOKESPERSON
        ? 'Talsmaður'
        : 'Verjandi',

 return formatMessage(message, {
  ...(hasDefenderType ? { defenderType: getFormattedDefenderType(workingCase.SessionArrangement) } : {}
  })
```

```
// idea 2
let formattingProps = {}

const getFormattedDefenderType = (sessionArrangement: SessionArrangements) => sessionArrangements ===
      SessionArrangements.ALL_PRESENT_SPOKESPERSON
        ? 'Talsmaður'
        : 'Verjandi',

//...
else {
      if (isProsecutionUser(user)) {
        message = defenderInfo.investigationCases.sections.defender.heading
      } else {
        message = defenderInfo.investigationCases.sections.defender.title
        formattingProps = {defenderType: getFormattedDefenderType(workingCase.SessionArrangement)}
      }
}
return formatMessage(message, formattingProps)
```
I prefer the second option, since then we don't have to repeat checking a condition. But this is not a blocker at all 🙏

apps/judicial-system/web/src/components/DefenderInfo/DefenderInfo.tsx
```diff
@@ -155,17 +188,11 @@ const DefenderInfo: FC<Props> = ({ workingCase, setWorkingCase }) => {
...
-                        force: true,
-                      },
-                    ],
-                    workingCase,
-                    setWorkingCase,
-                  )
+                  handleSetAndSendCaseToServer({
+                    requestSharedWithDefender:
+                      RequestSharedWithDefender.COURT_DATE,
+                    force: true,
```
thorhildurt, line 194: Non-related to the changes: What does this `force` property indicate?

apps/judicial-system/web/src/routes/Court/Indictments/Advocates/SelectDefender.tsx
```diff
@@ -14,106 +14,111 @@ import {
...
+    )
+  }
+
+  const toggleDefendantWaivesRightToCounsel = (
+    defendant: Defendant,
+    defendantWaivesRightToCounsel: boolean,
+  ) => {
+    const update: UpdateDefendant = {
+      defendantId: defendant.id,
+      defenderNationalId: defendantWaivesRightToCounsel
```
thorhildurt, line 61: to avoid checking `defentantWaivesRightToCounsel` repeatedly, could we combine this in the following object? 
```
    ...(!defendantWaivesRightToCounsel
        ? {
            defenderNationalId: defendant.defenderNationalId,
            defenderName: defendant.defenderName,
            defenderEmail: defendant.defenderEmail,
            defenderPhoneNumber: defendant.defenderPhoneNumber,
            defenderChoice: DefenderChoice.WAIVE,
          }
        : { defenderChoice: DefenderChoice.DELAY }),
```
But this would mean that the null assignment of the properties would be `undefined`, but maybe that is sufficient?

apps/judicial-system/web/src/routes/Court/Indictments/Advocates/SelectDefender.tsx
```diff
@@ -151,12 +155,53 @@ const SelectDefender: FC<Props> = ({ defendant }) => {
...
+              defendantId: defendant.id,
+              defenderName,
+              defenderNationalId,
+              defenderEmail,
+              defenderPhoneNumber,
+            })
+          }
+          onAdvocateNotFound={setDefenderNotFound}
+          onEmailChange={(defenderEmail: string | null) =>
+            handleUpdateDefendantState({
```
thorhildurt, line 178: Nit: Do we have to update per property? Could we not have a single `onChange` or `onSave` that would handle both email and phone number change to reduce the number of props defined for `InputAdvocate`?

thorhildurt (commented):
Looks good! Just left few comments and nits ☺️

apps/judicial-system/web/src/components/DefenderInfo/DefenderInfo.tsx
gudjong: As discussed this is to make sure we override the value even if the fields being set already have a value. The reason for the flag is to facilitate autofill with default values that we only want to set once.

apps/judicial-system/web/src/routes/Court/Indictments/Advocates/SelectDefender.tsx
gudjong: We could and probably would if there were more than two properties, but since they are only two I decided to implement it this way. No correct way really.

apps/judicial-system/web/src/routes/Court/Indictments/Advocates/SelectDefender.tsx
gudjong: This code is not part of changes in this PR, it was only moved. We could do what you suggest, but we would need to set all the fields to null in the else clause. The value null has a different meaning than undefined: null makes it all the way to the database, clearing the database field, but undefined is ignored. I am not going to change this now, but we could do it in future PRs.

apps/judicial-system/web/src/components/DefenderInfo/DefenderInfo.tsx
gudjong: This code was simply moved here as part of the PR, so I didn't really think about this, but you are right. I'm going to change this in I sligthtly different way, simply returning a formatted text directly in the conditinal above.

apps/judicial-system/web/src/routes/Court/Indictments/Advocates/SelectDefender.tsx
thorhildurt: Ah makes sense, thanks 🙏

thorhildurt (approved):
Looks good 👍

PR #17496, opened 2025-01-14
fix(services-bff): Fix for multiple sid cookies

snaerth:
# Fix for multiple sid cookies

## What

Some users were experiencing a bug, specifically a redirect loop between Stjórnborð and IDS. The issue was caused by two cookies with the same name (sid) but different values. I thought res.clearCookie would handle this, but it does not seem to be sufficient.

To resolve this, I’ve updated the approach in this PR and changed the cookie names to be more identifiable:  `minarsidur-sid` and `stjornbord-bff`.

## Checklist:

## Summary by CodeRabbit

- **New Features**
	- Dynamic session cookie naming based on environment
	- Enhanced session cookie management with support for multiple cookies

- **Bug Fixes**
	- Improved session ID retrieval and handling across services
	- Better error handling with more context in authentication processes

- **Refactor**
	- Updated method signatures in session cookie and authentication services
	- Standardized session cookie service methods to accept both request and response objects


eirikurn (commented):
Renaming session cookies is disruptive during release, requiring all users to sign in.

Do you know how these "rare cases" actually happen? Let's walk through this in the office and see if there's some way to prevent it from happening rather then to deal with it after the fact.

Alternatively we should perhaps still support the session cookie and remove it later.

PR #17502, opened 2025-01-15
chore(my-pages): Slugify document names

thordurhhh:
## What

Slugify document names
`"Þetta er nafnið sem hér verður.pdf" = "thetta-er-nafnid-sem-her-verdur.pdf"`

## Why

File names can be of all shapes and sizes. And all kinds of peculiar characters.
We need to unify the pdf names.

## Checklist:

- [x] I have performed a self-review of my own code
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] Formatting passes locally with my changes
- [x] I have rebased against main before asking for a review

## Summary by CodeRabbit

- **New Features**
  - Improved filename generation for document downloads using the `slugify` library
  - Enhanced URL-friendly filename formatting with custom character replacements

- **Bug Fixes**
  - Removed unnecessary `Content-length` header from document downloads

PR #17508, opened 2025-01-15
fix(my-pages): hotfix - my pages docs and settings

thordurhhh:
Hotfix 🔥 

* https://github.com/island-is/island.is/pull/17505
* https://github.com/island-is/island.is/pull/17502


thorkellmani (approved):
LGTM

PR #17511, opened 2025-01-15
fix(shared): InputController relies on FormContext being provided

RunarVestmann:
# InputController relies on FormContext being provided

## What

This is causing issues in numerous places since the FormProvider isn't always wrapped around the InputController.

## Why

Instead of having to manually wrap FormProvider around InputControllers in numerous places in the codebase I suggest we instead make it so that the InputController functions even though the FormContext isn't provided (like it used to do).

## Checklist:

## Summary by CodeRabbit

- **Refactor**
	- Updated form context handling in input controller
	- Improved error prevention by adding conditional context checks
	- Enhanced robustness of form value setting mechanism

### DIFF
--- a/libs/shared/form-fields/src/lib/InputController/InputController.tsx
+++ b/libs/shared/form-fields/src/lib/InputController/InputController.tsx
@@ -99,7 +99,7 @@ export const InputController = forwardRef(
       step,
       clearOnChange,
     } = props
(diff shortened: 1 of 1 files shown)


norda-gunni (approved):
👍

disaerna (approved):

PR #17541, opened 2025-01-17
feat(license-service): Barcode session

magnearun:
## What

Store in Redis session id when user creates a barcode
Check when creating a barcode if there is a key in redis for this user, license type, if the session id is the same we create the barcode, else we do not. 

## Why

We do not want users to be able to request a barcode for a licence on different devices within a short amount of time. 
This will prevent users from being able to for example request a barcode for their driving license to get into a bar and the immediately login to their island.is app on another persons phone and request a barcode again for that person to "scam" their way in. 

## Screenshots / Gifs

## Checklist:

## Release Notes

- **New Features**
  - Enhanced session management for barcode creation
  - Added session identifier (sid) to authentication flow

- **Improvements**
  - Introduced configurable barcode and session expiration times
  - Improved error handling for concurrent barcode sessions

- **Configuration Changes**
  - Added new configuration parameters for barcode and session expiration
  - Updated JWT payload and authentication interfaces

- **Bug Fixes**
  - Prevented multiple active barcode sessions for the same user and license type

- **Type Enhancements**
  - Introduced new problem type for bad sessions in the problem management system


libs/services/license/src/lib/barcode.service.ts
```diff
@@ -4,13 +4,16 @@ import {
...
 import { ConfigType } from '@nestjs/config'
 import { Inject, Injectable } from '@nestjs/common'
-import { Cache as CacheManager } from 'cache-manager'
+import { Cache as CacheManager, Milliseconds } from 'cache-manager'
 import { sign, VerifyOptions, verify } from 'jsonwebtoken'
 import { LICENSE_SERVICE_CACHE_MANAGER_PROVIDER } from './licenseCache.provider'
 import { LicenseConfig } from './license.config'
 
 export const BARCODE_EXPIRE_TIME_IN_SEC = 60
+export const BARCODE_SESSION_EXPIRE_TIME_IN_SEC = 1800 // 30 minutes
```
eirikurn, line 13: It's been requested to make these two constants configurable with environment variables.

libs/api/domains/license-service/src/lib/licenseService.service.ts
```diff
@@ -478,6 +484,27 @@ export class LicenseService {
...
+      )
+
+      if (activeBarcodeSession && activeBarcodeSession !== user.sid) {
+        // If the user has an active session for the license type, we should not create a new barcode
+        this.logger.info(
+          `User has an active session for license: ${licenseType}`,
+        )
+
+        throw new ProblemError({
+          type: ProblemType.BAD_SUBJECT,
```
eirikurn, line 502: ~Should create a new BAD_SESSION problem type 👍

libs/api/domains/license-service/src/lib/licenseService.service.ts
```diff
@@ -478,6 +484,27 @@ export class LicenseService {
     const licenseType = this.mapLicenseType(genericUserLicenseType)
     const client = await this.getClient<typeof licenseType>(licenseType)
 
+    const barcodeSessionKey = user.sub
+      ? this.getBarcodeSessionKey(licenseType, user.sub)
+      : undefined
```
eirikurn, line 489: ~Could refactor this whole block into something like this: 

```
const barcodeSessionKey = this.checkBarcodeSession(user, licenseType)
```

eirikurn (approved):
Approving with simple comments 👍

thoreyjona:
@magnearun can we get this to main tomorrow? 🙏

libs/api/domains/license-service/src/lib/licenseService.service.ts
```diff
@@ -469,6 +471,34 @@ export class LicenseService {
...
+    if (barcodeSessionKey) {
+      const activeBarcodeSession = await this.barcodeService.getSessionCache(
+        barcodeSessionKey,
+      )
+
+      if (activeBarcodeSession && activeBarcodeSession !== user.sid) {
+        // If the user has an active session for the license type, we should not create a new barcode
+        this.logger.info(
+          `User has an active session for license: ${licenseType}`,
+        )
```
thorkellmani, line 492: Might be nice to log the license type as a property instead of in the string

thorkellmani (approved):
Approved with one minor note

PR #17469, opened 2025-01-10
feat(ci): Docker build rework

AndesKrrrrrrrrrrr:
Lays the groundwork for replacing the script-in-script Docker build approach. This uses with a GitHub Actions-native workflow, making the build process more efficient, scalable, and easier to maintain.

- Fully matrixes all inputted projects
- `codegen` must be run _before_ running this action
- Mostly works locally
- Uses docker build cache for system and yarn/npm dependencies

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1209109220675086

## Summary by CodeRabbit

- **Chores**
	- Added a new GitHub Actions workflow for Docker image building.
	- Implemented automated Docker build process triggered by pull requests, workflow calls, or manual dispatch.
	- Configured build matrix and caching strategies to optimize Docker image creation.
	- Set up AWS ECR login and Docker Buildx for efficient image management.

PR #17551, opened 2025-01-17
fix(CI): Use the docker `get-cache`

AndesKrrrrrrrrrrr:
Try to enable the unused cache.


AndesKrrrrrrrrrrr:
Not needed if we use #17469. Closing

PR #17555, opened 2025-01-17
chore: lint --fix

AndesKrrrrrrrrrrr:
Result from running:
```
yarn nx run-many -t lint --fix
```


apps/web/components/ApiCatalogueFilter/ApiCatalogueFilter.tsx
```diff
@@ -1,96 +0,0 @@
-import {
```
RunarVestmann, line 1: Why delete this file, it's a component being used here: apps/web/screens/Organization/StafraentIsland/ApiCatalogue.tsx

RunarVestmann (changes requested):

apps/web/components/AnchorNavigation/AnchorNavigation.tsx
```diff
@@ -1,92 +0,0 @@
-import { useEffect, useMemo, useRef, useState } from 'react'
```
norda-gunni, line 1: Why are these files being emptied and left in place?

AndesKrrrrrrrrrrr: fixed

apps/web/pages/[slug]/index.ts
```diff
@@ -1,9 +1,10 @@
-import '@island.is/api/mocks'
 import { getServerSidePropsWrapper } from '@island.is/web/utils/getServerSidePropsWrapper'
 
 import withApollo from '../../graphql/withApollo'
 import { withLocale } from '../../i18n'
 import articleScreen from '../../screens/Article'
+
+import '@island.is/api/mocks'
```
Comment (RunarVestmann) [inline apps/web/pages/[slug]/index.ts:7]: I'm not sure if moving the api mocks import has any effect as I've never used mocks myself.

apps/web/pages/_app.tsx
```diff
@@ -1,7 +1,9 @@
-import '@island.is/api/mocks'
-import { globalStyles } from '@island.is/island-ui/core'
 import App from 'next/app'
 
+import { globalStyles } from '@island.is/island-ui/core'
+
+import '@island.is/api/mocks'
```
RunarVestmann, line 5: I'm not sure if moving the api mocks import has any effect as I've never used mocks myself.

RunarVestmann (approved):
Quick glance and running lots of pages locally makes it seem like this change doesn't have any effect so I'll approve.

sigruntg (approved):

kksteini (approved):
Skimmed over codeowner files.
LGTM

apps/web/pages/_app.tsx
AndesKrrrrrrrrrrr: Looks like this would only affect `API_MOCKS=true` behavior, which isn't properly supported anyways :shrug:

PR #17568, opened 2025-01-20
feat(app): simplify token refresh logic app

Herdismaria:
[ticket](https://www.notion.so/Log-users-out-if-token-refresh-is-unauthorized-1605a76701d680c1bb52d5ad5053af17?pvs=4)

## What

When token refresh fails, we should log the user out and return to login screen.

the function `getAppRoot` is used to determine if the user should be on the login or home screen. `getAppRoot` calls the function `checkIsAuthenticated` function which could result in failed refresh. If refresh fails we logout and call `getAppRoot` to return the user to the login screen. This could result in infinite loop.

Solution:
When checking if user is logged in with `checkIsAuthenticated` we should skip logout&redirect if refresh fails and instead just logout and return false to `getAppRoot`.

## Checklist:

## Release Notes

- **New Features**
  - Enhanced token refresh mechanism with improved error handling
  - Added user-friendly login expiration messages in English and Icelandic

- **Bug Fixes**
  - Improved authentication flow to prevent potential infinite logout loops
  - More robust token management with deduplication of refresh requests

- **Improvements**
  - More granular control over logout behavior during authentication processes
  - Better handling of token expiration scenarios


apps/native/app/src/stores/auth-store.ts
```diff
@@ -47,10 +51,14 @@ interface AuthStore extends State {
   cognitoDismissCount: number
   cognitoAuthUrl?: string
   cookies: string
-  fetchUserInfo(skipRefresh?: boolean): Promise<UserInfo>
-  refresh(): Promise<void>
+  refreshPromise: null | Promise<void | null>
```
thoreyjona, line 54: Are you using this variable anywhere?

thoreyjona (approved):
Great work 🥇 Just one small comment.

PR #17574, opened 2025-01-20
fix(driving-license): Driving license delivery section updates

albinagu:
## Summary by CodeRabbit

- **New Features**
  - Introduced a refreshed logo and updated visual elements across the application.
  - Redesigned the driving license application form with enhanced delivery options, offering clear choices for postal or district pickup and a dynamic payment summary reflecting delivery fees.
  - Expanded the applicant information section with mandatory contact fields and clearer guidance messages.
  - Added new messaging prompts to enhance user interaction regarding delivery methods and application information.

- **Bug Fixes**
  - Corrected a typographical error in certificate requirement terminology.

- **Removed Elements**
  - Eliminated outdated logo and phone components for a more streamlined user experience.
  - Removed the `useMsg` hook and associated functionality for message formatting.
  - Removed the `subSectionPhone` component from the application form.


jonnigs (approved):

libs/api/domains/driving-license/src/lib/drivingLicense.spec.ts
```diff
@@ -393,6 +393,7 @@ describe('DrivingLicenseService', () => {
         needsToPresentHealthCertificate: false,
         needsToPresentQualityPhoto: false,
         licenseCategory: DrivingLicenseCategory.B,
+        sendLicenseInMail: 0,
```
kksteini, line 396: I see we're using numbers and booleans for sendLicenseInMail.

Technically it is mostly ok to mix `0` and `false`, `1` and `true`.
But why not just use `true` and `false` or just numbers?

Strict comparison will fail though:
```javascript
> Boolean(0)
false
> Boolean(1)
true
> true == 1
true
> true === 1
false
```

libs/application/template-api-modules/src/lib/modules/templates/driving-license-submission/driving-license-submission.service.ts
```diff
@@ -146,11 +146,12 @@ export class DrivingLicenseSubmissionService extends BaseTemplateApiService {
     const needsHealthCert = calculateNeedsHealthCert(answers.healthDeclaration)
     const remarks = answers.hasHealthRemarks === 'yes'
     const needsQualityPhoto = answers.willBringQualityPhoto === 'yes'
-    const jurisdictionId = answers.jurisdiction
+    const jurisdictionId = Number(getValueViaPath(answers, 'delivery.jurisdiction'))
```
kksteini, line 149: Future todo: 
It seems we're only doing boolean comparisons with this and then defaulting to 37 on false.
Maybe we should make sure that there isn't a jurisdiction with the number 0 since it would default to Kópavogur.
Since Boolean(Number(0)) is false

kksteini (approved):

PR #17598, opened 2025-01-21
chore: Update node-gyp, ssh2 and fix codemagic build

eirikurn:
We should not specify a node-gyp version in our dependencies. Most dependencies automatically depend on the "latest" version of node-gyp to build their native addons.

- Our previous version of node-gyp did not support recent version of python (and likely caused other issues with Node v22). - Updating ssh2 to get a more recent version of cpu-features which supports ARM macs.
- Configures codemagic to use same node version as we do elsewhere.
- Configures codemagic to not request reviews for new android builds, as it is causing an error.

## Summary by CodeRabbit

- **Configuration**
	- Updated Node.js version to 20.15.0 across project workflows
	- Modified Android publishing settings to handle changes differently

- **Dependency Management**
	- Removed `node-gyp` dependency from project dependencies

### DIFF
--- a/codemagic.yaml
+++ b/codemagic.yaml
@@ -14,6 +14,8 @@ definitions:
     #      - pattern: "main"
     #        include: true
     #        source: true
+  environment: &environment
+    node: 20.15.0
   shared_envs: &shared_envs
     XCODE_WORKSPACE: 'IslandApp.xcworkspace'
     CM_CLONE_DEPTH: 200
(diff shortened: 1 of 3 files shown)


codemagic.yaml
```diff
@@ -14,6 +14,8 @@ definitions:
     #      - pattern: "main"
     #        include: true
     #        source: true
+  environment: &environment
+    node: 20.15.0
```
AndesKrrrrrrrrrrr, line 18: Consider adding a CI check to verify that this is updated in tandem with our `package.json` node version

eirikurn: Good point. Kristofer has offered to address in a separate PR. 🙏

snaerth (approved):
LGTM 👏

PR #17666, opened 2025-01-27
fix(application-system): Fix typing for updateOnSelect

norda-gunni:
# ...

## What

## Why

## Screenshots / Gifs

## Checklist:

## Release Notes

- **Improvements**
  - Simplified asynchronous select field configuration by changing `updateOnSelect` from array to single value.
  - Updated type definitions for async select fields to accommodate both single strings and arrays.
  - Streamlined state management for async select form fields.

- **Technical Updates**
  - Refined handling of selected values in async select components.
  - Improved logic for tracking and updating selected values.


libs/application/types/src/lib/Fields.ts
```diff
@@ -53,7 +53,7 @@ export type Context = {
 export type AsyncSelectContext = {
   application: Application
   apolloClient: ApolloClient<object>
-  selectedValue?: string[]
+  selectedValue?: string
```
helgifr, line 56: Shouldn't this be `string | string[]`? Since if the field that is being watched with UpdateOnSelect has `isMulti: true` then selectedValue will have an array of strings. Like for example we make primaryAsyncSelect have `isMulti: true`
```suggestion
  selectedValue?: string | string[]
```

norda-gunni: Good point, fixed.

helgifr (approved):
lgtm 👍

HjorturJ (approved):

PR #17708, opened 2025-01-29
chore(license-api): Remove unused assets build config

AndesKrrrrrrrrrrr:
This assets config causes `project.json` to be copied to `dist` as well. For some reason, Nx doesn't respect `.gitignore` in docker build, so all Nx commands after build fail when building with Docker.

Removing the unused/unnecessary asset config causing the problem.

## Summary by CodeRabbit

- **Chores**
	- Updated build configuration for the license API service by removing asset handling during the build process.
	- Added comment lines in the main files of various services for clarity.


thorkellmani (approved):
LGTM

PR #17713, opened 2025-01-29
Chore/consolidate chunking

AndesKrrrrrrrrrrr:
This build upon https://github.com/island-is/island.is/pull/17654, but maintains the chunker; doing some legwork for fully parallel jobs.


AndesKrrrrrrrrrrr:
Not needed. Worked on elsewhere.

PR #17715, opened 2025-01-29
ci(docker): replace docker Alpine with Debian

busla:
# Replace Alpine as our docker runtime image with Debian
- Runs on the same base image as our ARC runners
- ARC runners will be able to do most of the heavy work instead of caching docker layers for deps
- The Alpine size has reached a point that we can just as well switch to a stable Debian based image


scripts/ci/30_test.sh
```diff
@@ -52,20 +52,15 @@ fi
...
-  --projects="${AFFECTED_PROJECTS}"
-  --exclude="${services_to_skip[*]}"
-  --target=test
-  --parallel="${NX_PARALLEL}"
-  --verbose
-  --no-watchman
-  --debug
-  --ci
-  --detectLeaks=false # Detecting leaks in CI is flaky and takes a long time
-  --detectOpenHandles # Prevents shipping improperly async-ed code
```
pro-ingvar, line 65: Do we not want this option?

pro-ingvar (approved):
LGTM, well done

PR #17716, opened 2025-01-29
feat(j-s): Zip all files for defenders in indictment cases

oddsson:
# Zip all files for defenders in indictment cases

[Asana](https://app.asana.com/0/1199153462262248/1209091103413232/f)
[Asana](https://app.asana.com/0/1199153462262248/1209091103413222/f)

## What

In R-cases, we offer defenders to download all case files as a ZIP file. This PR enables that for S-cases as well. If the case files are not shared with the defender, only the ruling and court record will be included. If they are, all case files visible to the defender will be included.

This PR also implements a minor change in the names of the files in the ZIP file. They now are capitalized and include Icelandic letters.

## Why

So that defender users don't have to open each individual file to print out all the case files.

## Screenshots / Gifs

Screenshot 2025-01-29 at 14 02 55

## Checklist:

## Release Notes

- **New Features**
  - Introduced a `ZipButton` component for downloading all case documents.
  - Enhanced file download capabilities for indictment and request cases.

- **Improvements**
  - Expanded file retrieval methods to support additional case types, including indictment cases.
  - Optimized file fetching process with concurrent AWS S3 file retrieval.
  - Improved handling of case file categories for defenders.

- **User Experience**
  - Simplified document download process for judicial system users.
  - Removed the previous button for downloading documents, streamlining the interface.


apps/judicial-system/backend/src/app/modules/case/limitedAccessCase.service.ts
```diff
@@ -519,18 +526,17 @@ export class LimitedAccessCaseService {
 
   async getAllFilesZip(theCase: Case, user: TUser): Promise<Buffer> {
```
oddsson, line 527: **Question to @gudjong**

This PR introduces quite a bit of complexity to this function. This function now decides what case files should be included in the ZIP file based on the type of the case and whether or not the the case files are shared with the defender or not.

I'm a bit afraid that a regression will lead to files being included in the ZIP file that should not be included and users will be able to see files that they are not supposed to. 

Would you / how would you add tests for this function? Would you write test for this function in the service or would you test the controller? Can we maybe pair on this?

apps/judicial-system/web/src/components/ZipButton/ZipButton.tsx
```diff
@@ -0,0 +1,32 @@
...
+  courtCaseNumber?: string | null
+}
+
+const ZipButton: FC<Props> = (props) => {
+  const { formatMessage } = useIntl()
+  const { caseId, courtCaseNumber } = props
+
+  return (
+    <a
+      href={`${api.apiUrl}/api/case/${caseId}/limitedAccess/allFiles`}
```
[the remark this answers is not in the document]
oddsson: No, I think this is fine.

apps/judicial-system/backend/src/app/modules/case/limitedAccessCase.service.ts
gudjong: Great questions. I am not sure what would be the best way to test but lean towards testing the controller. I´ll take a look at the code and pairing the tests sounds good.

apps/judicial-system/backend/src/app/modules/case/limitedAccessCase.service.ts
```diff
@@ -540,23 +546,71 @@ export class LimitedAccessCaseService {
...
+      )
+    }
 
     return this.zipFiles(filesToZip)
   }
+
+  private getFileCategories = (theCase: Case) => {
+    if (isRequestCase(theCase.type)) {
+      return defenderCaseFileCategoriesForRequestCases
+    } else {
```
gudjong, line 608: Since you are returning from the function within the if statement, please consider reducing nesting by eliminating the else cases.

apps/judicial-system/backend/src/app/modules/case/limitedAccessCase.service.ts
```diff
@@ -540,23 +546,71 @@ export class LimitedAccessCaseService {
...
+
+      filesToZip.push(
+        {
+          data: await this.pdfService.getIndictmentPdf(theCase),
+          name: 'Ákæra.pdf',
+        },
+        {
+          data: await this.pdfService.getCaseFilesRecordPdf(
+            theCase,
+            theCase.policeCaseNumbers[0],
```
gudjong, line 583: There is a separate case files record for each police csae number, so we need to return them all. Also not that each case files record is identified by the police case number, not the court case number. Please make the necessary modifications.

apps/judicial-system/backend/src/app/modules/case/limitedAccessCase.service.ts
```diff
@@ -540,23 +546,71 @@ export class LimitedAccessCaseService {
...
-        data: await this.pdfService.getCourtRecordPdf(theCase, user),
-        name: 'þingbok.pdf',
-      },
-      {
-        data: await this.pdfService.getRulingPdf(theCase),
-        name: 'urskurður.pdf',
-      },
-    )
+    if (isIndictmentCase(theCase.type)) {
+      const defendant = theCase.defendants && theCase.defendants[0]
```
gudjong, line 571: We are already sending subpoeanas for all defendants to the court and police. Even if no indictments are created at this time with more than one defendant it is easy to have the implementation handle multiple defendants. Please consider suppurting multiple defendants here so we don´t need to revisit this code when prosecutors start creating indictments with multiple defendants.

gudjong (changes requested):
This looks really good. I have one change request regarding multiple case files records and one suggestion to support multiple defendants right away so we do not need to revisit this code when prosecutors start sending us indictments with multiple defendants.

I am also wondering if we shouldn't also include files from merged cases. The defenders can access from the client and including them in the ZIP file sounds prudent.

apps/judicial-system/backend/src/app/modules/case/limitedAccessCase.service.ts
```diff
@@ -520,43 +525,103 @@ export class LimitedAccessCaseService {
...
+        {
+          data: await this.pdfService.getCourtRecordPdf(theCase, user),
+          name: 'Þingbók.pdf',
+        },
+        {
+          data: await this.pdfService.getRulingPdf(theCase),
+          name: 'Úrskurður.pdf',
+        },
+      )
     }
```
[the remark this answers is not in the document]
gudjong: Good point, but I want to catch errors for individual files. Will make a change.

PR #17758, opened 2025-01-31
feat(app-sys): shared overview component

jonnigs:
## What

Make a shared component that handles the overview screen

## Why

So it doesn't need to be custom every time

## Screenshots / Gifs

Screenshot 2025-02-14 at 12 35 00

## Checklist:

## Summary by CodeRabbit

- **New Features**
  - Introduced a new configuration mechanism for overview fields, allowing detailed displays of applicant information, numerical summaries, and file attachments.
  - Added new UI components to render comprehensive overview sections and file details.
  - Expanded messaging definitions for clearer, localized text in overviews.

- **Refactor**
  - Replaced legacy overview components and custom builders with streamlined, modular overview field configurations.
  - Updated integration and export structures for enhanced maintainability.

- **Style**
  - Added a thicker divider variant to improve visual separation in the interface.


libs/island-ui/core/src/lib/Divider/Divider.css.ts
```diff
@@ -15,4 +15,5 @@ export const weight = styleVariants({
   blue300: { background: theme.color.blue300 },
   blueberry200: { background: theme.color.blueberry200 },
   red200: { background: theme.color.red200 },
+  thick: { background: theme.border.color.black, height: '2px' },
```
Toti91, line 18: This config seems to just be controlling the color, so this should rather be just "black", and then we should create another styleVariant for "thickness" which is default 1px

libs/application/types/src/lib/Fields.ts
```diff
@@ -826,6 +828,39 @@ export interface DisplayField extends BaseField {
...
+  fileType?: FormText
+}
+
+export interface OverviewField extends BaseField {
+  readonly type: FieldTypes.OVERVIEW
+  component: FieldComponents.OVERVIEW
+  title: FormText
+  description?: FormText
+  backId?: string
+  fieldType: 'keyValue' | 'attachments'
```
Toti91, line 852: Maybe we should add "table" as well?

Toti91 (changes requested):
Very nice, 2 things

HjorturJ (approved):

libs/island-ui/core/src/lib/Divider/Divider.css.ts
jonnigs: This const is named `weight`, so I thought in made sense that is does more than just control the color...

libs/application/types/src/lib/Fields.ts
Toti91: I don't remember a specific application but it there are some. It's useful when displaying data from tablerepeater f.x.

libs/island-ui/core/src/lib/Divider/Divider.css.ts
Toti91: Yeah the name is wrong, should be renamed to color and then another config added named weight

Toti91 (approved):
LGTM

Toti91 (approved):

PR #17768, opened 2025-01-31
chore(ci): Use dynamic node version

AndesKrrrrrrrrrrr:
- Re-add removed version script
- Use only the major node version
- Use script instead of static (hardcoded) versions


AndesKrrrrrrrrrrr:
Possibly outdated. Closing.

PR #17847, opened 2025-02-05
fix(native-app): if organ donation errors and there is no data, only show error message

thoreyjona:
## What

When organ donation service is down we were accidentally showing half of the data (defaulting to user not being an organ donor even if we know nothing about it) along with an error message, when we should only be showing the error message if we don't have any data. 

## Screenshots / Gifs
Before:

After:

## Checklist:

### DIFF
--- a/apps/native/app/src/screens/health/health-overview.tsx
+++ b/apps/native/app/src/screens/health/health-overview.tsx
@@ -550,36 +550,37 @@ export const HealthOverviewScreen: NavigationFunctionComponent = ({
               }
             />
           )}
-          {isOrganDonationEnabled && (
-            <InputRow background>
-              <Input
-                label={intl.formatMessage({
-                  id: isOrganDonorWithLimitations
-                    ? 'health.organDonation.isDonorWithLimitations'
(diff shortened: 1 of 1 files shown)


apps/native/app/src/screens/health/health-overview.tsx
```diff
@@ -550,36 +550,37 @@ export const HealthOverviewScreen: NavigationFunctionComponent = ({
...
-                value={`${intl.formatMessage(
-                  {
+          {isOrganDonationEnabled &&
+            (organDonationRes.loading || organDonationRes.data) && (
+              <InputRow background>
+                <Input
+                  label={intl.formatMessage({
                     id: isOrganDonorWithLimitations
-                      ? 'health.organDonation.isDonorWithLimitationsDescription'
+                      ? 'health.organDonation.isDonorWithLimitations'
```
thoreyjona, line 559: Git is diffing this wrong 😅 I did only add the (organDonationRes.loading || organDonationRes.data) && at the top, nothing else changed

snaerth (approved):
LGTM

PR #17875, opened 2025-02-07
fix(application-ir): Provider failure when no estates found

kksteini:
User needs to be halted in the process if no estate data is found for them instead of continuing onwards to an empty drop down list.

### Before
Allowed through but since no data was found, the user can only interact with an empty drop down list of *estates*.
image

### After
image

## Checklist:

## Summary by CodeRabbit

- **New Features**
	- Enhanced error handling now provides clear, user-friendly messages when essential inheritance report data is missing.
	- Updated messaging guides users in situations where estate information cannot be located, offering concise feedback and direction.


jonnigs (approved):

libs/application/core/src/lib/messages.ts
```diff
@@ -385,6 +385,13 @@ export const coreErrorMessages = defineMessages({
     description:
       'Text that appears when district comissioner has validation fails in pre-data',
   },
+  errorDataProviderEstateValidationNothingFoundSummary: {
+    id: 'application.system:core.error.dataProviderEstateValidationNothingFoundSummary',
+    defaultMessage:
+      'Engin dánarbú fundust. Vinsamlegast hafðu samband við Sýslumenn ef þú telur að um villu sé að ræða.',
+    description:
+      'Text that appears when district comissioner has no estate data in pre-data',
+  },
```
jonnigs, line 394: Is this something that belongs in the core messages? Shouldn't this be in the message file within the applications template?

albinagu (approved):

libs/application/core/src/lib/messages.ts
kksteini: Ok, refactored it to pull from the application messages instead. Thanks :D

albinagu (approved):

libs/application/core/src/lib/messages.ts
kksteini: Well, ever since I did this change I've tried numerous ways of getting codegen to work (after importing messages from an application to the application template modules).

However, this seems to completely empty out the schema that is generated and I can not figure out why.
I've moved the error messages back into the core error messages, but I do notice there are other there.

I've found I think 1 or 2 instances of template-api-modules that do include messages from their own applications but I can not figure out what they're doing differently that allows them to not run into these schema issues.

libs/application/core/src/lib/messages.ts
```diff
@@ -390,6 +390,13 @@ export const coreErrorMessages = defineMessages({
     description:
       'Text that appears when district comissioner has validation fails in pre-data',
   },
+  errorDataProviderEstateValidationNothingFoundSummary: {
+    id: 'ir.application:errorDataProviderEstateValidationNothingFoundSummary',
```
kksteini, line 394:
```suggestion
    id: 'application.system:core.error.dataProviderEstateValidationNothingFoundSummary',
```

PR #17879, opened 2025-02-07
fix(j-s): Case completed without ruling

unakb:
[Asana](https://app.asana.com/0/1199153462262248/1209350537324465)

## What

Separate case completed without ruling from case decision. 

## Why

Because judges still make decisions in cases without rulings. 

## Checklist:

### DIFF
--- a/apps/judicial-system/backend/migrations/20250206152023-update-case.js
+++ b/apps/judicial-system/backend/migrations/20250206152023-update-case.js
@@ -1,0 +1,68 @@
+'use strict'
+
+const replaceEnum = require('sequelize-replace-enum-postgres').default
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    // Add the new boolean field
+    await queryInterface.addColumn('case', 'is_completed_without_ruling', {
+      type: Sequelize.BOOLEAN,
+      allowNull: true,
+    })
+
+    await queryInterface.sequelize.query(`
+      UPDATE "case"
+      SET "is_completed_without_ruling" = true
+      WHERE "decision" = 'COMPLETED_WITHOUT_RULING'
+    `)
+
+    await queryInterface.sequelize.query(`
+      UPDATE "case"
+      SET "decision" = 'ACCEPTING'
+      WHERE "decision" = 'COMPLETED_WITHOUT_RULING'
+    `)
+
+    // Replace the enum values
+    await replaceEnum({
+      queryInterface,
+      tableName: 'case',
+      columnName: 'decision',
+      newValues: [
+        'ACCEPTING',
+        'REJECTING',
+        'ACCEPTING_ALTERNATIVE_TRAVEL_BAN',
+        'ACCEPTING_PARTIALLY',
+        'DISMISSING',
+      ],
+      enumName: 'enum_case_decision',
+    })
+  },
+
+  down: async (queryInterface) => {
+    // Revert the enum values
+    await replaceEnum({
+      queryInterface,
+      tableName: 'case',
+      columnName: 'decision',
+      newValues: [
+        'ACCEPTING',
+        'REJECTING',
+        'ACCEPTING_ALTERNATIVE_TRAVEL_BAN',
+        'ACCEPTING_PARTIALLY',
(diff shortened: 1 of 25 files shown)


apps/judicial-system/backend/migrations/20250206152023-update-case.js
```diff
@@ -0,0 +1,68 @@
+'use strict'
+
+const replaceEnum = require('sequelize-replace-enum-postgres').default
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
```
gudjong, line 6: Please do all these updates in a single transaction. Otherwise, the db may end up in an inconsistent state. Updating the enum cannot be done in a transaction, but the rest can. Updating the enum would just need to be done after all the other updates are completed.

apps/judicial-system/backend/migrations/20250206152023-update-case.js
```diff
@@ -0,0 +1,68 @@
...
+      UPDATE "case"
+      SET "is_completed_without_ruling" = true
+      WHERE "decision" = 'COMPLETED_WITHOUT_RULING'
+    `)
+
+    await queryInterface.sequelize.query(`
+      UPDATE "case"
+      SET "decision" = 'ACCEPTING'
+      WHERE "decision" = 'COMPLETED_WITHOUT_RULING'
+    `)
```
gudjong, line 23: The two updates could be done in a single update query, resulting in quicker execution. Please consider merging the two.

apps/judicial-system/backend/migrations/20250206152023-update-case.js
```diff
@@ -0,0 +1,68 @@
...
+        'ACCEPTING_ALTERNATIVE_TRAVEL_BAN',
+        'ACCEPTING_PARTIALLY',
+        'DISMISSING',
+        'COMPLETED_WITHOUT_RULING', // Adding back the removed value
+      ],
+      enumName: 'enum_case_decision',
+    })
+
+    // Revert the enum value changes
+    await queryInterface.sequelize.query(`
```
gudjong, line 59: Please use a transaction.

apps/judicial-system/web/src/routes/Court/InvestigationCase/Ruling/Ruling.tsx
```diff
@@ -395,7 +392,7 @@ const Ruling = () => {
                   [
                     {
                       conclusion:
-                        decision === CaseDecision.ACCEPTING
+                        decision === CaseDecision.ACCEPTING && isRulingRequired
```
gudjong, line 395: Should we not set `conclusion` to `prosecutorDemands` rather than `demands`? `prosecutorDemands` are autofilled from `demands` and the court has a chance to edit that filed as it sees fit.

gudjong (approved):
Looks mostly good to me. I have a few comments suggestions, but approving.

apps/judicial-system/backend/migrations/20250206152023-update-case.js
unakb: I changed it to use transactions except for the replaceEnum part since that doesn't support transactions

apps/judicial-system/web/src/routes/Court/InvestigationCase/Ruling/Ruling.tsx
unakb: Discussed offline - demands has been used here instead of prosecutor demands for a few years and the autofill is only partially used in the conclusion, mostly rewritten, so we consider this a non-issue  for now at least.

apps/judicial-system/backend/src/app/modules/case/limitedAccessCase.service.ts
```diff
@@ -115,6 +115,7 @@ export const attributes: (keyof Case)[] = [
   'indictmentReviewDecision',
   'indictmentReviewerId',
   'hasCivilClaims',
+  'isCompletedWithoutRuling',
```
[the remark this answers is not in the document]
unakb: No because we don't want limited access users updating this field.

PR #17888, opened 2025-02-10
fix(email-service): Make heading use a tag instead of h1

magnearun:
Outlook mobile app render's text that looks like a url as a link, For example, our notification messages contain the text Ísland.is, but Outlook mobile incorrectly converts it into a hyperlink while dropping the Í, likely because it is an Icelandic character. As a result, our notifications always include a link to sland.is. Malicious actors took advantage of this behaviour in January by setting up a fake sland.is website that mimicked the official Ísland.is site.

## What

Adding a anchor tag without a href around the text is supposed to fix this. Will need testing on dev to confirm that this works, test email link look fine. 

## Screenshots / Gifs

## Checklist:


unakb (approved):
lgtm!

gudjong (approved):
LGTM

Pretty clever hack 😅

PR #17896, opened 2025-02-10
chore(ci): `codegen` in Dockerfile

AndesKrrrrrrrrrrr:
Without `codegen` the build will fail. Currently we do `codegen` on the runner, and copying generated files with `COPY . .`.

Running `codegen` when the files are present/copied already, should add minimal extra time.

### DIFF
--- a/scripts/ci/Dockerfile
+++ b/scripts/ci/Dockerfile
@@ -39,6 +39,12 @@ FROM deps AS src
 
 COPY . .
 
+ARG NX_TASKS_RUNNER=ci
+RUN --mount=type=secret,id=nx_cloud_access_token \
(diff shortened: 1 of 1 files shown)


busla (commented):
NX cloud cache is meant to be used for _affected_ runs, not as a general cache service. 

Running codegen can take up to [~5 minutes (cached)](https://github.com/island-is/island.is/actions/runs/13266209859/job/37034248116) which would be added to each job.

Furthermore, we are constrainted by NX cloud limits and _codegen_ is the reason why we are surpassing it.

AndesKrrrrrrrrrrr:
Fixed in #17963

PR #17908, opened 2025-02-11
feat(ci): `set-cache-url` action

AndesKrrrrrrrrrrr:
Create a new action to set the cache URL.

This is useful to set the cache url to a self-hosted cache, which subsequent actions will use when they use the standard https://github.com/actions/cache. This means that when using actions like `actions/setup-node` which uses the `actions/cache` action for caching, it will transparently use the self-hosted cache set with `set-cache-url`.

### DIFF
--- a/.github/actions/get-cache/action.yml
+++ b/.github/actions/get-cache/action.yml
@@ -20,18 +20,11 @@ outputs:
 runs:
   using: 'composite'
   steps:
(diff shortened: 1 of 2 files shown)


AndesKrrrrrrrrrrr:
Not needed, since we moved to `runs-on/cache`

PR #17927, opened 2025-02-11
feat(native-app): inbox bulk selection

thoreyjona:
## What

Add bulk selection to inbox documents.
* Select documents button in top right corner
* When clicking on select button another button Select all appears in the top left corner
* Icons change to circles and if the user clicks on any list item it gets selected
* If the user has at least one item selected an action bar appears at the bottom of the screen showing three actions Star, Archive and Mark as read.
* If the action returns success for the selected items we display a toast with a success message and update the cache for the documents (did not manage to get the updating of the cache work for archiving documents so I refresh the data in that case)
* By clicking the icon in the list we can take a short cut into the "select state" (skipping clicking the top right button)
* Added back the star for starred items in the list but we only show it now if the document is starred, not if it is not and it is not a button like it was before

https://github.com/user-attachments/assets/af7d2e9b-1f1c-4b57-9eee-7ad28a17713b

## Checklist:

## Summary by CodeRabbit

- **New Features**
  - Introduced bulk selection in the inbox, enabling users to select individual or all messages, with options to deselect and cancel selections.
  - Added an interactive action bar that allows performing bulk actions such as starring, archiving, and marking messages as read.
  - Implemented toast notifications to provide immediate feedback on the status of bulk actions.
  - Enhanced the visual presentation of message cards and list items to clearly indicate selection and bookmarking statuses.


apps/native/app/src/screens/inbox/components/action-bar.tsx
```diff
@@ -0,0 +1,142 @@
...
+
+export const ActionBar = ({
+  onClickStar,
+  onClickArchive,
+  onClickMarkAsRead,
+  loading = false,
+}: {
+  onClickStar: () => void
+  onClickArchive: () => void
+  onClickMarkAsRead: () => void
```
thoreyjona, line 55: If we want to make this a reusable component we can remove these actions from it and extract it, I asked Daði and he did not think we would use it in another places. I wanted to try to add as little code I could to the inbox.tsx file since it is already huge so I took this approach ☺️

apps/native/app/src/screens/inbox/inbox.tsx
```diff
@@ -675,6 +844,113 @@ export const InboxScreen: NavigationFunctionComponent<{
...
+          }}
+          onClickArchive={async () => {
+            const result = await bulkSelectActionMutation({
+              variables: {
+                input: { action: 'archive', documentIds: selectedItems },
+              },
+            })
+            if (result.data?.postMailActionV2?.success) {
+              resetSelectState()
+              res.refetch()
```
thoreyjona, line 893: I tried figuring out how to do this in the cache like for the other two operations but could not make it work despite following the documentation from apollo. I don't have any easy way to vizualise the cache since we upgraded to React Native 0.74 since then Flipper is not supported like it was before. Spent some time trying to set it up manually but no luck. Decided it was not worth the effort for now so refetching the data instead.

apps/native/app/src/screens/inbox/inbox.tsx
```diff
@@ -675,6 +844,113 @@ export const InboxScreen: NavigationFunctionComponent<{
...
+              showToastForBulkSelectAction({
+                variant: 'success',
+                title: intl.formatMessage({
+                  id: 'inbox.bulkSelect.archiveSuccess',
+                }),
+              })
+            } else {
+              showToastForBulkSelectAction({
+                variant: 'success',
```
eirikurn, line 902: ~Should we use another variant for errors?

eirikurn (approved):
LGTM

apps/native/app/src/screens/inbox/inbox.tsx
thoreyjona: Great catch!

apps/native/app/src/screens/inbox/components/toast.tsx
```diff
@@ -0,0 +1,122 @@
...
+
+  if (!visible) {
+    return null
+  }
+
+  return (
+    <Host
+      style={{ width: screenWidth - theme.spacing[4] }}
+      entering={FadeInDown}
+      exiting={FadeOutDown}
```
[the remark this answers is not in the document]
thoreyjona: Yes at least seems to work 🤔 
https://github.com/user-attachments/assets/5700eb98-e5de-4000-a0ca-32d441ca5df0

PR #17928, opened 2025-02-12
feat(j-s): Don't complete indictments until they are sent to the public prosecutors office

oddsson:
# Don't complete indictments until they are sent to the public prosecutors office

[Asana](https://app.asana.com/0/1199153462262248/1209309645508750)

## What

Don't move indictment cases to the past cases table, even though they are completed, until it has been sent to the public prosecutors office.

## Why

There is some work to be done after indictments are completed with a ruling or a fine. That might be forgotten if the case is moved down immediately. 

## Checklist:

## Summary by CodeRabbit

- **New Features**
  - Enhanced active case filtering now offers a more precise view of cases based on status and user roles.
  - Event logs have been integrated into case details, providing additional context on case events.


apps/judicial-system/web/src/routes/Shared/Cases/Cases.tsx
```diff
@@ -169,7 +172,18 @@ export const Cases: FC = () => {
...
+          const sentToPublicProsecutor = c.eventLogs?.some(
+            (log) =>
+              log.eventType === EventType.INDICTMENT_SENT_TO_PUBLIC_PROSECUTOR,
+          )
+          const isFineOrRuling =
+            c.indictmentRulingDecision === CaseIndictmentRulingDecision.FINE ||
+            c.indictmentRulingDecision === CaseIndictmentRulingDecision.RULING
+
+          return (
+            !isCompletedCase(c.state) ||
```
thorhildurt, line 184: One thought since we are checking also if the case is not completed yet based on the state. Should we not rather include this logic where we are setting if the case state as completed and then only checking `!isCompeltedCase(c.state)`? or would that result in unexpected tags?

thorhildurt (approved):
Only one thought about the location of the introduced logic

apps/judicial-system/web/src/routes/Shared/Cases/Cases.tsx
oddsson: Good point. I feel like we should maybe not do that, since we are checking for completed cases in soooo many other places and perhaps this condition doesn't apply there also.

Holla if you strongly disagree 👀 

Screenshot 2025-02-14 at 15 05 23

PR #17983, opened 2025-02-17
fix(j-s): Indictment completion date on confirmed case files

unakb:
[Asana](https://app.asana.com/0/1199153462262248/1209410676187746)

## What

Use indictment completion date instead of ruling date for the confirmation stamp on indictment case files.

## Why

Because the ruling date could be set to an earlier date but the completion always takes place when the user performs the complete action. 

## Checklist:

## Summary by CodeRabbit

- **New Features**
  - Introduced a court end time selection on the case summary page, allowing users to update and view precise completion dates.
  - Enhanced the indictments display by conditionally revealing merged case details when relevant.
  - Updated confirmed case PDFs to show more accurate ruling dates.

- **Documentation**
  - Added new descriptive text for court end time details on the summary screen.


apps/judicial-system/backend/src/app/modules/file/file.service.ts
```diff
@@ -179,12 +179,16 @@ export class FileService {
...
+    )
+
     return createConfirmedPdf(
       {
         actor: theCase.judge?.name ?? '',
         title: theCase.judge?.title,
         institution: theCase.judge?.institution?.name ?? '',
-        date: theCase.rulingDate,
+        date: completedEvent?.created || theCase.rulingDate,
       },
```
unakb, line 192: I decided to have it still fall back on the ruling date just in case, but that really shouldn't ever happen.

oddsson (approved):

PR #17985, opened 2025-02-17
fix(j-s): Hide ruling if case completed without ruling

unakb:
[Asana](https://app.asana.com/0/1199153462262248/1209412018641625)

## What

Disable ruling button for defenders if the case was completed without a ruling.
Don't include the ruling PDF in the downloadable folder of all files

## Why

If there is no ruling with the case we don't want to generate it as it's not going to be accurate and is not applicable. 

## Checklist:

## Summary by CodeRabbit

- **Refactor**
  - Updated document processing logic so that the ruling PDF is now added only when appropriate, ensuring correct document handling.

- **New Features**
  - Enhanced the case overview screen with conditional actions for the ruling button. A new localized message now appears to clearly indicate when a case is closed without a ruling.


thorhildurt (approved):
Tested and works well 👌

PR #17987, opened 2025-02-17
chore(j-s): Sanitise input before using in SQL queries

oddsson:
# Sanitise input before using in SQL queries

[Asana](https://app.asana.com/0/1199153462262248/1209412022978053)

## What

This PR fixes a bug when we are fetching connected cases for defendants in RVG. When we are fetching connected cases, we are querying by name in a `literal` sequalize query. If the name had a special character, f.x. an `'`, there would be an error causing a crash of the application. 

This has now been fixed by escaping the name and national id before querying. 

## Why

This is a bug and a SQL injection threat.

## Checklist:

### DIFF
--- a/apps/judicial-system/backend/src/app/modules/case/case.service.ts
+++ b/apps/judicial-system/backend/src/app/modules/case/case.service.ts
@@ -1696,13 +1696,16 @@ export class CaseService {
     ]
 
     if (defendant.noNationalId) {
+      const defendantName = defendant.name ?? ''
+      const defendantNationalId = defendant.nationalId ?? ''
+
       whereClause.push({
         id: {
           [Op.in]: Sequelize.literal(`
             (SELECT case_id
               FROM defendant
-              WHERE national_id = '${defendant.nationalId}'
-              AND name = '${defendant.name}'
+              WHERE national_id = ${this.sequelize.escape(defendantNationalId)}
+              AND name = ${this.sequelize.escape(defendantName)}
           )`),
         },
       })


apps/judicial-system/backend/src/app/modules/case/case.service.ts
```diff
@@ -1696,13 +1696,16 @@ export class CaseService {
...
+
       whereClause.push({
         id: {
           [Op.in]: Sequelize.literal(`
             (SELECT case_id
               FROM defendant
-              WHERE national_id = '${defendant.nationalId}'
-              AND name = '${defendant.name}'
+              WHERE national_id = ${this.sequelize.escape(defendantNationalId)}
+              AND name = ${this.sequelize.escape(defendantName)}
```
unakb, line 1708: So I did some reading and found it really hard to find anything about using escape to fix this issue, but I did find an alternative way to fix this query in the sequelize documentation that I think we should consider using instead 

`  getConnectedIndictmentCases(
    caseId: string,
    defendant: Defendant,
  ): Promise<Case[]> {
    return this.caseModel.findAll({
      include: [
        { model: Institution, as: 'court' },
        {
          model: Defendant,
          as: 'defendants',
          required: true,
          where: {
            national_id: defendant.noNationalId
              ? defendant.nationalId
              : { [Op.in]: normalizeAndFormatNationalId(defendant.nationalId) },
            ...(defendant.noNationalId && { name: defendant.name }),
          },
        },
      ],
      attributes: ['id', 'courtCaseNumber', 'type', 'state'],
      where: {
        isArchived: false,
        type: CaseType.INDICTMENT,
        id: { [Op.ne]: caseId },
        state: CaseState.RECEIVED,
      },
    })
  }`
  
  
Thoughts?

oddsson: I'm curious about why you think we should use this instead of using escape. The issue was caused by an un-escaped value in the query and that has been fixed by escaping that value.

I'm not against changing, just wanted to hear why you feel we should ☺️

oddsson: Thanks for the clarification. I agree with the article that this is not ideal since it's now a responsibility of the dev to remember to escape all user input in raw queries. I also feel however that because of that, we should not have any raw queries in our code at all. I'll change the code to use the query builder from Sequalize and add a ticket to Asana to remove all literal queries from the codebase 🚀

unakb: I think it's definitely worth looking in to doing that. In some cases I think there's no other way than using literal strings, like if a query is too complex to fit in to an ORM query pattern, but I admit I didn't look at any of the other ones we have so there very well might be a way to rewrite them as well 🤗. Also a good task to look in to when we go over speed issues (which we need to do soon)

unakb (approved):

PR #18009, opened 2025-02-18
fix(auth): Fix ids-api service startup due to lru cache dependency issue through path-scurry

eirikurn:
Same as was needed in [notification-service](https://github.com/island-is/island.is/pull/13395).

## Summary by CodeRabbit

- **Chores**
  - Enhanced our internal module integration to improve the build process. This update aims to boost overall stability and performance, contributing to a more reliable experience for end users.


GunnlaugurG (approved):
LGTM

PR #18030, opened 2025-02-19
fix(documents): Return bff link instead of direct dl service

thordurhhh:
## What

Return a BFF formatted link instead of a direct link to DL service from document service.

Perhaps we have another way to generate/access the client base url?

```
      CLIENT_BASE_URL: {
        dev: 'https://beta.dev01.devland.is',
        staging: 'https://beta.staging01.devland.is',
        prod: 'https://island.is',
      },
```      

## Why

The document service returns a direct link to the DL service today, we are handling that in the front end where we generate a bff url from the dl service link.

The APP needs a direct link to BFF formatted url to be able to link directly to the correct url for a document.
And then we can remove our bff format from the front end.

## Checklist:

### DIFF
--- a/apps/api/infra/api.ts
+++ b/apps/api/infra/api.ts
@@ -148,6 +148,11 @@ export const serviceSetup = (services: {
         dev: 'https://api.dev01.devland.is',
         staging: 'https://api.staging01.devland.is',
       },
+      CLIENT_BASE_URL: {
+        dev: 'https://beta.dev01.devland.is',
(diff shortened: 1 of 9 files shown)


thorkellmani (approved):
LGTM 🪨

eirikurn (changes requested):
The app does not have a BFF session cookie, so it can't use the BFF download URL. It needs to use its own token to download the file.

PR #18042, opened 2025-02-19
fix(native-app): minor health fixes

thoreyjona:
## What

* Make sure to show medicinePurchase level if percentage is 0
* When showing organ donation with limitation, show comment as well
* When loading organ donation info, make sure to add a skeleton to the label of the field as well since that has the information if the user is a donor or not, we don't want to possible show the wrong info while loading.

## Screenshots / Gifs
Loading state for organ donation at the bottom of the first screenshot and organ donation with limitation with a comment on the second
<div>
</div>

## Checklist:


eirikurn (approved):
LGTM

PR #18072, opened 2025-02-21
chore(j-s): add audit log when sending subpoena to police

thorhildurt:
# [Create audit log when delivering subpoena to police](https://app.asana.com/0/1199153462262248/1209317540798477)

## What
- Adding a new audit log when delivering subpoena to police
- We apply the auditTrailService wrapper on the `deliverSubpoenaToPolice`, and update the audit log functions accordingly to ensure we can log additional `details` properties. Note we have only used this service in the api layer, and we are applying it for the first time in backend
- For the `deliverSubpoenaToPolice` we additionally log:
    - `deliveredToPolice`
    - `subpoenaId`
    - `subpoenaCreated`
    - `policeSubpoenaId`
    - `subpoenaHash`
    - `subpoenaDeliveredToPolice`
    - `indictmentHash` 
- Note we leave out the optional fields for civil claimants, but that can be added if required

## Testing (optional)
- We need to test this on dev with a case which has `case.origin === LOKE`. A case on dev with case id `4537ad25-4dcb-4681-bd94-53f8fb5a2fa2` can be used to test this log. Simply issue a new 'fyrirkall' as a judge, then the log should appear. 
- Below is a screenshot when testing this against dev locally
Screenshot 2025-02-24 at 13 47 44

 
## Why
- To improve logs and confidence in events performed in the system for better traceability

## Checklist:

## Summary by CodeRabbit

- **New Features**
  - Enhanced subpoena delivery processes with integrated audit logging, ensuring increased traceability and accountability.
  - Introduced global configuration for advanced audit functionalities.
  - Added a new action for auditing: `DELIVER_SUBPOENA_TO_POLICE`.

- **Refactor**
  - Improved logging mechanisms with more structured input and asynchronous capabilities.
  - Optimized internal operations by removing unused components for cleaner, more efficient processing.


apps/judicial-system/backend/src/app/modules/subpoena/internalSubpoena.controller.ts
```diff
@@ -82,11 +87,33 @@ export class InternalSubpoenaController {
...
+        subpoenaCreated: subpoena.created,
+        policeSubpoenaId: currentSubpoena.subpoenaId,
+        subpoenaHash: currentSubpoena.hash,
+        subpoenaDeliveredToPolice: new Date(),
+        indictmentHash: theCase.indictmentHash,
+      }
+    }
+
+    return this.auditTrailService.audit(
+      'internalSubpoena',
```
unakb, line 107: I wonder if it makes more sense to have the user's name that triggered the sending here than "internalSubpoena"?

apps/judicial-system/backend/src/app/modules/subpoena/internalSubpoena.controller.ts
```diff
@@ -118,11 +145,16 @@ export class InternalSubpoenaController {
...
     )
 
-    return this.subpoenaService.deliverSubpoenaFileToPolice(
-      theCase,
-      defendant,
-      subpoena,
-      deliverDto.user,
+    return this.auditTrailService.audit(
+      'internalSubpoena',
+      AuditedAction.DELIVER_SUBPOENA_TO_POLICE,
```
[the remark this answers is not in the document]
thorhildurt: Completely agree! Ah I just forgot to remove the audit wrapper here 🙈 So that would be definitely confusing to send the same `AuditedAction` type for two different endpoints. I just started logging this one, but that didn't work because the `policeSubpoenaId` is never generated there, so I ended up using the above endpoint which provides us with all required details.

apps/judicial-system/backend/src/app/modules/subpoena/internalSubpoena.controller.ts
thorhildurt: Makes sense, will update the user 👍

unakb (approved):
Nice 👌

PR #18074, opened 2025-02-21
fix(native-app): share document fixes

thoreyjona:
## What
When sharing documents that opened a webview or were rendered as html then we were using an old outdated downloadUrl that did not work after the bff changes. 

Now when sharing a document that opens a webview we share the correct url. When sharing a file that is of type html we create a .html file with the html code and share that. Tested it both on android device and ios device.
**note:** When sharing the html file in an email it is sometimes rendered as a text file. That is just how the email client handles it. 

## Checklist:

## Summary by CodeRabbit

- **New Features**
	- Enhanced file sharing now offers context-aware support, allowing users to seamlessly share documents as PDF, HTML, or URL.
	- Improved cross-platform file operations with the integration of an advanced file system library, ensuring smoother performance on both Android and iOS devices.
	- New dependency added for file system access, expanding functionality for file operations within the app.
- **Bug Fixes**
	- Updated error logging format for passkey registration to improve clarity in console outputs.

### DIFF
--- a/apps/native/app/src/screens/document-detail/utils/share-file.tsx
+++ b/apps/native/app/src/screens/document-detail/utils/share-file.tsx
@@ -7,9 +7,14 @@ import { DocumentV2 } from '../../../graphql/types/schema'
 interface ShareFileProps {
   document: DocumentV2
   pdfUrl?: string
+  fallbackUrl?: string | null
 }
 
-export const shareFile = async ({ document, pdfUrl }: ShareFileProps) => {
+export const shareFile = async ({
+  document,
+  pdfUrl,
+  fallbackUrl,
+}: ShareFileProps) => {
   if (!document || !document.subject || !document.sender) {
     return
   }
@@ -24,7 +29,7 @@ export const shareFile = async ({ document, pdfUrl }: ShareFileProps) => {
       subject: document.subject,
       message: `${document.sender.name} \n ${document.subject}`,
       type: pdfUrl ? 'application/pdf' : undefined,
-      url: pdfUrl ? `file://${pdfUrl}` : document.downloadUrl!,
+      url: pdfUrl ? `file://${pdfUrl}` : fallbackUrl ?? document.downloadUrl!,
     })
   } catch (error) {
     // noop
--- a/apps/native/app/src/screens/document-detail/utils/share-file.tsx
+++ b/apps/native/app/src/screens/document-detail/utils/share-file.tsx
@@ -1,4 +1,5 @@
 import Share from 'react-native-share'
+import RNFS from 'react-native-fs'
 
 import { isAndroid } from '../../../utils/devices'
 import { authStore } from '../../../stores/auth-store'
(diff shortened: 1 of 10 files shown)


apps/native/app/src/screens/document-detail/utils/share-file.tsx
```diff
@@ -1,30 +1,56 @@
...
 
+  if (isHtml) {
+    const utf8Prefix = '<meta charset="UTF-8">'
+    const formattedHtml = utf8Prefix + content
+    try {
+      const path = `${RNFS.TemporaryDirectoryPath}/${document.subject}.html`
+      await RNFS.writeFile(path, formattedHtml, 'utf8')
+      htmlUrl = path
+    } catch (error) {
+      // noop
```
thoreyjona, line 39: I'm thinking if we should add error message if this or the sharing fails? Just to let the user know that it did not work. Toast (when we have added that globally) for example

apps/native/app/src/screens/document-detail/utils/share-file.tsx
```diff
@@ -1,30 +1,56 @@
...
       subject: document.subject,
       message: `${document.sender.name} \n ${document.subject}`,
-      type: pdfUrl ? 'application/pdf' : undefined,
-      url: pdfUrl ? `file://${pdfUrl}` : document.downloadUrl!,
+      type: pdfUrl ? 'application/pdf' : isHtml ? 'text/html' : undefined,
+      url: pdfUrl
+        ? `file://${pdfUrl}`
+        : isHtml
+        ? `file://${htmlUrl}`
+        : content ?? document.downloadUrl!,
```
eirikurn, line 53: We should never use `document.downloadUrl`.

apps/native/app/src/screens/document-detail/utils/share-file.tsx
eirikurn: Yeah. In the meantime we may want to log the error in datadog and return?

apps/native/app/src/screens/document-detail/utils/share-file.tsx
```diff
@@ -1,30 +1,56 @@
...
       title: document.subject,
       subject: document.subject,
       message: `${document.sender.name} \n ${document.subject}`,
-      type: pdfUrl ? 'application/pdf' : undefined,
-      url: pdfUrl ? `file://${pdfUrl}` : document.downloadUrl!,
+      type: pdfUrl ? 'application/pdf' : isHtml ? 'text/html' : undefined,
+      url: pdfUrl
+        ? `file://${pdfUrl}`
+        : isHtml
+        ? `file://${htmlUrl}`
```
eirikurn, line 52: ~I wonder if we can just use a data URL instead of saving these as files?

apps/native/app/src/screens/document-detail/utils/share-file.tsx
```diff
@@ -1,30 +1,56 @@
...
 
   if (isAndroid) {
     authStore.setState({ noLockScreenUntilNextAppStateActive: true })
   }
 
+  if (isHtml) {
+    const utf8Prefix = '<meta charset="UTF-8">'
+    const formattedHtml = utf8Prefix + content
+    try {
+      const path = `${RNFS.TemporaryDirectoryPath}/${document.subject}.html`
```
eirikurn, line 35: Should encode the subject somehow to avoid invalid filename characters?

eirikurn (commented):
👍

apps/native/app/src/screens/document-detail/utils/share-file.tsx
thoreyjona: Good idea! But I don't think we are logging errors in datadog in any place in the app 😅 At least I have never done it, need to look into it

PR #18075, opened 2025-02-21
feat(application-system): Feedback for uploaded file with malware tag from GuardDuty

HjorturJ:
# Give users feedback when they upload malicious files.

## What

Let users know when they upload a file that is flagged by GuardDuty as malware.

## Why

So that we can start quarantining infected files in the back-end without our users not understanding why their file wasnt accessible by the application reviewers.

Note: This feature adds a minimum of 1 second to all file uploads, as after the file has been uploaded, we wait 1 second then fetch the malware tags from AWS.
For larger files there are also longer and longer re-fetch delays to give GuardDuty more time to scan the files.
At the time of the PR, even 100mb files have always been scanned after the second attempt (3 total seconds of wait)

After detection
image

## Checklist:

## Summary by CodeRabbit

- **New Features**
  - Introduced integrated file malware scanning during uploads, providing real-time status updates for enhanced safety.
  - Improved upload progress tracking with a refined indicator that remains animated until completion.
  - Added clearer error messaging to notify users when files are flagged for potential malware.


libs/application/ui-components/src/utilities/FileUploadController/index.tsx
```diff
@@ -206,27 +246,42 @@ export const FileUploadController = ({
...
             key: res.key,
+            url: res.url
           },
         })
       } catch {
         setUploadError(formatMessage(coreErrorMessages.fileUpload))
       }
     })
+
+    await Promise.all(uploadPromises)
```
norda-gunni, line 272:
```suggestion
    await Promise.allSettled(uploadPromises)
```
Consider using `Promise.allSettled()` here because `Promise.all()` fails fast and returns with the first failure but allSettled returns, as you may suspect, when all the promises have been settled.

norda-gunni (commented):
One tiny question

libs/application/ui-components/src/utilities/FileUploadController/index.tsx
HjorturJ: ah, did not know this, perfect

norda-gunni (approved):
LGTM

libs/api/domains/file-upload/src/lib/file-upload.resolver.ts
```diff
@@ -1,16 +1,25 @@
...
 @Resolver()
 export class FileUploadResolver {
   constructor(private fileStorageService: FileStorageService) {}
 
+  @Query(() => [TagType])
+  async getAttachmentTags(
+    @Args('url') url: string,
+  ): Promise<Tag[]> {
+    return this.fileStorageService.getAttachmentTags(url)
+  }
```
eirikurn, line 17: This resolver is not application/attachment specific. The upload bucket and resolver are shared for multiple use cases. Let's rename this to `fileUploadTags` instead.

Also, it should accept an upload key, or validate that the `url` points to our upload bucket, so this resolver can't be used to lookup tags of files in other buckets.

libs/api/domains/file-upload/src/lib/getAttachmentTags.model.ts
```diff
@@ -0,0 +1,10 @@
+import { Field, ObjectType } from '@nestjs/graphql'
+
+@ObjectType()
+export class TagType {
+  @Field(() => String)
+  Key!: string
```
eirikurn, line 6: This does not match our [graphql naming convention](https://docs.devland.is/technical-overview/api-design-guide/graphql-naming-conventions). Please map the tags to use camelCase fields and namespace the object type (eg `FileUploadTag`).

libs/nest/aws/src/lib/s3.service.ts
```diff
@@ -158,6 +160,28 @@ export class S3Service {
...
+      const results = await this.s3Client.send(command)
+      
+      return results?.TagSet ?? []
+    } catch (error) {
+      this.logger.error(
+        `Error occurred while fetching file tags for key: ${key} in S3 bucket: ${bucket}`,
+        error,
+      )
+    }
+    return []
```
eirikurn, line 182: Should almost never swallow unexpected errors. Errors from user input should only be logged as warnings (eg bad url?). 

Should looking up tags for a file which does not exist return an empty array or throw an error? I prefer throwing an error so the client knows that something happened, instead of assuming the file has no tags.

eirikurn (changes requested):
Ah, this will be very nice to have!

libs/api/domains/file-upload/src/lib/file-upload.resolver.ts
HjorturJ: Gotcha, if I understood correctly then I think I've made all the changes requested :)

libs/api/domains/file-upload/src/lib/getAttachmentTags.model.ts
HjorturJ: Done.

libs/nest/aws/src/lib/s3.service.ts
HjorturJ: Ah good point, getting ahead of myself a bit here 😅 Fixed.

eirikurn (approved):

libs/api/domains/file-upload/src/lib/file-upload.resolver.ts
HjorturJ: Ah ok interesting, thanks for the explaination. I'm going to try and go for a refactor to use the malwareScanStatus way 😄

HjorturJ: Ok, I've made more changes, think this is better at least 😅 Let me know if you can think of any more improvements 😄

libs/api/domains/file-upload/src/lib/file-upload.resolver.ts
```diff
@@ -1,12 +1,47 @@
...
   constructor(private fileStorageService: FileStorageService) {}
 
+  @Query(() => MalwareScanStatus)
+  async malwareScanStatus(
+    @Args('filename') filename: string,
+  ): Promise<MalwareScanStatus> {
+    let scanStatus = undefined
+    for (let i = 0; i < MALWARE_FETCH_ATTEMPTS; i++) {
+      const waitTimeThisLoop = MALWARE_MIN_MS_BETWEEN_FETCHES * (Math.pow(2, i))
+      await new Promise((resolve) => setTimeout(resolve, waitTimeThisLoop))
```
eirikurn, line 21: ~Waiting immediately assumes that this resolver is always called right after upload finished. Good to check tags once before doing backoff retries for better performance when the scan is already done.

libs/api/domains/file-upload/src/lib/malwareScanStatus.model.ts
```diff
@@ -0,0 +1,9 @@
+import { registerEnumType } from '@nestjs/graphql'
+import { MalwareScanStatus } from '@island.is/application/types'
```
eirikurn, line 2: ~Can we refactor this enum someplace else? E.g. shared/types or shared/constants.

eirikurn (approved):
Nice refactor 🙌

libs/api/domains/file-upload/src/lib/malwareScanStatus.model.ts
HjorturJ: Sure, makes sense, done.

eirikurn (approved):

PR #18078, opened 2025-02-21
feat(native-app): vehicle fetching on home screen should be cache-first

thoreyjona:
## What
Vehicle fetching on home screen should be cache-first.

## Why
When releasing barcodes we could get a lot of more traffic to the app and we don't want to overload samgöngustofa with requests if not necessary.

## Checklist:

## Summary by CodeRabbit

- **New Features**
  - Improved speed and responsiveness in the vehicle display. This update optimizes data retrieval by prioritizing local cached information, leading to faster load times and smoother interactions when viewing vehicle details. Users should experience a more efficient and responsive interface overall.
  - Increased the number of vehicle items displayed per page from 10 to 15, allowing users to view more vehicles at once.

### DIFF
--- a/apps/native/app/src/screens/home/home.tsx
+++ b/apps/native/app/src/screens/home/home.tsx
@@ -181,6 +181,7 @@ export const MainHomeScreen: NavigationFunctionComponent = ({
         pageSize: 15,
       },
     },
+    fetchPolicy: 'cache-first',
     skip: !vehiclesWidgetEnabled,
   })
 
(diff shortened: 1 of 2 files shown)


eirikurn (approved):
Need to somehow test that this gets updated from the subpage. Is it using the exact same query?

thoreyjona:
@eirikurn we are fetching with`pageSize: 10` on the homescreen but `pageSize: 15` on the details page. I think  > 99% of users don't have more than 10 vehicles so I think it would be safe to have it as 10 for both places. I think it could be cached separately if we don't have the exact same variables but not 100% sure.

PR #18086, opened 2025-02-21
fix(j-s): Police case numbers input

unakb:
[Asana](https://app.asana.com/0/1199153462262248/1209457126282469)

## What

- Prevent users from being able to input the same police case number twice.
- Hide close icon on the first police case number if the case's origin was the police case system.

## Why

There shouldn't ever be any need to have the same police case number twice in the same case. It also leads to buggy behaviour.

## Checklist:

## Summary by CodeRabbit

- **New Features**
	- Enhanced validations prevent the removal of a crucial police number during case updates.
	- Improved error messaging and input controls now guide users when managing police numbers.
- **Refactor**
	- Improved code organization and readability through the introduction of helper functions.


apps/judicial-system/backend/src/app/modules/case/case.controller.ts
```diff
@@ -271,6 +272,15 @@ export class CaseController {
       )
     }
 
+    if (theCase.origin === CaseOrigin.LOKE && update.policeCaseNumbers) {
+      const mainPoliceCaseNumber = theCase.policeCaseNumbers[0]
+      if (!update.policeCaseNumbers.includes(mainPoliceCaseNumber)) {
```
gudjong, line 277: This may cause issues in later updates as the main police case number can be removed in a later update if it is moved to a position other than 0. Please make sure that the main (first) police case number is alse the main (first) police case number in the updated list.

gudjong (changes requested):
One change request in the backend logic.

apps/judicial-system/backend/src/app/modules/case/case.controller.ts
unakb: Ah, that's a good point I hadn't thought of. Will add inn condition to handle that 🤝

gudjong (approved):

PR #18116, opened 2025-02-25
fix(web): Change asset link extension tag

mannipje:
# Change asset link extension tag

## What

- Handle asset link extension tag when it is video and does not have dot extension in url
- Do not render extension tag if it is to long

## Why

- Because if file has no dot extension in the end then the tag text was far too long and had affect the site style

### Before
image

### After
image

## Checklist:

### DIFF
--- a/libs/island-ui/contentful/src/lib/AssetLink/AssetLink.tsx
+++ b/libs/island-ui/contentful/src/lib/AssetLink/AssetLink.tsx
@@ -11,14 +11,19 @@ export const AssetLink: FC<React.PropsWithChildren<AssetLinkProps>> = ({
   url,
   children,
 }) => {
+  let extension: string
   const parts = url.split('.')
-  const extension = parts[parts.length - 1].toUpperCase()
+  extension = parts[parts.length - 1].toUpperCase()
+
+  if (url.includes('videos.ctfassets') && extension.length >= 10) {
+    extension = 'VIDEO'
+  }
 
   const secureUrl = `${url.startsWith('//') ? 'https:' : ''}${url}`
 
   return (
     <FocusableBox href={secureUrl} border="standard" borderRadius="large">
-      <LinkCard background="white" tag={extension}>
+      <LinkCard background="white" tag={extension.length < 10 ? extension : ''}>
         {title || children}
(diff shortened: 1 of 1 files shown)


libs/island-ui/contentful/src/lib/AssetLink/AssetLink.tsx
```diff
@@ -11,14 +11,19 @@ export const AssetLink: FC<React.PropsWithChildren<AssetLinkProps>> = ({
   url,
   children,
 }) => {
+  let extension: string
   const parts = url.split('.')
-  const extension = parts[parts.length - 1].toUpperCase()
+  extension = parts[parts.length - 1].toUpperCase()
+
+  if (url.includes('videos.ctfassets') && extension.length >= 10) {
```
RunarVestmann, line 18: Do we need the length check here?

libs/island-ui/contentful/src/lib/AssetLink/AssetLink.tsx
```diff
@@ -11,14 +11,19 @@ export const AssetLink: FC<React.PropsWithChildren<AssetLinkProps>> = ({
   url,
   children,
 }) => {
+  let extension: string
```
RunarVestmann, line 14: I myself am not a huge fan of uninitialised variables and would prefer it if in line 16 you would define the variable inline.

libs/island-ui/contentful/src/lib/AssetLink/AssetLink.tsx
mannipje: Maybe, if the video file has like .mp4 file ending then the tag will be MP4. Or maybe we just want the tag to be VIDEO then we dont need it

RunarVestmann: Ah, that's why the length check is there. I see, then that's fine

RunarVestmann: Maybe it'd be more obvious if the conditional below would be moved up as well

RunarVestmann (approved):

PR #18120, opened 2025-02-25
chore: Fix redis cluster port conflicts in local development

eirikurn:
## What

- Use a shared redis-cluster docker container across different services.
- Change the published redis ports from 7000-7005 to 7010-1015 for local development.
- Use docker compose name instead of directory name to manage services.

## Why

- To allow devs to start multiple services on their workstation which use redis and caused conflicts previously. These services have a shared redis cluster in production and are using it to share cache. Replicating that is easier and better then running multiple redis-clusters in docker across non-conflicting ports.
- Also, port 7000 is now taken by default on Macs by the Airplay service. Moving the redis port to reduce setup steps and not force devs to disable the Airplay service.

## FYI

- Changing the names in docker means that new containers will be created with clean data. You may have to run DB migrations, move data between DBs and clean up old containers.
- Specifying the name in docker-compose.yml is a fairly recent docker-compose feature, so you might need to update docker-compose (or Docker for Mac if you are using that).

## Summary by CodeRabbit

- **Chores**
	- Updated connection settings across services to shift the host port range for Redis, ensuring consistent integration across all environments.
	- Adjusted development configurations to reflect the new connection endpoints.

- **Documentation**
	- Removed outdated troubleshooting guidance regarding port conflicts on certain operating systems.


lodmfjord (approved):

norda-gunni (approved):
LGTM

kksteini (approved):
codeowner files lgtm

magnearun (approved):

PR #18174, opened 2025-02-28
feat(j-s): Redirect with User Id

gudjong:
# Redirect with User Id

[Styðja user id í login og deep links](https://app.asana.com/0/1199153462262248/1209524374225368)

## What

- Supports redirects with user id. This makes it possible to create deep links that are targeted to a specific user and is helpful when national ids are connected to multiple users.
- Supports developer login in web client login redirects due to token expiration. This is usedul when testing as the developer login remains in effect even when re-login is necessary.
- Sanity checks redirects to make sure they match the activated user.

## Why

- Allowing the same national id to be connected with multiple users introduces problems with the existing deep links and web client login redirects due to token expiration. This implementation addresses those problems.

### Login with two users: http://localhost:4200/api/auth/login?nationalId=1111110001
https://github.com/user-attachments/assets/a8e75cf8-17b4-4250-a8bf-61f02c59b009

### Login with two users and a redirect: http://localhost:4200/api/auth/login?nationalId=1111110001&redirectRoute=/landsrettur/yfirlit/65684d47-8ff1-45a0-9b8c-95d2ffc3c4a5
https://github.com/user-attachments/assets/f98895ec-3936-4399-bb4c-6aa6d17b566c

### Login with two users and a user id: http://localhost:4200/api/auth/login/168c6882-f71f-4003-8486-169fe2458a34?nationalId=1111110001
https://github.com/user-attachments/assets/0baf1c9f-d3fc-45c8-b11b-a39c929c7f06

### Login with two users, a user id and a redirect: http://localhost:4200/api/auth/login/168c6882-f71f-4003-8486-169fe2458a34?nationalId=1111110001&redirectRoute=/landsrettur/yfirlit/65684d47-8ff1-45a0-9b8c-95d2ffc3c4a5
https://github.com/user-attachments/assets/902b5377-d960-4e59-900a-ba3c1824c2ba

### Auto relogin after auth cookie timeout
https://github.com/user-attachments/assets/7e81bef9-ce76-4b3c-8d73-da2a850e8e18

## Checklist:


apps/judicial-system/api/src/app/modules/auth/auth.controller.ts
```diff
@@ -424,34 +456,105 @@ export class AuthController {
...
+        allowedPrefixes.some((prefix) =>
+          requestedRedirectRoute.startsWith(prefix),
+        )
+        ? requestedRedirectRoute
+        : defaultRoute
+    }
+
+    if (isProsecutionUser(currentUser)) {
+      return getRedirectRoute(CASES_ROUTE, [
+        '/beinir',
```
oddsson, line 477: Consider using the routes from the lib, like `CASES_ROUTE`

oddsson (approved):

apps/judicial-system/api/src/app/modules/auth/auth.controller.ts
gudjong: These are route prefixes and as such they do not exists in the lib. However, adding such constants in the lib is a good idea. I would like to clean up all our routes in the future and I think I will leave this as is until then.

PR #18180, opened 2025-02-28
ci: Speed up `install.yml` workflow

AndesKrrrrrrrrrrr:
- Use new packing/installing with corepack
- Use `$HOME` instead of `/home/runner`


AndesKrrrrrrrrrrr:
Possibly outdated. Closing for now.

PR #18202, opened 2025-03-03
feat(app-sys): copy link shared component

jonnigs:
## What

Make a new shared component `buildCopyLinkField`

## Why

This component is used in a few places in the application system and is always custom

## Screenshots / Gifs

Screenshot 2025-03-03 at 16 31 02

## Checklist:

## Summary by CodeRabbit

- **New Features**
	- Introduced a new form field that enables users to easily copy links via a dedicated button with accessible labels and descriptions.
	- Integrated the copy link functionality into forms, offering options for both static and dynamic URL sources for enhanced user interaction.


libs/application/core/src/lib/fieldBuilders.ts
```diff
@@ -1117,3 +1118,18 @@ export const buildOverviewField = (
     children: undefined,
   }
 }
+
+export const buildCopyLinkField = (
```
norda-gunni, line 1122: would `copyableLinkField` be more understandable?

norda-gunni (commented):
One question

norda-gunni (approved):
That one question is totally minor so I'll approve anyway

libs/application/core/src/lib/fieldBuilders.ts
jonnigs: Everything else has the "build" prefix so I say we keep it in line with the rest :)

PR #18226, opened 2025-03-04
fix(j-s): Poll service status until served successfully

oddsson:
# Poll service status until served successfully

[Asana](https://app.asana.com/1/203394141643832/project/1199153462262248/task/1209431560357533)

## What

This PR brings a change in how we present information about the service status of an indictment. Before this change, we used to call into RLS to get the service status if there was no service status set on the case. Once the service status was set on a case, we skipped calling into RLS and just displayed what status we had.

This could lead to issues, f.x. if a case's service status is _failed_, because the server couldn't reach the defendant but then the defendant is served electronically. Then we would never show the service status as served because the case was already marked with a failed service status.

This PR changes this so that we don't stop calling RLS for updates until the case is marked with a successful service status.

## Screenshots / Gifs

https://github.com/user-attachments/assets/d7289a00-b07d-44f4-a040-0c621566adb4

## Checklist:

## Summary by CodeRabbit

- **Bug Fixes**
	- Refined the logic for handling subpoena-related requests to ensure actions only occur under the correct conditions, resulting in improved reliability and more accurate data processing.


thorhildurt (approved):

thorhildurt (approved):

unakb (approved):
Looks good :)

PR #18243, opened 2025-03-05
fix(j-s): clean-up traffic violation indictment count fields when a indictment subtype is removed

thorhildurt:
# [Bug fixes when removing indictment subtypes from a police case](https://app.asana.com/1/203394141643832/project/1199153462262248/task/1209363863185091)

## What
In this PR we perform a cascading clean-up when a specific subtype is removed on the police case overview
Screenshot 2025-03-06 at 14 46 48. 
A decision was made to try to start moving indictment count logic to the **backend** and avoid adding more complexity to the` Defendants.tsx`, `indictment.tsx` and `indictmentCount.tsx `components. 

### Key changes
- (1) Handle a cascading clean-up on the backend, where we compare current vs updated indictment count subtypes on a given indictment count when a police case subtype is removed in the `PoliceCaseInfo.tsx`. 
     - We do this to avoid adding more logic in the UI components, and rather have the backend handle the required updates. We could definitely explore better future options how we store and structure the data such that we wouldn't need to reset the fields this way, but I think this is ok for now.  
 - (2) Ensure that previous indictment count subtype selection is persisted when subtype changes are made 

## Why
- Because this is a bug and causes unexpected behaviour

## Test scenarios
- [x] (1) Add traffic violation subtype and fill out the indictment count fields. (2) Go back to the police case overview and remove the traffic violation subtype and add some other type. (3) Navigate back to the indictment count and all fields should be based on the new subtype
- [x] (1) Add 3 subtypes where one is traffic violation. (2) For the indictment count select all. (3) Go back to the police case overview and remove one subtype. (4) Go back to the indictment count, and the subtype selection should be maintained. 

## Screenshots / Gifs

https://github.com/user-attachments/assets/bba557ed-882e-4f9f-8979-ce384c542213

https://github.com/user-attachments/assets/67f85952-1af2-41ad-8a7f-dd9cee6e3f4c

## Checklist:

## Summary by CodeRabbit

- **New Features**
  - Enabled more detailed input options for updating case subtypes, allowing multiple selections for police case identifiers.
  - Improved handling of traffic violation indicators to ensure precise case updates.

- **Refactor**
  - Streamlined the filtering and update processes for subtype selections, resulting in a more consistent and clear user experience.

- **Tests**
  - Enhanced testing procedures to simulate transactional updates, ensuring robust and reliable operations.


apps/judicial-system/web/src/routes/Prosecutor/Indictments/Indictment/Indictment.tsx
```diff
@@ -282,9 +281,15 @@ const Indictment = () => {
...
   const initialize = useCallback(() => {
-    if (workingCase.indictmentCounts?.length === 0) {
+    const indictmentCounts = workingCase.indictmentCounts || []
+    if (indictmentCounts.length === 0) {
       handleCreateIndictmentCount()
     }
+    else {
+      console.log("initialise indictment")
+      // in case indictment subtypes have been modified in earlier step
+      setDriversLicenseSuspensionRequest(indictmentCounts)
```
thorhildurt, line 291: We also need to update the suspension request fields if the indictment count have changed due to earlier steps...

apps/judicial-system/backend/src/app/modules/indictment-count/indictmentCount.service.ts
```diff
@@ -38,16 +41,62 @@ export class IndictmentCountService {
...
-      : this.indictmentCountModel.update(update, {
-          where: { id: indictmentCountId, caseId },
-          returning: true,
+      }
+
+      const isTrafficViolationSubtypePresent =
+        hasTrafficViolationSubtype(updatedSubtypes) ||
+        (policeCaseNumberSubtypes.length === 1 &&
+          hasTrafficViolationSubtype(policeCaseNumberSubtypes))
+
```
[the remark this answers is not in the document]
thorhildurt: That is intentional...

apps/judicial-system/backend/src/app/modules/indictment-count/indictmentCount.service.ts
```diff
@@ -38,16 +41,62 @@ export class IndictmentCountService {
...
+      }
+
+      return this.indictmentCountModel.update(updatedIndictmentCount, {
+        where: { id: indictmentCountId, caseId },
+        returning: true,
+        transaction,
+      })
+    }
+
+    const updateWithInternalTransaction = () =>
```
unakb, line 92: I found this a little confusing at first and wondered whether it might be cleaner to have the updateWithTransaction take in an optional transaction and then handle it there if it's undefined, but then that might just make that function more messy so I can see why you wrote it this way. 🤗

unakb (approved):
Looks good!

apps/judicial-system/backend/src/app/modules/indictment-count/indictmentCount.service.ts
thorhildurt: Yes I get that, was thinking a bit how to handle the added transaction wrapper because we didn't need that one specifically before. Thought about having it all in one function and handle the "external" transaction property but found this to be a bit simpler, to have `updateWithTransaction` shared between "internal" transaction and "external transaction.

PR #18256, opened 2025-03-06
fix(contentful-entry-tagger): Fix prod url

RunarVestmann:
# Fix prod url

Since devland.is is out we need to change the prod url


busla (approved):
LGTM

PR #18282, opened 2025-03-07
fix(native-app): don't show see all button for finished applications if they 3 or less

thoreyjona:
## What

Currently we are always showing the 'See all' button for completed applications. Strange if you only have one completed application to be able to open up a whole new screen with that one application. 
Removing the see all button if completed applications are fewer than 4.

## Checklist:

## Summary by CodeRabbit

- **Refactor**
  - Simplified the conditions for enabling interactive elements and displaying controls based on the available items, resulting in a more straightforward and predictable user interface.

### DIFF
--- a/apps/native/app/src/screens/applications/components/applications-preview.tsx
+++ b/apps/native/app/src/screens/applications/components/applications-preview.tsx
@@ -1,3 +1,4 @@
+import React from 'react'
 import { useIntl } from 'react-intl'
 import { TouchableOpacity, View } from 'react-native'
 import { useTheme } from 'styled-components'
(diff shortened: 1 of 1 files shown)


apps/native/app/src/screens/applications/components/applications-preview.tsx
```diff
@@ -1,3 +1,4 @@
+import React from 'react'
```
eirikurn, line 1: ~Unneccessary?

eirikurn (approved):

apps/native/app/src/screens/applications/components/applications-preview.tsx
thoreyjona: My editor complains over using <> </> unless I import React 🤷‍♀️

PR #18353, opened 2025-03-12
feat: migrate staging to argocd (#18321)

AndesKrrrrrrrrrrr:
Reverts island-is/island.is#18352
Re-applies island-is/island.is#18321


brynjarorng (approved):
LGTM

PR #18368, opened 2025-03-13
fix(secondary school): Add alertMessage warning if selected special needs program

johannaagma:
## What

Add alertMessage warning if selected1 or more special needs program

## Why

To prevent users from accidentally selecting special needs programs

## Screenshots / Gifs

image

## Checklist:

## Summary by CodeRabbit

- **New Features**
  - Added a warning alert to the secondary school form that notifies users when a special needs program is selected.
  - Enhanced the selection process to indicate programs for special needs with tailored, localized alert messages.
  - Updated alert messaging for additional selections with revised titles and descriptions.

- **Refactor**
  - Improved the dynamic message formatting to integrate current form responses more accurately for localized displays.

### DIFF
--- a/libs/application/templates/secondary-school/src/forms/secondarySchoolForm/schoolSection/utils.ts
+++ b/libs/application/templates/secondary-school/src/forms/secondarySchoolForm/schoolSection/utils.ts
@@ -484,11 +484,9 @@ export const getAlertSpecialNeedsProgramMessage = (
     ) || []
 
   return {
-    ...(programNames.length === 1
-      ? school.selection.specialNeedsProgramSingleAlertDescription
-      : school.selection.specialNeedsProgramMultipleAlertDescription),
+    ...school.selection.specialNeedsProgramAlertDescription,
     values: {
-      programName: programNames.join(', '),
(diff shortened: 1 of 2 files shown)


sigruntg (approved):

libs/application/types/src/lib/Fields.ts
```diff
@@ -569,6 +569,7 @@ export interface AlertMessageField extends BaseField {
   alertType?: 'default' | 'warning' | 'error' | 'info' | 'success'
   message?: FormTextWithLocale
   links?: AlertMessageLink[]
+  watchValue?: string
```
jonnigs, line 572: I think this new attribute watchValue is a bit redundant since it's not adding any extra functionality from the condition.

I think the problem that you are trying to solve is to get access to an updated answers object to display the correct message. This should be possible with getValues from useFormContext(), example in the select form field

jonnigs: Sounds ok to me, if the user doesn't notice a flicker or anything

jonnigs (approved):

PR #18404, opened 2025-03-17
feat(license): Add photo to license metadata

eirikurn:
## Summary by CodeRabbit

- **New Features**
  - Expanded license records to include an optional photo. This update now supports displaying a license holder’s image (stored as a base64-encoded PNG) for driving, firearm, and hunting licenses.
  - Introduced a new method for fetching details about the other parent associated with parental leave applications.
  - Added a new structure for representing information related to another parent in parental leave applications.


libs/api/domains/license-service/src/lib/mappers/huntingLicenseMapper.ts
```diff
@@ -145,6 +145,7 @@ export class HuntingLicensePayloadMapper implements GenericLicenseMapper {
               description: [
                 { text: formatMessage(m.huntingLicenseDescription) },
               ],
+              photo: t.holderPhoto,
```
thoreyjona, line 148: Why not` ?? undefined` here like in other places?

thoreyjona (approved):
LGTM

libs/api/domains/license-service/src/lib/mappers/huntingLicenseMapper.ts
eirikurn: I did it in the other place to coorce nulls into undefined, but the types match here. Don't think it matters at all.

PR #18410, opened 2025-03-18
feat(j-s): Create Case from LÖKE

gudjong:
# Create Case from LÖKE

[Bæta kennitölu embættis við API fyrir stofnun máls úr LÖKE](https://app.asana.com/0/1199153462262248/1209626188347722/f)

## What

- Accepts an optional prosecutors office national id when creating a case from LÖKE.

## Why

- To be able to select the correct prosecutor in RVG.

## Checklist:

## Summary by CodeRabbit

- **New Features**
  - Enhanced the case creation process with stricter validations: a prosecutor identifier is now required, and an optional field for the prosecutor's office identifier has been added to improve data accuracy.
  
- **Bug Fixes**
  - Improved case creator identification logic to ensure only valid users are selected based on their institution's national ID.

- **Documentation**
  - Updated API documentation to reflect the revised validation rules and the inclusion of the optional office identifier field.


apps/judicial-system/backend/src/app/modules/case/dto/internalCreateCase.dto.ts
```diff
@@ -38,6 +38,14 @@ export class InternalCreateCaseDto {
   @ApiProperty({ type: String })
   readonly prosecutorNationalId!: string
 
+  @IsOptional()
+  @IsString()
+  @MinLength(10)
```
oddsson, line 43: Nit: Consider using `Length` instead for a bit cleaner code

Screenshot 2025-03-19 at 13 06 20

apps/judicial-system/xrd-api/jest.config.ts
```diff
@@ -8,6 +8,8 @@ export default {
   coverageDirectory: '<rootDir>/coverage/apps/judicial-system/xrd-api',
   displayName: 'judicial-system-xrd-api',
   testEnvironment: 'node',
+  detectOpenHandles: false,
+  detectLeaks: false,
```
oddsson, line 12: Is this debug code or meant for production?

oddsson (approved):

apps/judicial-system/xrd-api/jest.config.ts
gudjong: Meant to be like this. We want all our tests suits to have this config because otherwise we cannot run tests. Note that is is not production config, it is jest test config.

apps/judicial-system/backend/src/app/modules/case/dto/internalCreateCase.dto.ts
gudjong: Thanks, I wasn't aware of this class validator.

PR #18520, opened 2025-03-24
chore(ci): Rename and tag workflow steps and jobs

AndesKrrrrrrrrrrr:


AndesKrrrrrrrrrrr:
Tagging like in this PR could be valuable later. However, since the job name still includes the projects it's testing, this PR won't solve the issue of job name sprawl and getting metrics of top-N jobs by <x>.

PR #18525, opened 2025-03-24
chore: dedupe

AndesKrrrrrrrrrrr:
```
❯ yarn dedupe
➤ YN0000: ┌ Deduplication step
➤ YN0000: │ No packages can be deduped using the highest strategy
➤ YN0000: └ Completed in 0s 257ms
➤ YN0000: · Yarn 4.6.0
➤ YN0000: ┌ Project validation
➤ YN0057: │ @island.is/native-app: Resolutions field will be ignored
➤ YN0000: └ Completed
➤ YN0000: ┌ Resolution step
➤ YN0000: └ Completed in 0s 478ms
➤ YN0000: ┌ Post-resolution validation
➤ YN0086: │ Some peer dependencies are incorrectly met by dependencies; run yarn explain peer-requirements for details.
➤ YN0000: └ Completed
➤ YN0000: ┌ Fetch step
➤ YN0013: │ No new packages added to the project, but 798 were removed (- 122.33 MiB).
```

## Summary by CodeRabbit

- **Style**
	- Reinstated a layout constant to ensure consistent responsive design in sidebar and content areas.
- **Chores**
	- Restored internationalization support by re-adding the necessary message imports, ensuring localized text is properly displayed on affected pages.

PR #18526, opened 2025-03-24
fix(auth-nest-tools): Handle invalid token inputs

AndesKrrrrrrrrrrr:
Following https://github.com/island-is/island.is/pull/18525, I got errors from this library. Typing indicates that the code is invalid in the current form, hopefully fixing here :smile:

### DIFF
--- a/libs/auth-nest-tools/src/lib/create-key-provider.ts
+++ b/libs/auth-nest-tools/src/lib/create-key-provider.ts
@@ -87,7 +87,15 @@ const createMultiIssuerKeyProvider = ({
       return callback(null)
     }
 
-    const issuer = decodedJwtToken.payload.iss
+    const payload = decodedJwtToken.payload
+    if (typeof payload === 'string')
+      throw new Error(
+        "Token payload is a plain string, don't know what to do 🤷",
+      )
+    if (payload.iss === undefined)
+      throw new Error('Issuer field is undefined 😠')
+
+    const issuer = payload.iss
     const client = clients.get(issuer)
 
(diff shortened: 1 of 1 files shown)


libs/auth-nest-tools/src/lib/create-key-provider.ts
```diff
@@ -87,7 +87,15 @@ const createMultiIssuerKeyProvider = ({
...
     }
 
-    const issuer = decodedJwtToken.payload.iss
+    const payload = decodedJwtToken.payload
+    if (typeof payload === 'string')
+      throw new Error(
+        "Token payload is a plain string, don't know what to do 🤷",
+      )
+    if (payload.iss === undefined)
+      throw new Error('Issuer field is undefined 😠')
```
eirikurn, line 96: We always wrap condition bodies in brackets, even if it's just one statement. Actually thought eslint/prettier caught this.

eirikurn (approved):
Approved with 1 comment.

libs/auth-nest-tools/src/lib/create-key-provider.ts
AndesKrrrrrrrrrrr: Reverted braces change, since it caused an unrelated circular dependency :shrug:

We should set up the curly rule, which can be auto-fixed :wrench:

AndesKrrrrrrrrrrr: Also failed after revert... reapplied the braces :bow:

PR #18547, opened 2025-03-25
chore(j-s): Move civil claim upload

oddsson:
# Move civil claim upload

[Asana](https://app.asana.com/0/1199153462262248/1209602081769418/f)

## What

Move the upload input for civil claims from the case files screen to the processing screen.

## Why

On the processing screen, users choose whether or not there is a civil claim in a case. Today, the case files screen comes before the processing screen so users need to first upload a civil claim and then on the next step  say if there is a civil claim in the case.

## Screenshots / Gifs

Screenshot 2025-03-26 at 11 14 13

## Checklist:

## Summary by CodeRabbit

- **New Features**
  - Enhanced file upload management with an interactive drag-and-drop interface and improved progress feedback during processing.
  - Upgraded the civil claimant entry section with refined input options, including national ID handling.
  - Revamped the court selection display for clearer, more accessible headings.

- **Refactor**
  - Removed outdated civil claims functionality from file details views.
  - Standardized layout spacing for a more consistent user experience.


apps/judicial-system/web/src/utils/validate.ts
```diff
@@ -257,6 +257,7 @@ export const isHearingArrangementsStepValidIC = (
 
 export const isProcessingStepValidIndictments = (
   workingCase: Case,
+  allFilesDoneOrError: boolean,
```
gudjong, line 260: This method is also used in `formHelper` so you probably need to set a default value here or add the new argument to the call in `formHelper`.

gudjong (approved):
Accepted, but have one comment that needs to be addressed.

apps/judicial-system/web/src/utils/validate.ts
oddsson: Thanks! Good eye. This doesn't need to be a part of the validation function and has been removed.

PR #18572, opened 2025-03-27
fix(j-s): use transitionCase when court id is changed

thorhildurt:
# [Laga state breytingu ](https://app.asana.com/0/1199153462262248/1209797193351298)

## What
- Use transitionCase when court id is changed

## Why
- Because we want to use our internal state machine when states are changing

## Screenshots / Gifs

## Checklist:

### DIFF
--- a/apps/judicial-system/backend/src/app/modules/case/case.controller.ts
+++ b/apps/judicial-system/backend/src/app/modules/case/case.controller.ts
@@ -41,6 +41,7 @@ import type { User } from '@island.is/judicial-system/types'
 import {
   CaseOrigin,
   CaseState,
+  CaseTransition,
   CaseType,
   indictmentCases,
   investigationCases,
@@ -291,10 +292,7 @@ export class CaseController {
     const hasCourtIdUpdate =
       theCase.courtId && update.courtId && theCase.courtId !== update.courtId
     if (hasCourtIdUpdate) {
-      update.courtCaseNumber = null
-      update.judgeId = null
-      update.registrarId = null
-      update.state = CaseState.SUBMITTED
+      transitionCase(CaseTransition.SUBMIT, theCase, user)
     }
 
     return this.caseService.update(theCase, update, user) as Promise<Case> // Never returns undefined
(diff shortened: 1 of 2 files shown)


apps/judicial-system/backend/src/app/modules/case/case.controller.ts
```diff
@@ -291,10 +292,7 @@ export class CaseController {
     const hasCourtIdUpdate =
       theCase.courtId && update.courtId && theCase.courtId !== update.courtId
     if (hasCourtIdUpdate) {
-      update.courtCaseNumber = null
-      update.judgeId = null
-      update.registrarId = null
-      update.state = CaseState.SUBMITTED
+      transitionCase(CaseTransition.SUBMIT, theCase, user)
```
thorhildurt, line 295: I think it makes more sense to call this transition here when a certain change occurs in a case update compared to calling the transition case endpoint in from the client. In the client the option to change the court occurs in at least two places. 

Still making this call here or in the client, it will trigger 2x case updates compared to the previous update assignment.

apps/judicial-system/backend/src/app/modules/case/case.controller.ts
```diff
@@ -288,16 +289,31 @@ export class CaseController {
...
+    if (requiresCourtTransition) {
+      this.eventService.postEvent(
+        CaseTransition.MOVE,
+        updatedCase ?? theCase,
+        false,
+        {
+          from: theCase.court?.name,
+          to: updatedCase?.court?.name,
+        },
+      )
```
gudjong, line 313: Good that you remembered to post a Slack event, I had not thought of that.

You may actually want consider moving the move transition and event to the update method in the case service. I just noticed that that method is already handling the receive transition, so it may make sense to keep them together in one place.

gudjong (approved):
Nice and clean. One suggestion you may want to consider.

apps/judicial-system/backend/src/app/modules/case/case.controller.ts
thorhildurt: Good catch! Will refactor 👌

PR #18659, opened 2025-04-03
feat(j-s): Alternative Service Option

gudjong:
# Alternative Service Option

[Komast hjá því að gefa út fyrirkall í gáttinni](https://app.asana.com/0/1199153462262248/1209423380480960/f)

## What

- Adds the option to serve an indictment without issuing a subpoena.

## Why

- Sometimes it is necessary to serve an indictment without issuing a subpoena, for instance when a defendant cannot be reached.

### A subpoena has not been issued
https://github.com/user-attachments/assets/5d465894-929c-4b26-bbd2-06cedb7596e5

### A subpoena has already been issued
https://github.com/user-attachments/assets/2de08aeb-399d-412c-be08-75bed2382da1

## Checklist:

## Summary by CodeRabbit

- **New Features**
  - Introduced alternative service options for defendants, allowing users to indicate alternative service status and provide descriptions.
  - Updated subpoena processing to conditionally exclude defendants using alternative service.
  - Enhanced API responses, GraphQL queries, and UI components to display and manage these alternative service details.

- **Chores**
  - Performed a database schema update to incorporate the new alternative service fields.


apps/judicial-system/backend/migrations/20250401150502-gudjon,update-defendant.js
thorhildurt, line 1: Nit: The migration name contains your name, is that intended? 😅

apps/judicial-system/web/src/routes/Court/Indictments/Subpoena/Subpoena.strings.ts
```diff
@@ -1,69 +1,77 @@
...
+        'Notaður sem texti í staðfesta takka í modal glugga á Fyrirkallsskjá í ákærum.',
+    },
+    modalSecondaryButtonText: {
+      id: 'judicial.system.core:subpoena.modal_secondary_button_text',
+      defaultMessage: 'Hætta við',
+      description:
+        'Notaður sem texti í hætta takka í modal glugga á Fyrirkallsskjá í ákærum.',
+    },
+  }),
+  modalAlternativeServiceTitle: 'Skrá birtingu með öðrum hætti',
```
thorhildurt, line 73: Do we want to mix Contentfull definitions and these static probs?

apps/judicial-system/web/src/routes/Court/components/SubpoenaType/SubpoenaType.strings.ts
```diff
@@ -1,22 +1,30 @@
...
+        'Notaður sem texti fyrir Útivistarfyrirkall valkost á Fyrirkalls skjá í dómaraflæði í ákærum.',
+    },
+    arrest: {
+      id: 'judicial.system.core:court.subpoena_type.arrest',
+      defaultMessage: 'Handtökufyrirkall',
+      description:
+        'Notaður sem texti fyrir Handtökufyrirkall valkost á Fyrirkalls skjá í dómaraflæði í ákærum.',
+    },
+  }),
+  alternativeService: 'Birt með öðrum hætti',
```
thorhildurt, line 24: Same here as above 👀

thorhildurt (commented):
Mostly nit comments 😇 But when I was testing I noticed something unexpected when removing the check and navigating between the steps. 

https://github.com/user-attachments/assets/14f005f0-5607-4602-beb5-3210270d5333

apps/judicial-system/backend/migrations/20250401150502-gudjon,update-defendant.js
gudjong: Will fix

apps/judicial-system/web/src/routes/Court/Indictments/Subpoena/Subpoena.strings.ts
gudjong: I think this is a good idea. Have used it before.

apps/judicial-system/web/src/routes/Court/components/SubpoenaType/SubpoenaType.strings.ts
gudjong: I like this approach.

apps/judicial-system/web/src/routes/Court/Indictments/Subpoena/Subpoena.strings.ts
gudjong: Note, that if done correctly, all stings are stored close to where they are used. I think that this would be more of an isssue if strings were sprread around in different folders far away from where they are used.

gudjong:
> Mostly nit comments 😇 But when I was testing I noticed something unexpected when removing the check and navigating between the steps.
> 
>  Screen.Recording.2025-04-07.at.13.46.21.mov

This is actually working as intended. There is an explanation for this in the test description in the Asana task. Once an alternative service has been confirmed, it is very unlikely that either a new subpoena is issued or a new alternative service is confirmed. However, we don't want the system to prevent the court from doing that if they really want to. Rather than implementing some obscure way of providing such functionality, I decided to simply allow the court to uncheck the alternative service box and the ckeck it again to enable a new alternative service.

apps/judicial-system/web/src/routes/Court/Indictments/Subpoena/Subpoena.strings.ts
thorhildurt: Ah I should have been more clear, I am more talking about having a alternative file co-located as we do with the strings, just with a different type or structure. I am just not a fan of the different patterns within the same object defined here. Thus in some places in the code we apply a format function for some properties and others not...

But this is not a blocker and is a borderline nit (hehe) but constants are everywhere thus I am up for keeping them consistent and thus I would be up for breaking this up 😇. Could also be two objects within the string files: One for the Contentfull structure and one with our new raw structure.

thorhildurt (approved):

apps/judicial-system/web/src/routes/Court/Indictments/Subpoena/Subpoena.strings.ts
gudjong: Ok, I see what you mean. Let's just disagree and commit on this one 🫣

PR #18708, opened 2025-04-08
J s/enable civil claimant to upload files

thorhildurt:
# [Civil claimants can upload and view files](https://app.asana.com/1/203394141643832/project/1199153462262248/task/1208230818033628)

## What
- Enable civil claimants to upload files
- Enable civil claimants to view their own uploaded files (independent of case file access granted)
- Enable defenders to view their own uploaded files (independent of case file access granted)

## Why
- Required feature and behaviour

## Screenshots / Gifs

## Checklist:


apps/judicial-system/backend/src/app/modules/case/filters/case.filter.ts
```diff
@@ -383,26 +382,21 @@ const canDefenceUserAccessIndictmentCase = (
     CivilClaimant.isConfirmedSpokespersonOfCivilClaimant(
       user.nationalId,
       theCase.civilClaimants,
-    ) &&
-    !forUpdate
+    )
```
unakb, line 385: I wonder if we should rethink how we use the CaseWriteGuard (especially in terms of limited access cases) because if we remove this from here (I'm assuming it was necessary to be able to call the file endpoint) we're basically giving civil claimant spokespersons the same case update privileges as defenders. It's not a huge issue at the moment since defenders update privileges are extremely limited, but I worry it could become a hidden gotcha in the future if we add more update fields for defenders without realizing it opens them up for spokespersons, legal rights spokepersons, and any other users that are classified as the defender role.

unakb: Not something we'll fix here though of course 😊

unakb (approved):
Looks good 😊

PR #18764, opened 2025-04-11
fix(ci): `skipped` or `cancelled` jobs parsed as `success`

AndesKrrrrrrrrrrr:
We only check for non-failure, so if a job was skipped or cancelled, it's still a "success".

Additionally, if the "Success" job is cancelled, it's "skipped", thus a "success" to the merge queue, and would proceed to the merge queue even if jobs failed (because the "success" job didn't fail).

Updating to always run, and require all steps to be success or skipped.

## Summary by CodeRabbit

- **Chores**
	- Streamlined our automated dependency verification process to consolidate success checks and enhance error reporting behind the scenes.

### DIFF
--- a/.github/workflows/pullrequest.yml
+++ b/.github/workflows/pullrequest.yml
@@ -403,20 +403,15 @@ jobs:
         env:
           NEEDS: ${{ toJSON(needs) }}
         run: |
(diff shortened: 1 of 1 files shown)


pro-ingvar (approved):
LGTM 👍

PR #18777, opened 2025-04-11
chore(ci): Less verbose test output

AndesKrrrrrrrrrrr:
`build` is nice and tight output, but `tests` is a mess, and it's worse when searching. Reducing verbosity for easier failure finding. Also, we don't really need the debug output for _all_ workflow runs; should add debug flag as needed.

image

image

## Summary by CodeRabbit

- **Chores**
  - Streamlined the testing process by reducing extraneous log details for a cleaner output during test runs.


robertaandersen (approved):
LGTM

PR #18788, opened 2025-04-14
feat(j-s): victim lawyer get limited case access

thorhildurt:
# [Limited case access for victim lawyers](https://app.asana.com/1/203394141643832/project/1199153462262248/task/1209922824767033)

## What
- Grant lawyers of victims access for their assigned cases
- Victim lawyers should see their case in the case list 
     - if lawyer should get access to case when sent to court
     - if lawyer should get access when arraignment date is scheduled
     - if the lawyer should get access and case is concluded 

## Why
- Because they should have access of the overview of case they are assigned to

## Screenshots / Gifs
- Adalheidur is a victim lawyer in this case. This branch does not have the latest info card changes, but when that is merged the assignment for Adalheidur and relevant victim should be rendered in the info card. 
Screenshot 2025-04-14 at 14 42 14

## Checklist:

## Summary by CodeRabbit

- **Refactor**
  - Improved the logic for verifying user access to case records, ensuring clearer and more accurate permissions checks for defense users and victim lawyers.

- **New Features**
  - Enhanced case information retrieval to optionally include related victim details, ordered by creation date.
  - Added specific filtering criteria enabling defense users and victim lawyers to access cases based on role, lawyer status, and case state.
  - Introduced a utility method to identify victims associated with a specific lawyer under defined access conditions.


apps/judicial-system/backend/src/app/modules/case/filters/cases.filter.ts
```diff
@@ -252,6 +253,21 @@ const getDefenceUserCasesQueryFilter = (user: User): WhereOptions => {
                     },
                   ],
                 },
+                {
+                  [Op.and]: [
+                    { state: [CaseState.SUBMITTED, CaseState.RECEIVED] },
+                    {
+                      id: {
+                        [Op.in]: Sequelize.literal(`
```
thorhildurt, line 261: I don't understand why this doesn't work when fetching the case file list. Maybe there something else that I am missing 🤔 I have tested the query locally, and it works in postico but not in the app. It doesn't return a case which should be included on the case list. The only difference I spot in the console is the PII encoding: 
Screenshot 2025-04-14 at 15 39 24

thorhildurt: SOLVED

gudjong (approved):
LGTM, but did not test.

Some suggestions, no show stoppers.

apps/judicial-system/backend/src/app/modules/case/filters/cases.filter.ts
```diff
@@ -261,12 +266,76 @@ const getDefenceUserCasesQueryFilter = (user: User): WhereOptions => {
                 },
               ],
             },
+          ],
+        },
+        {
+          // victim lawyer assigned to a case
+          [Op.and]: [
+            { type: [...restrictionCases, ...investigationCases] },
```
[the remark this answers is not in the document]
thorhildurt: Originally I tried to keep them the same and re-use the conditions as much as I could, but since we define the defenders on restriction and investigation cases differently compared to lawyers per victims I decided to keep them separate.

PR #18845, opened 2025-04-23
fix(web): Add types to english search result query

mannipje:
# Add types to english search result query

## What

Added same types in query for english search result as is for icelandic search results.

## Why

So the link on english search results total matches the search results total on english version of the web.

#### IS
image

#### EN
image

#### IS
image

#### EN
image

## Checklist:

## Summary by CodeRabbit

- **Bug Fixes**
	- English search results now exclude the "WebQna" content type from the results count.

### DIFF
--- a/apps/web/screens/Search/Search.tsx
+++ b/apps/web/screens/Search/Search.tsx
@@ -84,6 +84,21 @@ import { ActionType, initialState, reducer } from './Search.state'
 
 const PERPAGE = 10
 
+const ALL_TYPES: `${SearchableContentTypes}`[] = [
+  'webArticle',
+  'webLifeEventPage',
+  'webDigitalIcelandService',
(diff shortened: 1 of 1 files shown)


apps/web/screens/Search/Search.tsx
```diff
@@ -995,12 +995,19 @@ const EnglishResultsLink: FC<
     QuerySearchResultsArgs
   >(GET_SEARCH_RESULTS_TOTAL)
 
+  const excludedType = SearchableContentTypes.WebQna
+
+  const types = Object.values(SearchableContentTypes).filter(
+    (type) => !excludedType.includes(type),
```
RunarVestmann, line 1001: I feel it would make a bit more sense if we instead had a list of `excludedTypes` because as it stands right now we are checking if a type is a substring of the webqna string.

PR #18838, opened 2025-04-22
chore(ci): Re-enable formatting PRs

AndesKrrrrrrrrrrr:
Write back formatting [was removed](https://github.com/island-is/island.is/pull/17771/files#diff-ce9a2f9036cc2a5aaf1f78839c4f65c7a12808822e0dd0f3552471921ac5e007R344). Now re-adding.

PR #18853, opened 2025-04-23
chore(ci): Move generated file cache out of working directory

AndesKrrrrrrrrrrr:
While re-adding auto-format (https://github.com/island-is/island.is/pull/18838), I noticed that the generated files tarball cache is saved to the working directory, and [then added via](https://github.com/island-is/island.is/pull/18838/files#diff-da13ca165c9f46ab090c40c10f687fd2f6090921883604e9f4775749c6c628ea) `git add .`

Moving to a more suitable location.

## Summary by CodeRabbit

- **Chores**
	- Improved dependency loading process by introducing a configurable cache path and refining environment variable handling for better file management.
	- Simplified generated files path configuration in workflows.
	- Removed cache restore and extraction steps for generated files in the Docker build process.


.github/actions/load-deps/action.yml
```diff
@@ -5,7 +5,7 @@ inputs:
   working-directory:
     description: 'Directory to run the action in'
     required: false
-    default: '.'
+    default: '/tmp/'
```
pro-ingvar, line 8: Yeah we need to find a better solution for this since every script depends on the working dir to be main/root of the monorepo.

Maybe add a new parameter to write the generated files?

pro-ingvar (changes requested):

.github/actions/load-deps/action.yml
AndesKrrrrrrrrrrr: thx, good catch :heart:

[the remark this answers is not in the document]
AndesKrrrrrrrrrrr: thx, trying it out :smile:

PR #18872, opened 2025-04-25
J s/fix email notifications for defenders in indictment cases

thorhildurt:
# [Defender is not receiving notifications for certain court dates](https://app.asana.com/0/home/1208883002434307/1209964067553878)

## What
- Add missing notifications when arraignment date is scheduled when a confirmed defender exists on the case.
- Send a notification to defender when assigned to a case if an court date is in the future and they have not received such notifications (TODO)
- Refactor court date email notifications to separate indictment from non-indictment.

## Why
- Defenders should get email notification + email invite to cases they are assigned to. 

## Screenshots / Gifs
Screenshot 2025-04-25 at 15 45 44

## Checklist:

### DIFF
--- a/apps/judicial-system/backend/src/app/modules/defendant/defendant.service.ts
+++ b/apps/judicial-system/backend/src/app/modules/defendant/defendant.service.ts
@@ -164,6 +164,7 @@ export class DefendantService {
       ) {
         // Defender was just confirmed by judge
         if (!oldDefendant.isDefenderChoiceConfirmed) {
+          // TODO: If there is an court date in the future send court date invite to defender
           messages.push({
             type: MessageType.DEFENDANT_NOTIFICATION,
             caseId: theCase.id,
--- a/apps/judicial-system/backend/src/app/modules/notification/services/caseNotification/caseNotification.service.ts
+++ b/apps/judicial-system/backend/src/app/modules/notification/services/caseNotification/caseNotification.service.ts
@@ -623,7 +623,7 @@ export class CaseNotificationService extends BaseNotificationService {
     })
   }
 
-  // this is only relevant for non-indictment cases
+  // NOTE: this is only relevant for non-indictment cases
   private sendCourtDateEmailNotificationToProsecutor(
     theCase: Case,
     user: User,
@@ -639,7 +639,7 @@ export class CaseNotificationService extends BaseNotificationService {
       arraignmentDate?.location,
       theCase.judge?.name,
       theCase.registrar?.name,
-      theCase.defenderName, 
+      theCase.defenderName,
       theCase.sessionArrangements,
     )
 
--- a/apps/judicial-system/backend/src/app/modules/notification/services/caseNotification/caseNotification.service.ts
+++ b/apps/judicial-system/backend/src/app/modules/notification/services/caseNotification/caseNotification.service.ts
@@ -54,6 +54,7 @@ import {
 } from '@island.is/judicial-system/types'
 
 import {
+  formatArraignmentDateEmailNotification,
   formatCourtHeadsUpSmsNotification,
   formatCourtIndictmentReadyForCourtEmailNotification,
   formatCourtOfAppealJudgeAssignedEmailNotification,
@@ -579,6 +580,50 @@ export class CaseNotificationService extends BaseNotificationService {
     }
   }
 
+  private sendArraignmentDateEmailNotification({
+    theCase,
+    user,
+    arraignmentDateLog,
+    recipientName,
+    recipientEmail,
+  }: {
+    theCase: Case
+    user: User
+    arraignmentDateLog: DateLog
+    recipientName: string
+    recipientEmail: string
+  }): Promise<Recipient> {
+    const { subject, body } = formatArraignmentDateEmailNotification({
+      formatMessage: this.formatMessage,
+      theCase,
+      arraignmentDateLog,
+    })
+
+    const calendarInvite = this.createICalAttachment(theCase)
+
+    return this.sendEmail({
+      subject,
+      html: body,
+      recipientName,
+      recipientEmail,
+      attachments: calendarInvite ? [calendarInvite] : undefined,
+    }).then((recipient) => {
+      if (recipient.success) {
+        // No need to wait
+        this.uploadCourtDateInvitationEmailToCourt(
+          theCase,
+          user,
+          subject,
+          body,
+          recipientEmail,
+        )
+      }
+
+      return recipient
+    })
+  }
+
+  // this is only relevant for non-indictment cases
   private sendCourtDateEmailNotificationToProsecutor(
     theCase: Case,
     user: User,
@@ -594,7 +639,7 @@ export class CaseNotificationService extends BaseNotificationService {
       arraignmentDate?.location,
       theCase.judge?.name,
       theCase.registrar?.name,
-      theCase.defenderName,
+      theCase.defenderName, 
       theCase.sessionArrangements,
     )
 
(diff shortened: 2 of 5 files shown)


apps/judicial-system/backend/src/app/modules/case/case.service.ts
```diff
@@ -1660,8 +1647,11 @@ export class CaseService {
         updatedCourtDate.date.getTime() !== courtDate?.date.getTime()
 
       if (hasUpdatedArraignmentDate || hasUpdatedCourtDate) {
-        // New arraignment date or new court date
-        await this.addMessagesForNewCourtDateToQueue(updatedCase, user)
+        await this.eventLogService.createWithUser(
```
thorhildurt, line 1650: This is part of the core changes. Instead of writing directly to the message queue, we trigger an event that will then handle the notifications depending on the case type.

apps/judicial-system/backend/src/app/modules/event-log/eventLog.service.ts
```diff
@@ -52,11 +54,7 @@ export class EventLogService {
       {
         eventType,
         caseId,
-        nationalId: user.nationalId,
-        userRole: user.role,
-        userName: user.name,
-        userTitle: user.title,
-        institutionName: user.institution?.name,
+        user,
```
thorhildurt, line 57: Changed this to make it easier to pass down user and avoid changing it to soon to a subtype of user in case functions downstream depend on the User type.

apps/judicial-system/backend/src/app/modules/notification/services/caseNotification/caseNotification.service.ts
```diff
@@ -579,6 +528,24 @@ export class CaseNotificationService extends BaseNotificationService {
     }
   }
 
+  private getCourtDateCalendarInvite = (
```
thorhildurt, line 531: This is for example a function which is now duplicated in three notification service: case, indictmentCase, defendant. 

As mentioned in the description, this would be an example of a method we could include in a common notification service that case, indictment, defendant notification services could inherit from.

thorhildurt: The duplicated methods are: 
- getCourtDateCalendarInvite
- uploadEmailToCourt
-

apps/judicial-system/backend/src/app/modules/event-log/eventLog.service.ts
```diff
@@ -98,7 +114,7 @@ export class EventLogService {
       return false
     } finally {
       if (caseId) {
-        this.addEventNotificationToQueue(eventType, caseId)
+        this.addEventNotificationToQueue(eventType, caseId, user)
       }
```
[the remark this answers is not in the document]
thorhildurt: We need it to send messages to the queue which already allows passing in the full user object, so nothing to fix here.

apps/judicial-system/backend/src/app/modules/notification/services/indictmentCaseNotification/indictmentCaseNotification.service.ts
```diff
@@ -168,11 +218,236 @@ export class IndictmentCaseNotificationService extends BaseNotificationService {
...
+    if (!user) {
+      // Nothing happens
+      return { delivered: true }
+    }
+    this.eventService.postEvent('SCHEDULE_COURT_DATE', theCase)
+
+    const promises: Promise<Recipient>[] = this.sendCourtDateEmailNotification(
+      theCase,
+      user,
+    )
```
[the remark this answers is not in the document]
thorhildurt: We already have try catch in postEvent which I think should be enough here

apps/judicial-system/backend/src/app/modules/defendant/defendant.service.ts
```diff
@@ -164,12 +164,23 @@ export class DefendantService {
       ) {
         // Defender was just confirmed by judge
         if (!oldDefendant.isDefenderChoiceConfirmed) {
+          // send general defender assignment email
```
unakb, line 167: I know this isn't a change in this PR, but the check in 166 is redundant because we already know that the !oldDefendant.isDefenderChoiceConfirmed because of the check in line 151. Maybe we should consider removing the if statement here?

apps/judicial-system/backend/src/app/modules/event-log/dto/createEventLog.dto.ts
```diff
@@ -1,45 +1,29 @@
...
 export class CreateEventLogDto {
   @IsNotEmpty()
   @IsEnum(EventType)
   @ApiProperty({ enum: EventType })
   readonly eventType!: EventType
 
+  @IsNotEmpty()
+  @IsObject()
+  @ApiProperty({ type: Object })
+  readonly user!: User
```
unakb, line 23: I think we should keep this as an optional field because the controller endpoint doesn't require a user and neither does the db, so we want to be able to create non-user connected events.

unakb: Actually now that I think about it, we should probably keep at least the nationalId as stand-alone and separate from the User type. The reason we had it that way in the first place was because we want to log certain events that _do_ have a nationalId but _don't_ have a user (like an unauthorized user trying to access the portal). In that case it would be troublesome to initialize a User type with all the required fields when all we need is the nationalId.

apps/judicial-system/backend/src/app/modules/event-log/eventLog.service.ts
unakb: This makes sense, but it doesn't completely fit in with the thinking behind the EventLog. It's not meant to be dependant on anything other than a date and event type with everything else being optional (and not necessarily interdependent).

libs/judicial-system/types/src/lib/notification.ts
```diff
@@ -33,6 +34,8 @@ export enum DefendantNotificationType {
   DEFENDER_ASSIGNED = 'DEFENDER_ASSIGNED',
   INDICTMENT_SENT_TO_PRISON_ADMIN = 'INDICTMENT_SENT_TO_PRISON_ADMIN',
   INDICTMENT_WITHDRAWN_FROM_PRISON_ADMIN = 'INDICTMENT_WITHDRAWN_FROM_PRISON_ADMIN',
+  // court date is sometimes scheduled before defender is assigned, thus we send them notification to follow-up with court date session
```
unakb, line 37: 👌

apps/judicial-system/backend/src/app/modules/notification/dto/notificationDispatch.dto.ts
```diff
@@ -19,4 +20,9 @@ export class EventNotificationDispatchDto {
   @IsEnum(EventNotificationType)
   @ApiProperty({ enum: EventNotificationType })
   readonly type!: EventNotificationType
+
+  @IsOptional()
+  @IsObject()
+  @ApiPropertyOptional({ type: Object })
+  readonly user?: User
```
unakb, line 27: I think you could skip adding the user in here, it's never currently used as the NotificationDispatch endpoint is only used by the system internally so there is no user to attach to the request.

apps/judicial-system/backend/src/app/modules/notification/notificationDispatch.service.ts
```diff
@@ -95,12 +98,44 @@ export class NotificationDispatchService {
     return this.messageService.sendMessagesToQueue(messages)
   }
 
+  private async dispatchCourtDateNotifications(
```
unakb, line 101: Nice 🙏

apps/judicial-system/backend/src/app/modules/notification/services/indictmentCaseNotification/indictmentCaseNotification.service.ts
```diff
@@ -168,11 +218,236 @@ export class IndictmentCaseNotificationService extends BaseNotificationService {
...
+    const arraignmentDate = DateLog.arraignmentDate(theCase.dateLogs)
+
+    const promises: Promise<Recipient>[] = []
+
+    // get only confirmed defenders
+    const defenders = _uniqBy(
+      theCase.defendants ?? [],
+      (d: Defendant) => d.defenderEmail,
+    )
+      .map(
```
unakb, line 331: I'm confused with this, why do we destructure and then reconstruct the exact same object?

unakb (commented):
Overall looks really good! Just a few things to think about mainly in terms of the event log 🤗

apps/judicial-system/backend/src/app/modules/defendant/defendant.service.ts
thorhildurt: Yes good catch! Will remove this statement

apps/judicial-system/backend/src/app/modules/event-log/dto/createEventLog.dto.ts
thorhildurt: Another related reason on why we require the `user` object is that in the added defender notifications  we call `uploadEmailToCourt` -> `courtService.createEmail` where the `user.name` and `user.institution.name` were being used in a code that is actually commented out 👀 

Thus I wonder if we want to keep the `name`, `institutionName` and `nationalId` in an optional object? Or just keep the `user` as optional here? At least in the court date notification that have the user dependency we check if the user is defined or not.

thorhildurt: We can also remove the `user` dependency completely because of the commented out code in `createEmail` in `apps/judicial-system/backend/src/app/modules/court/court.service.ts` and just add it if we need it.

apps/judicial-system/backend/src/app/modules/notification/services/indictmentCaseNotification/indictmentCaseNotification.service.ts
thorhildurt: We won't get the same object. We are only getting the unique required defender fields to avoid working with the full defendant object.

apps/judicial-system/backend/src/app/modules/event-log/eventLog.service.ts
```diff
@@ -98,7 +108,14 @@ export class EventLogService {
       return false
     } finally {
       if (caseId) {
-        this.addEventNotificationToQueue(eventType, caseId)
+        this.addEventNotificationToQueue({
+          eventType,
+          caseId,
+          userDescriptor: {
```
unakb, line 114: nice!

unakb (approved):
Nice! Makes sense to use a userDescriptor :D

PR #18901, opened 2025-04-30
feat(j-s): Switch User

gudjong:
# Switch User

[Skipta um emætti / hlutverk í user drop down þegar aðeins er val um tvo eligible notendur](https://app.asana.com/1/203394141643832/project/1199153462262248/task/1209918369657543)

## What

- If exactly two active users have the same national id, then switching users results in an immediate switch rather than navigation to the select user page.

## Why

- Better UX.

## Screenshots / Gifs

https://github.com/user-attachments/assets/ff5cd999-f307-4259-8341-a453c17d156f

## Checklist:

## Summary by CodeRabbit

- **New Features**
  - Improved user switching in the header menu, allowing users to activate a different eligible user via an integrated button.
- **Refactor**
  - Renamed the login component and related references to "SelectUser" for clarity and consistency. 
  - Updated import paths to reflect the new component name.


apps/judicial-system/web/src/components/Header/Header.tsx
```diff
@@ -113,11 +113,47 @@ const HeaderContainer = () => {
...
+          variant="text"
+          onClick={() => handleSelectUser(otherUser?.id)}
+          size="small"
+          preTextIcon="swapHorizontal"
+        >
+          {otherUser
+            ? otherUser.institution?.name
+            : 'Skipta um embætti / hlutverk'}
+        </Button>
+        {otherUser && (
```
oddsson, line 148: Did you run this by Guðný? I feel like this is a bit "out of place" in the user menu. What do you think?

gudjong: Yes I did. We are not always changing institution, we may have another user in the same institution with a different role.

gudjong: Showing the other user's title clarifies to the user what the other user does.

oddsson (approved):

PR #18914, opened 2025-04-30
fix(ci): Failure when newlint checking a project with no lints

AndesKrrrrrrrrrrr:
https://github.com/island-is/island.is/pull/18907 was failing due to a 0-lint project. Fixing here.

## Summary by CodeRabbit

- **Chores**
	- Improved error handling in the workflow so that the lint check step no longer fails if no lints are found, allowing the workflow to continue running.


robertaandersen (approved):
LGTM

PR #18959, opened 2025-05-07
feat(j-s): fetch criminal records from new DMR service

thorhildurt:
# [Fetch criminal records from new DMR service using required token configuration](https://app.asana.com/1/203394141643832/project/1199153462262248/task/1209942519304082)

## What
- Include `refresh_token` from the ids tokens during authorization
- Add `access_token` and `refresh_token` to session cookie
- Implement new criminal record fetch endpoint in our backend and API
- Implement the first iteration of the UI to trigger the criminal record upload

## Important points
- The new criminal record functionality is only available on dev. 
- We will begin by storing the tokens in the session cookie, similar as we do with other tokens. We will implement a more suitable storage for the refresh and access token in a follow-up PR
- We will implement the refresh_token functionality in a follow-up PR to generate new access_token once it expires
- We will improve the UI to enable the user to view the criminal record post fetch on the same screen. Currently the user has to navigate to the summary page to view the criminal record. 

## Why
- Enables prosecutors to efficiently fetch the latest criminal record for a given defendant

## Screenshots / Gifs

https://github.com/user-attachments/assets/fc7f7238-17e1-4e13-9251-bff76a0e7e2d

## Checklist:

## Summary by CodeRabbit

- **New Features**
  - Added support for fetching and uploading criminal record files for defendants via the case files section, enabled by a new feature flag.
  - Introduced a dedicated button in the case files UI to initiate criminal record file uploads.
  - Enhanced authentication flow to include refresh tokens for extended session management.

- **Improvements**
  - Updated feature flag configuration to enable controlled rollout of the criminal record upload feature.
  - Improved user feedback and error handling during criminal record file uploads.


apps/judicial-system/api/src/app/modules/auth/auth.config.ts
```diff
@@ -4,7 +4,7 @@ export const authModuleConfig = defineConfig({
   name: 'AuthModule',
   load: (env) => ({
     production: env.optional('NODE_ENV') === 'production',
-    scope: env.required('AUTH_IDS_SCOPE', 'openid profile'),
+    scope: env.required('AUTH_IDS_SCOPE', 'openid profile offline_access'),
```
unakb, line 7: Good to keep in mind and maybe write down in a follow up task that we need to also change this in the parameter store for staging and prod

apps/judicial-system/api/src/app/modules/auth/auth.controller.ts
```diff
@@ -388,6 +415,14 @@ export class AuthController {
         ...this.idToken.options,
         maxAge: EXPIRES_IN_MILLISECONDS,
       })
+      // expires in 5 minutes, can't be overridden
+      res.cookie(this.accessToken.name, accessToken, {
+        ...this.accessToken.options,
+      })
+      res.cookie(this.refreshToken.name, refreshToken, {
+        ...this.refreshToken.options,
+        maxAge: EXPIRES_IN_MILLISECONDS,
```
unakb, line 424: We should also make sure to check what the lifetime of the refreshToken is set to in the ids config by talking to the ids team to check if it's shorter than 4 hours.

unakb (approved):
Nice! Just some tiny things to note down just in case you haven't yet, but looks really good overall 🤗

apps/judicial-system/api/src/app/modules/auth/auth.config.ts
thorhildurt: As discussed async we noticed it is handled in the yaml charts instead of the parameter store

PR #19004, opened 2025-05-15
feat/terminate rental agreement

jonnigs:
## What

A new application that allows for the termination of a rental agreement. Either bey cancelation (Riftun) or dismissal (uppsögn)

## Why

This application is the counterpart to the new rental agreement application where rental agreements can be created

## Checklist:


libs/application/templates/hms/terminate-rental-agreement/src/utils/options.ts
```diff
@@ -0,0 +1,78 @@
...
+  },
+  {
+    value: terminationReasons[9],
+    label: terminationReasons[9],
+  },
+  {
+    value: terminationReasons[10],
+    label: terminationReasons[10],
+  },
+]
```
HjorturJ: Maybe throw a mapping function for this, or create an enum for these values so its a bit more clear what each of these mean in the code?

HjorturJ (approved):

PR #19042, opened 2025-05-20
feat(web): Add linked page to link in one column text

mannipje:
# Add linked page to link in one column text

## What

- Add Linked Page content type to link in One Column Text content type.
- Add organization subpage and organization parent subpage as possible reference in page on Linked page content type

## Why

- More stable using reference to page instead of using direct url 

### Now it is possible to set Linked page on link field in one column text
image

### And Linked page can reference an Organization subpage and Organization parent subpage in the page field
image

## Checklist:

## Summary by CodeRabbit

- **New Features**
  - Improved handling of linked references, enabling support for additional link types and more consistent URL generation for organization subpages.
- **Refactor**
  - Enhanced internal mapping logic for links to streamline and unify how different link types are processed.

### DIFF
--- a/libs/cms/src/lib/models/oneColumnText.model.ts
+++ b/libs/cms/src/lib/models/oneColumnText.model.ts
@@ -4,8 +4,9 @@ import { SystemMetadata } from '@island.is/shared/types'
 
 import { IOneColumnText } from '../generated/contentfulTypes'
 
-import { Link, mapLink } from './link.model'
+import { Link } from './link.model'
 import { mapDocument, SliceUnion } from '../unions/slice.union'
+import { mapReferenceLink } from './utils'
 
 @ObjectType()
 export class OneColumnText {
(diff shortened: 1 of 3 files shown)


libs/cms/src/lib/models/utils.ts
```diff
@@ -1,3 +1,12 @@
+import {
```
RunarVestmann, line 1: Could we have this as a type import?

libs/cms/src/lib/models/utils.ts
```diff
@@ -6,3 +15,61 @@ export const getRelativeUrl = (url: string) => {
...
+  }
+
+  return null
+}
+
+const generateOrganizationSubpageLinkFromLinkedPage = (
+  linkedPage: ILinkedPage,
+): Link | null => {
+  const contentTypeId = linkedPage.fields.page.sys.contentType.sys.id
+  let slug: string
```
RunarVestmann, line 41: Let's initialize this variable here as well just in case someone comes in the future and changes this function

libs/cms/src/lib/models/oneColumnText.model.ts
```diff
@@ -35,9 +36,9 @@ export const mapOneColumnText = ({
   typename: 'OneColumnText',
   id: sys.id,
   title: fields.title ?? '',
-  link: fields.link ? mapLink(fields.link) : null,
+  link: mapReferenceLink(fields.link),
```
RunarVestmann, line 39: The term reference link is already used elsewhere, could we maybe find a different name or somehow make the check here whether it is a link and either call mapLink or the generateOrganizationSubpageLinkFromLinkedPage?

RunarVestmann (commented):
A few comments

PR #19052, opened 2025-05-20
feat(j-s): refresh token logic for criminal record endpoint

thorhildurt:
# [Generate new access token via refresh token and update session cookies](https://app.asana.com/0/home/1208883002434307/1210184258936987)

## What
- This is the initial version to support the required refresh token logic to enable calling a new criminal record endpoint from the judicial system app
- We currently implement the refresh logic within our auth api, where in our client side `handleUploadCriminalRecord` method we specifically 
    - (1) call an endpoint within the auth api to verify and update the tokens if required. 
    - (2) Then we call our criminal record graphql endpoint, where the required access token should be up to date in the request headers sent
- There are definitely some improvements that can be made when incorporating this functionality (e.g. by proxying the request via the new refresh logic endpoint, or take advatange of the BFF service) but this is our initial proposed steps to get the new feature across the finish line. 

## Why
- Required to ensure we always have a valid access token through a given authentication session

## Screenshots / Gifs
Screenshot 2025-05-22 at 14 21 09

## Checklist:

## Summary by CodeRabbit

- **New Features**
  - Introduced an endpoint to refresh authentication tokens, improving session continuity for users.
  - Added client-side logic to proactively refresh tokens before file uploads.
  - Enhanced logout to revoke refresh tokens, increasing account security.

- **Bug Fixes**
  - Prevented file uploads if token refresh fails, enhancing security and reliability.


apps/judicial-system/api/src/app/modules/auth/auth.config.ts
```diff
@@ -4,6 +4,7 @@ export const authModuleConfig = defineConfig({
   name: 'AuthModule',
   load: (env) => ({
     production: env.optional('NODE_ENV') === 'production',
+    // TODO: requesting support from the ids team to set the absolute expiry for the refresh token
```
thorhildurt, line 7: https://islandis.slack.com/archives/C021XH6KJNS/p1747922673120869

apps/judicial-system/api/src/app/modules/auth/auth.controller.ts
```diff
@@ -251,6 +252,53 @@ export class AuthController {
...
+      const accessToken = req.cookies[IDS_ACCESS_TOKEN_NAME]
+      if (!this.authService.isTokenExpired(accessToken)) {
+        this.logger.debug('Token is valid')
+        res.status(200).send()
+        return
+      }
+      const refreshToken = req.cookies[IDS_REFRESH_TOKEN_NAME]
+      const idsTokens = await this.authService.refreshToken(refreshToken)
+
+      // maybe we don't need to verify again?
```
thorhildurt, line 270: Maybe we don't need this? 🤔

apps/judicial-system/api/src/app/modules/auth/auth.controller.ts
```diff
@@ -251,6 +252,53 @@ export class AuthController {
...
+        })
+      }
+      this.logger.debug('Token refresh successful')
+      res.status(200).send()
+    } catch (error) {
+      this.logger.error('Handling token expiry failed:', { error })
+
+      this.clearCookies(res)
+
+      res.redirect('/?villa=innskraning-gomul')
```
thorhildurt, line 296: This is a requirement in the docs if something fails during refresh

apps/judicial-system/web/src/services/api.ts
```diff
@@ -27,3 +27,18 @@ export const activate = async (userId: string) => {
...
+// TEMP: Initial implementation to trigger token refresh manually in the client.
+// Ideally it should be handled in a middleware like BFF
+export const prepareRequest = async () => {
+  const token = Cookie.get(CSRF_COOKIE_NAME)
+  const options = token ? { headers: { authorization: `Bearer ${token}` } } : {}
+
+  // check if tokens stored in the session cookie are expired
+  // and if expired auth api handles the refresh
+  const res = await fetch(`/api/auth/token-refresh`, options)
+  if (res.ok) {
```
oddsson, line 40: NIIIIIIT but something like 

`return { success: res.ok}` or

```
const success = res.ok
return { success }
```

might be a bit cleaner, IMO :)

oddsson (approved):

apps/judicial-system/api/src/app/modules/auth/auth.service.ts
```diff
@@ -50,6 +50,47 @@ export class AuthService {
...
+      return await response.json()
+    } else {
+      throw new UnauthorizedException(
+        `Authorization request failed with status ${response.status} - ${response.statusText}`,
+      )
+    }
+  }
+
+  getExpiry(token: string) {
+    const decodedToken = jwt.decode(token, { complete: true })
```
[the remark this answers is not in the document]
thorhildurt: We use this to check the expiry for the access token which is a jwt token 👍 Eiki was referring to the refresh token.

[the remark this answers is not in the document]
thorhildurt: We don't need to check the expiry for the refresh token since that is verified in the ids service.

PR #19066, opened 2025-05-21
chore(dsl): Linting and typing

AndesKrrrrrrrrrrr:
- Move ingress class name from `.metadata.annotations."ingress.class"` annotation to `.spec.ingressClassName`
- Some cleanup, linting, and typing

## Summary by CodeRabbit

- **New Features**
  - Added support for specifying an ingress class name for each ingress entry, allowing selection between internal and external ALB options.

- **Style**
  - Improved code readability and naming consistency without affecting application behavior.

### DIFF
--- a/apps/judicial-system/api/infra/judicial-system-api.ts
+++ b/apps/judicial-system/api/infra/judicial-system-api.ts
@@ -47,8 +47,8 @@ export const serviceSetup = (services: {
       },
       HIDDEN_FEATURES: {
         dev: '',
-        staging: 'CRIMINAL_RECORD_ENDPOINT',
-        prod: 'CRIMINAL_RECORD_ENDPOINT',
+        staging: '',
+        prod: '',
       },
     })
     .secrets({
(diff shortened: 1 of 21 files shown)


infra/src/dsl/output-generators/map-to-helm-values.ts
```diff
@@ -520,7 +520,7 @@ function getFeatureDeploymentNamespace(env: EnvironmentConfig) {
 export const HelmOutput: OutputFormat<HelmService> = {
   featureDeployment(s: ServiceDefinition, env): void {
     // Set feature-deployment prefix for URLs
-    Object.entries(s.ingress).forEach(([name, ingress]) => {
+    Object.values(s.ingress).forEach((ingress) => {
```
AndesKrrrrrrrrrrr, line 523: `name` was unused, so mapping over `.values()` should be equivalent.

infra/src/dsl/output-generators/map-to-helm-values.ts
```diff
@@ -348,26 +348,26 @@ function serializePostgres(
...
   ).map((host) =>
     ingressConf.public ?? true
       ? hostFullName(host, env)
       : internalHostFullName(host, env),
   )
 
+  const className =
+    ingressConf.public ?? true ? 'nginx-external-alb' : 'nginx-internal-alb'
   return {
+    ingressClassName: className,
```
AndesKrrrrrrrrrrr, line 367: See

https://github.com/island-is/helm-values/blob/main/helm/libs/api-template/templates/ingress.yaml#L39-L39

PR #19108, opened 2025-05-26
feat(j-s): Request cases tables for district court

unakb:
[Asana](https://app.asana.com/0/home/1203322535524092/1209953894609719)

## What

New tables for request cases for the district courts. 

## Why

Splitting up tables for faster loading and better UX

## Checklist:


apps/judicial-system/backend/src/app/modules/case-table/caseTable.cellGenerators.ts
```diff
@@ -212,6 +214,31 @@ const appealState: CaseTableCellGenerator = {
     generateAppealStateTag(c, user),
 }
 
+// Same as appealState but with a different title
+const appealCaseState = appealState
```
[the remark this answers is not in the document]
unakb: For now this is fine because they are the same, the only difference is the title.

apps/judicial-system/backend/src/app/modules/case-table/caseTable.cellGenerators.ts
```diff
@@ -150,6 +154,99 @@ const generateRequestCaseStateTag = (c: Case, user: TUser): TagValue => {
...
 
+const generateIndictmentCaseStateTag = (c: Case, user: TUser): TagValue => {
+  const {
+    state,
+    indictmentRulingDecision,
+    indictmentDecision,
+    indictmentReviewerId,
+  } = c
+
+  const courtDate = DateLog.courtDate(c.dateLogs)?.date
```
gudjong, line 167: Please note that `courtDate` in the old `CaseListEntry` is set to either `arraignmentDate` or `courtDate` depending on wheteher the case has a `indictmentDecision`. Therefore, you need to do something similar here to get the correct tag. Note that you are already doing this for the column `indictmentArraignmentDate`.

apps/judicial-system/backend/src/app/modules/case-table/caseTable.whereOptions.ts
```diff
@@ -251,27 +359,56 @@ const prosecutorsOfficeIndictmentAppealedWhereOptions = {
...
 }
 
-export const caseTableWhereOptions: Record<CaseTableType, WhereOptions> = {
-  [CaseTableType.COURT_OF_APPEALS_IN_PROGRESS]:
+const withUserFilter = (
+  whereOptions: WhereOptions,
+  user: TUser,
+): WhereOptions => {
+  return isDistrictCourtUser(user)
+    ? { ...whereOptions, court_id: user.institution?.id }
```
gudjong, line 367: Please consider what to do when `court_id` is `null`. This was handled in the old case filters, but I am not sure what to do with it now. Can we ignore it?

gudjong (approved):
A few comments and suggestions.

I am accepting the PR, but the suggestion regarding the indictment case state tag must be taken into account.

apps/judicial-system/backend/src/app/modules/case-table/caseTable.whereOptions.ts
unakb: Do we want all the courts to see cases that have no courtId? I'm not sure why we had it set that way in the old logic

apps/judicial-system/backend/src/app/modules/case-table/caseTable.cellGenerators.ts
```diff
@@ -150,6 +156,128 @@ const generateRequestCaseStateTag = (c: Case, user: TUser): TagValue => {
...
+  indictmentSubtypes?: IndictmentSubtypeMap | null,
+): string | undefined => {
+  if (!indictmentSubtypes) return undefined
+
+  const labels =
+    getAllReadableIndictmentSubtypes(indictmentSubtypes).map(capitalize)
+
+  if (labels.length === 0) return undefined
+
+  return labels.length === 1 ? labels[0] : `${labels[0]} +${labels.length - 1}`
```
gudjong, line 262: Please note that the + should be on a separate line, so this should return a string list.

PR #19109, opened 2025-05-26
fix: dev hpa scaling

brynjarorng:
# Fix dev hpa scaling

Dev currently allows too much scaling and this restricts it down to 1 min and 3 max

## Summary by CodeRabbit

- **Bug Fixes**
  - Updated default replica count settings so that services (excluding those with 'search-indexer' in their name) in both 'staging' and 'dev' environments now receive a default replica count configuration. Previously, this applied only to 'staging' environments.


robertaandersen (approved):
LGTM

PR #19120, opened 2025-05-26
feat(j-s): Sorting / Filtering

gudjong:
# Sorting / Filtering

[Útfæra röðun á dagsetningum](https://app.asana.com/1/203394141643832/project/1199153462262248/task/1210364672198426)
[Útfæra röðun á gildistíma](https://app.asana.com/1/203394141643832/project/1199153462262248/task/1210364672198423)
[Útfæra my cases](https://app.asana.com/1/203394141643832/project/1199153462262248/task/1210364672198435)

## What

- Adds a new case table cell value type for simple string values and uses it for select case table columns.
- Adds a new sort value to case table cells and sorts the new string values using that sort value.
- Uses the new sort value to correctly sort date columns and valid to columns.
- Adds support for "my cases" and uses it for district court tables.

## Why

- Part of case table port.

## Screenshots / Gifs

https://github.com/user-attachments/assets/1214af0b-c0df-400d-8b52-733dc687706d

## Checklist:

## Summary by CodeRabbit

- **New Features**
  - Added support for a new cell type in case tables, allowing display and sorting of single string values.
  - Introduced a "My cases" filter, enabling users to view only cases assigned to them when supported.
  - Case rows now indicate if a case belongs to the current user.

- **Improvements**
  - Enhanced sorting and rendering for string and string group cell types.
  - Updated table data structure to include sortable values for improved user experience.


apps/judicial-system/backend/src/app/modules/case-table/caseTable.cellGenerators.ts
```diff
@@ -278,20 +307,21 @@ const generateRequestCaseType = (
...
-}
-
-const defendants: CaseTableCellGenerator = {
+  generate: (c: Case): CaseTableCell<StringGroupValue> =>
+    generateCell({
+      strList: [
+        c.appealCaseNumber ?? '',
+        c.courtCaseNumber ?? '',
+        c.policeCaseNumbers.length > 1
+          ? `${c.policeCaseNumbers[0]} +${c.policeCaseNumbers.length - 1}`
```
thorhildurt, line 318: We are using similar logic formatting in other places when there are more values. What do you think about extracting this as a helper?

```
const getAdditionalCellValuesFormat = (values: string[]) =>
  values.length > 1 ? `${values[0]} +${values.length - 1}` : values[0]

```

apps/judicial-system/backend/src/app/modules/case-table/caseTable.cellGenerators.ts
```diff
@@ -300,46 +330,48 @@ const defendants: CaseTableCellGenerator = {
       separate: true,
     },
   },
-  generate: (c: Case): StringGroupValue | undefined =>
+  generate: (c: Case): CaseTableCell<StringGroupValue> =>
     c.defendants && c.defendants.length > 0
-      ? {
-          s: [
+      ? generateCell({
+          strList: [
```
thorhildurt, line 336: Same formatting here as above 👀

apps/judicial-system/web/src/routes/Shared/CaseTable/CaseTable.tsx
```diff
@@ -38,9 +40,16 @@ const hasCellValue = (cell: CaseTableCell): boolean => {
   return cell.value !== null && cell.value !== undefined
 }
 
+const compareString = (
+  a: string | undefined | null,
```
thorhildurt, line 44: NIT: We can also have a shorter syntax since we allow both `undefined` and `null`

```
const compareString = (
  a?: string,
  b?: string,
) => {
...
```

apps/judicial-system/web/src/routes/Shared/CaseTable/CaseTable.tsx
```diff
@@ -207,7 +229,20 @@ const CaseTable: FC = () => {
       {user && // Wait until we have a user
         (table ? (
           <>
-            <SectionHeading title={table.title} />
+            <Box display="flex" alignItems="center">
```
thorhildurt, line 232: The check is visible on error. We should also hide it on error, right?
Screenshot 2025-05-28 at 11 31 15

thorhildurt (approved):
Nice! 👏 Mostly nit picks so I will approve and you can decide if you want to address as part of this PR or in follow-up PRs.

apps/judicial-system/backend/src/app/modules/case-table/caseTable.cellGenerators.ts
gudjong: I like it, but usually don't bother until we have a third or fourth instance. But if you like, you can implement it in a later PR.

apps/judicial-system/web/src/routes/Shared/CaseTable/CaseTable.tsx
gudjong: I would rather not as the ? may lead to us forgetting to pass an argument to the function. The original forces you to call the function with two arguments. Besides `a?: string` is not enough, it must be `a?:string|null`.

apps/judicial-system/backend/src/app/modules/case-table/caseTable.cellGenerators.ts
gudjong: This is slightly different, the + is on a separate line here.

PR #19125, opened 2025-05-27
chore(j-s): Setup navigation for page groups

thorhildurt:
# [Implement navigation for the new case group lists](https://app.asana.com/1/203394141643832/project/1199153462262248/task/1210239887893584)

## What
- Add navigation to the new case group lists only for institutions that have their lists implemented. 
    - Centralise the navigation to the dashboard instead of adding route constants in place in components. 
- We have to manually add institution when a new list is ready for prod
- Note that once we release this the new case group lists will be the default

## Why
- Because we need to have navigation

## Screenshots / Gifs
https://github.com/user-attachments/assets/3add2d7b-5fda-4398-985b-6b5ebb6b7ede

## Checklist:

## Summary by CodeRabbit

- **New Features**
  - Navigation throughout the judicial system now dynamically directs users to personalized dashboard routes based on role and feature flags.
  - Introduced support for gradual rollout of case group lists with a new shared route for grouped case tables.

- **Bug Fixes**
  - Fixed a typo in the page layout message description.

- **Refactor**
  - Centralized and simplified dashboard route selection, removing duplicated and role-based conditional navigation logic.
  - Replaced static navigation URLs with dynamic user-context-based routes across multiple components and pages.

- **Documentation**
  - Updated type definitions and exports to include new route and user utilities.


libs/judicial-system/types/src/lib/user.ts
```diff
@@ -205,6 +205,23 @@ export const isCoreUser = (user?: InstitutionUser): boolean => {
   )
 }
 
+// TEMP: Use this check to gradually rollout case group to users
+// Once a case group list is ready for a role we include it here to release
+export const hasCaseGroupListsEnabled = (user?: InstitutionUser): boolean => {
```
gudjong, line 222: Please consider using the `is???User` methods rather than only the istitution. It is safer and prevent unexpected behaviour if we later add new user roles that should behave differently. Currently, only admin users are different, but later we may have others.

PR #19196, opened 2025-06-03
fix(j-s): preview when not uploaded to s3 

MargretFinnboga:
# ...

[pdf viewer - villa ](https://app.asana.com/1/203394141643832/project/1199153462262248/task/1210432043252793)

## What

- Adding previewUrl when files are added and then removing them 
- Using previewUrl if signed url cant be found

## Why

- Needed fixing 

## Checklist:

## Summary by CodeRabbit

- **Refactor**
  - Improved file handling across multiple file upload components by updating file open actions to consistently use the full file object instead of just the file ID.
  - Simplified and standardized the way files are opened from upload lists, enhancing reliability and maintainability of file interactions. 

- **Bug Fixes**
  - Enhanced error handling when opening files, ensuring a local preview is shown if a file cannot be fetched, providing a smoother user experience.


thorhildurt (commented):
Looks good and fixes the problem for the mention files 👏 

But I wonder if we can make this logic more dynamic and extract a function which can be called in all file upload components for `onOpenFile`? 

For example: if the file id is invalid (e.g `Test.pdf-9fe3c34e-7ea5-47e3-aec0-7e7dc9bc47e9`, we create a `previewUrl` otherwise we just use the previous implemented `onOpen(file.id)` function. 

Just thought of this when testing since I found another file preview not working when court uploads their update to criminal records: 
Screenshot 2025-06-04 at 13 44 38

apps/judicial-system/web/src/utils/hooks/useFileList/index.ts
```diff
@@ -41,7 +45,15 @@ const useFileList = ({ caseId, connectedCaseParentId }: Parameters) => {
       }
     },
     onError: () => {
-      toast.error(formatMessage(errors.openDocument))
+      if (
+        currentFile &&
+        currentFile.previewUrl &&
+        currentFile.id === fullAccessVariables?.input.id
+      ) {
+        openFile(currentFile.previewUrl)
```
thorhildurt: isn't the `currentFile.id === fullAccessVariables.input.id` the key condition here? Would it make sense to create the preview url here instead of creating it for all files on upload?

thorhildurt (commented):
Just left one nit but I noticed that the preview button isn't working here for the prosecutors:
Screenshot 2025-06-06 at 10 13 22

Otherwise it is looking good!

apps/judicial-system/web/src/utils/hooks/useFileList/index.ts
MargretFinnboga: I tried that, but i decided to try again, and got it to work 🙌🏼

apps/judicial-system/web/src/components/AccordionItems/IndictmentsCaseFilesAccordionItem/IndictmentsCaseFilesAccordionItem.tsx
```diff
@@ -492,7 +492,13 @@ const IndictmentsCaseFilesAccordionItem: FC<Props> = (props) => {
...
                   caseFile={item}
                   onReorder={handleReorder}
-                  onOpen={onOpen}
+                  onOpen={() =>
+                    onOpen({
+                      name: `${item.displayText}`,
+                      ...item,
+                      size: item.size ?? undefined,
+                    })
+                  }
```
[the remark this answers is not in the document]
thorhildurt: I think we just need to fix the definition here otherwise the file preview is working in the Skjalaskra 👌

PR #19205, opened 2025-06-04
feat(j-s): Case group lists for police prosecutors

thorhildurt:
# [Case group lists for police prosecutors](https://app.asana.com/1/203394141643832/project/1199153462262248/task/1209953894609722?focus=true)

## What
- Adding case lists for police prosecutors
- Adding "create new case" button for prosecutor
- Fix `created` column
- Fix column props in request cases for prosecutors
- Fix case number sort order for prosecutor

## Why
- Better UX of the the case lists in the justice system

### Police prosectutor
Screenshot 2025-06-04 at 11 06 00

### Prosecutor representative
Screenshot 2025-06-06 at 14 39 15

### District prosecutor
Screenshot 2025-06-06 at 14 52 50

## Checklist:


apps/judicial-system/web/src/routes/Shared/CaseTableGroups/CaseTableGroups.tsx
```diff
@@ -49,6 +54,16 @@ const CaseTableGroups: FC = () => {
               ))}
             </CasesDashboardLayout>
           ))}
+          {user && isProsecutionUser(user) ? (
+            <>
+              <SectionHeading title={'Stofna mál'} marginBottom={1} />
+              <Text marginBottom={3}>
```
oddsson: The `SectionHeading` component has a `description` prop for text that goes under the heading 👀 

Please consider using that for more consistent spacing and layout.

apps/judicial-system/web/src/routes/Shared/CaseTableGroups/CaseTableGroups.tsx
```diff
@@ -49,6 +54,16 @@ const CaseTableGroups: FC = () => {
               ))}
             </CasesDashboardLayout>
           ))}
+          {user && isProsecutionUser(user) ? (
+            <>
+              <SectionHeading title={'Stofna mál'} marginBottom={1} />
```
oddsson:
```suggestion
              <SectionHeading title="Stofna mál" marginBottom={1} />
```

apps/judicial-system/web/src/routes/Shared/CaseTableGroups/CaseTableGroups.tsx
```diff
@@ -49,6 +54,16 @@ const CaseTableGroups: FC = () => {
               ))}
             </CasesDashboardLayout>
           ))}
+          {user && isProsecutionUser(user) ? (
```
oddsson: The `user` parameter is optional in `isProsecutionUser` so you should be able to do 

```suggestion
          {isProsecutionUser(user) ? (
```

right?

thorhildurt: It is required in `CreateCaseButton`

oddsson: Ahh I see. Nevermind then :)

apps/judicial-system/web/src/routes/Shared/CaseTableGroups/CaseTableGroups.tsx
```diff
@@ -1,17 +1,22 @@
 import { FC, useContext } from 'react'
 import { useRouter } from 'next/router'
 
-import { AlertMessage } from '@island.is/island-ui/core'
-import { getCaseTableGroups } from '@island.is/judicial-system/types'
+import { AlertMessage, Text } from '@island.is/island-ui/core'
```
oddsson:
```suggestion
import { AlertMessage } from '@island.is/island-ui/core'
```

apps/judicial-system/web/src/routes/Shared/Cases/Cases.tsx
```diff
@@ -41,57 +37,15 @@ import CasesAwaitingConfirmationTable from '../../Prosecutor/components/CasesAwa
 import CasesAwaitingReview from '../../PublicProsecutor/Tables/CasesAwaitingReview'
 import ActiveCases from './ActiveCases'
 import { useCasesQuery } from './cases.generated'
+import { CreateCaseButton } from './CreateCaseButton'
 import { FilterOption, useFilter } from './useFilter'
 import { cases as m } from './Cases.strings'
 import * as styles from './Cases.css'
 
-interface CreateCaseButtonProps {
+export interface CreateCaseButtonProps {
```
oddsson: Is this only used in the `CreateCaseButton` component? Consider moving this there and rename it to `Props`

apps/judicial-system/web/src/routes/Shared/Cases/CreateCaseButton.tsx
```diff
@@ -0,0 +1,54 @@
...
+import * as constants from '@island.is/judicial-system/consts'
+import { capitalize } from '@island.is/judicial-system/formatters'
+import { core } from '@island.is/judicial-system-web/messages'
+import { ContextMenu } from '@island.is/judicial-system-web/src/components'
+import { UserRole } from '@island.is/judicial-system-web/src/graphql/schema'
+
+import { CreateCaseButtonProps } from './Cases'
+
+export const CreateCaseButton: FC<CreateCaseButtonProps> = (props) => {
+  const { user } = props
```
oddsson: Could we fetch the user from the `UserContext` directly, instead of passing it down as props?

thorhildurt: The user object is required in the create case button props but optional in the `UserContext`. If we fetch it from the `UserContext` we have to check again if it is defined which is already handled in the parent component.

apps/judicial-system/web/src/routes/Shared/CaseTableGroups/CaseTableGroups.tsx
oddsson: I've been thinking about this and I think it would be a bit cleaner to 

- remove this null check for `user`
- don't pass `user` into the `CreateCase` button as props.
- refactor the `CreateCase` component to not accept props
- fetch the user from `UserContext` inside `CreateCase` component
- null check the user inside the `CreateCase` component

**Reasons ⚡️**
- fetching users from context as closely to where it is used should be preferred 
- this enables us to simplify the `CreateCase` component (by eliminating props)

apps/judicial-system/backend/src/app/modules/case-table/caseTable.cellGenerators.ts
```diff
@@ -421,7 +421,10 @@ const generateCaseNumberSortValue = (
   policeCaseNumber: string,
   user: TUser,
 ): string | undefined => {
-  if (isProsecutionUser(user) || isDistrictCourtUser(user)) {
+  if (isProsecutionUser(user)) {
+    return getPoliceCaseNumberSortValue(policeCaseNumber)
```
gudjong, line 425: Should we only order by police case number for prosecutors?

apps/judicial-system/backend/src/app/modules/case-table/caseTable.whereOptions.ts
```diff
@@ -480,6 +481,8 @@ export const caseTableWhereOptions: Record<
     publicProsecutorIndictmentInReviewWhereOptions(user),
   [CaseTableType.PUBLIC_PROSECUTOR_INDICTMENT_REVIEWED]: (user) =>
     publicProsecutorIndictmentReviewedWhereOptions(user),
+  [CaseTableType.PROSECUTOR_INDICTMENT_DRAFT]: (user) =>
```
gudjong: Nitpick, but please consider adhering to the established conventions by naming the case table type and where option in an identical way: `PROSECUTOR_INDICTMENT_IN_DRAFT`.

apps/judicial-system/backend/src/app/modules/case-table/whereOptions/prosecutor.ts
```diff
@@ -35,7 +35,12 @@ const prosecutorRequestCasesSharedWhereOptions = (user: TUser) => ({
 
 export const prosecutorRequestCasesInProgressWhereOptions = (user: TUser) => ({
   ...prosecutorRequestCasesSharedWhereOptions(user),
-  state: [CaseState.DRAFT, CaseState.SUBMITTED, CaseState.RECEIVED],
+  state: [
+    CaseState.NEW,
```
gudjong, line 39: Nice.

apps/judicial-system/backend/src/app/modules/case-table/whereOptions/prosecutor.ts
```diff
@@ -54,7 +59,12 @@ export const prosecutorRequestCasesCompletedWhereOptions = (user: TUser) => ({
   ...prosecutorRequestCasesSharedWhereOptions(user),
   state: completedRequestCaseStates,
   appeal_state: {
-    [Op.not]: [CaseAppealState.RECEIVED, CaseAppealState.APPEALED],
+    [Op.or]: [
+      null,
```
gudjong, line 63: Having `null` in a list will not work.

apps/judicial-system/web/src/routes/Shared/Cases/CreateCaseButton.tsx
```diff
@@ -0,0 +1,55 @@
+import { useContext, useMemo } from 'react'
```
gudjong: Please consider putting the button in its own folder `.../Shared/CreateCaseButton` as the `.../Shared/Cases` folder will be removed soon.

libs/judicial-system/types/src/lib/tables/caseTable.ts
```diff
@@ -98,7 +99,8 @@ export const getCaseTableType = (
     }
   }
 
-  if (isPublicProsecutionUser(user)) {
+  // shared mapping for all prosecution users
+  if (isProsecutionUser(user)) {
```
gudjong, line 116: Please note that prosecutor representatives should not be routed to request case tables as they do not have access to them. Please fix.

libs/judicial-system/types/src/lib/tables/caseTableGroup.ts
```diff
@@ -40,32 +48,32 @@ const districtCourtTableGroups: CaseTableGroup[] = [
...
         type: CaseTableType.DISTRICT_COURT_INDICTMENTS_FINALIZING,
-        route: 'sakamal-i-fragangi',
+        route: CaseTableRoutes.INDICTMENT_FINALIZING,
         title: 'Sakamál í frágangi',
         description: 'Sakamál sem á eftir að senda til ríkissaksóknara.',
       },
       {
         type: CaseTableType.DISTRICT_COURT_INDICTMENTS_COMPLETED,
-        route: 'afgreidd-sakamal',
+        route: CaseTableRoutes.COMPLETED,
```
gudjong: Please change to `CaseTableRoute.INDICTMENT_COMPLETED`.

libs/judicial-system/types/src/lib/tables/caseTableGroup.ts
```diff
@@ -79,13 +87,13 @@ const courtOfAppealsTableGroups: CaseTableGroup[] = [
     tables: [
       {
         type: CaseTableType.COURT_OF_APPEALS_IN_PROGRESS,
-        route: 'mal-i-vinnslu',
-        title: 'Mál í vinnslu',
+        route: CaseTableRoutes.IN_PROGRESS,
+        title: 'Rannsóknarmál mál í vinnslu',
```
gudjong: Please change to "Rannsóknarmál í vinnslu".

libs/judicial-system/types/src/lib/tables/caseTableGroup.ts
```diff
@@ -156,37 +164,37 @@ const publicProsecutorsOfficeTableGroups: CaseTableGroup[] = [
...
         type: CaseTableType.PROSECUTORS_OFFICE_INDICTMENT_APPEAL_PERIOD_EXPIRED,
-        route: 'frestur-lidinn',
+        route: CaseTableRoutes.APPEALED_EXPIRED,
         title: 'Frestur liðinn',
         description: 'Áfrýjunarfrestur liðinn.',
       },
       {
         type: CaseTableType.PROSECUTORS_OFFICE_INDICTMENT_SENT_TO_PRISON_ADMIN,
-        route: 'mal-i-fullnustu',
+        route: CaseTableRoutes.SENT_TO_PRISON,
```
gudjong: This will change the route fo this table to `mal-til-fullnustu`. Is that intentional? If so, then that breaks the convention to have the route reflect the table title.

libs/judicial-system/types/src/lib/tables/tableGroups/prosecutorTableGroups.ts
```diff
@@ -0,0 +1,70 @@
...
+    },
+    {
+      type: CaseTableType.PROSECUTOR_REQUEST_CASES_ACTIVE,
+      route: CaseTableRoutes.ACTIVE,
+      title: 'Virk mál',
+      description: 'Virk gæsluvarðhöld, farbönn og vistanir.',
+    },
+    {
+      type: CaseTableType.PROSECUTOR_REQUEST_CASES_APPEALED,
+      route: CaseTableRoutes.REQUEST_APPEALED,
```
gudjong, line 24: Note that the idea was to have the route reflect the table title. This will result in the route being "kaerd-mal", but the title being "Mál í kæruferli". I would prefer keeping the established convension, but open to discussin alternative strategies.

libs/judicial-system/types/src/lib/tables/tableGroups/prosecutorTableGroups.ts
```diff
@@ -0,0 +1,70 @@
...
+      title: 'Mál í kæruferli',
+      description: 'Úrskurðir sem eru í kæruferli hjá Landsrétti.',
+    },
+    {
+      type: CaseTableType.PROSECUTOR_REQUEST_CASES_COMPLETED,
+      route: CaseTableRoutes.COMPLETED,
```
gudjong, line 30: Again, the route does not reflect the table title.

libs/judicial-system/types/src/lib/user.ts
```diff
@@ -68,20 +68,30 @@ export const isProsecutionUser = (user?: InstitutionUser): boolean => {
...
+      user.institution?.type &&
+      prosecutionOfficeTypes.includes(user.institution.type),
+  )
+}
 
 export const isPublicProsecutionUser = (user?: InstitutionUser): boolean => {
   return Boolean(
     user?.role &&
-      publicProsecutionRoles.includes(user.role) &&
+      user.role === UserRole.PROSECUTOR &&
```
gudjong: I think this will break some logic in the client. If you need to check for prosecutors only in the public prosecutors office, I think you need a new method to ensure you don't break previous uses of the old method. The new method could be called `isPublicProsecutorUser`.

gudjong (changes requested):
I have some questions, suggestions and change requests.

One perticular issue is with routes. My initial idea was to have routes in the client reflect table titles. So "Mál í vinnslu" has the route "mal-i-vinnslu". This PR (and perhaps other before it) breaks this convention in several places. I would really prefer that we did not.

apps/judicial-system/backend/src/app/modules/case-table/caseTable.cellGenerators.ts
thorhildurt: Yes 👍 Updated this after feedback from user on how which number they want to use to sort.

apps/judicial-system/backend/src/app/modules/case-table/whereOptions/prosecutor.ts
thorhildurt: I tested this yesterday in another list and it did work with a list using to `Op.or` operator explicitly with a list 👀 Which matches the discovery shared in [this thread](https://kolibri.slack.com/archives/C085U8XFNBT/p1749224684111469) as well.

libs/judicial-system/types/src/lib/tables/caseTableGroup.ts
thorhildurt: Good catch!

libs/judicial-system/types/src/lib/tables/caseTableGroup.ts
thorhildurt: Ah I changed it because I saw used `mal-til-fullnusu` elsewhere and was trying to have it consistent. But if we want to match the title I will refactor the route names so they will only be re-used where the table title is truly identical 👍

libs/judicial-system/types/src/lib/tables/tableGroups/prosecutorTableGroups.ts
thorhildurt: I dont have a strong opinion and it sounds good to have it match the titles. The convention might break if we change title but not the route or the other-way around but that is not a huge deal in my opinion

I will update the route names 👍

libs/judicial-system/types/src/lib/user.ts
thorhildurt: I don't think this will break a logic in the client since it should result in the same `boolean` value. The `publicProsecutionRoles` list only included one item, thus I changed it just to a direct comparison instead of using an iterative `(.includes`) check for a single item.

libs/judicial-system/types/src/lib/tables/caseTableGroup.ts
thorhildurt: Good catch!

libs/judicial-system/types/src/lib/tables/caseTable.ts
thorhildurt: Fixed this, added a specific check for the representatives first 👍

apps/judicial-system/backend/src/app/modules/case-table/whereOptions/prosecutor.ts
gudjong: You are absolutely right. Seems like I put this comment in the wrong place. There is another null a few lines above this, which is not within an or predicate.

libs/judicial-system/types/src/lib/tables/caseTableGroup.ts
gudjong: Thank you 🙏

libs/judicial-system/types/src/lib/tables/tableGroups/prosecutorTableGroups.ts
gudjong: Yeh, I have an idea that makes it less likely we break the routes. I may implement it during the current cool down.

libs/judicial-system/types/src/lib/user.ts
gudjong: Ah, I see, I missed the fact that there was only one role in the list.

However, the point with using the lists is to prevent potential issues if we add roles later. That way we don't have to remember to add a new role in multiple comparisons in multiple places. We just add the new role to an "allowed" list. I don't feel strongly about this, but this was causing us problems in the past and therefore these lists were created.

The same thing actually applies to other lists, such as `completedRequestStates`, `completedIndictmentCaseStates`, `restrictionCases`, `indictmentCases`, etc. Rather than having to remember to list all the right values, we use a list, even if they contain only one value in some cases.

apps/judicial-system/backend/src/app/modules/case-table/whereOptions/prosecutor.ts
gudjong: That line is not part of the current PR changes so I will fix that in a different PR.

gudjong (approved):
Accepted with some minor comments.

apps/judicial-system/backend/src/app/modules/case-table/whereOptions/prosecutor.ts
thorhildurt: Ahh okay 😌

PR #19228, opened 2025-06-05
App/postholf 2 way

snaerth:
# WIP

## What

## Why

## Screenshots / Gifs

## Checklist:


apps/native/app/src/components/pdf-viewer/pdf-viewer.tsx
```diff
@@ -0,0 +1,63 @@
...
+    const extraProps = {
+      activityIndicatorProps: {
+        color: '#0061ff',
+        progressTintColor: '#ccdfff',
+      },
+    }
+
+    return (
+      <Pdf
+        spacing={0}
```
snaerth, line 26: Almost exactly the same component but moved to its own file. The only difference is spacing 0 property for the pdf to expand fully to its container without shadow

apps/native/app/src/config.ts
```diff
@@ -1,12 +1,14 @@
...
+  EnvironmentConfig,
   environmentStore,
   useEnvironmentStore,
 } from './stores/environment-store'
 
-// Initial environment
-export const environments = {
+type EnvironmentId = 'prod' | 'staging' | 'dev' | 'local' | 'mock'
+
+export const environments: Record<EnvironmentId, EnvironmentConfig> = {
```
snaerth, line 11: Was going to add to the config but it was not needed 😅, added typing that definitely does not hurt

apps/native/app/src/graphql/client.ts
```diff
@@ -63,37 +57,31 @@ const retryLink = new RetryLink({
...
-          `[GraphQL error]: Message: ${message}, Location: ${JSON.stringify(
-            locations,
-          )}, Path: ${JSON.stringify(path)}`,
-        ),
-      )
-    }
+const errorLink = onError(({ graphQLErrors, networkError }) => {
+  if (graphQLErrors) {
+    graphQLErrors.map((graphQLError) =>
+      console.log(`[GraphQL error]: ${JSON.stringify(graphQLError, null, 2)}`),
```
snaerth, line 63: Logs errors in much more readable way to the console in development

apps/native/app/src/screens/document-detail/document-detail.tsx
```diff
@@ -393,9 +354,9 @@ export const DocumentDetailScreen: NavigationFunctionComponent<{
     }
   }, componentId)
 
-  useNavigationComponentDidAppear(() => {
+  useEffect(() => {
```
snaerth: This event did not trigger on Android development build and also not the native event Navigation.events().
 I Don't see the harm in using useEffect on mount instead, unless I am missing something 🤔

apps/native/app/src/screens/document-detail/components/document-list-item.tsx
```diff
@@ -0,0 +1,164 @@
...
+  const theme = useTheme()
+  const { getOrganizationLogoUrl } = useOrganizationsStore()
+
+  const height = useSharedValue(0)
+  const containerOpacity = useSharedValue(isAndroid ? 1 : 0)
+  const isExpanded = useSharedValue(isOpen)
+  const shouldAnimate = useSharedValue(false)
+
+  useEffect(() => {
+    if (!isAndroid) {
```
snaerth, line 86: The opacity animation is to prevent a small delay in rendering the message body. However in development on Android the opacity was very laggy and there for I removed it for Android

apps/native/app/src/screens/document-detail/document-communications.tsx
```diff
@@ -0,0 +1,283 @@
...
+} from './components/document-list-item'
+import {
+  FloatingBottomContent,
+  FloatingBottomFooter,
+} from './components/floating-bottom-footer'
+import { DocumentReplyScreenProps } from './document-reply'
+
+type FlatListItem = DocumentComment | { __typename: 'Skeleton'; id: string }
+
+const LAST_REPLY_HEIGHT = 100
```
snaerth: We do not know before hand the height of the last message body, so this is a arbitrary value

apps/native/app/src/screens/document-detail/document-communications.tsx
```diff
@@ -0,0 +1,283 @@
...
+  const onReplyPress = () => {
+    if (!document?.sender?.name) {
+      return
+    }
+
+    const passProps: DocumentReplyScreenProps = {
+      senderName: document.sender.name,
+      documentId,
+      subject: document?.subject,
+      onReplySuccess: (showFirstReplyInfo) => {
```
snaerth: Passing props when closing native modals in  React native navigation is not possible. A callback prop however works rather nice and better than having context or store in this instance in my opinion

apps/native/app/src/screens/document-detail/document-communications.tsx
```diff
@@ -0,0 +1,283 @@
...
+
+  const keyExtractor = useCallback(
+    (item: FlatListItem, index: number) => item.id ?? `comment-${index}`,
+    [],
+  )
+
+  const handleContentSizeChange = () => {
+    if (shouldScroll) {
+      setShouldScroll(false)
+      setTimeout(() => {
```
snaerth, line 152: React Native Flatlist is not very good at scrolling to the bottom of a list, especially when data is not ready on first render. ScrollToBottom did not work well on both Platforms but I found that scrollToOffset worked best.
This however does not work all the time on Android and very rarely fails on Ios, when that happens the scroller stops somewhere in the middle if the list, especially if it is long.
I think I tried almost every solution out there

apps/native/app/src/screens/document-detail/document-detail.tsx
```diff
@@ -180,81 +188,33 @@ const { useNavigationOptions, getNavigationOptions } =
...
 
-interface PdfViewerProps {
-  url: string
-  subject: string
-  senderName: string
-  onLoaded: (path: string) => void
-  onError: (err: Error) => void
-}
-
-const PdfViewer = React.memo(
```
snaerth, line 191: Moved to its own file

apps/native/app/src/utils/lifecycle/get-app-root.ts
```diff
@@ -27,7 +27,8 @@ export async function getAppRoot(): Promise<Layout> {
               name: ComponentRegistry.LoginScreen,
             },
           },
-        ].concat(isAuthenticated ? onboardingScreens : []),
+          ...(isAuthenticated ? onboardingScreens : []),
```
snaerth, line 30: Fix type error

apps/native/app/ios/Podfile
```diff
@@ -53,6 +53,8 @@ target 'IslandApp' do
   pod 'FirebaseABTesting', :modular_headers => true
   pod 'FirebaseInstallations', :modular_headers => true
   pod 'RNFS', :path => '../../../../node_modules/react-native-fs'
+  pod 'glog', :podspec => '../../../../node_modules/react-native/third-party-podspecs/glog.podspec', :modular_headers => true
+  pod 'DoubleConversion', :podspec => '../../../../node_modules/react-native/third-party-podspecs/DoubleConversion.podspec', :modular_headers => true
```
eirikurn: Why these pod and gem changes?

apps/native/app/src/graphql/fragments/document.fragment.graphql
```diff
@@ -1,5 +1,24 @@
 directive @client on FIELD
 
+fragment DocumentTiket on DocumentV2 {
```
eirikurn:
```suggestion
fragment DocumentTicket on DocumentV2 {
```

apps/native/app/src/graphql/queries/inbox.graphql
```diff
@@ -36,6 +36,7 @@ query GetDocumentsCategoriesAndSenders {
 query GetDocument($input: DocumentInput!, $locale: String) {
   documentV2(input: $input, locale: $locale) {
     ...ListDocument
+    ...DocumentTiket
```
eirikurn:
```suggestion
    ...DocumentTicket
```

apps/native/app/src/graphql/fragments/document.fragment.graphql
```diff
@@ -9,6 +28,8 @@ fragment ListDocument on DocumentV2 {
   categoryId
   bookmarked
   isUrgent
+  replyable
+  closedForMoreReplies
```
eirikurn, line 31: Let's double check how these new fields end up feature flagged on main. If they throw an error we may want to feature flag the query with [@include or @skip](https://www.apollographql.com/docs/apollo-server/schema/directives#default-directives).

apps/native/app/src/screens/document-detail/document-detail.tsx
eirikurn: Yeah, good question. I don't see this doing anything. Only thing I can think of is the pdfviewer affecting the push transition animation somehow.

However, this seems to work in prod, so I'm worried that maybe the dev build is bad because of some Podfile version changes. 🤔

apps/native/app/src/screens/document-detail/document-communications.tsx
```diff
@@ -0,0 +1,283 @@
...
+      <CaseNumberWrapper>
+        <Typography variant="eyebrow" color={theme.color.purple400}>
+          {intl.formatMessage({ id: 'documentCommunications.caseNumber' })}
+        </Typography>
+        <Typography variant="body3">
+          #{document?.ticket?.id ?? ticketId}
+        </Typography>
+      </CaseNumberWrapper>
+      {(!showFirstReplyInfo || !replyable) && (
+        <TopLine scrollY={scrollY} offsetTop={31} />
```
eirikurn: Do you need to hardcode `offsetTop` or can you place this inside ListWrapper with top=0?

apps/native/app/src/ui/lib/navigation-bar-sheet/navigation-bar-sheet.tsx
```diff
@@ -67,20 +75,24 @@ const CloseIcon = styled.Image`
...
+  includeContainer?: boolean
+}
+
+export function NavigationBarSheet({
+  title,
+  onClosePress,
+  style,
+  showLoading,
+  closable = true,
+  includeContainer = false,
```
eirikurn: ~I wonder if there is ever a use case for rendering the Header without container / horizontal margin. Currently every single use of NavigationBarSheet passes in `style={{ marginHorizontal: 16 }}`.

Should just make it the default / only option until we find that use case :)

apps/native/app/src/ui/lib/list/list-item.tsx
```diff
@@ -60,13 +61,30 @@ const Row = styled.View`
...
 `
 
+const SubtitleWrapper = styled.View<{ replyable?: boolean }>`
+  flex-direction: row;
+  align-items: center;
+  flex: 1;
+  ${({ replyable, theme }) =>
+    replyable &&
+    css`
+      margin-right: ${theme.spacing[2]}px;
```
eirikurn: ~Why does this need right margin only when replyable?

apps/native/app/src/ui/lib/button/button.tsx
```diff
@@ -101,10 +102,16 @@ const Text = styled.Text<{
...
+  noMargin?: boolean
+  iconPosition?: 'left' | 'right'
+}>`
   width: 16px;
   height: 16px;
-  margin-left: ${(props) => (props.noMargin ? '0' : '8px')};
+  margin-left: ${({ noMargin, iconPosition }) =>
+    noMargin ? '0' : iconPosition === 'left' ? '0' : '8px'};
+  margin-right: ${({ noMargin, iconPosition }) =>
+    noMargin ? '0' : iconPosition === 'right' ? '0' : '8px'};
```
eirikurn: Should skip the fragment and configure flex gap on Host?

apps/native/app/src/ui/lib/button/button.tsx
```diff
@@ -101,10 +102,16 @@ const Text = styled.Text<{
   text-align: ${(props) => (props.isUtilityButton ? 'left' : 'center')};
 `
 
-const Icon = styled.Image<{ noMargin?: boolean }>`
+const Icon = styled.Image<{
+  noMargin?: boolean
+  iconPosition?: 'left' | 'right'
```
eirikurn: ~I try to avoid non-directional words in UI code, as not all cultures are LTR. Eg css prefers `start` and `end`.

apps/native/app/src/contexts/user-provider.tsx
```diff
@@ -0,0 +1,35 @@
...
+type UserContextType = {
+  user: GetProfileQuery['getUserProfile'] | null
+  refetch(): void
+  loading: boolean
+  error?: ApolloError
+}
+
+export const UserContext = createContext<UserContextType | undefined>(undefined)
+
+export const UserProvider = ({ children }: PropsWithChildren) => {
```
eirikurn: Not sure if this provider is a good idea:
- It is not integrated into the auth store lifecycle afaics. Does it do a failing API request when I'm not logged in? When does it refetch? When I log in? When I switch users?
- It adds one more API request and load on our servers everytime users open the app, even when we don't need it.
- If we want this in a context, we should add it to the current userinfo request logic.

I think it's fine to fetch the user in document-reply. You can also pass the email from the reply mutation to onReplySuccess to show it on the document-communication screen.

apps/native/app/src/screens/document-detail/document-reply.tsx
```diff
@@ -0,0 +1,241 @@
...
+
+const Row = styled(Container)<{
+  hasBottomBorder?: boolean
+  paddingVertical?: Spacing
+}>(({ hasBottomBorder = true, paddingVertical = 1, theme }) => ({
+  flexDirection: 'row',
+  paddingTop: theme.spacing[paddingVertical],
+  paddingBottom: theme.spacing[paddingVertical],
+  ...(hasBottomBorder && {
+    borderBottomWidth: theme.border.width.standard,
```
eirikurn: Should use [StyleSheet.hairlineWidth](https://reactnative.dev/docs/stylesheet#hairlinewidth) for most UI borders like these. Eg design uses 0.5 width border for a retina screen.

apps/native/app/src/screens/document-detail/components/document-list-item.tsx
```diff
@@ -0,0 +1,164 @@
...
+  const source = getOrganizationLogoUrl(sender, 75, true)
+
+  return (
+    <Animated.View style={containerAnimatedStyle}>
+      <Wrapper
+        onPress={handlePress}
+        highlightColor={closeable ? theme.shade.shade100 : theme.color.white}
+        hasTopBorder={hasTopBorder}
+      >
+        <>
```
eirikurn, line 130: ~Should skip fragment?

apps/native/app/src/screens/document-detail/document-reply.tsx
```diff
@@ -0,0 +1,241 @@
...
+    showModal(ComponentRegistry.SettingsScreen)
+  }
+
+  useEffect(() => {
+    // Make sure we only instantiate the modal once
+    if (!userProfile?.email && !hasShownModal) {
+      setHasShownModal(true)
+      showModal(ComponentRegistry.RegisterEmailScreen)
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
eirikurn: ~Why not list all deps?

apps/native/app/src/screens/document-detail/document-reply.tsx
```diff
@@ -0,0 +1,241 @@
...
+              >
+                {intl.formatMessage({ id: 'documentReply.from' })}
+                {': '}
+              </Typography>
+              <Typography
+                variant="eyebrow"
+                numberOfLines={1}
+                ellipsizeMode="tail"
+              >
+                {userProfile?.email}
```
eirikurn, line 203: Could not see an email in the video. Perhaps due to test environment?

apps/native/app/src/screens/document-detail/document-communications.tsx
eirikurn: ~Weird. This is a very common list pattern in all messaging apps. Maybe the trick revolves around somehow reversing the FlatList?

apps/native/app/src/screens/document-detail/document-reply.tsx
```diff
@@ -0,0 +1,241 @@
...
+  const [sendMessage, { loading: sendMessageLoading }] =
+    useDocumentReplyMutation({
+      onCompleted: (data) => {
+        const id = data.documentsV2Reply?.id
+
+        if (id) {
+          // Successful reply, clear message
+          setMessage('')
+          onReplySuccess?.(isFirstReply)
+          dismissModal(componentId)
```
eirikurn, line 111: For the first reply, the modal is on top of document-detail and needs to  somehow navigate to the communication screen with proper first reply props.

apps/native/app/src/screens/document-detail/document-communications.tsx
```diff
@@ -0,0 +1,283 @@
...
+                  id: 'documentCommunications.initialReply',
+                },
+                {
+                  email: user?.email ?? '',
+                },
+              )}
+              hasBorder
+            />
+          </Container>
+        )}
```
eirikurn: Looks like these appear above the FlatList, but in the design these alerts are floating in the bottom near the CTA. Am I missing something?

apps/native/app/src/screens/document-detail/document-communications.tsx
```diff
@@ -0,0 +1,283 @@
...
+    (item: FlatListItem, index: number) => item.id ?? `comment-${index}`,
+    [],
+  )
+
+  const handleContentSizeChange = () => {
+    if (shouldScroll) {
+      setShouldScroll(false)
+      setTimeout(() => {
+        flatListRef.current?.scrollToOffset({
+          offset: 75 * (comments.length - 1),
```
eirikurn: Should not hard code row height?

eirikurn (changes requested):
Sweet! Hope you haven't missed my PR reviews too much :)

apps/native/app/src/ui/lib/navigation-bar-sheet/navigation-bar-sheet.tsx
eirikurn: Using style instead of includeContainer for consistency until we just make it the default everywhere in another PR.

apps/native/app/ios/Podfile
eirikurn: Reverting to lower pre-summer build risks.

apps/native/app/src/screens/document-detail/document-detail.tsx
eirikurn: Reverting to lower pre-summer build risks

eirikurn (approved):

PR #19278, opened 2025-06-11
chore(j-s): Create and Update Return Values

gudjong:
# Create and Update Return Values

[Skila flötum object út create og update aðgerðum[(https://app.asana.com/1/203394141643832/project/1199153462262248/task/1209798354588864)

## What

## Why

## Screenshots / Gifs

## Checklist:

### DIFF
--- a/apps/judicial-system/web/src/routes/Prosecutor/Indictments/Defendant/Defendant.tsx
+++ b/apps/judicial-system/web/src/routes/Prosecutor/Indictments/Defendant/Defendant.tsx
@@ -368,6 +368,7 @@ const Defendant = () => {
         const createdCase = await createCase(workingCase)
 
         if (createdCase) {
+          // TODO: Do not create an empty defendant?
           workingCase.defendants?.forEach(async (defendant, index) => {
             if (
               index === 0 &&


apps/judicial-system/backend/src/app/modules/court-session/courtSession.service.ts
```diff
@@ -32,6 +32,8 @@ export class CourtSessionService {
   constructor(
     private readonly courtSessionRepositoryService: CourtSessionRepositoryService,
     private readonly courtDocumentService: CourtDocumentService,
+    // TODO: Move to a repository service - models should only be used in repository services
```
thorhildurt, line 35: Do we have task for this?

apps/judicial-system/backend/src/app/modules/indictment-count/indictmentCount.service.ts
```diff
@@ -144,27 +144,7 @@ export class IndictmentCountService {
...
-    })
-
-    if (!indictmentCount) {
-      throw new InternalServerErrorException(
-        `Could not find indictment count ${indictmentCountId} of case ${caseId}`,
-      )
-    }
-
-    return indictmentCount
+    return updatedIndictmentCounts[0]
```
thorhildurt, line 147: Sidenote: This update pattern is in multiple places. I was thinking if we should consider extracting it as a db utility and use for all current updates where we apply this pattern. 

For example we have the same error check in multiple places where we check the range of the updated rows before returning the first instance of the returned list of updated rows. 

What do you think?

apps/judicial-system/backend/src/app/modules/user/user.service.ts
```diff
@@ -124,7 +125,7 @@ export class UserService {
       throw new NotFoundException(`Could not update user ${userId}`)
```
thorhildurt, line 125: Nit: I find this throwing `NotFoundException` misleading here, because if the affected rows > 0, entries were definitely found but something unexpected happened sine we were only expecting one entry to be updated. Maybe we can throw `InternalServerErrorException` with "Unexpected number of rows (${numberOfAffectedRows}) affected when updating user ${user.id}"

Also the rows would be updated despite us throwing this exception here, right?

thorhildurt: But we could also standardise this pattern based on my previous comment above. Not part of this PR though, but we could add it to our backlog.

apps/judicial-system/api/src/app/modules/case/case.resolver.ts
```diff
@@ -68,6 +68,7 @@ export class CaseResolver {
 
   @ResolveField(() => String, { name: 'penalties', nullable: true })
   getPenalties(@Parent() theCase: Case, @CurrentGraphQlUser() user: User) {
+    // TODO: Move role base access control to the backend
```
thorhildurt, line 71: Do we have a task for this?

thorhildurt (approved):
Approving but left some comments regarding potential future refactoring 👀 

Also I noticed in the task that there was a point about limited exposed graphql fields in the client. Will we address that in a follow-up or was that not needed because of the changes in this PR?

gudjong:
> Approving but left some comments regarding potential future refactoring 👀
> 
> Also I noticed in the task that there was a point about limited exposed graphql fields in the client. Will we address that in a follow-up or was that not needed because of the changes in this PR?

The task discription had a dump of what I was thinking at the time. Some of the ideas could not realistically implemented at this time including limiting the GraphQL queries. That would entail  a separate query for each page or group of pages. No need to keep the comments for future refernces, we can create a new task if this bothers us again in the future.

apps/judicial-system/backend/src/app/modules/court-session/courtSession.service.ts
gudjong: No, but this is included as a concer in our architecture project. I have been gradually creating these repository services although it has been a couple of months since I touched this.

apps/judicial-system/backend/src/app/modules/indictment-count/indictmentCount.service.ts
gudjong: We could do that, but only if we can ensure type safty, perhaps using a template function. But this doesn't really bother me. Duplication is not always bad.

apps/judicial-system/backend/src/app/modules/user/user.service.ts
gudjong: The error is only thrown when the number of affected rows is zero, i.e. the update failed. If the number of rows is creater than one, we only log an error which will show up in the logs. But we can probably stop checking for multiple row updates in most cases because we are almost always updating based on primary keys.

Using a helper function is an option, but I don't really mind duplication in this case.

apps/judicial-system/api/src/app/modules/case/case.resolver.ts
gudjong: No, I didn't think this was that important. However, we discussed a bug this morning that will address this.

PR #19284, opened 2025-06-11
fix(application-system): Update buildOverviewField

veronikasif:
## What

Updated buildOverviewField 
- Fixed spacing 
- Added async items

## Screenshots / Gifs

Before:
image

After:
image

## Checklist:

### DIFF
--- a/libs/application/templates/examples/example-inputs/src/utils/overviewUtils.ts
+++ b/libs/application/templates/examples/example-inputs/src/utils/overviewUtils.ts
@@ -1,11 +1,14 @@
+import { ApolloClient } from '@apollo/client'
 import { getValueViaPath } from '@island.is/application/core'
-import { m } from '../lib/messages'
 import {
   AttachmentItem,
   ExternalData,
   FormValue,
   KeyValueItem,
 } from '@island.is/application/types'
+import { friggSchoolsByMunicipalityQuery } from '../graphql/sampleQuery'
+import { m } from '../lib/messages'
+import { FriggSchoolsByMunicipality } from './types'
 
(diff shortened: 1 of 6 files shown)


libs/application/templates/examples/example-inputs/src/utils/overviewUtils.ts
```diff
@@ -71,6 +74,43 @@ export const getOverviewItems = (
...
+      width: 'half',
+      keyText: 'Half width',
+      valueText:
+        getValueViaPath<string>(answers, 'applicant.phoneNumber') ?? '',
+    },
+    {
+      width: 'half',
+      keyText: 'Half width',
+      valueText: municipality,
+    },
```
jonnigs, line 113: I'm probably missing some context, but why is there a need to do a async GraphQL fetch in the overview?

veronikasif: We are using `buildAsyncSelectField` in some applications so we only have the selected item's id, not the full object or display label. We need to fetch the correct display label based on those Ids

jonnigs: Is it possible to fetch the options with a dataProvider in the prereq step or with an "onEntry" in the template? Then the select fields don't have to be async since they will be in the externalData and we can avoid the complexity of fetching the data again for the overview?

veronikasif: In some cases, we are retrieving very large lists (over 100 items) and would prefer not to store it in externalData

PR #19314, opened 2025-06-16
fix(j-s): qa feedback

thorhildurt:
# [Case list fixes for prosecutor lists ](https://app.asana.com/1/203394141643832/project/1199153462262248/task/1210557142168756)

## What
After a feedback from QA this PR fixes
- Only includes custody, travel ban, admission to faculty for active cases in restriction cases for prosecutors
- Correcting incorrect state column, should be appeal state column in active cases
- Adding missing prosecutor columns in waiting for confirmation table
## Why
- Not the expected designs

## Checklist:

### DIFF
--- a/apps/judicial-system/backend/src/app/modules/case-table/caseTable.whereOptions.ts
+++ b/apps/judicial-system/backend/src/app/modules/case-table/caseTable.whereOptions.ts
@@ -205,7 +205,7 @@ const prisonDoneWhereOptions = {
   valid_to_date: { [Op.lt]: fn('NOW') },
 }
 
-// Prison Admin Indictments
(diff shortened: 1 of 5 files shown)


apps/judicial-system/backend/src/app/modules/case-table/caseTable.whereOptions.ts
```diff
@@ -205,7 +205,7 @@ const prisonDoneWhereOptions = {
   valid_to_date: { [Op.lt]: fn('NOW') },
 }
 
-// Prison Admin Indictments
+// Prison Admin restriction cases
```
unakb, line 208: ❤️

PR #19370, opened 2025-06-19
feat(j-s): Add court case number to stepper

unakb:
[Asana](https://app.asana.com/1/203394141643832/project/1199153462262248/task/1210457401556074)

## What

Display the court case number above the side stepper for district court users 
Also made the UI a bit nicer in the smallest screen sizes 

## Why

So it's easier for users to keep track of which case they're currently working on without needing to scroll to the top. 

## Checklist:

## Summary by CodeRabbit

- **New Features**
  - The court case number is now displayed below the case title for district court users when available.
- **Style**
  - Improved layout spacing for the case title and court case number to ensure consistent appearance.

### DIFF
--- a/apps/judicial-system/web/src/components/PageLayout/PageLayout.tsx
+++ b/apps/judicial-system/web/src/components/PageLayout/PageLayout.tsx
@@ -17,6 +17,7 @@ import {
 import { getStandardUserDashboardRoute } from '@island.is/judicial-system/consts'
 import {
   isDefenceUser,
+  isDistrictCourtUser,
   isIndictmentCase,
 } from '@island.is/judicial-system/types'
 import {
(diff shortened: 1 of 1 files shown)


apps/judicial-system/web/src/components/PageLayout/PageLayout.tsx
```diff
@@ -156,6 +163,11 @@ const SidePanel: FC<SidePanelProps> = ({
                 { caseType: workingCase.type },
               )}
             </Text>
+            <Text>
+              {showCourtCaseNumber && workingCase.courtCaseNumber
+                ? workingCase.courtCaseNumber
+                : '\u00A0'}
```
oddsson, line 169: Could we do 

```
{showCourtCaseNumber && workingCase.courtCaseNumber && (
    <Text>
        {workingCase.courtCaseNumber}
    </Text>
)}
```
instead? I think the `'\u00A0'` part is not needed 🤔

oddsson (commented):
One thought ..

apps/judicial-system/web/src/components/PageLayout/PageLayout.tsx
unakb: We could, but the reason I put it there is to prevent the side panel list from moving down when the case number is added in the receive step, so we'd be reintroducing that issue

oddsson: Ah, I see. NVM then :)

PR #19427, opened 2025-06-24
feat(application-system-api): Change getReturnUrl to also return cancelUrl for payment

baering:
## What

Include a `cancelUrl` when initialising payments

## Why

#19422 adds the ability to define a special cancel url that user is redirected to when they cancel a payment

## Screenshots / Gifs

## Checklist:

### DIFF
--- a/libs/application/api/payment/src/lib/payment.service.ts
+++ b/libs/application/api/payment/src/lib/payment.service.ts
@@ -396,6 +396,8 @@ export class PaymentService {
     const onUpdateUrl = new URL(this.config.paymentApiCallbackUrl)
     onUpdateUrl.pathname = '/application-payment/api-client-payment-callback'
 
+    const { returnUrl, cancelUrl } = await this.getReturnUrls(applicationId)
+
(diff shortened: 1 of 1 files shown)


libs/application/api/payment/src/lib/payment.service.ts
```diff
@@ -683,6 +686,9 @@ export class PaymentService {
     returnUrl.pathname = `umsoknir/${applicationSlug}/${applicationId}`
     returnUrl.search = 'done'
 
-    return returnUrl.toString()
+    const cancelUrl = new URL(this.config.clientLocationOrigin)
+    cancelUrl.pathname = `umsoknir/${applicationSlug}`
```
norda-gunni:
```suggestion
    cancelUrl.pathname = `umsoknir/${applicationSlug}` // Not including the applicationId to avoid getting forwarded back to the payment screen since the application will be in the payment state
```

norda-gunni (approved):
Approved but adding an optional suggestion for a clarification comment on the cancelUrl construction

PR #19433, opened 2025-06-24
chore: Formatting

AndesKrrrrrrrrrrr:
## Summary by CodeRabbit

- **Style**
  - Improved formatting and indentation across multiple files for enhanced code readability and consistency.
  - Consolidated and standardized import statements and array definitions to single-line or multi-line formats as appropriate.
  - Reformatted JSX and function parameter declarations for clarity and uniformity.
  - Updated comments and removed redundant blank lines or trailing spaces.

- **Chores**
  - Expanded and reorganized the Prettier ignore list for better structure and coverage.
  - Updated JSON and configuration files to use consistent array formatting.
  - Added a new path mapping entry in the TypeScript configuration.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,17 +1,34 @@
 # Add files here to ignore them from prettier formatting
+# Uses .gitignore syntax
 
-/dist
-/coverage
+# CI/.github
 .github/actions/dist*
 .github/workflows/stale-prs.yml
+
+# Schema and generated
 /libs/api/domains/application/gen
+gen/
+*.generated.ts
+schema.d.ts
+schema.tsx
+schema.ts
+/infra/helm/
+/dist
+/coverage
+
+# Cache, temp
 /tmp
 /cache
 /.cache
+/*-cache
+/.cypress-cache
+/.nx/cache
+/.nx/workspace-data
+
+# Tooling?
 /tools/**/*.js
 **/*.hbs
 .yarn
-/infra/helm/
-/.nx/cache
-/.nx/workspace-data
+
+# Assets
 apps/web/public/assets/pdf.worker.min.mjs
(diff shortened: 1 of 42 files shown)


.prettierignore
```diff
@@ -1,17 +1,34 @@
...
 .github/actions/dist*
 .github/workflows/stale-prs.yml
+
+# Schema and generated
 /libs/api/domains/application/gen
+gen/
+*.generated.ts
+schema.d.ts
+schema.tsx
+schema.ts
```
baering: Do we not want our generated schema files to be formatted according to prettier rules?

For example when running codegen there are usually hooks in codegen files
- [Example from web](https://github.com/island-is/island.is/blob/main/apps/web/codegen.yml#L23)
- [Example from app](https://github.com/island-is/island.is/blob/main/apps/native/app/codegen.yml#L16)

image

If I'm not mistaken this change would skip all these schemas, right?

AndesKrrrrrrrrrrr: Yeah, it'd be nice if they were formatted as well, but they are generated with bad formatting, causing the format check to fail. That's why I decided to ignore them.

I saw that our codegen has a `...After` hook/command, which in some runs `prettier` on the generated files ([example](https://github.com/island-is/island.is/blob/d69aad34299e5632c52d1a3664a59a01b65b2239/apps/web/codegen.yml#L24))

AndesKrrrrrrrrrrr: Actually, the generated files shouldn't even be present when running format check :thinking: 
I did, however, get [formatting errors](https://github.com/island-is/island.is/actions/runs/15860061384/job/44715007824#step:4:260) at one point, even though I removed the codegen loading step. Hmm... maybe that was just

baering (approved):

PR #19456, opened 2025-06-25
feat(application-system): add custom HMS scope

jonnigs:
## What

Add custom scope to HMS applications

## Why

So applications can be filled out by someone with auth delegations, for example for companies

## Checklist:


HjorturJ (approved):

libs/auth-api-lib/seeders/data/scope-hms.ts
```diff
@@ -0,0 +1,13 @@
+import { createScope } from './helpers'
+
+export const up = createScope({
+  name: '@island.is/applications/hms',
+  displayName: 'Húsnæðis og mannvirkjastofnun',
+  description: 'Umsóknir hjá Húsnæðis og mannvirkjastofnun',
+
+  delegation: {
+    custom: true,
+    procuringHolders: true,
```
valurefugl, line 10: not general mandate?

jonnigs: The only options here are custom, procurationHolder and legalGuardians. I have general mandate in the template for the application it self
image

valurefugl: not entirely sure if/how this will work. but approving the PR. maybe general mandate needs to be added to the scope in the admin UI? @saevarma might have more insight.

pro-ingvar (approved):
LGTM

PR #19459, opened 2025-06-26
feat(medical-and-rehabilitation-payments): Update Questions Screen

veronikasif:
[[TS-1017] Questions Screen](https://dit-iceland.atlassian.net/jira/software/c/projects/TS/boards/82?issueParent=69125&selectedIssue=TS-1017)
[[TS-1070] Move "Færðu greiðslur frá öðru landi...?" to separate screen](https://dit-iceland.atlassian.net/jira/software/c/projects/TS/boards/82?selectedIssue=TS-1070)

## Checklist:

### DIFF
--- a/libs/api/domains/social-insurance/src/lib/models/general/countries.model.ts
+++ b/libs/api/domains/social-insurance/src/lib/models/general/countries.model.ts
@@ -1,0 +1,10 @@
+import { Field, ObjectType } from '@nestjs/graphql'
+
+@ObjectType('SocialInsuranceGeneralCountries')
+export class Countries {
+  @Field({ nullable: true })
+  code?: string
+
+  @Field({ nullable: true })
+  name?: string
+}
--- a/libs/api/domains/social-insurance/src/lib/models/general/educationalInstitutions.model.ts
+++ b/libs/api/domains/social-insurance/src/lib/models/general/educationalInstitutions.model.ts
@@ -1,0 +1,10 @@
+import { Field, ObjectType } from '@nestjs/graphql'
+
+@ObjectType('SocialInsuranceGeneralEducationalInstitutions')
+export class EducationalInstitutions {
+  @Field({ nullable: true })
+  name?: string
+
+  @Field({ nullable: true })
+  nationalId?: string
+}
--- a/libs/clients/social-insurance-administration/src/lib/socialInsuranceAdministrationClient.type.ts
+++ b/libs/clients/social-insurance-administration/src/lib/socialInsuranceAdministrationClient.type.ts
@@ -145,7 +145,9 @@ export type IncomePlanStatus = 'Accepted' | 'Cancelled' | 'InProgress'
 export interface Occupation {
   isSelfEmployed: boolean
   isStudying: boolean
+  educationalInstitution?: string
   isPartTimeEmployed: boolean
+  receivingPaymentsFromOtherCountry: boolean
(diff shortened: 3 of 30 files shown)


libs/api/domains/social-insurance/src/lib/models/general/countries.model.ts
```diff
@@ -0,0 +1,10 @@
+import { Field, ObjectType } from '@nestjs/graphql'
+
+@ObjectType('SocialInsuranceGeneralCountries')
+export class Countries {
+  @Field({ nullable: true })
+  code?: string
+
+  @Field({ nullable: true })
+  name?: string
```
thorkellmani: If both properties are undefined, the object is pretty meaningless and should be filtered out of the response. Shouldn't at least one of the properties be required?

Also, this seems to represent a single country and thus should be singular.
```suggestion
@ObjectType('SocialInsuranceGeneralCountry')
export class Country {
  @Field({ nullable: true })
  code?: string

  @Field({ nullable: true })
  name?: string
```

libs/api/domains/social-insurance/src/lib/models/general/educationalInstitutions.model.ts
```diff
@@ -0,0 +1,10 @@
+import { Field, ObjectType } from '@nestjs/graphql'
+
+@ObjectType('SocialInsuranceGeneralEducationalInstitutions')
+export class EducationalInstitutions {
+  @Field({ nullable: true })
+  name?: string
+
+  @Field({ nullable: true })
+  nationalId?: string
+}
```
thorkellmani: Same as above, one of the fields should be required at least. Are there any institutions which don't have a name or a national id?


```suggestion
@ObjectType('SocialInsuranceGeneralEducationalInstitution')
export class EducationalInstitution {
  @Field({ nullable: true })
  name?: string

  @Field({ nullable: true })
  nationalId?: string
}
```

libs/clients/social-insurance-administration/src/lib/socialInsuranceAdministrationClient.type.ts
```diff
@@ -145,7 +145,9 @@ export type IncomePlanStatus = 'Accepted' | 'Cancelled' | 'InProgress'
 export interface Occupation {
```
thorkellmani, line 145: This is a nice dto, why not make a graphql model "socialInsuranceOccupation" (maybe more specific to the application type), make a new resolver if needed and field resolve each property? Seems better to have a single query for kinda related data instead of all the general queries.

thorkellmani (changes requested):
Some graphql naming issues.

Why have all these different queries in a general resolver? Why not just one good query representing the domain object? By having a new query for each endpoint the graphql layer just become REST. 
I made modelling suggestion in the last comment.

libs/api/domains/social-insurance/src/lib/models/general/countries.model.ts
veronikasif: Updated to singular and made `code` required

libs/api/domains/social-insurance/src/lib/models/general/educationalInstitutions.model.ts
veronikasif: Updated to singular and made `name` required

thorkellmani (approved):
LGTM 🤝  🚀

PR #19516, opened 2025-06-30
feat(admin-portal): payments module

magnearun:
New Payment module for Admin portal

## Why

Allow look up of payment flows by flow id and payer nationalId

## Screenshots / Gifs

Screenshot 2025-06-30 at 13 26 17
Screenshot 2025-06-30 at 13 26 27

## Checklist:

## Summary by CodeRabbit

**New Features**
- Introduced a Payments module in the admin portal with integrated navigation and access control.
- Added payment flow listing and detailed views, including event histories and statuses with localized UI tags.
- Implemented search and cursor-based pagination for payment flows with filtering by payer ID and payment ID.
- Enabled viewing detailed payment events and their reasons within payment flow details.
- Added GraphQL queries and REST API endpoints to fetch payment flows and associated events.

**Bug Fixes**
- Enhanced pagination accuracy by counting distinct payment flows in database queries.

**Documentation**
- Added README and setup instructions for the Payments admin portal module.

**Chores**
- Added TypeScript, Babel, ESLint, Jest, and GraphQL code generation configurations to support the new Payments module.


apps/services/payments/src/app/paymentFlow/paymentFlow.service.ts
```diff
@@ -693,4 +721,68 @@ export class PaymentFlowService {
...
+    after?: string,
+    before?: string,
+  ): Promise<GetPaymentFlowsPaginatedDTO> {
+    const where: WhereOptions<PaymentFlow> = {}
+
+    if (payerNationalId && isValid(payerNationalId)) {
+      where.payerNationalId = payerNationalId
+    }
+
+    if (search && search !== '') {
```
baering, line 736: `isGuid(search)`?

apps/services/payments/src/app/paymentFlow/paymentFlow.service.ts
```diff
@@ -693,4 +721,68 @@ export class PaymentFlowService {
...
+      where.id = search
+    }
+
+    const paginatedResult = await paginate({
```
baering, line 740: Won't `where.id = search` always look for an exact match, so that the result will always just be of length 1 or 0?

Would it need to be more like `where.id = { [Op.like]: `%${searchString}%` }` (or maybe we want to give the user some control of the `%` and map `*` -> `%` 🤷 ) to require pagination?

apps/services/payments/src/types/payment.ts
```diff
@@ -12,6 +12,23 @@ export enum PaymentStatus {
...
+}
+
+export enum PaymentFlowEventReason {
+  PAYMENT_STARTED = 'payment_started',
+  PAYMENT_COMPLETED = 'payment_completed',
+  PAYMENT_FAILED = 'payment_failed',
+  DELETED_ADMIN = 'deleted_admin',
+  DELETED_AUTO = 'deleted_auto',
+  OTHER = 'other',
+}
```
baering, line 30: We should either infer this from [the model](https://github.com/island-is/island.is/pull/19516/files#diff-1244aaf2a029399e9364eadd41de38a359686b8f164e235947aa50204dae07deR69) or let the model use these, right?

libs/api/domains/payments/src/lib/dto/getPaymentFlow.response.ts
```diff
@@ -55,4 +54,7 @@ export class GetPaymentFlowResponse {
 
   @Field(() => Date)
   updatedAt!: Date
+
+  @Field(() => [PaymentFlowEvent], { nullable: true })
+  events?: PaymentFlowEvent[]
```
baering: I would like `includeEvents` to be behind a scope, so that only admins can request events.

So perhaps it needs to be in a new resolver that extends `GetPaymentFlowResponse` with `events`.

apps/services/payments/src/app/paymentFlow/paymentFlow.service.ts
magnearun: Yes it will (and I think should ) result in 1 or 0, since it is the guid of the flow, if I copy a flow id and paste into the search I don't want to see flows with similar id's, since I'm searching for a specific one. The pagination is more needed when searching by nationalId, then a single user can have multiple flows

libs/api/domains/payments/src/lib/dto/getPaymentFlow.response.ts
magnearun: Yes that makes sense

libs/portals/admin/payments/src/utils/paymentStatusTag.ts
```diff
@@ -0,0 +1,43 @@
...
+          id: 'admin-portal.payments:statusUnpaid',
+          defaultMessage: 'Ógreitt',
+        }),
+      }
+    case PaymentsGetFlowPaymentStatus.invoice_pending:
+      return {
+        variant: 'blue',
+        message: formatMessage({
+          id: 'admin-portal.payments:statusInvoicePending',
+          defaultMessage: 'Reikningur í vinnslu',
```
baering: Krafa gefin út, bíður greiðslu.

apps/services/payments/src/app/app.module.ts
```diff
@@ -12,13 +14,16 @@ import { CardPaymentModule } from './cardPayment/cardPayment.module'
 import { InvoicePaymentModule } from './invoicePayment/invoicePayment.module'
 import { SequelizeConfigService } from '../sequelizeConfig.service'
 import { JwksModule } from './jwks/jwks.module'
+import { environment } from '../environments'
 
 @Module({
   imports: [
-    ProblemModule,
+    AuditModule.forRoot(environment.audit),
```
baering: Isn't this unused?

baering (approved):
LGTM ✅

robertaandersen (approved):
LGTM

PR #19569, opened 2025-07-08
fix(rental-agreement): Add assignee management to SIGNING state 

addi:
https://app.asana.com/1/203394141643832/project/1208271027457465/task/1209555373134642

## What

Introduces 'assignToInstitution' and 'clearAssignees' actions to the SIGNING state in the rental agreement template. The SIGNING state now assigns the application to a specific institution on entry and clears assignees on exit. Also removes the EDIT event transition from the SIGNING state.

## Why

So the state of the application can be changed through the api.

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * The relevant institution is now automatically assigned as an assignee when entering the signing phase of the rental agreement application.

* **Changes**
  * It is no longer possible to revert from the signing phase back to the review phase via the edit option.

### DIFF
--- a/libs/application/templates/rental-agreement/src/lib/RentalAgreementTemplate.ts
+++ b/libs/application/templates/rental-agreement/src/lib/RentalAgreementTemplate.ts
@@ -170,6 +170,8 @@ const RentalAgreementTemplate: ApplicationTemplate<
         },
       },
       [States.SIGNING]: {
+        entry: 'assignToInstitution',
+        exit: 'clearAssignees',
         meta: {
           name: States.SIGNING,
           status: 'inprogress',
@@ -197,9 +199,6 @@ const RentalAgreementTemplate: ApplicationTemplate<
           ],
         },
         on: {
-          [DefaultEvents.EDIT]: {
-            target: States.INREVIEW,
-          },
           [DefaultEvents.APPROVE]: {
             target: States.COMPLETED,
           },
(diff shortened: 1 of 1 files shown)


libs/application/templates/rental-agreement/src/lib/RentalAgreementTemplate.ts
```diff
@@ -170,6 +170,8 @@ const RentalAgreementTemplate: ApplicationTemplate<
         },
       },
       [States.SIGNING]: {
+        entry: 'assignToInstitution',
+        exit: 'clearAssignees',
```
jonnigs, line 174: Will HMS only need to do state transfers from this state? Should we clear the assignees on exit?

libs/application/templates/rental-agreement/src/lib/RentalAgreementTemplate.ts
```diff
@@ -242,6 +241,15 @@ const RentalAgreementTemplate: ApplicationTemplate<
...
       }),
+      assignToInstitution: assign((context) => {
+        const { application } = context
+
+        set(application, 'assignees', [
+          InstitutionNationalIds.HUSNAEDIS_OG_MANNVIRKJASTOFNUN,
+        ])
+
+        return context
+      }),
```
jonnigs, line 252: This overwrites the current array of assignees. Is the definitely what we want rather than add it to the list of assignees?

libs/application/templates/rental-agreement/src/lib/RentalAgreementTemplate.ts
addi: Yes, no one should do state changes after the exit.

libs/application/templates/rental-agreement/src/lib/RentalAgreementTemplate.ts
addi: Yes, no one other than HMS should change it after it has started the signing process.

jonnigs (approved):

PR #19578, opened 2025-07-09
chore(j-s): migrate current defendant verdict variables to the new verdict model 

thorhildurt:
# [Migrate old defendant verdicts](https://app.asana.com/1/203394141643832/project/1199153462262248/task/1210698669328840)

## What
- Migrate all defendant verdict data for cases that have been completed with ruling
    - Public prosecutor
    - Completing court cases in court 
- Set service_status to `NOT_APPLICABLE` to have a way to identify verdict prior to the new delivery verdicts to police feature
- Note: we still have the deprecated properties on defendant, but they will be removed in a follow-up task

## Why
- We want to use the same data model for both old and new verdicts
- We want to have a consistent way in the judicial system to manage verdicts 

## Testing
- RVG: Court
    - Create new case which ends with ruling ✅
    - Use existing case which ends with ruling ✅
    - Existing multiple defendants ✅
- RVG: Public prosecutor office
    - Handle the new case ✅
    - Use existing case ✅
    - Check case lists with column variables ✅
- Digital mailbox api
    - get verdict ✅
    - Send verdict appeal ✅

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * Introduced a nested verdict object for defendants, consolidating verdict-related data and fields.
  * Added new endpoints and controllers for handling verdict appeal decisions and restricted verdict updates.
  * Implemented new guards for validating verdict existence and enhanced route protection.

* **Refactor**
  * Shifted verdict-related fields from the defendant level to a nested verdict object across backend, API, and frontend.
  * Updated GraphQL queries and DTOs to reflect the new verdict structure.
  * Revised UI components and hooks to use verdict-level data and actions.

* **Bug Fixes**
  * Improved error handling in guards to return more appropriate HTTP status codes.
  * Enhanced state management to support partial verdict updates.

* **Chores**
  * Updated migrations and data models to support the new verdict structure.
  * Renamed message keys and properties for clarity and consistency.

* **Tests**
  * Adjusted and expanded tests to align with the new verdict data structure and guard logic.


apps/judicial-system/api/src/app/modules/defendant/models/defendant.model.ts
```diff
@@ -4,23 +4,18 @@ import {
...
   Gender,
-  InformationForDefendant,
   PunishmentType,
-  ServiceRequirement,
   SubpoenaType,
-  VerdictAppealDecision,
 } from '@island.is/judicial-system/types'
 
 import { Subpoena } from '../../subpoena'
+import { Verdict } from '../../verdict/models/verdict.model'
```
gudjong: Please use barrel file import for consistency. It should work here in the api project.

apps/judicial-system/api/src/app/modules/defendant/models/defendant.model.ts
```diff
@@ -69,21 +64,11 @@ export class Defendant {
   @Field(() => DefendantPlea, { nullable: true })
   readonly defendantPlea?: DefendantPlea
 
-  @Field(() => ServiceRequirement, { nullable: true })
```
gudjong, line 72: The fields that are removed from the model should also be removed from `updateDefendant.input.ts`.

Please consider doing that, so we do not forget.

apps/judicial-system/backend/migrations/20250708145248-update-verdict.js
```diff
@@ -0,0 +1,27 @@
+'use strict'
+
+const COMPLETED = 'COMPLETED'
+const RULING = 'RULING'
+
+module.exports = {
+  async up(queryInterface) {
+    console.log('VERDICT MIGRATION')
```
gudjong: Did you intend to include a console log?

apps/judicial-system/backend/src/app/app.module.ts
```diff
@@ -15,6 +15,7 @@ import { courtClientModuleConfig } from '@island.is/judicial-system/court-client
 import { messageModuleConfig } from '@island.is/judicial-system/message'
 
 import { LawyerRegistryModule } from './modules/lawyer-registry/lawyerRegistry.module'
+import { VerdictModule } from './modules/verdict/verdict.module'
```
gudjong: Please follow the convetion and import the verdict module in the same way as other modules are imported. Please also consider modifying the laywer registry module import above to follow the same convention.

apps/judicial-system/backend/src/app/modules/case-table/caseTable.cellGenerators.ts
```diff
@@ -961,9 +962,15 @@ const subpoenaServiceState: CaseTableCellGenerator<TagValue> = {
...
     defendants: {
       model: Defendant,
-      attributes: ['serviceRequirement', 'verdictViewDate'],
       order: [['created', 'ASC']],
       separate: true,
+      includes: {
+        verdict: {
+          model: Verdict,
+          attributes: ['serviceRequirement', 'serviceDate'],
+          order: [['created', 'ASC']],
```
gudjong: There is no need for ordering here as there is only one verdict per defendant.

apps/judicial-system/backend/src/app/modules/case-table/caseTable.cellGenerators.ts
```diff
@@ -1062,9 +1069,15 @@ const indictmentReviewDecision: CaseTableCellGenerator<TagPairValue> = {
...
     defendants: {
       model: Defendant,
-      attributes: ['verdictAppealDate'],
       order: [['created', 'ASC']],
       separate: true,
+      includes: {
+        verdict: {
+          model: Verdict,
+          attributes: ['appealDate'],
+          order: [['created', 'ASC']],
```
gudjong: No need for ordering.

apps/judicial-system/backend/src/app/modules/case-table/whereOptions/publicProsecutionOffice.ts
```diff
@@ -165,8 +170,9 @@ export const publicProsecutionOfficeIndictmentsAppealedWhereOptions = () => ({
           [Op.and]: [
             literal(`EXISTS (
               SELECT 1 FROM defendant
+              JOIN verdict ON defendant.id = verdict.defendant_id
```
gudjong: The defendant is no longer needed here, so the query can be simplified and the join eliminated.

apps/judicial-system/backend/src/app/modules/case/case.service.ts
```diff
@@ -316,6 +317,12 @@ export const include: Includeable[] = [
         order: [['created', 'DESC']],
         separate: true,
       },
+      {
+        model: Verdict,
+        as: 'verdict',
+        required: false,
+        order: [['created', 'DESC']],
```
gudjong: Ordering not needed here.

apps/judicial-system/backend/src/app/modules/case/case.service.ts
```diff
@@ -401,6 +408,12 @@ export const include: Includeable[] = [
             order: [['created', 'DESC']],
             separate: true,
           },
+          {
+            model: Verdict,
+            as: 'verdict',
+            required: false,
+            order: [['created', 'DESC']],
```
gudjong: Ordering not needed. Please consider removing here and everywhere verdict is being retrieved.

apps/judicial-system/backend/src/app/modules/case/case.service.ts
```diff
@@ -461,6 +474,12 @@ export const caseListInclude: Includeable[] = [
         order: [['created', 'DESC']],
         separate: true,
       },
+      {
+        model: Verdict,
+        as: 'verdict',
+        required: false,
+        order: [['created', 'DESC']],
```
gudjong: Ordering not needed.

apps/judicial-system/backend/src/app/modules/case/guards/test/indictmentCaseExistsForDefendantGuard.spec.ts
```diff
@@ -82,6 +83,12 @@ describe('Indictment Case Exists For Defendant Guard', () => {
                 as: 'subpoenas',
                 order: [['created', 'DESC']],
               },
+              {
+                model: Verdict,
+                as: 'verdict',
+                required: false,
+                order: [['created', 'DESC']],
```
gudjong: Please remove ordering.

apps/judicial-system/backend/src/app/modules/case/internalCase.service.ts
```diff
@@ -1336,6 +1337,12 @@ export class InternalCaseService {
               as: 'subpoenas',
               order: [['created', 'DESC']],
             },
+            {
+              model: Verdict,
+              as: 'verdict',
+              required: false,
+              order: [['created', 'DESC']],
```
gudjong: Please remove ordering.

apps/judicial-system/backend/src/app/modules/case/limitedAccessCase.service.ts
```diff
@@ -194,6 +195,12 @@ export const include: Includeable[] = [
         order: [['created', 'DESC']],
         separate: true,
       },
+      {
+        model: Verdict,
+        as: 'verdict',
+        required: false,
+        order: [['created', 'DESC']],
```
gudjong: Please remove ordering.

apps/judicial-system/backend/src/app/modules/verdict/filters/verdict.filter.ts
```diff
@@ -16,6 +20,9 @@ export const canUserEditVerdict = (
   if (isDistrictCourtUser(user)) {
     return canDistrictCourtUserEditVerdict(theCase)
   }
+  if (isPublicProsecutionOfficeUser(user)) {
+    return canPublicProsecutionUserAccessCase(theCase)
```
gudjong: This is a bit inconsistent. Here, we are using the proper access check for public prosecution users, but above we are using an incomplete check for disctrict court users where we only check if the case is completed even thought district court users should only have access to cases which are assigned to their court.

Although this may not make much difference where this method is used (presumably because the proper access has already been checked elsewhere) please consider being more consistent when checking if users can edit the verdict. Otherwise, this may become confucing and difficult to maintain later on.

apps/judicial-system/backend/src/app/modules/verdict/internalVerdict.controller.ts
```diff
@@ -0,0 +1,122 @@
...
+  CaseCompletedGuard,
+  CaseExistsGuard,
+  CaseTypeGuard,
+  CurrentCase,
+} from '../case'
+import { CurrentDefendant, Defendant } from '../defendant'
+import { DefendantNationalIdExistsGuard } from '../defendant/guards/defendantNationalIdExists.guard'
+import { Verdict } from '../verdict/models/verdict.model'
+import { VerdictService } from '../verdict/verdict.service'
+import { UpdateVerdictAppealDto } from './dto/updateVerdictAppeal.dto'
```
gudjong: Please consider using `InternalUpdateVerdictDto` instead. These are realy identical dtos'

apps/judicial-system/backend/src/app/modules/verdict/test/verdictControllerGuards.spec.ts
```diff
@@ -10,8 +10,7 @@ describe('VerdictController - Top-level guards', () => {
   verifyGuards(VerdictController, undefined, [
     JwtAuthUserGuard,
     RolesGuard,
-    MinimalCaseExistsGuard,
-    MinimalCaseAccessGuard,
+    CaseExistsGuard,
```
gudjong: Since you have reverted to using `CaseExistsGuard` you can now revert to using the standard guard pattern. `VerdictWriteGuard` is no longer needed and instaed you can use `CaseReadGuard` and /or `CaseWriteGuard` and `CaseCompletedGuard`.

Also note that `ValidateVerdictGuard` would be better implemented as `VerdictExistsGuard` following the standard guard pattern. You would then use `DefendantExistsGuard` before `VerdictExistsGuard` and instead of querying the database for the verdict simply look it up in the existing defendant.

apps/judicial-system/backend/src/app/modules/verdict/verdict.controller.ts
```diff
@@ -38,13 +34,7 @@ import { VerdictService } from './verdict.service'
...
 @Controller('api/case/:caseId/defendant/:defendantId/verdict')
 @ApiTags('verdicts')
-@UseGuards(
-  JwtAuthUserGuard,
-  RolesGuard,
-  MinimalCaseExistsGuard,
-  MinimalCaseAccessGuard,
-  ValidateVerdictGuard,
-)
+@UseGuards(JwtAuthUserGuard, RolesGuard, CaseExistsGuard, ValidateVerdictGuard)
```
gudjong: Please refer to the previous comments about guards in the internal verdict controller. Since you revert to the regular `CaseExistsGuard` it would be more appropriate to follow the standard guard pattern, thus avoiding extra database queries and reimplementing user access control

libs/judicial-system/types/src/lib/defendant.ts
```diff
@@ -34,6 +34,7 @@ export enum ServiceStatus {
   IN_PERSON = 'IN_PERSON',
   FAILED = 'FAILED',
   EXPIRED = 'EXPIRED', // If a subpoena expires
+  NOT_APPLICABLE = 'NOT_APPLICABLE', // Note: Only used for old verdicts prior to delivering verdict to police feature
```
gudjong: Would `NOT_SERVED` perhaps be a better name?

gudjong (changes requested):
Some comments and canges requested in a few places.

I consentrated on the api and backend. It would be good if someone else could review other service and the front end.

apps/judicial-system/api/src/app/modules/defendant/models/defendant.model.ts
thorhildurt: Thanks, missed this model

apps/judicial-system/backend/migrations/20250708145248-update-verdict.js
thorhildurt: No, this was used for debugging and testing so will remove

apps/judicial-system/backend/src/app/modules/case-table/whereOptions/publicProsecutionOffice.ts
thorhildurt: Is that because of the the includes in cellGenerators? We should then be able to reference the verdict fields via defendant.{x}?

libs/judicial-system/types/src/lib/defendant.ts
thorhildurt: I think that indicates that they were not served at all. The passed verdicts probably were served manually by RLS and its almost impossible to know the actually status of those. So its true that they were not served via Birtingur or electronically. Maybe `NOT_SERVED_VIA_SYSTEM` gives more clarity, but I think I will just keep `NOT_APPLICABLE` which should indicate that for the old records status is not relevant.

apps/judicial-system/backend/src/app/modules/verdict/test/verdictControllerGuards.spec.ts
thorhildurt: But I wonder if I don't use `VerdictWriteGuard`, and which relies on `canUserAccessCase` via `CaseReadGuard`, are we then only relying on the `RolesRules` for specifying which role can update? 
Since in `canUserAccessCase` in `CaseReadGuard`, we technically support all roles (due to condition for each role) 🤔 Is that the expected design pattern to have the access case guard more generic and rely more on role rules to filter the roles?

EDIT (for my learning): In `CaseReadGuard` => `canUserAccessCase` => `canDistrictCourtUserAccessCase` for indictment cases, we allow more states which probably can solved with the intersection of the guards in combination with `CaseCompletedGuard`, as you mentioned in your feedback, right?

apps/judicial-system/backend/src/app/modules/verdict/guards/verdictExistGuard.guard.ts
```diff
@@ -0,0 +1,33 @@
...
+  CanActivate,
+  ExecutionContext,
+  Injectable,
+  NotFoundException,
+} from '@nestjs/common'
+
+@Injectable()
+export class VerdictExistGuard implements CanActivate {
+  async canActivate(context: ExecutionContext): Promise<boolean> {
+    const request = context.switchToHttp().getRequest()
```
gudjong, line 12: You can skip this part. The case is verified in another guard and no need to repeat it.

apps/judicial-system/backend/src/app/modules/verdict/internalVerdict.controller.ts
```diff
@@ -0,0 +1,121 @@
...
+      `Appeal deadline has passed for case ${caseId}. Deadline was ${appealDeadline.toISOString()}`,
+    )
+  }
+}
+
+@Controller('api/internal/case/:caseId')
+@ApiTags('internal verdict')
+@UseGuards(
+  TokenGuard,
+  MinimalCaseExistsGuard,
```
gudjong: This must be `CaseExistsGuard` for the subsequent defendant an verdict guards to work.

apps/judicial-system/backend/src/app/modules/verdict/internalVerdict.controller.ts
```diff
@@ -0,0 +1,121 @@
...
+  new CaseTypeGuard(indictmentCases),
+  CaseCompletedGuard,
+)
+export class InternalVerdictController {
+  constructor(
+    private readonly verdictService: VerdictService,
+    @Inject(LOGGER_PROVIDER) private readonly logger: Logger,
+  ) {}
+
+  @UseGuards(DefendantNationalIdExistsGuard, VerdictExistGuard)
```
gudjong: Since this is the only method in the controller, you could move these guards to the controller level after the other controller level guards. I don't have a strong opinion on this though.

apps/judicial-system/backend/src/app/modules/verdict/verdict.controller.ts
```diff
@@ -16,69 +16,76 @@ import {
...
 
-  @UseGuards(VerdictWriteGuard)
   @RolesRules(
     districtCourtJudgeRule,
     districtCourtRegistrarRule,
     districtCourtAssistantRule,
+    publicProsecutorStaffRule,
   )
-  @Patch(':verdictId')
+  @Patch(':defendantId')
```
gudjong: This works, but I would probably do .../defendant/:defendantId/verdict. That is more consistent with paths used in other controllers, for example the case controller. Still, I don't have a strong opinion.

gudjong (changes requested):
One more change needed regarding guards in the internal verdict controller.
Some minor other comments and suggestions.
Otherwise, the backend is looking really nice.

apps/judicial-system/backend/src/app/modules/verdict/test/verdictControllerGuards.spec.ts
gudjong: This is the basic idea in the standard pattern:

1. Authenticate the caller: `JwtAuthUserGuard` for user facing endpoints and `TokenGuard` for internal endpoints.
2. Make sure the current user's role has access to the endpoint: `RolesGuard`
3. Make sure the case exists: `CaseExistsGuard`.
4. Make sure the current user can read or write the case: `CaseReadGuard` or `CaseWriteGuard`.
5. Other case guards, such as guards that make sure the case is in the correct state.
6. Exists guards for sub elements, in your example first `DefendantExists` and the `VerdictExists`.

There is some overlap in these guards, for example the roles guards for the endpoint and the access guards for the case. Read and write access is only checked on the case level with the read and write guards and could be more granular.

This is all open to change if we feel it is needed, but we should have some common patterns so it is easier to maintain and harder to introduce security issues.

Note that in your case you should strictly speaking be using `CaseWriteGuard` in the update endpoint since you are updating the case. However, for district court and public prosecutor office users, the read and write guards are equivelent because they can write all cases they can read. For consistency, I suggest you use `CaseWriteGuard`.

thorhildurt: Thanks you @gudjong for outlining this so clearly 🙏 , will extract the standard pattern to my own documentation. This is helpful to understand better also in contrast to the new pattern we have for example in the victim endpoint where we experimented with only fetching the minimal case.

apps/judicial-system/backend/src/app/modules/verdict/internalVerdict.controller.ts
thorhildurt: Ah yes of course, we need that for the DefendantNationalIdExistsGuard when filtering the target defendant. Good catch.

apps/judicial-system/backend/src/app/modules/verdict/internalVerdict.controller.ts
thorhildurt: Will move since it is very likely we will need this for all future updates via this controller 👍

apps/judicial-system/backend/src/app/modules/verdict/verdict.controller.ts
thorhildurt: Will make it consistent 👍

apps/judicial-system/backend/migrations/20250708145248-update-verdict.js
```diff
@@ -0,0 +1,26 @@
...
+
+module.exports = {
+  async up(queryInterface) {
+    return queryInterface.sequelize.transaction((t) =>
+      queryInterface.sequelize.query(
+        `INSERT INTO "verdict" (id, defendant_id, case_id, service_requirement, service_status, service_date, appeal_decision, appeal_date, service_information_for_defendant)
+                    SELECT md5(random()::text || clock_timestamp()::text)::uuid, d.id, d.case_id, d.service_requirement, 'NOT_APPLICABLE', d.verdict_view_date, d.verdict_appeal_decision, d.verdict_appeal_date, d.information_for_defendant
+                    FROM "case" as c JOIN "defendant" as d ON d.case_id = c.id 
+                    WHERE c.state = '${COMPLETED}' AND c.indictment_ruling_decision = '${RULING}'`,
+        {
```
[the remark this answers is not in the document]
thorhildurt: Makes sense, but why are you showing diff for lines that do not change at all for example int he up method? I see that the where condition is different and the down

apps/judicial-system/backend/src/app/modules/verdict/test/verdictControllerGuards.spec.ts
```diff
@@ -1,21 +1,42 @@
...
-import { ValidateVerdictGuard } from '../guards/validateVerdict.guard'
-import { VerdictWriteGuard } from '../guards/verdictWrite.guard'
+import {
+  CaseCompletedGuard,
+  CaseExistsGuard,
+  CaseTypeGuard,
+  CaseWriteGuard,
+} from '../../case'
+import { DefendantExistsGuard } from '../../defendant'
+import { VerdictExistGuard } from '../guards/verdictExistGuard.guard'
```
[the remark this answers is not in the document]
thorhildurt: I think this is done in a follow-up PR

apps/judicial-system/web/src/components/FormProvider/limitedAccessCase.graphql
```diff
@@ -64,6 +62,14 @@ query LimitedAccessCase($input: CaseQueryInput!) {
...
         policeSubpoenaId
       }
+      verdict {
+        id
+        appealDate
+        appealDecision
+        serviceRequirement
+        serviceDate
+        serviceInformationForDefendant
+      }
```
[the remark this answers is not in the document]
thorhildurt: I am not sure all of these are relevant 🤔 For example I don't see any top-level fields defined via case.graphql

apps/judicial-system/web/src/components/InfoCard/DefendantInfo/DefendantInfo.tsx
```diff
@@ -104,10 +104,14 @@ export const DefendantInfo: FC<DefendantInfoProps> = (props) => {
         </Text>
       )}
       {displayVerdictViewDate &&
-        defendant.serviceRequirement &&
-        defendant.serviceRequirement !== ServiceRequirement.NOT_REQUIRED && (
+        defendant.verdict?.serviceRequirement &&
+        defendant.verdict?.serviceRequirement !==
+          ServiceRequirement.NOT_REQUIRED && (
```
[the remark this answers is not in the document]
thorhildurt: That is expected, serviceDate is set when serviceRequirement is not applicaple

apps/judicial-system/web/src/utils/hooks/useVerdict/updateVerdict.graphql
```diff
@@ -1,5 +1,5 @@
 mutation UpdateVerdict($input: UpdateVerdictInput!) {
   updateVerdict(input: $input) {
-    id
+    defendantId
```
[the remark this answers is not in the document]
thorhildurt: We only rely on the defendantId, since a unique entry in the verdict table exists per defendantId

libs/judicial-system/types/src/lib/defendant.ts
```diff
@@ -34,6 +34,7 @@ export enum ServiceStatus {
   IN_PERSON = 'IN_PERSON',
   FAILED = 'FAILED',
   EXPIRED = 'EXPIRED', // If a subpoena expires
+  NOT_APPLICABLE = 'NOT_APPLICABLE', // Note: ONLY used for old verdicts prior to delivering verdicts automatically to the police
```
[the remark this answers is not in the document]
thorhildurt: NOT_APPLICABLE in this context is to distinguish migrated verdicts from new ones. So it makes sense it is not used anywhere except if we need to roll back the migration or if we need to do further statistical analaysis in the future. We can know Service status for old verdicts.

apps/judicial-system/backend/src/app/modules/case/internalCase.service.ts
```diff
@@ -1338,6 +1339,11 @@ export class InternalCaseService {
               as: 'subpoenas',
               order: [['created', 'DESC']],
             },
+            {
+              model: Verdict,
+              as: 'verdict',
+              required: false,
+            },
```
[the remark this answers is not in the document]
thorhildurt: Wait I do have   
`@HasOne(() => Verdict, { foreignKey: 'defendantId' })`
 on my defendant.model.ts 🤔

PR #19609, opened 2025-07-16
feat(j-s): Deliver verdict document to police

thorhildurt:
# [Deliver verdict document to police to be serviced ](https://app.asana.com/1/203394141643832/project/1199153462262248/task/1210583209106869)

## What
- Implement the endpoint we will trigger when a verdict should be delivered to police to be served
- Use a new RLS endpoint `CreateDocument` to send only the **required** data to police. We will add the supplement data in a follow-up PR. 

## Why
- To be enable the police entity to forward verdicts that need to be served to defendants via Birtingur or electronically via island.is

## Screenshots
- Successful response and given verdict is update as expected 🥳 

Swagger
Screenshot 2025-07-25 at 15 39 23

Database
Screenshot 2025-07-25 at 15 40 00

## Checklist:

## Summary by CodeRabbit

- New Features
  - Deliver verdicts to the National Commissioner’s Office and receive a delivery status response.
  - Audit trail now records verdict delivery events.

- Refactor
  - Standardized guard naming for verdict checks.

- Tests
  - Expanded tests to cover verdict delivery flow, audit logging, and new service integrations.

- Chores
  - Simplified subpoena delivery routing.
  - Minor comment and documentation cleanup.


libs/judicial-system/message/src/lib/message.ts
```diff
@@ -95,6 +96,8 @@ export const messageEndpoint: { [key in MessageType]: string } = {
   CIVIL_CLAIMANT_NOTIFICATION: 'civilClaimantNotification',
   INDICTMENT_CASE_NOTIFICATION: 'indictmentCaseNotification',
   EVENT_NOTIFICATION_DISPATCH: 'eventNotification/dispatch',
+  DELIVER_VERDICT_TO_NATIONAL_COMMISSIONERS_OFFICE:
```
gudjong: Please use the established nameing convention: `DELIVER_TO_NATIONAL_COMMISSIONERS_OFFICE_VERDICT`. Also, please group related constants together to make it easy to find them.

apps/judicial-system/backend/src/app/modules/verdict/dto/updateVerdictExternalPoliceDocumentId.dto.ts
```diff
@@ -0,0 +1,10 @@
+import { IsNotEmpty, IsString } from 'class-validator'
+
+import { ApiProperty } from '@nestjs/swagger'
+
+export class UpdateVerdictExternalPoliceDocumentIdDto {
```
gudjong: It feels like overkill to create a separate dto for this as it is not really used as a dto but only a type in an internal update method. I see what you are trying to accomplish with this, but there are other ways to do it. One way is used in the case module, where we define a type `UpdateCase` in the case service to determine which fields can be modified in internal methods. Once defind we can pass dtos or custom objects containing mutable fields.

apps/judicial-system/backend/src/app/modules/verdict/guards/verdictExists.guard.ts
```diff
@@ -0,0 +1,48 @@
...
+      throw new BadRequestException('Missing defendant id')
+    }
+
+    const verdictId = request.params.verdictId
+
+    if (!verdictId) {
+      throw new BadRequestException('Missing verdict id')
+    }
+
+    const verdict = theCase.defendants?.find(
```
gudjong: As mentioned in a review to a previous PR, it would make sense to the guard `DefendantExists` before this guard and then use `request.defendant` to refer to the defendant just like you refer to the case.

apps/judicial-system/backend/src/app/modules/verdict/internalVerdict.controller.ts
```diff
@@ -78,9 +91,69 @@ const validateVerdictAppealUpdate = ({
 export class InternalVerdictController {
   constructor(
     private readonly verdictService: VerdictService,
+    private readonly auditTrailService: AuditTrailService,
     @Inject(LOGGER_PROVIDER) private readonly logger: Logger,
   ) {}
 
+  @UseGuards(
+    new CaseTypeGuard(indictmentCases),
+    DefendantExistsGuard,
```
gudjong: Nice.

apps/judicial-system/backend/src/app/modules/verdict/verdict.service.ts
```diff
@@ -10,16 +12,56 @@ import { InjectModel } from '@nestjs/sequelize'
...
+    @Inject(forwardRef(() => FileService))
+    private readonly fileService: FileService,
+    @Inject(forwardRef(() => PoliceService))
+    private readonly policeService: PoliceService,
     @Inject(LOGGER_PROVIDER) private readonly logger: Logger,
   ) {}
 
+  private async update(
+    verdictId: string,
+    update: UpdateVerdictDto | UpdateVerdictExternalPoliceDocumentIdDto,
```
gudjong: Please consider defining a new `UpdateVerdict` type rather than combining dto types like this.

apps/judicial-system/backend/src/app/modules/verdict/dto/updateVerdictExternalPoliceDocumentId.dto.ts
thorhildurt: Ah I get your point 💡 I will remove this dto and rather define a clearer response type for the create document in the police service. Then I will have a service based type as you suggest which works both for client based and internal updates based on the Verdict model schema 👍

apps/judicial-system/backend/src/app/modules/case-table/caseTable.cellGenerators.ts
```diff
@@ -977,12 +977,15 @@ const subpoenaServiceState: CaseTableCellGenerator<TagValue> = {
       return generateCell()
     }
 
-    const verdictInfo = c.defendants?.map<[boolean, Date | undefined]>((d) => [
-      true,
-      d.verdict?.serviceRequirement === ServiceRequirement.NOT_REQUIRED
-        ? c.rulingDate
-        : d.verdict?.serviceDate,
-    ])
+    // TODO: fix this in the database so we can always fetch the service date
```
thorhildurt, line 980: https://app.asana.com/1/203394141643832/project/1199153462262248/task/1210887973090543

apps/judicial-system/backend/src/app/modules/police/models/createDocument.response.ts
```diff
@@ -0,0 +1,6 @@
+import { ApiProperty } from '@nestjs/swagger'
+
+export class CreateDocumentResponse {
```
gudjong: Since we are not being used in any controller endpoint, this could just as easily be an interface defined in and exported from `police.service.ts`. This works though, so you decide.

apps/judicial-system/backend/src/app/modules/verdict/models/verdict.model.ts
```diff
@@ -83,7 +83,7 @@ export class Verdict extends Model {
 
   @Column({ type: DataType.DATE, allowNull: true })
   @ApiPropertyOptional({ type: Date })
-  serviceDate?: Date // birtingardagsetningin
+  serviceDate?: Date | null
```
gudjong: This should not be necessary. All optional fields in our backend models are nullable, but we never explicitly define them with `| null`.

libs/judicial-system/types/src/lib/indictmentCase.ts
```diff
@@ -9,7 +9,7 @@ import { getIndictmentAppealDeadlineDate, hasDatePassed } from './dates'
   - The second boolean indicates whether all defendant appeal deadlines have expired.
 */
 export const getIndictmentVerdictAppealDeadlineStatus = (
-  verdictInfo?: [boolean, Date | undefined][],
+  verdictInfo?: [boolean, Date | undefined | null][],
```
gudjong: I don't think null is needed here if you remove the `null` part from the verdict model (see previous comment).

apps/judicial-system/backend/src/app/modules/verdict/models/verdict.model.ts
thorhildurt: Yes, exactly but somehow when assigning the `serviceDate = null`, typescript complains which doesn´t make sense according to the column declaration 🤔 I will investigate a bit further, but chatti was also not able to solve this without setting `| null` in the prop definition.

thorhildurt: I rather moved the null type definition to the interface in the verdict service to avoid adding the null typing in several places.
But this seems to be an issue with TS type system because it interprets serviceDate only as possibly `undefined`, but not as `null` although it is specified in the decorators.

PR #19613, opened 2025-07-17
feat(j-s): update digital mailbox response for verdict and case

thorhildurt:
# [Set hasRulingBeenServed in case response](https://app.asana.com/1/203394141643832/project/1199153462262248/task/1210583209106877)
# [Set labels for instructions for defendant from court](https://app.asana.com/1/203394141643832/project/1199153462262248/task/1210799630659732)

## What
- hasRulingBeenServed was hardcoded, now it is set according to defendant verdict status
- Removed the harcoded string in the verdict response and set labels for verdict information for defendant based on what we currently show in RVG

## Why
- Because we don't want hardcoded values which were just set as poc

## Checklist:


apps/judicial-system/digital-mailbox-api/src/app/modules/cases/models/case.response.ts
```diff
@@ -43,16 +43,19 @@ export class CaseResponse {
     )
     const subpoenaCreatedDate = subpoenaDateLog?.created?.toString() ?? '' //TODO: Change to created from subpoena db entry?
     const subpoenas = defendant.subpoenas ?? []
+    const verdict = defendant.verdict
 
     return {
       caseId: internalCase.id,
       data: {
         caseNumber: `${t.caseNumber} ${internalCase.courtCaseNumber}`,
+        // TODO: Rename to hasSubpoenaBeenServed?
```
thorhildurt, line 52: Just added these TODOs as a reminders

libs/judicial-system/types/src/lib/verdict.ts
```diff
@@ -0,0 +1,89 @@
...
+  CONDITIONAL_SENTENCE_AND_BREACH_OF_PROBATION_TRANSLATION = 'CONDITIONAL_SENTENCE_AND_BREACH_OF_PROBATION_TRANSLATION',
+  DRIVING_RIGHTS_REVOKED_TRANSLATION = 'DRIVING_RIGHTS_REVOKED_TRANSLATION',
+  ALTERNATIVE_FINES_TRANSLATION = 'ALTERNATIVE_FINES_TRANSLATION',
+  COMMUNITY_SERVICE = 'COMMUNITY_SERVICE',
+  FINES_AND_COSTS = 'FINES_AND_COSTS',
+  ITEM_CONFISCATION = 'ITEM_CONFISCATION',
+}
+
+// TODO: set final descriptions
+export const InformationForDefendantMap: Map<
```
thorhildurt: This strictly speaking isn't a type but added here since this is used both in js client and in digital mailbox as a source of truth

apps/judicial-system/digital-mailbox-api/src/app/modules/cases/models/case.response.ts
```diff
@@ -43,16 +43,19 @@ export class CaseResponse {
...
         hasBeenServed:
           subpoenas.length > 0
-            ? isSuccessfulServiceStatus(subpoenas[0].serviceStatus)
+            ? // TODO: Don't we need to adjust the subpoena checked here in case of multiple subpoenas issued?
```
gudjong: A new subpoena invalidates all previous subpoenas, so only the latest subpoena is valid. I suppose that is the reason you only need to check the lates subpoena on the case lever.

libs/judicial-system/types/src/lib/verdict.ts
gudjong: Please consider using a name starting with a lower case letter as that is our convention.

apps/judicial-system/digital-mailbox-api/src/app/modules/cases/models/case.response.ts
thorhildurt: Ah and the latest is at the top since we order the subpoenas in descending order, got it 👍

PR #19667, opened 2025-07-25
chore(j-s): Event Log Improvements

gudjong:
# Event Log Improvements

[Event log cleanup](https://app.asana.com/1/203394141643832/project/1199153462262248/task/1210877803206173)

## What

- No longer passes event logs to the front end.
- Extracts relevant dates from the event log and passes them to the front end.
- Improves database transaction handling.

## Why

- Cleaner and more readable code.

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * Added new date fields to case data, including indictment review and appeal result access dates for different user roles.
* **Bug Fixes**
  * Improved accuracy and consistency of event date retrieval throughout the system.
* **Refactor**
  * Simplified and unified event log handling, removing redundant fields and logic from case data and GraphQL queries.
  * Centralized event log date retrieval using new helper methods supporting multiple event types and user roles.
  * Streamlined transaction management for defendant and event log updates.
  * Enhanced modularity by splitting event log update logic into specialized methods.
  * Replaced inline event log filtering with encapsulated utility calls across components and services.
  * Updated case and defendant event log date retrieval to use generalized methods supporting multiple event types and user roles.
  * Removed conditional appeal state and appeal date adjustments from case list processing.
  * Simplified court case number display and removed prosecutor text in tables.
  * Refined appeal result access display with a dedicated component.
  * Improved event log creation to respect transaction boundaries ensuring atomic updates.
  * Removed event log data from GraphQL queries and replaced with direct date fields.
  * Adjusted appeal withdraw option logic to depend solely on case state and appeal date.
* **Chores**
  * Updated import statements for Sequelize usage.
  * Removed unused or redundant fields from API responses and models.
  * Improved test setups with transaction mocking for database operations.


apps/judicial-system/backend/src/app/modules/case/interceptors/completedAppealAccessed.interceptor.ts
```diff
@@ -20,25 +22,31 @@ import { EventLogService } from '../../event-log'
...
+    private readonly eventLogService: EventLogService,
+  ) {}
 
   intercept(context: ExecutionContext, next: CallHandler) {
     const request = context.switchToHttp().getRequest()
     const user: User = request.user?.currentUser
 
     return next.handle().pipe(
-      map((data) => {
+      map(async (data) => {
```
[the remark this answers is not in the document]
gudjong: I have fixed this.

apps/judicial-system/backend/src/app/modules/case/interceptors/completedAppealAccessed.interceptor.ts
```diff
@@ -20,25 +22,31 @@
...
-            user,
-          )
+          await this.sequelize.transaction(async (transaction) => {
+            this.eventLogService.createWithUser(
+              EventType.APPEAL_RESULT_ACCESSED,
+              data.id,
+              user,
+              transaction,
+            )
+          })
```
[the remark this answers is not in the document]
gudjong: I will add a return instead.

apps/judicial-system/backend/src/app/modules/defendant/defendant.service.ts
```diff
@@ -282,41 +284,46 @@ export class DefendantService {
...
+        if (update.isSentToPrisonAdmin) {
+          this.createDefendantEvent(
+            {
+              caseId: theCase.id,
+              defendantId: defendant.id,
+              eventType: DefendantEventType.SENT_TO_PRISON_ADMIN,
+            },
+            transaction,
+          )
+        }
```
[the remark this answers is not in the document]
gudjong: Fixed.

apps/judicial-system/backend/src/app/modules/defendant/models/defendantEventLog.model.ts
```diff
@@ -20,14 +20,33 @@ import { Defendant } from './defendant.model'
...
-  static getDefendantEventLogTypeDate(
+  static getEventLogByEventType(
     eventType: DefendantEventType,
-    defendantEventLogs?: DefendantEventLog[],
-  ) {
-    return defendantEventLogs?.find(
-      (defendantEventLog) => defendantEventLog.eventType === eventType,
-    )?.created
+    eventLogs: DefendantEventLog[] | undefined,
+  ): DefendantEventLog | undefined {
```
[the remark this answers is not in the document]
gudjong: Good catch. I have fixed this.

apps/judicial-system/backend/src/app/modules/defendant/models/defendantEventLog.model.ts
```diff
@@ -20,14 +20,33 @@
...
+      return undefined
+    }
+
+    return relevantLogs[0]
+  }
+
+  static getEventLogDateByEventType(
+    eventType: DefendantEventType,
+    eventLogs: DefendantEventLog[] | undefined,
+  ): Date | undefined {
```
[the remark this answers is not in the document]
gudjong: Fixed

apps/judicial-system/backend/src/app/modules/event-log/models/eventLog.model.ts
```diff
@@ -18,12 +18,36 @@ import { Case } from '../../case/models/case.model'
...
+        eventTypes.includes(eventLog.eventType) &&
+        (!userRole || eventLog.userRole === userRole),
     )
+
+    if (relevantLogs.length === 0) {
+      return undefined
+    }
+
+    return relevantLogs[0]
+  }
```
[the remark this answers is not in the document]
gudjong: That was my intention. Thanks for pointing this out.

apps/judicial-system/backend/src/app/modules/defendant/defendant.service.ts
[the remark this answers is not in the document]
gudjong: I have fixed this.

apps/judicial-system/backend/src/app/modules/defendant/models/defendantEventLog.model.ts
```diff
@@ -20,14 +20,33 @@ import { Defendant } from './defendant.model'
...
+      .filter((eventLog) => eventTypes.includes(eventLog.eventType))
+      // Don't assume any ordering
+      .sort((a, b) => b.created.getTime() - a.created.getTime())
+
+    if (relevantLogs.length === 0) {
+      return undefined
+    }
+
+    return relevantLogs[0]
+  }
```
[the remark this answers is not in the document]
gudjong: I have fixed this.

apps/judicial-system/backend/src/app/modules/event-log/models/eventLog.model.ts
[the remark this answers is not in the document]
gudjong: I have added this.

apps/judicial-system/api/src/app/modules/case-list/interceptors/caseList.interceptor.ts
```diff
@@ -48,34 +27,10 @@ export class CaseListInterceptor implements NestInterceptor {
...
-            indictmentRulingDecision,
-            rulingDate,
-            defendants,
-          })
-
-          return {
-            ...theCase,
-            ...indictmentInfo,
-          }
+          return theCase
```
gudjong, line 33: This intercepter is deprecated, is only used for defender tables and will be removed when defender tables have been move to the new table architecture. This PR removes some code that is not needed for the current defender table.

apps/judicial-system/api/src/app/modules/case-list/models/caseList.model.ts
```diff
@@ -1,44 +1,35 @@
 import { Field, ID, ObjectType } from '@nestjs/graphql'
 
 import {
-  CaseAppealDecision,
```
gudjong, line 4: This model is deprecated like the case list interceptor. This PR removes some fields that are not needed for the current defender tables. The code will be removed when the defender tables have been moved to the new table architecture.

apps/judicial-system/api/src/app/modules/case/models/case.model.ts
```diff
@@ -27,7 +27,6 @@ import {
 } from '@island.is/judicial-system/types'
 
 import { CivilClaimant, Defendant } from '../../defendant'
-import { EventLog } from '../../event-log'
```
gudjong, line 30: Event logs are no longer available to the frontend. Instead, select date fields have been added.

apps/judicial-system/backend/src/app/formatters/fineSentToPrisonAdminPdf.ts
```diff
@@ -64,19 +66,15 @@ export const createFineSentToPrisonAdminPdf = (
...
-        eventLog.eventType === DefendantEventType.SENT_TO_PRISON_ADMIN,
-    )
-    .sort(
-      (a, b) => new Date(b.created).getTime() - new Date(a.created).getTime(),
-    )[0]?.created
-
-  const getSignatureDate = theCase.eventLogs?.find(
-    (eventLog) => eventLog.eventType === EventType.INDICTMENT_REVIEWED,
-  )?.created
+  const sentToPrisonAdminDate = DefendantEventLog.getEventLogDateByEventType(
```
gudjong, line 69: Event logs in the backend are now accessed via helper methods in the event log classes.

apps/judicial-system/backend/src/app/modules/case/case.service.ts
```diff
@@ -205,7 +205,6 @@ export interface UpdateCase
   courtDate?: UpdateDateLog
   postponedIndefinitelyExplanation?: string
   civilDemands?: string
-  caseSentToCourtDate?: string
```
gudjong, line 208: This is not an updateable field.

apps/judicial-system/backend/src/app/modules/case/case.service.ts
```diff
@@ -312,7 +311,6 @@ export const include: Includeable[] = [
         as: 'eventLogs',
         required: false,
         where: { eventType: defendantEventTypes },
-        order: [['created', 'DESC']],
```
gudjong, line 315: We no longer order event logs when reading from the database. Sorting is perfomed in the helper methods in the event log classes. That is a more robust implementation.

apps/judicial-system/backend/src/app/modules/case/case.service.ts
```diff
@@ -1652,19 +1647,6 @@ export class CaseService {
...
         updatedArraignmentDate.date.getTime() !==
           arraignmentDate?.date.getTime()
-      const courtDate = DateLog.courtDate(theCase.dateLogs)
-      const updatedCourtDate = DateLog.courtDate(updatedCase.dateLogs)
-      const hasUpdatedCourtDate =
-        updatedCourtDate &&
-        updatedCourtDate.date.getTime() !== courtDate?.date.getTime()
-
-      if (hasUpdatedArraignmentDate || hasUpdatedCourtDate) {
-        await this.eventLogService.createWithUser(
```
gudjong, line 1662: Event log creation should be done in the same database transaction that is used for case updates. Therefore, this code has been moved to a more appropriate place.

apps/judicial-system/backend/src/app/modules/case/case.service.ts
```diff
@@ -1691,14 +1673,19 @@ export class CaseService {
     )
   }
 
-  async findById(caseId: string, allowDeleted = false): Promise<Case> {
+  async findById(
+    caseId: string,
+    allowDeleted = false,
+    transaction?: Transaction,
```
gudjong, line 1679: When getting a case from the database after an update we should read the case in the same transaction as the updae. This ensures that subsequent comparisons between the old and new case versions are actually checking the current changes and not changes for other concurrent changes.

apps/judicial-system/backend/src/app/modules/case/case.service.ts
```diff
@@ -1975,18 +2045,18 @@ export class CaseService {
     const schedulingNewArraignmentDateForIndictmentCase =
       isIndictmentCase(theCase.type) && Boolean(updatedArraignmentDate)
 
+    if (isReceivingCase) {
```
gudjong, line 2048: This code does not belong within the database transaction.

apps/judicial-system/backend/src/app/modules/case/case.service.ts
```diff
@@ -2323,16 +2402,11 @@ export class CaseService {
...
-
-    if (isIndictmentCase(theCase.type)) {
-      receivalDate =
-        theCase.eventLogs?.find(
-          (eventLog) => eventLog.eventType === EventType.INDICTMENT_CONFIRMED,
-        )?.created ?? nowFactory()
-    } else {
-      receivalDate = nowFactory()
-    }
+    const receivalDate =
```
gudjong, line 2405: The receival date was wrong for request cases.

apps/judicial-system/backend/src/app/modules/case/interceptors/completedAppealAccessed.interceptor.ts
```diff
@@ -20,25 +22,31 @@ import { EventLogService } from '../../event-log'
...
           (isProsecutionUser(user) ||
             isDefenceUser(user) ||
             isPrisonStaffUser(user))
         ) {
-          this.eventLogService.createWithUser(
-            EventType.APPEAL_RESULT_ACCESSED,
-            data.id,
-            user,
-          )
+          await this.sequelize.transaction(async (transaction) => {
```
gudjong, line 42: We wont this to happen in a database transaction to guarantee that we do not unintentionally create duplicate events. An even better approach would be to  use a transaction from the call being intercepted, but we do not have support for that at the moment.

apps/judicial-system/backend/src/app/modules/defendant/defendant.service.ts
```diff
@@ -282,41 +284,46 @@ export class DefendantService {
...
     defendant: Defendant,
     update: UpdateDefendantDto,
     user: User,
   ): Promise<Defendant> {
-    const updatedDefendant = await this.updateDatabaseDefendant(
-      theCase.id,
-      defendant.id,
-      update,
-    )
+    const updatedDefendant = await this.sequelize.transaction(
```
gudjong, line 304: The defendant update and the event log creation should happen in the same database transaction.

apps/judicial-system/backend/src/app/modules/defendant/limitedAccessDefendant.controller.ts
```diff
@@ -54,11 +54,6 @@ export class LimitedAccessDefendantController {
     this.logger.debug(
       `Updating limitedAccess defendant ${defendantId} of case ${caseId}`,
     )
-    return this.defendantService.updateRequestCaseDefendant(
-      theCase,
-      defendant,
-      updateDto,
-      user,
-    )
+    return this.defendantService.update(theCase, defendant, updateDto, user)
```
gudjong, line 57: This method can probably be limited to indictments only, but at least it should not use the request case update. Also, it would probably be better to limit the update to prison admin staff, but we do not have a roles rule to support at the moment.

apps/judicial-system/backend/src/app/modules/defendant/models/defendantEventLog.model.ts
```diff
@@ -20,14 +20,33 @@ import { Defendant } from './defendant.model'
...
+  ): DefendantEventLog | undefined {
+    if (!eventLogs) {
+      return undefined
+    }
+
+    const eventTypes = Array.isArray(eventType) ? eventType : [eventType]
+    const relevantLogs = eventLogs
+      .filter((eventLog) => eventTypes.includes(eventLog.eventType))
+      // Don't assume any ordering
+      .sort((a, b) => b.created.getTime() - a.created.getTime())
```
gudjong, line 35: Sorting is not done here rather than in database queries.

apps/judicial-system/backend/src/app/modules/event-log/eventLog.service.ts
```diff
@@ -72,28 +74,24 @@ export class EventLogService {
...
     if (!allowMultiple.includes(event.eventType)) {
       const where = Object.fromEntries(
         // The user name and title do not matter when checking for duplicates
         Object.entries({
           eventType,
           caseId,
-          nationalId,
-          userRole,
-          institutionName,
+          userRole: allowOnePerUserRole.includes(eventType)
```
gudjong, line 85: The logic for one-time events was wrong. We should not allow duplicate events except is special cases, such as when logging appeal result access per user role.

apps/judicial-system/backend/src/app/modules/event-log/eventLog.service.ts
```diff
@@ -72,28 +74,24 @@ export class EventLogService {
...
+            ? userRole
+            : undefined,
         }).filter(([_, value]) => value !== undefined),
       )
 
-      const eventExists = await this.eventLogModel.findOne({ where })
+      const eventExists = await this.eventLogModel.findOne({
+        where,
+        transaction,
```
gudjong, line 93: We want to use the same transaction here as in the event log creation to prevent accidental creation of duplicate events.

apps/judicial-system/web/src/components/Table/Table.tsx
```diff
@@ -97,24 +93,6 @@ const Table: FC<TableProps> = (props) => {
     }
   }
 
-  const renderProsecutorText = (
```
gudjong, line 100: This component will be removed when the defender tables have been moved to the new table architecture. This PR cleans up some unused code.

apps/judicial-system/web/src/components/DeprecatedContextMenuOptions/WithdrawAppealMenuOption.tsx
```diff
@@ -1,11 +1,7 @@
-import { FC, useCallback, useContext, useState } from 'react'
+import { FC, useState } from 'react'
 import { useIntl } from 'react-intl'
 
 import { IconMapIcon } from '@island.is/island-ui/core'
-import {
```
gudjong, line 5: This component is only used in the defender tables and will be removed when the defender tables have been moved to the new table architecture. This PR removes some unnecessary code.

apps/judicial-system/web/src/routes/Shared/Cases/MobileCase.tsx
```diff
@@ -80,11 +78,7 @@ const MobileCase: FC<PropsWithChildren<Props>> = ({
       <Text title={theCase.policeCaseNumbers?.join(', ')}>
         {displayFirstPlusRemaining(theCase.policeCaseNumbers)}
       </Text>
-      {theCase.courtCaseNumber && (
-        <Text>{`${courtAbbreviation ? `${courtAbbreviation}: ` : ''}${
-          theCase.courtCaseNumber
-        }`}</Text>
-      )}
+      {theCase.courtCaseNumber && <Text>theCase.courtCaseNumber</Text>}
```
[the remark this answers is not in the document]
gudjong: Good catch. My mistake. I have fixed this.

apps/judicial-system/backend/src/app/modules/case-table/caseTable.cellGenerators.ts
```diff
@@ -718,9 +717,12 @@ const caseSentToCourtDate: CaseTableCellGenerator<StringValue> = {
     },
   },
   generate: (c: Case): CaseTableCell<StringValue> => {
-    const caseSentToCourtEvent = EventLog.caseSentToCourtEvent(c.eventLogs)
+    const caseSentToCourtDate = EventLog.getEventLogDateByEventType(
+      [EventType.CASE_SENT_TO_COURT, EventType.INDICTMENT_CONFIRMED],
```
thorhildurt, line 721: Why do we need to also need to send "indictment confirmed" here?

apps/judicial-system/backend/src/app/modules/case/case.service.ts
thorhildurt: Why not? Is it not part of another update?

thorhildurt: Ah or is it just changing the case to be update in memory prior to the actual database update?

apps/judicial-system/backend/src/app/modules/case/case.service.ts
thorhildurt: What was the issue? Was it that it was always set to the current date?

apps/judicial-system/backend/src/app/modules/case/interceptors/case.interceptor.ts
```diff
@@ -90,8 +92,33 @@ const transformCase = (theCase: Case) => {
...
+      theCase.eventLogs,
+    ),
+    indictmentSentToPublicProsecutorDate: EventLog.getEventLogDateByEventType(
+      EventType.INDICTMENT_SENT_TO_PUBLIC_PROSECUTOR,
+      theCase.eventLogs,
+    ),
+    defenceAppealResultAccessDate: EventLog.getEventLogDateByEventType(
+      EventType.APPEAL_RESULT_ACCESSED,
+      theCase.eventLogs,
+      UserRole.DEFENDER,
```
thorhildurt, line 110: Here we are passing a static user role although another user roles might be triggering the case fetch which includes the interceptor logic. 

Why is then the user role needed? Just to showcase that a given date is only relevant for a given user role?

apps/judicial-system/backend/src/app/modules/defendant/limitedAccessDefendant.controller.ts
thorhildurt: So it was just wrongly named `updateRequestCaseDefendant`? or was it ever used to update requests?

apps/judicial-system/backend/src/app/modules/defendant/models/defendantEventLog.model.ts
thorhildurt: Wait, why are we then sorting? Cant we just remove the sort?

thorhildurt: Aren´t we using the sort to set the order to then return the latest log of the returned `relevantLogs`? Maybe I am just misunderstanding the comment 😇  So no sorting had been completed when we call this function and thus we should not assume we have a certain ordering and thus we sort the event logs here 👍

apps/judicial-system/backend/src/app/modules/event-log/models/eventLog.model.ts
```diff
@@ -18,12 +18,39 @@ import { Case } from '../../case/models/case.model'
...
+    if (!eventLogs) {
+      return undefined
+    }
+
+    const eventTypes = Array.isArray(eventType) ? eventType : [eventType]
+    const relevantLogs = eventLogs
+      .filter(
+        (eventLog) =>
+          eventTypes.includes(eventLog.eventType) &&
+          (!userRole || eventLog.userRole === userRole),
```
thorhildurt, line 35: What is the purpose of userRole in the context of events? is is access control or just to indicate that a given event is only relevant for a certain role?

apps/judicial-system/backend/src/app/modules/event-log/models/eventLog.model.ts
```diff
@@ -18,12 +18,39 @@ import { Case } from '../../case/models/case.model'
   timestamps: false,
 })
 export class EventLog extends Model {
-  static caseSentToCourtEvent(eventLogs?: EventLog[]) {
-    return eventLogs?.find(
-      (eventLog) =>
-        eventLog.eventType === EventType.CASE_SENT_TO_COURT ||
-        eventLog.eventType === EventType.INDICTMENT_CONFIRMED,
-    )
+  static getEventLogByEventType(
```
thorhildurt, line 21: NIT: Apart from the user role, this helper identical to the helper in `DefendantEventLog`. Is it possible to extract shared logic to a suitable place in the code? where we can also add tests?

apps/judicial-system/backend/src/app/modules/case/interceptors/case.interceptor.ts
thorhildurt: Ah I think I get it now, is just because we have the samt event type for several roles and we use it as a way to distinguish a given type 👍

apps/judicial-system/backend/src/app/modules/event-log/models/eventLog.model.ts
thorhildurt: Pretty certain is it the latter as I also answered to myself above 😁 "It is just because we have the samt event type for several roles and we use it as a way to distinguish a given type 👍"

thorhildurt (approved):
Great work to move those to the backend only 💪 I will approve as I have mostly clarifying questions and minor nits.

apps/judicial-system/backend/src/app/modules/case/case.service.ts
gudjong: Yes, what I should have said is that this code is not adding to the current database transaction. So, just like other code above that modifies the `update` object, it does not need to be inside the transaction itself.

apps/judicial-system/backend/src/app/modules/defendant/limitedAccessDefendant.controller.ts
gudjong: The name was correct. There is another method called `upateIndictmentCaseDefendant` that should be used instead of the `updateRequestCaseDefendant`. The code behaved `ok` because none of the side effects associated with either method was or would have been triggered.

gudjong: Since this controller endpoint is not limited to indictment cases, I use the `update` method, which in turn calls the correct update method depending on the case type. However, as limited access users do not make changes to request cases it would be better to limit the endpoint to indicment cases. I did not make that change in this PR.

apps/judicial-system/backend/src/app/modules/defendant/models/defendantEventLog.model.ts
gudjong: The sorting is needed, because we are not assuming any ordering. Previously, we assumed that the event log was sorted by cerated in a descending order.

gudjong: So, instead of assuming the order, we enforce it here 😊

apps/judicial-system/backend/src/app/modules/case-table/caseTable.cellGenerators.ts
gudjong: This is the way `caseSentToCourtEvent` was implemented, so I am simply preserving the previous behaviour.

The reason, however, that both are need is that one event is used for request cases and the other one for indictment cases. Both give the date when the case was sent to the court. I am not sure why different events are used, but there could be some historical reason for it.

apps/judicial-system/backend/src/app/modules/case/interceptors/case.interceptor.ts
gudjong: Exactly, this is just getting the time a defender accessed the appeal result, if any.

apps/judicial-system/backend/src/app/modules/event-log/models/eventLog.model.ts
gudjong: Actually, the user role has a dual purpose:

1. Just like any other additional into, user national id,  name, title and institution name, it is informative. That is, gives us information about who triggered the event.
2. For some events (currently only `APPEAL_RESULT_ACCESSED`) duplicate events are not allowed unless a new event is triggered by a role which has not previously triggered that same event for the case.

apps/judicial-system/backend/src/app/modules/event-log/models/eventLog.model.ts
gudjong: Actually, the two methods handle different enum types, so I don't think it would be a good idea to mix the two.

gudjong:
@thorhildurt great review and good questions. I hope I have answered all your concerns.

PR #19671, opened 2025-07-28
feat(core): Trace SID in logs, APM, settings and problems

eirikurn:
## What

Implement a tracing session id which can be used to diagnose user issues in 3rd level support.

This "traceSid" is a SHA 256 digest of the authentication session id (sid) + the current date. It is shown to authenticated users in these places:

- In the "My settings" page with a copy button.
- When they encounter issues which render a large Problem component in my pages and the admin portal.
- Later: In settings in the island.is app.
- Later: When they encounter problems in the application system.
- Maybe later: Somewhere in admin portal.

If they share this traceSid with the Ísland.is service desk (eg in a screenshot), it can be used to look up the following things in Datadog:

- Basic information about all authenticated API requests associated with this traceSid (method, URL, user agent).
- Backend logs from those authenticated requests (info, warnings, errors).
- APM traces for some of those authenticated requests (spans, timings, errors).

This PR also refactors the Tooltip component to @island.is/react-spa/shared and implements a CopyButton for reuse.

## Why

To help developers troubleshoot problems that occur in production.

## Screenshots / Gifs

image
image
image
image

## Summary by CodeRabbit

- **New Features**
  - Error pages and My Pages show a per-user debug tracking code (traceSid) with a one-click copy button.
  - Problem components accept and display optional debug text (falls back to the user trace code).

- **UX**
  - Added reusable Copy and Tooltip components for consistent copy/tooltip interactions.

- **Style**
  - Empty-state component no longer forces a large size; it respects caller-provided sizing.


andes-it:
This PR has been inactive for over a week. The `deploy-feature` label has been removed. Re-add the label to re-enable automatic deploy previews.

lodmfjord (approved):
LGTM, looking forward to seeing the results on the dev before moving to prod.

thorkellmani (approved):
Yeah, this is nice. Should be very helpful.

However, I don't think that the SID in the `<Problem />` is accessible enough. The gray-on-white color scheme is kinda hard to spot if you don't know it's there. Could the contrast be increased?

disaerna (approved):
Island UI ✅

eirikurn:
> Yeah, this is nice. Should be very helpful.
> 
> However, I don't think that the SID in the `<Problem />` is accessible enough. The gray-on-white color scheme is kinda hard to spot if you don't know it's there. Could the contrast be increased?

Good point. The idea was that it is not super relevant to the user / decorative, so it doesn't require the same contrast. But we may go a different route after talking with legal.

PR #19686, opened 2025-07-30
feat)j-s): Order Numbers from LÖKE

gudjong:
# Order Numbers from LÖKE

[Raða númerum frá LÖKE í tímaröð á skjánum "Ákærði"](https://app.asana.com/1/203394141643832/project/1199153462262248/task/1210917200360983)

## What

- Orders police case numbers retrieved from LÖKE by date.

## Why

- Improved UX.

## Screenshots / Gifs

image

## Checklist:

### DIFF
--- a/apps/judicial-system/web/src/routes/Prosecutor/Indictments/Defendant/Defendant.tsx
+++ b/apps/judicial-system/web/src/routes/Prosecutor/Indictments/Defendant/Defendant.tsx
@@ -1,4 +1,4 @@
-import { useCallback, useContext, useEffect, useState } from 'react'
+import { useCallback, useContext, useMemo } from 'react'
 import { useIntl } from 'react-intl'
 import { AnimatePresence, motion } from 'motion/react'
 import { useRouter } from 'next/router'
@@ -105,6 +105,7 @@ const getPoliceCasesForUpdate = (
   )
 
 const Defendant = () => {
+  const router = useRouter()
   const { user } = useContext(UserContext)
   const { workingCase, setWorkingCase, isLoadingWorkingCase, caseNotFound } =
     useContext(FormContext)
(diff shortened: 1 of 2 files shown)


apps/judicial-system/web/src/routes/Prosecutor/Indictments/Defendant/Defendant.tsx
```diff
@@ -160,8 +156,26 @@ const Defendant = () => {
...
+  const policeCaseInfo = useMemo(() => {
+    if (!data || !data.policeCaseInfo || data.policeCaseInfo.length === 0) {
+      return []
+    }
+
+    const [first, ...rest] = data.policeCaseInfo
+
+    const sortedRest = rest.sort((a, b) => {
+      // We want info with missing dates to be at the end of the list
+      const aDate = a.date ?? '9'
```
thorhildurt, line 168: Why '9'? 🤔

apps/judicial-system/web/src/routes/Prosecutor/Indictments/Defendant/Defendant.tsx
```diff
@@ -160,8 +156,26 @@ const Defendant = () => {
...
   })
 
-  const handleCreatePoliceCase = async (policeCaseInfo?: PoliceCase) => {
-    const newPoliceCase = policeCaseInfo ?? { number: '' }
+  const policeCaseInfo = useMemo(() => {
+    if (!data || !data.policeCaseInfo || data.policeCaseInfo.length === 0) {
+      return []
+    }
+
+    const [first, ...rest] = data.policeCaseInfo
```
thorhildurt, line 164: Why do we store the first element in the `policeCaseInfo` array specifically? Why not just sort the entire list?

apps/judicial-system/web/src/routes/Prosecutor/Indictments/Defendant/Defendant.tsx
thorhildurt: NIT: What using dates instead of relying on string sort?

```
  const aDate = a.date ? new Date(a.date) : new Date()
  const bDate = b.date ? new Date(b.date) : new Date()

  // ascending order, such that the missing date (set to today) will be at the end
  return aDate.getTime() - bDate.getTime()
```

thorhildurt (approved):

thorhildurt (approved):
Approved with minor questions and suggestion

apps/judicial-system/web/src/routes/Prosecutor/Indictments/Defendant/Defendant.tsx
gudjong: We could use date, but there are two reasons for not doing it:
1. Using date would be more expensive.
2. Using new Date is not completely reliable in the front end. The user's local clock could affect the ordering. We have been trying to eliminate the use of new Date in the front end, but we are not fully there yet.
In any case, using '9' has the desired effect.

apps/judicial-system/web/src/routes/Prosecutor/Indictments/Defendant/Defendant.tsx
gudjong: That's a good question. The first police case number has a special meaning in LÖKE and for us as well. It is the police case number that all other police case numbers are joined under and should always be the first police case number.

However, this number will already be the first police case number in RVG, so it will never be shown in the list of not-yet-added police case numbers. So, we could just as well order all the numbers. I am going to reconsider this in a follow up PR.

PR #19699, opened 2025-08-01
chore(j-s): statistics return the data constraints

thorhildurt:
# [Statistics return the data set limits](https://app.asana.com/1/203394141643832/project/1199153462262248/task/1210917143848777)

## What
- Return the dataset limits for request, indictments and subpoenas based on the creation date

## Why
- To make it more visible to the users that each data set has different ranges

## Screenshots / Gifs
https://github.com/user-attachments/assets/b37df72b-e55a-4084-ba47-9c52b362b619

## Checklist:

## Summary by CodeRabbit

- New Features
  - Statistics for requests, indictments, and subpoenas now include a minimum available date (minDate).
  - Date filters are initialized and constrained by minDate, ensuring selections stay within available data.
  - Initial minDate is auto-fetched and applied across statistics views for consistent filtering.

- Refactor
  - Statistics pages split into lightweight wrappers and body components to fetch and pass minDate while preserving existing behavior.


apps/judicial-system/web/src/routes/Admin/Statistics/new/RequestStatistics.tsx
```diff
@@ -108,11 +108,27 @@ const RequestStatistics = () => {
...
       </Box>
     </StatisticPageLayout>
   )
 }
 
+const RequestStatistics = () => {
+  // We extract the initial call to fetch the request statistics data to a specific parent component
+  // to fetch defined statistical constraints (minDate) once. The child component is
+  // currently re-rendered on each filter change.
+  const { data } = useRequestCaseStatisticsQuery({
```
thorhildurt, line 123: Note: Currently we use the same query to fetch constraints and aggregations, thus when fetching the `minDate` here we are also triggering an aggregation but not referencing those results. We then make a second call in the Body child component below.

It would make more sense to split those up to make this more efficient but to avoid creating bunch of new code while our statistics is a bit early stage and might change depending on feedback, I decided to re-use the current query for simplicity. 

But please lmk if you have other thoughts on this.

oddsson: That sounds reasonable to me. Could you please create a ticket to refactor this when the time is right?

thorhildurt: Given the work happening in the current statistical iteration, I will see if it makes sense to refactor this and will create ticket if that makes sense when wrapping up the current work I am doing.

